PR #10425, opened 2022-09-28
DAOS-11534 build: add -fpic -pie in linkflag for program

wiliamhuang:
Per SDL requirement, "-pie" is added for linking C executable files. "-buildmode=pie" is added for linking Go executable files.

Required-githooks: true

Signed-off-by: Lei Huang <lei.huang@intel.com>

### DIFF
--- a/site_scons/daos_build.py
+++ b/site_scons/daos_build.py
@@ -154,6 +154,7 @@ def library(env, *args, **kwargs):
 def program(env, *args, **kwargs):
     """build Program with relative RPATH"""
     denv = env.Clone()
+    denv.AppendUnique(LINKFLAGS=['-fpic','-pie'])
     denv.Replace(RPATH=[])
     add_rpaths(denv, kwargs.get('install_off', '..'), False, True)
     prog = denv.Program(*args, **kwargs)
--- a/site_scons/daos_build.py
+++ b/site_scons/daos_build.py
@@ -154,7 +154,7 @@ def library(env, *args, **kwargs):
 def program(env, *args, **kwargs):
     """build Program with relative RPATH"""
     denv = env.Clone()
-    denv.AppendUnique(LINKFLAGS=['-fpic', '-pie'])
+    denv.AppendUnique(LINKFLAGS=['-pie'])
     denv.Replace(RPATH=[])
     add_rpaths(denv, kwargs.get('install_off', '..'), False, True)
     prog = denv.Program(*args, **kwargs)
@@ -167,6 +167,7 @@ def program(env, *args, **kwargs):
 def test(env, *args, **kwargs):
     """build Program with fixed RPATH"""
     denv = env.Clone()
+    denv.AppendUnique(LINKFLAGS=['-pie'])
     denv.Replace(RPATH=[])
     add_rpaths(denv, kwargs.get("install_off", None), False, True)
     testbuild = denv.Program(*args, **kwargs)
--- a/site_scons/daos_build.py
+++ b/site_scons/daos_build.py
@@ -154,7 +154,7 @@ def library(env, *args, **kwargs):
 def program(env, *args, **kwargs):
     """build Program with relative RPATH"""
     denv = env.Clone()
-    denv.AppendUnique(LINKFLAGS=['-fpic','-pie'])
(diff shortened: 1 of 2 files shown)


jolivier23:
Just use build as the title component.

wiliamhuang:
> Just use build as the title component.

Thanks, @jolivier23 ! Just updated it.

site_scons/daos_build.py
```diff
@@ -154,6 +154,7 @@ def library(env, *args, **kwargs):
 def program(env, *args, **kwargs):
     """build Program with relative RPATH"""
     denv = env.Clone()
+    denv.AppendUnique(LINKFLAGS=['-fpic','-pie'])
```
jolivier23, line 157: You need to fix the pylint warning here.  

Do we also need to add these to CCFLAGS

jolivier23 (changes requested):

site_scons/daos_build.py
wiliamhuang: CCFLAGS includes "-fpic" already. I am going to fix the pylint warning soon. Thank you!

wangshilong (dismissed):
Learn a bit from these two options, looks reasonable.

site_scons/daos_build.py
```diff
@@ -154,6 +154,7 @@ def library(env, *args, **kwargs):
 def program(env, *args, **kwargs):
     """build Program with relative RPATH"""
     denv = env.Clone()
+    denv.AppendUnique(LINKFLAGS=['-fpic', '-pie'])
```
jolivier23, line 157: maybe move this to site_scons/compiler_setup.py in base_setup ?

jolivier23 (changes requested):
See if you can move it to base_setup()

wiliamhuang:
> See if you can move it to base_setup()

@jolivier23 Flag "-pie" only works for linking executable files. It is not applicable for linking a shared library file. That's why I put it in program() only.

jolivier23:
> > See if you can move it to base_setup()
> 
> @jolivier23 Flag "-pie" only works for linking executable files. It is not applicable for linking a shared library file. That's why I put it in program() only.

Do you need fpic there?

wiliamhuang:
> Do you need fpic there?

Yes. "-pie" requires "-fpic".

wiliamhuang:
> Do you need fpic there?

I was wrong last time. "-fpic" is required when compiling and linking in one command. In our case, "-fpic" is used to compile source code and generate object files. It is not required in linking step. With and without "-fpic" make no difference here.

jolivier23 (dismissed):

soumagne (dismissed):

src/control/SConscript
```diff
@@ -118,8 +118,13 @@ def install_go_bin(denv, gosrc, libs, name, install_name):
     # Propagate useful GO environment variables from the caller
     if 'GOCACHE' in os.environ:
         denv['ENV']['GOCACHE'] = os.environ['GOCACHE']
+    if not is_release_build(denv):
+        flag_pie = ' '
+    else:
+        flag_pie = '-buildmode=pie '
```
ashleypittman, line 124: Any reason to make this conditional?  Why wouldn't we always want a pie build?

ashleypittman (commented):
Hard to believe we missed this.

src/control/SConscript
wiliamhuang: https://github.com/daos-stack/daos/blob/4829d750ee7e896304ba1751158358d12fc2b919/src/control/SConscript#L65-L67
"-race" is used for Go linking in debug mode. "-race" and "-buildmode=pie" can not be used at the same time, so I made it conditional, turning on "-buildmode=pie" only for non-debug mode.

mjmac: I agree with ashley that it would be better to add this to `get_build_flags()`. You could convert it to the same style as `get_build_tags()`, where a list of strings is built and then a single string joined by `' '` is returned. That would be cleaner.

ashleypittman: Thank you, that's much clearer about what is happening and why now.

ashleypittman (approved):

PR #10432, opened 2022-09-28
DAOS-623 build: Add SCONS_ENV option

jolivier23:
Allow one to propagate external environment with SCONS_ENV=user

Required-githooks: true

### DIFF
--- a/site_scons/prereq_tools/base.py
+++ b/site_scons/prereq_tools/base.py
@@ -600,16 +600,22 @@ class PreReqComponent():
         self.__opts = variables
         self.configs = None
 
-        real_env = self.__env['ENV']
+        self.add_opts(EnumVariable('SCONS_ENV', "Default SCons environment mode",
+                                   'default', ['default', 'user'], ignorecase=1))
 
-        for var in ["HOME", "TERM", "SSH_AUTH_SOCK",
-                    "http_proxy", "https_proxy",
-                    "PKG_CONFIG_PATH", "MODULEPATH",
-                    "MODULESHOME", "MODULESLOADED",
-                    "I_MPI_ROOT"]:
-            value = os.environ.get(var)
-            if value:
-                real_env[var] = value
+        if self.__env.get("SCONS_ENV") == 'user':
+            real_env = os.environ
+        else:
+            real_env = self.__env['ENV']
+
+            for var in ["HOME", "TERM", "SSH_AUTH_SOCK",
+                        "http_proxy", "https_proxy",
+                        "PKG_CONFIG_PATH", "MODULEPATH",
+                        "MODULESHOME", "MODULESLOADED",
+                        "I_MPI_ROOT"]:
+                value = os.environ.get(var)
+                if value:
+                    real_env[var] = value
 
         libtoolize = 'libtoolize'
         if self.__env['PLATFORM'] == 'darwin':


daltonbohning (approved):
LGTM, and even wonder if the default should be `user`. Though that would change the default behavior

jolivier23:
> LGTM, and even wonder if the default should be `user`. Though that would change the default behavior

I just modifed it...Let's see what happens in CI with the different default.

daltonbohning (approved):

soumagne (approved):

site_scons/prereq_tools/base.py
```diff
@@ -600,16 +600,22 @@ def __init__(self, env, variables, config_file=None, arch=None):
...
-        for var in ["HOME", "TERM", "SSH_AUTH_SOCK",
-                    "http_proxy", "https_proxy",
-                    "PKG_CONFIG_PATH", "MODULEPATH",
-                    "MODULESHOME", "MODULESLOADED",
-                    "I_MPI_ROOT"]:
-            value = os.environ.get(var)
-            if value:
-                real_env[var] = value
+        self.add_opts(EnumVariable('SCONS_ENV', "Default SCons environment inheritance",
+                                   'minimal', ['minimal', 'full'], ignorecase=1))
```
ashleypittman, line 604: ignorecase is interesting, should we be using that elsewhere?

site_scons/prereq_tools/base.py
```diff
@@ -600,16 +600,22 @@ def __init__(self, env, variables, config_file=None, arch=None):
...
+
+        if self.__env.get("SCONS_ENV") == 'full':
```
ashleypittman, line 606: I wonder if we should move the VIRTUAL_ENV code here from https://github.com/daos-stack/daos/blob/master/SConstruct#L349-L352 then you could have 'minimal', 'venv', 'full' with the default being venv.  We also have some special handling of PATH for mpicc elsewhere in the tree

ashleypittman (commented):
Another option would be to check for $COMPILER in os.environ['PATH'] and env['ENV']['PATH'] and if the results are different then warn here.

ashleypittman:
There's also https://github.com/daos-stack/daos/blob/master/SConstruct#L331-L338 to look at as well.

daltonbohning:
What's left to do with this one? FWIW, `SCONS_ENV=full` works for my purposes on Frontera, where gcc, cmake, and some others are in non-standard paths.

ashleypittman:
For what it's worth I've found out how to do this with .files in a users home directory if that would be preferable.

daltonbohning:
> For what it's worth I've found out how to do this with .files in a users home directory if that would be preferable.

I'd rather have an option than define something external before building. But what is the general flow of doing this with `.files`?

daltonbohning:
> The page I was working from is this one. https://steelpangolin.wordpress.com/2014/02/06/modifying-scons-environment-variables-globally/

As simple as that approach is, I think this PR is less error-prone

daltonbohning:
@jolivier23 @ashleypittman Could we get an updated version of this PR, please? After the latest scons changes this check is now failing:
https://github.com/daos-stack/daos/blob/aa64f5092d4c3b2107bd38c392c9a9437c322c33/SConstruct#L348-L349

daltonbohning:
> @jolivier23 @ashleypittman Could we get an updated version of this PR, please? After the latest scons changes this check is now failing:

From @ashleypittman, we need init the PreReqComponent before the stdatomic check
https://github.com/daos-stack/daos/blob/f99382ba1006ada422ab84d03a6808c38131c7d5/SConstruct#L347-L352

mchaarawi (approved):

daltonbohning:
FYI that @ashleypittman this change in #11291. Maybe we can close this PR?

mchaarawi:
> FYI that @ashleypittman this change in #11291. Maybe we can close this PR?

that PR is still a draft? this looks ready to land :-)

ashleypittman:
> > FYI that @ashleypittman this change in #11291. Maybe we can close this PR?
> 
> that PR is still a draft. this looks ready to land :-)

I'm fixing some other issues at the same time so it's still in testing.

For what it's worth both PRs add the SCONS_ENV option (I copied the code from here) however that option is still not enough to solve the initial problem however it is a step forwards.

daltonbohning (approved):

ashleypittman:
I think you could close this now.

PR #10433, opened 2022-09-28
DAOS-11706 control: Add pool svc list to property query

mjmac:
The internal handlers already support it; we just need to
expose the pool replicas as a ranklist string.

Moves the pool property handling code from lib/control to
lib/daos, as this is a better home for it.


src/mgmt/srv_drpc.c
```diff
@@ -1147,8 +1146,23 @@ add_props_to_resp(daos_prop_t *prop, Mgmt__PoolGetPropResp *resp)
 			if (resp_props[j]->strval == NULL)
 				D_GOTO(out, rc = -DER_NOMEM);
 		} else if (daos_prop_has_ptr(entry)) {
-			D_ERROR("pointer-value props not supported\n");
-			D_GOTO(out, rc = -DER_INVAL);
+			switch (entry->dpe_type) {
+			case DAOS_PROP_PO_SVC_LIST:
+				if (entry->dpe_val_ptr == NULL) {
+					D_ERROR("svc rank list unset\n");
+					D_GOTO(out, rc = -DER_INVAL);
```
tanabarr, line 1153: nit: I guess I hear conflicting advice about the use of D_GOTO, I thought the current guidance was to avoid using it.

tanabarr (approved):

kccain (approved):
It might be good in this change or a future one to (in daos_test pool_properties() after the daos_pool_query() call) inspect the DAOS_PROP_PO_SVC_LIST entry.

src/mgmt/srv_drpc.c
kjacque: I was never totally clear on that myself. I've been continuing to use it where it's convenient.

PR #10436, opened 2022-09-29
DAOS-11715 pool: worked around a coverity issue

wiliamhuang:
To work around a coverity issue: the return value rc of rdb_remove_replicas is not used/checked.

Required-githooks: true

Signed-off-by: Lei Huang <lei.huang@intel.com>


daltonbohning (approved):

jolivier23 (approved):
More like "worked around a coverity issue" since it's not really a fix to just print the error.

jolivier23:
I take it by adding gatekeeper that this should no longer be a draft?

PR #10437, opened 2022-09-29
DAOS-623 engine: Change D_FREE not to log NULL.

ashleypittman:
This caused a number of compile warnings so rework code to avoid
these.

### DIFF
--- a/.github/workflows/landing-builds.yml
+++ b/.github/workflows/landing-builds.yml
@@ -17,7 +17,6 @@ on:
 
 jobs:
 
-
   # Build a base Docker image, and save it with a key based on the hash of the dependencies, and a
   # copy of the git hash.  Do not attempt to load the cache here, but rather start each build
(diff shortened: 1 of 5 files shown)


.github/workflows/landing-builds.yml
```diff
@@ -17,6 +17,7 @@ on:
 
 jobs:
 
+
```
jolivier23, line 20: Is this one intentional?

jolivier23 (approved):

.github/workflows/landing-builds.yml
ashleypittman: Ah yes, that was to trigger the landings-builds to run so we get full distro coverage to be sure there are no undetected compile errors introduced.  I'll back it out once the GHA just have completed.

frostedcmos (approved):

soumagne (approved):

PR #10438, opened 2022-09-29
DAOS-10663 dfs: add new DFS readdirplus API

mchaarawi:
Same as readdir but also does a stat on the entry for convenience.

Required-githooks: true

### DIFF
--- a/src/client/dfs/dfs.c
+++ b/src/client/dfs/dfs.c
@@ -3447,8 +3447,8 @@ dfs_lookup(dfs_t *dfs, const char *path, int flags, dfs_obj_t **_obj,
 }
 
 int
-dfs_readdir(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
-	    struct dirent *dirs)
+readdir_int(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
+	    struct dirent *dirs, struct stat *stbufs)
 {
 	daos_key_desc_t	*kds;
 	char		*enum_buf;
@@ -3500,6 +3500,18 @@ dfs_readdir(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
 				       kds[i].kd_key_len + 1, "%s", ptr);
 			D_ASSERT(len >= kds[i].kd_key_len);
 			ptr += kds[i].kd_key_len;
+
+			/** stat the entry if requested */
+			if (stbufs) {
+				rc = entry_stat(dfs, DAOS_TX_NONE, obj->oh, dirs[key_nr].d_name,
+						kds[i].kd_key_len, NULL, true, &stbufs[key_nr],
+						NULL);
+				if (rc) {
+					D_ERROR("Failed to stat entry %s: %d (%s)\n",
+						dirs[key_nr].d_name, rc, strerror(rc));
+					D_GOTO(out, rc);
+				}
+			}
 			key_nr++;
 		}
 		number = *nr - key_nr;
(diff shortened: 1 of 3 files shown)


src/include/daos_fs.h
```diff
@@ -602,8 +602,32 @@ dfs_punch(dfs_t *dfs, dfs_obj_t *obj, daos_off_t offset, daos_size_t len);
...
+ *
+ * \param[in]	dfs	Pointer to the mounted file system.
+ * \param[in]	obj	Opened directory object.
+ * \param[in,out]
+ *		anchor	Hash anchor for the next call, it should be set to
+ *			zeroes for the first call, it should not be changed
+ *			by caller between calls.
+ * \param[in,out]
+ *		nr	[in]: number of dirents allocated in \a dirs.
+ *			[out]: number of returned dirents.
```
ashleypittman, line 618: [nit] This isn't strictly correct although the intent is clear.

src/client/dfs/dfs.c
```diff
@@ -3447,8 +3447,8 @@ dfs_lookup(dfs_t *dfs, const char *path, int flags, dfs_obj_t **_obj,
 }
 
 int
-dfs_readdir(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
-	    struct dirent *dirs)
+readdir_int(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
```
ashleypittman, line 3450: You could also just call this dfs_readdirplus() and have dfs_readdir() call it with a NULL value for stbufs to avoid a second wrapper function later on.

src/client/dfs/dfs.c
```diff
@@ -3500,6 +3500,18 @@ dfs_readdir(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
 				       kds[i].kd_key_len + 1, "%s", ptr);
 			D_ASSERT(len >= kds[i].kd_key_len);
 			ptr += kds[i].kd_key_len;
+
+			/** stat the entry if requested */
+			if (stbufs) {
+				rc = entry_stat(dfs, DAOS_TX_NONE, obj->oh, dirs[key_nr].d_name,
+						kds[i].kd_key_len, NULL, true, &stbufs[key_nr],
+						NULL);
```
ashleypittman, line 3508: This function is inherently racy against different clients so it would be possible to define the API such that a return value indicated an error but if *nr was non-zero then at least that many dirents/stbufs are valid.  The anchor itself would be in a undefined state in that case however so I'm not sure how a user would recover from that case.

ashleypittman (approved):
Approved, but see comments.

src/client/dfs/dfs.c
mchaarawi: although we don't adhere to this much, but we shouldn't call API functions internally, in case we add some API tracing later.

src/client/dfs/dfs.c
mchaarawi: if i understand your comment, that's the model of the relaxed mode, where these races are acceptable. i would argue that this race is also acceptable in the balanced mode, but in that case we can used a dtx for the readdir and stat. in any case this isn't supported till daos 3.0 anyway.

ashleypittman: I was thinking if a race was encountered and the stat failed the API would be able to return *some* valid data but it seems a corner case.

mchaarawi: ok i see what you mean, and i think that's OK. indeed that a user cannot recover against a race error where the stat failed because the anchor and the number of items returned do not match. so either they have to save the initial passed in anchor and restart from that, or restart from the beginning.

src/client/dfs/dfs.c
```diff
@@ -3514,6 +3526,20 @@ dfs_readdir(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
...
+{
+	return readdir_int(dfs, obj, anchor, nr, dirs, NULL);
+}
+
+int
+dfs_readdirplus(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
+		struct dirent *dirs, struct stat *stbufs)
+{
+	return readdir_int(dfs, obj, anchor, nr, dirs, stbufs);
+}
```
daltonbohning, line 3541: Do we need a NULL check on `stbufs`?

mchaarawi: i made it as optionally populated if not NULL:
https://github.com/daos-stack/daos/pull/10438/files#diff-dd80cb8b1b118e01084005bad5c40fabe2a4b05744db070f7f3dc2142aa98c60R3505

daltonbohning: Right. I guess if it's NULL, it's the same as calling `dfs_readdir`

daltonbohning (approved):

johannlombardi (approved):

PR #10440, opened 2022-09-29
DAOS-11728 object: refine sgl iov_len handling in EC data recovery

liuxuezhao:
Fix a bug of after obj_ec_recov_data() did not set sgl's iov_len. 
Added a test case to verify it.

Required-githooks: true

### DIFF
--- a/src/common/misc.c
+++ b/src/common/misc.c
@@ -322,7 +322,7 @@ daos_sgl_get_bytes(d_sg_list_t *sgl, bool check_buf, struct daos_sgl_idx *idx,
 	daos_size_t len;
 
 	if (p_buf_len != NULL)
-	*p_buf_len = 0;
+		*p_buf_len = 0;
 
 	if (idx->iov_idx >= sgl->sg_nr)
 		return true; /** no data in sgl to get bytes from */
@@ -330,7 +330,9 @@ daos_sgl_get_bytes(d_sg_list_t *sgl, bool check_buf, struct daos_sgl_idx *idx,
 	len = check_buf ? sgl->sg_iovs[idx->iov_idx].iov_buf_len :
 		sgl->sg_iovs[idx->iov_idx].iov_len;
 
-	D_ASSERT(idx->iov_offset < len);
+	D_ASSERT(idx->iov_offset <= len);
+	if (len == 0)
+		goto next_iov;
 	/** Point to current idx */
 	if (p_buf != NULL)
 		*p_buf = sgl->sg_iovs[idx->iov_idx].iov_buf + idx->iov_offset;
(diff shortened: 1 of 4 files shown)


wangdi1 (approved):

src/common/misc.c
```diff
@@ -322,15 +322,17 @@ daos_sgl_get_bytes(d_sg_list_t *sgl, bool check_buf, struct daos_sgl_idx *idx,
...
+		*p_buf_len = 0;
 
 	if (idx->iov_idx >= sgl->sg_nr)
 		return true; /** no data in sgl to get bytes from */
 
 	len = check_buf ? sgl->sg_iovs[idx->iov_idx].iov_buf_len :
 		sgl->sg_iovs[idx->iov_idx].iov_len;
 
-	D_ASSERT(idx->iov_offset < len);
+	D_ASSERT(idx->iov_offset <= len);
```
NiuYawei, line 333: I think the 'iov_offset' should never equal to 'len' since we'll move to next iov once "iov_offset == len", right? (see following code in this function)

src/common/misc.c
```diff
@@ -322,15 +322,17 @@ daos_sgl_get_bytes(d_sg_list_t *sgl, bool check_buf, struct daos_sgl_idx *idx,
...
+	if (len == 0)
+		goto next_iov;
```
NiuYawei, line 335: so we choose to advance the iov_idx but don't set p_buf and return 0 p_buf_len to caller? and caller will have to call into this function again? I'm not sure if caller can handle this well or not, but it looks to me it's not necessary to implement such function in this way. (I think we'd always return a non-zero buf_len until reach the end)

src/common/misc.c
liuxuezhao: one possible case is that if user passed in sgl with one iov with zero length, then the idx->iov_offset possily == len == 0, it is extreme cornel case but seems possible.

src/common/misc.c
liuxuezhao: now the only user is daos_sgl_processor and do memcpy, for the case of 0 len it will set the buf_len = 0, so the memcpy will do nothing. that seems fine?

src/common/misc.c
NiuYawei: what's the use case of 0 length iov? I tend to think we'd assert iov length > 0 to avoid potential bug.

src/common/misc.c
NiuYawei: yes, if caller can handle this, it won't cause error. my point is why we want to make this function to return a 0 buf_len?

NiuYawei (approved):

wangdi1 (approved):

PR #10441, opened 2022-09-29
DAOS-11679 dtx: misc patch for dtx related issues

Nasf-Fan:
The patch fixes the following DTX related issues:

1. Use the address of tree handle for dtx_rpc_post.
   If we directly use the value of tree handle as dtx_rpc_post()'s
   input parameter, then it may be DAOS_HDL_INVAL before DTX helper
   ULT being scheduled. That will cause memory leak.

2. Abort remote DTX entry before local one.
   The DTX abort maybe triggered by dtx_leader_end for RPC timeout
   on some remote DTX participant(s). Under such case, client side
   RPC sponsor may also hit the RPC timeout and resends related RPC
   to the leader. To avoid DTX abort and resend RPC forwarding being
   executed in parallel, we will abort local DTX after remote done,
   before that the logic of handling resent RPC on server will find
   the local pinned DTX entry then notify related client to resend
   RPC sometime later.

3. More frequent CPU yield during DTX RPCs dispatch.
   It is possible that the DTX may contains a lot of participants.
   When we commit or abort such DTX, the leader needs to send RPCs
   to all related participant. That will take a lot of CPU cycles.
   Under such case, related DTX operation ULT will yield to avoid
   blocking others for too long time. The patch reduces the yield
   interval from per 64-RPCs to per 32-RPCs.

4. Dynamically load DTX participants information (dtx_memberships).
   For the transaction with very large participants information,
   handle related DTX (such as for DTX resync, for cleanup stale
   DTX entries) may take some time, expecially under the case of
   system very busy. If we pre-load all related MBS information
   in DRAM before really handle them, it will hold a lot of DRAM
   resource for long time. That may cause server OOM. The patch
   adjusts DTX MBS load policy as loading large MBS when use it
   instead of pre-loading.

5. Use absolute shard index to reassemble CPD RPC.
   The absolute shard index was changed accidentally to local index
   within single redundancy group when handle EC parity rotation.
   That will cause unexpected CPD RPC reassemble on leader. On the
   other hand, the reassemble logic on leader has some defect that
   may cause access DRAM out of boundary. The patch also fixes it.

### DIFF
--- a/src/dtx/dtx_common.c
+++ b/src/dtx/dtx_common.c
@@ -202,7 +202,13 @@ dtx_cleanup_stale_iter_cb(uuid_t co_uuid, vos_iter_entry_t *ent, void *args)
 	    DTX_CLEANUP_THD_AGE_LO)
 		return 1;
 
-	D_ALLOC(dsp, sizeof(*dsp) + ent->ie_dtx_mbs_dsize);
+	D_ASSERT(ent->ie_dtx_mbs_dsize > 0);
+
+	if (ent->ie_dtx_mbs_dsize > DTX_INLINE_MBS_SIZE)
+		D_ALLOC_PTR(dsp);
+	else
+		D_ALLOC(dsp, sizeof(*dsp) + sizeof(*mbs) + ent->ie_dtx_mbs_dsize);
+
 	if (dsp == NULL)
 		return -DER_NOMEM;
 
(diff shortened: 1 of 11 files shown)


Nasf-Fan:
Hit DAOS-11178, to be retested.

jolivier23 (commented):
Ready for review or should it be draft pr?

jolivier23 (approved):

wangdi1 (approved):

liuxuezhao (approved):

Nasf-Fan:
Ping @daos-stack/daos-gatekeeper , thanks!

PR #10442, opened 2022-09-29
DAOS-7821 cq: Fix a number of python flake issues.

ashleypittman:
Fix all flake issues in scons code and eradicate a number
of classes of flake error and add then to the GHA checking.
Update flake action to report all errors correctly.
Update pylint handling to work with new one-line files correctly.
Change pylint reporting to report invalid-name and f-string warnings last.


daltonbohning (approved):

otandka (approved):
Triaging the warnings is helpful.  I gather the goal is not to eliminate all the pylint warnings?

Good improvements on variable names.

It would be much easier to review if the non-functional changes were separated.  The diff is half the size using diff -w --ignore-blank-lines.  Most of the rest are line wrapping changes.

src/tests/ftest/pool/management_race.py
```diff
@@ -110,7 +111,8 @@ def test_pool_management_race(self):
                 del_recreate=False, test_pool=t_pool, thread_num=index, test_loop=test_loop)
         # Setup the thread manager for del_and_recreate_pool
         thread_manager.add(
-            del_recreate=True, test_pool=self.pool[pool_number], thread_num=index, test_loop=test_loop)
+            del_recreate=True, test_pool=self.pool[pool_number],
+            thread_num=index, test_loop=test_loop)
```
otandka, line 115: The pylint complaint here has some merit.  What is the expected value of index after the loop completes?  What if self.pool is empty.  Better would be an explicit value.
```suggestion
            thread_num=len(self.pool), test_loop=test_loop)
```

ashleypittman:
> Triaging the warnings is helpful. I gather the goal is not to eliminate all the pylint warnings?

The overall goal is to eliminate all pylint warnings, although large PRs that touch a number of files like this tend to be really difficult to get through test and approval and cause a lot of merge conflicts so for this PR I've taken the view of fixing only the "simple" issues that can be reviewed easily/quickly.  The attribute-outside-of-init and possibly-undefined-loop-variable issues are both likely bugs but they need to be resolved by people more familiar with those specific tests than I am, hence this PR does not try and address them in order to make reviewing easier/quicker.

Hopefully the "noise" of this style of PR will reduce soon and we can focus on the harder work of resolving the more complex issues.

> Good improvements on variable names.
> 
> It would be much easier to review if the non-functional changes were separated. The diff is half the size using diff -w --ignore-blank-lines. Most of the rest are line wrapping changes.

The functional changes should be limited the tools in this PR but point taken.  It's in fixing the code issues that we identify the tool issues.

otandka (commented):
Were most of these non-functional changes generated by running a formatting tool?

src/tests/ftest/control/dmg_server_set_logmasks.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/python
```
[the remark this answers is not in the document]
daltonbohning: > Are these only imported

Yes. We shouldn't have a shebang for non-executable files, and we shouldn't use `/usr/bin/python`. There are a lot of these remaining, which are slowly being cleaned up.

src/tests/ftest/pool/management_race.py
ashleypittman: I agree this complaint has merit and I hope that next time someone is working on this test with context of what it's trying to do that they address this.

jolivier23:
I pushed a new patch to resolve the merge conflict.  There is really no change here, just a file that we modified in this patch was removed in another commit.

daltonbohning (approved):

ashleypittman:
> Were most of these non-functional changes generated by running a formatting tool?

No, I started by running "flake8 --statistics" and picked the error clases with low numbers of errors, then for the classes I thought I could eliminate entirely I opened affected files in an editor and resolved them along with any other easy flake/pylint fixes for the same file and removed the #! when I remembered.  The other error clases either have many more instances to resolve or the changes required are more in-depth.

otandka (approved):

PR #10443, opened 2022-09-29
DAOS-11706 control: Move system.Rank* to lib/ranklist

mjmac:
The system package imposes a lot of extra dependencies on
parts of the code that just want to work with ranks or
ranklists. Having ranks in the system package also makes
import cycles more likely without extra work to avoid them.

Required-githooks: true


kjacque (approved):
Makes sense to me. There's always the tradeoff to be concerned with, breaking out small packages vs. trying to have larger packages. But I like this approach personally.

tanabarr (approved):

mjmac:
One unrelated failure: [DAOS-11749](https://daosio.atlassian.net/browse/DAOS-11749)

PR #10444, opened 2022-09-29
DAOS-11740 control: Reject invalid number of pool create ranks

mjmac:
If the pool create request has specified more ranks than are
available in the system, reject it.

Required-githooks: true


kjacque (approved):

ravalsam (approved):
Tested locally and it works fine.

PR #10445, opened 2022-09-30
DAOS-11664 agent: Add config option to ignore interfaces

kjacque:
In some environments clients may be connected to multiple networks that support the server's fabric provider, but some of them should not or cannot be used for DAOS. This new config option allows unwanted fabric interfaces to be excluded by name, rather than defining a full fabric_ifaces configuration. Excluded interfaces will not be suggested to clients.

### DIFF
--- a/src/control/cmd/daos_agent/config.go
+++ b/src/control/cmd/daos_agent/config.go
@@ -26,16 +26,17 @@ const (
 
 // Config defines the agent configuration.
 type Config struct {
-	SystemName       string                    `yaml:"name"`
-	AccessPoints     []string                  `yaml:"access_points"`
-	ControlPort      int                       `yaml:"port"`
-	RuntimeDir       string                    `yaml:"runtime_dir"`
-	LogFile          string                    `yaml:"log_file"`
-	LogLevel         common.ControlLogLevel    `yaml:"control_log_mask,omitempty"`
-	TransportConfig  *security.TransportConfig `yaml:"transport_config"`
-	DisableCache     bool                      `yaml:"disable_caching,omitempty"`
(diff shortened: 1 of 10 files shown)


mjmac (commented):
This accomplishes the goal, but I think it highlights some problems with the way the code is organized. After checking out the PR and looking it the changes in wider context, I understand why you're adding multiple `WithIgnoredDevices()` methods, but it doesn't seem like the ideal way to handle this.

I know it's a more invasive change, but to me it seems like it would be less error-prone from a code maintenance standpoint to pass the config around and delegate responsibility for interpreting the config to the receivers. If we were planning to extract these things for standalone use outside of the agent I could see a stronger case being made for keeping them purely functional and unaware of the config, but I don't think that's going to happen.

Another idea for centralizing responsibility would be to create a type alias for `common.StringSet` with a custom `UnmarshalYAML()` method so that you can do the parsing once. Or just add it onto `common.StringSet` directly, actually. Probably useful in other places. In general, we should be moving toward doing config/ui parsing as part of the receiver types directly rather than writing little variations on the same parsers all the time.

tanabarr (approved):

knard-intel (approved):
LGTM

mjmac (approved):

kjacque:
The failure is in a script vetting the commit title. There is a (new?) script checking component name in the commit message. I think "agent" should be an accepted component name, as we have used it often in the past. I'd like to change that in a separate PR though.

tanabarr (approved):

jolivier23:
This does not need forced landing.   The jira_query script error isn't fatal.

Right now, most of the components we accept are directories though we don't look into src/control/cmd (though in such case, it would be daos_agent rather than agent).  But you can also add agent to the list.

PR #10446, opened 2022-09-30
DAOS-10138 chk: Schedule PS reconf before DONE

liw:
Prevent automatic PS reconfigurations in the CR "check" mode, in order
to avoid racing with the chk module's actions. Schedule PS
reconfigurations when the chk module is ready before moving a pool to
the DONE phase.

Fix a segfault due to the use of an invalid svc variable when destroying
a stopped rsvc replica:

    pool_svc_locate_cb (id=0x18, path=0x7fa947c6d2e0) at
      src/pool/srv_pool.c:953
    ds_rsvc_stop (class=DS_RSVC_CLASS_POOL, id=0x7fa96c5f0860,
      caller_term=18446744073709551615, destroy=<optimized out>) at
      src/rsvc/srv.c:1009
    ds_rsvc_stop_handler (rpc=0x7fa96c5f0660) at src/rsvc/srv.c:1437
    crt_handle_rpc (arg=0x7fa96c5f0660) at src/cart/crt_rpc.c:1654

Required-githooks: true

### DIFF
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -5422,7 +5422,8 @@ pool_svc_schedule_reconf(struct pool_svc *svc, bool for_chk)
 	D_DEBUG(DB_MD, DF_UUID": begin\n", DP_UUID(svc->ps_uuid));
 
 	if (!for_chk && engine_in_check()) {
-		D_DEBUG(DB_MD, DF_UUID": end: skip in check mode\n", DP_UUID(svc->ps_uuid));
+		D_DEBUG(DB_MD, DF_UUID": end: skip automatic reconf in check mode\n",
+			DP_UUID(svc->ps_uuid));
 		return;
 	}
 
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -116,6 +116,15 @@ reconf_end(struct pool_svc_reconf *reconf)
 }
 
 static void
+reconf_wait(struct pool_svc_reconf *reconf)
+{
+	ABT_mutex_lock(reconf->psc_mutex);
+	while (reconf->psc_in_progress)
+		ABT_cond_wait(reconf->psc_cv, reconf->psc_mutex);
+	ABT_mutex_unlock(reconf->psc_mutex);
+}
+
+static void
 reconf_cancel_and_wait(struct pool_svc_reconf *reconf)
 {
 	/*
@@ -1583,7 +1592,7 @@ pool_svc_check_node_status(struct pool_svc *svc)
 	D_PRINT(fmt, ## __VA_ARGS__);								\
 } while (0)
 
-static void pool_svc_schedule_reconf(struct pool_svc *svc);
+static void pool_svc_schedule_reconf(struct pool_svc *svc, bool for_chk);
 static void pool_svc_cancel_and_wait_reconf(struct pool_svc *svc);
 
 static int
@@ -1645,7 +1654,7 @@ pool_svc_step_up_cb(struct ds_rsvc *rsvc)
 	 * reconfigurations or the last MS notification.
 	 */
 	svc->ps_reconf.psc_force_notify = true;
-	pool_svc_schedule_reconf(svc);
+	pool_svc_schedule_reconf(svc, false /* for_chk */);
 	reconf_scheduled = true;
 
 	rc = ds_pool_iv_prop_update(svc->ps_pool, prop);
@@ -5413,7 +5422,7 @@ out:
 }
 
 static void
-pool_svc_schedule_reconf(struct pool_svc *svc)
+pool_svc_schedule_reconf(struct pool_svc *svc, bool for_chk)
 {
 	struct pool_svc_reconf *reconf = &svc->ps_reconf;
 	enum ds_rsvc_state	state;
@@ -5421,7 +5430,7 @@ pool_svc_schedule_reconf(struct pool_svc *svc)
 
 	D_DEBUG(DB_MD, DF_UUID": begin\n", DP_UUID(svc->ps_uuid));
 
-	if (engine_in_check()) {
+	if (!for_chk && engine_in_check()) {
 		D_DEBUG(DB_MD, DF_UUID": end: skip in check mode\n", DP_UUID(svc->ps_uuid));
 		return;
 	}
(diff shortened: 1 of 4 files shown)


src/pool/srv_pool.c
```diff
@@ -5464,6 +5473,19 @@ pool_svc_cancel_and_wait_reconf(struct pool_svc *svc)
...
+ * Perform PS reconfigurations (if necessary) synchronously. This is currently
+ * for the chk module only.
+ */
+void
+ds_pool_svc_reconf(struct ds_pool_svc *svc)
+{
+	struct pool_svc *s = pool_ds2svc(svc);
+
+	pool_svc_schedule_reconf(s, true /* for_chk */);
+	reconf_wait(&s->ps_reconf);
```
kccain, line 5486: Refamiliarizing with the code after reviewing earlier and having some older discussions about cat_recovery I don't recall well. Apologies if it's a silly question: In this scenario (for_chk == true), will it be the case that there is no pool_svc_reconf_ult running? i.e., the reconf_cancel_and_wait() call in pool_svc_schedule_reconf() will only lock and unlock the mutex and not actually request to cancel an in progress ULT? Just wondering otherwise if there would be a race to be unblocked on the CV wait between this reconf_wait() call and the reconf_cancel_and_wait() call.

kccain (approved):

src/pool/srv_pool.c
```diff
@@ -5413,15 +5422,15 @@ pool_svc_reconf_ult(void *arg)
...
 {
 	struct pool_svc_reconf *reconf = &svc->ps_reconf;
 	enum ds_rsvc_state	state;
 	int			rc;
 
 	D_DEBUG(DB_MD, DF_UUID": begin\n", DP_UUID(svc->ps_uuid));
 
-	if (engine_in_check()) {
+	if (!for_chk && engine_in_check()) {
 		D_DEBUG(DB_MD, DF_UUID": end: skip in check mode\n", DP_UUID(svc->ps_uuid));
```
Nasf-Fan, line 5434: The log message is some misguided for the new patch.

src/pool/srv_pool.c
```diff
@@ -115,6 +115,15 @@ reconf_end(struct pool_svc_reconf *reconf)
 	reconf->psc_canceled = false;
 }
 
+static void
+reconf_wait(struct pool_svc_reconf *reconf)
+{
+	ABT_mutex_lock(reconf->psc_mutex);
+	while (reconf->psc_in_progress)
+		ABT_cond_wait(reconf->psc_cv, reconf->psc_mutex);
```
Nasf-Fan, line 123: How long the wait may take? It is possible that "dmg check stop" coming in during such wait, we may need some interface to wakeup the waiting reconf under such case to avoid being blocked for very long time?

src/pool/srv_pool.c
liw: Currently there should be no PS reconf ULTs when the chk module calls this function. On feature/cat_recovery I've added engine_in_check calls to pool_svc_schedule_reconf so that the PS won't schedule any automatic reconf ULTs in the "check" mode, which is set to true when the daos_engine process starts.

Sorry, I don't get the problem with the "race" if there were existing PS reconf ULTs. Could you elaborate a little, @kccain?

src/pool/srv_pool.c
liw: @Nasf-Fan, if we see this message, then we are indeed "in check mode" and we will indeed "skip" scheduling a new PS reconf. If we are in check mode and for_chk is true, then we won't print this message. We're good, no? Or, do you want me to add "automatic", such as "end: skip automatic reconf in check mode", to imply that this message won't be printed if this is not an automatic reconf request?

src/pool/srv_pool.c
liw: I think the simplest way is to just do not wait for the reconf, like we don't wait for the automatic reconfs. Because the PS is already available and the pool can already work. Tests that must verify the reconf result can just wait for a minute or so.

liw:
Force-pushed in order to update the commit message; all code changes are in the new commit.

src/pool/srv_pool.c
kccain: I was confused. In check mode, even if a PS reconf ULT were running (let's call it ULT execution A) at the time of this pool_svc_reschedule_reconf(), then the reconf_cancel_and_wait() call would cause that ULT to be cancelled, then would cause another ULT (let's call it ULT execution B) to occur. Then the reconf_wait() call here would wait for execution B to be done.

My confusion was if somehow we could have 2 threads waiting on reconf->psc_cv and a cancelled ULT calls cond_broadcast then we don't know which of the 2 threads waiting on psc_cv would be awoken first.

Looks like in your latest commit though this reconf_wait() call is removed.

src/pool/srv_pool.c
Nasf-Fan: I mean that the message "skip in check mode" is inconsistent with "for_chk" parameter, because if "for_chk" is true, the pool_svc_schedule_reconf() is not real skipped. So some suggestion:

1. Adjust the log message, such as you said "skip automatic reconf ...", or
2. Rename "for_chk" as "force", and if "force" is true, then not skip; otherwise if it is in check mode, then skip.

src/pool/srv_pool.c
Nasf-Fan: What will happen if pool stop during reconfs? Because the user may wants to restart daos check before reconfs complete, right?

src/pool/srv_pool.c
liw: There will be at most one PS reconf ULT at any moment during the term of a PS leader. We implement that by always cancel any existing PS reconf ULT and wait for it to terminate first (the reconf_cancel_and_wait call) in pool_svc_schedule_reconf. A new PS reconf ULT will take care of what any canceled one should do.

In this very chk scenario, besides the absence of existing PS reconf ULTs, I think it suffices to only trigger a best-effort reconf.

src/pool/srv_pool.c
liw: If some code calls ds_pool_svc_stop, and if this is the PS leader with a running reconf ULT, then the reconf ULT will be canceled. I think the PS reconf is not part of CR and the CR state of this pool should move to `DONE` regardless of the PS reconf.

src/pool/srv_pool.c
liw: I anticipate that we will remove `for_chk` once we replace `engine_in_check` with some real per-pool flag. So I'll only change the log message here. Hopefully that satisfies your requirement.

Nasf-Fan (approved):

PR #10451, opened 2022-09-30
DAOS-11605 common: ad-hoc allocator for umem

gnailzenh:
Main data structures:
- ad_blob
  Represent the Metadata blob, all the space allocations happen with
  the logic address space of this blob.

- ad_arena
  An ad_blob is divided into 16MB pages, an arena represents a 16MB page
  and manage all the allocatino metadata for this page.
    
- ad_group
  A group is a set objects with the same size, a arena can have many
  groups with different sizes.
    
supported API
- create/open/close a blob
- reserve
- tx_publish/cancel
- tx_begin/end
    
status:
- passed the simplest test
- need a lot more tests and review

Signed-off-by: Liang Zhen <liang.zhen@intel.com>


src/common/ad_mem.c
```diff
@@ -3,5 +3,1578 @@
...
+
+	nr = nr_saved = 0;
+	at = at_saved = -1;
+
+	for (i = n = 0; i < bmap_sz; i++) {
+		uint64_t free_bits = ~used[i];
+		uint64_t tmp;
+
+		if (reserved)
+			free_bits |= ~reserved[i];
```
liuxuezhao, line 989: should it be "&="?

src/common/ad_mem.c
```diff
@@ -3,5 +3,1578 @@
...
+		} else {
+			at = i * 64 + j - 1;
+			nr = 1;
+		}
+
+		for (; j < 64; j++) {
+			if (nr == *bits) /* done */
+				goto out;
+
+			if (isset(free_bits, j)) {
```
liuxuezhao, line 1015: isset/isclr/setbit/clrbit maros seems only works with "uint8 []" type (see /usr/include/sys/param.h),
if use this macros, may define all bitmap to be "uint8_t []" or "char []".

src/common/ad_mem.h
```diff
@@ -0,0 +1,401 @@
...
+	/** link chain on ad_blob LRU */
+	d_list_t		  ar_link;
+	/** arena type */
+	int			  ar_type;
+	/** refcount */
+	int			  ar_ref;
+	/** number of groups */
+	int			  ar_grp_nr;
+	/** published or not */
+	bool			  ar_published;
```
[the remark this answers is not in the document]
gnailzenh: a unpublished arena will stay in DRAM until another call to published (don't try to undo it for simplicity)

src/common/ad_mem.c
```diff
@@ -3,5 +3,1578 @@
...
+	D_FREE(grp);
+}
+
+struct ad_arena *
+alloc_arena(struct ad_blob *blob, bool force)
+{
+	struct ad_arena *arena = NULL;
+
+	if (!force)
+		arena = d_list_pop_entry(&blob->bb_gps_lru);
```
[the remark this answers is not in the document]
gnailzenh: right

src/common/ad_mem.c
```diff
@@ -3,5 +3,1578 @@
...
+	struct ad_arena		*arena;
+	struct ad_arena_df	*ad;
+	struct umem_store_iod	 iod;
+	d_sg_list_t		 sgl;
+	d_iov_t			 iov;
+	uint32_t		 id;
+	int			 rc;
+
+	/* NB: store::stor_size can be changed by upper level stack */
+	blob->bb_pgs_nr = ((blob_size(blob) + ARENA_SIZE_MASK) >> ARENA_SIZE_BITS);
```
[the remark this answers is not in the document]
gnailzenh: correct

src/common/ad_mem.c
```diff
@@ -3,5 +3,1578 @@
...
+		return -DER_INVAL;
+
+	if (specs_nr >= ARENA_GRP_SPEC_MAX)
+		return -DER_INVAL;
+
+	spec = &bd->bd_asp[arena_type];
+	if (spec->as_specs_nr != 0)
+		return -DER_EXIST;
+
+	ad_tx_increase(tx, &bd->bd_asp_nr, sizeof(bd->bd_asp_nr));
```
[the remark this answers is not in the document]
gnailzenh: I've changed this  and make only for int32 now

src/common/ad_mem.c
gnailzenh: correct, we need to do some typecast at here

src/common/ad_mem.h
```diff
@@ -0,0 +1,323 @@
...
+	d_list_t		 tx_redo;
+	d_list_t		 tx_ar_pub;
+	d_list_t		 tx_gp_pub;
+	d_list_t		 tx_gp_free;
+	uint32_t		 tx_redo_act_nr;
+	uint32_t		 tx_redo_payload_len;
+	struct umem_act_item	*tx_redo_act_pos;
+};
+
+/** action paramenters for free() */
```
jolivier23, line 29: need to fix these or they will show up on every PR

src/common/ad_mem.h
```diff
@@ -0,0 +1,401 @@
...
+	 */
+	uint64_t		ad_incarnation;
+	/** external blob ID, an arena can be created for external blob */
+	uint64_t		ad_blob_id;
+	/** blob address */
+	uint64_t		ad_addr;
+	/** for future use */
+	uint64_t		ad_reserved[2];
+	/** 64 bytes (512 bits) for each, each bit represents 32K(minimum group size) */
+	uint64_t		ad_bmap[ARENA_GRP_BMSZ];
```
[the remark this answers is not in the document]
gnailzenh: I fixed the comment

gnailzenh:
Sorry I force pushed for a few times, just want to keep the history clean before the first version that is ready for review. Now I think it's ready for review, although it can only pass the simplest test, but at least the code is in reasonable shape.

src/common/ad_mem.c
```diff
@@ -3,5 +3,1857 @@
...
+
+	blob->bb_arena_last[0]	= bd->bd_asp[0].as_arena_last;
+	/* already published arena[0], clear the reserved bit */
+	clrbit(blob->bb_bmap_rsv, arena->ar_df->ad_id);
+
+	ad_iod_set(&iod, blob_ptr2addr(blob, arena->ar_df), ARENA_HDR_SIZE + BLOB_HDR_SIZE);
+	ad_sgl_set(&sgl, &iov, arena->ar_df, ARENA_HDR_SIZE + BLOB_HDR_SIZE);
+
+	D_ASSERT(store->stor_ops);
+	rc = store->stor_ops->so_write(store, &iod, &sgl);
```
NiuYawei, line 367: I'm wondering what will be stored in such a large superblock, since no WAL to guarantee the atomicity of this 32k superblock update, you need to update the 4k block contains magic after all other block updates done, so that we could tell the superblock consistent by check the magic on load time.

And I'm not sure why we don't rely on WAL to write ad superblock? because it's only updated on format time?

src/common/ad_mem.c
```diff
@@ -3,5 +3,1857 @@
...
+
+	store->stor_size = bd->bd_size;
+	store->stor_addr = bd->bd_addr;
+	blob->bb_pgs_nr = ((blob_size(blob) + ARENA_SIZE_MASK) >> ARENA_SIZE_BITS);
+
+	rc = blob_init(blob);
+	if (rc)
+		goto failed;
+
+	rc = blob_load(blob);
```
NiuYawei, line 458: I think the whole tmpfs loading (including the ad superblock) should be moved to generic umem interface and being shared by BMEM, allocator can assume the data is already loaded from SSD to tmpfs.

NiuYawei (commented):
Haven't review the allocator algorithm yet, just few quick comments related to the meta store and generic umem interfaces.

I'll continue on the allocator part review soon.

src/common/ad_mem.h
```diff
@@ -0,0 +1,327 @@
...
+	/** unused arena LRU */
+	d_list_t		 bb_ars_lru;
+	/** group being reserved (not published), pinned at here */
+	d_list_t		 bb_gps_rsv;
+	/** unused group LRU */
+	d_list_t		 bb_gps_lru;
+	int			 bb_ars_lru_size;
+	int			 bb_gps_lru_size;
+	/** file descriptor of MD file */
+	int			 bb_fd;
```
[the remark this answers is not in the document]
gnailzenh: ok, will remove it later.

src/common/ad_mem.h
```diff
@@ -0,0 +1,327 @@
...
+	struct ad_arena_spec	bd_asp[ARENA_SPEC_MAX];
+	/** loading incarnation */
+	uint64_t		bd_incarnation;
+	/** it is DRAM reference of blob */
+	uint64_t		bd_back_ptr;
+	/** XXX: SPDK blob ID  */
+	uint64_t		bd_id;
+	/** start address in the SPDK blob */
+	uint64_t		bd_addr;
+	/** size of the blob */
```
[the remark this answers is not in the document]
gnailzenh: sure, will fix the comment

src/common/ad_mem.c
gnailzenh: This is the format, so let's do it in a simple way, after the format, all the update to superblock will go through WAL.

src/common/ad_mem.c
gnailzenh: the thing bothered me is, page cache is supposed to be managed by the allocator... we have to allocate pages while loading information from SSD right?

src/common/ad_mem.c
```diff
@@ -3,5 +3,1917 @@
...
+	/* no cached arena, allocate it now */
+
+	arena = alloc_arena(blob, false);
+	if (!arena)
+		return -DER_NOMEM;
+
+	/* NB: don't need to write to WAL,
+	 * NB: stale pointer can be detected by incarnation.
+	 */
+	ad->ad_back_ptr = (unsigned long)arena;
```
NiuYawei, line 648: Do we really need to put a DRAM pointer into durable format?

NiuYawei (commented):
Few more comments about where to put page cache management.

src/common/ad_mem.c
```diff
@@ -3,5 +3,1917 @@
...
+	}
+
+	while ((group = d_list_pop_entry(&tx->tx_gp_pub, struct ad_group, gp_link))) {
+		group->gp_publishing = false;
+		if (err) { /* keep the refcount and pin it */
+			d_list_add(&group->gp_link, &blob->bb_gps_rsv);
+			continue;
+		}
+		for (i = 0; i < group->gp_bit_nr; i++)
+			clrbit(group->gp_arena->ar_bmap_rsv, group->gp_bit_at + i);
```
NiuYawei, line 1590: Same here. Isn't the reserve bit cleared (turned into allocation bits) on publish?

NiuYawei (commented):
few comments/questions on implementation details.

src/common/ad_mem.c
gnailzenh: I agree, I actually thought about this, but I decided to put them together in the first step, because it is little difficult to abstract things before I have a few pieces. I can do this in follow-on patches.

src/common/ad_mem.c
```diff
@@ -3,5 +3,1917 @@
...
+		D_ERROR("No available arena\n");
+		return rc;
+	}
+
+	if (ad->ad_magic != ARENA_MAGIC) {
+		D_ERROR("Invalid arena magic: %x/%x\n", ad->ad_magic, ARENA_MAGIC);
+		return -DER_PROTO;
+	}
+
+	if (ad->ad_incarnation != blob_incarnation(blob)) {
```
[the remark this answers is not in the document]
gnailzenh: instead of maintaining another index in DRAM, I decided to use blob-map to index almost everything. For example, for reserve() implementation,  I need to maintain an ephemeral bitmap in DRAM  (this bitmap disappear if the engine crashes), as addition of the bitmap in blob. To avoid another index, the ad_group_df has a backpointer to ad_group, the backpointer can be different each time we start the engine (it's never written to WAL or checkpointed intentionally, but I can be written to SSD by other changes made to ad_group_df), so I'm using the incarnation to check if the backpointer is still valid (it's a. up-to-date pointer, or a stale pointer from the previous run of the engine).

src/common/ad_mem.c
gnailzenh: as I explained, it can avoid another index, which is good for performance.

src/common/ad_mem.c
```diff
@@ -3,5 +3,1917 @@
...
+	ad_tx_increase(tx, &bd->bd_asp_nr);
+
+	spec->as_type	    = arena_type;
+	spec->as_specs_nr   = specs_nr;
+	spec->as_arena_last = AD_ARENA_ANY;
+	for (i = 0; i < specs_nr; i++)
+		spec->as_specs[i] = specs[i];
+
+	blob->bb_arena_last[arena_type] = AD_ARENA_ANY;
+	ad_tx_copy(tx, spec, sizeof(*spec), spec, AD_TX_REDO | AD_TX_UNDO);
```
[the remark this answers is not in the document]
gnailzenh: this should be replaced by ad_tx_snap, I don't think we have any use case for ad_tx_copy for now, thanks

src/common/ad_mem.c
```diff
@@ -3,5 +3,1917 @@
...
+				/* XXX: other sized groups may have space. */
+				arena->ar_full = true;
+				return 0;
+			}
+		}
+		D_DEBUG(DB_TRACE, "Found group=%p [r=%d, f=%d] for size=%d in arena=%d\n",
+			grp, grp->gp_unit_rsv, grp->gp_df->gd_unit_free,
+			(int)size, arena->ar_df->ad_id);
+
+		addr = group_reserve_addr(grp, act);
```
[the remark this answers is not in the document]
gnailzenh: No, but we can customize a different arena type which can support large group size :-)
but arena registration is not tested yet.

src/common/ad_mem.c
```diff
@@ -3,5 +3,1917 @@
...
+			tried = true;
+		}
+
+		D_DEBUG(DB_TRACE, "reserve from arena==%d\n", arena->ar_df->ad_id);
+		addr = arena_reserve_addr(arena, size, act);
+		if (!addr) { /* full arena */
+			D_DEBUG(DB_TRACE, "full arena=%d\n", arena->ar_df->ad_id);
+			arena_decref(arena);
+			if (tried)
+				return 0;
```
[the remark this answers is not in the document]
gnailzenh: this version cannot reuse arena yet, but my plan is creating arena index, and compute arena weight based on free space of each arena (it's going to be a lazy check, not going to compute weight on each allocate/free), so after freeing space an arena can be reused at some point.

src/common/ad_mem.c
```diff
@@ -3,5 +3,1917 @@
...
+	else
+		rc = err;
+
+	while ((arena = d_list_pop_entry(&tx->tx_ar_pub, struct ad_arena, ar_link))) {
+		arena->ar_publishing = false;
+		if (err) { /* keep the refcount and pin it */
+			d_list_add(&arena->ar_link, &blob->bb_ars_rsv);
+			continue;
+		}
+		clrbit(blob->bb_bmap_rsv, arena->ar_df->ad_id);
```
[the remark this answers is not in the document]
gnailzenh: No, only one transaction can publish an arena/group, it will be move back to the global list if it failed/canceled. The next transaction will do it again.

src/common/ad_mem.c
gnailzenh: This is called on successful committed transaction, if the transaction is canceld/aborted, we don't reset these bits.

src/common/ad_mem.h
```diff
@@ -0,0 +1,327 @@
...
+	/** specifications of registered arena types */
+	struct ad_arena_spec	bd_asp[ARENA_SPEC_MAX];
+	/** loading incarnation */
+	uint64_t		bd_incarnation;
+	/** it is DRAM reference of blob */
+	uint64_t		bd_back_ptr;
+	/** XXX: SPDK blob ID  */
+	uint64_t		bd_id;
+	/** start address in the SPDK blob */
+	uint64_t		bd_addr;
```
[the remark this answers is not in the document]
gnailzenh: OK, then I will remove it.

src/common/ad_mem.c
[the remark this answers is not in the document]
NiuYawei: Can multiple transactions reserve same arena? I mean if tx A reserved arena 1 and yield on wal commit, can tx B reserve the same arena?

src/common/ad_mem.c
NiuYawei: I mean could the same reserve bit being set by another tx while this tx yield on wal commit?

src/common/ad_mem.c
```diff
@@ -3,5 +3,2003 @@
...
+		D_ASSERT(arena->ar_unpub);
+		arena_decref(arena);
+	}
+
+	while ((group = d_list_pop_entry(&blob->bb_gps_lru, struct ad_group, gp_link)))
+		free_group(group, true);
+	while ((arena = d_list_pop_entry(&blob->bb_ars_lru, struct ad_arena, ar_link)))
+		free_arena(arena, true);
+
+	if (blob->bb_bmap_rsv)
```
wangshilong, line 252: [minor] D_FREE inside already checked NULL, no need check in the caller.

src/common/ad_mem.c
```diff
@@ -3,5 +3,2003 @@
...
+}
+
+static void
+blob_set_opened(struct ad_blob *blob)
+{
+	struct ad_blob_df *bd = blob->bb_df;
+
+	bd->bd_incarnation = d_timeus_secdiff(0);
+	bd->bd_back_ptr	   = (unsigned long)blob;
+
```
wangshilong, line 306: (unsigned long) is 4 byte, bd_back_ptr is uint64, should we use uint64_t here?

src/common/ad_mem.h
```diff
@@ -0,0 +1,327 @@
...
+	int			 bb_ars_lru_size;
+	int			 bb_gps_lru_size;
+	/** file descriptor of MD file */
+	int			 bb_fd;
+	/** reference counter */
+	int			 bb_ref;
+	/** is dummy blob, for unit test */
+	bool			 bb_dummy;
+	/** opened blob */
+	bool			 bb_opened;
```
wangshilong, line 271: [minor] this might declared as to save some memory:
unsigned int                                      bb_dump:1,
                                                          bb_opened:1,

src/include/daos_srv/ad_mem.h
```diff
@@ -90,4 +94,69 @@ ad_tx_redo_act_first(struct ad_tx *tx);
 struct umem_action *
 ad_tx_redo_act_next(struct ad_tx *tx);
 
+static inline int
+ad_tx_decrease(struct ad_tx *tx, int32_t *addr)
+{
```
wangshilong, line 99: uint32_t addr?

src/common/ad_mem.c
```diff
@@ -3,5 +3,2003 @@
...
+		if (blob->bb_ars_lru_size < ARENA_LRU_SIZE) {
+			blob->bb_ars_lru_size++;
+			return;
+		}
+		/* release an old one from the LRU */
+		arena = d_list_pop_entry(&blob->bb_ars_lru, struct ad_arena, ar_link);
+	}
+	if (arena->ar_blob)
+		blob_decref(arena->ar_blob);
+	if (arena->ar_addr_sorter)
```
wangshilong, line 1997: D_FREE() check NULL inside.

src/include/daos_srv/ad_mem.h
gnailzenh: yes, it's address of integer

src/common/ad_mem.h
gnailzenh: this is the superblock, so I think it doesn't matter? if it's a data structure which has many instances, then I would agree.

src/common/ad_mem.c
gnailzenh: No, unsigned long is 8 bytes on 64-bit machine, it has the same size as pointer, so if we want to convert it to uint64_t,
 we need to do this:
 back_ptr = (uint64_t)(unsigned long)blob;
we cannot cast pointer to uint64_t directly, compiler will complain.

src/common/ad_mem.c
gnailzenh: right, I will remove in follow-on patches.

src/common/ad_mem.h
wangshilong: sure

src/common/ad_mem.c
wangshilong: you are right, i thought 'unsigned long' is different with 'unsigned long long', but they are both 8 bytes..

src/common/ad_mem.c
```diff
@@ -3,5 +3,2004 @@
...
+
+	bit_at = (gd->gd_addr - ad->ad_addr - ARENA_HDR_SIZE) >> GRP_SIZE_SHIFT;
+	bit_nr = (gd->gd_unit_nr * gd->gd_unit) >> GRP_SIZE_SHIFT;
+	D_DEBUG(DB_TRACE, "publishing group=%p, bit_at=%d, bits_nr=%d\n", group, bit_at, bit_nr);
+
+	ad_tx_setbits(tx, ad->ad_bmap, bit_at, bit_nr);
+	ad_tx_increase(tx, &ad->ad_grp_nr);
+
+	ad_tx_set(tx, gd, 0, sizeof(*gd), AD_TX_REDO | AD_TX_LOG_ONLY);
+	ad_tx_snap(tx, gd, offsetof(struct ad_group_df, gd_bmap[0]), AD_TX_REDO);
```
wangshilong, line 1407: above ad_tx_xxx function might fail

src/common/ad_mem.c
```diff
@@ -3,5 +3,2004 @@
...
+		return -DER_INVAL;
+
+	if (specs_nr >= ARENA_GRP_SPEC_MAX)
+		return -DER_INVAL;
+
+	spec = &bd->bd_asp[arena_type];
+	if (spec->as_specs_nr != 0)
+		return -DER_EXIST;
+
+	ad_tx_increase(tx, &bd->bd_asp_nr);
```
wangshilong, line 1842: check return value.

src/common/ad_mem.c
```diff
@@ -3,5 +3,2004 @@
...
+
+	spec->as_type	    = arena_type;
+	spec->as_specs_nr   = specs_nr;
+	spec->as_arena_last = AD_ARENA_ANY;
+	for (i = 0; i < specs_nr; i++)
+		spec->as_specs[i] = specs[i];
+
+	blob->bb_arena_last[arena_type] = AD_ARENA_ANY;
+	ad_tx_snap(tx, spec, sizeof(*spec), AD_TX_REDO);
```
wangshilong, line 1851: this as well..

src/common/ad_tx.c
```diff
@@ -268,42 +327,47 @@ ad_tx_assign(struct ad_tx *tx, void *addr, daos_size_t size, uint32_t val)
...
-	it_redo->it_act.ac_set.addr = ad_ptr2addr(addr);
-	it_redo->it_act.ac_set.size = size;
-	it_redo->it_act.ac_set.val = c;
-	AD_TX_ACT_ADD(tx, it_redo, ACT_REDO);
+	if (!(flags & AD_TX_LOG_ONLY))
+		memset(addr, c, size);
+
+	if (flags & AD_TX_REDO) {
+		D_ALLOC_ACT(it_redo, UMEM_ACT_SET, size);
+		if (it_redo == NULL)
```
wangshilong, line 363: Do we need free above memory if AD_TX_UNDO is set, or  cleanup will be done in the tx abort?

src/common/ad_tx.c
```diff
@@ -146,18 +152,62 @@ ad_tx_end(struct ad_tx *tx, int err)
...
 	if (err)
-		rc = ad_tx_act_replay(&tx->tx_undo);
+		rc = ad_tx_act_replay(tx, &tx->tx_undo);
 
 	/* TODO write actions in redo list to WAL */
-
+	rc = tx_complete(tx, err);
 	if (rc == 0) {
 		ad_tx_act_cleanup(&tx->tx_undo);
 		ad_tx_act_cleanup(&tx->tx_redo);
```
wangshilong, line 161: sorry, we don't need run act cleanup if some error happen?

src/common/ad_mem.c
liuxuezhao: seems cannot reserve/set by another tx again, as the "ad->ad_bmap" is already set?

src/common/ad_mem.c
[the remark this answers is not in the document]
liuxuezhao: ad_blob_df::bd_bmap is set so next reservation of other tx will not select that bit?

src/common/ad_tx.c
gnailzenh: No we don't need to, tx_end() will handle all these (even it doesn't, it should)

liuxuezhao (approved):

src/common/ad_mem.c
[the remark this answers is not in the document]
NiuYawei: Oh, right. Hmm, then I'm wondering why do we need reserve bit for arena?

src/common/ad_mem.c
NiuYawei: yeah, seems so. Then I'm wondering what's the purpose of reserve bit in such case? I can understand reserve bit for a particular allocation, but I don't quite see why do we need reserve bit arena or group? Arena & group are shared by transactions anyway, right?

src/common/ad_tx.c
```diff
@@ -146,18 +161,62 @@ ad_tx_end(struct ad_tx *tx, int err)
...
+		struct umem_act_item	*it_redo = NULL;
+
+		D_ALLOC_ACT(it_redo, UMEM_ACT_COPY, size);
+		if (it_redo == NULL)
+			return -DER_NOMEM;
+
+		act_copy_payload(&it_redo->it_act, addr, size);
+		it_redo->it_act.ac_copy.addr = blob_ptr2addr(tx->tx_blob, addr);
+		it_redo->it_act.ac_copy.size = size;
+		AD_TX_ACT_ADD(tx, it_redo, ACT_REDO);
```
wangshilong, line 215: [style] Above codes could be cleaned to share, only different is last line.
AD_TX_ACT_ADD(tx, act_item, undo ? ACT_UNDO : ACT_REDO)

src/common/ad_tx.c
```diff
@@ -218,6 +277,41 @@ ad_tx_copy(struct ad_tx *tx, void *addr, daos_size_t size, void *ptr, uint32_t f
 	return 0;
 }
 
+static int
```
wangshilong, line 280: [defect] return value here should be uint32_t

liuxuezhao (approved):

src/common/tests/ad_mem_tests.c
```diff
@@ -0,0 +1,384 @@
...
+	struct ad_reserv_act acts[rsv_count];
+	daos_off_t	     addrs[rsv_count];
+	int		     i;
+	uint32_t	     arena = AD_ARENA_ANY;
+
+	printf("multiple reserve and cancel\n");
+	for (i = 0; i < rsv_count; i++) {
+		addrs[i] = ad_reserve(adt_bh, 0, alloc_size, &arena, &acts[i]);
+		if (addrs[i] == 0) {
+			fprintf(stderr, "failed allocate\n");
```
wangshilong, line 161: [minor] misssed cancel.
if (i > 0)
    ad_cancel(acts, i)

PR #10456, opened 2022-09-30
DAOS-11697 test: bad_evict.py test baduuid case, manipulation of UUID can be a no-op and cause negative test failure

dinghwah:
Description: Check pool uuid 4th byte before update an invalid ID.
Test-repeat: 10
Test-tag: bad_evict
Required-githooks: true

Signed-off-by: Ding Ho ding-hwa.ho@intel.com

### DIFF
--- a/src/tests/ftest/pool/bad_evict.py
+++ b/src/tests/ftest/pool/bad_evict.py
@@ -63,7 +63,10 @@ class BadEvictTest(TestWithServers):
             saveduuid = (ctypes.c_ubyte * 16)(0)
             for index, _ in enumerate(saveduuid):
                 saveduuid[index] = self.pool.pool.uuid[index]
-            self.pool.pool.uuid[4] = 244
+            if self.pool.pool.uuid[4] != 244:
+                self.pool.pool.uuid[4] = 244
+            else:
+                self.pool.pool.uuid[4] == 255
 
         self.pool.uuid = self.pool.pool.get_uuid_str()
 


src/tests/ftest/pool/bad_evict.py
```diff
@@ -63,7 +63,10 @@ def test_evict(self):
             saveduuid = (ctypes.c_ubyte * 16)(0)
             for index, _ in enumerate(saveduuid):
                 saveduuid[index] = self.pool.pool.uuid[index]
-            self.pool.pool.uuid[4] = 244
+            if self.pool.pool.uuid[4] != 244:
+                self.pool.pool.uuid[4] = 244
+            else:
+                self.pool.pool.uuid[4] == 255
```
shimizukko, line 69:
```suggestion
                self.pool.pool.uuid[4] = 255
```
Typo?

shimizukko (approved):

src/tests/ftest/pool/bad_evict.py
dinghwah: Yes. Thanks.

src/tests/ftest/pool/bad_evict.py
```diff
@@ -63,7 +63,10 @@ def test_evict(self):
             saveduuid = (ctypes.c_ubyte * 16)(0)
             for index, _ in enumerate(saveduuid):
                 saveduuid[index] = self.pool.pool.uuid[index]
-            self.pool.pool.uuid[4] = 244
+            if self.pool.pool.uuid[4] != 244:
+                self.pool.pool.uuid[4] = 244
+            else:
+                self.pool.pool.uuid[4] = 255
```
daltonbohning, line 69: Why 244 and 255?

dinghwah: Original bad_evict testcase choose 244 (0xf4) and over-written to byte-4 of pool uuid as a negative testcase, when a pool origin uuid byte-4 is 0xf4 test passed on this negative test.

daltonbohning: But where does 0xf4 come from? I.e. where is it set, and why THAT number? Looking at 244 and 255 here out of context, they seem like magic numbers

dinghwah: This is an old testcase to evict an invalid/junk pool-uuid. We can choose a random number, but it would not gain any coverage.

shimizukko (approved):

PR #10460, opened 2022-10-01
DAOS-11679 dtx: dispatch large transaction via multiple steps

Nasf-Fan:
If a large transaction has sub-requests to dispatch to a lot of DTX participants, then we may have to split the dispatch process to multiple steps; otherwise, the dispatch process may trigger too many in-flight or in-queued RPCs that will hold too much resource as to server maybe out of memory.


Nasf-Fan:
Not need it any longer.

jolivier23 (commented):
is this ready for review?

Nasf-Fan:
Related fixes has been landed to master via https://github.com/daos-stack/daos/pull/10569

PR #10485, opened 2022-10-05
DAOS-11601 object: refine daos_recx_ep_list_set

liuxuezhao:
Refine daos_recx_ep_list_set() to release the requirement of parity epoch mismatch check for client side,
because the client-side retry for parity epoch mismatch is for the case of EC aggregation unstable case that
parity ext updated to different parity shards asynchronously. This patch set the parity epoch as max of
parity ext and EC aggregation epoch boundary, which means once the EC aggregation done we can release
the requirement of parity ext epoch mismatch check for client-side retry.

Required-githooks: true

### DIFF
--- a/src/object/srv_obj.c
+++ b/src/object/srv_obj.c
@@ -1519,7 +1519,8 @@ obj_local_rw_internal(crt_rpc_t *rpc, struct obj_io_context *ioc,
 		if (get_parity_list) {
 			parity_list = vos_ioh2recx_list(ioh);
 			if (parity_list != NULL) {
-				daos_recx_ep_list_set(parity_list, orw->orw_nr, 0, 0);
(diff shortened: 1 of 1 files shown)


wangdi1 (approved):

Nasf-Fan (approved):

jolivier23:
can you add a little description here?  What does the patch do, why is it needed, etc?

liuxuezhao:
> can you add a little description here? What does the patch do, why is it needed, etc?

sure, I added some detailed description at the commit msg, please check again, thanks @jolivier23

jolivier23:
> > can you add a little description here? What does the patch do, why is it needed, etc?
> 
> sure, I added some detailed description at the commit msg, please check again, thanks @jolivier23

Thanks

PR #10487, opened 2022-10-05
DAOS-11750 object: wait for 0 xstream tls stop

wangdi1:
In ds_migrate_stop(), migrate_fini_one_ult only wait non-0 xstream ULTs to stop, so it should also wait for 0 xstream migrate ULT to stop before destroying migrate tls.

Required-githooks: true

### DIFF
--- a/src/object/srv_obj_migrate.c
+++ b/src/object/srv_obj_migrate.c
@@ -2703,6 +2703,17 @@ ds_migrate_stop(struct ds_pool *pool, unsigned int version)
 	if (tls->mpt_opc == RB_OP_REINT)
 		pool->sp_reintegrating--;
 	migrate_pool_tls_put(tls);
+	tls->mpt_fini = 1;
+	/* Wait for xstream 0 migrate ULT(migrate_ult) stop */
+	if (tls->mpt_ult_running) {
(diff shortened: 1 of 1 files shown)


jolivier23 (approved):

wangdi1:
failed with 1G exceed failure.

src/object/srv_obj_migrate.c
```diff
@@ -2703,6 +2703,16 @@ ds_migrate_stop(struct ds_pool *pool, unsigned int version)
 	if (tls->mpt_opc == RB_OP_REINT)
 		pool->sp_reintegrating--;
 	migrate_pool_tls_put(tls);
+	/* Wait for xstream 0 migrate ULT(migrate_ult) stop */
+	if (tls->mpt_ult_running) {
```
liuxuezhao, line 2707: seems no place set "tls->mpt_fini = 1" for XS 0? above "dss_thread_collective(migrate_fini_one_ult, &arg, 0);" did not do that for XS0.

wangdi1: ah, right. Thanks.

liuxuezhao (approved):

jolivier23 (approved):

PR #10490, opened 2022-10-05
DAOS-11828 dtx: handle excluded participant as prepared when DTX resync

Nasf-Fan:
During DTX resync, if non-leader is excluded, handle it as 'prepared'. If other non-leaders are also 'prepared' then related DTX maybe still committable or 'corrupted'. The subsequent DTX resync logic will handle related things, see dtx_verify_groups().

More fixes:

1. "-DER_INPROGRESS" for DTX RPC reply is normal, not error.

2. DTX RPC should not skip the in-drain target.

### DIFF
--- a/src/dtx/dtx_rpc.c
+++ b/src/dtx/dtx_rpc.c
@@ -301,15 +301,18 @@ dtx_req_list_cb(void **args)
 					drr->drr_rank, drr->drr_tag);
 				return;
 			case -DER_EXCLUDED:
-				/* If non-leader is excluded, handle it
-				 * as 'prepared'. If other non-leaders
-				 * also 'prepared' then related DTX is
-				 * committable. Fall through.
+				/*
+				 * If non-leader is excluded, handle it as 'prepared'. If other
+				 * non-leaders are also 'prepared' then related DTX maybe still
+				 * committable or 'corrupted'. The subsequent DTX resync logic
+				 * will handle related things, see dtx_verify_groups().
+				 *
+				 * Fall through.
 				 */
 			case DTX_ST_PREPARED:
 				if (dra->dra_result == 0 ||
 				    dra->dra_result == DTX_ST_CORRUPTED)
-					dra->dra_result = drr->drr_result;
+					dra->dra_result = DTX_ST_PREPARED;
 				break;
 			case DTX_ST_CORRUPTED:
 				if (dra->dra_result == 0)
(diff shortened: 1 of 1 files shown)


src/dtx/dtx_rpc.c
```diff
@@ -301,15 +301,18 @@ dtx_req_list_cb(void **args)
...
+				 * If non-leader is excluded, handle it as 'prepared'. If other
+				 * non-leaders are also 'prepared' then related DTX maybe still
+				 * committable or 'corrupted'. The subsequent DTX resync logic
+				 * will handle related things, see dtx_verify_groups().
+				 *
+				 * Fall through.
 				 */
 			case DTX_ST_PREPARED:
 				if (dra->dra_result == 0 ||
 				    dra->dra_result == DTX_ST_CORRUPTED)
```
NiuYawei, line 314: DTX_ST_CORRUPTED will be turned into DTX_ST_PREPARED as well?

src/dtx/dtx_rpc.c
```diff
@@ -506,8 +509,8 @@ btr_ops_t dbtree_dtx_cf_ops = {
 #define DTX_CF_BTREE_ORDER	20
 
 static int
-dtx_classify_one(struct ds_pool *pool, daos_handle_t tree, d_list_t *head,
-		 int *length, struct dtx_entry *dte, int count, d_rank_t my_rank, uint32_t my_tgtid)
+dtx_classify_one(struct ds_pool *pool, daos_handle_t tree, d_list_t *head, int *length,
+		 struct dtx_entry *dte, int count, d_rank_t my_rank, uint32_t my_tgtid, int opc)
```
NiuYawei, line 513: I didn't see 'opc' is used.

src/dtx/dtx_rpc.c
Nasf-Fan: Yes, generally, if a DTX entry is marked as 'CORRUPTED', it means that former DTX resync logic cannot find enough DTX participants to decide the DTX status. Now, in this cycle of DTX resync, we have more participants to claim the DTX status as 'PREPARED', then it is possible to decide its status this time. But if still cannot (processed via dtx_verify_groups()), then it will be handled as 'CORRUPTED' again, harmless.

src/dtx/dtx_rpc.c
Nasf-Fan: I will refresh the patch to remove it.

NiuYawei (approved):

liuxuezhao (approved):

Nasf-Fan:
Ping @daos-stack/daos-gatekeeper , thanks!

PR #9633, opened 2022-07-06
DAOS-11063 vos: check and reset NONEXIST in iter_next and probe

wangdi1:
Only reset nonexist error after iter_next and probe, to keep the
return value from other callback in vos_iterate_interal().

Signed-off-by: Di Wang <di.wang@intel.com>

PR #10493, opened 2022-10-05
DAOS-11063 vos: check and reset NONEXIST in iter_next and probe (#9633)

wangdi1:
Only reset nonexist error after iter_next and probe, to keep the return value from other callback in vos_iterate_interal().


jolivier23 (approved):
clean cherry pick ?

PR #10496, opened 2022-10-05
DAOS-11724 test: pool/svc.py - Update the number of ranks to stop

shimizukko:
In pool/svc.py, we stop 0, 1, or 2 ranks based on the expected number
of PS replicas. However, there's a limit for the number of ranks we can
stop, which are:

Number of PS replicas - Number of ranks we can stop
1 - 0
3 - 1
5 - 2

We were stopping 2 ranks when the number of PS replicas was 3. Update
the test so that we only stop 1 rank; and stop 2 ranks only when the
number of PS replicas is 5.

Skip-unit-tests: true
Test-tag: test_pool_svc
Required-githooks: true
Signed-off-by: Makito Kano <makito.kano@intel.com>

### DIFF
--- a/src/tests/ftest/pool/svc.py
+++ b/src/tests/ftest/pool/svc.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 '''
   (C) Copyright 2018-2022 Intel Corporation.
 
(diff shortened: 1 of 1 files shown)


daltonbohning (approved):

src/tests/ftest/pool/svc.py
```diff
@@ -141,23 +145,23 @@ def test_pool_svc(self):
...
-                    self.log.info(error)
-                    self.fail(
-                        "Error stopping a pool non-leader - "
-                        "DaosServerManager.stop_ranks([{}])".format(non_leader))
-
-                self.pool.wait_for_rebuild(to_start=True, interval=1)
-                self.pool.wait_for_rebuild(to_start=False, interval=1)
-                # Verify the pool leader has not changed
-                self.check_leader(pool_leader, False)
+                if svc_params[1] == 5:
```
daltonbohning, line 148: Just for robustness, you might consider
```suggestion
                if svc_params[1] >= 5:
```

daltonbohning (approved):

liw (approved):

PR #10506, opened 2022-10-06
DAOS-11587 bio: sysdb changes for metadata on ssd

wangshilong:
1. move init thread from main xstream to sys xstream
2. first detected meta SSD will be selected as sys device to store
SMD, vos db pool blobid is stored as 'root' blob.
3. VOS DB creation/open won't rely on SMD any more.
4. Meta/WAL blob might reside in the same blobstore as Data, 
change blob cluster size from 1GB to 32MB to simplify codes.
5. Fix two minors issues for iod_dma_completion()

Required-githooks: true


src/bio/bio_context.c
```diff
@@ -576,22 +610,31 @@ bio_blob_open(struct bio_io_context *ctxt, bool async)
...
+		rc = smd_pool_get_blob(ctxt->bic_pool_id, xs_ctxt->bxc_tgt_id,
+				       &blob_id);
+		if (rc != 0) {
+			D_ERROR("Failed to find blobID for xs:%p, pool:"DF_UUID", tgt:%d\n",
+				xs_ctxt, DP_UUID(ctxt->bic_pool_id), xs_ctxt->bxc_tgt_id);
+			rc = -DER_NONEXIST;
+			goto out_free;
+		}
+	} else if (is_sys){
+		spdk_bs_get_super(bbs->bb_bs, show_bs_cb, bma);
```
NiuYawei, line 631: Is this a synchronous operation?

src/bio/bio_context.c
```diff
@@ -576,22 +610,31 @@ bio_blob_open(struct bio_io_context *ctxt, bool async)
...
+	if (!is_sys || !bio_is_meta_on_ssd_configured()) {
+		rc = smd_pool_get_blob(ctxt->bic_pool_id, xs_ctxt->bxc_tgt_id,
+				       &blob_id);
+		if (rc != 0) {
+			D_ERROR("Failed to find blobID for xs:%p, pool:"DF_UUID", tgt:%d\n",
+				xs_ctxt, DP_UUID(ctxt->bic_pool_id), xs_ctxt->bxc_tgt_id);
+			rc = -DER_NONEXIST;
+			goto out_free;
+		}
+	} else if (is_sys){
```
NiuYawei, line 630: Why we need to check 'is_sys' here?

src/bio/bio_context.c
```diff
@@ -672,7 +717,7 @@ __bio_ioctxt_open(struct bio_io_context **pctxt, struct bio_xs_context *xs_ctxt,
 	}
 
 	ctxt->bic_xs_blobstore = bxb;
-	rc = bio_blob_open(ctxt, false);
+	rc = bio_blob_open(ctxt, false, flags & BIO_MC_FL_SYSDB);
```
NiuYawei, line 720: So the super blob could be opened twice? One for meta blob open, the other for WAL blob open? I'm wondering if I should change the WAL design a bit to support Meta & WAL on same blob, so that we could put meta & WAL blob for sysdb on the super blob.

src/bio/bio_xstream.c
```diff
@@ -21,6 +21,7 @@
 #include <spdk/blob.h>
 #include "bio_internal.h"
 #include <daos_srv/smd.h>
+#include <daos_srv/vos.h>
```
NiuYawei, line 24: Why do we need vos.h in BIO?

src/bio/bio_xstream.c
```diff
@@ -1463,23 +1569,50 @@ bio_xsctxt_alloc(struct bio_xs_context **pctxt, int tgt_id, bool self_polling)
...
+			if (rc)
+				goto out;
+			rc = init_xs_blobstore_ctxt(ctxt, tgt_id, SMD_DEV_TYPE_WAL);
+			if (rc)
+				goto out;
+		}
+		if (use_sysdb)
+			rc = vos_db_init(db_path);
+		else
+			rc = vos_db_init_ex(db_path, "self_db", true, true);
```
NiuYawei, line 1587: ah, I see why you include vos.h. These functions should not be called in BIO.

src/bio/bio_xstream.c
```diff
@@ -892,7 +891,34 @@ create_bio_bdev(struct bio_xs_context *ctxt, const char *bdev_name,
...
+		D_ERROR("invalid numeric value: %s (name %s)\n", dst, name);
+		return -DER_INVAL;
+	}
+
+	if (value & (~NVME_ROLE_ALL))
+		return -DER_INVAL;
+
+	/* AIO */
+	if (value == 0)
+		value = NVME_ROLE_ALL;
```
NiuYawei, line 915: This is temporary workaround for testing? If so, it needs be removed before landing.

src/bio/bio_xstream.c
```diff
@@ -908,10 +934,50 @@ init_bio_bdevs(struct bio_xs_context *ctxt)
...
+
+		*sys_bdev = bdev;
+		break;
 	}
+
+	return rc;
+}
+
+static int
+init_bio_nosys_bdevs(struct bio_xs_context *ctxt, struct spdk_bdev *except_bdev)
```
NiuYawei, line 957: I don't see why do we need separate init_bio_sys_bdev() and init_bio_nosys_bdevs().

src/bio/bio_xstream.c
```diff
@@ -971,6 +1037,8 @@ fini_bio_bdevs(struct bio_xs_context *ctxt)
 		d_list_del_init(&d_bdev->bb_link);
 		destroy_bio_bdev(d_bdev);
 	}
+	vos_db_fini();
```
NiuYawei, line 1040: Please, no VOS in BIO.

src/bio/bio_xstream.c
```diff
@@ -1038,33 +1106,55 @@ get_bio_blobstore(struct bio_blobstore *bb, struct bio_xs_context *ctxt)
...
+{
+	switch (st) {
+	case SMD_DEV_TYPE_DATA:
+		return NVME_ROLE_DATA;
+	case SMD_DEV_TYPE_META:
+		return NVME_ROLE_INDEX;
+	case SMD_DEV_TYPE_WAL:
+		return NVME_ROLE_WAL;
+	default:
+		D_ASSERT(1);
```
NiuYawei, line 1120: What's this D_ASSERT(1) for? You mean D_ASSERT(0)?

src/bio/bio_xstream.c
```diff
@@ -1038,33 +1106,55 @@ get_bio_blobstore(struct bio_blobstore *bb, struct bio_xs_context *ctxt)
...
-				  bb_link);
-	lowest_tgt_cnt = chosen_bdev->bb_tgt_cnt;
-
 	/*
 	 * Traverse the list and return the device with the least amount of
 	 * mapped targets.
 	 */
 	d_list_for_each_entry(d_bdev, &nvme_glb.bd_bdevs, bb_link) {
-		if (d_bdev->bb_tgt_cnt < lowest_tgt_cnt) {
+		rc = bdev_name2roles(d_bdev->bb_name);
```
NiuYawei, line 1143: Don't use bdev name generated by SPDK, use the base name from daos_nvme.conf.

src/bio/bio_xstream.c
```diff
@@ -1108,18 +1198,31 @@ find_xs_bio_dev(struct bio_xs_context *ctxt, int tgt_id, enum smd_dev_type st,
 	 * Lookup device mapped to @tgt_id and @st in the per-server
 	 * metadata, if found, create blobstore on the mapped device.
 	 */
+	if (tgt_id != -1) {
```
NiuYawei, line 1201: "-1" means sys 0 xstream? You'd use macro for it.

src/include/daos_srv/vos.h
```diff
@@ -243,18 +243,19 @@ vos_self_fini(void);
...
- * \param scm_sz [IN]	Size of SCM for the pool
- * \param blob_sz[IN]	Size of blob for the pool
- * \param flags [IN]	Pool open flags (see vos_pool_open_flags)
- * \param poh	[OUT]	Returned pool handle if not NULL
+ * \param path		[IN]	Path of the memory pool
+ * \param uuid		[IN]    Pool UUID
+ * \param scm_sz	[IN]	Size of SCM for the pool
+ * \param blob_sz	[IN]	Size of blob for the pool
+ * \param flags		[IN]	Pool open flags (see vos_pool_open_flags)
+ * \param is_sys	[IN]	sys vos pool or not
```
NiuYawei, line 251: We can tell it by 'path' or a new flag, this extra parameter isn't necessary.

src/include/daos_srv/vos.h
```diff
@@ -281,16 +282,18 @@ vos_pool_destroy(const char *path, uuid_t uuid);
...
  *
- * \param path	 [IN]	Path of the memory pool
- * \param uuid	 [IN]	Pool UUID
- * \param flags  [IN]	Pool open flags (see vos_pool_open_flags)
- * \param poh	 [OUT]	Returned pool handle
+ * \param path		[IN]	Path of the memory pool
+ * \param uuid		[IN]	Pool UUID
+ * \param flags		[IN]	Pool open flags (see vos_pool_open_flags)
+ * \param metrics	[IN]	Metrics
+ * \param is_db		[IN]	Pool is for sys db or not
```
NiuYawei, line 289: Ditto.

NiuYawei (changes requested):

src/bio/bio_context.c
wangshilong: This should be not needed.

src/bio/bio_context.c
```diff
@@ -574,22 +615,38 @@ bio_blob_open(struct bio_io_context *ctxt, bool async)
...
+					       &blob_id);
+			if (rc != 0) {
+				D_ERROR("Failed to find blobID for xs:%p, pool:"DF_UUID", tgt:%d\n",
+					xs_ctxt, DP_UUID(ctxt->bic_pool_id), xs_ctxt->bxc_tgt_id);
+				rc = -DER_NONEXIST;
+				goto out_free;
+			}
+		} else {
+			ba->bca_inflights = 1;
+			spdk_bs_get_super(bbs->bb_bs, get_super_cb, bma);
```
NiuYawei, line 638: Minor: spdk_bs_get_super() is a synchronous call, it just get the blob ID from spdk_blob_store->super_blob.

src/bio/bio_xstream.c
```diff
@@ -25,7 +25,8 @@
 /* These Macros should be turned into DAOS configuration in the future */
 #define DAOS_MSG_RING_SZ	4096
 /* SPDK blob parameters */
-#define DAOS_BS_CLUSTER_SZ	(1ULL << 30)	/* 1GB */
+#define DAOS_BS_DATA_CLUSTER_SZ	(1ULL << 30)	/* 1GB */
+#define DAOS_BS_META_CLUSTER_SZ	(1ULL << 25)	/* 32MB */
```
NiuYawei, line 29: The cluster size is per blobstore, so I think we can't use mixed cluster size when meta blob and data blob resides in same device/blobstore.

src/bio/bio_xstream.c
```diff
@@ -466,7 +458,7 @@ xs_poll_completion(struct bio_xs_context *ctxt, unsigned int *inflights,
 
 struct spdk_blob_store *
 load_blobstore(struct bio_xs_context *ctxt, char *bdev_name, uuid_t *bs_uuid,
-	       bool create, bool async,
+	       bool create, bool async, unsigned int roles,
```
NiuYawei, line 461: Why don't you pass 'cluster_size' directly? Anyway, I don't think we should use different cluster size since the data blob and meta blob could reside in same device/blobstore.

src/engine/sched.c
```diff
@@ -1149,7 +1149,7 @@ should_enqueue_req(struct dss_xstream *dx, struct sched_req_attr *attr)
 		return false;
 
 	/* For VOS xstream only */
-	return dx->dx_main_xs;
+	return dss_xstream_has_context(dx);
```
NiuYawei, line 1152: It's "For VOS xstream only" so far.

src/engine/sched.c
```diff
@@ -1882,7 +1882,7 @@ watchdog_enabled(struct dss_xstream *dx)
 	if (sched_unit_runtime_max == 0)
 		return false;
 
-	return dx->dx_xs_id == 0 || (sched_watchdog_all && dx->dx_main_xs);
+	return dx->dx_xs_id == 0 || (sched_watchdog_all && dss_xstream_has_context(dx));
```
NiuYawei, line 1885: I don't think we should change anything related to watchdog in this PR.

src/engine/srv.c
```diff
@@ -454,28 +454,41 @@ dss_srv_handler(void *arg)
...
 	}
 
-	if (dx->dx_main_xs) {
+	if (dss_xstream_has_context(dx)) {
 		ABT_thread_attr attr;
 
 		/* Initialize NVMe context for main XS which accesses NVME */
-		rc = bio_xsctxt_alloc(&dmi->dmi_nvme_ctxt, dmi->dmi_tgt_id, false);
+		if (dx->dx_xs_id == 0) {
+			rc = bio_sys_xsctxt_alloc(&dmi->dmi_nvme_ctxt, dmi->dmi_tgt_id, false);
```
NiuYawei, line 462: This won't work, since SMD is required when allocating per-xstream context for sys xstream. (To store which device is assigned to sys xstream), so for sys xstream, it should do:
- bio_glbctxt_alloc();
- Initialized sysdb & setup smd;
- bio_xsctxt_alloc();

src/bio/bio_xstream.c
```diff
@@ -1384,13 +1454,89 @@ bio_xsctxt_free(struct bio_xs_context *ctxt)
...
+	if (ctxt->bxc_dma_buf != NULL) {
+		dma_buffer_destroy(ctxt->bxc_dma_buf);
+		ctxt->bxc_dma_buf = NULL;
+	}
+
+	D_FREE(ctxt);
+}
+
+int
+bio_init_xs_data_blobstore_ctxt(struct bio_xs_context *ctxt, int tgt_id)
```
NiuYawei, line 1525: I don't quite see why do we need this.

NiuYawei (changes requested):

src/bio/bio_xstream.c
wangshilong: if meta blob and data blob resides in the same device, we will use DATA_CLUSTER_SZ

src/engine/srv.c
wangshilong: Sure

src/engine/sched.c
wangshilong: OK

src/engine/sched.c
wangshilong: sure.

src/bio/bio_xstream.c
wangshilong: This is used for testing: vos_self_nvme_init, it will try to use sys xs for data testing, however in normal case, we don't allocate datablobstore for sys xs before, or maybe i should add a flag to bio_xsctxt_alloc(), indicate we should always allocate DATA blobstore even it is sys xs.

NiuYawei: I don't think we need a separate function for this. sys xstream for engine could be differentiated from standalone VOS by target ID.

src/bio/bio_xstream.c
NiuYawei: That'll be too large for sysdb and RDB. I think we should always use 32MB cluster size to keep it simple.

src/engine/srv.c
wangshilong: Above won't work, because init sysdb need sys xstream blobstore prepared, and allocating context for sys xstream don't need SMD. so right logic shall be:

1. bio_glbctxt_alloc()
    init meta/wal blobstore properly---> this don't rely on SMD.
2. init sysdb && setup smd.

wangshilong: problem is if we want to use SMD in sys xs for RDB, we might need do that after step 2(SMD is ready)

src/bio/bio_internal.h
```diff
@@ -19,7 +19,7 @@
 
 #define BIO_DMA_PAGE_SHIFT	12	/* 4K */
 #define BIO_DMA_PAGE_SZ		(1UL << BIO_DMA_PAGE_SHIFT)
-#define BIO_XS_CNT_MAX		48	/* Max VOS xstreams per blobstore */
+#define BIO_XS_CNT_MAX		DSS_MAX_VOS_TGT_CNT	/* Max VOS xstreams per blobstore */
```
NiuYawei, line 22: DAOS server engine is on the top layer, we can't use engine stuff in BIO. This restriction could be defined in BIO or SMD.

src/bio/bio_xstream.c
```diff
@@ -1162,7 +1166,7 @@ find_xs_bio_dev(struct bio_xs_context *ctxt, int tgt_id, enum smd_dev_type st,
 	 * Lookup device mapped to @tgt_id and @st in the per-server
 	 * metadata, if found, create blobstore on the mapped device.
 	 */
-	if (tgt_id != SYS_TGT_ID) {
+	if (tgt_id != DSS_SYS_TGT_ID) {
```
NiuYawei, line 1169: This should be defined in SMD or BIO, and standalone VOS needs be considered too. We can leave standalone VOS support in follow-in PR though.

src/bio/bio_xstream.c
```diff
@@ -766,6 +789,11 @@ create_bio_bdev(struct bio_xs_context *ctxt, const char *bdev_name,
 	}
 
 	D_INIT_LIST_HEAD(&d_bdev->bb_link);
+	rc = bdev_name2roles(bdev_name);
+	if (rc < 0)
+		goto error;
```
NiuYawei, line 794: so we assume the daos_nvme.conf will always be re-generated after upgrading?

src/bio/bio_xstream.c
```diff
@@ -1038,47 +1043,100 @@ get_bio_blobstore(struct bio_blobstore *bb, struct bio_xs_context *ctxt)
...
+	default:
+		D_ASSERT(0);
+		return NVME_ROLE_DATA;
+	}
+}
+
+static bool
+is_bdev_match(struct bio_bdev *d_bdev, unsigned int bdev_roles)
+{
+	return d_bdev->bb_roles == 0 || d_bdev->bb_roles & bdev_roles;
```
NiuYawei, line 1065: Why bb_roles == 0 means a match?

src/bio/bio_xstream.c
```diff
@@ -1282,8 +1361,9 @@ bio_blobstore_free(struct bio_xs_blobstore *bxb, struct bio_xs_context *ctxt)
 	if (is_bbs_owner(ctxt, bbs))
 		bio_fini_health_monitoring(bbs);
 }
+
 /*
- * Finalize per-xstream NVMe context and SPDK env.
+ * Finalize sys xstream NVMe context and SPDK env.
```
NiuYawei, line 1366: This comment change should be reverted.

src/bio/bio_xstream.c
```diff
@@ -1472,25 +1554,34 @@ bio_xsctxt_alloc(struct bio_xs_context **pctxt, int tgt_id, bool self_polling)
 		}
 	}
 
-	/* Initialize per-xstream blobstore context */
+	/* Initialize main xstream blobstore context */
```
NiuYawei, line 1557: This comment should be reverted.

src/engine/srv_internal.h
```diff
@@ -326,17 +326,24 @@ dss_xs2tgt(int xs_id)
...
+		if (xs_id >= dss_sys_xs_nr + dss_tgt_nr)
+			return DSS_INVALID_TGT_ID;
 		return xs_id - dss_sys_xs_nr;
 	}
 
-	if (xs_id < dss_sys_xs_nr)
-		return -1;
 	return (xs_id - dss_sys_xs_nr) /
 	       (dss_tgt_offload_xs_nr / dss_tgt_nr + 1);
 }
```
NiuYawei, line 341: I don't see why this code needs be changed.

src/engine/srv_internal.h
```diff
@@ -326,17 +326,24 @@ dss_xs2tgt(int xs_id)
...
 
+static inline bool
+dss_xstream_has_context(struct dss_xstream *dx)
```
NiuYawei, line 344: dss_xstream_has_nvme()? Also, shouldn't we consider the tiers passed by control plane to avoid create nvme context on sys xstream when pmem tier is provided?

src/include/daos_srv/bio.h
```diff
@@ -447,15 +452,27 @@ enum {
...
+/*
+ * Initialize Sys xstream data NVMe context.
+ *
+ * \param[IN] ctxt		Per-xstream NVMe context to be returned
+ * \param[IN] tgt_id		Target ID (mapped to a VOS xstream)
+ *
+ * \returns		Zero on success, negative value on error
+ */
+int
+bio_init_xs_data_blobstore_ctxt(struct bio_xs_context *ctxt, int tgt_id);
```
NiuYawei, line 474: This should be removed.

src/include/daos_srv/daos_engine.h
```diff
@@ -51,6 +51,13 @@ extern unsigned int	 dss_instance_idx;
 /** Bypass for the nvme health check */
 extern bool		 dss_nvme_bypass_health_check;
 
+/** Max number of vos targets per engine */
+#define			 DSS_MAX_VOS_TGT_CNT	96
+/* System xstream target ID */
+#define			 DSS_SYS_TGT_ID		1024
+/* Invalid target id */
+#define			 DSS_INVALID_TGT_ID	-1
```
NiuYawei, line 59: Don't define these in daos engine, define them in BIO or SMD.

src/vos/vos_common.c
```diff
@@ -705,19 +707,33 @@ vos_self_nvme_init(const char *vos_path, uint32_t tgt_id)
...
 	if (rc)
 		goto out;
 
-	self_mode.self_nvme_init = true;
+	if (use_sys_db)
+		rc = vos_db_init(vos_path);
+	else
+		rc = vos_db_init_ex(vos_path, "self_db", true, true);
+	if (rc)
+		goto out;
```
NiuYawei, line 726: What's the purpose of moving these from vos_self_init() to here?

src/bio/bio_xstream.c
```diff
@@ -1472,25 +1554,34 @@ bio_xsctxt_alloc(struct bio_xs_context **pctxt, int tgt_id, bool self_polling)
...
 		if (rc)
 			goto out;
 	}
 
+	if (tgt_id == DSS_SYS_TGT_ID) {
+		rc = bio_sys_smd_init();
+		if (rc)
+			goto out;
+	}
+
```
NiuYawei, line 1577: I didn't see SMD updated after assign blobstore to sys xstream.

NiuYawei (changes requested):

src/bio/bio_xstream.c
wangshilong: I haven't got conclusion from Tom yet,  just assume we will for simple reason.

src/bio/bio_xstream.c
wangshilong: Originally this is for AIO testing purpose, one case is roles might be 0 if metadata on SSD not enabled? maybe we should make roles 0 match DATA instead.

src/bio/bio_xstream.c
wangshilong: OK!

src/bio/bio_xstream.c
wangshilong: sure

src/bio/bio_xstream.c
wangshilong: Sure

src/bio/bio_xstream.c
wangshilong: You mean we also need write SMD for sys stream mapping? i think we discussed that you mentioned that this will  be not used even we updated SMD for sys xs.

src/vos/vos_common.c
wangshilong: will move back, this is history reason.

src/include/daos_srv/daos_engine.h
wangshilong: ok!

src/include/daos_srv/bio.h
wangshilong: sure.

src/engine/srv_internal.h
wangshilong: Sure, will add a Todo here.

NiuYawei (approved):

jolivier23 (approved):

src/bio/bio_xstream.c
```diff
@@ -1474,23 +1568,32 @@ bio_xsctxt_alloc(struct bio_xs_context **pctxt, int tgt_id, bool self_polling)
...
+		/* No Meta/WAL blobstore if Metadata on SSD is not configured */
+		if (st != SMD_DEV_TYPE_DATA && !bio_is_meta_on_ssd_configured())
+			break;
+
 		rc = init_xs_blobstore_ctxt(ctxt, tgt_id, st);
 		if (rc)
 			goto out;
 	}
 
+	if (tgt_id == BIO_SYS_TGT_ID) {
```
NiuYawei, line 1584: Minor: It's the init thread being responsible for initializing/finalizing SMD, this could be revised when we make patch to support standalone VOS.

I think the standalone VOS support will be similar to sys xstream support, the only difference is that for sys xstream, we don't assign data blobstore, but for standalone VOS, we should assign blobstores according to the roles passed by bdev name.

src/engine/srv_internal.h
```diff
@@ -339,4 +339,14 @@ dss_xs2tgt(int xs_id)
...
 }
 
+/*
+ * FIXME: control plane shall pass tier information to
+ * avoid create nvme context if Metadata on SSD is not enabled.
+ */
+static inline bool
+dss_xstream_has_nvme(struct dss_xstream *dx)
+{
+	return dx->dx_main_xs != 0 || dx->dx_xs_id == 0;
```
NiuYawei, line 349: So the per-xstream NVMe context will be created on sys xstream after this PR, I'm not sure if it could cause any test failures, since sys xstream will consume some reserved hugepages too.

It reminds me that the hugepage nr auto calculation on control plane needs be adjusted for meta data on SSD.

src/vos/vos_common.c
```diff
@@ -678,7 +679,7 @@ vos_self_nvme_fini(void)
 #define VOS_NVME_NR_TARGET	1
 
 static int
-vos_self_nvme_init(const char *vos_path, uint32_t tgt_id)
+vos_self_nvme_init(const char *vos_path, uint32_t tgt_id, bool use_sys_db)
```
NiuYawei, line 682: Minor: this 'use_sys_db' flag isn't used, it could be removed in follow-on patch.

NiuYawei (approved):

src/bio/bio_xstream.c
wangshilong: Sure

src/engine/srv_internal.h
wangshilong: Indeed

src/vos/vos_common.c
wangshilong: I missed your comments since there is no email notifications, i will remove this in this PR.

NiuYawei (approved):

wangshilong:
@jolivier23 would you consider this PR?  standalone vos and integration parts will be pushed as follow-on PR.
Thanks!

src/bio/bio_context.c
```diff
@@ -492,8 +505,18 @@ bio_blob_create(uuid_t uuid, struct bio_xs_context *xs_ctxt, uint64_t blob_sz,
...
 
-		rc = smd_pool_add_tgt(uuid, xs_ctxt->bxc_tgt_id, ba->bca_id,
-				      blob_sz);
+		if (!(flags & BIO_MC_FL_SYSDB) || !bio_is_meta_on_ssd_configured()) {
+			rc = smd_pool_add_tgt(uuid, xs_ctxt->bxc_tgt_id, ba->bca_id,
+					      blob_sz);
+		} else if (st == SMD_DEV_TYPE_META) {
+			ba->bca_inflights = 1;
+			spdk_bs_set_super(bbs->bb_bs, ba->bca_id,
+					  set_super_cb, &bma);
```
NiuYawei, line 514: Should we set INVALID super blob ID when we delete super blob? Or it's done by SPDK already?

wangshilong: As discussed, now bio won't check super exist before set, and spdk_bs_set_super() just overwrite previous super blob id, so it is fine to not set invalid when delete super blob.

jolivier23 (approved):

PR #10513, opened 2022-10-07
DAOS-11870 object: log data inconsistency as D_WARN

Nasf-Fan:
That will be helpful for log collection.


Nasf-Fan:
Not need it any longer.

PR #10521, opened 2022-10-07
DAOS-11695 test: Add test users to CI env

daltonbohning:
Skip-test: true
Skip-build: true
Doc-only: true
Test-tag: test_harness_pre_launch_users

Required-githooks: true

Signed-off-by: Dalton Bohning <dalton.bohning@intel.com>

### Gatekeeper:

* [x] You are the appropriate gatekeeper to be landing the patch.
* [x] The PR has 2 reviews by people familiar with the code.
* [x] Any appropriate watchers have had a chance to review the PR.
* [x] Review comments are sufficiently resolved, particularly by prior reviewers that requested
      changes.
* [x] Githooks were used. If not, there is a sufficient reason to move forward.  Check copyrights
      if githooks were not used.
* [x] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future
      PRs.
* [x] All builds have passed.  Check non-required builds to ensure they don't indicate any problem
* [x] No new NLT or valgrind warnings.  Check the classic view. This step should only matter if
* [x] Ensure sufficent testing is done.   Check feature pragmas and test tags. Check that tests
* [x] Quick-build or Quick-functional is not used.
* [x] Pay attention to PRs that may affect compatibility between versions and ensure it has been
      addressed.
* [x] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If
* [x] Check the commit message when landing. Check the standard


daltonbohning:
@brianjmurrell @phender Before running this on CI nodes, I'd like to sanity check the commands. These are the commands ultimately ran on each test node:
```
# Create two groups
groupadd daos_test_groupx -r
groupadd daos_test_groupy -r

# Delete users if pre-existing
if $(id -u daos_test_userx1 &> /dev/null); then userdel -f -r daos_test_userx1; fi
if $(id -u daos_test_userx2 &> /dev/null); then userdel -f -r daos_test_userx2; fi
if $(id -u daos_test_usery1 &> /dev/null); then userdel -f -r daos_test_usery1; fi

# Create users
useradd -m -g daos_test_groupx daos_test_userx1
useradd -m -g daos_test_groupx daos_test_userx2
useradd -m -g daos_test_groupy daos_test_usery1
```

src/tests/ftest/user_setup.py
```diff
@@ -0,0 +1,137 @@
...
+    parser.add_argument(
+        "-s", "--sudo",
+        action="store_true",
+        help="Run all commands with privileges")
+
+    args = parser.parse_args()
+    logging.info("Arguments: %s", args)
+
+    nodes = NodeSet(args.nodes or get_local_host())
+    users = args.users.split(',') if args.users else None
```
phender, line 125: This should either default to empty list (so you don't have a problem on line 67) or the progarm should quit if no users are specified:
```suggestion
    users = args.users.split(',') if args.users else []
```

Alternatively, the string `split()` could be avoided by defining the `--user` argument with `nargs="*"` and then `required=True` should ensure at least one user was provided.

src/tests/ftest/user_setup.py
```diff
@@ -0,0 +1,137 @@
...
+        type=str,
+        help="Comma-separated list of nodes to create users on")
+    parser.add_argument(
+        '-u', '--users',
+        type=str,
+        required=True,
+        help="Comma-separated list of user accounts to create")
+    parser.add_argument(
+        "-g", "--group",
+        type=str,
```
phender, line 110: Should this have `required=True` set?

src/tests/ftest/user_setup.py
```diff
@@ -0,0 +1,137 @@
...
+        help="Group to add users to. Will be created if non-existent")
+    parser.add_argument(
+        "-d", "--dir",
```
phender, line 113: Should this be required too or is there a case where it makes sense to create a group without creating the user?

src/tests/ftest/user_setup.py
```diff
@@ -0,0 +1,137 @@
...
+    '''
+    logging.basicConfig(
+        format="%(asctime)s %(levelname)-5s %(message)s",
+        datefmt=r"%Y/%m/%d %I:%M:%S", level=logging.DEBUG)
+
+    parser = argparse.ArgumentParser(prog="user_setup.py")
+
+    parser.add_argument(
+        '-n', '--nodes',
+        type=str,
```
phender, line 101: Optionally the `default=get_local_host()` could be used here with the benefit of appearing in the help

src/tests/ftest/scripts/main.sh
```diff
@@ -190,6 +190,20 @@ fi
 
 export DAOS_APP_DIR=${DAOS_APP_DIR:-$DAOS_TEST_SHARED_DIR}
 
+# Create two test users in the same group
+if ! ./user_setup.py -n "${TEST_NODES}" -g daos_test_groupx -u daos_test_userx1,daos_test_userx2 -s; then
+    exit "${PIPESTATUS[0]}"
+else
+    rc=0
```
phender, line 197: This should not be required
```suggestion
```

src/tests/ftest/scripts/main.sh
```diff
@@ -190,6 +190,20 @@ fi
...
+fi
+
+# Create a third test user in a different group
+if ! ./user_setup.py -n "${TEST_NODES}" -g daos_test_groupy -u daos_test_usery1 -s; then
+    exit "${PIPESTATUS[0]}"
+else
+    rc=0
```
phender, line 204: This should not be required
```suggestion
```

src/tests/ftest/scripts/main.sh
```diff
@@ -190,6 +190,20 @@ fi
 
 export DAOS_APP_DIR=${DAOS_APP_DIR:-$DAOS_TEST_SHARED_DIR}
 
+# Create two test users in the same group
+if ! ./user_setup.py -n "${TEST_NODES}" -g daos_test_groupx -u daos_test_userx1,daos_test_userx2 -s; then
```
phender, line 194: Optional readability improvement:
```suggestion
if ! ./user_setup.py -n "${TEST_NODES}" -g daos_test_group_x -u daos_test_user_x1,daos_test_user_x2 -s; then
```

src/tests/ftest/scripts/main.sh
```diff
@@ -190,6 +190,20 @@ fi
...
+    exit "${PIPESTATUS[0]}"
+else
+    rc=0
+fi
+
+# Create a third test user in a different group
+if ! ./user_setup.py -n "${TEST_NODES}" -g daos_test_groupy -u daos_test_usery1 -s; then
```
phender, line 201: Optional readability improvement:
```suggestion
if ! ./user_setup.py -n "${TEST_NODES}" -g daos_test_group_y -u daos_test_user_y1 -s; then
```

phender (changes requested):

src/tests/ftest/user_setup.py
daltonbohning: Maybe. At this point, I'm mostly trying to make sure I have the commands in order, then I plan to make the script itself robust

src/tests/ftest/user_setup.py
daltonbohning: `-d` is for when we don't want the user's home directory to be the default. I found this necessary on boro, since I don't have permission to create users in `/home` (it's shared across the system). So I use `-d /tmp`.
In CI, I don't what we should do

src/tests/ftest/user_setup.py
daltonbohning: Thanks. I'll clean up the interface before moving out of draft

src/tests/ftest/scripts/main.sh
daltonbohning: Agree. Thanks

src/tests/ftest/user_setup.py
ashleypittman: nargs="*" means a list of zero or more, nargs="+" means a list of one or more which I think is what you're looking for here.

src/tests/ftest/user_setup.py
```diff
@@ -0,0 +1,137 @@
...
+"""
+
+import argparse
+import logging
+import os
+import sys
+
+from ClusterShell.NodeSet import NodeSet
+
+from util.logger_utils import get_console_handler
```
ashleypittman, line 15: I'm somewhat familiar with this error but don't know the solution.

All other python files in src/test/ftest/*.py which import from util hit the same problem, and I've had to suppress it.  The check passes for me locally but fails in github actions. I think part of the reason is there is a python module called "util" but from memory it's present both in my setup and in GHA so that doesn't entirely explain it.

Anyway, you should just suppress this error for now as not a regression and I'll take another look next time I'm doing a parse of the pylint code.


https://github.com/daos-stack/daos/blob/master/src/tests/ftest/config_file_gen.py#L8-L9
https://github.com/daos-stack/daos/blob/master/src/tests/ftest/launch.py#L30
https://github.com/daos-stack/daos/blob/master/src/tests/ftest/process_core_files.py#L14
https://github.com/daos-stack/daos/blob/master/src/tests/ftest/slurm_setup.py#L8

src/tests/ftest/user_setup.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        nodes (NodeSet): nodes on which to create the group
+        group (str): the group to create
+        sudo (bool): whether to execute commands with sudo
+
+    Returns:
+        bool: whether the command was successful on all nodes
+    '''
+    command = _sudo(sudo, 'groupadd {} -f'.format(group))
```
ashleypittman, line 50: This file is python3 right, so should be using f-strings?

src/tests/ftest/user_setup.py
```diff
@@ -0,0 +1,137 @@
...
+        if not run_remote(LOGGER, nodes, command).passed:
+            return False
+    return True
+
+
+def main():
+    '''Set up test env users.
+
+    Returns:
+        int: 0 on success. positive int on failure
```
ashleypittman, line 91: nitpick, this code could in theory return a positive int in a way that wouldn't cause a failure.  The exit code of a program is a signed char with negative values reserved for signals so only 1-127 are valid "failure" return values for this function.  If this function returned 256 for example then when cast to a signed int it would be 0 and the check would pass.  Likewise if it was 501 then that would translate to -11 or SEGV.

This might seem academic and I'm not requesting changes here but we've seen exactly this issue in IOR where we had a data corruption issue where every byte was wrong and as the buffer was 2^N in size the error count and therefore exit code was a multiple of 256 and IOR exited 0, leading to silent failure and a test that we thought was passing actually failing.

src/tests/ftest/harness/pre_launch.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class HarnessPreLaunchTest(TestWithoutServers):
+    """Harness pre-launch test cases.
+
+    :avocado: recursive
+    """
+
+    def test_harness_pre_launch_users(self):
+        """Verify all expected users are setup correctly.
```
ashleypittman, line 21: This comment should include the name of the script that's supposed to have setup the users correctly, otherwise in years to come someone is going to have to go digging or looking in the git history to see what file was added/edited at the same time.

src/tests/ftest/harness/pre_launch.py
```diff
@@ -0,0 +1,41 @@
...
+
+        :avocado: tags=all
+        :avocado: tags=vm
+        :avocado: tags=harness
+        :avocado: tags=test_harness_pre_launch_users
```
ashleypittman, line 26: Just for my benefit, when will this test get run?

ashleypittman (commented):
Only comment here is should the user/group data be in main.sh or would it be better in user_setup.py itself or a config file.  How does this play with running the tests locally or under ansible.

@knard-intel might like to comment further.

src/tests/ftest/user_setup.py
daltonbohning: > is there a case where it makes sense to create a group without creating the user

At this time, no. But later it would be easy enough to add an option like `--groups-only`

src/tests/ftest/harness/pre_launch.py
daltonbohning: Very good point

src/tests/ftest/user_setup.py
daltonbohning: Since the name of the script is `user_setup.py`, I changed this to a positional and used `nargs='+',`

src/tests/ftest/user_setup.py
daltonbohning: For our purposes, we always want a group, but linux doesn't require a group when creating a user. The default is to put the user in a group named after the user

daltonbohning:
> should the user/group data be in main.sh or would it be better in user_setup.py itself or a config file. How does this play with running the tests locally or under ansible.

@ashleypittman I'm not sure. In general, I think our infrastructure is lacking a clear way to handle this

src/tests/ftest/harness/pre_launch.py
ashleypittman: Can I ask that you make this into a function so that when we have tests that need this setup they can call the function to verify it's there and will fail correctly if it's not.

src/tests/ftest/scripts/main.sh
```diff
@@ -190,6 +190,17 @@ fi
...
+if ! ./user_setup.py -n "${TEST_NODES}" -g daos_test_group_x -s \
+        daos_test_user_x1 daos_test_user_x2; then
+    exit "${PIPESTATUS[0]}"
+fi
+
+# Create a third test user in a different group
+if ! ./user_setup.py -n "${TEST_NODES}" -g daos_test_group_y -s daos_test_user_y1; then
+    exit "${PIPESTATUS[0]}"
+fi
+
```
daltonbohning, line 203: From discussion with Phil, we decided it would be better to move this logic into launch.py so setup is done in other environments as well

src/tests/ftest/util/user_utils.py
```diff
@@ -9,7 +9,7 @@
 import os
 
 # pylint: disable=import-error,no-name-in-module
```
ashleypittman, line 11: You might not need this pylint annotation now.

src/tests/ftest/harness/pre_launch.py
daltonbohning: The users are now created by launch.py, so they will exist when the tests run

daltonbohning: Also added functions in user_utils.py that can be used in test cases to sanity check existence

src/tests/ftest/launch.py
```diff
@@ -29,15 +29,16 @@
 # from util.distro_utils import detect
 # pylint: disable=import-error,no-name-in-module
 from process_core_files import CoreFileProcessing
-from util.logger_utils import get_console_handler, get_file_handler
-from util.results_utils import create_html, create_xml, Job, Results, TestResult
-from util.run_utils import get_local_host, run_local, run_remote, RunException
-from util.user_utils import get_chown_command, get_getent_command, get_groupadd_command, \
-    get_useradd_command, get_userdel_command
+sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), "util"))
+# pylint: disable=import-outside-toplevel
```
ashleypittman, line 33: I see what you're trying to do here but I don't think that's the correct option, and I'd rather changes to the import/pylint setup were done separately.

See also this line, the util directory should be added to pylint automatically.
https://github.com/daos-stack/daos/blob/master/utils/cq/daos_pylint.py#L369

daltonbohning: > I don't think that's the correct option

What do you think the correct option is? The only "proper" way I can think of is to convert the entire ftest directory into a python module and install it. The problem is launch.py is one level up from util. And some of the files it imports have other imports from the util directory. And those imports will fail, because the relative import path is one level up.

> I'd rather changes to the import/pylint setup were done separately

Do you mean separate PRs? Wouldn't that introduce a pylint error in the meantime?

> the util directory should be added to pylint automatically

There was somewhere else in the process flagging this, but I don't remember where.

Anyway, these two lines are a partial cherry-pick from https://github.com/daos-stack/daos/pull/10621/files

src/tests/ftest/launch.py
```diff
@@ -1747,6 +1764,120 @@ def setup_test_files(self, args, yaml_dir):
...
+            LaunchException: if there is an error querying or creating the user
+
+        """
+        user_exists = False
+        try:
+            # Query and verify user
+            logger.info('Querying user %s', user)
+            result = run_remote(logger, hosts, f'id {user}')
+            if not result.passed:
+                logger.info('here1')
```
daltonbohning, line 1821: TODO - remove

daltonbohning:
Work is complete, but this PR conflicts with #10621 - waiting for it to land first

src/tests/ftest/launch.py
```diff
@@ -2225,12 +2357,14 @@ def _setup_test_directory(self, test):
...
         logger.debug("-" * 80)
         test_dir = os.environ["DAOS_TEST_LOG_DIR"]
+        user_dir = os.environ["DAOS_TEST_USER_DIR"]
         logger.debug("Setting up '%s' on %s:", test_dir, test.host_info.all_hosts)
         commands = [
             f"sudo -n rm -fr {test_dir}",
             f"mkdir -p {test_dir}",
             f"chmod a+wr {test_dir}",
             f"ls -al {test_dir}",
+            f"mkdir -p {user_dir}"
```
ashleypittman, line 2367: Is this needed?

src/tests/ftest/launch.py
```diff
@@ -2123,12 +2249,13 @@ def display_disk_space(path):
         except RunException:
             pass
 
-    def prepare(self, test, repeat):
+    def prepare(self, test, repeat, user_create):
```
ashleypittman, line 2252: If this is only called locally then you could rename it to _prepare.

src/tests/ftest/harness/launch_setup.py
```diff
@@ -0,0 +1,50 @@
+"""
+(C) Copyright 2021-2022 Intel Corporation.
```
ashleypittman, line 2:
```suggestion
(C) Copyright 2022 Intel Corporation.
```

src/tests/ftest/launch.py
daltonbohning: Yes, because someone **could** override the path to be anywhere else:
https://github.com/daos-stack/daos/blob/8b4fd08e41d1848a2088aba6b43811412447069a/src/tests/ftest/launch.py#L1155-L1156

And `useradd` doesn't create parent directories.

src/tests/ftest/launch.py
daltonbohning: I'll defer to @phender on whether this should be `prepare` or `_prepare`, but generally I agree with you.

src/tests/ftest/harness/launch_setup.py
daltonbohning: Thanks. Will fix

src/tests/ftest/launch.py
ashleypittman: Ah, I misread this as a duplicate of the mkdir above but I see it's two different directories now.

src/tests/ftest/launch.py
ashleypittman: If it's only called from self.prepare then it should be _prepare, if it's called from elsewhere then the name is correct.  Given that you change the arguments and only change the same object that tells me it's not called from elsewhere.

daltonbohning: I think we aren't in the mindset of always naming "private" functions appropriately, but we should be.

ashleypittman: This is clearly defined in the python docs, it's appropriate to change things to private if you've identified something that can be.

https://docs.python.org/3.8/tutorial/classes.html#private-variables

daltonbohning: Agree. Done

src/tests/ftest/launch.py
```diff
@@ -1733,6 +1748,115 @@ def setup_test_files(self, args, yaml_dir):
...
+
+        Returns:
+            str: the group's gid
+
+        """
+        # Query the group and get the gid
+        logger.info('Querying group %s', group)
+        result = getent(logger, hosts, 'group', group)
+        if result.passed:
+            return result.output[0].stdout[0].split(':')[2]
```
phender, line 1771: I'd be careful with assuming the expected output is always on the first line, e.g., `stdout[0]`.  A regex here may be safer.  Also note that `result.passed` does not mean that all the hosts returned the same output, that's what `result.homogeneous` indicates.

phender (changes requested):

src/tests/ftest/harness/launch_setup.py
```diff
@@ -8,23 +8,23 @@
...
-    def test_harness_pre_launch_users(self):
-        """Verify all expected users are setup correctly by launch.py.
+    def test_harness_launch_setup_users(self):
+        """Verify all expected users are setup correctly by launch.py -u.
 
         :avocado: tags=all
         :avocado: tags=vm
         :avocado: tags=harness
-        :avocado: tags=test_harness_pre_launch_users
+        :avocado: tags=test_harness_launch_setup_users
```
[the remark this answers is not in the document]
daltonbohning: Done

src/tests/ftest/launch.py
```diff
@@ -1733,6 +1748,115 @@ def setup_test_files(self, args, yaml_dir):
...
+            create (bool, optional): whether to create the group if non-existent. Default is False
+
+        Raises:
+            LaunchException: if there is an error querying or creating the user
+
+        """
+        logger.info('Querying user %s', user)
+        result = run_remote(logger, hosts, f'id {user}')
+        if result.passed:
+            if re.findall(rf'groups={gid}\(', result.output[0].stdout[0]):
```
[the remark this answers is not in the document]
daltonbohning: I added a helper function that robust enough to even handle extraneous output like from SSH, RSA keys, etc.

src/tests/ftest/launch.py
daltonbohning: Fixed

phender (approved):

src/tests/ftest/harness/launch_setup.py
```diff
@@ -0,0 +1,46 @@
...
+        :avocado: tags=harness
+        :avocado: tags=HarnessLaunchSetupTest,test_harness_launch_setup_users
+        """
+        hostlist_clients = self.get_hosts_from_yaml(
+            "test_clients", "server_partition", "server_reservation", "/run/hosts/*")
+        expected_users = self.params.get('client_users', '/run/*')
+        for user_group in expected_users:
+            user, group = user_group.split(':')
+            self.log.info('Checking if group %s exists', group)
+            gids = get_group_id(self.log, hostlist_clients, group).keys()
```
ashleypittman, line 32: I can't work this out, get_group_id gives a dict of host/gid but this then takes keys() so it will return a tuple of hosts, which you then call list() on below?

daltonbohning: `get_group_id` returns a dict like `{<gid>: <hosts>}`, so `list(keys())` effectively gives all the unique gids. We don't need the hosts here, but the function returns the dictionary format to be more complete/extensible

ashleypittman: Ah, so it's `gid: host` not `host: gid`.  That makes a lot more sense.

daltonbohning: Right. The main reason being that `hosts` is a `NodeSet`, which isn't hashable

ashleypittman (approved):

daltonbohning:
Unit tests were skipped since this PR only touches functional tests.
Only VM tests were ran because the changes are verified by the new harness test:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-10521/29/tests

daltonbohning:
@daos-stack/daos-gatekeeper Please use the following commit message on landing:

```
DAOS-11695 test: Add test users to CI env

- Add a --user_create option to launch.py to trigger setting up user accounts from the test yaml
  - launch.py looks for client_users in the yaml
- Add a harness test to verify this functionality, and verify running commands remotely as the created users
- 
```

PR #10525, opened 2022-10-10
DAOS-11043 common: Blob backed Memory(BMEM) interface for umem

sherintg:
This commit contains a new allocator which is a modified version of pmemobj. The new allocator operates on volatile memory and provides transaction functionality similar to that of pmemobj.

The BMEM interfaces internally use the new allocator functionalties to implement the required backend for hosting metadata on volatile memory backed by blob.

Note: The current commit default to BMEM instead of PMDK. One can override this by setting DAOS_USE_PMDK=1 in the server config file.

Allow-unstable-test: true

* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it
* [ ] Commit messages follows the guidelines outlined

### Gatekeeper:

* [ ] The PR has 2 reviews by people familiar with the code.
* [ ] Any appropriate watchers have had a chance to review the PR.
* [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested
      changes.
* [ ] Githooks were used. If not, there is a sufficient reason to move forward.  Check copyrights
      if githooks were not used.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future
      PRs.
* [ ] All builds have passed.  Check non-required builds to ensure they don't indicate any problem
* [ ] No new NLT or valgrind warnings.  Check the classic view. This step should only matter if
* [ ] Ensure sufficent testing is done.   Check feature pragmas and test tags. Check that tests
* [ ] Pay attention to PRs that may affect compatibility between versions and ensure it has been
      addressed.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If
* [ ] Check the commit message when landing. Check the standard


jolivier23:
> Bug-tracker data: Ticket title is 'Implement new alternative umem backend' Status is 'In Progress' Labels: 'md_on_ssd' Job should run at elevated priority (3) Errors are Unknown component https://daosio.atlassian.net/browse/DAOS-11043

It doesn't like umem. You can either change it to another directory name under src/ or src/client or add umem as a component in VALID_COMPONENTS in ci/jira_query.py

This is not required but if you push again, this will eliminate the noise.

src/common/dav/alloc.h
```diff
@@ -0,0 +1,38 @@
...
+static inline void *
+Realloc(void *ptr, size_t size)
+{
+	void *new_ptr;
+
+	D_REALLOC_NZ(new_ptr, ptr, size);
+	return new_ptr;
+}
+
+#define	Free(ptr)	D_FREE(ptr)
```
jolivier23, line 36: One side effect and you can test with D_LOG_MASK=debug DD_MASK=mem is you may get logging for a variable named "ptr" in Malloc but the name will be whatever is passed to the macro here.  That shouldn't cause any issue but may be a little confusing to the log reader.   In such case, it may be better to wrap in an inline function.   I presume the compiler will make it efficient.

What I had in mind with my comment on the prior patch was to wrap the routines, rather than the underlying malloc/free/realloc simiilar to what we do with D_STRNDUP or D_ASPRINTF...the idea being that we log these routines differently.   I'm honestly fine either way though, just clarifying.

jolivier23 (commented):
You might also check if NLT is doing any fault injection.  That could cause "new" failures in some places.

src/common/dav/alloc.h
sherintg: Even the NLT failures were a result of inconsistent use of D_ALLOC/* macros. There were no real double free getting called. It was just that D_FREE was called on allocations done using posix_mem_align() and strdup(). I have made a new commit where all of the dav routines will now use the DAOS D_ALLOC/FREE/* macros for internal allocation.

src/vos/tests/vts_ts.c
```diff
@@ -474,6 +474,7 @@ lru_array_stress_test(void **state)
 		assert_int_equal(evicted, LRU_ARRAY_SIZE);
 	}
 
+	srand(1); /* The below test suite may fail for random seed value*/
```
jolivier23, line 477: why is this necessary here?   Do we use random elsewhere in the code differently than before?

jolivier23:
I merged latest base branch and tested the failing core tests.  I wasn't able to reproduce the rebuild failure

src/vos/tests/vts_ts.c
sherintg: This particular test  sometimes fails if the random number seed is different. During vos create the new code is calling uuid_create() which inturn is using rand. This inturn is causing this test suite to fail sometimes as the below for loop expects 70:30 proportion of lru_alloc vs lru_evict calls to be made and sometimes it is not exactly met.

jolivier23: Ok, It's probably fine but I tried this several times with and without the envirable and it doesn't crash.

jolivier23 (approved):

PR #10532, opened 2022-10-10
DAOS-11884 cart: Add new API to set a number of remote endpoints 

frostedcmos:
Add new API to set number of remote tags for a secondary provider
Round-robin across all secondary tags when sending RPC from a secondary cluster

Signed-off-by: Alexander Oganezov <alexander.a.oganezov@intel.com>

### DIFF
--- a/src/cart/crt_group.c
+++ b/src/cart/crt_group.c
@@ -1849,6 +1849,14 @@ crt_group_config_path_set(const char *path)
 	return 0;
 }
 
+int
+crt_group_num_remote_tags_set(crt_group_t *group, int *tags, int num_entries)
+{
+	struct crt_grp_priv	*grp_priv = NULL;
+
+	grp_priv = crt_grp_pub2priv(grp);
+}
+
 /**
  * Save attach info to file with the name
  * "<singleton_attach_path>/grpid.attach_info_tmp".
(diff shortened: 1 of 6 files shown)


daltonbohning (commented):
ftest LGTM

NiuYawei (approved):

src/cart/crt_group.c
```diff
@@ -1849,6 +1849,35 @@ crt_group_config_path_set(const char *path)
 	return 0;
 }
 
+int
+crt_nr_secondary_remote_tags_set(int idx, int num_tags)
+{
+	struct crt_prov_gdata *prov_data;
+
+	D_DEBUG(DB_ALL, "secondary_idx=%d num_tags=%d\n", num_tags, idx);
```
kjacque, line 1857: Looks like order on the parameters is reversed

frostedcmos: good catch, thanks

kjacque (approved):

kjacque:
There are a couple of odd test failures. I saw pool/create_capacity.py fail once on another PR against the feature branch. We may have an intermittent bug in the branch. With that in mind, I'm going to go ahead and land.

PR #10535, opened 2022-10-10
DAOS-11796 control: Consistently resolve MS replica addresses

mjmac:
In environments where access_points hostnames can resolve
to multiple IP addresses in a nondeterministic manner, we
can run into problems due to MS peers not recognizing each
other. This patch works around the problem by pinning each
replica to the lowest IP address in the set of addresses
associated with each replica's hostname.

### DIFF
--- a/src/control/server/server.go
+++ b/src/control/server/server.go
@@ -166,7 +166,7 @@ func (srv *server) logDuration(msg string, start time.Time) {
 
 // createServices builds scaffolding for rpc and event services.
 func (srv *server) createServices(ctx context.Context) error {
-	dbReplicas, err := cfgGetReplicas(srv.cfg, net.ResolveTCPAddr)
+	dbReplicas, err := cfgGetReplicas(srv.cfg, net.LookupIP)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tanabarr (approved):

kjacque (approved):

kjacque:
Is there a master version of this PR, too? Or did that already land?

mjmac:
> Is there a master version of this PR, too? Or did that already land?

No, not yet. I was initially thinking that this was just a workaround for the 2.2.x series, and I wanted to get feedback on the approach from Aurora testing. I may go ahead and land it on master while I consider a more intrusive change for the 2.4+ series. I haven't yet decided whether or not it's an issue that the raft stuff only knows about a single IP address for a given hostname. Maybe it's actually fine?

kjacque:
> I haven't yet decided whether or not it's an issue that the raft stuff only knows about a single IP address for a given hostname. Maybe it's actually fine?

When I was looking into similar issues previously, I had gotten the idea into my head that we'd probably want raft to know about all the IPs associated with a hostname so we could match against any of them. But with this solution I don't think that would be necessary.

PR #10540, opened 2022-10-11
DAOS-11260 dfuse: Apply timeouts to dfuse data cache.

ashleypittman:
Use the information we keep on when files are opened to track
how old any data in the kernel page cache might be and only
set the keep_cache flag if it's recent.  Extend the current
container setting from being an on/off to also accepting a time value.

Test-tag: dfuse


docs/user/filesystem.md
```diff
@@ -651,24 +651,26 @@ to be set to 0 or off, except dentry-dir-time which defaults to dentry-time
...
-dfuse-data-cache should be set to "on", "true", "off" or "false" if set, other values will
-log an error, and result in the cache being off.  The O\_DIRECT flag for open
-files will be honored with this option enabled, files which do not set
-O\_DIRECT will be cached.
+dfuse-data-cache can be set to a time value or "on", "true", "off" or "false" if set, other values
+will log an error, and result in the cache being off.  The O\_DIRECT flag for open files will be
+honored with this option enabled, files which do not set O\_DIRECT will be cached.  Data caching
+is controlled by dfuse passing a flag to the kernel on open, if data-cache is enabled then it will
+be allowed for files if that file is already open, and timeout value will be the duration between
+a previous close call which reduced the open count to zero and the next subsequent call to open.
```
daltonbohning, line 666: grammar nit
```suggestion
dfuse-data-cache can be set to a time value or "on", "true", "off" or "false". If set, other values
will log an error and result in the cache being off.  The O\_DIRECT flag for open files will be
honored with this option enabled. Files which do not set O\_DIRECT will be cached.  Data caching
is controlled by dfuse passing a flag to the kernel on open. If data-cache is enabled then it will
be allowed for files if that file is already open, and timeout value will be the duration between
a previous close call which reduced the open count to zero and the next subsequent call to open.
```

daltonbohning (approved):

daltonbohning (approved):

src/client/dfuse/dfuse_core.c
```diff
@@ -653,7 +655,7 @@ dfuse_set_default_cont_cache_values(struct dfuse_cont *dfc)
 	dfc->dfc_dentry_timeout     = 1;
 	dfc->dfc_dentry_dir_timeout = 5;
 	dfc->dfc_ndentry_timeout    = 1;
-	dfc->dfc_data_caching       = true;
+	dfc->dfc_data_timeout       = 60 * 10;
```
mchaarawi, line 658: is there some guidance from somewhere that 10 mins is best for default?

mchaarawi (approved):
i feel i should run that ai benchmark from lrz with this to verify it doesn't break the performance that we fixed for them in master. but i can do that later when i have some time.

src/client/dfuse/dfuse_core.c
ashleypittman: No, that's something we need to work out through benchmarking.

wiliamhuang (approved):

PR #10541, opened 2022-10-11
DAOS-11538 proto: chk start flag for handling danglong pool

Nasf-Fan:
Dangling pool means that related pool only exists in MS database, but does not exist on daos targets. To check dangling pools, all daos targets need to report their known pools (shards) when start the check instance. But sometime, the admin may only want to check some special pools when start the check instance. Under such case, the check engines/leader will only report and handle the pools in the check list. That will skip the check of dangling pools.

With this patch, the dangling pool will be handled under following two cases:

1. DANGLING_POOL flag is specified when start the check instance. then in spite of the pools list is specified or not, dangling pools will always be handled.

2. If not specify DANGLING_POOL flag, but the all pools (without specified pools list) are checked from the scratch, then the dangling pools will also be handled.


Nasf-Fan:
@mjmac @ryon-jensen , would you please to help review the patch? Thanks!

PR #10542, opened 2022-10-11
DAOS-11890 vos: Process removals to merge adjacent minor epoch extents

jolivier23:
Minor epoch writes can cause errors when inserting removal records.

Modify the algorithm when finding records to remove to merge overlapping extents with the same minor epoch to avoid -DER_NO_PERM.

Replace the hack here with an assertion.  Fix the evtree test so it fails when this happens (note, it would have failed if that fix were in place previously).

Required-githooks: true

### Gatekeeper:

* [ ] The PR has 2 reviews by people familiar with the code.
* [ ] Any appropriate watchers have had a chance to review the PR.
* [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested
      changes.
* [ ] Githooks were used. If not, there is a sufficient reason to move forward.  Check copyrights
      if githooks were not used.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future
      PRs.
* [ ] All builds have passed.  Check non-required builds to ensure they don't indicate any problem
* [ ] No new NLT or valgrind warnings.  Check the classic view. This step should only matter if
* [ ] Ensure sufficent testing is done.   Check feature pragmas and test tags. Check that tests
* [ ] Pay attention to PRs that may affect compatibility between versions and ensure it has been
      addressed.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If
* [ ] Check the commit message when landing. Check the standard


src/vos/evtree.c
```diff
@@ -848,12 +850,31 @@ evt_find_visible(struct evt_context *tcx, const struct evt_filter *filter,
...
+		 *  step is required to merge records at the same major epoch so we don't attempt
+		 *  to insert overlapping removal records.  Once this step is complete, we can sort
+		 *  the remaining records and return to the caller.
 		 */
-		d_list_for_each_entry(le, &covered, le_link) {
+		while ((le = d_list_pop_entry(&covered, struct evt_list_entry, le_link)) != NULL) {
 			this_ent = &le->le_ent;
+			d_list_for_each_entry_safe(next_le, tmp_le, &covered, le_link) {
+				next_ent = &next_le->le_ent;
+				if (next_ent->en_ext.ex_lo > this_ent->en_ext.ex_hi)
```
NiuYawei, line 862: It should be "next lo" > "this hi" + 1?

src/vos/evtree.c
```diff
@@ -848,12 +850,31 @@ evt_find_visible(struct evt_context *tcx, const struct evt_filter *filter,
...
+					break;
+				if (next_ent->en_epoch != this_ent->en_epoch)
+					continue;
+				if (next_ent->en_sel_ext.ex_lo <=
+				    (this_ent->en_sel_ext.ex_hi + 1)) {
```
NiuYawei, line 867: This check could be removed since we've done adjacent/overlapping check above.

src/vos/evtree.c
```diff
@@ -2568,14 +2589,11 @@ evt_ent_array_fill(struct evt_context *tcx, enum evt_find_opc find_opc,
...
-						DF_RECT" and "DF_RECT"\n", DP_RECT(rect),
-						DP_RECT(&rtmp));
-					rc = 0;
-					goto out;
-				}
+				D_ASSERTF(rect->rc_minor_epc != EVT_MINOR_EPC_MAX,
+					  "Should never have overlap with removals: " DF_RECT
+					  " overlaps with " DF_RECT "\n",
+					  DP_RECT(&rtmp), DP_RECT(rect));
+
```
NiuYawei, line 2596: We regard that caller inserts overlapped removal records as bug, and this assert is to capture the bug?

src/vos/evtree.c
jolivier23: this.hi + 1 == next.lo means the extents are adjacent and can be merged.  So I don't want to skip that case.

src/vos/evtree.c
jolivier23: Good point.  Will change tomorrow

src/vos/evtree.c
jolivier23: Yes.  The only way to insert removal records is through evt_remove_all and this should never insert overlapped records.  The code here was added by Xuezhao to work around the bug I'm fixing with this patch.

src/vos/evtree.c
NiuYawei: but you break the loop and skip that case here, right?

jolivier23: You are right, it is fixed now

NiuYawei (dismissed):

src/vos/evtree.c
jolivier23: Ok, actually, it was not quite right.   I think it's more clear now with a bit of rearrangement but that check was checking the physical record whereas the subsequent check was checking the selected record.   Changing it as I did in prior patch you approved triggered the assertion because it wasn't checking the selected record anymore.

NiuYawei (approved):

ryon-jensen (approved):

PR #10544, opened 2022-10-12
DAOS-11893 control: Include num ctxs in GetAttachInfo

kjacque:
- Save number of secondary provider contexts in Management Service.
- Include the number of contexts for each provider with the rank URIs in GetAttachInfo.
- Update Raft unit test fixtures for the MS updates.

### Gatekeeper:

* [x] You are the appropriate gatekeeper to be landing the patch.
* [x] The PR has 2 reviews by people familiar with the code.
* [x] Any appropriate watchers have had a chance to review the PR.
* [x] Review comments are sufficiently resolved, particularly by prior reviewers that requested
      changes.
* [x] Githooks were used. If not, there is a sufficient reason to move forward.  Check copyrights
      if githooks were not used.
* [x] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future
      PRs.
* [x] All builds have passed.  Check non-required builds to ensure they don't indicate any problem
* [x] No new NLT or valgrind warnings.  Check the classic view. This step should only matter if
* [x] Ensure sufficent testing is done.   Check feature pragmas and test tags. Check that tests
* [x] Quick-build or Quick-functional is not used.
* [x] Pay attention to PRs that may affect compatibility between versions and ensure it has been
      addressed.
* [x] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If
* [x] Check the commit message when landing. Check the standard


src/control/system/raft/database.go
```diff
@@ -559,9 +561,11 @@ func (db *Database) GroupMap() (*GroupMap, error) {
 		}
 
 		gm.RankEntries[srv.Rank] = RankEntry{
-			PrimaryURI:    srv.PrimaryFabricURI,
-			SecondaryURIs: srv.SecondaryFabricURIs,
-			Incarnation:   srv.Incarnation,
+			PrimaryURI:       srv.PrimaryFabricURI,
+			NumPrimaryCtxs:   srv.FabricContexts,
```
knard-intel, line 565: Nit, but to be consistent should be prefixed with Primary ?

src/control/system/member.go
```diff
@@ -161,17 +161,18 @@ func (ms MemberState) isTransitionIllegal(to MemberState) bool {
...
-	Info                string        `json:"info"`
-	FaultDomain         *FaultDomain  `json:"fault_domain"`
-	LastUpdate          time.Time     `json:"last_update"`
+	Rank                    ranklist.Rank `json:"rank"`
+	Incarnation             uint64        `json:"incarnation"`
+	UUID                    uuid.UUID     `json:"uuid"`
+	Addr                    *net.TCPAddr  `json:"addr"`
+	PrimaryFabricURI        string        `json:"fabric_uri"`
+	SecondaryFabricURIs     []string      `json:"secondary_fabric_uris"`
+	FabricContexts          uint32        `json:"fabric_contexts"`
```
knard-intel, line 170: Nit, but to be consistent should be prefixed with Primary ?

knard-intel (approved):
LGTM, just some details with field structure name consistency.

src/control/common/proto/mgmt/svc.pb.go
```diff
@@ -1132,6 +1133,13 @@ func (x *GetAttachInfoResp_RankUri) GetProviderIdx() uint32 {
 	return 0
 }
 
+func (x *GetAttachInfoResp_RankUri) GetNumCtxs() uint32 {
+	if x != nil {
+		return x.NumCtxs
+	}
+	return 0
```
frostedcmos, line 1140: i wonder if 1 is a better default. all servers can be assumed to have at least 1 context

kjacque: This is autogenerated protobuf code so we can't set defaults here unfortunately. A good point though, we should set defaults when we read the config.

src/control/system/raft/database.go
kjacque: Good catch, you're right, I will update the name.

src/control/common/proto/mgmt/svc.pb.go
kjacque: Verified in the code that we're already checking for this default case if the field is unset in the yaml. If a user explicitly specifies 0 endpoints, that is an invalid config.

knard-intel (approved):
LGTM

kjacque:
The one failed test looked like a bunch of network timeout errors on the fabric during teardown. My instinct is that it has nothing to do with these changes.

kjacque:
Functional HW passed on the re-run. The only failure is a codespell failure outside of this changeset (which is fixed in another PR on this branch).

mjmac (approved):
Looks OK to me.

Out of curiosity, and unrelated to this patch specifically, have you tested migrating from a non-multiprovider system? Do we need to do any migrations when the new server starts up with an older sysdb?

kjacque:
> Out of curiosity, and unrelated to this patch specifically, have you tested migrating from a non-multiprovider system? Do we need to do any migrations when the new server starts up with an older sysdb?

I think it should be a non-issue. I haven't tested this case yet, but I've tried to keep the multiprovider additions strictly additions, so as to avoid disturbing the behavior in the single provider case, and maintain compatibility. I think it would be a good idea to explicitly check this as part of the test plan, though.

PR #10547, opened 2022-10-12
DAOS-7821 test: Resolve all python warnings in NLT test code.

ashleypittman:
Update to remove both flake and pylint warnings across entire file.


utils/node_local_test.py
```diff
@@ -56,36 +56,39 @@ class NLTestTimeout(NLTestFail):
     """Used to indicate that an operation timed out"""
 
 
-instance_num = 0
+instance_num = 0  # pylint: disable=invalid-name
```
daltonbohning, line 59: Why not just rename?
```suggestion
INSTANCE_NUM= 0
```

utils/node_local_test.py
```diff
@@ -56,36 +56,39 @@ class NLTestTimeout(NLTestFail):
...
+        with open(json_file, 'r') as ofh:
+            self._bc = json.load(ofh)
         self.agent_dir = None
         self.wf = None
         self.args = None
         self.max_log_size = None
         self.valgrind_errors = False
-        self.lt = CulmTimer()
+        self.log_timer = CulmTimer()
         self.lt_compress = CulmTimer()
```
daltonbohning, line 92: If we rename `lt`, should we also rename `lt_compress` to be consistent?

utils/node_local_test.py
```diff
@@ -2805,7 +2800,7 @@ def test_cont_copy(self):
                '--src',
                src_dir.name,
                '--dst',
-               'daos://{}/{}'.format(self.pool.uuid, self.container)]
+               f'daos://{self.pool.uuid}/{self.container}']
```
daltonbohning, line 2803: Side note - If you ever get the chance, we should switch from pool UUID to label. Mohamad suggested before that the datamover tools support labels only to be more user friendly and simplify logic

utils/node_local_test.py
ashleypittman: Because it's not a constant.  I was cheating a bit with suppressions this one but as it's only used in two places and they're right next to each other in the file I thought silencing was the easiest way.

utils/node_local_test.py
ashleypittman: Done.

utils/node_local_test.py
ashleypittman: Good point.  There's another pylint issue in this file that the Pool class has a id() method whose name is invalid however it's also widely used so changing it would expand the number of changes here.

Best would be to change the Class to only support labels and then self.pool could be used throughout, then the question is would it be better to make code changes and update all the invocation sites in this PR or separately?  I can do either.

soumagne (approved):

ashleypittman:
Only test failure is also happening on master.

PR #10556, opened 2022-10-13
DAOS-10666 dfs: Get only needed props during mount

kjacque:
Instead of requesting all props, including ACL, dfs_mount should request only those that it needs. All of the requested props are available to users with get_prop access on the container.

### Gatekeeper:

### DIFF
--- a/src/client/dfs/dfs.c
+++ b/src/client/dfs/dfs.c
@@ -1950,13 +1950,19 @@ out:
 int
 dfs_mount(daos_handle_t poh, daos_handle_t coh, int flags, dfs_t **_dfs)
 {
-	dfs_t			*dfs;
-	daos_prop_t		*prop;
-	struct daos_prop_entry	*entry;
-	struct daos_prop_co_roots *roots;
-	struct dfs_entry	root_dir;
-	int			amode;
-	int			rc;
+	dfs_t				*dfs;
+	daos_prop_t			*prop;
+	struct daos_prop_entry		*entry;
+	struct daos_prop_co_roots	*roots;
+	struct dfs_entry		root_dir;
+	int				amode;
+	int				rc;
+	int				i;
+	int				num_props = 4;
+	uint32_t			props[] = {DAOS_PROP_CO_LAYOUT_TYPE,
+						   DAOS_PROP_CO_ROOTS,
+						   DAOS_PROP_CO_OWNER,
+						   DAOS_PROP_CO_OWNER_GROUP};
 
 	if (_dfs == NULL)
 		return EINVAL;
@@ -1965,9 +1971,12 @@ dfs_mount(daos_handle_t poh, daos_handle_t coh, int flags, dfs_t **_dfs)
 	if (get_daos_obj_mode(flags) == -1)
 		return EINVAL;
 
-	prop = daos_prop_alloc(0);
+	prop = daos_prop_alloc(num_props);
 	if (prop == NULL)
 		return ENOMEM;
+
+	for (i = 0; i < num_props; i++)
+		prop->dpp_entries[i].dpe_type = props[i];
 
 	rc = daos_cont_query(coh, NULL, prop, NULL);
 	if (rc) {


ashleypittman:
This looks like it would work, it would be good to get a test for it if we can.  I'll try and work on one tomorrow so it would be good to include in this PR and include the fix and the test in the same commit.

kjacque:
> This looks like it would work, it would be good to get a test for it if we can. I'll try and work on one tomorrow so it would be good to include in this PR and include the fix and the test in the same commit.

Thanks! I only verified manually. A new test would be great (maybe part of NLT?).

jolivier23 (dismissed):

src/client/dfs/dfs.c
```diff
@@ -1950,13 +1950,19 @@ dfs_disconnect(dfs_t *dfs)
...
+	struct daos_prop_co_roots	*roots;
+	struct dfs_entry		root_dir;
+	int				amode;
+	int				rc;
+	int				i;
+	int				num_props = 4;
+	uint32_t			props[] = {DAOS_PROP_CO_LAYOUT_TYPE,
+						   DAOS_PROP_CO_ROOTS,
+						   DAOS_PROP_CO_OWNER,
+						   DAOS_PROP_CO_OWNER_GROUP};
```
ashleypittman, line 1965: I think it's probably better to only get two of these, then perform a second fetch if needed on L2025.

kjacque: Fixed.

src/client/dfs/dfs.c
```diff
@@ -2147,16 +2147,69 @@ dfs_disconnect(dfs_t *dfs)
...
-	int			rc;
+	dfs_t				*dfs;
+	daos_prop_t			*prop;
+	struct daos_prop_entry		*entry;
+	struct daos_prop_co_roots	*roots;
+	struct dfs_entry		root_dir;
+	int				amode;
+	int				rc;
+	int				i;
+	#define				NUM_PROPS 3
```
mchaarawi, line 2209: this sounds weird to have a #define in here. maybe just use 3 without a define ..
or better just add an int that is 3 or 5 depending on the layout?

src/client/dfs/dfs.c
```diff
@@ -2147,16 +2147,69 @@ dfs_disconnect(dfs_t *dfs)
...
+		D_GOTO(out, rc = daos_der2errno(rc));
+	}
+
+	/* Convert the owner information to uid/gid */
+	entry = daos_prop_entry_get(owner_props, DAOS_PROP_CO_OWNER);
+	D_ASSERT(entry != NULL);
+	rc = daos_acl_principal_to_uid(entry->dpe_str, &dfs->uid);
+	if (rc == -DER_NONEXIST)
+		/** Set uid to nobody */
+		rc = daos_acl_principal_to_uid("nobody@", &dfs->uid);
```
ashleypittman, line 2176: Do you need to convert this from der2errno?  Same on L2184

src/client/dfs/dfs.c
```diff
@@ -2147,16 +2147,69 @@ dfs_disconnect(dfs_t *dfs)
...
+	int			rc = 0;
+
+	owner_props = daos_prop_alloc(2);
+	if (owner_props == NULL)
+		return ENOMEM;
+
+	owner_props->dpp_entries[0].dpe_type = DAOS_PROP_CO_OWNER;
+	owner_props->dpp_entries[1].dpe_type = DAOS_PROP_CO_OWNER_GROUP;
+
+	rc = daos_cont_query(coh, NULL, owner_props, NULL);
```
mchaarawi, line 2164: instead of calling daos_cont_query twice, why not call it once and depending on the dfs layout, call it with three props or 5?

mchaarawi (changes requested):

src/client/dfs/dfs.c
ashleypittman: I wondered this, you don't know if you need the extra two until you've used the first set do you?

mchaarawi: ah yes indeed, since we have to get the SB oid to be able to query the layout from there. i missed that part.
so querying non existing props returns errors?

ashleypittman:
> This looks like it would work, it would be good to get a test for it if we can. I'll try and work on one tomorrow so it would be good to include in this PR and include the fix and the test in the same commit.

I had a quick go at this and without multiple users to assign a container to it's not all that easy to setup standalone.  The multi-user code will cover this functionality though (I've cherry-picked this PR into that one for now).

src/client/dfs/dfs.c
mchaarawi: i understand that it's better to query only the props that are needed and we should update this. that definitely is better. 
bit i don't understand though why querying all properties returns an error depending on the ACL. Unless we have ACLs that restrict querying some properties but not others, shouldn't querying 3 properties return the same error as querying all properties if there are ACL permission issues?

mchaarawi: > Calling daos_cont_query() twice was my idea to streamline the normal case although it wouldn't be wrong to query all five up-front.

on that front it's cleaner to me to just query all 5 to reduce query calls, prop alloc, prop frees.

ashleypittman: We have ACLs that restrict querying some properties but not others, but none of the affected properties are used by dfs.

mchaarawi: > > Calling daos_cont_query() twice was my idea to streamline the normal case although it wouldn't be wrong to query all five up-front.
> 
> on that front it's cleaner to me to just query all 5 to reduce query calls, prop alloc, prop frees.

but i'll retract my request for change there as indeed querying 3 is the normal case, and we shouldn't call query twice unless we are reading 2.0 container which i don't anticipate would happen much

src/client/dfs/dfs.c
kjacque: I was trying to have something that would flag at compile time if the array was the wrong length. I agree it looks a bit weird to have a #define here though.

mchaarawi: i think there is no need to have a variable. it's just use right below in 2 places, so just using 3 is fine IMO

src/client/dfs/dfs.c
kjacque: Will fix.

mchaarawi (approved):

ashleypittman (dismissed):

src/client/dfs/dfs.c
```diff
@@ -2165,10 +2218,13 @@ dfs_mount(daos_handle_t poh, daos_handle_t coh, int flags, dfs_t **_dfs)
 	if (get_daos_obj_mode(flags) == -1)
 		return EINVAL;
 
-	prop = daos_prop_alloc(0);
+	prop = daos_prop_alloc(num_props);
```
ashleypittman, line 2221: You could probably do something like the following which would be nicer as it's automatically the correct size.
```suggestion
	prop = daos_prop_alloc(ARRAY_SIZE(props));
```

kjacque: Good idea. I think this will be a bit more foolproof in the long run.

ashleypittman (approved):
Much better.

mchaarawi (approved):

kjacque:
The failing tests are unrelated IOR issues, and are being seen in other PRs. The ticket where these issues are being resolved: https://daosio.atlassian.net/browse/DAOS-11936

PR #10558, opened 2022-10-13
DAOS-11146 object: cache dc_pool and dc_cont

wangshilong:
Whenever we want to access pool and container pointer in daos client stack,
looking a per-proccess handle hash table by daos handle is required.
Performance could not scale well for multiple threads because of pthread read lock scability.

To avoid above issue, It is possible to store container and pool pointers in dc_object directly,
extra reference will be held for pool/cont if object is opened. during IO, accessing these
pointers are safe, even pool/cont are closed, server will return errors.

Benchmarking showed fio 4k random write(32 threads) improved from 381k iops to
1500k iops on Rocky8.

Required-githooks: true

### DIFF
--- a/src/object/obj_tx.c
+++ b/src/object/obj_tx.c
@@ -19,6 +19,7 @@
 #include <daos/mgmt.h>
 #include <daos/dtx.h>
 #include <daos/task.h>
+#include <daos/container.h>
 #include "obj_ec.h"
 #include "obj_rpc.h"
 #include "obj_internal.h"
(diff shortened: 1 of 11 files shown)


src/object/obj_tx.c
```diff
@@ -355,6 +365,18 @@ dc_tx_alloc(daos_handle_t coh, daos_epoch_t epoch, uint64_t flags,
 	*ptx = tx;
 
 	return 0;
+out:
+	dc_cont_put(dc);
+
+	if (tx) {
+		if (tx->tx_pool)
+			dc_pool_put(tx->tx_pool);
+		if (tx->tx_req_cache)
```
mchaarawi, line 374: no need to check if !NULL for calling D_FREE

mchaarawi (approved):

src/object/obj_tx.c
wangshilong: Fix

mchaarawi (approved):

wiliamhuang (approved):

PR #10566, opened 2022-10-13
DAOS-11396 control: Enable hugepage support for tmpfs

mjmac:
Unless scm_hugepages_disabled is true in the engine storage configuration,
the per-engine tmpfs will be mounted with the huge=always option.

Required-githooks: true

### DIFF
--- a/src/control/server/storage/scm/provider.go
+++ b/src/control/server/storage/scm/provider.go
@@ -611,6 +611,9 @@ func (p *Provider) MountRamdisk(target string, params *storage.RamdiskParams) (*
 	if params.Size > 0 {
 		opts = append(opts, fmt.Sprintf("size=%dg", params.Size))
 	}
+	if !params.DisableHugepages {
+		opts = append(opts, "huge=always")
+	}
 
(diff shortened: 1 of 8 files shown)


johannlombardi (approved):

src/control/server/storage/scm/provider.go
```diff
@@ -611,6 +611,9 @@ func (p *Provider) MountRamdisk(target string, params *storage.RamdiskParams) (*
 	if params.Size > 0 {
 		opts = append(opts, fmt.Sprintf("size=%dg", params.Size))
 	}
+	if !params.DisableHugepages {
+		opts = append(opts, "huge=always")
```
johannlombardi, line 615: Shall we use "mpol=local" too?

mjmac: In a prior commit, (line 609 above this), I added support for `mpol=prefer:%d` where we prefer to allocate memory on the numa node being used by the engine. I don't feel strongly either way, but based on https://man7.org/linux/man-pages/man2/set_mempolicy.2.html it appears that local is the default policy anyhow. I didn't make that configurable, but it wouldn't be difficult to expose that as a tunable so that the effect of different memory policies can be tested.

PR #10577, opened 2022-10-14
DAOS-11915 vos: Reregistration of btree doesn't fail

ryon-jensen:
It was discovered by the ddb tool that when vos_self_init/fini are called repeatedly, the container btree class is never really 'unregistered' so registering again causes it to fail with -DER_EXIST. This changes that behavior so that if registering and the class already exists, it will check if it's registering the same thing, then returns success. The registration process does not allocate memory or resources so there is no harm in multiple registrations.

Signed-off-by: Ryon Jensen <ryon.jensen@intel.com>

### Gatekeeper:

### DIFF
--- a/src/common/btree.c
+++ b/src/common/btree.c
@@ -4183,8 +4183,12 @@ dbtree_class_register(unsigned int tree_class, uint64_t tree_feats,
 		return -DER_INVAL;
 
 	/* XXX should be multi-thread safe */
-	if (btr_class_registered[tree_class].tc_ops != NULL)
-		return -DER_EXIST;
+	if (btr_class_registered[tree_class].tc_ops != NULL) {
+		if (btr_class_registered[tree_class].tc_ops != ops ||
+		    btr_class_registered[tree_class].tc_feats != tree_feats)
+			return -DER_EXIST;
+		return 0;
+	}
 
 	/* These are mandatory functions */
 	D_ASSERT(ops != NULL);
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -1148,7 +1148,7 @@ events_handler(void *arg)
 	D_DEBUG(DB_MD, DF_UUID": starting\n", DP_UUID(svc->ps_uuid));
 
 	for (;;) {
-		struct pool_svc_event  *event;
+		struct pool_svc_event  *event = NULL;
 		bool			stop;
 
 		ABT_mutex_lock(events->pse_mutex);


src/common/btree.c
```diff
@@ -4183,8 +4183,12 @@ dbtree_class_register(unsigned int tree_class, uint64_t tree_feats,
 		return -DER_INVAL;
 
 	/* XXX should be multi-thread safe */
-	if (btr_class_registered[tree_class].tc_ops != NULL)
-		return -DER_EXIST;
+	if (btr_class_registered[tree_class].tc_ops != NULL) {
+		if (btr_class_registered[tree_class].tc_ops != ops ||
+		    btr_class_registered[tree_class].tc_feats != tree_feats)
+			return -DER_EXIST;
+		return 0;
```
NiuYawei, line 4190: This change looks good to me, but I don't see why repeated vos_self_init() could fail with -DER_EXIST, the -DER_EXIST errors should have been ignored by most dbtree_class_register() callers, including the vos_self_init().

src/pool/srv_pool.c
```diff
@@ -1148,7 +1148,7 @@ events_handler(void *arg)
 	D_DEBUG(DB_MD, DF_UUID": starting\n", DP_UUID(svc->ps_uuid));
 
 	for (;;) {
-		struct pool_svc_event  *event;
+		struct pool_svc_event  *event = NULL;
```
kccain, line 1151: seems like a reasonable change. Was there a situation in which a stale / already freed event was processed?

src/common/btree.c
ryon-jensen: I didn't see that it was ignored. Would you prefer that I change it so that each call of dbtree_class_register() ignores the DER_EXIST error instead of handling it within dbtree_class_register()?

src/pool/srv_pool.c
ryon-jensen: I have an older compiler that I was using and was giving me `error: ‘event’ may be used uninitialized in this function [-Werror=maybe-uninitialized]`. I don't have to commit this change but thought it was a good one. If the event pointer is never set within the if statement, then the handle_event would receive an uninitialized pointer.

ryon-jensen:
I'm not sure why the Jira report failed yet.

src/common/btree.c
NiuYawei: oh, I was looking into vos_self_init() -> vos_self_nvme_init(), but seems vos_self_init() -> vos_mod_init() does return error in such case.
I'm fine with current approach.

NiuYawei (approved):

src/pool/srv_pool.c
ryon-jensen: I don't think it can actually happen, but if 'stop' == false and d_list_empty(...) is true??

ryon-jensen:
Requested @daos-stack/daos-gatekeeper with forced-landing label because of Jira Report failure which causes a red X :-(

src/pool/srv_pool.c
liw: In that case we will call sched_cond_wait and check the "stop" and "d_list_empty" condition again, that is, we will not break the inner loop.

PR #10583, opened 2022-10-15
DAOS-11674 doc: document default EC cell size change in DAOS 2.2

Michael-Hennecke:
document that ec_cell_sz changed from 1MiB to 64kiB in DAOS 2.2
provide better guidance how to select the optimal EC cell size

Doc-only: true
Signed-off-by: Michael Hennecke <michael.hennecke@intel.com>

### DIFF
--- a/docs/user/container.md
+++ b/docs/user/container.md
@@ -260,7 +260,7 @@ The table below summarizes the available container properties.
 | rf\_lvl                 | Yes             | Redundancy Level which is the level in the fault domain hierarchy to use for object placement|
 | health                  | No              | Current state of the container|
 | alloc\_oid              | No              | Maximum allocated object ID by container allocator|
-| ec\_cell                | Yes             | Erasure code cell size for erasure-coded objects|
+| ec\_cell\_sz            | Yes             | Erasure code cell size for erasure-coded objects|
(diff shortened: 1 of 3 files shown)


docs/user/container.md
```diff
@@ -455,23 +455,50 @@ Server Checksumming   on
...
-DAOS erasure code implementation uses a fixed cell size that applies to all
-objects in the container. The cell size in DAOS is the size of each data and
-parity fragments (also called sometimes chunks). The cell size can be set at
-container creation time via the property:
+The DAOS erasure code implementation uses a fixed cell size that applies to all
+objects in the container.
+The cell size in DAOS is the size of a single data and parity fragment.
+By default, a container's `ec_cell_sz` property is inherited from the pool's
+default `ec_cell_sz`, which was 1MiB in DAOS 2.0 and has been reduced to
+64kiB in DAOS 2.2.  The container cell size can also be set at
```
kccain, line 463: Very minor I think (unless I am not understanding something).
There are uses of kiB and KiB in the documentation (some of which is output from commands - not sure if there is an inconsistency in command output text)

For the parts of the patch that are written documentation (not pasted command output), does it make sense to use the same case e.g., KiB everwhere versus kIB some places and KiB in other places?

kccain (approved):

johannlombardi (approved):

PR #10589, opened 2022-10-17
DAOS-4713 bio: introduce retry-able DER_NVME_IO

NiuYawei:
Introduce retry-able error DER_NVME_IO, once a fetch request hits NVMe I/O error when reading from SSD, this -DER_NVME_IO will be returned to client, then the client will use the same mechanism for -DER_CSUM error to retry fetch from replicas (if any).

This patch also includes code changes for auto SSD faulty detection, the auto faulty feature is disabled by default now, once it's enabled, -DER_NVME_IO will be returned for NVMe write as well, then client won't fail the I/O but keep retry until the SSD is marked as faulty and the affected target being marked as DOWN.

Required-githooks: true

Signed-off-by: Niu Yawei <yawei.niu@intel.com>

### Gatekeeper:


src/object/cli_obj.c
```diff
@@ -4886,10 +4912,7 @@ obj_ec_valid_shard_get(struct obj_auxi_args *obj_auxi, uint8_t *tgt_bitmap,
 	if (rc) {
 		if (force_ec_degrade) {
 			obj_auxi->no_retry = 1;
-			if (obj_auxi->csum_retry)
-				rc = -DER_CSUM;
-			else if (obj_auxi->tx_uncertain)
-				rc = -DER_TX_UNCERTAIN;
+			rc = retry_errcode(obj_auxi);
```
wangshilong, line 4915: logic changes a bit here, rc here might be overwrote with EIO incorrectly?

rc1 = retry_errorcode(obj_auxi);
if (rc1 != DER_IO)
  rc = rc1

src/object/cli_obj.c
```diff
@@ -5073,7 +5097,7 @@ obj_fetch_shards_get(struct dc_object *obj, daos_obj_fetch_t *args, unsigned int
 		rc = obj_ec_fetch_shards_get(obj, args, map_ver, obj_auxi, shard, shard_cnt);
 		if (rc)
 			D_GOTO(out, rc);
-	} else if ((obj_auxi->csum_retry || obj_auxi->tx_uncertain)) {
+	} else if ((obj_auxi->csum_retry || obj_auxi->tx_uncertain || obj_auxi->nvme_io_err)) {
```
wangshilong, line 5100: maybe a helper() obj_auxi_retryable()

src/bio/bio_xstream.c
```diff
@@ -159,6 +159,25 @@ bypass_health_collect()
...
 }
 
+struct bio_faulty_criteria	glb_criteria;
+
+/* TODO: Make it configurable through control plane */
+static inline void
+set_faulty_criteria(void)
+{
+	glb_criteria.fc_enabled = false;
+	glb_criteria.fc_max_io_errs = 5;
```
wangshilong, line 169: don't have much knowledge about SSD, but 5 looks too aggresive?

NiuYawei: I'm not sure too, let's adjust the number when we enable the auto faulty feature.

src/object/cli_obj.c
NiuYawei: Yes, I'll fix it. thanks

src/object/cli_obj.c
NiuYawei: sure

wangshilong (dismissed):

src/bio/bio_buffer.c
```diff
@@ -1046,6 +1052,13 @@ nvme_rw(struct bio_desc *biod, struct bio_rsrvd_region *rg)
 	if (daos_io_bypass & IOBP_NVME)
 		return;
 
+	/* No locking for BS state query here is tolerable */
+	if (xs_ctxt->bxc_blobstore->bb_state == BIO_BS_STATE_FAULTY) {
+		D_ERROR("Blobstore is marked as FAULTY.\n");
+		biod->bd_result = -DER_NVME_IO;
```
Nasf-Fan, line 1058: Check glb_criteria.fc_enabled for update case before set NVME_IO?

src/bio/bio_xstream.c
```diff
@@ -159,6 +159,25 @@ bypass_health_collect()
...
+	glb_criteria.fc_max_csum_errs = 5;
+
+	d_getenv_bool("DAOS_NVME_AUTO_FAULTY_ENABLED", &glb_criteria.fc_enabled);
+	d_getenv_int("DAOS_NVME_AUTO_FAULTY_IO", &glb_criteria.fc_max_io_errs);
```
Nasf-Fan, line 173: Sorry, not understand why need fc_max_io_errs. Based on what consideration when we set it?

src/object/cli_obj.c
```diff
@@ -4839,15 +4869,18 @@ obj_retry_next_shard(struct dc_object *obj, struct obj_auxi_args *obj_auxi,
...
-
-		return -DER_TX_UNCERTAIN;
+		return retry_errcode(obj_auxi, 0);
 	}
 
 	return rc;
 }
 
+static inline bool
+obj_auxi_retryable(struct obj_auxi_args *obj_auxi)
```
Nasf-Fan, line 4879: "retryable" is some misguided in client IO logic, because client will retry the IO request under many cases, for example: RPC timeout, DER_INPROGRESS, stale pool map, and the cases you mention here.

In your case, the retry is for next replica, so we need a more proper name, such as "obj_auxi_need_retry_redudancy"?

src/bio/bio_buffer.c
NiuYawei: Ok.

src/object/cli_obj.c
NiuYawei: Sure.

src/bio/bio_xstream.c
NiuYawei: discussed this with Fanyong offline.

wangshilong (approved):

Nasf-Fan (approved):

src/bio/bio_xstream.c
```diff
@@ -159,6 +159,25 @@ bypass_health_collect()
...
 
+struct bio_faulty_criteria	glb_criteria;
+
+/* TODO: Make it configurable through control plane */
+static inline void
+set_faulty_criteria(void)
+{
+	glb_criteria.fc_enabled = false;
+	glb_criteria.fc_max_io_errs = 5;
+	glb_criteria.fc_max_csum_errs = 5;
```
ryon-jensen, line 170: There is a user settable pool property (DAOS_PROP_PO_SCRUB_THRESH) that the checksum scrubber as a threshold. Wondering if we should use the same property here instead of hard coding it? Could be in a future patch.

ryon-jensen (approved):

src/bio/bio_xstream.c
NiuYawei: Yes, all these hardcoding numbers are not for the production feature. We have to make the faulty criteria configurable through control plane (probably not per-pool but a global configuration?) when we officially support auto device faulty.

src/object/obj_internal.h
```diff
@@ -592,7 +593,8 @@ obj_retry_error(int err)
 	       err == -DER_EXCLUDED || err == -DER_CSUM ||
 	       err == -DER_TX_BUSY || err == -DER_TX_UNCERTAIN ||
 	       err == -DER_NEED_TX || err == -DER_NOTLEADER ||
-	       err == -DER_UPDATE_AGAIN || daos_crt_network_error(err);
+	       err == -DER_UPDATE_AGAIN || err == -DER_NVME_IO ||
```
jolivier23, line 596: I guess this is rare but it makes my eyes hurt.  Seems like we could either change these error codes (so they are all in the same range) or we could define a function to convert the error code into an index to a static array that gives us true or false.  This would take far fewer checks with our FOREACH macro setup.  Anyway, not a comment on this patch, just something I don't like seeing.

So it would be something like
idx = d_err2idx(errno);
return error_is_retryable[idx];

We already do something similar for d_errstr conversions.

NiuYawei: agreed, let's fix it in a separate patch.

PR #10642, opened 2022-10-21
DAOS-11554 control: Don't log missing lib as error (#10327)

kjacque:
The use case where libfabric is present but UCX libs aren't installed is normal.

- Log missing dynamically loaded libs at debug level only.
- Improve error messages so it's easier to tell which library failed to load.


kjacque:
The codespell failure is not related to this PR.

PR #10654, opened 2022-10-24
DAOS-623 build: Use AddMethod for adding methods to environments.

ashleypittman:


site_scons/daos_build.py
```diff
@@ -235,8 +227,16 @@ def configure_mpi(env):
...
     print("No MPI installed")
     return False
+
+
+def setup(env):
+    """Add daos specific methods to environment"""
+    env.AddMethod(_add_build_rpath, 'd_add_build_rpath')
+    env.AddMethod(_configure_mpi, 'd_configure_mpi')
+    env.AddMethod(_run_command, 'd_run_command')
+    env.AddMethod(_add_rpaths, 'd_add_rpaths')
```
ashleypittman, line 242: This is the real change in this PR, the rest is just fluff.

1. Is this worthwhile?
2. do we want a common 'd_' prefix for all our methods to make it clear they're non-standard.
3. Should we copy the CamelCase naming of existing methods?
4. Assuming this is worthwhile should we do it for all daos_build functions (which will touch more code than this PR).

jolivier23 (approved):

site_scons/daos_build.py
jolivier23: For #2, I think we already have other precedent where we add stuff and don't name it specially.  For instance, CheckFlag and CheckFlagCC.   I think it would be fine to do something like AddBuildRPATH

ashleypittman: That would mean that we'd replace daos_build.library() with env.Library(), yet there is already a method with that name, the one that we're wrapping/

jolivier23: You could do something like DAOSLibrary, I suppose

jolivier23 (approved):

SConstruct
```diff
@@ -50,6 +45,8 @@ API_VERSION = f'{API_VERSION_MAJOR}.{API_VERSION_MINOR}.{API_VERSION_FIX}'
 
 def update_rpm_version(version, tag):
     """ Update the version (and release) in the RPM spec file """
+
+    # pylint: disable=consider-using-f-string
```
daltonbohning, line 49: Can we use f-strings in this file?

utils/sl/fake_scons/SCons/Script/__init__.py
```diff
@@ -192,7 +191,28 @@ def Alias(self, *_args, **_kw):
...
+
+            def __index__(self):
+                return 0
+
+            def __getitem__(self, x):
+                """Fake __getitem__"""
+
+            def __setitem__(self, x, value):
+                """Fake __setitem__"""
+                return
```
daltonbohning, line 206: Why are these fake classes needed?

daltonbohning (approved):

SConstruct
ashleypittman: Yes, but if you look at this file there's what should be a function stuck inside a giant `if` block, and it's for tagging releases on GitHub so it's not really tested.  What we should do is move it out of the top-level and put it in it's own code under site_scons somewhere.  Rather than rock the boat on that I just silenced the code-block.

utils/sl/fake_scons/SCons/Script/__init__.py
ashleypittman: In general or these ones specifically?  In general so we can test the scons interface with pylint as scons does some odd things, these ones are so we can remove the pylint disabling suppression from https://github.com/daos-stack/daos/blob/master/SConstruct#L332-L333

It's really just making this object behave like a scons Environment so that the checks pass for code we know works.

daltonbohning: Yes, in general. Thanks for the explanation

PR #10668, opened 2022-10-25
DAOS-11919 control: Add concurrent pool ops protection

mjmac:
In certain circumstances, a PoolCreate and PoolDestroy may
race, leading to unexpected behavior. This commit adds a
new locking mechanism at the pool database level to prevent
concurrent write operations by more than one parent gRPC
handler.

Additionally, we now call raft.Barrier() as part of the
leadership check in order to ensure that the leader has caught
up with any outstanding logs before proceeding on to any logic
that depends on having leadership.

### DIFF
--- a/src/control/server/mgmt_pool.go
+++ b/src/control/server/mgmt_pool.go
@@ -18,6 +18,8 @@ import (
 
 	mgmtpb "github.com/daos-stack/daos/src/control/common/proto/mgmt"
 	"github.com/daos-stack/daos/src/control/drpc"
+	"github.com/daos-stack/daos/src/control/fault"
+	"github.com/daos-stack/daos/src/control/fault/code"
 	"github.com/daos-stack/daos/src/control/lib/daos"
 	"github.com/daos-stack/daos/src/control/lib/ranklist"
 	"github.com/daos-stack/daos/src/control/server/engine"
@@ -42,6 +44,20 @@ type poolServiceReq interface {
 	SetUUID(uuid.UUID)
 	GetSvcRanks() []uint32
 	SetSvcRanks(rl []uint32)
+}
+
+func (svc *mgmtSvc) makeLockedPoolServiceCall(ctx context.Context, method drpc.Method, req poolServiceReq) (*drpc.Response, error) {
+	ps, err := svc.getPoolService(req.GetId())
+	if err != nil {
+		return nil, err
+	}
+	lock, err := svc.sysdb.TakePoolLock(ctx, ps.PoolUUID)
+	if err != nil {
+		return nil, err
+	}
+	defer lock.Release()
+
+	return svc.makePoolServiceCall(lock.InContext(ctx), method, req)
 }
 
 func (svc *mgmtSvc) makePoolServiceCall(ctx context.Context, method drpc.Method, req poolServiceReq) (*drpc.Response, error) {
@@ -88,12 +104,12 @@ func (svc *mgmtSvc) resolvePoolID(id string) (uuid.UUID, error) {
 
 // getPoolService returns the pool service entry for the given UUID.
 func (svc *mgmtSvc) getPoolService(id string) (*system.PoolService, error) {
-	uuid, err := svc.resolvePoolID(id)
+	poolUUID, err := svc.resolvePoolID(id)
 	if err != nil {
 		return nil, err
 	}
 
-	ps, err := svc.sysdb.FindPoolServiceByUUID(uuid)
+	ps, err := svc.sysdb.FindPoolServiceByUUID(poolUUID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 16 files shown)


src/control/system/raft/database.go
```diff
@@ -898,14 +907,41 @@ func (db *Database) FindPoolServiceByLabel(label string) (*system.PoolService, e
 	return nil, system.ErrPoolLabelNotFound(label)
 }
 
+func (db *Database) TakePoolLock(ctx context.Context, uuid uuid.UUID) (*PoolLock, error) {
```
kjacque, line 910: Needs documentation comment

src/control/server/mgmt_pool.go
```diff
@@ -459,6 +482,17 @@ func (svc *mgmtSvc) checkPools(ctx context.Context, psList ...*system.PoolServic
 			continue
 		}
 
+		lock, err := svc.sysdb.TakePoolLock(parent, ps.PoolUUID)
+		if err != nil {
+			if fault.IsFaultCode(err, code.SystemPoolLocked) {
+				svc.log.Noticef("skipping cleanup on pool %s: %s", ps.PoolUUID, err)
+				continue
```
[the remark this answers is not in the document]
mjmac: On startup, on leadership transition, on pool creation. Those are the automatic methods. The admin could also issue the destroy command manually.

src/control/system/raft/database.go
```diff
@@ -260,6 +261,8 @@ func NewDatabase(log logging.Logger, cfg *DatabaseConfig) (*Database, error) {
 			SchemaVersion: CurrentSchemaVersion,
 		},
 	}
+	// NB: We may remove this once the locking stuff is solid.
+	db.poolLocks.log = log
```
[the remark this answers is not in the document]
mjmac: Yeah, I've definitely thought about this. I think that there is a benefit to enforcing some discipline in what gets logged, though. What I've seen many times in the past is that debug/trace level logging gets added and then never gets reduced over time, leading to massive log output whenever those levels are turned on, making support more difficult due to size of logs and difficulty in sorting through them. There isn't a hard rule, but I think generally we should be looking for opportunities to remove debug logging when it's clearly not contributing any value.

I'm open to the idea that we should eventually make the control plane logging system more ~complex~ sophisticated (more levels, more subsystems), but I don't think we're there yet.

src/control/system/raft/database.go
mjmac: Thanks, missed this. Will add it if there is any other feedback that requires another push, or in a follow-up. I'd like to get this landed to resolve some intermittent failures that we're still seeing.

src/control/system/raft/database.go
```diff
@@ -898,14 +907,41 @@ func (db *Database) FindPoolServiceByLabel(label string) (*system.PoolService, e
...
+	if err := db.CheckLeader(); err != nil {
+		return nil, err
+	}
+	db.Lock()
+	defer db.Unlock()
+
+	lock, err := getCtxLock(ctx)
+	if err != nil && err != errNoCtxLock {
+		return nil, err
+	} else if lock != nil {
```
tanabarr, line 920: This smells a bit, it feels more comfortable sticking to the premise that if err exists then we shouldn't be using the value returned from the call (lock). Under what circumstances (error value) will we need to be interrogating the returned lock value and is there a more sane way to do this that doesn't require using the value returned in an error condition?

src/control/system/raft/pool_lock.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, errors.New("nil context in AddContextLock()")
+	}
+	if lock == nil {
+		return nil, errors.New("nil lock in AddContextLock()")
+	}
+
+	cl, err := getCtxLock(parent)
+	if err != nil && err != errNoCtxLock {
+		return nil, err
+	} else if cl != nil {
```
tanabarr, line 87: same applies as per previous comment re. using a value returned alongside an error

tanabarr (dismissed):
No blocking concerns, just a preference regarding coding style expressed.

src/control/system/raft/database.go
mjmac: If you look at the implementation of `getCtxLock()`, you'll see that it never returns a non-nil lock and a non-nil error at the same time. I'll admit that the above construction does not make this clear, however. 

The logic is that we check the context for an existing lock. If a lock exists, err will be nil, and we check the lock to see if it's a valid lock for the pool UUID. If no lock exists in the context, then we return the sentinel `errNoCtxLock` error with a nil lock and proceed down to taking a new lock on the pool.

I'll see if I can clean this up a bit to reduce the ambiguity.

tanabarr (approved):

kjacque (approved):

PR #10694, opened 2022-10-27
DAOS-11610 dfuse: Re-use dfuse buffers.

ashleypittman:


src/client/dfuse/ops/read.c
```diff
@@ -58,44 +57,39 @@ dfuse_cb_read(fuse_req_t req, fuse_ino_t ino, size_t len, off_t position, struct
 		mock_read = true;
 	}
 
-	D_ALLOC(buff, len);
-	if (!buff)
-		D_GOTO(err, rc = ENOMEM);
+	if (len > ev->de_iov.iov_buf_len) {
+		D_ERROR("Fuse read buffer not large enough %zx > %zx\n", len,
+			ev->de_iov.iov_buf_len);
+	}
```
daltonbohning, line 63: In practice, when will this happen?

src/tests/ftest/dfuse/caching_check.py
```diff
@@ -62,8 +62,7 @@ def test_caching_check(self):
         max_mib = int(IorMetrics.Max_MiB)
 
         # unmount dfuse and mount again with caching enabled
-        pcmd(self.hostlist_clients,
-             self.dfuse.get_umount_command(), expect_rc=None)
+        pcmd(self.hostlist_clients, self.dfuse.get_umount_command(), expect_rc=None)
```
daltonbohning, line 65: FYI - this will conflict with my #10696 - it's a race now!

src/tests/ftest/dfuse/caching_check.py
```diff
@@ -74,7 +73,10 @@ def test_caching_check(self):
         out = self.run_ior_with_pool(fail_on_warning=False)
         with_caching = IorCommand.get_ior_metrics(out)
         # verify cached read performance is multiple times greater than without caching
+        # Log all the values first, then do the assert so that failures can be checked easily.
         for base_read in base_read_arr:
             actual_change = percent_change(base_read[0][max_mib], with_caching[0][max_mib])
             self.log.info('assert actual_change > min_change: %f > %f', actual_change, read_x)
+        for base_read in base_read_arr:
+            actual_change = percent_change(base_read[0][max_mib], with_caching[0][max_mib])
```
daltonbohning, line 81: Thank you. I should have done this when I last touch this code... :)

src/tests/ftest/dfuse/caching_check.yaml
```diff
@@ -34,7 +34,7 @@ ior:
   transfer_size: '1M'
   block_size: '64M'
   dfs_oclass: "EC_2P1G1"
-  read_x: 5  # 500%
+  read_x: 4.5  # 450%
```
daltonbohning, line 37: This is fine, but FYI we could lower more if needed

daltonbohning (approved):

src/client/dfuse/ops/read.c
ashleypittman: When the kernel changes to support larger buffers.

src/tests/ftest/dfuse/caching_check.yaml
ashleypittman: It passed 10/10 with this so I'm OK with the new value.

src/client/dfuse/ops/read.c
```diff
@@ -58,44 +57,39 @@ dfuse_cb_read(fuse_req_t req, fuse_ino_t ino, size_t len, off_t position, struct
 		mock_read = true;
 	}
 
-	D_ALLOC(buff, len);
-	if (!buff)
-		D_GOTO(err, rc = ENOMEM);
+	if (len > ev->de_iov.iov_buf_len) {
+		D_ERROR("Fuse read buffer not large enough %zx > %zx\n", len,
```
mchaarawi, line 61: why is this an error if no error is returned?

mchaarawi (commented):
- please add some description on what the PR is doing.
- would be better if slab allocator changes are moved to a separate PR

jolivier23 (dismissed):

src/client/dfuse/ops/read.c
ashleypittman: With fuse you can specify max_read_size or pass 0 at init time to mean "as large as the kernel supports", as dfuse passes 0 it has no way of knowing what the maximum request size will be so it hard-codes to 1mb.  If this error happens then there's better kernel support and we need to change the buffer size but the code will still work as expected.  The error is only here to highlight to us if the kernel limit ever does change.

mchaarawi: then INFO or maybe WARN would be more appropriate in this case?

ashleypittman: Updated to WARN and I've added a comment to the source.

I didn't want to use INFO as we want to know and change dfuse if the kernel value ever changes.

mchaarawi (approved):

daltonbohning (approved):

PR #10705, opened 2022-10-28
DAOS-11981 tools: Check rc in label prop handler

mjmac:
Return an error if setting the property string fails.

Required-githooks: true

### DIFF
--- a/src/control/cmd/daos/property.go
+++ b/src/control/cmd/daos/property.go
@@ -103,8 +103,11 @@ var propHdlrs = propHdlrMap{
 			}
 			e.dpe_type = C.DAOS_PROP_CO_LABEL
 			cStr := C.CString(v)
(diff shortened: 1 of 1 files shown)


mchaarawi (approved):

mjmac:
One known failure in the scrubber tests: [DAOS-11950](https://daosio.atlassian.net/browse/DAOS-11950)

ashleypittman (approved):
Yes, this fixes it.  I've verified the fix in https://github.com/daos-stack/daos/pull/10689 although there are still a couple more leaks I've yet to find.

ashleypittman:
Oh, what should the component be in the commit message?

mjmac:
> Oh, what should the component be in the commit message?

I think it should be "tools", because that what this is. The `daos` utility isn't part of the control plane. How do I fix that error?

PR #10713, opened 2022-10-31
DAOS-11993 vos: new flags VOS_OBJ_KILL_DKEY for  vos_obj_del_key

Nasf-Fan:
During traverse some table under the sysdb, the caller holds reference
on related object, it will cause vos_obj_hold() to return -DER_BUSY if
use DAOS_INTENT_KILL when call vos_obj_del_key(), that is triggered by
the iteration callback. In fact, such restriction is only meaningful
when delete the object instead of the dkey under the object. The patch
introduces new flags VOS_OBJ_KILL_DKEY for vos_obj_del_key() to bypass
related check inside vos_obj_hold().

### Gatekeeper:

### DIFF
--- a/src/vos/vos_obj.c
+++ b/src/vos/vos_obj.c
@@ -667,7 +667,7 @@ vos_obj_del_key(daos_handle_t coh, daos_unit_oid_t oid, daos_key_t *dkey,
 	daos_handle_t		 toh;
 	int			 rc;
 
-	rc = vos_obj_hold(occ, cont, oid, &epr, 0, VOS_OBJ_VISIBLE,
+	rc = vos_obj_hold(occ, cont, oid, &epr, 0, VOS_OBJ_VISIBLE | VOS_OBJ_KILL_DKEY,
 			  DAOS_INTENT_KILL, &obj, NULL);
 	if (rc == -DER_NONEXIST)
 		return 0;
--- a/src/vos/vos_obj.h
+++ b/src/vos/vos_obj.h
@@ -60,6 +60,8 @@ enum {
 	VOS_OBJ_CREATE		= (1 << 1),
 	/** Hold for object specific discard */
 	VOS_OBJ_DISCARD		= (1 << 2),
+	/** Hold the object for delete dkey */
+	VOS_OBJ_KILL_DKEY	= (1 << 3),
 };
 
 /**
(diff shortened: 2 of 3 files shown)


src/vos/vos_obj.c
```diff
@@ -668,7 +668,7 @@ vos_obj_del_key(daos_handle_t coh, daos_unit_oid_t oid, daos_key_t *dkey,
 	int			 rc;
 
 	rc = vos_obj_hold(occ, cont, oid, &epr, 0, VOS_OBJ_VISIBLE,
-			  DAOS_INTENT_KILL, &obj, NULL);
+			  DAOS_INTENT_PUNCH, &obj, NULL);
```
jolivier23, line 671: What is the iterator doing here?   I presume KILL was used for a reason, such as the fact that it checks the reference count on the object before removing stuff from the tree.   Depending on the operations the iterator is performing, this type of thing could be unsafe.

As an example that probably doesn't apply to sysdb.  When aggregation comes back from a yield over an evtree, it doesn't re-probe because it knows the evtree isn't going anywhere and it has a cached state.   But an API like vos_obj_del_key would simply remove the whole evtree causing problems for the aggregation iterator.

The reference count and object eviction for KILL is overkill because what you really want is a reference count on the key tree but it would prevent scenarios such as these.   That said, it's probably not an issue for smd but at the very least should be documented in the API documentation.

Nasf-Fan: In DAOS check, we have table "chk" in sysdb, each pool has an akey under such table. Sometimes, we need to travel the table to delete the pools under such table.

jolivier23 (approved):
not a big fan of using PUNCH since that's not what we are doing and it can be misleading but if it fixes the immediate issue, I'm fine with it.

jolivier23 (approved):

Nasf-Fan:
Ping reviewers. Thanks!

src/vos/vos_obj.c
Nasf-Fan: Yes, the chk logic iterates the table, and conditionally delete some keys from the table via iteration callback.

Nasf-Fan:
Ping reviewers, would you please to help review the new patch? Thanks!

jolivier23 (approved):

NiuYawei (approved):

wangdi1 (approved):

PR #10714, opened 2022-10-31
DAOS-11943 objects: online extending fixes

wangdi1:
1. Cleanup extend handler to use pool_svc_update_map(), similar as other pool update handler.

2. Change extend pool map status to NEW->UP->UPIN, to match with reintegrated process. NEW: just added the rank. UP: extending targets. UPIN: extended/healthy targets.

3. A few fixes for punch over online extending, so for extending/reintegrating/excluding shards, it should use DELAY forward flag. And add tests to verify it.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -5419,6 +5419,9 @@ dc_obj_update(tse_task_t *task, struct dtx_epoch *epoch, uint32_t map_ver,
 		D_GOTO(out_task, rc);
 	}
 
+	if (args->flags & DAOS_COND_MASK)
+		obj_auxi->cond_modify = 1;
+
 	rc = obj_shards_2_fwtgts(obj, map_ver, tgt_bitmap, shard, shard_cnt, 1,
 				 OBJ_TGT_FLAG_FW_LEADER_INFO, obj_auxi);
 	if (rc != 0)
@@ -5443,9 +5446,6 @@ dc_obj_update(tse_task_t *task, struct dtx_epoch *epoch, uint32_t map_ver,
 
 	if (DAOS_FAIL_CHECK(DAOS_DTX_COMMIT_SYNC))
 		obj_auxi->flags |= ORF_DTX_SYNC;
-
-	if (args->flags & DAOS_COND_MASK)
-		obj_auxi->cond_modify = 1;
 
 	D_DEBUG(DB_IO, "update "DF_OID" dkey_hash "DF_U64"\n",
 		DP_OID(obj->cob_md.omd_id), obj_auxi->dkey_hash);
@@ -6146,6 +6146,9 @@ dc_obj_punch(tse_task_t *task, struct dc_object *obj, struct dtx_epoch *epoch,
 			D_GOTO(out_task, rc);
 	}
 
+	if (api_args->flags & DAOS_COND_MASK)
+		obj_auxi->cond_modify = 1;
+
 	rc = obj_shards_2_fwtgts(obj, map_ver, NIL_BITMAP, shard, shard_cnt, grp_cnt,
 				 OBJ_TGT_FLAG_FW_LEADER_INFO, obj_auxi);
 	if (rc != 0)
@@ -6153,9 +6156,6 @@ dc_obj_punch(tse_task_t *task, struct dc_object *obj, struct dtx_epoch *epoch,
 
 	if (daos_fail_check(DAOS_FAIL_TX_CONVERT))
 		D_GOTO(out_task, rc = -DER_NEED_TX);
-
-	if (api_args->flags & DAOS_COND_MASK)
-		obj_auxi->cond_modify = 1;
 
 	if (DAOS_FAIL_CHECK(DAOS_DTX_COMMIT_SYNC))
 		obj_auxi->flags |= ORF_DTX_SYNC;
--- a/src/placement/jump_map.c
+++ b/src/placement/jump_map.c
@@ -93,10 +93,8 @@ layout_find_diff(struct pl_jump_map *jmap, struct pl_obj_layout *original,
 		if (reint_tgt != original_target) {
 			pool_map_find_target(jmap->jmp_map.pl_poolmap,
 					     reint_tgt, &temp_tgt);
-			if (pool_target_avail(temp_tgt, PO_COMP_ST_UPIN |
-							PO_COMP_ST_UP |
-							PO_COMP_ST_DRAIN |
-							PO_COMP_ST_NEW))
+			if (pool_target_avail(temp_tgt, PO_COMP_ST_UPIN | PO_COMP_ST_UP |
+					      PO_COMP_ST_DRAIN))
 				remap_alloc_one(diff, index, temp_tgt, true, NULL);
 			else
 				/* XXX: This isn't desirable - but it can happen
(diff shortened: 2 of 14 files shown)


src/pool/srv_pool.c
```diff
@@ -5381,6 +5381,9 @@ static int pool_find_all_targets_by_addr(struct pool_map *map,
  * \param[in]	svc		pool service
  * \param[in]	opc		update operation (e.g., POOL_EXCLUDE)
  * \param[in]	exclude_rank	for excluding ranks (rather than targets)
+ * \param[in]	extend_rank_list ranks list to be extended.
```
liw, line 5384: [Not requesting changes] I think we should consider merging `extend_rank_list` with `tgts`/`tgt_addrs` going forward, as the set of parameters of this function is growing out of hand.

liw (approved):
Also discussed with Wang Di about potential considerations related to upgrading.

src/placement/jump_map.c
```diff
@@ -256,14 +254,15 @@ get_num_domains(struct pool_domain *curr_dom, uint32_t allow_status, pool_comp_t
...
+	if (allow_status & PO_COMP_ST_UP)
 		return num_dom;
 
 	if (curr_dom->do_children == NULL || curr_dom->do_comp.co_type == fdom_lvl) {
 		next_target = &curr_dom->do_targets[num_dom - 1];
 		status = next_target->ta_comp.co_status;
 
-		while (num_dom - 1 > 0 && status == PO_COMP_ST_NEW) {
+		while (num_dom - 1 > 0 && status == PO_COMP_ST_UP &&
+		       next_target->ta_comp.co_fseq == 1) {
```
liuxuezhao, line 265: just confirm that "status == PO_COMP_ST_UP && next_target->ta_comp.co_fseq == 1" only possible for the case of POOL_EXTEND (you added code in this PR) right?
And for the other case of "status == PO_COMP_ST_UP" that is POOL_REINT, that case's co_fseq should not equal to 1?

src/placement/pl_map_common.c
```diff
@@ -287,23 +287,23 @@ determine_valid_spares(struct pool_target *spare_tgt, struct daos_obj_md *md,
...
-				DP_OID(md->omd_id),
-				DP_TARGET(spare_tgt),
+				DP_OID(md->omd_id), DP_TARGET(spare_tgt),
 				md->omd_ver);
 			spare_avail = false;
 			goto next_fail;
 		}
 
+		if (spare_tgt->ta_comp.co_in_ver < md->omd_ver &&
+		    spare_tgt->ta_comp.co_status == PO_COMP_ST_UP)
```
liuxuezhao, line 297: you add this "if" here and without change indent of below section? some comment maybe helpful, thanks.

src/pool/srv_pool.c
```diff
@@ -5427,32 +5431,57 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc,
...
-	 * If an empty target ID list is provided, convert from target
-	 * addresses.
-	 */
-	if (tgts->pti_number == 0) {
-		D_ASSERT(tgts->pti_ids == NULL);
-		D_ASSERT(tgt_addrs != NULL);
-		D_ASSERT(inval_tgt_addrs != NULL);
-		rc = pool_find_all_targets_by_addr(map, tgt_addrs, tgts,
-						   inval_tgt_addrs);
+	if (opc == POOL_EXTEND) {
```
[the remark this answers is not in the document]
wangdi1: hmm, extend is a bit different as other pool map update ops, since extend needs to add domains/targets to the pool map,  then update the target status, while other operations only update target status.  So I tend to think this if () means to generate the new pool map and extract the target lists,   then call ds_pool_map_tgts_update() to update the status.

Though if we have better algorithm for extending, it might be clearer to do extend in ds_pool_map_tgts_update() indeed.  I will check what I can do in the following patches.

src/pool/srv_pool.c
wangdi1: yeah, indeed, I do not like current parameters list as well, a bit out control here. Though sharing the parameters might make it a bit complex.  I will check. Thanks.

src/object/cli_obj.c
```diff
@@ -1012,13 +1012,17 @@ obj_shard_tgts_query(struct dc_object *obj, uint32_t map_ver, uint32_t shard,
 	shard_tgt->st_shard	= shard;
 	shard_tgt->st_shard_id	= obj_shard->do_id.id_shard;
 	shard_tgt->st_tgt_idx	= obj_shard->do_target_idx;
-	if (obj_auxi->cond_modify && (obj_shard->do_rebuilding || obj_shard->do_reintegrating))
+	if ((obj_auxi->cond_modify || obj_is_modification_opc(obj_auxi->opc)) &&
```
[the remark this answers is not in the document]
wangdi1: Say one object S1 is created and written some records,  then adds new ranks/targets, then do punch during extending, which then will punch on both "new" and "old" shards, but the shards only exist on the "old" shard, so the punch will fail with "-ENOEXIST" on the "new" shard.

So the solution here is that we set the DELAY flags on the reintegrating(new) shard, so during server forward, it will send "punch" to old/healthy shard first, then it will do create/punch by special flag on the reintegrating(DELAY) shard, instead of only punch.

src/object/cli_obj.c
```diff
@@ -6181,6 +6185,9 @@ dc_obj_punch(tse_task_t *task, struct dc_object *obj, struct dtx_epoch *epoch,
 			D_GOTO(out_task, rc);
 	}
 
+	if (api_args->flags & DAOS_COND_MASK)
+		obj_auxi->cond_modify = 1;
```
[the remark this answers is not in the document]
wangdi1: ah, yes, I will update it if it needs to refresh patch, otherwise I will update in a different patch.

src/placement/jump_map.c
wangdi1: >just confirm that "status == PO_COMP_ST_UP && next_target->ta_comp.co_fseq == 1" only possible for the case of POOL_EXTEND (you added code in this PR) right?

right.

>And for the other case of "status == PO_COMP_ST_UP" that is POOL_REINT, that case's co_fseq should not equal to 1?

correct, since reintegration can only be done from excluded(DOWNOUT) target, so co_fseq must > 1, since its initial value is 1 anyway.

src/placement/pl_map_common.c
wangdi1: ah, this should be removed. Thanks!

src/object/cli_obj.c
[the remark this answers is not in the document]
liuxuezhao: OK, as discussed by skype, here should be unnecessary for normal update.

src/object/cli_obj.c
```diff
@@ -984,7 +984,8 @@ obj_shard_tgts_query(struct dc_object *obj, uint32_t map_ver, uint32_t shard,
 	shard_tgt->st_shard	= shard;
 	shard_tgt->st_shard_id	= obj_shard->do_id.id_shard;
 	shard_tgt->st_tgt_idx	= obj_shard->do_target_idx;
-	if (obj_auxi->cond_modify && (obj_shard->do_rebuilding || obj_shard->do_reintegrating))
+	if ((obj_auxi->cond_modify || obj_is_punch_opc(obj_auxi->opc)) &&
```
liuxuezhao, line 987: probably only punch with DAOS_COND_PUNCH need to delay forward? so seems original check "if (obj_auxi->cond_modify && (obj_shard->do_rebuilding || obj_shard->do_reintegrating))" is good enough and need not change here?

liuxuezhao (approved):

wangdi1:
all tests passed.

liw (approved):

PR #10715, opened 2022-10-31
DAOS-11959 control: Improve server/engine unit tests (#10682)

mjmac:
Fixes two intermittent issues seen while running this
unit test suite:
  * Ensure that the context is not canceled before the
    child test process starts
  * Don't regenerate the fake daos_engine binary if
    it already exists


mjmac:
One unrelated failure: [DAOS-11919](https://daosio.atlassian.net/browse/DAOS-11919)

kjacque (approved):

PR #10716, opened 2022-10-31
DAOS-11887 test: Remove control/dmg_system_leader_query

mjmac:
This test was written before multiple access points was
supported, and would need to be updated in order to avoid
intermittent failures. Additionally, the test is redundant
in that the leader query/discovery mechanisms are tested
constantly as part of the normal course of running tests.

Required-githooks: true


daltonbohning (dismissed):

mjmac:
@daos-stack/daos-gatekeeper: Doesn't really seem useful to waste CI time on re-running this for green. It just removes tests.

phender (approved):

daltonbohning (approved):

PR #10731, opened 2022-11-02
DAOS-11605 common: free blob on destroy

gnailzenh:
- A blob is pinned in memory because arenas in LRU keeps refcount on it, so there is no way to actually free it on close/destroy. With this patch, all groups & arenas in LRU can release refcounts on close, so close & destroy are able to free memory.

- Blob incarnation was overwritten on open(), it is fixed by this patch.

- Change allocation size of the performance test from 64 to 256 bytes, because arena will overflow when the size is 64 (too many groups), this requires a seprate fix in follow-on patch.

Signed-off-by: Liang Zhen <liang.zhen@intel.com>

### Gatekeeper:

### DIFF
--- a/src/common/ad_mem.c
+++ b/src/common/ad_mem.c
@@ -9,12 +9,15 @@
 
 #define AD_MEM_DEBUG	0
 
-static struct ad_arena *alloc_arena(struct ad_blob *blob, bool force);
-static void free_arena(struct ad_arena *arena, bool force);
-static struct ad_group *alloc_group(struct ad_arena *arena, bool force);
-static void free_group(struct ad_group *grp, bool force);
+static struct ad_arena *arena_alloc(struct ad_blob *blob, bool force);
+static void arena_free(struct ad_arena *arena, bool force);
+static void arena_unbind(struct ad_arena *arena, bool reset);
 
-static int blob_fini(struct ad_blob *blob);
+static struct ad_group *alloc_group(struct ad_arena *arena, bool force);
+static void group_free(struct ad_group *grp, bool force);
+static void group_unbind(struct ad_group *grp, bool reset);
+
+static void blob_fini(struct ad_blob *blob);
 static int blob_register_arena(struct ad_blob *blob, unsigned int arena_type,
 			       struct ad_group_spec *specs, unsigned int specs_nr,
 			       struct ad_tx *tx);
@@ -103,7 +106,7 @@ group_decref(struct ad_group *grp)
 	D_ASSERT(grp->gp_ref > 0);
 	grp->gp_ref--;
 	if (grp->gp_ref == 0)
-		free_group(grp, false);
+		group_free(grp, false);
 }
 
 static void
@@ -118,7 +121,7 @@ arena_decref(struct ad_arena *arena)
 	D_ASSERT(arena->ar_ref > 0);
 	arena->ar_ref--;
 	if (arena->ar_ref == 0)
-		free_arena(arena, false);
+		arena_free(arena, false);
 }
 
 static inline int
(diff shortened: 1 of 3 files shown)


src/common/ad_mem.c
```diff
@@ -318,6 +308,43 @@ blob_set_opened(struct ad_blob *blob)
...
+	}
+
+	while ((arena = d_list_pop_entry(&blob->bb_ars_rsv, struct ad_arena, ar_link))) {
+		D_ASSERT(arena->ar_unpub);
+		arena_decref(arena);
+	}
+
+	D_DEBUG(DB_TRACE, "Unbind groups and arenas in LRU\n");
+	d_list_for_each_entry(group, &blob->bb_gps_lru, gp_link)
+		group_unbind(group, false);
```
wangshilong, line 336: In group_unbind(similar as following arena_unbind) we will modify gp_df->gd_back_ptr, there wont change to write it back, so we are ok to ignore this update right, a bit confusing though.

src/common/tests/ad_mem_tests.c
```diff
@@ -840,7 +841,7 @@ adt_teardown(void **state)
 	int	rc;
 
 	printf("close ad_blob\n");
-	rc = ad_blob_close(adt_bh);
+	rc = ad_blob_destroy(adt_bh);
```
wangshilong, line 844: There are still a reference leak that blob_fini() is not called properly.

In ad_blob_prep_open() if it is dummy_blob, ref will be 2:
                      blob = dummy_blob;
 469                         D_DEBUG(DB_TRACE, "found dummy blob, refcount=%d\n", blob->bb_ref);
 470                         blob_addref(blob);

maybe 
blob_decref(adt_bh.bh_blob)
need be called before destroy?

src/common/ad_mem.c
gnailzenh: yes, this is how back_ptr works, we don't need it to be written back to storage. 
We use the durable arena & group, which is addressed by offset, to index in-memory data structure. So we don't have to create another huge tree or hash table to index them.
"Incarnation" is the way that we can detect stale pointers.

src/common/tests/ad_mem_tests.c
gnailzenh: To be clear, the global variable "dummy_blob" has no refcount (it is reset to NULL when the last opener closed), multiple openers can find the same "dummy_blob" on open, instead of creating their own instance. 

The main reason dummy_blob has no refcount is because I don't want to pin it in memory so I can test blob_open(). If it has refcount, then the global blob is never going to be freed, so blob_open() will not be tested at all.

liuxuezhao (approved):

src/common/tests/ad_mem_tests.c
wangshilong: I added some codes it showed there are still leaks somewhere:
[root@daos1 daos]# git diff .
diff --git a/src/common/ad_mem.c b/src/common/ad_mem.c
index d91a228551..bc46f35d12 100644
--- a/src/common/ad_mem.c
+++ b/src/common/ad_mem.c
@@ -562,6 +562,7 @@ ad_blob_destroy(struct ad_blob_handle bh)

        /* TODO: remove the file */
        blob_close(blob);
+       printf("(blob->bb_ref: %d\n", blob->bb_ref);
        blob_decref(blob);
        return 0;
 }
close ad_blob
(blob->bb_ref: 2
[==========] 16 test(s) run.

wangshilong (approved):
anyway, I could try to fix that problem.

src/common/tests/ad_mem_tests.c
gnailzenh: thanks, I will check

gnailzenh: found the issue, it's an issue in test program, I forget to publish/cancel the last reserve in adt_delayed_free_1(), because it's not a bug in allocator itself, so I will submit another patch to fix it.

PR #10744, opened 2022-11-02
DAOS-11829 object: add a DAOS API to allow users to set an anchor at …

mchaarawi:
…specific key

- add also DFS API on top of that to allow readdir to resume at an anchor set at a particular dir entry.

Required-githooks: true

### Gatekeeper:


mchaarawi:
ping reviewers (@liuxuezhao @wangdi1 @jolivier23)

daltonbohning (commented):
ftest LGTM

src/object/cli_obj.c
```diff
@@ -6112,6 +6136,101 @@ dc_obj_list_rec(tse_task_t *task)
...
+	obj_auxi->dkey_hash = obj_dkey2hash(obj->cob_md.omd_id, args->dkey);
+	grp_idx = obj_dkey2grpidx(obj, obj_auxi->dkey_hash, map_ver);
+	if (grp_idx < 0) {
+		D_ERROR(DF_OID" can not find grp %d\n", DP_OID(obj->cob_md.omd_id), grp_idx);
+		D_GOTO(err_obj, rc = grp_idx);
+	}
+
+	if (obj_auxi->is_ec_obj) {
+		rc = obj_ec_get_parity_or_alldata_shard(obj_auxi, map_ver, grp_idx, args->dkey,
+							&shard_cnt);
```
wangdi1, line 6191: This won't work for parity rotation EC objects,  since EC parity object actually allocate sub anchors for each ec shards within one group, and then track these sub anchors individually.  

So I assume you will need allocate sub anchors for all shards within one EC group, and set anchors individually on each shard.  see obj_list_common()->sub_anchors_prep().

Hmm, set anchor seems only useful for directory, but since EC object is rarely used for directory, probably just disable this for EC object?

mchaarawi: on obj_list_common() we call sub_anchors_prep() only if shard_cnt > 1
here shard_cnt is always 1 right? maybe i don't understand..
but i cannot assume that users would not use EC for directory im afraid

wangdi1: >on obj_list_common() we call sub_anchors_prep() only if shard_cnt > 1
here shard_cnt is always 1 right?

shard_cnt will be > 1 if parity rotation is enabled or all parity failed (in degraded mode).

> but i cannot assume that users would not use EC for directory im afraid

it can return failure? since this API is specifically for directory, which will rarely use EC object anyway. or I miss other use case?

mchaarawi: we cannot assume that users will always use non-EC for directory.
let's say a middleware layer like dfuse uses this function to implement opendir/seekdir/readdir. then a user creates an EC directory (above dfuse control). normal readdir will not work anymore on this directory because it calls this function. so we shouldn't assume this at this layer.

wangdi1: >let's say a middleware layer like dfuse uses this function to implement opendir/seekdir/readdir. then a user creates an EC directory (above dfuse control).

I understand the object type is not controlled by dfuse, though the directory object type can be controlled by hint/property?  so as long as we make it clear in our document, it should be fine? Besides, this does not seems a "general" API to me? But I might miss some use cases other than directory entry fetch?

If you insist to implement this for EC object, then it will need prepare the top anchor to create sub anchors for each shard as obj_list_common(),  then forward the request to all shards, in complete_callback(), it will set the those reply anchors in each sub_anchors individually see anchor_update_check_eof()->update_sub_anchor_cb().

So I am not sure if it is worth to implement these complexity for EC object, unless this api will be used to implement standard dfs_api() ?

jolivier23: If the dkey tree exists on a shard, the da_buf will be the same independent of which shard fills it in.  So if there is any extra work to do there to fill out multiple subanchors, I think it can be done independently.   But I guess the issue may be that you don't know which shard may have the dkey tree allocated so that is why you need to send to all?

jolivier23: In fact, it doesn't really even matter which shard you send it to if a tree exists for the dkey/akey you are trying to turn into an anchor.

jolivier23: So is there no way to just reject EC object class for directories?  It is kind of terrible and probably not what the user actually wants.   It's really no different than replication, right?

Actually, that brings up an interesting point.   This should work as is if the EC object is just using replication (e.g. no full stripe writes), no?

mchaarawi: ok so turns out with current master, we cannot event do dkey_list on EC object - get a segfault. so maybe that sub-anchor stuff is already broken? @wangdi1 ?

wangdi1: hmm, we have some simple ec enumeration tests in daos_obj.c & daos_obj_ec.c. Maybe there is some bugs there. stack trace? create a ticket?

mchaarawi: yes already done:
https://daosio.atlassian.net/browse/DAOS-12063

mchaarawi: @wangdi1 after reviewing this more, in this case since we have a dkey passed in for the set_anchor, shard_cnt will always be 1. in this case, for enumeration, nothing is done for the sub anchor:
https://github.com/daos-stack/daos/blob/master/src/object/cli_obj.c#L6048
So im still confused by all this subanchor stuff and what it has to do with this case.. i do see that if we use EC directory, the enumeration does not start from the correct place though, so there is something wrong with this and EC.

mchaarawi: OK i think i figured out why shard_nr was one in my case. it should be 4 but it was getting overwritten. now that i have this, i need to figure out how to allocate the subanchor, since in the case of enumeration, it relies on the user provided sgl, but in this case here, there is no sgl since the user is not doing enumeration at this point. any idea?

mchaarawi: actually i was wrong. shard_cnt is always returned as 1 here:
https://github.com/daos-stack/daos/blob/master/src/object/cli_obj.c#L5860

mchaarawi: ok i see. updated the code to support EC object. i pulled in your fix for enumeration for EC object to test it and it works fine with EC directory (tested locally). probably should land your PR https://github.com/daos-stack/daos/pull/10834 first then rebase this after.

jolivier23 (dismissed):
Looks like it needs rebase now

jolivier23 (dismissed):

src/object/cli_obj.c
```diff
@@ -6123,6 +6197,115 @@ dc_obj_list_rec(tse_task_t *task)
...
+		D_ERROR(DF_OID" can not find grp %d\n", DP_OID(obj->cob_md.omd_id), grp_idx);
+		D_GOTO(err_obj, rc = grp_idx);
+	}
+
+	if (obj_auxi->is_ec_obj) {
+		struct shard_anchors	*sub_anchors = NULL;
+
+		rc = obj_ec_get_parity_or_alldata_shard(obj_auxi, map_ver, grp_idx, args->dkey,
+							&shard_cnt);
+		shard_cnt = obj_get_grp_size(obj);
```
wangdi1, line 6255: probably it should use the shard_cnt got from obj_ec_get_parity_or_alldata_shard(), in case parity rotation is not enabled (shard_cnt = 1), and also obj_get_grp_size() will also include the extending shard during online OSA, which is not needed for enumeration.

src/object/cli_obj.c
```diff
@@ -3989,6 +4035,15 @@ obj_comp_cb_internal(struct obj_auxi_args *obj_auxi)
 
 	if (obj_is_enum_opc(obj_auxi->opc))
 		rc = obj_list_comp(obj_auxi, &iter_arg);
+	else if (obj_auxi->opc == DAOS_OBJ_RPC_KEY2ANCHOR) {
+		daos_obj_key2anchor_t *obj_arg = dc_task_get_args(obj_auxi->obj_task);
+
+		if (!obj_arg->anchor->da_sub_anchors || !obj_is_ec(obj_auxi->obj))
+			goto out;
+		rc = obj_auxi_shards_iterate(obj_auxi, k2a_update_sub_anchor_cb, NULL);
```
wangdi1, line 4043: Hmm, seems no different with other key enumeration operations.  I think obj_list_comp() might just work by using obj_list_dkey_cb().

wangdi1 (dismissed):
Probably need some update, though it can be done once PR-10498 is landed, not sure which one will be landed first.

src/object/cli_obj.c
mchaarawi: better to keep things separate for better readability?

src/object/cli_obj.c
mchaarawi: well i think shard_cnt was always returned as 1 so i had to do this, but i will check again to make sure.

mchaarawi: ok i confirmed that this is required as obj_ec_get_parity_or_alldata_shard always returning 1 for shard_cnt.. since i assume dkey is not NULL. so to be able to go query all shards, we have to do this.

wangdi1: ah, hmm, probably should be

if (obj_ec_parity_rotate_enable())
         shard_cnt = daos_oclass_grp_size();

wangdi1 (approved):

mchaarawi:
@wangdi1 only 1 failure here in: FTEST_osa.OSAOnlineExtend
that should be unrelated to this PR, right?

wangdi1:
> @wangdi1 only 1 failure here in: FTEST_osa.OSAOnlineExtend
> that should be unrelated to this PR, right?

yes, DAOS-12031

mchaarawi:
@jolivier23 need another review from you please since i had to rebase with conflicting PR

jolivier23 (approved):

PR #10745, opened 2022-11-02
DAOS-11401 common: umem cache API for tracking dirty pages

jolivier23:
Adds basic umem cache implementation supporting 16MB page size
Implements synchronous copy
Tracks changes to 16KB chunks
Implements initial checkpoint ULT

Remaining work:
Better testing coverage
Hook up with ad_mem
More sophistication in the scheduler, as it stands it has same priority as aggregation and garbage collection.
Asynchronous copy support
Page eviction support

Required-githooks: true

Co-authored-by: Liang Zhen <liang.zhen@intel.com>

### Gatekeeper:


src/common/mem.c
```diff
@@ -1573,4 +1573,362 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+			d_list_add_tail(&page->pg_link, &cache->ca_pgs_waiting);
+
+			memset(&page->pg_bmap[0], 0, sizeof(page->pg_bmap));
+			inflight++;
+		}
+
+		store->stor_ops->so_wal_query_committed(store, &committed_tx);
+
+		d_list_for_each_entry_safe(page, tmp, &cache->ca_pgs_waiting, pg_link) {
+			if (page->pg_last_checkpoint > committed_tx)
```
NiuYawei, line 1912: We have an ID compare callback provided, see the comment in wal_id_cmp(), it tries to handle sequence overflow case, which is very unlikely though.

src/include/daos/mem.h
```diff
@@ -862,4 +863,217 @@ struct umem_action {
...
+
+#define UMEM_CACHE_BMAP_SZ        (1 << (UMEM_CACHE_PAGE_SZ_SHIFT - UMEM_CACHE_CHUNK_SZ_SHIFT - 6))
+
+/** 16 MB page */
+struct umem_page {
+	/** page ID */
+	unsigned int		 pg_id;
+	/** refcount */
+	int			 pg_ref;
+	uint64_t                 pg_dirty : 1, /** Page is dirty */
```
NiuYawei, line 882: The dirty bit looks redundant to me, we could see if the page is 'dirty' by comparing the 'pg_last_inflight' with global 'last checkpointed ID', right?

src/common/mem.c
```diff
@@ -1573,4 +1573,362 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+next_bmap:
+		offset += UMEM_CACHE_CHUNK_SZ << 6;
+		page_addr += UMEM_CACHE_CHUNK_SZ << 6;
+	}
+	sgl->sg_nr_out = sgl->sg_nr = nr;
+	store_iod->io_nr            = nr;
+}
+
+int
+umem_cache_checkpoint(struct umem_cache *cache, uint64_t wal_tx)
```
NiuYawei, line 1831: This 'wal_tx' is supposed to be the "checkpoint to ID" passed in by caller? If it's not useful, we can remove it from the parameters.

jolivier23: Yeah, it should be the last ID committed but not necessarily to the WAL.   It slightly more efficient to pass it in than to walk through the pages and calculate it.

src/include/daos/mem.h
jolivier23: Removed.

src/common/mem.c
jolivier23: Will fix this

src/include/daos/mem.h
```diff
@@ -862,4 +863,217 @@ struct umem_action {
 	};
 };
 
+#define UMEM_CACHE_PAGE_SZ_SHIFT  24 /* 16MB */
+#define UMEM_CACHE_PAGE_SZ        (1 << UMEM_CACHE_PAGE_SZ_SHIFT)
+#define UMEM_CACHE_PAGE_SZ_MASK   (UMEM_CACHE_PAGE_SZ - 1)
+
+#define UMEM_CACHE_CHUNK_SZ_SHIFT 14 /* 16KB */
```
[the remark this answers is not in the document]
jolivier23: Good question. It would mean the bit mask would take 512 bytes instead of 128 per page.  So it would mean 32MB instead of 8MB for a 1TB store

src/include/daos/mem.h
```diff
@@ -862,4 +863,217 @@ struct umem_action {
 	};
 };
 
+#define UMEM_CACHE_PAGE_SZ_SHIFT  24 /* 16MB */
```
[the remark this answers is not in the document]
jolivier23: Possibly it is too large.  We can play with these numbers when we get some experience.  Every different option will have tradeoffs.

src/common/mem.c
```diff
@@ -1573,4 +1573,428 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+		 *  them from being moved to another list by an I/O operation.
+		 */
+		page->pg_waiting = 1;
+		if (store->stor_ops->so_wal_id_cmp(store, page->pg_last_inflight, chkpt_id) > 0)
+			chkpt_id = page->pg_last_inflight;
+	}
+
+	do {
+		/** first try to add up to MAX_INFLIGHT pages to the waiting queue */
+		while (inflight < MAX_INFLIGHT &&
```
NiuYawei, line 1931: Ideally, the MAX_INFLIGHT should be based on the amount of data to be flushed but not the page count. Shouldn't we iterate the page list to generate an "umem_store_iod" for multiple pages?

src/common/mem.c
```diff
@@ -1573,4 +1573,428 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+	 *  handle that.
+	 */
+	struct umem_store_region cd_regions[UMEM_CACHE_BMAP_SZ / 2];
+	d_iov_t                  cd_iovs[UMEM_CACHE_BMAP_SZ / 2];
+	/** Handle for the underlying I/O operations */
+	daos_handle_t            cd_fh;
+};
+
+static void
+page2chkpt(struct umem_page *page, struct umem_checkpoint_data *chkpt_data)
```
NiuYawei, line 1807: Can we expand this function to generate 'chkpt_data' for multiple pages?

src/common/mem.c
```diff
@@ -1573,4 +1573,428 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+
+		wait_cb(store, page->pg_last_checkpoint, &committed_tx, arg);
+
+		D_ASSERT(store->stor_ops->so_wal_id_cmp(store, committed_tx,
+							page->pg_last_checkpoint) >= 0);
+
+		/** Since the flush API only allows one at a time, let's just do one at a time
+		 *  before copying another page.  We can revisit this later if the API allows
+		 *  to pass more than one fh.
+		 */
```
NiuYawei, line 1980: What's the major difficult of putting multiple pages into single flush handle (umem_checkpoint_data)?

src/include/daos/mem.h
```diff
@@ -100,7 +100,7 @@ struct umem_store_ops {
 	int	(*so_flush_prep)(struct umem_store *store, struct umem_store_iod *iod,
 				 daos_handle_t *fh);
 	int	(*so_flush_copy)(daos_handle_t fh, d_sg_list_t *sgl);
-	int	(*so_flush_post)(daos_handle_t fh, int err);
+	int (*so_flush_post)(daos_handle_t fh, int err);
```
NiuYawei, line 103: unexpected change?

src/pool/srv_pool_chkpt.c
```diff
@@ -0,0 +1,143 @@
...
+				ctx.cc_dmi->dmi_tgt_id);
+			break;
+		}
+		if (rc != 0) {
+			D_ERROR("Issue with VOS checkpoint (tgt_id: %d): " DF_RC "\n",
+				ctx.cc_dmi->dmi_tgt_id, DP_RC(rc));
+			sleep_time = 60000; /* wait longer if there's an error */
+		}
+
+		sched_req_sleep(child->spc_chkpt_req, sleep_time);
```
NiuYawei, line 101: I'm wondering if we also need to query WAL usage more frequently to see if WAL space is under pressure and trigger checkpoint promptly. The timing of checkpointing could be optimized in follow-on PRs.

src/bio/bio_wal.c
```diff
@@ -158,6 +158,24 @@ bio_wal_id_cmp(struct bio_meta_context *mc, uint64_t id1, uint64_t id2)
...
+		notify_cb(si->si_commit_id, arg);
+		return -DER_ALREADY;
+	}
+
+	si->si_ckp_wait_id  = chkpt_id;
+	si->si_ckp_wait_arg = arg;
+	si->si_notify_cb    = notify_cb;
+
+	return 0;
+}
```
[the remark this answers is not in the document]
jolivier23: Yeah, I can look reworking that.

src/common/mem.c
```diff
@@ -1573,4 +1573,428 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+		 *  be committed to the WAL before it can be flushed.
+		 */
+		d_list_del(&page->pg_link);
+		d_list_add_tail(&page->pg_link, &cache->ca_pgs_dirty);
+	}
+
+	if (page->pg_last_inflight == wr_tx)
+		return;
+
+	if (store->stor_ops->so_wal_id_cmp(store, wr_tx, page->pg_last_inflight) > 0)
```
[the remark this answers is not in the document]
jolivier23: ok

src/vos/vos_pool.c
```diff
@@ -176,20 +176,102 @@ vos_wal_id_cmp(struct umem_store *store, uint64_t id1, uint64_t id2)
...
+	struct umem_store    *store;
+	int                   rc;
+
+	pool = vos_hdl2pool(ctx->cc_vos_pool_hdl);
+	D_ASSERT(pool != NULL);
+
+	umm   = vos_pool2umm(pool);
+	store = &umm->umm_pool->up_store;
+
+	rc = bio_wal_ckp_start(store->stor_priv, &tx_id);
```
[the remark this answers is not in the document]
jolivier23: I started using it to avoid calling commit if we haven't done anything since last time.

src/common/mem.c
```diff
@@ -1573,4 +1573,428 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+			page->pg_last_checkpoint = page->pg_last_inflight;
+
+			rc = store->stor_ops->so_flush_copy(chkpt_data->cd_fh,
+							    &chkpt_data->cd_sg_list);
+			/** If this fails, it means invalid argument, so assertion here is fine */
+			D_ASSERT(rc == 0);
+
+			page->pg_copying = 0;
+
+			page_insert_sorted(store, page, &cache->ca_pgs_waiting);
```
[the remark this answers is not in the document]
jolivier23: The idea is to keep the pipeline full.  Suppose MAX_INFLIGHT were 2 and you had two pages in the queue.  We will issue a wait for the checkpoint to commit.   By picking the lowest value to wait for, we can add another entry to the queue and kick it off while the higher checkpoint is in process of committing to the WAL.  Since the list is small, the overhead is pretty negligible.

src/common/mem.c
jolivier23: We can potentially look into improvements.   There is a tradeoff here.   Keeping pages separate is simpler and allows me to preallocate enough space to hold the iod for a page ahead of time as I know there will be a limit on possible iod entries within one 16 MB block.  We could certainly look at pushing multiple pages at a time.  But in that case, I think we would want to try to select the least recently updated pages first.

src/common/mem.c
jolivier23: It's possible, yes.  SEe the comment above.   One assumption that is simpler with doing this on a single page basis is the size of the umem_chkpoint_data struct is fixed.  If I handle multiple pages, I either have to add some dynamic sizing or split a page or rollback page iod if there isn't enough space.  Anyway, this seemed simpler.

We did discuss this before your break a bit.   One way to handle more than one page at a time is change the post callback to handle more than one fh.

anyway, if you feel it would be better to combine pages, I can certainly look into that.   But this is currently working so I think it may be better to commit to a line in the sand.

src/common/mem.c
jolivier23: Basically the allocation of the iod struct becomes a little less static and I don't know how many iod entries I could have until I read the bitmap.  Right now, the chkpt_data struct allows for a maximum of number of chunks/2 since worst case would be every other chunk.

Anyway, it could be done, it's just not how I did it to start.

src/include/daos/mem.h
jolivier23: I have clang-format enabled so it probably did this.

src/pool/srv_pool_chkpt.c
jolivier23: Indeed, this is a first cut to get things working.   This code will need revisiting.   I don't think we should be at same priority as gc and aggregation.

jolivier23:
@NiuYawei @gnailzenh can we land as is and I can address the various comments in a follow-up patch?  I'd like to unblock anyone waiting on this.

src/pool/srv_target.c
```diff
@@ -397,6 +403,7 @@ pool_child_delete_one(void *uuid)
 
 	D_ASSERT(d_list_empty(&child->spc_cont_list));
 	d_list_del_init(&child->spc_list);
+	ds_stop_chkpt_ult(child);
```
NiuYawei, line 406: I think we'd better stop checkpoint ULT when ds_pool_child drop the last reference in ds_pool_child_put(), otherwise, there could be potential deadlock issue:

1. I/O request ULT calls bio_wal_reserve(), it find out that WAL space is insufficient, then bio_wal_reserve() yield and wait for checkpoint to reclaim WAL space.
2. Pool is destroyed by force, pool_child_delete_one() is called to stop checkpoint ULT.
3. bio_wal_reserve() will wait forever since the WAL space won't be reclaimed.

I'm not sure if we still support force pool destroy today, but seems to me stopping such internal ULTs at the end (after all other users dropped reference) is a good practice.

src/vos/vos_pool.c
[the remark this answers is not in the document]
NiuYawei: Sure, we could simplify the APIs in follow-on PRs.

NiuYawei (approved):
Another missing part on top my head is to complete replay:
1. Load meta pages before replay;
2. call touch_page in replay callback;

It related to umem page cache, not sure if it should be included in this PR.

src/common/mem.c
```diff
@@ -1573,4 +1573,362 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+	struct umem_page *end_page = umem_cache_off2page(cache, addr + size - 1) + 1;
+
+	while (page != end_page) {
+		D_ASSERT(page->pg_ref >= 1);
+		page->pg_ref--;
+		page++;
+	}
+
+	return 0;
+}
```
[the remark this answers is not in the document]
jolivier23: yeah

src/common/mem.c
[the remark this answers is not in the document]
jolivier23: Maybe could assert >=0.  It can be equal

src/pool/srv_pool_chkpt.c
```diff
@@ -0,0 +1,143 @@
...
+	}
```
gnailzenh, line 102: do we want to run the last round checkpoint even the ULT is exiting.

NiuYawei (approved):

src/common/mem.c
```diff
@@ -1573,4 +1573,473 @@ umem_tx_publish(struct umem_instance *umm, struct umem_rsrvd_act *rsrvd_act)
...
+
+int
+umem_cache_pin(struct umem_store *store, umem_off_t addr, daos_size_t size)
+{
+	struct umem_cache *cache     = store->cache;
+	struct umem_page *page      = umem_cache_off2page(cache, addr);
+	struct umem_page *end_page  = umem_cache_off2page(cache, addr + size - 1) + 1;
```
gnailzenh, line 1688: why do we need to "+1" at here? e.g. if addr=0, size=4K, we only pin one page not two right?

src/common/dav/wal_tx.c
```diff
@@ -259,11 +272,17 @@ dav_wal_tx_set_bits(void *hdl, void *addr, uint32_t pos, uint16_t num_bits)
...
 	struct dav_tx		*tx = utx2wtx(dav_hdl->do_utx);
 	struct wal_action	*wa_redo;
+	int                      rc;
 
 	D_ASSERT(hdl != NULL);
 	if (addr == NULL)
 		return -DER_INVAL;
 
+	rc = umem_cache_touch(dav_hdl->do_store, dav_hdl->do_utx->utx_id,
+			      mdblob_addr2offset(tx->wt_dav_hdl, addr), sizeof(uint64_t));
```
gnailzenh, line 282: interesting, collocated metadata can reduce the write amplification, but it could potentially increase the chance of TX restart while checkpointing.

gnailzenh (approved):

PR #10791, opened 2022-11-07
DAOS-9799 chk: Allow checker policies to be set all-interactive

mjmac:
Provides two new set-policy flags:
  * dmg check set-policy --all-interactive
  * dmg check set-policy --reset-defaults

These flags can be used instead of setting each policy's action
individually.

Additionally, adds UI support for the new checker flags added
in a previous commit.
  * dmg check start --auto=[on|off] --failout=[on|off]

Required-githooks: true


mjmac:
hw-medium failures are unrelated.

PR #10746, opened 2022-11-03
DAOS-3916 container: fix metadata times in CONT_OPEN_BYLABEL

kccain:
In commit d37f4fb the interoperability mechanism was not properly implemented for RPC CONT_OPEN_BYLABEL due to an oversight in rpc.h.

Without this change, when daos_cont_open() is invoked specifying a valid container label, the daos_cont_info_t information contains junk values in structure members ci_md_otime and ci_md_mtime. The problem is seen in client log messages when running daos_test -p (pool tests). The test still passes. It is also visible if daos_test -c (container tests) is changed to open containers by label (instead of UUID string typically done). Container test co_mdtimes() eplicitly fails in this scenario.

With this change, the RPC protocol version mechanism is corrected for opcode CONT_OPEN_BYLABEL, to use latest (v7) message definitions when applicable. Also, the daos_test co_mdtimes() test is updated to exercise open by label (and therefore catch any future regression).

Required-githooks: true

Signed-off-by: Kenneth Cain <kenneth.c.cain@intel.com>

### Before requesting gatekeeper:

* [x] Two review approvals and any prior change requests have been resolved.
* [x] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [x] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [x] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate watchers.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficent testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
[...]

PR #9985, opened 2022-08-15
DAOS-3916 container: query metadata open, close/modify times

kccain:
With this change, daos_cont_open() and daos_cont_query() return the
most recent open and close/modify times in the daos_cont_info_t
output argument. The new information is returned as hybrid logical
clock (HLC) values. Container operations that only read the metadata
state will not update these times, while other operations will update
the modify time. This is envisioned as a basic mechanism for a user
to identify containers not used recently (from a metadata standpoint,
not an IO standpoint).

A future patch is envisioned to provide a pool list containers
interface with some filtering criteria, to find containers that
may fit some user-determined criteria for migration or removal.

Main changes summary
- container properties KVS includes new key/val for metadata times
  (using same DAOS_POOL_GLOBAL_VERSION=2 for master/release 2.4 dev).
- pool/container upgrade code changed to initialize metadata times.
- libdaos API minor version incremented (v2.3.0 -> v2.4.0)
- daos_cont_info_t.ci_pad, .ci_redun_lvl repurposed to make space
  for the new time fields while keeping the same structure size.
  (ci_redun_lvl is otherwise available through container properties).
- daos_test container test added (co_mdtimes).
- daos_test container tests modified to check redundancy level via
  property value rather than daos_cont_info_t field.
- daos utility output (both JSON and human-readable when run with -v)
  contains the new metadata time information.
- CaRT protocol for CONT_OPEN, CONT_OPEN_BYLABEL, and CONT_QUERY
  existing version (6) maintained, and new version (7) added that
  returns metadata open and close/modify times.
- engine code register and handle protocol v6 or v7 RPCs.
- client code registers and uses uses only the new/v7 protocol.
  (Possible future change: client query engine then register v6 or v7).

Other changes:
- change dmg and pool functional tests to expect pool layout v2
  following recent DAOS_POOL_GLOBAL_VERSION update. DAOS-11282.
- satisfy pylint suggestion for f-string in SConstruct, API_VERSION.
- satisfy jira_query.py failure, add 'md' as a valid component
   in PR/commit titles.

Features: pool container

Required-githooks: true

Signed-off-by: Kenneth Cain <kenneth.c.cain@intel.com>

PR #10793, opened 2022-11-07
DAOS-3916 container: add number of handles to rdb, query output

kccain:
Before this change, the daos_cont_info_t structure returned by daos_cont_open() and daos_cont_query() includes an unnecessary value, ci_redun_fac (it is otherwise available in the properties). Additionally, callers to these APIs and daos_pool_filter_containers() have no visibility into the number of open handles on a container.

With this change, the daos_cont_info_t structure replaces the uint32_t member ci_redun_fac with ci_nhandles. The daos utility output (both human-readable and JSON format) is updated to output this quantity. The libdaos API minor version is also updated. The container open and query CaRT RPC protocols are  updated to return (where previously a pad value was specified) this 32-bit quantity. The protocol version is not changed, since it was already updated for this feature (metadata time tracking, and filtering based on metadata criteria). The daos_pool_filter_containers() API and implementation is updated to allow filtering based on a container's number of open handles. The daos_test pool and container test suites are updated to utilize the changes in the API functions and returned information.

The daos_engine container open, query and close handling is updated to get and increment/decrement the number of handles maintained in rdb. And engine code that performs pool/container metadata layout upgrades is updated to initialize the number of handles to zero when upgrading an older container (e.g., one created with DAOS v2.2). The pool filter containers handling is updated to process filters that include conditional tests based on the number of container open handles.

Other changes include:
- a small fix to pool_cont_filter_is_valid() engine code to inspect the key and comparison function specified in the RPC against the proper minimum and maximum bounds.
- a small change to the pool/container upgrade code for metadata times, updating modify time only (not open + modify times) for containers that already have metadata times in their rdb. This will fix a future upgrade scenario (from DAOS v2.4 to the following version).

Required-githooks: true

Signed-off-by: Kenneth Cain <kenneth.c.cain@intel.com>

### Gatekeeper:


kccain:
Build 4 is in progress (all except functional hw medium and large have run, no test failures so far). However, the master on which this is based has some known test failures that I expect to see in that medium/large hw testing. While that is getting sorted out independently, in parallel I would like to open this up for code reviews.

src/control/cmd/daos/container.go
```diff
@@ -733,7 +733,7 @@ func printContainerInfo(out io.Writer, ci *containerInfo, verbose bool) error {
 	if verbose {
 		rows = append(rows, []txtfmt.TableRow{
 			{"Pool UUID": ci.PoolUUID.String()},
-			{"Container redundancy factor": fmt.Sprintf("%d", *ci.RedundancyFactor)},
+			{"Number of open handles": fmt.Sprintf("%d", *ci.NumHandles)},
```
mchaarawi, line 736: as i understand from this PR, you are changing the output of daos cont query to not include:
`Container redundancy factor : x`
and have instead
`Number of open handles   : x`
?

the RF is something useful really that users would like to see on cont query. im ok with the change to the info structure to remove that since we can grab that programmatically from the props, but visually on cont query it has been helpful to have the RF there..

src/include/daos_types.h
```diff
@@ -132,8 +132,8 @@ typedef struct {
 	uuid_t			ci_uuid;
 	/** Epoch of latest persistent snapshot */
 	daos_epoch_t		ci_lsnapshot;
-	/** Redundancy factor */
-	uint32_t		ci_redun_fac;
+	/** Number of open handles */
+	uint32_t		ci_nhandles;
```
mchaarawi, line 136: why can't we have both (adding ci_nhandles at the end)?

src/tests/suite/daos_container.c
```diff
@@ -2240,9 +2263,16 @@ co_rf_simple(void **state)
...
-		rc = daos_cont_query(arg->coh, &info, NULL, NULL);
+		daos_prop_t	*prop_out = daos_prop_alloc(1);
+
+		assert_non_null(prop_out);
+		prop_out->dpp_entries[0].dpe_type = DAOS_PROP_CO_REDUN_FAC;
+
+		rc = daos_cont_query(arg->coh, &info, prop_out, NULL);
 		assert_rc_equal(rc, 0);
-		assert_int_equal(info.ci_redun_fac, DAOS_PROP_CO_REDUN_RF2);
+		assert_int_equal(prop_out->dpp_entries[0].dpe_val, DAOS_PROP_CO_REDUN_RF2);
```
mchaarawi, line 2273: we have internal functions:
dc_cont_hdl2redunfac and dc_cont_hdl2redunlvl
does it make sense to make a public API for those? (not necessarily with this PR)

src/tests/suite/daos_container.c
```diff
@@ -52,7 +54,28 @@ co_create(void **state)
 				    &info, arg->async ? &ev : NULL);
 		assert_rc_equal(rc, 0);
 		WAIT_ON_ASYNC(arg, ev);
+		assert_int_equal(info.ci_nhandles, 0);
```
mchaarawi, line 57: not quite sure i get why this should be 0 and not 1. we just opened 1 handle.

src/tests/suite/daos_container.c
```diff
@@ -52,7 +54,28 @@ co_create(void **state)
...
 		print_message("container opened\n");
+
+		/* Open a second time to verify num handles was incremented */
+		rc = daos_cont_open(arg->pool.poh, str, DAOS_COO_RO, &coh2,
+				    &info, arg->async ? &ev : NULL);
+		assert_rc_equal(rc, 0);
+		WAIT_ON_ASYNC(arg, ev);
+		assert_int_equal(info.ci_nhandles, 1);
```
mchaarawi, line 65: same here, since we have 2 handles now, it should be 2?

src/tests/suite/daos_container.c
```diff
@@ -52,7 +54,28 @@ co_create(void **state)
...
+		assert_rc_equal(rc, 0);
+		WAIT_ON_ASYNC(arg, ev);
+		print_message("container closed (coh2)\n");
+
+		/* Query with first handle (coh), verify num handles was decremented 2 -> 1 */
+		info.ci_nhandles = 0xABCD0123;
+		rc = daos_cont_query(coh, &info, NULL, arg->async ? &ev : NULL);
+		assert_rc_equal(rc, 0);
+		WAIT_ON_ASYNC(arg, ev);
+		assert_int_equal(info.ci_nhandles, 1);
```
mchaarawi, line 78: this one sounds correct because one handle is open.. so not sure i get it

src/include/daos_types.h
kccain: It would break the ABI that to my understanding cannot be done until the next major API version increment (that cannot be done in DAOS 2.4).

src/tests/suite/daos_container.c
kccain: Like for the fetching of the container metadata (open, modify) times I decided that the number of open handles returned should reflect the state before the open, instead of including the effect of the open.

src/tests/suite/daos_container.c
kccain: Here this is  a container query (as opposed to the open whose "post" execution effect would be to increment the number of open handles). So in this case query returns the number that reflects the execution of the previous open.

src/include/daos_types.h
mchaarawi: you are already breaking the API, which IMO is actually worse

src/tests/suite/daos_container.c
mchaarawi: this is counter intuitive IMO. idk about the time values, but for this particular count, for a user we really should return the actual count of open handles at that point where they are doing the checking. the user has just opened the container. getting back 0 open handle count sounds wrong IMO.

mchaarawi (changes requested):
we should probably add some more robust testing part of this PR:
1) verify open count from several ranks / threads opening the container
2) verify the behavior with l2g/g2l functions

src/include/daos_types.h
kccain: Other than the daos_test are there other users of the ci_redun_fac structure member? Also in a previous commit I removed the other ci_redun_lvl member and we discussed the reasoning for making this subsequent change to ci_redun_fac as well https://github.com/daos-stack/daos/pull/9985#discussion_r952610355

This patch is a "nice to have" addition to what I've already implemented for daos_pool_filter_containers(). So, if we think that the API change (removing ci_redun_fac) is too severe, we can always just decide to abandon this patch (and give up the ability to track number of open handles). I think that would be a bad decision, since it seems this could be useful for a datamover use case (e.g., filter a pool's containers based on this and other criteria, to decide what data to migrate).

mchaarawi: "Other than the daos_test are there other users of the ci_redun_fac structure member?"
i don't know.. it's in a public header, so anyone can be using it. in the middleware i control / support, it's not used directly.
my comment is just in relation to your comment: "It would break the ABI"
you are already doing worse than that by breaking the API :-) so my point is that just adding the variables is not going to be worse that replacing an existing one.

mchaarawi: just to be clear, my change request is not for this. im fine if you want to just replace it here. hopefully no one is using it..
The main issue for this PR for me is the reported count in the num handles and not reporting the RF in the daos cont query (which seems unnecessary to remove in this case since you can just retrieve that from the open handle?)

src/tests/suite/daos_container.c
kccain: Yes it is counter-intuitive. I'm going to look at returning nhandles to reflect the effect of the current open being executed.

And while doing this I will review the container metadata open time and whether it can be pre-updated to be consistent with what will be done with nhandles here. Open time has one tricky special case to account for (no update at all) - perhaps it is that case that made it seemingly difficult to pre-update the open time (when we do want the update to occur).

src/tests/suite/daos_container.c
kccain: will pre-increment nhandles in the next push

src/tests/suite/daos_container.c
kccain: yes, will fix in the next push

src/tests/suite/daos_container.c
kccain: I have local commits that both pre-increment nhandles, and  pre-update the open time - will have both in the next push

src/container/srv_container.c
```diff
@@ -1981,6 +2046,17 @@ cont_open(struct rdb_tx *tx, struct ds_pool_hdl *pool_hdl, struct cont *cont,
 			DP_CONT(cont->c_svc->cs_pool_uuid, cont->c_uuid), out->coo_lsnapshot);
 	}
 
+	/* Get number of open handles (pre-incremented to reflect the effects of this open) */
+	if (cont_proto_ver >= CONT_PROTO_VER_WITH_NHANDLES) {
+		rc = get_nhandles(tx, cont, NHANDLES_PRE_INCREMENT, &nhandles);
```
liw, line 2051: [Not requesting changes] It looks like we don't need the special "PRE_<OP>" concepts or doing updates in functions named "get_<...>". What would the problem be if we simply write:

    lookup(tx, &nhandles)
    nhandles++ /* increment the variable for preforming the update call */
    update(tx, &nhandles)
    commit(tx)
    if tx cannot commit
        if necessary, do not publish nhandles and decrement the variable if needed

I think we may not even need the last `if` statement for cont_open.

liw (dismissed):

src/container/srv_container.c
kccain: I was trying to consolidate the logic for looking up and updating this property across multiple calling locations, rather than copy/paste the code in each location. Agree that the word "get_" in the function can be misleading the way it has been implemented here.

kccain:
> Test stage Functional Hardware Medium completed with status FAILURE. https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-10793/7/execution/node/1084/log

That is an instance of intermittent failure https://daosio.atlassian.net/browse/DAOS-12054

Running Functional HW Medium tests again (another build 8 triggered).

jolivier23 (commented):
SCons change looks fine to me.   Though I think we may have already bumped it for 2.4 so it really only affects internal usage.

kccain:
Addressed file conflict due to the landing of PR https://github.com/daos-stack/daos/pull/10746, and have pushed for CI testing.
Adding @kjacque to the review as well.

src/include/daos_types.h
kjacque: I do find it a bit odd to break the API on a minor API version change. I don't feel strongly enough to block on it, but I do hope it doesn't come back to bite us.

kjacque (approved):

mchaarawi (dismissed):
lgtm. although i prefer you get a +1 from @liw for the server side changes before you request landing as i'm not too familiar with that part.

src/include/daos_types.h
kjacque: Oh, if it was added recently, e.g. after branching 2.2, that's fine. I thought it was already in a release, which could mean users expect it and we shouldn't remove it. But if it's only in master, in active development, I think we don't have to worry about it.

liw (approved):

src/container/srv_container.c
```diff
@@ -1971,28 +1954,27 @@ cont_open(struct rdb_tx *tx, struct ds_pool_hdl *pool_hdl, struct cont *cont,
...
+		out_v7->coo_md_otime = mdtimes.otime;
+		out_v7->coo_md_mtime = mdtimes.mtime;
+	}
 
-			out_v7->coo_md_otime = mdtimes.otime;
-			out_v7->coo_md_mtime = mdtimes.mtime;
-		} else {	/* CONT_OPEN_BYLABEL */
-			struct cont_open_bylabel_v7_out *out_v7 = crt_reply_get(rpc);
+	if (mdtimes_in_reply && (opc_get(rpc->cr_opc) == CONT_OPEN_BYLABEL)) {
+		struct cont_open_v7_out *out_v7 = crt_reply_get(rpc);
```
kccain, line 1974: The type of out_v7 pointer variable declared here for CONT_OPEN_BYLABEL should have been struct cont_open_bylabel_v7_out *

This was the cause of the single test failure in build 9 - thought I had tested locally before pushing, but was mistaken. Fixing in a next/final push.

liw (approved):

kccain:
Build 12 passed all tests that were run. However, there was a site reliability issue on functional hardware large, and medium clusters preventing those tests from running. I added a note about it in https://daosio.atlassian.net/browse/SRE-372.

Will try to retrigger functional hardware testing.

kccain:
All tests passed, calling for final reviews

liw (dismissed):

mjmac (commented):
Go changes look fine to me. Did this ever get a test run with `Features: container` ? If not, my only concern would be that there might be some daily/weekly tests that are broken by the new container query output. I would hope not, but you know how that goes...

kccain:
> Go changes look fine to me. Did this ever get a test run with `Features: container` ? If not, my only concern would be that there might be some daily/weekly tests that are broken by the new container query output. I would hope not, but you know how that goes...

it has not. I'm making an attempt to do so in a separate PR with identical code content (so as to not disturb the review state of this PR, i.e., not wanting to have to keep asking the reviewers to repeatedly visit PR and give +1).
https://github.com/daos-stack/daos/pull/10888
that PR for some reason didn't select any tests when run with Quick-Functional: true and Features: container so I have some debugging to do to see why that happened.

kccain:
Oh, the branch now has conflicts so maybe it doesn't matter. Sigh. Time to push and test again.

kccain:
PR passed all (expanded, with Features: container) tests after merging latest master to address file conflicts.

As discussed with @mjmac , adding daos-gatekeeper now. This PR has been reviewed numerous times and been given +1 by 2 or more developers, with the most recent changes to the patch being only for merging master & fixing file conflicts, and running expanded testing.

liw (approved):

PR #10821, opened 2022-11-10
DAOS-12059 gurt: fix potential double free

wangshilong:
in d_hhash_create()/d_hash_table_create(), if
d_hash_table_create_inplace() failed, @htable will be freed, but its pointer will be still returned.

This is not safe, and might cause double free if caller cleanup to free memory by checking if pointer is NULL or not.

Required-githooks: true

### DIFF
--- a/src/gurt/hash.c
+++ b/src/gurt/hash.c
@@ -875,8 +875,10 @@ d_hash_table_create(uint32_t feats, uint32_t bits, void *priv,
 		D_GOTO(out, rc = -DER_NOMEM);
 
 	rc = d_hash_table_create_inplace(feats, bits, priv, hops, htable);
-	if (rc)
+	if (rc) {
 		D_FREE(htable);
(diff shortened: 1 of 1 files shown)


src/gurt/hash.c
```diff
@@ -875,8 +875,10 @@ d_hash_table_create(uint32_t feats, uint32_t bits, void *priv,
 		D_GOTO(out, rc = -DER_NOMEM);
 
 	rc = d_hash_table_create_inplace(feats, bits, priv, hops, htable);
-	if (rc)
+	if (rc) {
 		D_FREE(htable);
```
liuxuezhao, line 879: D_FREE internally do the "(ptr) = NULL;", so probably need not explicitly zero it again.

wangshilong: You are right..probably there are other memory corruptions cause my problem. thanks.

PR #10825, opened 2022-11-10
DAOS-10752 control: System Property to Pool Property

ryon-jensen:
This change adds the ability for the system wide mgmt service properties to update pool properties and to be used on pool creation. This way existing background tasks that use pool properties can continue to use them, but the user can experience system wide properties. Additional work still needs to be done to remove the ability to set the pool properties at the pool level with dmg.

This approach also makes it possible to have a
"system pool property" and a "pool override property" so a user can set at system, then override at the
pool level.

Two new system properties were added for checksum
scrubbing mode and threshold.

Required-githooks: true

Signed-off-by: Ryon Jensen <ryon.jensen@intel.com>

### Gatekeeper:


src/control/server/mgmt_system.go
```diff
@@ -1138,7 +1145,56 @@ func (svc *mgmtSvc) SystemSetProp(ctx context.Context, req *mgmtpb.SystemSetProp
...
+		pspr.Id = ps.PoolUUID.String()
+		pspr.SvcRanks = ranklist.RanksToUint32(ps.Replicas)
+		dResp, err := svc.makePoolServiceCall(ctx, drpc.MethodPoolSetProp, pspr)
+		if err != nil {
+			return nil, err
+		}
+
+		if err = proto.Unmarshal(dResp.Body, resp); err != nil {
+			return nil, errors.Wrap(err, "unmarshal PoolSetProp response")
+		}
```
tanabarr, line 1194: Should probably check resp.Status indicates success here.

tanabarr (changes requested):

src/control/server/mgmt_pool.go
```diff
@@ -437,6 +441,57 @@ func (svc *mgmtSvc) PoolCreate(ctx context.Context, req *mgmtpb.PoolCreateReq) (
 	return resp, nil
 }
 
+func (svc *mgmtSvc) PoolCreateAddSystemProps(req *mgmtpb.PoolCreateReq) error {
```
kjacque, line 444: Could make this non-exported (lowercase first letter) if it's only used here.

src/control/server/mgmt_system.go
```diff
@@ -1138,7 +1145,56 @@ func (svc *mgmtSvc) SystemSetProp(ctx context.Context, req *mgmtpb.SystemSetProp
...
+		} else {
+			pspr.Properties[i].SetValueString(p.Value.String())
+		}
+	}
+
+	pools, err := svc.sysdb.PoolServiceList(false)
+	if err != nil {
+		return nil, err
+	}
+	for _, ps := range pools {
```
kjacque, line 1184: Could the `PoolSetSystemProp` function below share some of this logic? Seems like they're basically doing the same thing.

kjacque (commented):
Basically LGTM aside from Tom's comment. I just have a couple suggestions.

src/control/server/mgmt_system.go
ryon-jensen: Actually PoolSetSystemProp isn't used anymore. Deleted that function.

src/control/server/mgmt_system.go
```diff
@@ -1192,6 +1192,9 @@ func (svc *mgmtSvc) SystemSetProp(ctx context.Context, req *mgmtpb.SystemSetProp
 		if err = proto.Unmarshal(dResp.Body, resp); err != nil {
 			return nil, errors.Wrap(err, "unmarshal PoolSetProp response")
 		}
+		if resp.Status != 0 {
+			return nil, errors.Errorf("SystemSetProp: %d\n", resp.Status)
```
kjacque, line 1196: suggestion - You can translate the status from DAOS into a useful string using the `Status` type defined in src/control/lib/daos. Something like this:

```suggestion
			return nil, errors.Wrapf(daos.Status(resp.Status), "setting system prop on pool %s", pspr.Id)
```

kjacque (dismissed):
OK with me. No changes needed but I have a small suggestion.

tanabarr (dismissed):

src/control/server/mgmt_system.go
```diff
@@ -1138,7 +1145,59 @@ func (svc *mgmtSvc) SystemSetProp(ctx context.Context, req *mgmtpb.SystemSetProp
 		return nil, err
 	}
 
-	return &mgmtpb.DaosResp{}, nil
+	// for any system property that has a pool property, update all pools
```
mjmac, line 1148: This is a pretty large increase to the `SystemSetProp()` RPC handler with logic that is ancillary to its primary purpose. It would be better to break it out into a helper function that can be unit tested in isolation, I think.

It's not a rule, exactly, more of a goal, but in general we are trying to keep these RPC handlers relatively thin in order to encourage better separation of concerns and easier unit testing.

src/control/server/mgmt_system.go
```diff
@@ -1138,7 +1145,59 @@ func (svc *mgmtSvc) SystemSetProp(ctx context.Context, req *mgmtpb.SystemSetProp
...
+	var poolSysProps []*daos.PoolProperty
+	for k := range req.GetProperties() {
+		p, ok := svc.systemProps.Get(k)
```
mjmac, line 1151: I don't know if this came in from my example (which may have been incomplete), or there was some other misunderstanding, but it doesn't appear that the system-level values are being used when setting the pool-level values.

The `svc.systemProps` field is initialized as a map of prop key strings to the corresponding property objects, but it is not the actual store for the property values. I understand why this may be confusing. We do it this way to decouple the definition of the system property map from the storage of the map in the system db, in order to keep the storage layer more generic and more easily testable.

In any case, unless I'm missing something, you're going to want to get the value here in addition to the key and use it to set the value of the `PoolProperty` here so that the new value is persisted on the pool(s).

mjmac (commented):
I think this is close, but not quite where it needs to be yet. Are there plans to add unit and system-level test coverage?

src/control/server/mgmt_system.go
ryon-jensen: It does seem to be working the way *I* expected it to. :-D I've manually tested by changing the pool_scrub_mode system property between off and lazy and each time the pool property is updated appropriately. Would system.SetUserProperties update the svc.systemProps map?

src/control/server/mgmt_system.go
ryon-jensen: Refactored ... let me know what you think.

src/control/lib/daos/system_prop.go
```diff
@@ -402,8 +408,68 @@ func (spm SystemPropertyMap) UpdateCompPropVal(key SystemPropertyKey, sourceFn f
 	return errors.Errorf("system property %q does not exist", key)
 }
 
+type poolPropValue struct {
```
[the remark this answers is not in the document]
ryon-jensen: Updated, thanks!

src/control/server/mgmt_pool.go
```diff
@@ -437,6 +441,57 @@ func (svc *mgmtSvc) PoolCreate(ctx context.Context, req *mgmtpb.PoolCreateReq) (
...
 
+func (svc *mgmtSvc) poolCreateAddSystemProps(req *mgmtpb.PoolCreateReq) error {
+	poolSysProps := make(map[uint32]*daos.PoolProperty)
+	for sp := range svc.systemProps.Iter() {
+		pp, found := sp2pp(sp)
+		if !found {
+			continue
+		}
+
+		svc.log.Debugf("Converting System Property '%+v' to Pool Property '%+v'", sp, pp)
```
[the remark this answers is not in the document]
ryon-jensen: Fixed

src/control/server/mgmt_system.go
kjacque: Spent some time going through the code to understand this. It doesn't seem like it should be updated, since the prop it's manipulating should be a copy. Would like to know at least why it seems to be working, if it's an unexpected side-effect from somewhere else in the code.

mjmac (changes requested):

mjmac (approved):

kjacque (approved):

ryon-jensen:
Ready for @daos-stack/daos-gatekeeper

PR #10831, opened 2022-11-11
DAOS-623 build: Improve handling of go code in scons.

ashleypittman:
Move the logic for setting up the go compiler into it's own module.
Disable always-rebuild option for go code to decrease build times.
Add a Scanner for automatically detecting dependencies in go code.
Register dependencies properly so they don't trigger rebuilds.
Mask out build-id from checking.
Change from using InstallAs to Install as the filename isn't changing.

This means that the go build does not called on every scons invocation
and on a test system brings the time of a no-changes build down from
20 seconds to 8 seconds. 


src/control/SConscript
```diff
@@ -59,160 +28,109 @@ def is_release_build(benv):
...
-    # Repository path shared by DAOS Go package import paths
-    repopath = join("github.com", "daos-stack", "daos")
 
-    mod_src = join(gosrc, "cmd", name, "main.go")  # Module src
-    install_src = get_install_src_dir(repopath, name)
-    install_bin = get_install_bin_path(denv, install_name)
-    build_bin = join('$BUILD_DIR', 'src/control/bin', name)
-    src = [mod_src]
+    gosrc = Dir('.').srcnode().abspath
+    sources = Glob(f'cmd/{name}/*.go')
```
jolivier23, line 71: Glob? I thought we discouraged this

site_scons/go_builder.py
```diff
@@ -0,0 +1,87 @@
...
+    conf = Configure(env, custom_tests={'CheckGoVersion': _check_go_version})
+    if not conf.CheckGoVersion():
+        print('no usable Go compiler found (yum install golang?)')
+        Exit(1)
+    conf.Finish()
+    env.d_go_bin = go_bin
+    env.Append(SCANNERS=Scanner(function=_scan_go_file, skeys=['.go']))
+
+
+def setup(env):
```
jolivier23, line 85: Only question is whether this should be done via a tool like the preprocess code in site_tools/extra (imported in SConstruct with this

325     env = Environment(TOOLS=['extra', 'default', 'textfile'])

jolivier23 (approved):

src/control/SConscript
ashleypittman: I think for c files we should be discouraging it, if you add a c file you should add it to the scons file or it won't get compiled.  Here we're adding dependencies so we need to use glob though as if you added a .go file and didn't update the scons then it wouldn't trigger rebuild even if the first build worked.

site_scons/go_builder.py
ashleypittman: Oh yeah, maybe.  Let me take a look at that.

ashleypittman: Done, good call.

tanabarr (approved):
LGTM as a positive cleanup but my review should be taken with a pinch of salt as my familiarity with the scons API is limited.

site_scons/site_tools/go_builder.py
```diff
@@ -78,6 +79,12 @@ def _check_go_version(context):
         print('no usable Go compiler found (yum install golang?)')
         Exit(1)
     conf.Finish()
+
+    # Must be run from the top of the source dir in order to pick up the vendored modules.
+    # Propagate useful GO environment variables from the caller
+    if 'GOCACHE' in os.environ:
+        env['ENV']['GOCACHE'] = os.environ['GOCACHE']
+
     env.Append(SCANNERS=Scanner(function=_scan_go_file, skeys=['.go']))
```
jolivier23, line 88: Does this code get used without adding it to tool import?

jolivier23 (approved):

site_scons/site_tools/go_builder.py
jolivier23: Nevermind, I was looking at the diff since last time

tanabarr (approved):

jolivier23 (approved):

PR #10850, opened 2022-11-16
DAOS-11605 common: fill gap between admem and umem

liuxuezhao:
1. DAOS_MD_ON_SSD ENV set as 0 for PMEM, 1 for BMEM, 2 for ADMEM.
2. Hook with umempobj_create/_open/_close/_get_heapusage etc.
3. Add ad_root()/add_base() API to support umempobj_get_rootptr()/query base,
    reserve another 32KB in arena 0, and limit root obj size as 32KB.
4. create blob file and fallocate space
5. Fix ad mem internal tx handling, and several other bug fixing

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/bio/bio_xstream.c
+++ b/src/bio/bio_xstream.c
@@ -226,7 +226,7 @@ bio_nvme_init(const char *nvme_conf, int numa_node, unsigned int mem_size,
 		bio_spdk_max_unmap_cnt = UINT32_MAX;
 	D_INFO("SPDK batch blob unmap call count is %u\n", bio_spdk_max_unmap_cnt);
 
-	d_getenv_bool("DAOS_MD_ON_SSD_ENABLED", &md_on_ssd_enabled);
+	d_getenv_bool("DAOS_MD_ON_SSD", &md_on_ssd_enabled);
 	D_INFO("MD on SSD is %s\n", md_on_ssd_enabled ? "enabled" : "disabled");
 
 	/* Hugepages disabled */
--- a/src/common/ad_mem.c
+++ b/src/common/ad_mem.c
@@ -535,7 +535,7 @@ blob_close(struct ad_blob *blob)
  * NB: superblock is stored in the first arena.
  */
 int
-ad_blob_create(char *path, unsigned int flags, struct umem_store *store,
+ad_blob_create(const char *path, unsigned int flags, struct umem_store *store,
 	       struct ad_blob_handle *bh)
 {
 	struct ad_blob		*blob;
@@ -625,7 +625,7 @@ failed:
 }
 
 int
-ad_blob_open(char *path, unsigned int flags, struct umem_store *store,
+ad_blob_open(const char *path, unsigned int flags, struct umem_store *store,
 	     struct ad_blob_handle *bh)
 {
 	struct ad_blob		*blob;
@@ -1003,11 +1003,15 @@ arena_reserve(struct ad_blob *blob, unsigned int type, struct umem_store *store_
 	setbits64(ad->ad_bmap, 0, 2);
 
 	D_CASSERT(ARENA_UNIT_SIZE == BLOB_HDR_SIZE);
-	if (id == 0) { /* the first arena */
-		/* Blob header(superblock) is stored in arena zero, it consumes 64K.
+	if (id == 0) {
+		/* Arena 0 reserves 128KB totally -
+		 * Arena header ad_arena_df)			ARENA_HDR_SIZE (64KB)
+		 * Blob header (superblock ad_blob_df)		BLOB_HDR_SIZE (32KB)
+		 * Root obj (export to user by ad_root())	AD_ROOT_OBJ_SIZE (32KB)
 		 * NB: the first arena is written straightway, no WAL
 		 */
-		setbit64(ad->ad_bmap, 2);
+		setbit64(ad->ad_bmap, 2); /* for blob header */
+		setbit64(ad->ad_bmap, 3); /* for root obj */
 	}
 	/* DRAM only operation, mark the arena as reserved */
 	D_ASSERT(!isset64(blob->bb_bmap_rsv, id));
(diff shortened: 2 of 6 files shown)


src/common/mem.c
```diff
@@ -34,7 +34,13 @@ struct umem_tx_stage_item {
...
 #ifdef DAOS_PMEM_BUILD
 
-static int use_bmem = 1; /* Use BMEM by default */
+enum {
+	DAOS_MD_PMEM	= 0,
+	DAOS_MD_BMEM	= 1,
+	DAOS_MD_ADMEM	= 2,
+};
+
+static int daos_md_backend = DAOS_MD_BMEM; /* Use BMEM by default */
```
NiuYawei, line 43: This comment looks confusing, DAOS_MD_ON_SSD isn't set by default, so by default, the daos_md_backend will actually be set to DAOS_MD_PMEM in umempobj_setting_init().

If we want to use PMEM as default value, we need to fix this comment; If we want to use BMEM as default value, the umempobj_setting_init() should be fixed.

src/common/mem.c
```diff
@@ -109,25 +130,36 @@ umempobj_create(const char *path, const char *layout_name, int flags,
...
-	if (!dav_hdl) {
-		D_ERROR("Failed to create pool %s, size="DF_U64": errno = %d\n",
-			path, poolsize, errno);
-		D_FREE(umm_pool);
-		return NULL;
-	}
-	if (store != NULL)
-		umm_pool->up_store = *store;
-	umm_pool->up_priv = dav_hdl;
+		/* TODO: Do checkpoint here to write back allocator heap */
```
NiuYawei, line 145: I think this is for both BMEM and AD? Or ad_blob_create() will internally flush heap data to SSD (and doesn't need to rely on WAL)?

src/common/mem.c
```diff
@@ -173,28 +209,39 @@ umempobj_open(const char *path, const char *layout_name, int flags, struct umem_
...
 
-		if (store != NULL)
-			umm_pool->up_store = *store;
 		umm_pool->up_priv = pop;
 		return umm_pool;
-	}
+	case DAOS_MD_BMEM:
+		/* TODO mmap tmpfs file */
+		/* TODO Load all meta pages from SSD */
+		/* TODO Replay WAL */
```
NiuYawei, line 217: My original idea is that umempobj_open() calls umem common function to do mmap first, then load meta blob and replay WAL, the last step is to call dav_obj_open() or ad_blob_open() according to the daos_md_backend.

src/common/mem.c
```diff
@@ -331,6 +425,13 @@ umempobj_set_slab_desc(struct umem_pool *ph_p, struct umem_slab_desc *slab)
 		rc = dav_class_register((dav_obj_t *)ph_p->up_priv, &davslab);
 		/* update with the new slab id */
 		slab->class_id = davslab.class_id;
+		break;
+	case DAOS_MD_ADMEM:
+		/* NOOP for ADMEM now */
```
NiuYawei, line 430: You need to fake a slab ID for each registered slabs, it won't work otherwise.

src/common/mem.c
```diff
@@ -1149,10 +1254,19 @@ set_offsets(struct umem_instance *umm)
...
 
-		umm->umm_base = (uint64_t)dav_get_base_ptr(pop);
+		umm->umm_base = (uint64_t)dav_get_base_ptr(dav_pop);
+		break;
+	case UMEM_CLASS_ADMEM:
+		bh.bh_blob = (struct ad_blob *)umm->umm_pool->up_priv;
+		umm->umm_base = (uint64_t)ad_get_base_ptr(bh);
+		break;
+	default:
+		D_ASSERTF(0, "bad umm->umm_id %d\n", umm->umm_id);
```
NiuYawei, line 1268: so mmap + set base will be implemented in follow-on patches?

src/bio/bio_xstream.c
```diff
@@ -227,7 +227,7 @@ bio_nvme_init(const char *nvme_conf, int numa_node, unsigned int mem_size,
 		bio_spdk_max_unmap_cnt = UINT32_MAX;
 	D_INFO("SPDK batch blob unmap call count is %u\n", bio_spdk_max_unmap_cnt);
 
-	d_getenv_bool("DAOS_MD_ON_SSD_ENABLED", &md_on_ssd_enabled);
+	d_getenv_bool("DAOS_MD_ON_SSD", &md_on_ssd_enabled);
```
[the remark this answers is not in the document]
liuxuezhao: here as md_on_ssd_enabled is a bool, seems d_getenv_bool() can work. In mem.c need to use d_getenv_int().

src/common/mem.c
liuxuezhao: agreed, I feel confuse too when I read the code, in this PR it actually just kept same behavior with original code.
let's change umempobj_settings_init() to make it match with the comment, but not sure if it will cause trouble in CI. will see...

src/common/mem.c
liuxuezhao: I don't quite understand the original comment for BMEM here.
For AD, it internally call store->stor_ops->so_write() to write the ARENA HDR + BLOB HDR in ad_blob_create().

src/common/mem.c
liuxuezhao: make sense to me, now BMEM do the open+mmap internally, we can refine it at following patch.

src/common/mem.c
liuxuezhao: right, looks like the caller of umempobj_set_slab_desc() already set slab->class_id (see vos_register_slabs), let's see if will hit problem in following test.

src/common/mem.c
liuxuezhao: right, I'll add a few code later. thx

src/common/mem.c
liuxuezhao: hmm, changing default to BMEM and hit a NLT failure (not very sure if it is related although), I'll change back to PMEM by default and remove the comment. To avoid CI failure.

src/bio/bio_xstream.c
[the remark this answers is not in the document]
wangshilong: btw I was confused with API, when i tried to set DAOS_MD_ON_SSD=true, it did not work, instead we need DAOS_MD_ON_SSD=1

wangshilong (dismissed):

NiuYawei (approved):

src/common/ad_mem.c
```diff
@@ -2790,9 +2805,10 @@ ad_root(struct ad_blob_handle bh, size_t size)
 
 /** Query base pointer */
 void *
-ad_get_base_ptr(struct ad_blob_handle bh)
+ad_base(struct ad_blob_handle bh)
 {
 	struct ad_blob	*blob = bh.bh_blob;
 
-	return ad_addr2ptr(bh, blob_addr(blob));
+	D_ASSERT((uintptr_t)ad_addr2ptr(bh, blob_addr(blob) == (uintptr_t)blob->bb_mmap));
```
liuxuezhao, line 2812: sorry here with a bug, should be "D_ASSERT((uintptr_t)ad_addr2ptr(bh, blob_addr(blob)) == (uintptr_t)blob->bb_mmap);" will fix it when refresh it next time.

gnailzenh (approved):

PR #10859, opened 2022-11-17
DAOS-12107 control: Verbose output for daos container list

0xE0F:
Adds verbose output as follow:

```
$:~/src/daos/install/bin> sudo ./daos container list -V denisb
UUID                                 Label   Type    ObjClass RF
----                                 -----   ----    -------- --
c29a1242-c77a-486d-904f-1a6b321e2e0a nvme    POSIX   SX       0
33dc4dd4-259c-411b-a93b-ba83e18d642c nfs     POSIX   EC_4P2G2 0
ee224070-5589-4aa9-9847-5bd2dc8a99e9 scratch unknown          0
```

Signed-off-by: Denis Barakhtanov <denis.barahtanov@croit.io>

### Gatekeeper:

### DIFF
--- a/src/control/cmd/daos/container.go
+++ b/src/control/cmd/daos/container.go
@@ -527,31 +527,7 @@ func listContainers(hdl C.daos_handle_t) ([]*ContainerID, error) {
 	return out, nil
 }
 
-func printContainers(out io.Writer, contIDs []*ContainerID) {
-	if len(contIDs) == 0 {
-		fmt.Fprintf(out, "No containers.\n")
-		return
-	}
-
-	uuidTitle := "UUID"
-	labelTitle := "Label"
-	titles := []string{uuidTitle, labelTitle}
-
-	table := []txtfmt.TableRow{}
-	for _, id := range contIDs {
-		table = append(table,
-			txtfmt.TableRow{
-				uuidTitle:  id.UUID.String(),
-				labelTitle: id.Label,
-			})
-	}
-
-	tf := txtfmt.NewTableFormatter(titles...)
-	tf.InitWriter(out)
-	tf.Format(table)
-}
-
-func printContainersVerbose(out io.Writer, contsInfo []*containerInfo) {
+func printContainers(out io.Writer, contsInfo []*containerInfo, verbose bool) {
 	if len(contsInfo) == 0 {
 		fmt.Fprintf(out, "No containers.\n")
 		return
@@ -561,23 +537,28 @@ func printContainersVerbose(out io.Writer, contsInfo []*containerInfo) {
 	labelTitle := "Label"
 	typeTitle := "Type"
 	oclassTitle := "ObjClass"
+	dclassTitle := "DirClass"
+	fclassTitle := "FileClass"
 	rfTitle := "RF"
-	titles := []string{uuidTitle, labelTitle, typeTitle, oclassTitle, rfTitle}
+	titles := []string{uuidTitle, labelTitle}
+	if verbose {
+		titles = append(titles, typeTitle, oclassTitle, dclassTitle, fclassTitle, rfTitle)
+	}
 
 	table := []txtfmt.TableRow{}
 	for _, info := range contsInfo {
-		rfValue := ""
-		if info.RedundancyFactor != nil {
-			rfValue = fmt.Sprintf("%d", *info.RedundancyFactor)
+		row := txtfmt.TableRow{
+			uuidTitle:  info.ContainerUUID.String(),
+			labelTitle: info.ContainerLabel,
 		}
-		table = append(table,
-			txtfmt.TableRow{
-				uuidTitle:   info.ContainerUUID.String(),
-				labelTitle:  info.ContainerLabel,
-				typeTitle:   info.Type,
-				oclassTitle: info.ObjectClass,
-				rfTitle:     rfValue,
-			})
+		if verbose {
+			row[typeTitle] = info.Type
+			row[oclassTitle] = info.ObjectClass
+			row[dclassTitle] = info.DirObjectClass
+			row[fclassTitle] = info.FileObjectClass
+			row[rfTitle] = fmt.Sprintf("%d", info.RedundancyFactor)
+		}
+		table = append(table, row)
 	}
 
 	tf := txtfmt.NewTableFormatter(titles...)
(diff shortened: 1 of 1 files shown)


mjmac:
Created #10905 to test from an internal branch.

mjmac:
> Created #10905 to test from an internal branch.

Passed all tests. Pylint failure seems to be some kind of issue on master.

mjmac:
@0xE0F: PR looks good, but I think you'll need to rebase it on the latest master in order to bring in some changes to the container query output. I'll push the updated PR to an internal branch for full testing.

0xE0F:
@mjmac Thank you !
I rebased to the latest master and updated to a bit more concise version.

Maybe it's a bit of a stretch, but would the team consider using verbose output as a default one ?
Or the additional RPCs to get info are the deal breaker ?

Sample of output:
```
denis@dev:~/daos/install/bin> sudo ./daos container list -V denisb
UUID                                 Label    Type    ObjClass DirClass FileClass RF
----                                 -----    ----    -------- -------- --------- --
1daf60cc-1725-4694-9abc-f4f3b4da3c06 scratch2 POSIX   UNKNOWN  S1       S1        0
fb7c9ae6-b088-45d3-94f7-7e48a7746d6f nfs      POSIX   UNKNOWN  SX       SX        0
1c668e65-ab19-498f-83f3-35a879ce09b1 nvme     POSIX   UNKNOWN  SX       SX        0
59dc1615-aa01-43a3-b479-07fbce117fe6 scratch  unknown                             0
094e32de-7739-4789-bc19-bb518070e636 scratch3 POSIX   UNKNOWN  SX       S1        0
```

mchaarawi (changes requested):
I have some concerns with this PR.
some of the information printed in verbose mode is specific to only POSIX containers. I don't think this should be printed with the list containers command. 
I'm talking about all the object classes which only apply to POSIX container, are not container properties but more like just data from the generic container list perspective.
so my suggestion is for the container list in verbose mode to print only container properties (RF, Rlevel, etc.) but not include type specific data for containers like the object class for posix containers.
This also appears to cause overhead by mounting the container again from the go code on daos cont create since query is run from there. this can all be avoided if you exclude that type specific data.

src/control/cmd/daos/container.go
```diff
@@ -562,14 +589,33 @@ func (cmd *containerListCmd) Execute(_ []string) error {
 			"unable to list containers for pool %s", cmd.PoolID())
 	}
 
+	contInfo := make([]*containerInfo, 0, len(contIDs))
+	for _, contID := range contIDs {
+		if cmd.Verbose {
+			info, err := cmd.containerInfo(&contID.UUID)
```
mchaarawi, line 595: In a production setting where a pool will have containers owned by different users, isn't this going to fail everytime with ENOPERM for containers they don't have access to?
so i question the usefulness in this case.

0xE0F:
> 

yeah, I do agree that printing attributes that are not part of the container type not the best idea.
The whole point of this PR is to make a container management a bit more convenient to use in case when there are dozens of container with different types and options.
And it's a bit hard to find that specific container with posix type, rf=1 and oclass=S1 when all you have in the output is label and UUID.
This leads to give a container labels like `smb_rf1_S1`, `scratch_rf0_SX` etc. 
Or list attributes for each container until one is found.
Same goes for the shelling out and use `daos` cmd in the pipeline with other tools.

What about grouping container by type? Maybe something like this:

```
denis@dev:~/daos/install/bin> sudo ./daos container list -V denisb
POSIX containers
----------------

UUID                                 Label    ObjClass DirClass FileClass RF
----                                 -----    -------- -------- --------- --
1daf60cc-1725-4694-9abc-f4f3b4da3c06 scratch2 UNKNOWN  S1       S1        0
fb7c9ae6-b088-45d3-94f7-7e48a7746d6f nfs      UNKNOWN  SX       SX        0
1c668e65-ab19-498f-83f3-35a879ce09b1 nvme     UNKNOWN  SX       SX        0
094e32de-7739-4789-bc19-bb518070e636 scratch3 UNKNOWN  SX       S1        0


OTHER containers
----------------

UUID                                 Label    RF
----                                 -----    --
59dc1615-aa01-43a3-b479-07fbce117fe6 scratch  1
```

src/control/cmd/daos/container.go
0xE0F: fair enough, I'm not using multiple owners at the moment.
On the other hand, the ones without access can be omitted, or perhaps reading container attributes allowed for anyone ?

mchaarawi: > On the other hand, the ones without access can be omitted

But that is not really a container list anymore (and not even a verbose list), is it? it would very weird that list-containers return one thing, and list-containers --verbose returns a subset. sounds like you need a completely different option to return containers that you are an owner (or have read permissions on) and return those.  

> perhaps reading container attributes allowed for anyone

the problem is that attributes you are reading here for POSIX containers are not really attributes from the container perspective, but just regular data (POSIX containers store them in the SB object which is just a regular object from container permissions sake) so it involves a regular object fetch that you should have read permissions for.

mjmac: @kccain pointed out yesterday that there is a new-ish container API that might be useful here: https://github.com/daos-stack/daos/blob/11e2bddb4575df47092a91ec9cc23973a0b5d067/src/include/daos_pool.h#L611

In theory, that should allow us to construct a filter to only show containers meeting certain criteria. I'm not sure if there is a way to express read permission as a filter criterion, though.

kccain: Not related to container owners -  but applicable to outputting some common properties: This new interface combines list containers behavior with the basic part of a container query (the daos_cont_info_t information).  However it does not return properties. I tried to leave a placeholder in one of the data structures in case we might decide to try to add properties in later, but didn't think too exactly about how it would be implemented. Except to think that perhaps a more sophisticated implementation of the cart bulk transfer could be done -  i.e., list_cont_bulk_create() that in this implementation currently refers to a single linear buffer of size ncont * sizeof(struct daos_pool_cont_info2). Maybe the pci_reserved[] pointers in that structure could have some additional dynamically allocated storage to accommodate properties coming back from the engine in the bulk transfer.

0xE0F: Looks like permission issue might get on the way with the verbose output.
I do agree with @mchaarawi that skipping containers without proper access does not express `list` command semantic at all.

I guess tooling have to deal with containers one by one 😭 
Maybe one day with another interface it would be straightforward to implement 😬

PR #10873, opened 2022-11-18
DAOS-12119 control: Handle -DER_NONEXIST on retried PoolDestroy

mjmac:
In the event that a PoolDestroy RPC is retried and the pool is
destroyed while the client is waiting to retry, the server will
return a -DER_NONEXIST in response to the final retry. As this
is not really an error, just catch it and signal success.

Test-repeat-vm: 10

Required-githooks: true


kjacque (approved):

tanabarr (approved):
looks like a nice clean solution

PR #10890, opened 2022-11-22
DAOS-12013 test: Update telemetry test for I/O DTX committed values

sydneyvanda:
Change object class from SX to RP_2GX, increase test case to 2 server ranks.
Remove TODO for verifying I/O DTX committable metrics (no longer needed).

Quick-Functional: true
Skip-func-hw-test-large: true
Test-tag: test_io_latency_telemetry test_ior_latency_telemetry test_ior_dtx_telemetry offline_extend_oclass

Signed-off-by: Sydney Vanda <sydney.m.vanda@intel.com>

### Gatekeeper:


src/tests/ftest/control/dmg_telemetry_io_latency.py
```diff
@@ -315,13 +315,15 @@ def verify_io_dtx_committed_metrics(self, baseline_data, metrics_data, test_metr
         # test ops = (create dir + punch obj + punch dir) * repetitions
         test_ops *= repetitions
         num_operations += test_ops
+        # 2 DTX entries on 2 replicas
+        num_operations *= 2
```
kjacque, line 319: Instead of hard-coding the number of replicas, could we import the number of servers from the yaml file?

Nasf-Fan:
I am thinking that whether it is worth for us to spend time on maintaining the test to verify the count of committed DTX entries. This time, it is because we changed DTX commit logic that does not generate committed DTX entry for single replicated object, then this test got failure. But in the future, we may change related DTX logic again, that may break the test at other point.

On the other hand, even if there is potential bug that may cause committed DTX entry is not generated, it also needs some completed IO mode/race, but current test IO mode is too simple, that almost cannot trigger related corner cases; otherwise, our other tests (with similar IO mode) have already hit test failures.

In the real world, a DTX entry can be removed by DTX aggregation at any time. The count of committed DTX entries cannot stand for related transactions count. It is not a hard standard for the user/application except DTX aggregation. But DTX aggregation is not the verified purpose of the test.

Sorry, my suggestion maybe some late. But it is better than still waste time on the maintaining...

sydneyvanda:
> I am thinking that whether it is worth for us to spend time on maintaining the test to verify the count of committed DTX entries. This time, it is because we changed DTX commit logic that does not generate committed DTX entry for single replicated object, then this test got failure. But in the future, we may change related DTX logic again, that may break the test at other point.
> 
> On the other hand, even if there is potential bug that may cause committed DTX entry is not generated, it also needs some completed IO mode/race, but current test IO mode is too simple, that almost cannot trigger related corner cases; otherwise, our other tests (with similar IO mode) have already hit test failures.
> 
> In the real world, a DTX entry can be removed by DTX aggregation at any time. The count of committed DTX entries cannot stand for related transactions count. It is not a hard standard for the user/application except DTX aggregation. But DTX aggregation is not the verified purpose of the test.
> 
> Sorry, my suggestion is some late. But it may be better than still waste time on the maintaining... please consider.

Understandable, maybe we should just remove test altogether then instead of worrying about constantly maintaining it? I will post your comment in ticket to get others opinion/approval.

src/tests/ftest/control/dmg_telemetry_io_latency.py
```diff
@@ -315,13 +315,16 @@ def verify_io_dtx_committed_metrics(self, baseline_data, metrics_data, test_metr
         # test ops = (create dir + punch obj + punch dir) * repetitions
         test_ops *= repetitions
         num_operations += test_ops
+        # 2 DTX entries on 2 replicas
+        test_servers = self.params.get("test_servers", "/run/*")
+        num_operations *= test_servers
```
Nasf-Fan, line 320: This caused the CI test failure:
05:02:17 ERROR| TypeError: can't multiply sequence by non-int of type 'float'

phender, line 320: The better way to get the number of server hosts is to use the class that manages starting the servers, but I suspect you want engine count here:
```suggestion
        num_operations *= self.server_managers[0].engines
```
Instead, if you do really need the number of hosts:
```
        num_operations *= len(self.server_managers[0].hosts)
```

phender (changes requested):

daltonbohning:
Superseded by #11451

PR #10901, opened 2022-11-24
DAOS-11048 bio: monitor WAL and Meta SSDs

wangshilong:
Currently only data SSDs monitoring is supported, this patch tries
enable health collection for WAL and Meta SSDs as well.

Auto fault detection and state transit for WAL/Meta SSDs are not
supported yet.

Required-githooks: true

### DIFF
--- a/src/bio/bio_monitor.c
+++ b/src/bio/bio_monitor.c
@@ -202,12 +202,13 @@ bio_dev_set_faulty(struct bio_xs_context *xs, enum smd_dev_type st)
 }
 
 static inline struct bio_dev_health *
-xs_ctxt2dev_health(struct bio_xs_context *ctxt)
+cb_arg2dev_health(void *cb_arg)
 {
-	D_ASSERT(ctxt != NULL);
+
+	struct bio_xs_blobstore **bxb_ptr = (struct bio_xs_blobstore **)cb_arg;
 	struct bio_xs_blobstore *bxb;
 
-	bxb = bio_xs_context2xs_blobstore(ctxt, SMD_DEV_TYPE_DATA);
+	bxb = *bxb_ptr;
 	/* bio_xsctxt_free() is underway */
 	if (bxb == NULL)
 		return NULL;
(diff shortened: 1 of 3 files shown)


src/bio/bio_monitor.c
```diff
@@ -811,8 +837,16 @@ bio_bs_monitor(struct bio_xs_context *xs_ctxt, uint64_t now)
 		D_ERROR("State transition on target %d failed. %d\n",
 			bbs->bb_owner_xs->bxc_tgt_id, rc);
 
-	if (!bypass_health_collect())
-		collect_raw_health_data(xs_ctxt);
+	if (!bypass_health_collect()) {
+		D_ALLOC_PTR(gsca);
```
NiuYawei, line 841: Instead of allocating and freeing a 'gsca', can't we just pass '&xs_ctxt->bxc_xs_blobstore[st]" as argument for completion callback?

src/bio/bio_monitor.c
```diff
@@ -201,13 +201,20 @@ bio_dev_set_faulty(struct bio_xs_context *xs, enum smd_dev_type st)
...
+cb_arg2dev_health(struct get_spdk_cb_arg *gsca)
 {
-	D_ASSERT(ctxt != NULL);
+
+	struct bio_xs_context	*ctxt = gsca->ctxt;
 	struct bio_xs_blobstore *bxb;
 
-	bxb = bio_xs_context2xs_blobstore(ctxt, SMD_DEV_TYPE_DATA);
+	D_ASSERT(ctxt != NULL);
+	bxb = bio_xs_context2xs_blobstore(ctxt, gsca->st);
```
NiuYawei, line 217: bio_xs_context2bx_blobstore() should be used in this case. And I think we pass the address of bxc_xs_blobstore[st] as completion argument, we could just check if (*arg == NULL) here to tell if bio_xsctxt_free() is underway, right?

NiuYawei (changes requested):

src/bio/bio_xstream.c
```diff
@@ -1838,9 +1838,7 @@ bio_nvme_poll(struct bio_xs_context *ctxt)
 		bxb = ctxt->bxc_xs_blobstores[st];
 		if (bxb && bxb->bxb_blobstore &&
 		    is_bbs_owner(ctxt, bxb->bxb_blobstore))
-			bio_bs_monitor(ctxt, now);
-		/* Only support DATA for now */
-		break;
+			bio_bs_monitor(ctxt, st, now);
```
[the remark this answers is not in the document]
wangshilong: You are right, just tried to enable for monitoring metrics works, we disabled fault detection and device state transition for WAL/Meta, need enable them too, also fault detection and device state transition won't work for sys device.

NiuYawei (approved):

jolivier23 (approved):

PR #7678, opened 2021-12-15
DAOS-9163 pool: Do map_dist for restarted ranks

liw:
Request pool map distributions for potentially restarted ranks, so that
any rank who has restarted without triggering a SWIM DEAD event will
eventually receive maps of the pools that this rank belongs to.

Introduce POOL_GROUP_MAP_STATUS to ensure that a consistent set of map
status is used to determine the up ranks in various cases.

Signed-off-by: Li Wei <wei.g.li@intel.com>

PR #10916, opened 2022-11-28
DAOS-10162 control:Support auto reint target by swimming

wangzhaorong-cestc:
Admin should be able to set a DAOS pool property and disable auto-rebuild, 
in this case, even if an auto-evicted target is marked as DOWN to declare death, 
rebuild should not be immediately scheduled for this target because admin might 
fix the problem in minutes, simultaneously, the storage system should continue to
server I/O requests in degraded mode instead of waiting. 

Signed-off-by: wangzhaorong <wangzhaorong@cestc.cn>

### DIFF
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -1099,18 +1099,21 @@ discard_events(d_list_t *queue)
 	}
 }
 
-static int pool_svc_exclude_rank(struct pool_svc *svc, d_rank_t rank);
+static int pool_svc_change_rank(struct pool_svc *svc, d_rank_t rank, crt_opcode_t opc);
 
 static void
 handle_event(struct pool_svc *svc, struct pool_svc_event *event)
 {
 	daos_prop_t		prop = {0};
 	struct daos_prop_entry *entry;
+	struct pool_domain	*dom = NULL;
 	int			rc;
+	uint8_t			co_status;
 
-	/* Only used for exclude the rank for the moment */
+	/* used for exclude or reint the rank*/
 	if ((event->psv_src != CRT_EVS_GRPMOD && event->psv_src != CRT_EVS_SWIM) ||
-	    event->psv_type != CRT_EVT_DEAD || pool_disable_exclude) {
+	    (event->psv_type != CRT_EVT_ALIVE && event->psv_type != CRT_EVT_DEAD) ||
+	    (event->psv_type == CRT_EVT_DEAD && pool_disable_exclude)) {
 		D_DEBUG(DB_MD, "ignore event: "DF_PS_EVENT" exclude=%d\n", DP_PS_EVENT(event),
 			pool_disable_exclude);
 		goto out;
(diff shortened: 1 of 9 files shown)


gnailzenh:
Di, xuezhao, could you please check this patch? thanks

src/pool/srv_pool_map.c
```diff
@@ -116,10 +123,20 @@ update_one_tgt(struct pool_map *map, struct pool_target *target,
 				DP_TARGET(target));
 			return -DER_BUSY;
 		case PO_COMP_ST_DOWNOUT:
+			if (!is_command && target->ta_comp.co_flags & PO_COMPF_COMMAND) {
+				D_DEBUG(DB_MD, DF_TARGET" must be reinted by admin command\n",
+					DP_TARGET(target));
+				return -DER_NO_PERM;
```
wangdi1, line 129: So if the target is excluded by cmd, then swim can not reint the target?  This is a bit strange. IMHO.  I thought DAOS_SELF_HEAL_AUTO_REINT should control this?  i.e. the user should understand the target will be reintegrated once it is startup.

Hmm, if there are duplicate reintegrations due to this, the 2nd can always be ignored?  Or I miss some cases?

wangzhaorong-cestc: We want the disk and the rank  to be marked down by the command must also be marked up by the command.
A rank has multiple disks, when there are bad disks, slow disks and other scenarios, we hope that the target on these faulty disks in the rank does not go up, which can solve this scenario

wangdi1: I see. thanks for clarification. Though I wonder

1. should we add an option in the cmd (dmg pool exclude) to indicate these ranks/targets are not allow to be reintegrated automatically (by swim). Or it should be by default. @liw any thought here?
2. pool map version needs to updated for adding new flag?

src/pool/srv_pool.c
```diff
@@ -5904,21 +5954,21 @@ ds_pool_tgt_exclude_out(uuid_t pool_uuid, struct pool_target_id_list *list)
 {
 	return pool_update_map_internal(pool_uuid, POOL_EXCLUDE_OUT, false,
 					list, NULL, NULL, NULL, NULL, NULL,
-					NULL);
+					NULL, false);
```
wangdi1, line 5957: Hmm, even if the rank/tgt is excluded by dmg, then following rebuild process will call this function to set the target status to "DOWNOUT",  so it should tell if the "exclude out" is from dmg or swim?

src/pool/srv_pool.c
```diff
@@ -6158,25 +6208,38 @@ ds_pool_update_handler(crt_rpc_t *rpc)
...
+		exclude_rank = false;
+		break;
+	case POOL_EXCLUDE:
+		exclude_rank = true;
+		break;
+	default:
+		break;
+	}
+
+	rc = pool_svc_update_map(svc, opc, exclude_rank /* exclude_rank */,
```
wangdi1, line 6237: exclude_rank and opc seems duplicate parameters?

src/pool/srv_pool_map.c
liw: > 1. should we add an option in the cmd (dmg pool exclude) to indicate these ranks/targets are not allow to be reintegrated automatically (by swim). Or it should be by default. @liw any thought here?

Well, I don't have a clear preference, as long as the reintegration process can handle the churn. By the way, thanks for reminding me of this patch; I will revise my #7678 to accommodate the changes here.

src/pool/srv_pool.c
wangzhaorong-cestc: well， exclude_rank here  is to distinguish between swim detection and command. The swim detection will mark rank down, but the command will not mark rank down

src/pool/srv_pool.c
wangzhaorong-cestc: I I don't think it's necessary. Why should we distinguish

src/pool/srv_pool_map.c
wangzhaorong-cestc: > 

Flag setting and cancellation are accompanied by target status changes

wangzhaorong-cestc: Any suggestions?

PR #10923, opened 2022-11-28
DAOS-12119 control: Handle correct error on retried PoolDestroy

mjmac:
The previous commit checked for the wrong error in determining
whether or not a PoolDestroy failure is fatal.

Required-githooks: true

### DIFF
--- a/src/control/system/errors.go
+++ b/src/control/system/errors.go
@@ -7,6 +7,7 @@
 package system
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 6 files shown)


src/control/system/errors.go
```diff
@@ -233,6 +234,37 @@ type ErrPoolNotFound struct {
 	byLabel *string
 }
 
+func (err *ErrPoolNotFound) MarshalJSON() ([]byte, error) {
```
tanabarr, line 237: why is there a need for custom serialisation? rank and uuid have those methods already, maybe I am missing something else?

mjmac: The fields are unexported, intentionally. I decided that this was the least-worst solution to allow us to get the error on the dmg side.

tanabarr: ah, why do they need to not be exported?

mjmac: > ah, why do they need to not be exported?

Because the intent is for the error types to be opaque outside of the package. That decision could be revisited at some point, but I didn't want to get into it for this PR. If we do change it so that the fields are exported, then we could remove the custom marshal/unmarshal methods on the type without any ripple effects.

kjacque (approved):

PR #10930, opened 2022-11-29
DAOS-623 build: Remove hwloc and cmocka requires from build

ashleypittman:


src/control/SConscript
```diff
@@ -108,7 +108,7 @@ def scons():
 
     denv.d_setup_go()
 
-    prereqs.require(denv, 'ofi', 'hwloc')
```
mjmac, line 111: We do use hwloc in the control plane, just FYI. If this works though then I don't have a problem with it.

site_scons/components/__init__.py
```diff
@@ -221,50 +219,34 @@ def define_common(reqs):
...
+    reqs.define('hwloc', libs=['hwloc'], headers=['hwloc.h'], package='hwloc-devel')
+
+    reqs.define('protobufc', libs=['protobuf-c'], headers=['protobuf-c/protobuf-c.h'],
+                package='protobuf-c-devel')
 
 
 def define_ompi(reqs):
     """OMPI and related components"""
-    reqs.define('hwloc', headers=['hwloc.h'], libs=['hwloc'],
-                package='hwloc-devel')
```
jolivier23, line 267: I don't like removing these.   With package= line, it doesn't try to build hwloc or protobufc.  It is there primarily for configure checks.   Now, if you try to build daos without protobufc or hwloc installed, you will get a build failure.   I think it's better to do the configure checks up front for things that are not always automatically installed in the distro.

jolivier23 (changes requested):
I don't have a strong opinion but I prefer to keep the configure checks in place.

daltonbohning (commented):
Could I ask that we maintain some previous behavior of building protobuf?
With the environment we use on Frontera, this fails to build. And historically, build changes are incredibly difficult to resolve

site_scons/components/__init__.py
ashleypittman: I simply moved this so it was in defined in the same place as others, the check should be exactly the same.  If you try to build daos without protobufc or hwloc installed then it'll tell you what package to install (assuming you're on el8)

site_scons/components/__init__.py
```diff
@@ -221,50 +219,34 @@ def define_common(reqs):
...
-        reqs.define('uuid', headers=['uuid/uuid.h'])
-    else:
-        reqs.define('uuid', libs=['uuid'], headers=['uuid/uuid.h'],
-                    package='libuuid-devel')
+    reqs.define('uuid', libs=['uuid'], headers=['uuid/uuid.h'], package='libuuid-devel')
+
+    reqs.define('hwloc', libs=['hwloc'], headers=['hwloc.h'], package='hwloc-devel')
+
+    reqs.define('protobufc', libs=['protobuf-c'], headers=['protobuf-c/protobuf-c.h'],
+                package='protobuf-c-devel')
```
ashleypittman, line 245: This is the definitions for hwloc and protobufc.

I did remove a couple of definitions from this file but they weren't ones that we ever used or checked for.

site_scons/components/__init__.py
daltonbohning: > it'll tell you what package to install

On Frontera, we can't install any packages to system paths because we don't have root permission.
Frontera is technically not a production environment, but if it's not too much maintenance to allow scons to optionally build for us...

ashleypittman:
> Could I ask that we maintain some previous behavior of building protobuf? With the environment we use on Frontera, this fails to build. And historically, build changes are incredibly difficult to resolve

Ok, added it back in although I can tidy it up a bit.

Even with the packages installed our build is still building the dependences, is there any way to check for the installed deps and only build if required @jolivier23?  I can work on adding something if this doesn't exist today.

ashleypittman:
https://github.com/daos-stack/daos/pull/11100 updates the builder to make protobuf optional and adds it to the docker builders so that supersedes this PR, hence making it private.

There are some changes in the build in this PR that remove dead code or unnecessary complexity so I'll keep this open for now and cherry-pick those changes in a different PR.

src/gurt/tests/SConscript
```diff
@@ -14,27 +14,11 @@ def scons():
...
-        print("***************************************************\n")
-        # Just returning is the desired behavior here.  The utest target
-        # has yet to be defined so the build will just fail.
-        return
+    Import('env', 'gurt_targets')
 
     test_env = env.Clone()
-    prereqs.require(test_env, "mercury", "uuid", "cmocka")
-    test_env.AppendUnique(LIBS=['pthread', 'm'])
+    test_env.require('mercury', 'uuid')
```
jolivier23, line 20: why is there no require for cmocka here?

ashleypittman: There is no provision for building cmocka from source so no munging of include or lib dirs are needed.

jolivier23 (dismissed):
I guess it gets required initially so subsequent require lines are not really necessary since it's not a component we build?

ashleypittman:
> I guess it gets required initially so subsequent require lines are not really necessary since it's not a component we build?

I'm not sure.  I don't think it did get required initially, it must just be added to 'libs' where it's needed.

ashleypittman:
Actually cmocks is just an include right so it needs to be on the include path and as we don't build from source it's always in /usr/include?

daltonbohning (commented):
I'd like to build with this on Frontera, but
1. Frontera queues are down right now so I can't.
2. This has merge conflicts with #11291, which I also need.

daltonbohning (approved):
LGTM and builds on Frontera

soumagne (approved):

src/cart/utils/SConscript
```diff
@@ -1,32 +1,26 @@
...
-    prereqs.require(env, 'protobufc')
+    Import('env', 'conf_dir')
 
     env.d_add_build_rpath()
 
+    cenv = env.Clone()
+
+    cenv.AppendUnique(CPPPATH=[Dir('../../mgmt').srcnode()])
+
+    cenv.require('protobufc', 'mercury')
```
ashleypittman, line 19: I think this change is what broke protobufc, the previous code was incorrectly modifying env so this was the right change but then protobufc should have been added back in src/test/ftest/cart.

PR #10933, opened 2022-11-30
DAOS-11663 common: fault injection to simulate DAOS  inconsistency

Nasf-Fan:
In includes the following:

1. DAOS_CHK_CONT_ORPHAN
   Take effect when destory the container, that will cause the
   container shards to be left on related targets then simulate
   the case of orphan container.

2. DAOS_CHK_CONT_BAD_LABEL
   Take effect when change the container label, that will cause
   the container label in the container service does not match
   the one in the container property.

### Gatekeeper:


src/include/daos/common.h
```diff
@@ -825,6 +825,10 @@ enum {
 #define DAOS_FORCE_EC_AGG_PEER_FAIL	(DAOS_FAIL_UNIT_TEST_GROUP_LOC | 0x9a)
 #define DAOS_FAIL_TX_CONVERT		(DAOS_FAIL_UNIT_TEST_GROUP_LOC | 0x9b)
 
+#define DAOS_CHK_POOL_ORPHAN		(DAOS_FAIL_UNIT_TEST_GROUP_LOC | 0xa0)
+#define DAOS_CHK_CONT_ORPHAN		(DAOS_FAIL_UNIT_TEST_GROUP_LOC | 0xa1)
+#define DAOS_CHK_CONT_BAD_LABLE		(DAOS_FAIL_UNIT_TEST_GROUP_LOC | 0xa2)
```
mjmac, line 830: typo: Should be `DAOS_CHK_CONT_BAD_LABEL`

mjmac (commented):
One small nit in the spelling of "label". I am not very familiar with daos_fail_loc, but I'll look into it.

src/mgmt/srv_pool.c
```diff
@@ -231,6 +231,9 @@ ds_mgmt_destroy_pool(uuid_t pool_uuid, d_rank_list_t *ranks)
 		return -DER_INVAL;
 	}
 
+	if (DAOS_FAIL_CHECK(DAOS_CHK_POOL_ORPHAN))
+		D_GOTO(out, rc = 0);
```
mjmac, line 235: Do we need this? We already have `dmg faults mgmt-svc pool $pool POOL_NONEXIST_ON_MS` which as far as I can tell does the same thing. It removes the pool entry from the MS so that the checker finds the orphaned pool.

src/container/srv_container.c
```diff
@@ -3135,6 +3135,9 @@ set_prop(struct rdb_tx *tx, struct ds_pool *pool,
 	if (rc != 0)
 		goto out;
 
+	if (DAOS_FAIL_CHECK(DAOS_CHK_CONT_BAD_LABLE))
+		goto out;
```
mjmac, line 3139: I don't think we need this either. We have `dmg faults mgmt-svc pool $pool POOL_BAD_LABEL` which will force the label in the MS to be different from the label stored as a property.

src/container/srv_target.c
```diff
@@ -1298,7 +1298,9 @@ ds_cont_tgt_destroy_handler(crt_rpc_t *rpc)
 	D_DEBUG(DB_MD, DF_CONT": handling rpc %p\n",
 		DP_CONT(in->tdi_pool_uuid, in->tdi_uuid), rpc);
 
-	rc = ds_cont_tgt_destroy(in->tdi_pool_uuid, in->tdi_uuid);
+	if (!DAOS_FAIL_CHECK(DAOS_CHK_CONT_ORPHAN))
+		rc = ds_cont_tgt_destroy(in->tdi_pool_uuid, in->tdi_uuid);
```
mjmac, line 1302: I think this is all we really need now, right? I think we can probably just use the `daos_debug_set_params` utility for this. I could make dmg shell out to that utility because it's not very nice to use, though.

src/mgmt/srv_pool.c
Nasf-Fan: Good to know that. I will remove it.
I think we also need the fault injection to simulate POOL_NONEXIST_ON_ENGINE, means to destroy the pool shards on targets, but with the pool entry kept on MS.

src/container/srv_container.c
Nasf-Fan: This is for container label, not pool label.

src/container/srv_target.c
Nasf-Fan: It am not sure whether it is convenient for @shimizukko to directly use the C API in his CR demo scripts or we have some utils named **daos_debug_set_params**?

src/container/srv_container.c
mjmac: Ah, right. Of course... I didn't read it carefully.

shimizukko (approved):

PR #10947, opened 2022-12-01
DAOS-11880 engine: Scale size of metrics shmem region

kjacque:
Previously the metrics shared memory region was created with a static size. However many metrics are now created on a per-target basis. The size needs to scale with the number of targets.

The estimates of the number of standard metrics and number of target metrics are higher than the actual number, to allow some buffer space for future additions.

### Gatekeeper:

### DIFF
--- a/src/engine/init.c
+++ b/src/engine/init.c
@@ -624,13 +624,19 @@ server_id_cb(uint32_t *tid, uint64_t *uid)
 	}
 }
 
+static uint64_t
+metrics_region_size(int num_tgts)
+{
+	const uint64_t	est_std_metrics = 1024; /* high estimate to allow for pool links */
+	const uint64_t	est_tgt_metrics = 128; /* high estimate */
+
+	return (est_std_metrics + est_tgt_metrics * num_tgts) * D_TM_METRIC_SIZE;
+}
+
 static int
 server_init(int argc, char *argv[])
 {
 	uint64_t		bound;
-	uint64_t		metrics_size;
-	const uint64_t		est_std_metrics = 1024; /* high estimate to allow for pool links */
-	const uint64_t		est_tgt_metrics = 128; /* high estimate */
 	unsigned int		ctx_nr;
 	int			rc;
 	struct engine_metrics	*metrics;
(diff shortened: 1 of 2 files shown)


src/engine/init.c
```diff
@@ -628,6 +628,9 @@ static int
 server_init(int argc, char *argv[])
 {
 	uint64_t		bound;
+	uint64_t		metrics_size;
+	const uint64_t		est_std_metrics = 1024; /* high estimate to allow for pool links */
+	const uint64_t		est_tgt_metrics = 128; /* high estimate */
```
liw, line 633: [Not requesting changes] Would be better to put these into a new function that takes a number of targets and returns a size. Because `server_init` involves many different "things," we should avoid introducing several variables for a single "thing."

liw (approved):

src/engine/init.c
kjacque: That is good advice, thank you. If I get any comments requiring me to re-push, I'll make this change.

liw: Thanks. :)

mjmac: For what it's worth, I also think that it would be a good idea to encapsulate this logic in a `d_tm_` helper. Currently we only use the library on the server, but there are loose plans to use it on the client side as well (e.g. [DAOS-8851](https://daosio.atlassian.net/browse/DAOS-8851)).

kjacque: I felt like this didn't belong in gurt as a d_tm_* function because it contains knowledge of the actual usage of the metrics in the server. (Pools as links to other regions, targets, etc.) I think it makes a good static helper function within this file though.

The calculation for clients would be a little different anyway, I think, since the metric set would be different.

kjacque: Got some errors in tests, so I went ahead and made the change @liw suggested.

jolivier23 (approved):

liw (approved):
Thank you.

Michael-Hennecke (approved):
The patch itself works, great.
 
But see my comment in https://daosio.atlassian.net/browse/DAOS-11880 regarding shmem segment cleanup...

jolivier23:
Do you not have githooks installed?

kjacque:
> Do you not have githooks installed?

I have them installed. Maybe they're not up to date, if something's changed semi-recently. I recall having to fiddle with them at some point when going back and forth a lot between master and release branches.

jolivier23:
> > Do you not have githooks installed?
> 
> I have them installed. Maybe they're not up to date, if something's changed semi-recently. I recall having to fiddle with them at some point when going back and forth a lot between master and release branches.

With the new method in utils/githooks/README.md, you shouldn't have to touch it anymore when switching branches.   We replaced symbolic links with just pointing the git config at the utils/githooks directory instead.

PR #10955, opened 2022-12-01
DAOS-12181 ds3: Don't use stack allocated variables

chucktuffli:
The function ds3_obj_read() allocates IOV and SGL variables on the stack for the underlying call to dfs_read(). This pattern works for synchronous calls, but it causes a segmentation violation for asynchronous read operations (i.e., when the event structure passed to dfs_read is not NULL).

Fix this by adding an SG list and IOV structures to the ds3_obj and using those variables in the call to dfs_read().

Signed-off-by: Chuck Tuffli <chuck.tuffli@hpe.com>

### Gatekeeper:

### DIFF
--- a/src/client/ds3/ds3_internal.h
+++ b/src/client/ds3/ds3_internal.h
@@ -58,8 +58,6 @@ struct ds3_bucket {
 struct ds3_obj {
 	/** DFS object handle */
 	dfs_obj_t *dfs_obj;
-	d_sg_list_t	ds3_sgl;
-	d_iov_t		ds3_iov;
 };
 
 /** DAOS S3 Upload Part handle */
--- a/src/client/ds3/ds3_internal.h
+++ b/src/client/ds3/ds3_internal.h
@@ -58,6 +58,8 @@ struct ds3_bucket {
 struct ds3_obj {
 	/** DFS object handle */
 	dfs_obj_t *dfs_obj;
+	d_sg_list_t	ds3_sgl;
+	d_iov_t		ds3_iov;
 };
 
 /** DAOS S3 Upload Part handle */
--- a/src/client/ds3/object.c
+++ b/src/client/ds3/object.c
@@ -6,6 +6,11 @@
 
 #include "ds3_internal.h"
 
+typedef struct ds3_obj_args {
+	d_iov_t	iov;
+	d_sg_list_t sg;
+} ds3_obj_args_t;
+
 /* helper */
 static bool
 ends_with(const char *str, const char *suffix)
@@ -199,6 +204,32 @@ ds3_obj_set_info(struct ds3_object_info *info, ds3_bucket_t *ds3b, ds3_obj_t *ds
 			     info->encoded_length, 0);
 }
 
+static int
+ds3_obj_read_int_cb(void *args, daos_event_t *ev, int ret)
+{
+	D_FREE(args);
+	return 0;
+}
+
+static int
+ds3_obj_read_int(void *buf, daos_off_t off, daos_size_t *size, ds3_bucket_t *ds3b, ds3_obj_t *ds3o,
+	     daos_event_t *ev)
+{
+	ds3_obj_args_t	*args;
+
+	D_ALLOC_PTR(args);
+	if (args == NULL)
+		return -DER_NOMEM;
+
+	d_iov_set(&args->iov, buf, *size);
+	args->sg.sg_nr     = 1;
+	args->sg.sg_iovs   = &args->iov;
+	args->sg.sg_nr_out = 1;
+
+	daos_event_register_comp_cb(ev, ds3_obj_read_int_cb, args);
+	return -dfs_read(ds3b->dfs, ds3o->dfs_obj, &args->sg, off, size, ev);
+}
+
 int
 ds3_obj_read(void *buf, daos_off_t off, daos_size_t *size, ds3_bucket_t *ds3b, ds3_obj_t *ds3o,
 	     daos_event_t *ev)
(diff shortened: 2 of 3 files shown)


src/client/ds3/ds3_internal.h
```diff
@@ -58,6 +58,8 @@ struct ds3_bucket {
 struct ds3_obj {
 	/** DFS object handle */
 	dfs_obj_t *dfs_obj;
+	d_sg_list_t	ds3_sgl;
+	d_iov_t		ds3_iov;
```
johannlombardi, line 62: I assume that we could have concurrent read operations against the same S3 object. In this case, they will all use the same sgl and iov and causes corruption, no?

johannlombardi (changes requested):

src/client/ds3/ds3_internal.h
mchaarawi: that's reasonable given the ds3 API accepts an ev. although the sgl you are passing would replace the `void *buf` since the sgl is the memory buffer descriptor. If we do that, then should also update the ds3 write API to match.

alternatively one could allocate a per IO struct and free it on completion by registering a completion cb on the event. No strong preference on my side, but slight to the former.

johannlombardi: It sounds like libds3 users won't need to manage scatter-gather list, so it sounds to me that we should hide the complexity of managing the sgl from the user. I would personally then be in favor of keeping the API simple with just a buf and manage the sgl internally (possibly as in a completion callback as suggested by Mohamad).

mssawant: Presently, for every S3 PUT/GET operation a new `ds3_obj` is allocated, thus I think it is safe to embed iov and sgl in `ds3_obj`.

johannlombardi: Thanks. I still think that we should use a callback. @chucktuffli, is that something you could please look into? I think that you can use the dfs read path as an example: https://github.com/daos-stack/daos/blob/master/src/client/dfs/dfs.c#L4646
Thanks in advance.

chucktuffli: The last push removed the SGL/IOV from the object and follows a pattern similar to `dfs_read` in that it a) uses stack variables for the synchronous case (i.e., event is NULL) and b) it allocates the SGL/IOV for the asynchronous case and frees the memory in a callback registered via `daos_event_register_comp_cb()`

johannlombardi: Thanks. Asking for folks to review it (otherwise it does not appear in review list).

johannlombardi (approved):

mchaarawi (dismissed):

src/client/ds3/object.c
```diff
@@ -199,21 +204,51 @@ ds3_obj_set_info(struct ds3_object_info *info, ds3_bucket_t *ds3b, ds3_obj_t *ds
...
-
 	if (ds3b == NULL || buf == NULL || ds3o == NULL)
 		return -EINVAL;
 
-	d_iov_set(&iov, buf, *size);
-	rsgl.sg_nr     = 1;
-	rsgl.sg_iovs   = &iov;
-	rsgl.sg_nr_out = 1;
-	return -dfs_read(ds3b->dfs, ds3o->dfs_obj, &rsgl, off, size, ev);
+	if (ev == NULL) {
```
mssawant, line 240: Alternatively, `ev` NULL check can be moved to `ds3_obj_read_int()` before invoking `daos_event_register_comp_cb()` or merge `ds3_obj_read_int()` and `ds3_obj_read()`?

chucktuffli: I implemented it with a separate function for the async case as that is what `dfs_read` does. But I'm fine combining the functions if that is preferred.

mssawant: Since we are trying to avoid stack allocated variables here, may be we should do that unconditionally with respect to `ev` 's presence.
Presently we are allocating this structure only in the async case and rgw is not using async path as of now.
We can allocate `ds3_obj_args_t` for both sync and async by invoking `ds3_obj_read_int()` and configuring the callback if `ev`'s not NULL. That way we can avoid code duplication and have the same code path.

It would have been great to have altogether separate functions for sync and async from readability perspective,
e.g. `ds3_obj_read(void *buf, daos_off_t off, daos_size_t *size, ds3_bucket_t *ds3b, ds3_obj_t *ds3o)` and `ds3_obj_read_async(void *buf, daos_off_t off, daos_size_t *size, ds3_bucket_t *ds3b, ds3_obj_t *ds3o,
	     daos_event_t *ev)`
Check `Java_io_daos_dfs_DaosFsClient_dfsRead()` and `Java_io_daos_dfs_DaosFsClient_dfsReadAsync()`. 
But let's not do that here.
I see different conventions at client and dfs levels though, may be this is fine, but then we must not use stack in either case for consistency, I think. I will leave it up to you.

mchaarawi: no we should not do allocation in case of the blocking case, just to avoid the unecessary overhead IMO.

mssawant: I am okay, since we are explicitly just handling the async case here. @mchaarawi, another question I had is, I believe `dfs_read()` would return immediately being async, the actual operation result would be received in the call back's `ret` param, right? we are not really using that or propagating that in this case, what do you think? or is it that we need to handle it separately?

chucktuffli: > no we should not do mem allocation in case of the blocking case, just to avoid the unnecessary overhead IMO.

Would it still make sense to check for a NULL `ev` in `ds3_obj_read_int()`? If so, would the following be correct?
```diff
diff --git a/src/client/ds3/object.c b/src/client/ds3/object.c
index fd9e7c838..f80fd5dfb 100644
--- a/src/client/ds3/object.c
+++ b/src/client/ds3/object.c
@@ -217,6 +217,9 @@ ds3_obj_read_int(void *buf, daos_off_t off, daos_size_t *size, ds3_bucket_t *ds3
 {
        ds3_obj_args_t  *args;

+       D_ASSERT(ev);
+       daos_event_errno_rc(ev);
+
        D_ALLOC_PTR(args);
        if (args == NULL)
                return -DER_NOMEM;
```

johannlombardi (approved):

mssawant (approved):

johannlombardi:
Please inspect.

mchaarawi (approved):

johannlombardi:
Thanks a lot for working on this and providing a patch!

PR #10961, opened 2022-12-01
DAOS-12168 dfs: add new API to check consistency of DFS namespace

mchaarawi:
New API to scan through a DFS container and mark all OIDs as seen in an OID table, then scan the OID table to check for unmarked objects that are were not seen in the DFS namespace and print or punch those leaked objects.

Required-githooks: true

### Gatekeeper:


src/client/dfs/dfs.c
```diff
@@ -6479,3 +6478,246 @@ dfs_obj_anchor_set(dfs_obj_t *obj, uint32_t index, daos_anchor_t *anchor)
...
+	d_iov_t		marker;
+	bool		mark_data = true;
+	int		rc;
+
+	oit.cookie = (uint64_t)arg;
+
+	/** open the entry name and get the oid */
+	rc = dfs_lookup_rel(dfs, parent, name, O_RDONLY, &obj, NULL, NULL);
+	if (rc) {
+		D_ERROR("dfs_open() if %s failed: %d\n", name, rc);
```
wiliamhuang, line 6499: Is "dfs_open() if" typo?

src/include/daos_fs.h
```diff
@@ -1122,6 +1121,27 @@ dfs_removexattr(dfs_t *dfs, dfs_obj_t *obj, const char *name);
...
+
+enum {
+	DFS_CHECK_PRINT		= (1 << 0),
+	DFS_CHECK_REMOVE	= (1 << 1),
+	DFS_CHECK_LINK_LF	= (1 << 2),
+};
+
+/**
+ * Scan the DFS namespace and check if there are any leaked objects. Add those object to
+ * "Lost+Found" directory.
```
liuxuezhao, line 1133: the comment looks can be improved a little bit, as not always add to LF (based on the flags).

src/client/dfs/dfs.c
```diff
@@ -6479,3 +6478,246 @@ dfs_obj_anchor_set(dfs_obj_t *obj, uint32_t index, daos_anchor_t *anchor)
...
+	}
+
+	/** descend into directories */
+	if (S_ISDIR(obj->mode))	{
+		daos_anchor_t		anchor = {0};
+		uint32_t		nr_entries = DFS_NR_ITERATE;
+
+		while (!daos_anchor_is_eof(&anchor)) {
+			rc = dfs_iterate(dfs, obj, &anchor, &nr_entries, DFS_MAX_NAME * nr_entries,
+					 oit_mark_cb, (void *)oit.cookie);
```
liuxuezhao, line 6526: just confirm do we have a limitation for the #layers of nested directory in dfs? if no limitation such recursion possibly cause stack overflow in extreme case?

liuxuezhao (approved):

src/client/dfs/dfs.c
ashleypittman, line 6499: Agreed, the called function is dfs_lookup_rel() but the log says dfs_open(), this looks like a cut-and-paste error or similar.

src/client/dfs/dfs.c
```diff
@@ -6479,3 +6478,246 @@ dfs_obj_anchor_set(dfs_obj_t *obj, uint32_t index, daos_anchor_t *anchor)
...
+		D_GOTO(out_obj, rc);
+
+	d_iov_set(&marker, &mark_data, sizeof(mark_data));
+	rc = daos_oit_mark(oit, oid, &marker, NULL);
+	/*
+	 * If the entry exists but the file or directory are empty, the corresponding oid itself has
+	 * not been written to, so it doesn't exist in the OIT. The mark operation would return
+	 * NONEXIST in this case, so check and avoid returning an error in this case.
+	 */
+	if (rc && rc != DER_NONEXIST) {
```
ashleypittman, line 6514:
```suggestion
	if (rc && rc != -DER_NONEXIST) {
```

src/client/dfs/dfs.c
ashleypittman: Agree, particularly with an anchor on the stack this could run out of space quickly.  There's not a patricianly good alternative I can easily think of though.

I don't think symlinks will be dereferenced however a corrupt container could contain loops which would crash this code no matter how big the stack was.

src/client/dfs/dfs.c
```diff
@@ -6479,3 +6478,246 @@ dfs_obj_anchor_set(dfs_obj_t *obj, uint32_t index, daos_anchor_t *anchor)
...
+				enum daos_otype_t	otype = daos_obj_id2type(oids[i]);
+				struct timespec		now;
+				char			name[DFS_MAX_NAME + 1];
+				daos_size_t		len;
+
+				entry.uid = geteuid();
+				entry.gid = getegid();
+				oid_cp(&entry.oid, oids[i]);
+				if (daos_is_array_type(otype))
+					entry.mode = S_IFREG | 0755;
```
ashleypittman, line 6668: Will symlinks show up as files here?

src/client/dfs/dfs.c
```diff
@@ -6479,3 +6478,246 @@ dfs_obj_anchor_set(dfs_obj_t *obj, uint32_t index, daos_anchor_t *anchor)
...
+
+			if (flags & DFS_CHECK_LINK_LF) {
+				struct dfs_entry	entry = {0};
+				enum daos_otype_t	otype = daos_obj_id2type(oids[i]);
+				struct timespec		now;
+				char			name[DFS_MAX_NAME + 1];
+				daos_size_t		len;
+
+				entry.uid = geteuid();
+				entry.gid = getegid();
```
ashleypittman, line 6665: These will be loop invariants.

src/client/dfs/dfs.c
```diff
@@ -6479,3 +6478,246 @@ dfs_obj_anchor_set(dfs_obj_t *obj, uint32_t index, daos_anchor_t *anchor)
...
+					entry.mode = S_IFREG | 0755;
+				else
+					entry.mode = S_IFDIR | 0755;
+				rc = clock_gettime(CLOCK_REALTIME, &now);
+				if (rc)
+					D_GOTO(out_lf, rc = errno);
+				entry.atime = entry.mtime = entry.ctime = now.tv_sec;
+				entry.atime_nano = entry.mtime_nano = entry.ctime_nano =
+					now.tv_nsec;
+				entry.chunk_size = dfs->attr.da_chunk_size;
```
ashleypittman, line 6677: I think it would be easier to call this once before the loop as well and just have all recovered files having the same timestamp.

src/client/dfs/dfs.c
```diff
@@ -6479,3 +6478,246 @@ dfs_obj_anchor_set(dfs_obj_t *obj, uint32_t index, daos_anchor_t *anchor)
...
+			D_ERROR("dfs_iterate() failed: %d\n", rc);
+			D_GOTO(out_oit, rc);
+		}
+
+		nr_entries = DFS_NR_ITERATE;
+	}
+
+	/** Create lost+found directory and properly link unmarked oids there. */
+	if (flags & DFS_CHECK_LINK_LF) {
+		rc = dfs_open(dfs, NULL, "lost+found", S_IFDIR | 0755, O_CREAT | O_RDWR, 0, 0, NULL,
```
ashleypittman, line 6616: I expect the filename will be unique over time but there's no reason why this couldn't use a unique directory name here which might be useful if this utility is going to be run multiple times.

As this is accessible via a API then perhaps we should pass in (or out) the directory names used here.

src/client/dfs/dfs.c
```diff
@@ -6479,3 +6478,246 @@ dfs_obj_anchor_set(dfs_obj_t *obj, uint32_t index, daos_anchor_t *anchor)
...
+		}
+	}
+
+	/** list all unmarked oids */
+	memset(&anchor, 0, sizeof(anchor));
+	while (!daos_anchor_is_eof(&anchor)) {
+		nr_entries = DFS_NR_ITERATE;
+		rc = daos_oit_list_unmarked(oit, oids, &nr_entries, &anchor, NULL);
+		if (rc) {
+			D_ERROR("daos_oit_list_unmarked failed: "DF_RC"\n", DP_RC(rc));
```
ashleypittman, line 6630:
```suggestion
			D_ERROR("daos_oit_list_unmarked() failed: "DF_RC"\n", DP_RC(rc));
```

src/include/daos_fs.h
```diff
@@ -1122,6 +1121,27 @@ dfs_removexattr(dfs_t *dfs, dfs_obj_t *obj, const char *name);
 int
 dfs_listxattr(dfs_t *dfs, dfs_obj_t *obj, char *list, daos_size_t *size);
 
+
+enum {
+	DFS_CHECK_PRINT		= (1 << 0),
+	DFS_CHECK_REMOVE	= (1 << 1),
+	DFS_CHECK_LINK_LF	= (1 << 2),
```
ashleypittman, line 1128: Better documentation would make sense here, having reviewed the PR I know what "LF" means but if someone just opens the header file without context it won't be so obvious.

src/include/daos_fs.h
```diff
@@ -1122,6 +1121,27 @@ dfs_removexattr(dfs_t *dfs, dfs_obj_t *obj, const char *name);
...
+};
+
+/**
+ * Scan the DFS namespace and check if there are any leaked objects. Add those object to
+ * "Lost+Found" directory.
+ *
+ * \param[in]	dfs	Pointer to the mounted file system.
```
ashleypittman, line 1135: This looks like it's out of date with the code and will need fixing so the Doxygen is inline with the code.
```suggestion
 * \param[in]	poh	Open pool handle
```

src/client/dfs/dfs.c
mchaarawi: yes indeed

src/include/daos_fs.h
mchaarawi: yes, sorry i updated the API later and forgot about the comment.

src/client/dfs/dfs.c
mchaarawi: yes we have actually discussed this in the WG. we are not handling this problem here of a loop caused by the user.
but yes i don't have an alternative option for now, and i think we can tackle this later. ill make a note of it to be tracked and discussed with stakeholders.

src/client/dfs/dfs.c
mchaarawi: thx!

src/client/dfs/dfs.c
mchaarawi: no they will show as dirs, which i know is wrong. the L+F is still a work in progress and still has some holes. i would like to address those later after landing the initial implementation to the feature branch (DAOS-12176).

src/client/dfs/dfs.c
mchaarawi: sure

src/client/dfs/dfs.c
mchaarawi: im not quite sure what you mean here. are you talking about the name of the directory lost+found or something else as you are referring to directories (plural)?

ashleypittman: This implements traditional unix semantics of putting all files into a single "lost_and_found" directory, I was wondering if creating one subdirectory per invocation and putting all recovered files from that run into the newly created subdirectory.

You'd then get "lost+found/<timestamp>/<objectid>" with a different timestamp for each time the utility is run.  Among other things it would allow users of the API iterate over newly created files without having to handle files recovered from a previous run.

mchaarawi: ok makes sense to me. i'll discuss this at the WG meeting. for now let's tackle this with the l+f ticket.

wiliamhuang (approved):

mchaarawi:
@ashleypittman any more feedback aside from the l+f improvements you suggested?

mchaarawi:
> Looks good to me. I assume more work is needed before we can use it on more complex containers (for example ones with symlinks)? It would be good to add some system testing once we think it's feature complete.

actually symlinks will not cause issues as they are not stored in separate objects. so there are not going to be any leaked symlinks because the value is stored directly in the entry itself.
as for system testing, this is exactly why this is going to a feature branch and not master. there is a separate plan for way more system test that there is in this PR.

PR #10966, opened 2022-12-02
DAOS-12146 cq: Update pylint to work with VS code.

ashleypittman:
Required-githooks: true


utils/cq/daos_pylint.py
```diff
@@ -552,9 +563,13 @@ def main():
...
                         default='{path}:{line}:{column}: {message-id}: {message} ({symbol})')
     parser.add_argument('--reports', choices=['y', 'n'], default='y')
-    parser.add_argument('--output-format', choices=['text'])
+    parser.add_argument('--output-format', choices=['text', 'json'], default='text')
     parser.add_argument('--rcfile', default=rcfile)
     parser.add_argument('--diff', action='store_true')
+    parser.add_argument('--version', action='store_true')
+
+    # Args that VS Code uses.
+    parser.add_argument('--import')
```
daltonbohning, line 572: Would it make more sense to handle the `sys.path.append(arg[9:])` here instead of globally at the top of the file?

daltonbohning (approved):

utils/cq/daos_pylint.py
ashleypittman: No, you need to change the sys.path before importing pylint.  The only reason this is here is so that if this argparse doesn't complain about the option.

kccain (dismissed):

daltonbohning (approved):

kccain (approved):

ashleypittman:
The one test failure appears to be DAOS-12399

PR #10975, opened 2022-12-05
DAOS-11605 ad_mem: fix arena free size

wangshilong:
There might be some free space from arena unused for specific non-pow2 group unit, those should be included in free size of arena.

Groups might be reset with aligned-pow2 group which will be able to use all arena space later.


liuxuezhao:
@wangshilong this PR does looks like fixed a problem, however liang's https://github.com/daos-stack/daos/pull/10927 fixed similar issue. just FYI.

wangshilong:
> @wangshilong this PR does looks like fixed a problem, however liang's #10927 fixed similar issue. just FYI.

Indeed, that fixes a lot of problems.

PR #10981, opened 2022-12-06
DAOS-12058 common: Fixed leaks in BMEM WAL reserve/commit.

sherintg:
This fix ensures that all internal data related to dav transaction are cleaned up before invoking WAL reserve/commit calls.

### Gatekeeper:

### DIFF
--- a/src/common/dav/tx.c
+++ b/src/common/dav/tx.c
@@ -507,7 +507,25 @@ dav_tx_begin(dav_obj_t *pop, jmp_buf env, ...)
 
 		VALGRIND_START_TX;
 	} else if (tx->stage == DAV_TX_STAGE_NONE) {
+		struct umem_wal_tx *utx;
+
 		DAV_DBG("");
+		utx = pop->do_utx;
+		if (utx == NULL) {
+			utx = dav_umem_wtx_new(pop);
+			if (utx == NULL) {
+				err = ENOMEM;
+				goto err_abort;
+			}
+		}
+		err = dav_wal_tx_reserve(pop);
+		if (err) {
+			D_ERROR("so_wal_reserv failed, "DF_RC"\n", DP_RC(err));
+			D_FREE(utx);
+			pop->do_utx = NULL;
+			goto err_abort;
+		}
+		tx = get_tx();
 
 		VALGRIND_START_TX;
 
@@ -545,26 +563,7 @@ dav_tx_begin(dav_obj_t *pop, jmp_buf env, ...)
 
 	DAV_SLIST_INSERT_HEAD(&tx->tx_entries, txd, tx_entry);
 
-	if (tx->stage == DAV_TX_STAGE_NONE) {
-		struct umem_wal_tx *utx;
-
-		tx->stage = DAV_TX_STAGE_WORK;
-		utx = pop->do_utx;
-		if (utx == NULL) {
-			utx = dav_umem_wtx_new(pop);
-			if (utx == NULL) {
-				err = ENOMEM;
-				goto err_abort;
-			}
-		}
-		err = dav_wal_tx_reserve(pop);
-		if (err) {
-			D_ERROR("so_wal_reserv failed, "DF_RC"\n", DP_RC(err));
-			D_FREE(utx);
-			pop->do_utx = NULL;
-			goto err_abort;
-		}
-	}
+	tx->stage = DAV_TX_STAGE_WORK;
 
 	/* handle locks */
 	va_list argp;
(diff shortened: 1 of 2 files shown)


src/common/dav/tx.c
```diff
@@ -507,7 +507,25 @@ dav_tx_begin(dav_obj_t *pop, jmp_buf env, ...)
...
+			}
+		}
+		err = dav_wal_tx_reserve(pop);
+		if (err) {
+			D_ERROR("so_wal_reserv failed, "DF_RC"\n", DP_RC(err));
+			D_FREE(utx);
+			pop->do_utx = NULL;
+			goto err_abort;
+		}
+		tx = get_tx();
```
ivadmitry, line 528: get_tx() returns per-thread tx structure pointer, it is the same for all ULTs on this pthread. It is not a ULT local store data. Re-loading the same tx address doesn't serve the purpose of UTL's data separation.

src/common/dav/tx.c
```diff
@@ -545,26 +563,7 @@ dav_tx_begin(dav_obj_t *pop, jmp_buf env, ...)
...
-		}
-		err = dav_wal_tx_reserve(pop);
-		if (err) {
-			D_ERROR("so_wal_reserv failed, "DF_RC"\n", DP_RC(err));
-			D_FREE(utx);
-			pop->do_utx = NULL;
-			goto err_abort;
-		}
-	}
+	tx->stage = DAV_TX_STAGE_WORK;
```
ivadmitry, line 566: We can't access tx after the context switch.

ivadmitry (approved):
This patch partly resolves issues caused by multiple UTLs entering tx_begin() and tx_end() but not comprehensively: the tx->stage flow switch still requires the future work.

src/common/dav/tx.c
```diff
@@ -507,7 +507,25 @@ dav_tx_begin(dav_obj_t *pop, jmp_buf env, ...)
 
 		VALGRIND_START_TX;
 	} else if (tx->stage == DAV_TX_STAGE_NONE) {
```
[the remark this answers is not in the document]
jolivier23: we could consider using something private to the ult.   But it's a good point and bit us in PMDK tx_end before, fixed in this commit to pmdk

`commit 33596644174cc3de6610482fc64b86e6e010059f
Author: Piotr Balcer <piotr@balcer.eu>
Date:   Fri Oct 8 11:18:38 2021 +0200

    obj: make sure nothing is accessed TX_STAGE_NONE cb
    
    The transactional TX_STAGE_NONE callback is executed after
    the end of the transaction indicating that transaction
    can now execute on the same thread. However, currently
    the pmemobj_tx_end() function will actually read from
    the per-thread transaction state after this callback.
    
    This patch addresses the problem by reading all the necessary
    state prior to the callback.`

src/common/dav/tx.c
NiuYawei: Since wal_tx_reserve() guarantees first call first return, it should be fine, but I strongly suggest move this into the "else if (tx->stage == DAV_TX_STAGE_NONE)" branch and add a assert of "tx->stage == DAV_TX_STAGE_NONE" before change it.

src/common/dav/tx.c
```diff
@@ -507,7 +507,25 @@ dav_tx_begin(dav_obj_t *pop, jmp_buf env, ...)
...
+		struct umem_wal_tx *utx;
+
 		DAV_DBG("");
+		utx = pop->do_utx;
+		if (utx == NULL) {
```
NiuYawei, line 514: The 'do_utx' could be non-NULL if it was created by a reserve call before? I'm wondering if we should differentiate the two cases of redo from reserve and redo from tx, so that code will be less bug prone? Though that could introduce more complexities. Just a suggestion to consider, not necessary to make change in this PR.

NiuYawei (changes requested):

src/common/dav/tx.c
sherintg: Similar assert is already present in line 543. We reach this point only if tx->stage is DAV_TX_STAGE_NONE or DAV_TX_STAGE_WORK.

src/common/dav/tx.c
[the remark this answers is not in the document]
sherintg: The latest change addresses the following:
- the tx structure elements for the ULT are initialized only after wal_reserve
- the tx structure elements are cleaned up and stage changed to DAV_TX_STAGE_NONE before call to wal_commit

Since the ULT does not yield while in stage other then DAV_TX_STAGE_NONE, I do not see tx getting shared now.

On the call back, the latest fix will cache the callback data before clearing tx and calling wal_commit . Then it calls the callback after wal_commit. This is fine since the callback was already called after stage is set to DAV_TX_STAGE_NONE.

src/common/dav/tx.c
sherintg: Yes, this call is redundant. Is it ok I clean it up if future PR?

src/common/dav/tx.c
sherintg: We will relook at this. Right now, this code is to handle case where we want to support certain atomic operation that does not want its change to be immediately committed but committed later along with next transaction that happens on this vos object.

src/common/dav/tx.c
NiuYawei: You mean the "FATAL("Invalid stage..)" above? That looks assert on different things to me, my proposal is like:

else if (tx->stage == DAV_TX_STAGE_NONE) {
    wal_reserve();
    D_ASSERT(tx->stage == DAV_TX_STAGE_NONE);  /* Nobody change tx stage while I yield in wal_reserve() */
    tx->stage = DAV_TX_STAGE_WORK;
}

I'm fine with cleanup it in follow-on PR.

NiuYawei (approved):

jolivier23 (approved):

gnailzenh (approved):

PR #10988, opened 2022-12-07
DAOS-12216 gurt: increase share memory for telemetry

wangshilong:
Creating pool failed if target number exceeds 22, increase shared memory from 1MiB to 2MiB.

Required-githooks: true


wiliamhuang (approved):

kjacque (commented):
I believe this PR is no longer needed. See #10947 which scales the size of the telemetry shmem based on the number of targets.

wangshilong:
> I believe this PR is no longer needed. See #10947 which scales the size of the telemetry shmem based on the number of targets.

Thanks for pointing out, this PR might not be included when i tested feature/vos_on_blob branch.

PR #11023, opened 2022-12-12
DAOS-623 build: Improve Go builder for scons

mjmac:
The include scanning code needs to take into account
import aliases, so it can't just check that a line
starts with the import string.

Required-githooks: true

### DIFF
--- a/site_scons/site_tools/go_builder.py
+++ b/site_scons/site_tools/go_builder.py
@@ -33,9 +33,11 @@ def _scan_go_file(node, _env, _path):
                     includes.append(File(os.path.join(src_dir, header)))
                 else:
                     includes.append(f'../../../include/{header}')
-        elif line.strip().startswith(GOINC_PREFIX):
-            deps_dir = line[len(GOINC_PREFIX) + 1:-1]
-            includes.extend(Glob(f'src/{deps_dir}/*.go'))
+        else:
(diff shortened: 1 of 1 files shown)


ashleypittman:
I'm surprised not to have seen more revisions for this!  Can you post an example where the current scanner doesn't work and I'll test it?

mjmac:
> I'm surprised not to have seen more revisions for this! Can you post an example where the current scanner doesn't work and I'll test it?

Modify anything under `src/control/proto/mgmt/`. The scanner doesn't pick up these changes because most imports alias it to `mgmtpb`, so the line doesn't start with the expected prefix.

I've also started using `--implicit-deps-changed` for every build just to ensure that things are rebuilt as expected, but that may just be flailing.

jolivier23 (approved):

ashleypittman:
It may be overkill but it looks like "go list --json <file.go>" will also list the imports and possibly improve this.  I can code something up today if you like.

ashleypittman (approved):

ashleypittman:
Alternative implemented in https://github.com/daos-stack/daos/pull/11041 and if it works it should be much more resilient, can you take a look please?

mjmac:
Closing in favor of the approach taken in #11041.

PR #11027, opened 2022-12-12
DAOS-12255 control: Correct hugepage affinity with multi-engine-per-NUMA

tanabarr:
When multiple engines coexist on the same NUMA node (and there is an
equal number of engines per NUMA node), divide the total number of
hugepages proportionally across NUMA nodes when allocating.

Features: control

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/server/server_utils.go
+++ b/src/control/server/server_utils.go
@@ -209,22 +209,22 @@ func getFabricNetDevClass(cfg *config.Server, fis *hardware.FabricInterfaceSet) 
 // Detect the number of engine configs assigned to each NUMA node and return error if engines are
 // distributed unevenly across NUMA nodes. Otherwise return sorted list of NUMA nodes in use.
 // Configurations where all engines are on a single NUMA node will be allowed.
-func getEngineNUMANodes(log logging.Logger, engineCfgs []*engine.Config) ([]int, error) {
+func getEngineNUMANodes(log logging.Logger, engineCfgs []*engine.Config) ([]string, error) {
 	nodeMap := make(map[int]int)
 	for _, ec := range engineCfgs {
 		nodeMap[int(ec.Storage.NumaNodeIndex)] += 1
 	}
 
 	var lastCount int
-	nodes := make([]int, 0, len(engineCfgs))
+	nodes := make([]string, 0, len(engineCfgs))
 	for k, v := range nodeMap {
 		if lastCount != 0 && v != lastCount {
 			return nil, FaultEngineNUMAImbalance(nodeMap)
 		}
 		lastCount = v
-		nodes = append(nodes, k)
+		nodes = append(nodes, fmt.Sprintf("%d", k))
 	}
-	sort.Ints(nodes)
+	sort.Strings(nodes)
 
 	return nodes, nil
 }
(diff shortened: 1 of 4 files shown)


src/control/server/server_utils.go
```diff
@@ -208,20 +208,23 @@ func getFabricNetDevClass(cfg *config.Server, fis *hardware.FabricInterfaceSet)
...
-			nodes = []string{nn}
-			break
+		nodeMap[int(ec.Storage.NumaNodeIndex)] += 1
+	}
+
+	var lastCount int
+	nodes := make([]int, 0, len(engineCfgs))
+	for k, v := range nodeMap {
+		if lastCount != 0 && v != lastCount {
+			log.Errorf("different number of engines assigned to each NUMA node (%+v),"+
```
kjacque, line 221: Seems like more of a Notice to me

src/control/server/server_utils_test.go
```diff
@@ -555,6 +556,35 @@ func TestServer_prepBdevStorage(t *testing.T) {
 			expMemSize:      16384,
 			expHugePageSize: 2,
 		},
+		"4 engines; hugepage alloc across numa 0,1": {
```
kjacque, line 559: Would be interesting to add a test for the case where there is an uneven distribution of engines.

src/control/server/server_utils.go
tanabarr: It is an error condition, having a different number of engines running on each NUMA will cause problems so I think it should be raised as an error in the log but for extraneous circumstances we want to allow the server to start in this case. As an aside, it was previously a Notice but wasn't being printed for some reason when the log mask was set to DEBUG.

src/control/server/server_utils_test.go
tanabarr: On inspection decided that it should be an error if engine configs are imbalanced across numa nodes (unless all engines are on a single node). Added unit test case for uneven distribution as requested.

src/control/server/server_utils.go
tanabarr: I changed this to return an error.

kjacque (dismissed):

src/control/server/server_utils.go
```diff
@@ -291,7 +298,9 @@ func prepBdevStorage(srv *server, iommuEnabled bool) error {
 		// allocation as some overhead may result in one or two being unavailable.
 		prepReq.HugePageCount = srv.cfg.NrHugepages / len(numaNodes)
 		prepReq.HugePageCount += common.ExtraHugePages
-		prepReq.HugeNodes = strings.Join(numaNodes, ",")
+		prepReq.HugeNodes = strings.Trim(
+			strings.Join(strings.Split(fmt.Sprint(numaNodes), " "), ","),
+			"[]")
```
mjmac, line 303: 🤨 ... You're tying this functionality to the current implementation of how the language prints an integer slice. Is it likely to change? No. But I don't think the language compatibility guarantee extends to `fmt.Sprint()` output.

It looks like the only place you're using the `numaNodes` value returned by `getEngineNUMANodes()` is here, so why not just generate the string slice in that function while you're iterating over the nodes anyhow?

tanabarr: that's what it's used to do and I will change back, I was going back on forth on this myself

tanabarr: done

kjacque (approved):

tanabarr:
single known CI failure:
- TelemetryIOLatency DAOS-12013

mjmac (approved):

PR #8426, opened 2022-03-09
DAOS-10037 cart: Abort RPCs to dead ranks

liw:
WIP

Signed-off-by: Li Wei <wei.g.li@intel.com>

PR #11030, opened 2022-12-13
DAOS-10037 cart: Improve the abort code

liw:
There are a few abort problems in two future use cases.

  - In a daos_engine, the SWIM xstream aborts all in-flight RPCs sent by
    all xstreams to a specific rank. This will be implemented by https://github.com/daos-stack/daos/pull/8426.
    When called by the SWIM xstream, crt_req_abort may
      - introduce data races on crt_rpc_priv fields with the xstream
	owning the RPC;
      - race with crt_req_send_internal, crt_hg_req_send_cb, etc., and
	is incomplete in the handling of certain RPC states;
      - invoke the RPC completion callback on the SWIM xstream instead
	of the xstream owning the RPC.

  - In dfuse, multiple libdaos application threads send, progress, and
    abort RPCs in the same context. We might introduce abort calls to
    implement dfuse operation-level timeouts or faster shutdowns. The
    first two problems under the daos_engine use case also apply here.

This patch addresses those problems in the following manner.

  - Add crt_rpc_priv.crp_mutex to prevent data races on crt_rpc_priv
    fields.
      - Extend and fix the existing fine-to-coarse lock order to prevent
	deadlocks. See the comment on crt_rpc_priv.crp_mutex.
      - Remove the recursions in the
	crt_req_send_internal-crt_req_hg_add_lookup path.
      - Remove crp_on_wire, which is superseded by crp_mutex and
	crp_state.
      - Remove the unused RPC_STATE_REPLY_RECEVED.

  - Use crt_context.cc_bh_timeout to avoid calling RPC completion
    callbacks in undesirable threads. (We could also introduce a new
    queue tracking RPCs to be aborted.)
      - Set crt_rpc_priv.crp_timeout_ts to 0 in crt_req_abort so that
	the progress thread will complete the RPC. This will return
	-DER_TIMEDOUT instead of -DER_CANCELED for an aborted RPC. It is
	not an issue for daos since daos already treats -DER_CANCELED as
        a network error that shall be retried.

  - Print which RPCs can't be aborted in crt_ctx_epi_abort. This has
    been helpful during the development.

  - Remove the recursion in crt_context_req_untrack.

Required-githooks: true

### Gatekeeper:


liw:
@frostedcmos, @liuxuezhao, could you let me know if the approach I use in this patch is acceptable or not, please? I'm not sure about the change on when to start the RPC timer.

liw:
The test failure also happens with other PRs and therefore is unrelated.

src/cart/crt_context.c
```diff
@@ -1202,50 +1362,49 @@ crt_context_req_untrack(struct crt_rpc_priv *rpc_priv)
...
-		D_MUTEX_LOCK(&crt_ctx->cc_mutex);
-		rc = crt_req_timeout_track(tmp_rpc);
-		D_MUTEX_UNLOCK(&crt_ctx->cc_mutex);
-		if (rc != 0)
-			RPC_ERROR(tmp_rpc,
-				"crt_req_timeout_track failed, rc: %d.\n", rc);
+		D_MUTEX_LOCK(&tmp_rpc->crp_mutex);
+		D_MUTEX_LOCK(&epi->epi_mutex);
+		if (tmp_rpc->crp_state == RPC_STATE_QUEUED && credits_available(epi) > 0) {
+			tmp_rpc->crp_state = RPC_STATE_INITED;
```
liuxuezhao, line 1375: original calling of "crt_req_timeout_track" removed, not quite sure you mean it is unnecessary?

src/cart/crt_context.c
```diff
@@ -902,6 +1037,15 @@ crt_req_timeout_hdlr(struct crt_rpc_priv *rpc_priv)
 		d_tm_inc_counter(crt_ctx->cc_timedout, 1);
 
 	switch (rpc_priv->crp_state) {
+	case RPC_STATE_QUEUED:
+		RPC_ERROR(rpc_priv, "aborting waiting to group %s, rank %d, tgt_uri %s\n",
+			  grp_priv->gp_pub.cg_grpid, tgt_ep->ep_rank, rpc_priv->crp_tgt_uri);
+		D_MUTEX_LOCK(&rpc_priv->crp_epi->epi_mutex);
+		d_list_del_init(&rpc_priv->crp_epi_link);
```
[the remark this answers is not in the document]
liw: I think you're right that we need to RPC_DECREF here. I've modified crt_context_req_untrack a little (otherwise it can't handle QUEUED request) and used it here.

src/cart/crt_context.c
liw: Yes, this patch changes crt_context_req_track to call crt_req_timeout_track for both waiting and inflight requests. Hence, the waiting request here should already be in the timeout binheap.

liw (commented):
Thanks, Xuezhao.

src/cart/crt_context.c
liuxuezhao: ok, original code will not add to timeout binheap if it is queued, but seems not bad to track timeout for QUEUED case also.

liuxuezhao (approved):
LGTM, there are a few test failures, not very sure if it is related although.

liw:
> LGTM, there are a few test failures, not very sure if it is related although.

Yeah, I've been investigating the rebuild_rc failure. It appears that the progress ULT on the system xstream of rank 1 was blocked by some Argobots lock or eventual, because that xstream had not been able to receive any RPCs since while it could still run other ULTs.

liw:
> > LGTM, there are a few test failures, not very sure if it is related although.
> 
> Yeah, I've been investigating the rebuild_rc failure. It appears that the progress ULT on the system xstream of rank 1 was blocked by some Argobots lock or eventual, because that xstream had not been able to receive any RPCs since, while it could still run other ULTs.

This turns out to be a deadlock regression. I've added a fix commit "Fix URI lookup deadlocks".

src/cart/crt_context.c
liw: It does have some downside: I guess we didn't track timeouts for waiting requests because we wanted to avoid timeouts due to local waits. I'm also a little concerned if we will see such timeouts as a result of this patch.

On the other hand, tracking timeouts for waiting requests makes it possible for users to limit the time spent on waiting for an RPC. For example, in the future, if we track the deadline of an incoming RPC in an engine, and the RPC handler needs to send an other RPC to a different engine, then we may want to get a -DER_TIMEDOUT for the outgoing RPC before the deadline of the incoming RPC.

Personally, I'm not sure which way is _definitely_ better, and would be happy to revert this change if you guys prefer the current behavior.

frostedcmos (commented):
looks good to me, just need to go more thoroughly again to understand all the changes. mostly minor things inline

liw:
> Test stage Functional Hardware Medium UCX Provider completed with status FAILURE. https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-11030/20/display/redirect

Looks like some `scp` error caused this failure...

frostedcmos:
"Start the RPC timer before crt_req_send returns, rather than
afterwards when the queued RPC moves to the inflight queue, so
that the an RPC is always in crt_context.cc_bh_timeout."

I am concerned about this particular change, as the timer now starts for all wait-queued rpcs as well. Let's assume scenario where server takes 6 seconds to process an RPC, timeout is set to 10, and inflight depth is 1. Client asynchronously submitting two rpcs to the same target will now result in a second rpc timing out, while previous behavior would allow both rpcs to complete.  At the same time client submitting 2 rpcs synchronously would still work.

src/cart/crt_context.c
liw: @frostedcmos also commented on this change. Since everyone, including myself, turns out to be a bit concerned, I'm working on restoring the previous behavior.

src/cart/crt_hg.c
```diff
@@ -1318,7 +1314,7 @@ crt_hg_req_send(struct crt_rpc_priv *rpc_priv)
 		}
 		rpc_priv->crp_state = RPC_STATE_FWD_UNREACH;
 	} else {
-		rpc_priv->crp_on_wire = 1;
+		rpc_priv->crp_state = RPC_STATE_REQ_SENT;
```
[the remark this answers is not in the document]
liw: Yes, this is the only place where `crp_state` becomes `REQ_SENT`. With the addition of `crp_mutex`, we no longer need to set `REQ_SENT` before calling `crt_hg_req_send`, and thus no longer need the `crp_on_wire` flag.

src/cart/crt_context.c
```diff
@@ -458,6 +460,64 @@ crt_rpc_complete(struct crt_rpc_priv *rpc_priv, int rc)
...
 }
 
+/* Array of pointers */
+struct pointers {
+	void		**p_buf;
+	uint32_t	  p_cap;
+	uint32_t	  p_len;
+};
+
+static int
```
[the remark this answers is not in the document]
liw: Moved to gurt as d_vec_pointers.

src/cart/crt_context.c
```diff
@@ -458,6 +460,64 @@ crt_rpc_complete(struct crt_rpc_priv *rpc_priv, int rc)
...
+pointers_append(struct pointers *pointers, void *pointer)
+{
+	if (pointers->p_len == pointers->p_cap) {
+		void		**buf;
+		uint32_t	  cap;
+
+		if (pointers->p_cap == 0)
+			cap = 8;
```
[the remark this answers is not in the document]
liw: This is not a magic number; its meaning is clear. Moreover, the number is only referenced here once. I prefer not to introduce a macro scenarios like this.

src/cart/crt_context.c
```diff
@@ -458,6 +460,64 @@ crt_rpc_complete(struct crt_rpc_priv *rpc_priv, int rc)
...
+		else
+			cap = 2 * pointers->p_cap;
```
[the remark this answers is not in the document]
liw: AFAIK doubling is the most common strategy.

src/cart/crt_context.c
```diff
@@ -710,36 +803,83 @@ int
...
+	struct pointers		 epis;
+	int			 i;
 
-	D_RWLOCK_RDLOCK(&crt_gdata.cg_rwlock);
+	/*
+	 * To avoid acquiring cg_rwlock, cc_mutex, and epi_mutex in wrong
+	 * orders, we hold at most one of them at a time.
+	 */
+
+	rc = pointers_init(&ctxs, 8 /* cap */);
```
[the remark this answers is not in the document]
liw: The number depends on the type of the objects we store in this instance of the vector; I'm afraid using a common macro might discourage people from considering what a suitable number should be for each vector.

src/cart/crt_context.c
liw: Restored to the previous behavior (i.e., not starting timer until inflight).

liw:
- Restored the previous timeout semantics (i.e., RPC timers do not start until RPCs are in flight).
- Released crp_mutex before calling RPC completion callback, to avoid deadlock concerns. Renamed code that locks/unlocks crp_mutex to maintain readability: crt_rpc_lock, crt_rpc_unlock, crt_rpc_complete_and_unlock.
- Addressed Alex's comments.

frostedcmos (changes requested):
Need a bit more time to review but in a meantime, change in context_destroy (noted inline) is needed.

liw (commented):
Thanks, Alex.

src/cart/crt_rpc.c
```diff
@@ -1344,14 +1349,22 @@ crt_req_send_internal(struct crt_rpc_priv *rpc_priv)
...
+			if (rc == 0)
+				D_GOTO(send_immediately, rc);
+			else
 				D_ERROR("crt_req_hg_addr_lookup() failed, "
 					"rc %d, opc: %#x.\n", rc, req->cr_opc);
 		}
 		break;
 	case RPC_STATE_ADDR_LOOKUP:
+send_immediately:
 		rc = crt_req_send_immediately(rpc_priv);
```
frostedcmos, line 1361: not sure i like the label in the middle of the case statement. can we just create the outside label and jump there instead?

frostedcmos (commented):
Looks good, 2 comments inline

src/cart/crt_context.c
```diff
@@ -635,39 +700,9 @@ crt_context_destroy(crt_context_t crt_ctx, int force)
...
-
-	crt_provider_put_ctx_idx(ctx->cc_primary, provider, ctx->cc_idx);
-	d_list_del(&ctx->cc_link);
-
-	D_RWLOCK_UNLOCK(&crt_gdata.cg_rwlock);
-
-	D_MUTEX_DESTROY(&ctx->cc_mutex);
-	D_DEBUG(DB_TRACE, "destroyed context (idx %d, force %d)\n", ctx->cc_idx, force);
-	D_FREE(ctx);
+	CTX_DECREF(ctx);
```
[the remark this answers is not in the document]
liw: I don't have a good idea, only an average one: Instead of calling `CTX_DEFREF(ctx)`, we wait for the reference count of `ctx` to drop to 1, then call `crt_context_free`.

src/cart/crt_context.c
```diff
@@ -635,39 +700,9 @@ crt_context_destroy(crt_context_t crt_ctx, int force)
...
-		D_ERROR("destroy context (idx %d, force %d), "
-			"d_hash_table_destroy_inplace failed, rc: %d.\n",
-			ctx->cc_idx, force, rc);
-		if (!force)
-			D_GOTO(err_unlock, rc);
-	}
-
-	d_binheap_destroy_inplace(&ctx->cc_bh_timeout);
-
 	D_MUTEX_UNLOCK(&ctx->cc_mutex);
```
liw, line 703: Noticed a bug in this patch: `crt_context_destroy` acquires `cc_mutex` and `epi_mutex` in the wrong order, and `crt_ctx_epi_abort` should not release and acquire `cc_mutex`.

liw: Fixed.

src/cart/crt_context.c
```diff
@@ -616,45 +606,73 @@ crt_ctx_epi_abort(d_list_t *rlink, void *arg)
...
+		rc = crt_ctx_epi_abort(epi, flags);
+		if (rc != 0)
+			break;
+	}
+
+out_epis:
+	for (i = 0; i < epis.p_len; i++) {
+		struct crt_ep_inflight *epi = epis.p_buf[i];
+
+		d_hash_rec_decref(&ctx->cc_epi_table, &epi->epi_link);
```
frostedcmos, line 658: should this be wrapped in D_MUTEX_LOCK/UNLOCK on ctx->cc_mutex?

src/cart/crt_context.c
```diff
@@ -762,19 +808,26 @@ crt_rank_abort(d_rank_t rank)
 	 * orders, we hold at most one of them at a time.
 	 */
 
+	/*
+	 * Acquiring crt_context_destroy_lock ensures that the contexts in ctxs
+	 * will not be destroyed. We don't block if the lock is busy, because a
+	 * crt_context_destroy call may take the lock for quite some time.
+	 */
+	rc = pthread_rwlock_trywrlock(&crt_context_destroy_lock);
```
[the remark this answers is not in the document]
liw: Added D_RWLOCK_TRYWRLOCK.

src/cart/crt_context.c
liw: Indeed. Fixed.

frostedcmos (approved):

liuxuezhao (approved):

PR #11036, opened 2022-12-13
DAOS-12250 control: Improve debug logging

mjmac:
Remove boilerplate from gRPC handlers by logging
request and result automatically via an interceptor.

Required-githooks: true

### DIFF
--- a/src/control/server/interceptors.go
+++ b/src/control/server/interceptors.go
@@ -9,6 +9,7 @@ package server
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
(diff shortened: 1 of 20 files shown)


src/control/server/interceptors.go
```diff
@@ -205,3 +209,52 @@ func unaryStatusInterceptor(ctx context.Context, req interface{}, info *grpc.Una
...
+			}
+		}
+
+		// Log the unwrapped error if it's not a sentinel error.
+		if logErr != nil {
+			if !isSentinelErr(logErr) {
+				log.Errorf("gRPC handler for %T failed: %s (elapsed: %s)", req, logErr, elapsed)
+			}
+			return res, err
+		}
```
tanabarr, line 253:
```suggestion
		} else {
		        // Log the unwrapped error if it's not a sentinel error.
			log.Errorf("gRPC handler for %T failed: %s (elapsed: %s)", req, logErr, elapsed)
		}
		if logErr != nil {
			return res, err
		}
```

tanabarr (approved):
Nice change, good cleanup!

src/control/server/interceptors.go
mjmac: To me, it's clearer the other way. The intent is to avoid logging the request at all if it results in a sentinel error (e.g. ErrNotReplica), because the control client's leader/replica resolution mechanism will produce a lot of log spam otherwise. I expect that this will continue to evolve as we evaluate the logging going forward, though.

PR #11041, opened 2022-12-14
DAOS-623 build: Update the go builder and tweak commit hooks.

ashleypittman:
Required-githooks: true

### DIFF
--- a/site_scons/site_tools/go_builder.py
+++ b/site_scons/site_tools/go_builder.py
@@ -17,8 +17,9 @@ def _scan_go_file(node, env, _path):
 
     src_dir = os.path.dirname(str(node))
     includes = []
-    rc = subprocess.run([env.d_go_bin, 'list', '--json', os.path.basename(str(node))],
-                        cwd=src_dir, stdout=subprocess.PIPE, check=True)
+    path_name = str(node)[12:]
+    rc = subprocess.run([env.d_go_bin, 'list', '--json', '-mod=vendor', path_name],
+                        cwd='src/control', stdout=subprocess.PIPE, check=True)
     data = json.loads(rc.stdout.decode('utf-8'))
     for dep in data['Deps']:
         if not dep.startswith('github.com/daos-stack/daos'):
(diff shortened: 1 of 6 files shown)


site_scons/site_tools/go_builder.py
```diff
@@ -3,20 +3,32 @@
...
 
-def _scan_go_file(node, _env, _path):
+def _scan_go_file(node, env, _path):
     """Scanner for go code"""
-    contents = node.get_contents()
+
     src_dir = os.path.dirname(str(node))
     includes = []
+    rc = subprocess.run([env.d_go_bin, 'list', '--json', os.path.basename(str(node))],
+                        cwd=src_dir, capture_output=True, check=True)
```
mjmac, line 21: Have you timed a build with this and without? Unless I'm misunderstanding, this will call `go list --json` for every single .go file in our tree, right? Seems like a lot of overhead.

ashleypittman: that was a concern.  It does however only call it for files that have changed so even if the first build is a little slower subsequent ones will see little difference.

ashleypittman: Ah, it doesn't work in CI as this is causing it to download some dependencies.  Needs more thinking about.

ashleypittman: Ok, I got this working.  Looking at the Jenkins logs there doesn't seem to be a measurable difference.

ashleypittman:
Looking at the build log in Jenkins for the "build on el8 job" the configure/startup checks for scons take 22 seconds with this patch and 23 seconds without it so whilst there are clearly a lot of calls to external processes as a result of this PR they don't have an effect on overall build time.  Scons doesn't re-check the dependencies of a file unless that file changes so for developers who aren't modifying go then the re-build time will be entirely unchanged.

jolivier23 (approved):

mjmac (approved):
LGTM. Tested it locally and it seems to pick things up correctly.

jolivier23 (approved):

PR #10891, opened 2022-11-23
DAOS-10578 control: Add daos_server config generate command

tanabarr:
Decouple logic that scans network and storage remotely across multiple
hosts from the logic that generates the output config from a single
storage and network set. Then expose internal config generation logic as
part of the Control API so it can be used in cmd/daos_server directly
and implement config generate functionality in cmd/daos_server using the
Control API with storage and network sets gathered from the locally
running process.

- Fix regression in HostFabricInterfaces where interface/provider
  combination priority was not returned (DAOS-12274).
- Implement daos_server config generate command which will generate
  file for localhost only (DAOS-10578).
- Improve algorithm for calculating number of worker and service threads
  (DAOS-10272).

Required-githooks: true

Signed-off-by: Tom Nabarro <tom.nabarro@intel.com>
Co-authored-by: Kris Jacque <kristin.jacque@intel.com>

### Before requesting gatekeeper:

* [x] Two review approvals and any prior change requests have been resolved.
* [x] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [x] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [x] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [x] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate watchers.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
[...]

PR #11047, opened 2022-12-15
DAOS-12274 control: Detect fabric provider priority

kjacque:
For the time being, the priority is determined by the order returned from the underlying fabric library for libfabric. By default all other fabric providers (except ucx+tcp) are highest priority (0).

- Add ability to harvest a priority for each FabricInterface provider in lib/hardware.
- Ensure the priority is included for various network scan commands (daos_agent, daos_server, dmg).

Features: control

### Gatekeeper:


src/control/lib/hardware/fabric.go
```diff
@@ -136,7 +314,7 @@ func (s fabricInterfaceMap) update(name string, fi *FabricInterface) {
 		// always possible to add to providers or net interfaces
 		if fi.Providers != nil {
 			if cur.Providers == nil {
-				cur.Providers = common.NewStringSet()
+				cur.Providers = newTestFabricProviderSet()
```
tanabarr, line 317: as this is just creating an empty set, use NewFabricProviderSet() and remove newTestFabricProviderSet() function completely

src/control/lib/hardware/fabric.go
```diff
@@ -136,7 +314,7 @@ func (s fabricInterfaceMap) update(name string, fi *FabricInterface) {
...
 			}
 			cur.Providers.Add(fi.Providers.ToSlice()...)
```
tanabarr, line 319: Nice, so the FabricInterface will end up with a provider list where each provider in that list will keep the highest priority (lowest value) that has been encountered in iterating through fi_info output. There might be an edge case where `cxi+tcp` gets the same priority as a more performant provider on an interface on NUMA-1 (i.e. `ofi+verbs` on ib1 has priority 1 which would be the same as `cxi+tcp` on the same interface), but we can probably deal with this as a follow-on.

tanabarr (changes requested):
Looks like a nice solution, just need to remove the unnecessary function `newTestFabricProviderSet()`. The edge case I alluded to in the comments can be addressed at a later date.

src/control/lib/hardware/fabric.go
kjacque: Good catch, there was a bit of rename shuffling here. I'll fix this.

src/control/lib/hardware/fabric.go
kjacque: Ah, I can push ucx+tcp lower if you like, although if we are using ucx, and it's possible for ofi to use verbs, I would expect ucx+dc/rc/ud to be available as well (and at highest priority so UCX would be preferred). One thing that is tricky is adjusting in relation to the ofi providers since they're in entirely different packages with internals hidden from each other.

kjacque: Adjusted the priority down. Let's see if this causes any issues.

src/control/lib/hardware/fabric.go
```diff
@@ -22,6 +22,184 @@ import (
...
+	for _, p := range providers {
+		set.Add(p)
+	}
+
+	return set
+}
+
+// newTestFabricProviderSet creates a new set of FabricProviders. The priority is derived from the order
+// of the provider strings.
+func newTestFabricProviderSet(providers ...string) *FabricProviderSet {
```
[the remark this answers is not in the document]
kjacque: Moved into test code.

src/control/cmd/daos_agent/network.go
```diff
@@ -83,12 +83,13 @@ func fabricInterfaceSetToHostFabric(fis *hardware.FabricInterfaceSet, filterProv
 
 		for _, name := range netIFs.ToSlice() {
```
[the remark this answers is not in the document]
kjacque: Fixed.

tanabarr (approved):

knard-intel (approved):
LGTM

kjacque:
Merged into #10891

PR #11050, opened 2022-12-15
DAOS-11040 common: add lmdb backend kv for SMD

wangshilong:
LMDB is added as new backend kv store for sys DB, this new backend sys DB will be used to store SMD for MD-ON-SSD case.

For PMDK case, SMD will still be stored in PMEM for backward compatibility purpose.

Required-githooks: true

### DIFF
--- a/src/bio/bio_xstream.c
+++ b/src/bio/bio_xstream.c
@@ -199,7 +199,8 @@ set_faulty_criteria(void)
 
 int
 bio_nvme_init(const char *nvme_conf, int numa_node, unsigned int mem_size,
-	      unsigned int hugepage_size, unsigned int tgt_nr, bool bypass_health_collect)
+	      unsigned int hugepage_size, unsigned int tgt_nr, struct sys_db *db,
+	      bool bypass_health_collect)
 {
 	char		*env;
 	int		 rc, fd;
@@ -290,6 +291,12 @@ bio_nvme_init(const char *nvme_conf, int numa_node, unsigned int mem_size,
 	D_INFO("Set per-xstream DMA buffer upper bound to %u %uMB chunks\n",
 	       bio_chk_cnt_max, size_mb);
 
+	rc = smd_init(db);
+	if (rc != 0) {
+		D_ERROR("Initialize SMD store failed. "DF_RC"\n", DP_RC(rc));
+		goto free_cond;
+	}
+
 	spdk_bs_opts_init(&nvme_glb.bd_bs_opts, sizeof(nvme_glb.bd_bs_opts));
 	nvme_glb.bd_bs_opts.cluster_sz = DAOS_BS_CLUSTER_SZ;
 	nvme_glb.bd_bs_opts.num_md_pages = DAOS_BS_MD_PAGES;
@@ -310,13 +317,15 @@ bio_nvme_init(const char *nvme_conf, int numa_node, unsigned int mem_size,
 	rc = bio_spdk_env_init();
 	if (rc) {
 		nvme_glb.bd_nvme_conf = NULL;
-		goto free_cond;
+		goto fini_smd;
 	}
 	bio_spdk_inited = true;
 	set_faulty_criteria();
 
 	return 0;
 
+fini_smd:
+	smd_fini();
 free_cond:
 	ABT_cond_free(&nvme_glb.bd_barrier);
 free_mutex:
@@ -1181,24 +1190,6 @@ alloc_xs_blobstore(void)
 	return bxb;
 }
 
-static inline struct bio_bdev *
-find_sys_bdev(unsigned int role)
-{
-	struct bio_bdev	*d_bdev;
-
-	D_ASSERT(!d_list_empty(&nvme_glb.bd_bdevs));
-	/*
-	 * Temporarily use the first meta/WAL device, needs be improved when we support
-	 * meta/WAL device hotplug.
-	 */
-	d_list_for_each_entry(d_bdev, &nvme_glb.bd_bdevs, bb_link) {
-		if (is_role_match(d_bdev->bb_roles, role))
-			return d_bdev;
-	}
-
-	return NULL;
-}
-
 static int
 assign_roles(struct bio_bdev *d_bdev, unsigned int tgt_id)
 {
@@ -1242,16 +1233,6 @@ assign_xs_bdev(struct bio_xs_context *ctxt, int tgt_id, enum smd_dev_type st,
 	int			 rc;
 
 	*dev_state = SMD_DEV_NORMAL;
-	if (tgt_id == BIO_SYS_TGT_ID || tgt_id == BIO_STANDALONE_TGT_ID) {
-		d_bdev = find_sys_bdev(dev_type2role(st));
-		if (d_bdev == NULL) {
-			D_ERROR("Failed to find sys bdev\n");
-			return NULL;
-		}
-		/* Assume the sys device in NORMAL state for this moment */
-		return d_bdev;
-	}
-
 	rc = smd_dev_get_by_tgt(tgt_id, st, &dev_info);
 	if (rc == -DER_NONEXIST) {
 		d_bdev = choose_device(tgt_id, st);
--- a/src/bio/bio_xstream.c
+++ b/src/bio/bio_xstream.c
@@ -1246,8 +1246,24 @@ assign_xs_bdev(struct bio_xs_context *ctxt, int tgt_id, enum smd_dev_type st,
 			D_ERROR("Failed to assign roles. "DF_RC"\n", DP_RC(rc));
 			return NULL;
 		}
-
-		d_bdev->bb_tgt_cnt++;
+		/*
+		 * Now a device will be assigned to SYS_TGT_ID for RDB
+		 * (the mapping will be recorded in target table), but we should not
+		 * treat the SYS_TGT mapping equally with other VOS targets mappings.
+		 *
+		 * Let's take an example, if there is a config having 4 meta SSDs and 3 targets,
+		 * how should we assign SSDs?
+		 *
+		 * 1. Assign 3 SSDs to 3 VOS targets and sys target (sys target share SSD with
+		 * one of VOS target), leave one SSD unused, or;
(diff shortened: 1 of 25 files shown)


src/engine/srv.c
```diff
@@ -1242,11 +1247,40 @@ dss_srv_fini(bool force)
...
 
+static int
+dss_sys_db_init(struct sys_db **db)
+{
+	int	 rc;
+	bool	 md_on_ssd_enabled;
+	/* walkaround*/
+	char	*lmm_db_path = getenv("DAOS_LMM_DB_PATH");
+
+	d_getenv_bool("DAOS_MD_ON_SSD", &md_on_ssd_enabled);
```
NiuYawei, line 1258: Is this a temporary workaround? I think we'd decide using vos sysdb when system is in old format, right? (We could tell it via the parameters passed from control plane? Need to confirm this with control plane team).

src/gurt/hash.c
```diff
@@ -947,6 +947,9 @@ d_hash_table_destroy_inplace(struct d_hash_table *htable, bool force)
 	uint32_t		 i;
 	int			 rc = 0;
 
+	if (htable->ht_buckets == NULL)
+		return 0;
```
NiuYawei, line 951: What's the change for?

src/bio/bio_xstream.c
```diff
@@ -1177,24 +1186,6 @@ alloc_xs_blobstore(void)
 	return bxb;
 }
 
-static inline struct bio_bdev *
-find_sys_bdev(unsigned int role)
```
NiuYawei, line 1181: I think all other special handling code for sysdb VOS in BIO and VOS should be reverted too, do you plan to make it in a follow-on patch?

src/common/sys_lmdb.c
```diff
@@ -0,0 +1,511 @@
...
+
+static int
+lmm_db_fetch(struct sys_db *db, char *table, d_iov_t *key, d_iov_t *val)
+{
+	MDB_val			 db_key, db_data;
+	struct lmm_sys_db	*ldb = db2lmm(db);
+	int			 rc;
+	bool			 end_tx = false;
+
+	if (ldb->db_txn == NULL) {
```
NiuYawei, line 233: This could be racy if multiple threads calls lmm_db_fetch() concurrently, if you assume 'db_txn' accessing is always protected by 'lmm_db_lock()', you'd add an assert here.

BTW, does lmdb support nested transaction? If it does, we could simplify it a bit.

src/common/sys_lmdb.c
```diff
@@ -0,0 +1,511 @@
...
+txn_abort:
+	mdb_txn_abort(ldb->db_txn);
+out:
+	ldb->db_txn = NULL;
+	return rc;
+}
+
+#define MAX_SMD_TABLE_LEN	32
+static int
+lmm_db_generate_key(char *table, d_iov_t *key, MDB_val *db_key)
```
NiuYawei, line 186: Instead of generating new keys in this way, what about using multiple dbs? (MDB_dbi per table, can single lmdb file support multiple MDB_dbi?)

src/common/sys_lmdb.c
```diff
@@ -0,0 +1,511 @@
...
+		D_ERROR("mismatch value for table: %s, expected: %lu, got: %lu\n",
+			table, val->iov_len, db_data.mv_size);
+		rc = -DER_MISMATCH;
+		goto out;
+	}
+	memcpy(val->iov_buf, db_data.mv_data, db_data.mv_size);
+
+out:
+	if (end_tx) {
+		mdb_txn_abort(ldb->db_txn);
```
NiuYawei, line 261: Abort even fetch is successful?

src/common/sys_lmdb.c
```diff
@@ -0,0 +1,511 @@
...
+		if (rc)
+			break;
+	}
+	/* reach end */
+	if (rc == MDB_NOTFOUND)
+		rc = 0;
+
+	mdb_cursor_close(cursor);
+tx_end:
+	mdb_txn_abort(ldb->db_txn);
```
NiuYawei, line 367: Always abort tx?

src/engine/srv.c
wangshilong: Yup, this is a walkaround for me to testing.

src/gurt/hash.c
wangshilong: oh, a bug that hash table might be freed twice bug when i tried to use multiple DBs, but i found multiple DBs could not work, then switch back, but did not remove it.

src/common/sys_lmdb.c
wangshilong: I tried, it almost work, but when i killed daos server and restart, data lost. i see docs:

"Database names are keys in the unnamed database, and may be read but not written."

then i reverted to this way, it could work anyway.

src/common/sys_lmdb.c
wangshilong: It supported nested transaction, but when starting nested transaction, mdb_txn_begin must be passed parent handle, so it did not actually simplify anything even it support..

src/common/sys_lmdb.c
wangshilong: Yup, a bit strange. there are APIs:

mdb_txn_begin
mdb_txn_commit
mdb_txn_abort
mdb_txn_reset
mdb_txn_renew

No APIs like mdb_txn_end()..i search some example in the website, they used mdb_txn_abort() for read operations.

src/engine/srv.c
NiuYawei: Ok. Let's make it in a follow-on patch.

I think we might need to parse the passed in daos_nvme.conf to see if meta SSD is configured, see dss_xstream_has_nvme().

src/bio/bio_xstream.c
```diff
@@ -1242,16 +1233,6 @@ assign_xs_bdev(struct bio_xs_context *ctxt, int tgt_id, enum smd_dev_type st,
...
-	if (tgt_id == BIO_SYS_TGT_ID || tgt_id == BIO_STANDALONE_TGT_ID) {
-		d_bdev = find_sys_bdev(dev_type2role(st));
-		if (d_bdev == NULL) {
-			D_ERROR("Failed to find sys bdev\n");
-			return NULL;
-		}
-		/* Assume the sys device in NORMAL state for this moment */
-		return d_bdev;
-	}
-
```
NiuYawei, line 1254: Now a device will be assigned to SYS_TGT_ID for RDB (the mapping will be recorded in target table), but I don't think we should treat the SYS_TGT mapping equally with other VOS targets mappings.

Let's take an example, if there is a config having 4 meta SSDs and 3 targets, how should we assign SSDs?

1. Assign 3 SSDs to 3 VOS targets and sys target (sys target share SSD with one of VOS target), leave one SSD unused, or;
2. Assign 1 SSD to sys target, assign the other 3 SSDs to VOS targets;

I think we'd use the 1st policy to assign SSDs, right?

src/common/sys_lmdb.c
```diff
@@ -0,0 +1,517 @@
...
+}
+
+#define MAX_SMD_TABLE_LEN	32
+static int
+lmm_db_generate_key(char *table, d_iov_t *key, MDB_val *db_key)
+{
+	char	*new_key;
+	int	 table_len;
+
+	table_len = strnlen(table, MAX_SMD_TABLE_LEN);
```
NiuYawei, line 196: Shouldn't we return -DER_INVAL when table length is larger than MAX_SMD_TABLE_LEN?

src/common/sys_lmdb.c
```diff
@@ -0,0 +1,517 @@
...
+		mdb_dbi_close(lmm_db.db_env, lmm_db.db_dbi);
+		free(lmm_db.db_file);
+	}
+
+	free(lmm_db.db_path);
+	memset(&lmm_db, 0, sizeof(lmm_db));
+}
+
+int
+lmm_db_init_ex(const char *db_path, const char *db_name, bool force_create, bool destroy_db_on_fini)
```
NiuYawei, line 441: Maybe we could improve the src/bio/smd/tests/smd_ut.c to make the backend sysdb configurable (specify backend sysdb by argument)? So that we could run SMD unit tests over lmdb based sysdb.

src/bio/bio_xstream.c
wangshilong: sure.

src/bio/smd/smd_internal.h
```diff
@@ -16,6 +16,7 @@
 #include <abt.h>
 #include <daos_types.h>
 #include <daos/btree.h>
+#include <daos/sys_db.h>
 #include <daos/mem.h>
 #include <daos_srv/smd.h>
 #include <daos_srv/daos_engine.h>	/* Revise the potential layering issue later */
```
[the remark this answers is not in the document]
wangshilong: removed smd_internal.h, removed daos_engine.h from bio_internal.h hits some build failure, kept that for now.

src/bio/bio_xstream.c
```diff
@@ -1265,8 +1246,24 @@ assign_xs_bdev(struct bio_xs_context *ctxt, int tgt_id, enum smd_dev_type st,
...
+		 * 1. Assign 3 SSDs to 3 VOS targets and sys target (sys target share SSD with
+		 * one of VOS target), leave one SSD unused, or;
+		 * 2. Assign 1 SSD to sys target, assign the other 3 SSDs to VOS targets
+		 *
+		 * We use the 1st policy to assign SSDs and @bb_tgt_cnt won't be increased for
+		 * sys tgt id.
+		 *
+		 */
+		if (tgt_id != BIO_SYS_TGT_ID)
+			d_bdev->bb_tgt_cnt++;
```
NiuYawei, line 1266: The 'bb_tgt_cnt' is bit messy now, I'm wondering if we should cleanup it now.

NiuYawei (approved):

wangshilong:
> Test stage Functional Hardware Medium completed with status FAILURE. https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-11050/13/execution/node/905/log

Failures are known as DAOS-12277 and DAOS-12309

PR #11052, opened 2022-12-15
DAOS-11390 test: Add NLT/ftest support for control_metadata

mjmac:
Enable use of an external directory for control metadata by
default. Eventually we'll want to add support for testing with
a dedicated control metadata device with hw tests, as well
as fallback tests for the old SCM-based metadata mode.

Features: control
Required-githooks: true


src/tests/ftest/util/server_utils.py
```diff
@@ -230,6 +230,20 @@ def clean_files(self, verbose=True):
...
+            # Remove the control metadata directory
+            self.log.info("Cleaning up the %s directory.", str(path))
+            clean_commands.append("sudo rm -fr {}".format(path))
+
+        if self.manager.job.yaml.control_metadata.device:
+            device = self.manager.job.yaml.control_metadata.device
+            # Clean up the control metadata device
+            self.log.info("Cleaning up the %s device.", str(device))
+            clean_commands.append("sudo umount {}".format(device))
+            clean_commands.append("sudo wipefs -a {}".format(device))
```
kjacque, line 245: You'll probably want to reverse the order of these, since the path is the mount point for the device.

mjmac: Oh, I typoed the first one. I was thinking that we should `rm -fr $path/*` to nuke the contents but leave the directory itself alone.

mjmac: Thought about it some more and decided that you're right. Best to unmount/wipe first if there's a device, then just nuke the directory. I was thinking that we might want to preserve the directory but nuke the contents, but I think ultimately we'll want to ensure that the directory gets recreated by the control plane.

kjacque (approved):

src/tests/ftest/util/server_utils.py
```diff
@@ -230,6 +230,20 @@ def clean_files(self, verbose=True):
             verbose (bool, optional): display clean commands. Defaults to True.
         """
         clean_commands = []
+
+        if self.manager.job.yaml.control_metadata.device:
```
daltonbohning, line 234: Nothing in ftest is yet using `device`, right?

mjmac: Right. This is for the vos_on_blob branch, where we need at least the external directory for control metadata to be working soon. The plan is to eventually start testing with the control metadata device as well.

daltonbohning (approved):

PR #11066, opened 2022-12-16
DAOS-12279 control: Remove support for config-specified ranks

mjmac:
The ranks assigned to engines will be determined by the DAOS
Management Service. The old rank parameter in the engine
configuration section was a legacy remnant of pre-Control Plane
days and is a source of confusion and errors for new users.

Features: control

Required-githooks: true

### DIFF
--- a/src/control/server/config/server_test.go
+++ b/src/control/server/config/server_test.go
@@ -240,7 +240,6 @@ func TestServerConfig_Constructed(t *testing.T) {
 		engine.MockConfig().
 			WithSystemName("daos_server").
 			WithSocketDir("./.daos/daos_server").
-			WithRank(0).
 			WithTargetCount(16).
 			WithHelperStreamCount(4).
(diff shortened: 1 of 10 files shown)


ravalsam (approved):

tanabarr (commented):
Are we really sure that we don't need to be storing the rank in the superblock anymore @mjmac ? Maybe we just want to stop it being read from the config file instead of removing it from the config structure completely?

mjmac:
> Are we really sure that we don't need to be storing the rank in the superblock anymore @mjmac ? Maybe we just want to stop it being read from the config file instead of removing it from the config structure completely?

We are still storing the rank in the superblock: https://github.com/daos-stack/daos/blob/master/src/control/server/instance.go#L215

This PR just removes the ability to set the rank manually.

knard-intel (approved):
LGTM

PR #11090, opened 2022-12-20
DAOS-12277 test: Convert suite to use new-style dmg helpers

mjmac:
Use the new-style dmg helper to perform all pool operations.
Avoids malloc/RDMA problems as described in DAOS-10301.

### DIFF
--- a/src/common/tests_dmg_helpers.c
+++ b/src/common/tests_dmg_helpers.c
@@ -750,6 +750,123 @@ out:
 	return rc;
 }
 
+static int
+dmg_pool_target(const char *cmd, const char *dmg_config_file, const uuid_t uuid,
+		const char *grp, d_rank_t rank, int tgt_idx)
+{
+	char			uuid_str[DAOS_UUID_STR_SIZE];
+	int			argcount = 0;
+	char			**args = NULL;
+	struct json_object	*dmg_out = NULL;
+	int			rc = 0;
+
+	uuid_unparse_lower(uuid, uuid_str);
+	args = cmd_push_arg(args, &argcount, "%s ", uuid_str);
+	if (args == NULL)
+		D_GOTO(out, rc = -DER_NOMEM);
+
+	if (grp != NULL) {
+		args = cmd_push_arg(args, &argcount, "--sys=%s ", grp);
+		if (args == NULL)
+			D_GOTO(out, rc = -DER_NOMEM);
+	}
+
+	if (tgt_idx >= 0) {
+		args = cmd_push_arg(args, &argcount, "--target-idx=%d ", tgt_idx);
+		if (args == NULL)
+			D_GOTO(out, rc = -DER_NOMEM);
+	}
+
+	args = cmd_push_arg(args, &argcount, "--rank=%d ", rank);
+	if (args == NULL)
+		D_GOTO(out, rc = -DER_NOMEM);
+
+	rc = daos_dmg_json_pipe(cmd, dmg_config_file,
+				args, argcount, &dmg_out);
+	if (rc != 0) {
+		D_ERROR("dmg failed");
+		goto out_json;
+	}
+
+out_json:
+	if (dmg_out != NULL)
+		json_object_put(dmg_out);
+	cmd_free_args(args, argcount);
+out:
+	return rc;
+}
+
+int
+dmg_pool_exclude(const char *dmg_config_file, const uuid_t uuid,
+		 const char *grp, d_rank_t rank, int tgt_idx)
+{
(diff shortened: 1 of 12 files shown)


wangdi1 (approved):
Thanks! Michael

src/common/tests_dmg_helpers.c
```diff
@@ -750,6 +750,123 @@ dmg_pool_destroy(const char *dmg_config_file, const uuid_t uuid, const char *grp
...
+		if (args == NULL)
+			D_GOTO(out, rc = -DER_NOMEM);
+	}
+
+	rank_list.rl_ranks = ranks;
+	rank_list.rl_nr = rank_nr;
+
+	rank_str = d_rank_list_to_str(&rank_list);
+	if (rank_str == NULL)
+		D_GOTO(out, rc = -DER_NOMEM);
```
kccain, line 848: if rank_str == NULL here, then args will be not NULL and when the code reaches the out label the args will not be freed. So before jumping to the out label perhaps cmd_free_args() could be called here?

src/tests/suite/daos_obj.c
```diff
@@ -3165,9 +3165,9 @@ tgt_idx_change_retry(void **state)
 
 		/** exclude target of the replica */
 		print_message("rank 0 excluding target rank %u ...\n", rank);
```
kccain, line 3167: minor, may be worth changing the print_message to say "excluding rank" rather than "excluding target rank"

src/tests/suite/daos_obj.c
```diff
@@ -3223,8 +3223,9 @@ tgt_idx_change_retry(void **state)
 
 	if (arg->myrank == 0) {
 		print_message("rank 0 adding target rank %u ...\n", rank);
```
kccain, line 3225: minor, may be worth changing the print_message to say "adding rank" rather than "adding target rank"

src/tests/suite/daos_rebuild_common.c
```diff
@@ -79,13 +81,14 @@ rebuild_extend_tgt(test_arg_t **args, int args_cnt, d_rank_t rank,
...
+	int rc = 0;
 
 	for (i = 0; i < args_cnt; i++) {
-		if (!args[i]->pool.destroyed)
-			daos_extend_target(args[i]->pool.pool_uuid,
-					   args[i]->group,
-					   args[i]->dmg_config,
-					   rank, tgt_idx, nvme_size);
+		if (!args[i]->pool.destroyed) {
+			rc = dmg_pool_extend(args[i]->dmg_config, args[i]->pool.pool_uuid,
```
kccain, line 88: I see tgt_idx is now not passed (since dmg pool extend doesn't accept --target-idx argument?)
But I guess the tgt_idx argument to this function is now unused and could be removed here and by the callers?

kccain (changes requested):

src/common/tests_dmg_helpers.c
bfaccini: or simply goto the out_json label ?

mjmac: Oops, good catch. I moved things around a little; hopefully it makes sense now. I guess another option would be to just get rid of the `out_rankstr:` label and put the `D_FREE(rank_str)` under `out:`? It's safe to call that even on a null pointer, I think. Maybe more sloppy that way though.

src/tests/suite/daos_obj.c
mjmac: Done.

src/tests/suite/daos_obj.c
mjmac: Done.

src/tests/suite/daos_rebuild_common.c
kccain: Sounds good to defer on this for now

kccain (approved):

wangdi1 (approved):

src/include/daos/tests_lib.h
```diff
@@ -241,6 +241,58 @@ int dmg_pool_create(const char *dmg_config_file,
 int dmg_pool_destroy(const char *dmg_config_file,
 		     const uuid_t uuid, const char *grp, int force);
 
+/**
+ * Exclude an entire rank or a target on that rank from a pool.
+ *
+ * \param dmg_config_file
+ *			[IN]	DMG config file
```
ashleypittman, line 248: FYI, this annotation style is not doxygen, please don't use it.  We still need to convert existing files like this one to the proper format.

PR #11134, opened 2023-01-04
DAOS-8722 dfuse: Add common init function for inodes and checking to readdir.

ashleypittman:
Create a common inode init function to allow us to have a common code-path.
Add a magic value which we check for in close to ensure init has been called.

Use atomics to check for concurrent readdir operations, both on the handle
and on the inode.

Test-tag: dfuse
Required-githooks: true


mchaarawi (approved):

daltonbohning (approved):

ashleypittman:
Test failures appear to be DAOS-12241 and DAOS-12333

PR #11152, opened 2023-01-06
DAOS-12329 cart: Fix swim suspicion leaks

liw:
See the Jira ticket for the problems. This patch makes three changes:

  - Reset any corresponding suspicion when receiving a suspicion of a newer incarnation. This would have prevented the test failure observed in the Jira ticket.

  - Remove any corresponding suspicion when deleting a SWIM member.

  - Remove the request in server_utils_param.py for an overly short 10 ms SWIM ping timeout. The request was likely resulted from a misunderstanding.

  - Remove the rank removal in swim_crt_event_cb, for doing so in a crt event callback may lead to deadlocks.

Required-githooks: true

### Gatekeeper:


frostedcmos (dismissed):

src/tests/ftest/cart/rpc/swim_notification.yaml
```diff
@@ -5,7 +5,7 @@ ENV:
   default:
     #!filter-only : /run/env_CRT_CTX_SHARE_ADDR/sep
     #!filter-only : /run/tests/proto_np
-    - D_LOG_MASK: "WARN"
+    - D_LOG_MASK: "DEBUG"
```
frostedcmos, line 8: Just to make sure - do we end up generating small enough logs? Previously we changed to WARN due to some cart tests taking 100mb+, filling up CI system

liw: @frostedcmos, here we are changing only the swim_notification test. I had to investigate a failure of this test, and found that the `WARN`-level logs contained way too little info to be of any use. With the mask change to `DEBUG`, we get 3.8 MB logs from this test:

```
liw2@m2 swim_notification.py % du -Ah
2.9M	./cart_logs.wolf-104vm2
 11K	./jobdata
 66K	./sysinfo/post
 66K	./sysinfo/pre
133K	./sysinfo
222K	./test-results/1-._cart_swim_notification.py_CartRpcOneNodeSwimNotificationOnRankEvictionTest.test_cart_rpc_run-ENV-env_CRT_CTX_SHARE_ADDR-no_sep-env_CRT_PHY_ADDR_STR-ofi_tcp-hosts-hosts_1-tests-test_group-d273
223K	./test-results
3.8M	.
```

frostedcmos: Thanks for checking. Yes, a while ago we decided that we want most cart tests to run at warn/info levels until they fail, during which time a developer can switch temporarily to a more verbose mode (e.g. debug) to investigate further. The problem with this approach is that some tests can only fail once in a while, making it harder to investigate, so there is this tradeoff.
 
I am ok with keeping this test always at a debug level if you think this test falls into such category. Or perhaps some compromise with enabling only some subsystems as DEBUG while keeping the rest at info/warn/error.

We just need to be careful of not enabling excessive logging by default, especially once we start running the same tests under multiple providers and different settings (e.g. SEP on/off, multiprovider variation etc)

liw: OK, let's maintain what has been decided then. The failure I had to debug was 100% reproducible. I've reverted this mask change.

soumagne (approved):

src/cart/crt_swim.c
```diff
@@ -1534,6 +1536,10 @@ void crt_swim_rank_del_all(struct crt_grp_priv *grp_priv)
 			SWIM_STATUS_CHARS[cst->cst_state.sms_status],
 			cst->cst_state.sms_incarnation);
 		D_FREE(cst);
+
+		crt_swim_csm_unlock(csm);
```
soumagne, line 1540: so you're unlocking then deleting then locking back ?

liw: @soumagne, that's right. We lock `csm` in order to delete a rank from `crt_swim`-level tracking structures. Then, we want to delete states of this rank in the `swim` level, which requires locking `csm_ctx` (inside `swim_member_del`). Due to lock order considerations, we have to unlock `csm` first.

I think ideally we should remove locking from the `swim` module and rely solely on locking in the `crt_swim` module. Then the lock order considerations here would disappear, and the deletion would become one atomic change. Probably too much for the current problem and patch.

daltonbohning (commented):
ftest LGTM

frostedcmos (approved):

PR #11157, opened 2023-01-06
DAOS-12209 control: Don't clean up Creating pools on MS step-up

mjmac:
When a MS leader steps up, it checks the list of pools for any
not found in a Ready state. With this change, a pool in the
Creating state will be ignored rather than destroyed, on the
theory that the create may be completing on the previous leader.

Update the pool svc RAS event handler to automatically set
a Creating pool to Ready, on the theory that the pool create
has most likely completed, given that it is sending us a
svc ranks update event.

Also adds support for sending the pool service leader in the
PoolCreate response.

Finally, modify dc_mgmt_pool_find() to retry on -DER_NONEXIST
errors which can result from stale MS replica reads.


kccain (dismissed):
first pass review LGTM, consistent with our separate chats on the approach.

tanabarr (approved):

kccain (approved):

PR #11163, opened 2023-01-06
DAOS-12239 control: Check for ErrNotLeader in ResignLeadership

mjmac:
The call to raft.Barrier() was modified to return ErrNotLeader
on leadership loss in order to convey this information back to
dmg, but ResignLeadership() was not updated to include a check
for this error in order to avoid calling raft.LeadershipTransfer()
on a non-leader (which can cause a hang due to blocked channel read).

Required-githooks: true

### DIFF
--- a/src/control/system/raft/mocks.go
+++ b/src/control/system/raft/mocks.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2022 Intel Corporation.
+// (C) Copyright 2020-2023 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
--- a/src/control/system/raft/raft.go
+++ b/src/control/system/raft/raft.go
@@ -100,7 +100,7 @@ func IsRaftLeadershipError(err error) bool {
 // leadership state. No-op if there is only one replica configured
 // or the cause is a raft leadership error.
 func (db *Database) ResignLeadership(cause error) error {
-	if IsRaftLeadershipError(cause) {
+	if system.IsNotLeader(cause) || IsRaftLeadershipError(cause) {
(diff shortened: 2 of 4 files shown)


src/control/system/raft/mocks.go
```diff
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2022 Intel Corporation.
+// (C) Copyright 2020-2023 Intel Corporation.
```
kjacque, line 2: Looks like copyright updated without any other changes

kjacque (approved):
Minor nit, not blocking

tanabarr (approved):

src/control/system/raft/raft.go
```diff
@@ -110,6 +110,13 @@ func (db *Database) ResignLeadership(cause error) error {
 	}
 	db.log.Errorf("resigning leadership (%s)", cause)
 	return db.raft.withReadLock(func(svc raftService) error {
+		// One more belt-and-suspenders check to make sure we're
+		// actually the leader before we try to transfer leadership.
+		// This is important because trying to transfer leadership
+		// if we're not the leader will result in a blocked channel, i.e. hang.
+		if svc.State() != raft.Leader {
+			return nil
```
tanabarr, line 118: don't you want to log the fact that this resign leadership attempt was aborted?

mjmac: There will already be some logging to show that leadership was lost on the node.

src/control/system/raft/mocks.go
mjmac: Oops, I had made a change to that file, committed it, then backed out the changes, and forgot to revert the copyright change.

PR #11165, opened 2023-01-06
DAOS-12309 test: Introduce param to toggle log mast for pool create/destroy

saurabhtandan:
Adding a TestPool param to enable/disable elevating log
mask to DEBUG during pool create/destroy. Currently, this is
done by default for every pool create and destroy dmg call
hence adding significant test run time, causing tests to
timeout in multiple pool scenarios.

Quick-Functional: true
Test-repeat: 10
Test-tag: pool_create_tests test_container_boundary

### DIFF
--- a/src/tests/ftest/pool/create_capacity.py
+++ b/src/tests/ftest/pool/create_capacity.py
@@ -37,7 +37,11 @@ class PoolCreateTests(PoolTestBase):
         # available capacity, e.g. 0.6% for 100 pools.
         quantity = self.params.get("quantity", "/run/pool/*", 1)
         self.add_pool_qty(quantity, create=False)
+        # for multiple pool creation cases, enabling and disabling
+        # logmask setting to DEBUG explicitly to save run time..
+        self.dmg.server_set_logmasks("DEBUG", raise_exception=False)
         self.check_pool_creation(30)
+        self.dmg.server_set_logmasks(raise_exception=False)
 
         # Verify DAOS can be restarted in less than 2 minutes
         try:
--- a/src/tests/ftest/util/test_utils_pool.py
+++ b/src/tests/ftest/util/test_utils_pool.py
@@ -340,13 +340,13 @@ class TestPool(TestDaosApiBase):
         # Create a pool with the dmg command and store its CmdResult
         # Elevate engine log_mask to DEBUG before, then restore after pool create
         self._log_method("dmg.pool_create", kwargs)
-        if self.set_logmasks:
+        if self.set_logmasks is True:
             self.dmg.server_set_logmasks("DEBUG", raise_exception=False)
         try:
             data = self.dmg.pool_create(**kwargs)
             create_res = self.dmg.result
         finally:
-            if self.set_logmasks:
+            if self.set_logmasks is True:
                 self.dmg.server_set_logmasks(raise_exception=False)
 
         # make sure dmg exit status is that of the pool create, not the set-logmasks
@@ -445,7 +445,7 @@ class TestPool(TestDaosApiBase):
 
                 # Destroy the pool with the dmg command.
                 # Elevate log_mask to DEBUG, then restore after pool destroy
-                if self.set_logmasks:
+                if self.set_logmasks is True:
                     self.dmg.server_set_logmasks("DEBUG", raise_exception=False)
                     self.dmg.pool_destroy(pool=self.identifier, force=force, recursive=recursive)
                     self.dmg.server_set_logmasks(raise_exception=False)
--- a/src/tests/ftest/util/test_utils_pool.py
+++ b/src/tests/ftest/util/test_utils_pool.py
@@ -147,8 +147,7 @@ class TestPool(TestDaosApiBase):
         self.quantity = BasicParameter(None, 1)
         self.min_targets = BasicParameter(None, 1)
 
-        # Parameter to control setting and unsetting of log mask
-        # for pool create/destroy
+        # Parameter to control log mask enable/disable for pool create/destroy
         self.set_logmasks = BasicParameter(None, True)
 
         self.pool = None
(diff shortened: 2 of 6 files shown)


shimizukko (dismissed):

src/tests/ftest/pool/create_capacity.py
```diff
@@ -37,7 +37,11 @@ def test_create_pool_quantity(self):
         # available capacity, e.g. 0.6% for 100 pools.
         quantity = self.params.get("quantity", "/run/pool/*", 1)
         self.add_pool_qty(quantity, create=False)
+        # for multiple pool creation cases, enabling and disabling
+        # logmask setting to DEBUG explicitly to save run time..
+        self.dmg.server_set_logmasks("DEBUG", raise_exception=False)
```
kccain, line 42: I think this line would need to be placed before the self.add_pool_qty() call above that will actually be doing the pool creates

kccain (changes requested):
Looks good. I had been doing an experimental change that is not as compact as this test .yaml parameter approach. Can you take a look at what I've done in that experimental PR  https://github.com/daos-stack/daos/pull/11222 - in utils/apricot/apricot/test.py in tearDown though, to see if something similar can be done to make sure that set-logmasks DEBUG can be done once before the tearDown pool destroys, and then set-logmasks again to restore the engines log_mask?

src/tests/ftest/pool/create_capacity.py
saurabhtandan: `self.add_pool_qty(quantity, create=False)` is not creating the pool yet at this stage as we have `create=False`. It is just creating different TestPool objects at this stage for the quantify specified. In the next step i.e `self.check_pool_creation(30)` the pool actually gets created using the multiple TestPool objects and verified if pool creation was completed within the expected 30 sec time limit.

kccain: ah yes, thank you for correcting me on that detail that I missed.

saurabhtandan:
> Looks good. I had been doing an experimental change that is not as compact as this test .yaml parameter approach. Can you take a look at what I've done in that experimental PR #11222 - in utils/apricot/apricot/test.py in tearDown though, to see if something similar can be done to make sure that set-logmasks DEBUG can be done once before the tearDown pool destroys, and then set-logmasks again to restore the engines log_mask?

Thanks for the review. Regarding log mask in teardown() for pool destroy, unfortunately that is not possible at this point as pool destroys get taken care off by the `test.register_cleanup(remove_pool, test=test, pool=pool)` in add_pool(). But I hope the detailed plan mentioned on the ticket by Phil should help understand how can we tackle that issue in future.
But I noticed there was an extra log mask set and unset call being made which you rectified in your PR-11222, under test.py which I incorporated. Thanks for pointing that out.

kccain:
> > Looks good. I had been doing an experimental change that is not as compact as this test .yaml parameter approach. Can you take a look at what I've done in that experimental PR #11222 - in utils/apricot/apricot/test.py in tearDown though, to see if something similar can be done to make sure that set-logmasks DEBUG can be done once before the tearDown pool destroys, and then set-logmasks again to restore the engines log_mask?
> 
> Thanks for the review. Regarding log mask in teardown() for pool destroy, unfortunately that is not possible at this point as pool destroys get taken care off by the `test.register_cleanup(remove_pool, test=test, pool=pool)` in add_pool(). But I hope the detailed plan mentioned on the ticket by Phil should help understand how can we tackle that issue in future. But I noticed there was an extra log mask set and unset call being made which you rectified in your PR-11222, under test.py which I incorporated. Thanks for pointing that out.

OK. PR 112222 does something here but it is probably ugly (when it comes to Python, I am not an artist :)). I guess most of the functional tests (those that do not specify "set_logmasks: false" in their config .yaml file) will still get log_mask: DEBUG for their pool destroys. Which should cover most of our needs for the time being.

saurabhtandan:
> > > Looks good. I had been doing an experimental change that is not as compact as this test .yaml parameter approach. Can you take a look at what I've done in that experimental PR #11222 - in utils/apricot/apricot/test.py in tearDown though, to see if something similar can be done to make sure that set-logmasks DEBUG can be done once before the tearDown pool destroys, and then set-logmasks again to restore the engines log_mask?
> > 
> > 
> > Thanks for the review. Regarding log mask in teardown() for pool destroy, unfortunately that is not possible at this point as pool destroys get taken care off by the `test.register_cleanup(remove_pool, test=test, pool=pool)` in add_pool(). But I hope the detailed plan mentioned on the ticket by Phil should help understand how can we tackle that issue in future. But I noticed there was an extra log mask set and unset call being made which you rectified in your PR-11222, under test.py which I incorporated. Thanks for pointing that out.
> 
> OK. PR 112222 does something here but it is probably ugly (when it comes to Python, I am not an artist :)). I guess most of the functional tests (those that do not specify "set_logmasks: false" in their config .yaml file) will still get log_mask: DEBUG for their pool destroys. Which should cover most of our needs for the time being.

Exactly, that's the intention. It's only handful which might not set it otherwise all will have DEBUG log mask set by default.

kccain (dismissed):

shimizukko (dismissed):
LGTM. There's a typo in PR title.

saurabhtandan:
Failed `container/snapshot_aggregation.py` test failed due to known issue reported at DAOS-12326.

src/tests/ftest/util/test_utils_pool.py
```diff
@@ -147,6 +147,10 @@ def __init__(self, context, dmg_command, cb_handler=None,
         self.quantity = BasicParameter(None, 1)
         self.min_targets = BasicParameter(None, 1)
 
+        # Parameter to control setting and unsetting of log mask
```
daltonbohning, line 150: We shouldn't introduce more spelling errors
```
wrong-spelling-in-comment, Wrong spelling of a word 'unsetting' in a comment:
```
We should either use a different word or add this one to `utils/cq/words.dict`

saurabhtandan: Done. Thanks

daltonbohning (approved):

kccain (approved):

shimizukko (approved):

PR #11193, opened 2023-01-10
DAOS-12354 test: Reduce ior write size to fit all CI clusters

saurabhtandan:
It was observed that some clusters in CI had smaller size nvme dimms than others. Hence the test was passing on some and running out of space on another. So reducing the ior write size by half to satisfy this situation.

Quick-Functional: true
Test-repeat: 10
Test-tag: test_ec_online_rebuild_array


saurabhtandan:
The test was run 5 times on the cluster wolf-[201-203] (which is mentioned below as a smaller cluster). The test did not failed due to no space this time. But for one of the iterations one variant of the test fail exposed a possible seg fault scenario. A separate ticket has been opened for that issue https://daosio.atlassian.net/browse/DAOS-12419

daltonbohning (approved):

shimizukko (approved):

PR #11200, opened 2023-01-11
DAOS-12239 control: Check for ErrNotLeader in ResignLeadership (#11163)

mjmac:
The call to raft.Barrier() was modified to return ErrNotLeader
on leadership lost in order to convey this information back to
dmg, but ResignLeadership() was not updated to include a check
for this error in order to avoid calling raft.LeadershipTransfer()
on a non-leader (which can cause a hang due to blocked channel read).


knard-intel (approved):
LGTM

tanabarr (approved):

kjacque (approved):

PR #11208, opened 2023-01-12
DAOS-12372 mgmt: fix dev query assertion with MD-ON-SSD enabled

wangshilong:
Two fixes:
1. dev query ult should be run on the owner xstream, introduce a new api to return dev's owner xs target id.

2. SYS XS might be owner of device, fix to pass correct XS type when creating ults.

Required-githooks: true

### DIFF
--- a/src/bio/bio_context.c
+++ b/src/bio/bio_context.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2018-2022 Intel Corporation.
+ * (C) Copyright 2018-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -273,6 +273,24 @@ bio_bs_hold(struct bio_blobstore *bbs)
 out:
 	ABT_mutex_unlock(bbs->bb_mutex);
 	return rc;
+}
+
+struct bio_xs_blobstore *
+bio_xs_blobstore_by_devid(struct bio_xs_context *xs_ctxt, uuid_t dev_uuid)
+{
+	enum smd_dev_type	 st;
+	struct bio_xs_blobstore *bxb = NULL;
+
+	for (st = SMD_DEV_TYPE_DATA; st < SMD_DEV_TYPE_MAX; st++) {
+		bxb = xs_ctxt->bxc_xs_blobstores[st];
+		if (!bxb)
+			continue;
+
+		if (uuid_compare(bxb->bxb_blobstore->bb_dev->bb_uuid, dev_uuid) == 0)
+			return bxb;
+	}
+
+	return NULL;
 }
 
 /**
--- a/src/engine/ult.c
+++ b/src/engine/ult.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -329,7 +329,8 @@ sched_ult2xs(int xs_type, int tgt_id)
 {
 	uint32_t	xs_id;
 
-	D_ASSERT(tgt_id >= 0 && tgt_id < dss_tgt_nr);
+	if (xs_type == DSS_XS_VOS || xs_type == DSS_XS_OFFLOAD || xs_type == DSS_XS_IOFW)
+		D_ASSERT(tgt_id >= 0 && tgt_id < dss_tgt_nr);
 	switch (xs_type) {
 	case DSS_XS_SELF:
 		return DSS_XS_SELF;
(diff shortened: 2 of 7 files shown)


src/engine/ult.c
```diff
@@ -329,7 +329,9 @@ sched_ult2xs(int xs_type, int tgt_id)
 {
 	uint32_t	xs_id;
 
-	D_ASSERT(tgt_id >= 0 && tgt_id < dss_tgt_nr);
+	if (xs_type != DSS_XS_SYS)
```
NiuYawei, line 332: either we revert this change and ensure caller always pass 0 tgt_id when the tgt_id is not used, or we change it to "if (xs_type == DSS_XS_VOS || xs_type == DSS_XS_OFFLOAD || xs_type == DSS_XS_IOFW)"

src/mgmt/srv_query.c
```diff
@@ -41,58 +41,60 @@ bs_state_query(void *arg)
...
 
-	bs_arg.bs_arg_type = SMD_DEV_TYPE_DATA;
 	/*
 	 * Query per-server metadata (SMD) to get target ID(s) for given device.
 	 */
 	if (!uuid_is_null(bs_uuid)) {
-		rc = smd_dev_get_by_id(bs_uuid, &dev_info);
-		if (rc != 0) {
-			D_ERROR("Blobstore UUID:"DF_UUID" not found\n",
+		tgt_id = bio_dev_owner_xs_id(bs_uuid);
```
NiuYawei, line 67: The device list in nvme_glb.bd_bdevs is managed by init xstream, it's not thread safe, though the race window is small (the list is only changed on xstream start or hotplug), we just can iterate the list from other xstreams.

I know that there might be existing code having similar problems, maybe we'd use this opportunity to cleanup and fix all potential issues.

src/mgmt/srv_query.c
```diff
@@ -41,58 +41,60 @@ bs_state_query(void *arg)
...
-		/* Default tgt_id is the first mapped tgt */
-		tgt_id = dev_info->sdi_tgts[0];
 	} else {
 		D_ERROR("Blobstore UUID is not provided for state query\n");
 		return -DER_INVAL;
 	}
 
+	xs_type = tgt2xs_type(tgt_id);
+	if (xs_type == DSS_XS_SYS)
+		bs_arg.bs_arg_type = SMD_DEV_TYPE_META;
```
NiuYawei, line 80: What if SCM is configured (not md-on-ssd)? And I don't know why we need device type at all, since the device ID is already provided by user.

src/mgmt/srv_query.c
```diff
@@ -154,54 +156,56 @@ int
...
+
 	D_DEBUG(DB_MGMT, "Querying BIO Health Data for dev:"DF_UUID"\n",
 		DP_UUID(dev_uuid));
 	uuid_copy(mbh->mb_devid, dev_uuid);
 
+	xs_type = tgt2xs_type(tgt_id);
+	if (xs_type == DSS_XS_SYS)
+		mbh->mb_dev_type = SMD_DEV_TYPE_META;
+	else
+		mbh->mb_dev_type = SMD_DEV_TYPE_DATA;
```
NiuYawei, line 205: Ditto, it seems to me the device type isn't useful when the device ID is known.

src/mgmt/srv_query.c
```diff
@@ -585,13 +590,22 @@ ds_mgmt_dev_set_faulty(uuid_t dev_uuid, Ctl__DevManageResp *resp)
...
+	if (tgt_id < 0) {
+		D_ERROR("Device UUID:"DF_UUID" could be not found\n", DP_UUID(dev_uuid));
+		D_GOTO(out, rc = tgt_id);
+	}
+
+	xs_type = tgt2xs_type(tgt_id);
+	if (xs_type == DSS_XS_SYS)
+		faulty_info.bf_dev_type = SMD_DEV_TYPE_META;
+	else
+		faulty_info.bf_dev_type = SMD_DEV_TYPE_DATA;
```
NiuYawei, line 603: Ditto.

NiuYawei (changes requested):
The meta & wal device query/monitor will be done in a separate PR?

NiuYawei (approved):

src/bio/bio_context.c
```diff
@@ -275,6 +275,24 @@ bio_bs_hold(struct bio_blobstore *bbs)
 	return rc;
 }
 
+struct bio_xs_blobstore *
+bio_xs_blobstore_by_devid(struct bio_xs_context *xs_ctxt, uuid_t dev_uuid)
+{
+	enum smd_dev_type	 st;
+	struct bio_xs_blobstore *bxb = NULL;
+
+	for (st = SMD_DEV_TYPE_DATA; st < SMD_DEV_TYPE_MAX; st++) {
```
tanabarr, line 284: Very minor comment but I would prefer using 0 as a starting point is it doesn't use implicit knowledge that DATA is the first entry in the enum.

tanabarr (approved):

PR #11225, opened 2023-01-13
DAOS-12384 control: Support fabric authorization keys

mjmac:
Adds a new fabric_auth_key server config parameter
which will be set on the server side and also propagated to
client applications via the agent.

Bonus: Also adds a new client_env_vars section to the server
config that allows custom client-side environment variables
to be defined. The client environment may override these
variables with locally-defined values.

Required-githooks: true


src/cart/crt_hg.c
```diff
@@ -746,6 +746,8 @@ crt_hg_class_init(int provider, int idx, bool primary, hg_class_t **ret_hg_class
 	if (rc != 0)
 		D_GOTO(out, rc);
 
+	init_info.na_init_info.auth_key = getenv("CRT_AUTH_KEY");
```
frostedcmos, line 749: We should check for the provider here and do this only for provider == CRT_PROV_OFI_CXI.

Also, as a rule we read out envariables in data_init() call and store values in crt_gdata struct.

New envariables should be documented in src/cart/README.env
As a last thing, envariables should also be added to dump_envariables() function to help with the debug later on.

mjmac: > We should check for the provider here and do this only for provider == CRT_PROV_OFI_CXI.
> 
Auth keys are used by more than just CXI though, no? Looks like maybe also the GNI provider. In any case, it seems like it would be forward compatible to just set it and let mercury decide what to do with it (if anything).

> Also, as a rule we read out envariables in data_init() call and store values in crt_gdata struct.
> 
So are you saying that you would like to see a new `char *auth_key` member of that struct?

> New envariables should be documented in src/cart/README.env 
>
OK, I can add that.

> As a last thing, envariables should also be added to dump_envariables() function to help with the debug later on.
>
I can do that, but I wonder if there are security considerations there.

frostedcmos: We don't use auth keys for anything else today. My concern is that if this gets set in env/scripts and starts causing issues with TCP when people forget to unset it during testing. We also don't support GNI, nor planning to. @soumagne what is the current mercury behavior for non cxi providers when auth key is set? 

Yes something like char auth_key[MAX_KEY_SIZE], so that we limit the input to something reasonable like max=255. Otherwise we might get flagged by SDL/codescans. 

Good point on security considerations, perhaps then lets skip that part for now.

soumagne: There is no need for auth keys for tcp so no need to worry about it. I think it's harmless to set it like that since even for tcp we already have a check for it internally and depending on the provider (gni, CXI) the auth key has a different format.

mjmac: > Another thing to think about here is whether this is sufficient for a multi-provider case. Do we need 2 different auth keys in case of CXI primary, CXI secondary? Primary and secondary would be on different networks.

My understanding is that there will be a single VNI for DAOS traffic. I suppose that we could try to future-proof it by making it support multiple keys from the beginning, but I'm not sure what that looks like at the mercury level. @soumagne: Any thoughts on this?

soumagne: Each cart context / mercury class will take a single key. But yes if in the future we have multiple CXI  domains (primary + secondary) as Alex introduced I could see a need for having multiple auth keys (though still only one would be used per context / class).

mjmac: I spent some time trying to add the requested support for multiple auth keys, but it quickly gets complicated because we need to deal with multiple strings. All of that went OK, but then I got to `crt_hg_class_init()` and realized that I needed to copy the auth key string so that it could be safely used in the `init_info` struct, but I didn't see a good way to free that string later on (there's no `crt_hg_class_fini()`) without adding more infrastructure.

As all of this stuff is well outside my area of expertise, I would prefer to land my simple solution here to make sure that we are not blocked on Aurora testing, and leave the correct solution to the networking experts. OK?

mjmac: Actually, after re-reading the code, I guess it's safe to free the auth key string after `HG_init_opt()` is finished, as the mercury stuff just does a `sscanf()` to parse the key into a uint32/uint16 pair. I was worried that the string was used elsewhere. I'll just finish my patch.

frostedcmos: I am a bit confused here -- The handling of multiple auth keys should be similar to handling of multiple provider, interface, domain and port strings. Those are done in crt_init_opt():
So something like:
`auth_env = get_env("CRT_AUTH_KEY");

rc = __split_arg(auth_env, &key0, &key1);

rc = crt_na_config_init(true, primary_provider, iface0, domain0, port0, key0);`

And similarly for initing secondary providers (we only support 1 for now) for key1.

frostedcmos: Added the comment before the latest patch:) will re-read

src/cart/crt_init.c
```diff
@@ -606,6 +614,16 @@ crt_init_opt(crt_group_id_t grpid, uint32_t flags, crt_init_options_t *opt)
 				D_DEBUG(DB_ALL, "set group_config_path as %s.\n", path);
 		}
 
+		if (opt && opt->cio_auth_key)
+			auth_key_env = opt->cio_auth_key;
+		else {
+			auth_key_env = getenv("CRT_AUTH_KEY");
```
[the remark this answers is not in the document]
mjmac: Sure, I can do that. I was wondering what the `D_$FOO` stuff was about.

frostedcmos (approved):
Cart side changes LGTM

soumagne (approved):

tanabarr (approved):
Go changes look good

kjacque (commented):
The code itself looks good, but I find myself wondering about the implications of putting the auth key value into an arbitrary set of environment variables in the client net hint. I'm not totally opposed to the idea, but reading the code, it feels inconsistent. As things are now, the agent's responsibility deals with general concepts, and libdaos deals with translating those to client environment variables. The arbitrary list shifts that translation to the agent side.

Is the overall vision to shift all of our client net hint values into this variable list? If not, I think this case just adds some confusion. I like the client_env_vars section -- for that, this implementation makes sense -- but stuffing the fabric auth key into it muddies the waters.

mjmac:
> The code itself looks good, but I find myself wondering about the implications of putting the auth key value into an arbitrary set of environment variables in the client net hint. I'm not totally opposed to the idea, but reading the code, it feels inconsistent. As things are now, the agent's responsibility deals with general concepts, and libdaos deals with translating those to client environment variables. The arbitrary list shifts that translation to the agent side.
> 
> Is the overall vision to shift all of our client net hint values into this variable list? If not, I think this case just adds some confusion. I like the client_env_vars section -- for that, this implementation makes sense -- but stuffing the fabric auth key into it muddies the waters.
> 
I went back and forth on this, tbh. Where I ended is that it probably doesn't make sense to continue padding out the `client_net_hint` message and `dc_mgmt_net_cfg()` function for every little thing we want to convey to the client. There will almost certainly be more of this sort of thing, and in most cases they can be handled with just a simple `setenv()`. In cases where there is more complex client-side handling required (e.g. SRX stuff, OFI_DOMAIN/OFI_INTERFACE), then it does make sense to keep the extra code.

PR #11237, opened 2023-01-16
DAOS-11551 tools: Update daos cmd output for clone/copy (#11189)

mjmac:
Improve output options to ensure reliable output of container
clone and fs copy information (for both standard and JSON output).

### DIFF
--- a/src/utils/daos_hdlr.c
+++ b/src/utils/daos_hdlr.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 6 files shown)


tanabarr (approved):

src/utils/daos_hdlr.c
```diff
@@ -2226,11 +2201,21 @@ fs_copy_hdlr(struct cmd_args_s *ap)
...
+
+	D_ALLOC(ca, sizeof(struct dm_args));
+	if (ca == NULL)
+		D_GOTO(out, rc = -DER_NOMEM);
+	ap->dm_args = ca;
 
-	set_dm_args_default(&ca);
+	D_ALLOC(num, sizeof(struct fs_copy_stats));
+	if (num == NULL)
+		D_GOTO(out, rc = -DER_NOMEM);
```
ashleypittman, line 2215: Will this leak ca?

mjmac: Actually, it won't leak, even without the explicit cleanup I just pushed. The reason is that over in [utils.go:freeCmdArgs()](https://github.com/daos-stack/daos/pull/11237/files#diff-2a10e44da826c23ea5b3390a8c3db666d1bd672512fd0d53e6f2d3626f290638R193), we free the members regardless of whether or not the handler succeeded. That's why the `fs copy` fi test didn't fail.

mjmac:
Latest push reverts the unnecessary free on err (handled on the Go side), but does incorporate the Coverity fix from #11260.

tanabarr (approved):

PR #11245, opened 2023-01-18
DAOS-12408 cart: Split data_init()

frostedcmos:
- split data_init() into part that can be initialized multiple times and a part that should only be initialized once, such as mutex init

Required-githooks: true

Signed-off-by: Alexander A Oganezov <alexander.a.oganezov@intel.com>

### Gatekeeper:

### DIFF
--- a/src/cart/crt_init.c
+++ b/src/cart/crt_init.c
@@ -17,8 +17,48 @@
 struct crt_gdata crt_gdata;
 static volatile int   gdata_init_flag;
 struct crt_plugin_gdata crt_plugin_gdata;
-static pthread_once_t data_init_once = PTHREAD_ONCE_INIT;
 
+static void
+crt_lib_init(void) __attribute__((__constructor__));
+
+static void
+crt_lib_fini(void) __attribute__((__destructor__));
+
+/* Library initialization/constructor */
+static void
+crt_lib_init(void)
+{
+	int		rc;
+	uint64_t	start_rpcid;
+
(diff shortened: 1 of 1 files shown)


soumagne (changes requested):
I don't think that's correct to use pthread_once here ? it seems like either that init part should be done somewhere else or you should just be using a constructor/destructor?

frostedcmos:
> I don't think that's correct to use pthread_once here ? it seems like either that init part should be done somewhere else or you should just be using a constructor/destructor?

Why is not not correct to use pthread_once? it's designed for things that have to be initialized only once. we use it in other parts of the stack. I agree we could use library constructor/destructor too, but that seems like a bit of an overkill.

soumagne:
OK well "correct" was maybe not the right terminology but I just feel it's a bit messy :) since all you're doing here is trying to init some global data only once in the middle of other routines that are going be called concurrently... it's probably okay in practice but I feel like it would be much cleaner with a constructor/destructor and I don't see the code here that destroys the R/W lock (which should also be called only once).

soumagne:
actually I see the code now that destroys the R/W lock. I take back what I said, it's probably okay, if you feel like it's good like that I won't oppose to it but I still feel like it's a bit convoluted.

jgmoore-or (approved):

soumagne (approved):

jgmoore-or (approved):

PR #11258, opened 2023-01-19
DAOS-12417 security: Add ui.ACLPrincipalFlag type

mjmac:
Centralizes input handling for user/group principals, and
improves user-friendliness by automatically converting bare
names to principals as necessary.

Required-githooks: true


mjmac:
@kjacque: Tagged you for early review, mostly just to validate the approach. This came up as a result of a customer question. The docs/help do say that the format is `name@[domain]`, but it seemed like better UX to just append the `@`.

kjacque (approved):

tanabarr (approved):

knard-intel (approved):
LGTM

PR #11259, opened 2023-01-19
DAOS-12012 cart: Segfault in OFI-RXM when using SPDK DAOS bdev module

frostedcmos:
The crash happened due to bdev module using crt_context_create() and crt_context_destroy() in parallel from different threads, causing internal messup of context indecies and associated tables.

- Introduce list of free indecies for context ids and helper functions to get/put the index.
- Use free index instead of 'context number' as a context id.

Required-githooks: true

Signed-off-by: Alexander A Oganezov <alexander.a.oganezov@intel.com>

### Gatekeeper:

### DIFF
--- a/src/cart/crt_context.c
+++ b/src/cart/crt_context.c
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
--- a/src/cart/crt_context.c
+++ b/src/cart/crt_context.c
@@ -213,7 +213,6 @@ crt_context_provider_create(crt_context_t *crt_ctx, crt_provider_t provider, boo
 	struct crt_context	*ctx = NULL;
 	int			rc = 0;
 	size_t			uri_len = CRT_ADDR_STR_MAX_LEN;
-	int			cur_ctx_num;
 	int			ctx_idx;
 	int			max_ctx_num;
 	d_list_t		*ctx_list;
@@ -224,19 +223,20 @@ crt_context_provider_create(crt_context_t *crt_ctx, crt_provider_t provider, boo
 	}
 
 	D_RWLOCK_WRLOCK(&crt_gdata.cg_rwlock);
-	cur_ctx_num = crt_provider_get_cur_ctx_num(primary, provider);
 	max_ctx_num = crt_provider_get_max_ctx_num(primary, provider);
 	ctx_idx = crt_provider_get_ctx_idx(primary, provider);
 
-	if (ctx_idx < 0 || ctx_idx >= max_ctx_num || cur_ctx_num >= max_ctx_num) {
-		D_WARN("Provider: %d; Number of active contexts (%d) reached limit (%d).\n",
-		       provider, cur_ctx_num, max_ctx_num);
+	if (ctx_idx < 0 || ctx_idx >= max_ctx_num) {
+		D_WARN("Provider: %d; Context limit (%d) reached\n",
+		       provider, max_ctx_num);
+		crt_provider_put_ctx_idx(primary, provider, ctx_idx);
 		D_RWLOCK_UNLOCK(&crt_gdata.cg_rwlock);
 		D_GOTO(out, rc = -DER_AGAIN);
 	}
 
 	D_ALLOC_PTR(ctx);
 	if (ctx == NULL) {
+		crt_provider_put_ctx_idx(primary, provider, ctx_idx);
 		D_RWLOCK_UNLOCK(&crt_gdata.cg_rwlock);
 		D_GOTO(out, rc = -DER_NOMEM);
 	}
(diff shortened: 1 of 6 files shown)


soumagne (dismissed):

soumagne (dismissed):

src/cart/crt_context.c
```diff
@@ -225,8 +226,9 @@ crt_context_provider_create(crt_context_t *crt_ctx, crt_provider_t provider, boo
 	D_RWLOCK_WRLOCK(&crt_gdata.cg_rwlock);
 	cur_ctx_num = crt_provider_get_cur_ctx_num(primary, provider);
 	max_ctx_num = crt_provider_get_max_ctx_num(primary, provider);
+	ctx_idx = crt_provider_get_ctx_idx(primary, provider);
 
-	if (cur_ctx_num >= max_ctx_num) {
+	if (ctx_idx < 0 || ctx_idx >= max_ctx_num || cur_ctx_num >= max_ctx_num) {
```
mchaarawi, line 231: is cur_ctx_num still needed here?

src/cart/crt_init.c
```diff
@@ -174,6 +175,9 @@ prov_data_init(struct crt_prov_gdata *prov_data, crt_provider_t provider,
 	prov_data->cpg_max_unexp_size = max_unexpect_size;
 	prov_data->cpg_primary = primary;
 
+	for (i = 0; i < CRT_SRV_CONTEXT_NUM; i++)
+		prov_data->cpg_used_idx[i] = false;
```
mchaarawi, line 179: is that necessary? don't/can't you initialize that struct to {0} somewhere before you call this init?

src/cart/crt_hg.c
```diff
@@ -601,6 +601,30 @@ crt_provider_dec_cur_ctx_num(bool primary, int provider)
...
+	prov_data->cpg_used_idx[idx] = false;
+}
+
+int
+crt_provider_get_ctx_idx(bool primary, int provider)
+{
+	struct crt_prov_gdata	*prov_data = crt_get_prov_gdata(primary, provider);
+	int			i;
+
+	for (i = 0; i < CRT_SRV_CONTEXT_NUM; i++) {
```
mchaarawi, line 618: sorry if that seems naive:
Is that really an index required to be passed down to mercury or just any user defined id for the context?
do you need an array table here or just a context ID that is used in the context and here you just maintain how many context you use? 
i.e. can you just always return idx++; and all you maintain are how many active indexes there are at any given point (not like you had before where the id returned was the max index in use)?

src/cart/crt_context.c
frostedcmos: We still have to increment as other place needs to know total number of contexts, so we already have cur_ctx_num; so might as well do this sanity check

mchaarawi: sure, but can't this be done as part of crt_provider_get_ctx_idx() instead of having to call 2 functions one to increment and one to get the available context ID?

mchaarawi: ah my bad sorry i got confused with crt_provider_inc_cur_ctx_num and crt_provider_get_cur_ctx_num.
anyway it just seems unnecessary here to call crt_provider_get_cur_ctx_num() on line 226 anymore.

src/cart/crt_context.c
```diff
@@ -225,8 +226,9 @@ crt_context_provider_create(crt_context_t *crt_ctx, crt_provider_t provider, boo
 	D_RWLOCK_WRLOCK(&crt_gdata.cg_rwlock);
 	cur_ctx_num = crt_provider_get_cur_ctx_num(primary, provider);
 	max_ctx_num = crt_provider_get_max_ctx_num(primary, provider);
+	ctx_idx = crt_provider_get_ctx_idx(primary, provider);
```
mchaarawi, line 229: so if anything fails in this function, you don't call crt_provider_put_ctx_idx() to reset that idx back to unused. that seems problematic as you might run out of indexes in the table without any being actually in use.

mchaarawi (changes requested):

src/cart/crt_init.c
frostedcmos: we can, but i like explicit init

src/cart/crt_context.c
frostedcmos: Agreed, i ll increment/decrement as part of get/put index

src/cart/crt_hg.c
frostedcmos: we use context index in other parts of the cart code to access static cache uri/hg_addr tables, so it needs to be between {0 and CRT_SRV_CONTEXT_NUM-1}. Similarly it is used for endpoint address generation (base_addr + ctx_idx). Eventually we might get rid of those static tables, but for now it would be a big change and some places would still need a linear index

src/cart/crt_context.c
frostedcmos: good point will do a better cleanup of other things as well in next push

src/cart/crt_hg.c
mchaarawi: got it. thx

src/cart/crt_context.c
frostedcmos: On this one, the failures will cause context_destroy that will do the put

mchaarawi: well that's still a problem because crt_context_destroy can itself fail and not put the context that you take here (you are not checking the return value from destroy on the error path here in create).

frostedcmos: on the error path in the create we call context_destroy with force=true (second argument), which should cause context_destroy to force go-through everything ignoring intermediate rcs.  Having said that, force arg was not being used in 1 of the rcs, so fixing it next upload

mchaarawi (approved):

soumagne (approved):

jgmoore-or (approved):

PR #11299, opened 2023-01-24
DAOS-12241 test: Moving io/large_file_count.py back to large cluster

saurabhtandan:
io/large_file_count.py is consistently failing since moved to medium cluster. Reason being, test timeouts while running mdtest as the workload is handled by a single client now compared to 3 client nodes previously. Reducing the workload to accommodate the test on medium cluster will defeat the purpose of the test. Hence, moving back to large cluster.

Quick-Functional: true
Test-repeat: 10
Test-tag: test_largefilecount

### DIFF
--- a/src/tests/ftest/io/large_file_count.yaml
+++ b/src/tests/ftest/io/large_file_count.yaml
@@ -1,6 +1,6 @@
 hosts:
-  test_servers: 3
-  test_clients: 1
+  test_servers: 5
+  test_clients: 3
 timeout: 5000
 server_config:
   name: daos_server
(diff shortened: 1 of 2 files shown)


src/tests/ftest/io/large_file_count.yaml
```diff
@@ -1,6 +1,6 @@
 hosts:
-  test_servers: 3
-  test_clients: 1
+  test_servers: 5
+  test_clients: 3
```
daltonbohning, line 3: Have we considered 2 servers, 2 clients so it still fits on Medium?

daltonbohning (approved):

shimizukko (approved):

src/tests/ftest/io/large_file_count.yaml
ashleypittman: This test never has fit on medium, it's failed ever since it was moved there.  I think this is the correct way forward.

PR #11301, opened 2023-01-25
DAOS-12443 common: umem unit tests for bmem

sherintg:
- Added test unit tests for all of the umem apis with bmem as the backing interface.
- Addressed bugs in dav_free(), dav atomic apis.

### Gatekeeper:

### DIFF
--- a/src/common/dav/tx.c
+++ b/src/common/dav/tx.c
@@ -452,18 +452,21 @@ tx_create_wal_entry(struct ulog_entry_base *e, void *arg,
 int
 lw_tx_begin(dav_obj_t *pop)
 {
-	struct umem_wal_tx	*utx = pop->do_utx;
+	struct umem_wal_tx	*utx = NULL;
 	int			 rc;
 
-	D_ASSERT(utx == NULL);
-	utx = dav_umem_wtx_new(pop);
-	if (utx == NULL)
-		return obj_tx_fail_err(EINVAL, 0);
+	if (pop->do_utx == NULL) {
+		utx = dav_umem_wtx_new(pop);
+		if (utx == NULL)
+			return obj_tx_fail_err(EINVAL, 0);
+	}
 	rc = dav_wal_tx_reserve(pop);
 	if (rc) {
 		D_ERROR("so_wal_reserv failed, "DF_RC"\n", DP_RC(rc));
-		D_FREE(utx);
-		pop->do_utx = NULL;
+		if (utx) {
+			D_FREE(utx);
+			pop->do_utx = NULL;
+		}
 	}
 	return rc;
 }
@@ -510,11 +513,10 @@ dav_tx_begin(dav_obj_t *pop, jmp_buf env, ...)
 
 		VALGRIND_START_TX;
 	} else if (tx->stage == DAV_TX_STAGE_NONE) {
-		struct umem_wal_tx *utx;
+		struct umem_wal_tx *utx = NULL;
 
 		DAV_DBG("");
-		utx = pop->do_utx;
-		if (utx == NULL) {
+		if (pop->do_utx == NULL) {
 			utx = dav_umem_wtx_new(pop);
 			if (utx == NULL) {
 				err = ENOMEM;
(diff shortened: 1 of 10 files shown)


src/common/dav/tx.c
```diff
@@ -452,18 +452,21 @@ tx_create_wal_entry(struct ulog_entry_base *e, void *arg,
 int
 lw_tx_begin(dav_obj_t *pop)
 {
-	struct umem_wal_tx	*utx = pop->do_utx;
+	struct umem_wal_tx	*utx = NULL;
 	int			 rc;
 
-	D_ASSERT(utx == NULL);
```
NiuYawei, line 458: Could you explain a bit why this assert is removed?

src/common/dav/tx.c
```diff
@@ -452,18 +452,21 @@ tx_create_wal_entry(struct ulog_entry_base *e, void *arg,
...
-	utx = dav_umem_wtx_new(pop);
-	if (utx == NULL)
-		return obj_tx_fail_err(EINVAL, 0);
+	if (pop->do_utx == NULL) {
+		utx = dav_umem_wtx_new(pop);
+		if (utx == NULL)
+			return obj_tx_fail_err(EINVAL, 0);
+	}
 	rc = dav_wal_tx_reserve(pop);
```
NiuYawei, line 463: I think the WAL reserve should be called only once for a umem_wal_tx, maybe we'd combine the dav_umem_wtx_new() & dav_wal_tx_reserve() to single function?

src/common/dav/tx.c
sherintg: umem/dav_reserve() will create a wal entries without actually commiting it. If atomic api is called post this we will end up hitting this assert

src/common/dav/tx.c
sherintg: The current implementation allows creation of wal entries without a need to call wal_reserve. This helps in the implementation of dav_reserve() where the changes it does to memory can be persisted during the next transaction.

src/common/dav/tx.c
NiuYawei: Sorry, I forgot why reserve needs to generate WAL entries, but can we guarantee that the WAL entries for reserve are committed by the publish transaction? For example:
1. ULT_A does reserve;
2. ULT_B start & commit transaction B (or call atomic API B);
3. ULT_A commits transaction A to publish reservation;

The reserve won't be mistakenly committed by B, right?

sherintg: reserve will occasionally create wal entries if the allocation requires new chunk or zone to be created. We do not have undo mechanism for chunk/zone creation and this change needs to be committed along with the next commit that happens to the vos file. The commit of actual allocation block will happen only when ULT_A commits its transaction. In the above example when ULT_B does commit, it commits the changes related to chunk and zone creation and when ULT_A does tx_publish the redo entry for the allocation block within the chunk will be created.

NiuYawei: I see, thanks.

src/common/dav/tx.c
NiuYawei: Ok, but we need to avoid calling WAL reserve multiple times for a umem_wal_tx, maybe introducing a flag in umem_wal_tx to indicate WAL reserve is called? (or take UINT64_MAX as invalid/uninitialized WAL ID).

ivadmitry (approved):

src/common/dav/tx.c
sherintg: We call wal_reserve for only two cases:
1. Outer most tx begin
2. Beginning of an atomic operation
The only scenario where wal reserve gets called again within a transaction is when atomic calls are made from within a transaction. Since we cannot support this scenario, I am thinking whether placing an assert in the bmem wrappers would suffice?  Is that ok with you?

NiuYawei: Sure, then it'll be great to add an assert(utx_id == ULLONG_MAX) in dav_wal_tx_reserve().

sherintg: Done.

NiuYawei (approved):

ivadmitry (approved):

PR #11317, opened 2023-01-27
DAOS-8938 tools: Fix --path handling for daos cont commands

mjmac:
As --path support was bolted on during the transition from
--pool/--cont flags to positional arguments, some regressions
crept in. This commit refactors things to centralize logic
and responsibility for consistent application of the argument
handling rules.

Also adds positional arg support for the (get|set)-(attrs|props)
subcommands for consistency with the dmg command surface.

Features: daos_cmd
Required-githooks: true


src/tests/ftest/container/query_attribute.py
```diff
@@ -47,6 +72,7 @@ def test_container_query_attr(self):
         :avocado: tags=all,full_regression
         :avocado: tags=vm
         :avocado: tags=container
+        :avocado: tags=daos_cmd
```
daltonbohning, line 75: "features" are usually put on the same line. For this and other files
```suggestion
        :avocado: tags=container,daos_cmd
```

But... Is the idea to add `daos_cmd` to EVERY test that uses the `daos` command? Because that be would nearly every test in the framework, since we use `daos` to create containers 99% of the time. Most references are just buried in the framework. In fact, files that directly reference `DaosCommand`, etc. are slowly being changed to use wrapper utils.

src/tests/ftest/container/query_attribute.py
```diff
@@ -147,6 +150,91 @@ def test_container_query_attr(self):
         self.log.debug(str(actual_attrs))
         self.assertEqual(actual_attrs, expected_attrs)
 
+    def test_container_query_attrs(self):
+        """JIRA ID: DAOS-4640
```
daltonbohning, line 154: JIRA ID copy-paste?

src/tests/ftest/container/query_attribute.py
```diff
@@ -147,6 +150,91 @@ def test_container_query_attr(self):
...
+                        "Unexpected output for get_attr: {} != {}\n".format(
+                            actual_val, escape_to_not[attr_values[key]]))
+            else:
+                # Standard character string.
+                if actual_val != attr_values[key]:
+                    errors.append(
+                        "Unexpected output for get_attr: {} != {}\n".format(
+                            actual_val, attr_values[key]))
+
+        self.assertEqual(len(errors), 0, "; ".join(errors))
```
daltonbohning, line 236: There is also a common function for this
```suggestion
        from general_utils import report_errors
        report_errors(self, errors)
```

src/tests/ftest/util/daos_utils.py
```diff
@@ -297,7 +297,28 @@ def pool_set_attr(self, pool, attr, value, sys_name=None):
...
-            ("pool", "set-attr"), pool=pool, attr=attr, value=value,
+            ("pool", "set-attr"), pool=pool, attr=':'.join([attr, value]),
+            sys_name=sys_name)
+
+    def pool_set_attrs(self, pool, attrs, sys_name=None):
+        """Set multiple pool attributes.
+
+        Args:
+            pool (str): Pool UUID.
+            attrs (map): Attribute name/value map.
```
daltonbohning, line 308: `map` in python has a different meaning
```suggestion
            attrs (dict): Attribute name/value dictionary.
```

src/tests/ftest/util/daos_utils.py
```diff
@@ -379,6 +400,28 @@ def container_set_prop(self, pool, cont, prop, value):
             ("container", "set-prop"),
             pool=pool, cont=cont, prop=prop_value)
 
+    def container_set_props(self, pool, cont, props, sys_name=None):
+        """Set multiple container properties.
+
+        Args:
+            pool (str): Pool UUID.
+            cont (str): Container UUID.
+            props (map): Property name/value pairs.
```
daltonbohning, line 409:
```suggestion
            props (dict): Property name/value pairs.
```

src/tests/ftest/util/daos_utils.py
```diff
@@ -563,10 +606,32 @@ def container_set_attr(
...
-            sys_name=sys_name, attr=attr, value=val)
+            sys_name=sys_name, attr=':'.join([attr, val]))
+
+    def container_set_attrs(self, pool, cont, attrs, sys_name=None):
+        """Set multiple container attributes.
+
+        Args:
+            pool (str): Pool UUID.
+            cont (str): Container UUID.
+            attrs (map): Attribute key/val pairs.
```
daltonbohning, line 617:
```suggestion
            attrs (dict): Attribute key/val pairs.
```

src/tests/ftest/container/query_attribute.py
```diff
@@ -7,6 +7,31 @@
 from apricot import TestWithServers
 
 
+# Test container set-attr, get-attr, and list-attrs with different
+# types of characters.
+test_strings = [
```
daltonbohning, line 12: I think this should disable the pylint errors for the backslashes
```suggestion
# pylint: disable=anomalous-backslash-in-string
test_strings = [
```

src/tests/ftest/container/query_attribute.py
mjmac: > Is the idea to add daos_cmd to EVERY test that uses the daos command?

No, more just to find a set of tests that will exercise as much of the command surface as possible in order to verify that core changes to the command don't break existing functionality.

src/tests/ftest/container/query_attribute.py
mjmac: I figured I'd leave it as that was the original ticket for adding this test. I'm just adding a new testcase...

src/tests/ftest/util/daos_utils.py
mjmac: Oh, right. Thanks.

src/tests/ftest/container/query_attribute.py
mjmac: Done, thanks.

src/tests/ftest/container/query_attribute.py
daltonbohning: No worries. I don't think we have clear guidance on when to use or not use a JIRA ID like this anyway. Many tests don't have it at all. But either way, the git history is a better method to determine when cases were added/modified.

src/tests/ftest/container/query_attribute.py
mjmac: I'll try that, thanks.

src/tests/ftest/container/query_attribute.py
daltonbohning: FWIW, the `container` tag covers much of the `daos container` operations. If adding `daos_cmd` helps you / the control team, I'm okay with it. We still have a bit of work to do with tags, so it's hard to be picky :)

mjmac: There is other functionality being added that isn't covered by the `container` tag, e.g. datamover stuff, autotest, etc. The idea is to try to pick tests that we can't easily run in a single-node dev setup to verify that we haven't broken some important but obscure part of the tool.

mjmac:
Two consistent known failures:
- [DAOS-12013](https://daosio.atlassian.net/browse/DAOS-12013)
- [DAOS-12526](https://daosio.atlassian.net/browse/DAOS-12526)

src/tests/ftest/util/daos_utils.py
```diff
@@ -297,7 +297,28 @@ def pool_set_attr(self, pool, attr, value, sys_name=None):
 
         """
         return self._get_result(
-            ("pool", "set-attr"), pool=pool, attr=attr, value=value,
+            ("pool", "set-attr"), pool=pool, attr=':'.join([attr, value]),
+            sys_name=sys_name)
+
+    def pool_set_attrs(self, pool, attrs, sys_name=None):
+        """Set multiple pool attributes.
+
```
daltonbohning, line 305: Just FYI that once this PR lands, I'd like to personally reconcile the `set_attr` and `set_attrs` such that all tests just use your new version.

daltonbohning (commented):
ftest LGTM

tanabarr (approved):
go changes look good

kjacque (approved):

daltonbohning (commented):
ftest LGTM

kjacque (approved):

tanabarr (approved):

PR #11326, opened 2023-01-30
DAOS-12415 test: Allow for tests which may produce core files.

ashleypittman:
Add logic to not fail the DaosBuild test if it creates a forefile,
it sometimes creates a corefile from conftest which is expected
and should not cause a test failure.
Fix a bug where the program that generated the core file was read
incorrectly.
Add some debugging where the DaosBuild test fails.

Required-githooks: true

### DIFF
--- a/src/tests/ftest/launch.py
+++ b/src/tests/ftest/launch.py
@@ -2656,12 +2656,6 @@ class Launch():
             self._fail_test(self.result.tests[-1], "Process", message, sys.exc_info())
             return 256
 
-        # One of the dfuse tests intermittently creates core files which is known so make a
-        # special case for that test.
-        if str(test) == './dfuse/daos_build.py' and './conftest' in core_file_processing.exe_names:
-            logger.info('conftest core file exists from daos_build test, ignoring')
-            corefiles_processed = len(core_file_processing.exe_names) - 1
-
         if corefiles_processed > 0 and str(test) not in TEST_EXPECT_CORE_FILES:
             message = "One or more core files detected after test execution"
             self._fail_test(self.result.tests[-1], "Process", message, None)
--- a/src/tests/ftest/launch.py
+++ b/src/tests/ftest/launch.py
@@ -2656,11 +2656,11 @@ class Launch():
             self._fail_test(self.result.tests[-1], "Process", message, sys.exc_info())
             return 256
 
-        # One of the dfuse tests intermittently creates a core file which is known so make a
+        # One of the dfuse tests intermittently creates core files which is known so make a
         # special case for that test.
         if str(test) == './dfuse/daos_build.py' and './conftest' in core_file_processing.exe_names:
             logger.info('conftest core file exists from daos_build test, ignoring')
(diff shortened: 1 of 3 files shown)


src/tests/ftest/launch.py
```diff
@@ -2656,6 +2656,12 @@ def _process_core_files(self, test_job_results, test):
             self._fail_test(self.result.tests[-1], "Process", message, sys.exc_info())
             return 256
 
+        # One of the dfuse tests intermittently creates core files which is known so make a
+        # special case for that test.
+        if str(test) == './dfuse/daos_build.py' and './conftest' in core_file_processing.exe_names:
+            logger.info('conftest core file exists from daos_build test, ignoring')
+            corefiles_processed = len(core_file_processing.exe_names) - 1
```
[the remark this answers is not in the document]
ashleypittman: This did work although the local value of **corefiles_processed** might have been wrong in absolute terms.  If conftest was in the set then it would be the number of executables that created corefiles of interest and the variable was then compared against zero.

I've reworked this PR with your suggestions however.

phender (approved):
Changes look good, thanks.

ashleypittman:
This looks to be working again now, please review.

src/tests/ftest/process_core_files.py
```diff
@@ -101,8 +106,12 @@ def process_core_files(self, directory, delete):
                         command = ["lbzip2", "-d", "-v", os.path.join(core_dir, core_name)]
                         run_local(self.log, " ".join(command))
                         core_name = os.path.splitext(core_name)[0]
-                    self._create_stacktrace(core_dir, core_name)
-                    corefiles_processed += 1
+                    exe_name = self._get_exe_name(os.path.join(core_dir, core_name))
+                    self._create_stacktrace(core_dir, core_name, exe_name)
+                    if test in CORE_FILES_IGNORE and exe_name in CORE_FILES_IGNORE[test]:
+                        self.log.info('core file name configured for ignore, skipping')
```
phender, line 112: It would be nice to log the details, but that could be added later.:
```suggestion
                        self.log.debug('Excluding %s core file (%s) detected while running %s from processed core count', core_name, exe_name, test)
```

phender (approved):

src/tests/ftest/process_core_files.py
phender: @mjean308 when this lands and we have this capability in process_core_files.py, it may also make sense to replace the launch.py `TEST_EXPECT_CORE_FILES = ["./harness/core_files.py"]` code with a `CORE_FILES_IGNORE ` entry for the `./harness/core_files.py` test.

ashleypittman: I was thinking the same Phil.

mjean308: will do... I will create a ticket for this

mjean308 (approved):

daltonbohning (approved):

PR #11332, opened 2023-01-31
DAOS-12397 Support "append" in Hadoop-DAOS

zjf2012:
Implemented the optional append method in Hadoop-DAOS. 

Signed-off-by: jiafu zhang <jiafu.zhang@intel.com>


src/client/java/hadoop-daos/src/main/java/io/daos/fs/hadoop/DaosFileSource.java
```diff
@@ -38,24 +38,30 @@ public abstract class DaosFileSource {
...
-  protected DaosFileSource(DaosFile daosFile, int bufCapacity, long fileLen,
+  protected DaosFileSource(DaosFile daosFile, int bufCapacity, long fileLen, boolean append,
                            FileSystem.Statistics stats) {
     this.daosFile = daosFile;
     this.buffer = BufferAllocator.directNettyBuf(bufCapacity);
     selfManagedBuf = true;
     this.bufCapacity = buffer.capacity();
     this.fileLen = fileLen;
     this.stats = stats;
+    if (append) {
```
mchaarawi, line 49: do we need to require multithreaded access with the same handle?
like what if two files are appending to the same file at the same time? should this be serialized with some lock?

zjf2012: > do we need to require multithreaded access with the same handle? like what if two files are appending to the same file at the same time? should this be serialized with some lock?

I checked the append impl in HDFS, which is Hadoop's impl of Hadoop Filesystem. HDFS has a concept, file lease, in namenode which prevents concurrent update operations to same file. As talked in last meeting, DAOS doesn't have similar control to file. It seems that there is no API in client to lock a file.

So for now, I just assume there is no more 1 thread updating same file.

mchaarawi (approved):
im not familiar with Java code, so my review is mostly on the just on the file source.

johannlombardi (approved):

PR #11346, opened 2023-01-31
DAOS-12124 control: Fix pool create error after format

mjmac:
When the --size=% flag is used, the pool create command
performs a storage scan in order to determine the actual
available storage across the ranks. In cases where the
system has not started yet (i.e. pool create is run
immediately after storage format completes), the user may
receive a confusing error message. This patch changes the
flow to perform a system query before storage scan so that
the user receives a normal error if the MS hasn't started
yet.

Required-githooks: true

### DIFF
--- a/src/control/cmd/dmg/pool.go
+++ b/src/control/cmd/dmg/pool.go
@@ -10,7 +10,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"regexp"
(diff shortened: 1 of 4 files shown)


knard-intel (approved):
LGTM

tanabarr (approved):

kjacque (approved):

PR #11354, opened 2023-02-02
DAOS-10681 object: fix a bug in obj_ec_parity_lists_match

liuxuezhao:
In EC data recovery, possibly get different parity recx list from different parity shards. If they overlapped then possibly due to async vos agg, in that case should not return EIO.

Required-githooks: true

### Gatekeeper:


src/object/obj_ec.h
```diff
@@ -753,8 +754,16 @@ obj_ec_parity_lists_match(struct daos_recx_ep_list *lists_1,
 			     list_2->re_items[j].re_recx.rx_idx) ||
 			    (list_1->re_items[j].re_recx.rx_nr !=
 			     list_2->re_items[j].re_recx.rx_nr)) {
-				D_ERROR("got different parity recx in EC data recovery\n");
-				return -DER_IO;
+				if (!DAOS_RECX_OVERLAP(list_1->re_items[j].re_recx,
+						       list_2->re_items[j].re_recx)) {
+					D_ERROR("got different parity recx in EC data recovery\n");
```
wangdi1, line 759: Hmm, how about this case?
list1 [0-2M], [4-6M]
list2 [1M-3M], [5M-7M] 

should return -DER_IO ?

liuxuezhao: I think that case is failure case, because it happens when client try to recover one or multiple full-stripe, so only reasonable mismatch is that one parity return multiple partial exts, and another return less number of merged exts, right?

wangdi1: I mean in this case, if parity1 missing [2M-4M], [6M-7M], parity2 missing [0-1M], [3M-5M] due to the corruption, then it should return EIO, instead of keep retrying?

liuxuezhao: that case current code will return EIO, no?

src/object/obj_ec.h
```diff
@@ -741,8 +772,16 @@ obj_ec_parity_lists_match(struct daos_recx_ep_list *lists_1,
 	for (i = 0; i < nr; i++) {
 		list_1 = &lists_1[i];
 		list_2 = &lists_2[i];
-		if (list_1->re_nr != list_2->re_nr ||
-		    list_1->re_ep_valid != list_2->re_ep_valid) {
+		if (list_1->re_ep_valid != list_2->re_ep_valid) {
+			D_ERROR("got different ep_valid in EC data recovery\n");
+			return -DER_IO;
+		}
+		if (list_1->re_nr != list_2->re_nr) {
```
wangdi1, line 779: need compare the list_1 and list_2 even list_1->re_nr == list_2->re_nr ?

liuxuezhao: I think need not compare that case, when the re_nr equals, the two recx lists should be match exactly.

wangdi1: ah, it will compare in the following.

src/object/obj_ec.h
```diff
@@ -741,8 +772,16 @@ obj_ec_parity_lists_match(struct daos_recx_ep_list *lists_1,
...
+			if (obj_ec_parity_list_covered(list_1, list_2) ||
+			    obj_ec_parity_list_covered(list_2, list_1)) {
+				D_DEBUG(DB_IO, "parity list mismatch but fully covered\n");
```
wangdi1, line 782: Hmm, server will set the epoch with vos_ec_agg, so the epoch should be same here?  I mean we should compare the epoch, if they are same, it can just use these extents, instead of keep FETCH_AGAIN to wait for the aggregation? no?

liuxuezhao: if the recxs with same epoch, they should be already merged at server side before reply to client, see "daos_recx_ep_list_merge(parity_list, iods_nr)", so here at client-side, the different recxs should with different epoch already, right?

wangdi1: ah, right.

wangdi1 (approved):

Nasf-Fan (approved):

PR #11359, opened 2023-02-03
DAOS-6341 security: Add client methods to determine user permissions

kjacque:
If a client has a user's uid/gids, this method can be used to determine the user's permissions for a given set of pool and container properties.

The client is responsible for collecting and caching the pool and container properties containing the ownership and ACL.

Features: security

### Gatekeeper:


src/security/acl.h
```diff
@@ -0,0 +1,55 @@
...
+ * \param[in]	user_info		User whose permissions we are getting
+ * \param[in]	min_owner_perms		Special permissions granted to the owner, if any
+ * \param[out]	permissions		The user's calculated permissions (DAOS_ACL_PERM_*)
+ *
+ * \return	0		Success
+ *		-DER_INVAL	Invalid input
+ */
+int
+get_acl_permissions(struct daos_acl *acl, struct ownership *ownership, struct acl_user *user_info,
+		     uint64_t min_owner_perms, uint64_t *permissions);
```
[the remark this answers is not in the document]
kjacque: Fixed

src/security/cli_security.c
```diff
@@ -161,3 +167,215 @@ get_cred_from_response(Drpc__Response *response, d_iov_t *cred)
...
+static int
+get_perms(daos_prop_t *prop, uint32_t acl_prop, uint32_t owner_prop, uint32_t group_prop,
+	  struct acl_user *user_info, uint64_t min_owner_perms, uint64_t *perms)
+{
+	struct daos_acl		*acl = NULL;
+	struct ownership	ownership = {0};
+	int			rc;
+
+	/* These ACL and ownership variables point to the data in-place in the prop, and thus don't
+	   need to be freed here. */
```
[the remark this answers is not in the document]
kjacque: Fixed

src/security/cli_security.c
```diff
@@ -161,3 +167,215 @@
...
+		D_GOTO(err_user, rc);
+	}
+
+	D_ALLOC_ARRAY(user_info->groups, nr_gids);
+	if (user_info->groups == NULL)
+		D_GOTO(err_group, rc = -DER_NOMEM);
+
+	user_info->nr_groups = 0;
+	for (i = 0; i < nr_gids; i++) {
+		rc = daos_acl_gid_to_principal(gids[i], &(user_info->groups[i]));
```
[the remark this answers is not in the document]
kjacque: Fixed

src/security/acl.c
```diff
@@ -0,0 +1,177 @@
...
+		*perms = 0;
+		rc = 0;
+	}
+
+	return rc;
+}
+
+int
+get_acl_permissions(struct daos_acl *acl, struct ownership *ownership, struct acl_user *user_info,
+		     uint64_t min_owner_perms, uint64_t *perms)
```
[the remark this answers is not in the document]
kjacque: Fixed

ashleypittman:
Looking at this now I have a couple of questions:

1. The same call for both pool and containers in one?  Would there be a performance advantage to having separate calls?  Even with caching dfuse is going to call this frequently.
2. Does the result distinguish between RO and RW access?
3. Is there an easy way of knowing what properties I need to fetch prior to calling this?
4. It looks like the properties are strings, if dfuse is doing caching is it enough to call get_prop periodically and if the string hasn't changed then it doesn't need to call this function?

kjacque:
> 1. The same call for both pool and containers in one?  Would there be a performance advantage to having separate calls?  Even with caching dfuse is going to call this frequently.

As long as the props are cached, I'm not sure there would be an advantage, unless you want to check pool and container at different frequencies. If the intent is to check them at the same time, there isn't any advantage to splitting them up. In this implementation we exit early if the user doesn't have pool access.

> 2. Does the result distinguish between RO and RW access?

It doesn't yet. I can add an input parameter that could take the DAOS_COO_* flags that were used by dfuse to open the container, if that would be useful.

> 3. Is there an easy way of knowing what properties I need to fetch prior to calling this?

DAOS_PROP_(PO|CO)_ACL
DAOS_PROP_(PO|CO)_OWNER
DAOS_PROP_(PO|CO)_OWNER_GROUP

For the container you can call daos_cont_get_acl, and it will fetch these three properties. For the pools we don't have a user-facing get_acl API anymore (it's in the admin tool only), although this is a good argument for having one.

> 4. It looks like the properties are strings, if dfuse is doing caching is it enough to call get_prop periodically and if the string hasn't changed then it doesn't need to call this function?

Yes, if nothing has changed since last access by a user, we can safely assume they still have access. The owner and group props are strings, correct. The ACL prop is a complex data structure, but you could memcmp, using daos_acl_get_size to determine the full length of the structure.

ashleypittman:
> > 1. The same call for both pool and containers in one?  Would there be a performance advantage to having separate calls?  Even with caching dfuse is going to call this frequently.
> 
> As long as the props are cached, I'm not sure there would be an advantage, unless you want to check pool and container at different frequencies. If the intent is to check them at the same time, there isn't any advantage to splitting them up. In this implementation we exit early if the user doesn't have pool access.

Actually dfuse can open multiple containers in a single pool concurrently so actually I'd much prefer if we had two seperate calls for pools and containers.

> > 2. Does the result distinguish between RO and RW access?
> 
> It doesn't yet. I can add an input parameter that could take the DAOS_COO_* flags that were used by dfuse to open the container, if that would be useful.

Dfuse will need to know this, I supposed RO is implied if you have RW so it could try RW and if that fails also try RO but it would be good to get back two bits from a single call.

kjacque:
> Actually dfuse can open multiple containers in a single pool concurrently so actually I'd much prefer if we had two seperate calls for pools and containers.

Okay, I'll make that change.

> Dfuse will need to know this, I supposed RO is implied if you have RW so it could try RW and if that fails also try RO but it would be good to get back two bits from a single call.

Maybe I misunderstood the question. In this implementation, it returns all of the container access bits (including some that don't apply to dfuse, but I thought this might be generally useful for other utilities in the future). What's most interesting to dfuse are DAOS_ACL_PERM_READ and DAOS_ACL_PERM_WRITE, both of which you'll get from this call as it is.

src/include/daos_security.h
```diff
@@ -625,6 +626,47 @@ daos_acl_principal_from_str(const char *principal_str,
...
+ * \param[out]	perms		Bitmap representing the user's permissions. Bits are defined
+ *				in enum daos_acl_perm.
+ *
+ * \return	0		Success
+ *		-DER_INVAL	Invalid input
+ *		-DER_NONEXIST	UID or GID not found on the system
+ *		-DER_NOMEM	Could not allocate memory
+ */
+int
+daos_sec_get_pool_permissions(daos_prop_t *pool_prop, uid_t uid, gid_t gid, gid_t *supp_gids,
```
mchaarawi, line 646: would it be better to make those functions in the pool and container headers and rename them to something more user-friendly?
so daos_pool_get_perm() and daos_cont_get_perm()?

kjacque: Will do.

knard-intel (dismissed):
LGTM for what I understand.
However, do not understand why the test report step of the CI have failed as all tests were successfully passed.

src/security/tests/cli_security_tests.c
```diff
@@ -348,6 +351,388 @@ test_request_credentials_returns_raw_bytes(void **state)
...
+	assert_rc_equal(dc_sec_get_cont_permissions(prop_no_acl, uid, gid, NULL, 0,
+						    &perms),
+			-DER_INVAL);
+	daos_prop_free(prop_no_acl);
+
+	daos_prop_free(cont_prop);
+	daos_acl_free(acl);
+}
+
+static void
```
kccain, line 550: Just confirming, this is the same function as in the daos_test daos_container.c source file but I assume needs to be copy/pasted in each separate test file? Not a big deal at all IMO, just making sure I have it right.

kccain (approved):

kccain (dismissed):

knard-intel (dismissed):
LGTM

src/security/acl.c
```diff
@@ -0,0 +1,177 @@
...
+
+#include <daos/debug.h>
+#include <daos_errno.h>
+#include <daos_security.h>
+#include "acl.h"
+
+bool
+is_ownership_valid(struct ownership *ownership)
+{
+	if (daos_acl_principal_is_valid(ownership->user) &&
```
tanabarr, line 15: nit: you could replace "if" with "return" here

tanabarr (approved):
Nice work!

src/include/daos_pool.h
```diff
@@ -611,6 +611,26 @@ int
...
+ *
+ * \param[in]	pool_prop	Pool property containing DAOS_PROP_PO_ACL/OWNER/OWNER_GROUP entries
+ * \param[in]	uid		User's local uid
+ * \param[in]	gid		Gid of the user's primary group
+ * \param[in]	supp_gids	Gids of the user's supplementary groups, if any
+ * \param[in]	nr_supp_gids	Length of the supp_gids list
+ * \param[out]	perms		Bitmap representing the user's permissions. Bits are defined
+ *				in enum daos_acl_perm.
+ *
+ * \return	0		Success
```
ashleypittman, line 625: The date on this file needs updating.

src/include/daos/security.h
```diff
@@ -10,7 +10,16 @@
 #include <stdint.h>
 #include <stddef.h>
 #include <daos_types.h>
+#include <daos_prop.h>
 
+/**
+ * Structure representing a resource's ownership by user and group,
+ * respectively.
+ */
```
ashleypittman, line 18: I think this would probably fit on one line given our 100 character line length.
```suggestion
/** Structure representing a resource's ownership by user and group, respectively. */
```

src/include/daos/security.h
```diff
@@ -10,7 +10,16 @@
...
+struct ownership {
+	char *user;	/** name of the user owner */
+	char *group;	/** name of the group owner */
+};
```
ashleypittman, line 22: If this structure is to be public then it needs renaming to avoid namespace pollution with applications.

src/security/cli_security.c
```diff
@@ -161,3 +167,212 @@ get_cred_from_response(Drpc__Response *response, d_iov_t *cred)
...
+
+err_grps:
+	for (i = 0; i < user_info->nr_groups; i++)
+		D_FREE(user_info->groups[i]);
+	D_FREE(user_info->groups);
+err_group:
+	D_FREE(user_info->group);
+err_user:
+	D_FREE(user_info->user);
+err_out:
```
ashleypittman, line 317: You could probably just simplify all this by just calling free_user_info_strings().

ashleypittman (commented):
The other thing I wonder is do you need the distinction between primary and secondary groups at all, there's nothing special about the primary that I'm aware of so you could in theory just pass in a single list.

src/security/cli_security.c
```diff
@@ -161,3 +167,211 @@ get_cred_from_response(Drpc__Response *response, d_iov_t *cred)
...
+{
+	struct acl_user	user_info = {0};
+	int		rc;
+
+	if (prop == NULL) {
+		D_ERROR("null property parameter\n");
+		return -DER_INVAL;
+	}
+
+	if (gids == NULL && nr_gids > 0) {
```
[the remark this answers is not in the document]
kjacque: That feels a bit too strict to me. An array of zero length is at least safe. We could check nr_gids == 0 before trying to allocate anything though.

kjacque:
> The other thing I wonder is do you need the distinction between primary and secondary groups at all, there's nothing special about the primary that I'm aware of so you could in theory just pass in a single list.

This was a question I asked myself, too. In the end I went with splitting them because it was how the code was already conceptualized, and I wanted to limit how much I was rewriting the underlying code in particular. I can put them in one list if you prefer.

src/security/acl.c
kjacque: Fixed

src/include/daos_pool.h
kjacque: Fixed

src/security/cli_security.c
kjacque: Fixed

ashleypittman:
> > The other thing I wonder is do you need the distinction between primary and secondary groups at all, there's nothing special about the primary that I'm aware of so you could in theory just pass in a single list.
> 
> This was a question I asked myself, too. In the end I went with splitting them because it was how the code was already conceptualized, and I wanted to limit how much I was rewriting the underlying code in particular. I can put them in one list if you prefer.

I think the answer is in Unix the primary group matters (it's what files are created as) but you can switch to secondary groups so one is active but the user has the option to switch to others so for Unix the interface as you have here makes sense.  For daos that distinction doesn't (or might not?) exist which is why this feels clunky, however it's actually exactly the same as all other unix group interfaces and therefore not unreasonable that we should stay in-line.

tanabarr (approved):

PR #11373, opened 2023-02-04
DAOS-11405 vos: Simplify vos slab allocation

jolivier23:
This is a precursor to subsequent changes I will make in my evtree dynamic root patch.  Rather than defining exact slab values, we define common known ones and use the size, aligned to 32 bytes to calculate the slab id.  This allows us to use the slab allocator for integer key nodes as well since they didn't previously fit the hard coded size definition.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/vos/ilog.h
+++ b/src/vos/ilog.h
@@ -166,7 +166,7 @@ struct ilog_entry {
 	int32_t		ie_idx;
 };
 
-#define ILOG_PRIV_SIZE 408
+#define ILOG_PRIV_SIZE 600
 /* Information about ilog entries */
 struct ilog_info {
 	/** Status of ilog entry */
--- a/src/vos/vos_internal.h
+++ b/src/vos/vos_internal.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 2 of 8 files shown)


src/vos/ilog.h
```diff
@@ -166,7 +166,7 @@ struct ilog_entry {
 	int32_t		ie_idx;
 };
 
-#define ILOG_PRIV_SIZE 408
+#define ILOG_PRIV_SIZE 600
```
jolivier23, line 169: I plan to look into separating the slab array from the individual umm since it's the same for every pool.  This is remove the need to add 32 bytes for every slab entry here.  But that is for a different patch.

src/vos/vos_internal.h
```diff
@@ -1289,31 +1289,52 @@ vos_iterate_key(struct vos_object *obj, daos_handle_t toh, vos_iter_type_t type,
...
-enum {
-	VOS_SLAB_OBJ_NODE	= 0,
-	VOS_SLAB_KEY_NODE	= 1,
-	VOS_SLAB_SV_NODE	= 2,
-	VOS_SLAB_EVT_NODE	= 3,
-	VOS_SLAB_EVT_DESC	= 4,
-	VOS_SLAB_OBJ_DF		= 5,
-	VOS_SLAB_EVT_NODE_SM	= 6,
-	VOS_SLAB_MAX		= 7
+/** Define common slabs.  We can refine this for 2.4 pools but that is for next patch */
```
NiuYawei, line 1292: I think PMDK allocator has already internally defined/used common allocation classes, which is quite similar to "common slabs", the major difference of 'common slab' with the allocation classes is the granularity?

I think the point of using slab is to totally eliminate internal fragmentations when the object size is known (and the object is massively used). Imagine an object size is 33 bytes, with this "common slabs", it will still have almost 50% internal fragmentations.

jolivier23: I'll have to look at how they do allocation when not using slabs.  My understanding from prior discussions with Piotr was that they have some overhead, probably to store back pointer with allocation whereas the slab allocation doesn't because everything in the block is the same size.

My plan was to do this so that I can then resize the various tree nodes to fit in 512 for better allocation locality (e.g fewer blocks overall).  The other reason for doing this is because the existing code didn't handle various sizes of tree nodes in VOS.

Anyway, let me do some digging on the allocations in PMDK.

jolivier23: One more thing, I don't believe we have any situations where the overhead is 50%

ryon-jensen (approved):
Looks good to me. However, I don't know anything about the internal allocation of PMKD.

src/vos/vos_internal.h
jolivier23: Ok, I have dug into it a little.   The default PMDK allocation classes all have 16 byte headers attached to each allocation.    They have only a handfull of classes in the size ranges we are creating in DAOS.  Anyway, suppose you want to allocation something that is 128 bytes, if I am reading the code correclty, that would fall into the class with 192 bytes because of the header.

Here are the classes defined in the range we are talking about with existing slab allocations.
class 0 unit_size=262144 alignment=0 units_per_block=0, header_type=1, class_id=0
class 1 unit_size=128 alignment=0 units_per_block=2045, header_type=1, class_id=1
class 2 unit_size=192 alignment=0 units_per_block=1364, header_type=1, class_id=2
class 3 unit_size=320 alignment=0 units_per_block=818, header_type=1, class_id=3
class 4 unit_size=448 alignment=0 units_per_block=584, header_type=1, class_id=4
class 5 unit_size=704 alignment=0 units_per_block=372, header_type=1, class_id=5
class 6 unit_size=832 alignment=0 units_per_block=315, header_type=1, class_id=6
....

So, for evtree node, which is presently 768 bytes, we would be in the 832 byte class.   In my PR, we would be in the 768 class in this case, wasting at most 31 bytes in any given allocation to internal fragmentation.

For a point of reference, these are all of the custom ones defined with this patch
class 54 unit_size=32 alignment=0 units_per_block=8158, header_type=2, class_id=54
class 55 unit_size=64 alignment=0 units_per_block=4087, header_type=2, class_id=55
class 56 unit_size=96 alignment=0 units_per_block=2726, header_type=2, class_id=56
class 57 unit_size=128 alignment=0 units_per_block=2045, header_type=2, class_id=57
class 58 unit_size=160 alignment=0 units_per_block=1636, header_type=2, class_id=58
class 59 unit_size=192 alignment=0 units_per_block=1364, header_type=2, class_id=59
class 60 unit_size=224 alignment=0 units_per_block=1169, header_type=2, class_id=60
class 61 unit_size=288 alignment=0 units_per_block=1819, header_type=2, class_id=61
class 62 unit_size=384 alignment=0 units_per_block=1364, header_type=2, class_id=62
class 63 unit_size=416 alignment=0 units_per_block=1259, header_type=2, class_id=63
class 64 unit_size=576 alignment=0 units_per_block=1365, header_type=2, class_id=64
class 65 unit_size=672 alignment=0 units_per_block=1170, header_type=2, class_id=65
class 66 unit_size=768 alignment=0 units_per_block=1023, header_type=2, class_id=66

I plan to have fewer allocation classes (that we actually use) in a subsequent patch to try to avoid allocating too many slabs and also try to avoid alignment issues due to weird sizes.

NiuYawei: Sounds reasonable to me, thanks for looking into PMDK allocation classes.

NiuYawei (approved):

src/vos/vos_pool.c
```diff
@@ -519,93 +519,27 @@ vos_pool_destroy(const char *path, uuid_t uuid)
...
-				D_ASSERT(slab->class_id != 0);
-				break;
-			}
-		}
-
-		if (skip_set)
-			continue;
+		slab->alignment       = 0;
+		slab->unit_size       = (defined + 1) * 32;
+		slab->units_per_block = 1000;
```
gnailzenh, line 541: what is the default if we don't set this one?

gnailzenh (approved):

src/vos/vos_pool.c
jolivier23: It is a minimum specification.  If you look at my reply to Niu's comment, it looks like 256K is the default for all of the internal types.  Maybe we should follow suit or let PMDK pick.

PR #11374, opened 2023-02-05
DAOS-10681 container: add sc_discarding

wangdi1:
During reclaim or failure reclaim, the object are being discarded from the container, the aggregation should be aborted to let discarding proceed, to avoid EBUSY.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/container/srv_target.c
+++ b/src/container/srv_target.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -53,6 +53,12 @@ agg_rate_ctl(void *arg)
 
 	/* Abort current round of aggregation */
 	if (dss_ult_exiting(req) || pool->sp_reclaim == DAOS_RECLAIM_DISABLED)
+		return -1;
+
+	/* If the container is discarding the object mostly due to rebuild failure
+	 * reclaim, let's abort the aggregation to let discard proceed.
+	 */
+	if (cont->sc_discarding)
 		return -1;
 
 	/* System is idle, let aggregation run in tight mode */
--- a/src/include/daos_srv/container.h
+++ b/src/include/daos_srv/container.h
@@ -70,7 +70,8 @@ struct ds_cont_child {
 				 sc_stopping:1,
 				 sc_vos_agg_active:1,
 				 sc_ec_agg_active:1,
-				 sc_scrubbing:1;
+				 sc_scrubbing:1,
+				 sc_discarding:1;
 	uint32_t		 sc_dtx_batched_gen;
 	/* Tracks the schedule request for aggregation ULT */
 	struct sched_request	*sc_agg_req;
--- a/src/rebuild/scan.c
+++ b/src/rebuild/scan.c
@@ -787,6 +787,7 @@ rebuild_container_scan_cb(daos_handle_t ih, vos_iter_entry_t *entry,
 	vos_iter_param_t		param = { 0 };
 	struct vos_iter_anchors		anchor = { 0 };
 	daos_handle_t			coh;
+	struct ds_cont_child		*cont_child = NULL;
 	struct dtx_id			dti = { 0 };
 	struct dtx_epoch		epoch = { 0 };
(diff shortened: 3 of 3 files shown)


src/container/srv_target.c
```diff
@@ -55,6 +55,12 @@ agg_rate_ctl(void *arg)
 	if (dss_ult_exiting(req) || pool->sp_reclaim == DAOS_RECLAIM_DISABLED)
 		return -1;
 
+	/* If the container is discarding the object mostly due to rebuild failure
+	 * reclaim, let's abort the aggregation to let discard proceed.
+	 */
+	if (cont->sc_discarding)
+		return 1;
+
```
NiuYawei, line 63: This won't abort aggregation.

I remembered that you introduced two bits in ds_cont_child to indicate if the VOS or EC aggregation is running, why don't you reuse that two bits?

NiuYawei (changes requested):

src/container/srv_target.c
```diff
@@ -55,6 +55,12 @@ agg_rate_ctl(void *arg)
 	if (dss_ult_exiting(req) || pool->sp_reclaim == DAOS_RECLAIM_DISABLED)
 		return -1;
 
+	/* If the container is discarding the object mostly due to rebuild failure
+	 * reclaim, let's abort the aggregation to let discard proceed.
+	 */
+	if (cont->sc_discarding)
+		return 1;
```
liuxuezhao, line 62: seems return -1 is to abort?

src/container/srv_target.c
wangdi1: ah, it should return -1 here. Hmm, which bits you mean? I do not remember 

These are bits in ds_cont_child
 uint32_t                 sc_dtx_resyncing:1,
                                 sc_dtx_reindex:1,
                                 sc_dtx_reindex_abort:1,
                                 sc_props_fetched:1,
                                 sc_stopping:1,
                                 sc_vos_agg_active:1,
                                 sc_ec_agg_active:1,
                                 sc_scrubbing:1,
                                 sc_discarding:1;
        uint32_t                 sc_dtx_batched_gen;

Besides indicating VOS/EC aggregation won't help here. Since the intention is to let aggregation abort, and let discard proceed, i.e. favor discarding.

src/container/srv_target.c
wangdi1: ah, right. thanks

src/container/srv_target.c
NiuYawei: sc_vos_agg_active & sc_ec_agg_active. I thought you implemented this for caller to wait for agg ULT exited, but seems it was changed in master somehow.

Returning -1 here isn't correct, since the discard needs to wait for the aggregation ULT exited and aggregation ULT could yield at various places. (agg_rate_ctl() is only one of them).

NiuYawei (changes requested):

src/container/srv_target.c
wangdi1: Actually, this does not have to abort the aggregation, since vos_discard() will return -DER_BUSY if there is vos_aggregation on the same container. see aggregate_enter().  

So this patch is trying to let vos_aggregate() exit asap (aggregate_exit() is called), and using sc_vos(ec)_agg_active() won't make this exit quicker.

NiuYawei: The commit comment mentioned the goal is to avoid EBUSY, I was thinking that EBUSY wasn't tolerated. (I checked rebuild code now and seems it does retry on EBUSY)

Do we allow EC aggregation running?

Instead of keep adding bits into ds_cont_child for the same purpose, I strongly suggest we incorporate them into few common bits.

NiuYawei (approved):

liuxuezhao (approved):

src/container/srv_target.c
wangdi1: "Do we allow EC aggregation running?"

Yes, EC aggregation can be done with vos_discard at the same time. At least I do not see the reason to exclude them yet.

"Instead of keep adding bits into ds_cont_child for the same purpose, I strongly suggest we incorporate them into few common bits."

Hmm, this discarding state is different with vos/ec_agg_active, which means vos/ec agg state, but discarding state is used to notify the agg ULT to abort the agg. I do not see other related status in ds_cont_child(), though I totally agree with you for generalizing the common bits. @NiuYawei

PR #11400, opened 2023-02-08
DAOS-12570 test: multi-engine per socket test for the latest DAOS changes

dinghwah:
Description:
Update test with: daos_server scm reset and prepare –force --scm-ns-per-socket 2

Allow-unstable-test: true
Test-tag: server,-manual test_container_basics
Required-githooks: true
Signed-off-by: Ding Ho ding-hwa.ho@intel.com

### Gatekeeper:


tanabarr (approved):

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -246,6 +247,7 @@ def test_multiengines_per_socket(self):
             # Check for agent start status
             if results[0]['exit_status'] and "sudo systemctl" in cmd:
                 self.fail("#Fail on {0}".format(cmd))
+        time.sleep(15)
```
tanabarr, line 250: this should be waiting for ranks to join in a deterministic way, not just waiting for a static length of time, as it is this is brittle and may break (there is no guarantee that the ranks will join within this time)

tanabarr (changes requested):
I think it would be more reliable to check the output of query and run in a loop (with maybe a sleep between iterations) until all the expected ranks have joined before creating a pool. I imagine there are other places in test code that perform this that you can use as a reference.

src/tests/ftest/server/multiengine_persocket.py
dinghwah: Thanks for the review, 
the reason to use a fixed delay is due to after "daos_server scm prepare --force --scm-ns-per-socket=2", and "dmg storage format", observing 2 server-ranks joined, and some time 4 server-ranks joined.

tanabarr: right so I am suggesting that you wait until 4 ranks have joined before continuing by looping over the results of system query and not continuing until 4 ranks are detected

dinghwah: Issue is script configure server with "--scm-ns-per-socket=2", but detected with 2 ranks or 4 ranks up occasionally,.

rpadma2: @dinghwah : So you are saying sometimes only 2 ranks are detected (however long you wait?)... If so, is it not a bug?

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -201,6 +199,8 @@ def test_multiengines_per_socket(self):
         self.daos_server_scm_reset(step)
         engines_per_socket = self.params.get(
             "engines_per_socket", "/run/server_config/*", default=1)
+        reboot_after_format = self.params.get(
+            "reboot_after_format", "/run/server_config/*", default=False)
         self.daos_server_scm_prepare_ns(step, engines_per_socket)
```
tanabarr, line 204: @dinghwah let's try this just to see if we can get it behaving consistently, then we can tidy up after. The multiple reboots will mean we clear and recreate the PMem regions as well as namespaces, hopefully this will ensure the naming of the namespaces ndctl create are consistent each time.
```suggestion
        self.daos_server_scm_reset(step)
        time.sleep(15)
        self.host_reboot(self.hostlist_servers)
        engines_per_socket = self.params.get(
            "engines_per_socket", "/run/server_config/*", default=1)
        self.daos_server_scm_prepare_ns(step, engines_per_socket)
        time.sleep(15)
        self.host_reboot(self.hostlist_servers)
        self.daos_server_scm_prepare_ns(step, engines_per_socket)
        time.sleep(15)
        reboot_after_format = self.params.get(
            "reboot_after_format", "/run/server_config/*", default=False)
```

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -218,6 +218,8 @@ def test_multiengines_per_socket(self):
             self.fail(
                 "#PMem did not show after scm prepare --scm-ns-per-socket and {0} "
                 "retries".format(max_retry))
+        if reboot_after_format:
```
tanabarr, line 221: reboot shouldn't be needed here

src/tests/ftest/server/multiengine_persocket.py
tanabarr: This should always be 4, I presume this is because of the issue where namepaces are created with unexpected names e.g.  0, 0.2, 1, 1.2 instead of the usual 0, 0.1, 1, 1.1. @dinghwah Could you please try with the changes I've suggested above and see if the same issue occurs?

src/tests/ftest/server/multiengine_persocket.py
dinghwah: If it is needed 2 "daos_server  scm prepare --force --scm-ns-per-socket=2" ?

dinghwah: I updated the script with following steps: (2  self.daos_server_scm_prepare_ns) the pmem0.1 and pmem1.1 came up consistently.
step (1.A): daos_server scm reset --force
                 server reboot  <-----  This reboot can be removed.
step (1.1B): daos_server scm prepare –force --scm-ns-per-socket 2
                 server reboot
step (1.2B): daos_server scm prepare –force --scm-ns-per-socket 2

dinghwah: Please see the newly updated script. 
Thanks.

tanabarr (dismissed):

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -196,28 +195,32 @@ def test_multiengines_per_socket(self):
...
+        # (1) Scm reset and prepare --scm-ns-per-socket
         step = 1
-        self.log.info("===(%s)===Storage prepare --scm-ns-per-socket", step)
-        self.daos_server_storage_prepare_reset(step)
+        self.log.info("===(%s)===Scm reset and prepare --scm-ns-per-socket", step)
         engines_per_socket = self.params.get(
             "engines_per_socket", "/run/server_config/*", default=1)
-        self.daos_server_storage_prepare_ns(step, engines_per_socket)
+        self.daos_server_scm_reset(step)
+        self.host_reboot(self.hostlist_servers)
```
rpadma2, line 204: This method shutdown the node sleep for 300 seconds and assumes the node is booted. There is no check to verify whether the node is really active before proceeding to next command. This can cause issue... Maybe, think about it for future enhancement in the script. Since this is a manual test case, it is fine for now. For large scale testing, it will be nice to verify the nodes status.

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -135,21 +136,19 @@ def daos_server_storage_prepare_ns(self, step, engines_per_socket=1):
         cmd.sudo = False
         cmd.debug.value = False
         cmd.set_sub_command("scm")
-        cmd.sub_command_class.set_sub_command("reset")
+        cmd.sub_command_class.set_sub_command("prepare")
```
rpadma2, line 139: It will be nice to implement the sub-commands under daos_utils.py, daos_utils_base.py. In this way, other tests cases can utilize the subcommands as needed. Another enhancement to be considered for future.

dinghwah: Yes. This test to make sure the first multiple-engine per socket tests stabilize, and could be update to ../ftest/util/. in the future.

rpadma2 (dismissed):

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -123,9 +123,10 @@ def daos_server_storage_prepare_reset(self, step):
             self.fail(
                 "#({0}.A){1} failed, "
                 "please make sure the server equipped with PMem modules".format(step, cmd))
+        time.sleep(5)
```
phender, line 126: Is there a more deterministic way to determine the test can continue here?  We should avoid arbitrary sleeps in test code.  

Also checking `results[0]['exit_status']` is ill-advised as there could be multiple results (max of one per self.hostlist_servers) even if right now the test only uses one server host - as launch.py can easily override this.

Also using 'run_remote' in place of 'run_pcmd' would make checking the results easier as well as defining a DaosServerCommand.scm_seset() helper function.

src/tests/ftest/server/multiengine_persocket.py
phender: +1 to Ravi's comment and we should also move to use `run_remote()` instead of `run_pcmd()` for ease of use.

phender (changes requested):
All sleeps should be replaced my deterministic means of determining that the test can proceed or if it should fail at that point.

shimizukko (dismissed):

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -158,9 +158,25 @@ def host_reboot(self, hosts):
...
+            self.log.info("===>ping_result= %s", ping_result)
+            ping_status = re.search("1 received", ping_result)
+            if (time.time() - start) < 120:
+                time.sleep(5)
+            else:
+                time.sleep(1)
+        if time.time() - start >= ping_timeout:
+            self.fail("#{0} reboot failed, did not come back after {1} seconds",
+                      hosts, ping_timeout)
+        time.sleep(5)
```
tanabarr, line 178: why is this extra sleep needed, shouldn't hosts be up now?

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -231,9 +247,9 @@ def test_multiengines_per_socket(self):
             "sudo cp /etc/daos/daos_server.yml_4 /etc/daos/daos_server.yml",
             "sudo systemctl start daos_server.service"]
         for cmd in start_server_cmds:
-            results = run_pcmd(self.hostlist_servers, cmd, timeout=90)
+            results = run_remote(self.log, self.hostlist_servers, cmd, timeout=90)
         # Check for server start status
-        if results[0]['exit_status']:
+        if not results.passed:
```
tanabarr, line 252: shouldn't this be indented to be checking each of the commands that are iterated through in the above for block succeeded?

tanabarr (approved):
The improvements look like they make the test more deterministic, just a couple of suggestions. Phil's approval should be granted before landing.

src/tests/ftest/server/multiengine_persocket.py
dinghwah: This is expected, after server reboot, ping will still succeed, and then disabled, after server reboot completed, ssh will succeed with password, and broke the script. Extra sleep to allow ssh with password-less and continue the test.

tanabarr (dismissed):

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -159,9 +158,25 @@ def host_reboot(self, hosts):
...
         cmd = "sudo shutdown -r now"
-        run_pcmd(hosts, cmd, timeout=210)
+        run_remote(self.log, hosts, cmd, timeout=210)
         self.log.info("===Server %s rebooting... \n", hosts)
         time.sleep(reboot_waittime)
+        ping_timeout = 600
+        start = time.time()
+        ping_status = None
+        while ping_status is None and (time.time() - start) < ping_timeout:
+            ping_result = run_local(self.log, ["ping", "-c 1", str(hosts)], check=False).stdout
```
phender, line 168: I'm fairly certain this node designation is not supported for the ping command.  If `hosts`, being a NodeSet object, contains multiple hosts, `str(hosts)` will give you something like `wolf-[118-121]`.  We may need to issue multiple ping commands - one per host - where we could reduce the list of hosts for each host that responds.

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -159,9 +158,25 @@ def host_reboot(self, hosts):
...
+            self.log.info("===>ping_result= %s", ping_result)
+            ping_status = re.search("1 received", ping_result)
+            if (time.time() - start) < 120:
+                time.sleep(5)
+            else:
+                time.sleep(1)
+        if time.time() - start >= ping_timeout:
+            self.fail("#{0} reboot failed, did not come back after {1} seconds",
+                      hosts, ping_timeout)
+        time.sleep(5)
```
phender, line 178: We should not use a sleep here.  Instead, we could clush (`run_remote()`) a basic command, like `uname`, in a while loop until it passes on all the hosts or hits a timeout (to avoid an invite loop).

src/tests/ftest/server/multiengine_persocket.py
phender: Yes it should
```suggestion
            if not results.passed:
```

phender (changes requested):

src/tests/ftest/server/multiengine_persocket.yaml
```diff
@@ -76,7 +76,14 @@ server_config:
...
           scm_mount: /mnt/daos4
-
+  transport_config:
+    allow_insecure: True
+agent_config:
+  transport_config:
+    allow_insecure: True
+dmg:
+  transport_config:
+    allow_insecure: True
```
shimizukko, line 86: Why do we use `allow_insecure: True`? I thought we should avoid insecure mode unless it's absolutely necessary.

shimizukko (dismissed):

src/tests/ftest/server/multiengine_persocket.py
ashleypittman: there are 'fping' and similar commands for doing just this in parallel.

src/tests/ftest/server/multiengine_persocket.yaml
tanabarr: maybe something to do with it being to do with a manual test?

shimizukko (approved):

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -196,28 +217,32 @@ def test_multiengines_per_socket(self):
...
-        self.daos_server_storage_prepare_ns(step, engines_per_socket)
+        self.daos_server_scm_reset(step)
+        self.host_reboot(self.hostlist_servers)
+        self.daos_server_scm_prepare_ns(1.1, engines_per_socket)
+        self.host_reboot(self.hostlist_servers)
+        self.daos_server_scm_prepare_ns(1.2, engines_per_socket)
+
         cmd = "/usr/bin/ls -l /dev/pmem*"
-        results = run_pcmd(self.hostlist_servers, cmd, timeout=90)
+        results = run_remote(self.log, self.hostlist_servers, cmd, timeout=90)
```
phender, line 232: Replace lines 231-246 with:
```
    if not self.wait_for_result(self, self.check_pmem, 100, hosts=hosts, count=4, cmd_timeout=10, verbose=True):
            self.fail("4 pmem devices not found on all hosts within 100 seconds.")
```

Which will require the addition of this method:
```
    def check_pmem(self, hosts, count, cmd_timeout=10, verbose=True):

        storage = StorageInfo(self.log, hosts)
        try:
            storage.scan()            
        except StorageException:
            self.log.debug("Problem running StorageInfo.scan()"
        return (len(storage.pmem_devices) == count)
```

phender (changes requested):

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -118,50 +120,69 @@ def daos_server_storage_prepare_reset(self, step):
...
+            self.log.info("===>ping_result= %s", ping_result)
+            if expect_pass:
+                ping_status = re.search("1 received", ping_result)
+            else:
+                ping_status = re.search("0 received", ping_result)
+            if (time.time() - start) < 120:
+                time.sleep(20)
+            else:
+                time.sleep(1)
+        if time.time() - start >= ping_timeout:
```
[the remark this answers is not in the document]
dinghwah: done.

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -118,50 +120,69 @@ def daos_server_storage_prepare_reset(self, step):
...
         self.log.info("===Server %s rebooting... \n", hosts)
-        time.sleep(reboot_waittime)
+
+        self.ping_verify(hosts[0], expect_pass=False)
+        self.ping_verify(hosts[0], expect_pass=True)
+        cmd = "sudo uname"
+        if not run_remote(self.log, hosts, cmd, timeout=600).passed:
+            self.fail(
+                "#{0} reboot failed, sudo uname failed.".format(hosts))
+        self.log.info("===Server %s is up after reboot. \n", hosts)
```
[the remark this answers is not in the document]
dinghwah: done.

shimizukko (approved):

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -135,33 +134,112 @@ def daos_server_storage_prepare_ns(self, step, engines_per_socket=1):
...
+        start = time.time()
+
+        while not result_found and not timed_out:
+            timed_out = (time.time() - start) >= timeout
+            result_found = get_method(**kwargs)
+            self.log.debug("==Response found by %s(%s): %s", get_method, kwargs, result_found)
+            time.sleep(5)
+        return not timed_out
+
+    def check_ping(self, host, expected_pings=0, cmd_timeout=60, verbose=True):
```
rpadma2, line 172: My general recommendation would be to move wait_for_result, check_ping and check_ssh to a utils file. In this way, it can be utilized by other test cases.

src/tests/ftest/server/multiengine_persocket.py
```diff
@@ -135,33 +134,112 @@ def daos_server_storage_prepare_ns(self, step, engines_per_socket=1):
...
+            expected_pings (int, optional): number of succeed ping expected. default is 0.
+            cmd_timeout (int, optional): number of seconds to wait for a response to be found.
+            verbose (bool, optional): display check ping commands. Defaults to True.
+
+        Returns:
+            bool:  True if the expected number of pings were returned; False otherwise.
+        """
+        try:
+            result = run_local(
+                self.log, ["ping", "-c 1", str(host)], check=False,
```
ashleypittman, line 186: This is almost certainly wrong, I suspect you mean this:
```suggestion
                self.log, ["ping", "-c",  "1", str(host)], check=False,
```

How about passing the timeout onto the ping command with the -W option and using the returncode to determine success?

This function takes a "expected_pings" but only sends one ping so valid values that can be checked here are only 0 and 1 meaning the only thing this function can do if confirm that a host is up so you could remove complexity and use the return code.

phender, line 186: The ping command should be string.  If you are running this manually make sure you have up to date code.  We have been passing the command as a string since https://github.com/daos-stack/daos/commit/a63d9da754ec569b3eb0fe68f01c7c73535a52cb

phender (changes requested):

src/tests/ftest/server/multiengine_persocket.py
dinghwah: Please see the update.

src/tests/ftest/server/multiengine_persocket.py
dinghwah: Please see the update.

src/tests/ftest/server/multiengine_persocket.py
dinghwah: Please see the latest update with script fixes.

shimizukko (approved):

phender (approved):

PR #11412, opened 2023-02-09
DAOS-11624 test: convert str for stdout

wangshilong:
Required-githooks: true

### DIFF
--- a/src/tests/ftest/util/server_utils.py
+++ b/src/tests/ftest/util/server_utils.py
@@ -1195,7 +1195,7 @@ class DaosServerManager(SubprocessManager):
             result = run_remote(self.log, host, f"grep -E '{pattern}' {' '.join(log_files)}")
             for data in result.output:
                 if data.returncode == 0:
-                    matches = re.findall(fr'{pattern}', data.stdout)
+                    matches = re.findall(fr'{pattern}', str(data.stdout))
                     log_file_matches += len(matches)
             self.log.debug("Found %s matches on %s", log_file_matches, host)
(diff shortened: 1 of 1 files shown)


tanabarr (approved):

src/tests/ftest/util/server_utils.py
```diff
@@ -1195,7 +1195,7 @@ def search_log(self, pattern):
             result = run_remote(self.log, host, f"grep -E '{pattern}' {' '.join(log_files)}")
             for data in result.output:
                 if data.returncode == 0:
-                    matches = re.findall(fr'{pattern}', data.stdout)
+                    matches = re.findall(fr'{pattern}', str(data.stdout))
```
daltonbohning, line 1198: FYI that `data.stdout` is a list of output lines. E.g. `str(list)` looks like:
```
>>> str(['foo', 'bar'])
"['foo', 'bar']"
```
A more proper search is probably to convert to a single string of output:
```suggestion
                    matches = re.findall(fr'{pattern}', '\n'.join(data.stdout))
```

wangshilong: Sure, @phender mentioned that he is not going to add such kind of check and trying to use different approach, i am not sure if this PR is needed or not. or possibly reasonable for a quick fix.

@phender would you clarity it here?

wangshilong: @daltonbohning Feel free to help refresh if we still want to land this simple fix.

PR #11435, opened 2023-02-10
DAOS-12610 vos: Add iterator API to validate current position

jolivier23:
In some cases, we may yield in the middle of a callback and need to request a full reprobe.  The checksum scrubber is such a case. It could, in theory, read a value, start processing it for checksum scrubbing, and have the value deleted before it is able to continue.  If this happens, we can end up reading bad data, or worse, causing a system crash.

Features: scrubber

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/include/daos_srv/vos.h
+++ b/src/include/daos_srv/vos.h
@@ -1035,6 +1035,21 @@ int
 vos_iter_empty(daos_handle_t ih);
 
 /**
+ * When the callback executes code that may yield, in some cases, it needs to
+ * verify the value or key it is operating on still exists before continuing.
+ * This function will revalidate from the object layer down.   This is only
+ * supported in conjunction with recursive vos_iterate from VOS_ITER_OBJ.
+ *
+ * \param[in]	ih	The iterator handle
+ *
+ * \return		0 if iterator value is valid
+ *			VOS_ITER_* level that needs probe
+ *			<0 on error
+ */
+int
+vos_iter_validate(daos_handle_t ih);
+
+/**
  * Iterate VOS entries (i.e., containers, objects, dkeys, etc.) and call \a
  * cb(\a arg) for each entry.
  *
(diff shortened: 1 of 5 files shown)


jolivier23:
I ran this with scrubber feature and I ran my evtree patch on top of it here.  Both passed.  This could probably use a bit more targeted testing but I'd like to finish the evtree patch first and this will make it more stable.

https://github.com/daos-stack/daos/pull/11436

src/include/daos_srv/vos.h
```diff
@@ -1034,6 +1034,21 @@ vos_iter_process(daos_handle_t ih, vos_iter_proc_op_t op, void *args);
 int
 vos_iter_empty(daos_handle_t ih);
 
+/**
+ * When the callback executes code that may yield, in some cases, it needs to
+ * verify the value or key it is operating on still exists before continuing.
+ * This function will revalidate from the object layer down.   This is only
+ * supported in conjunction with recursive vos_iterate from VOS_ITER_OBJ.
```
NiuYawei, line 1041: It looks to me an issue existing in EC aggregation & rebuild enumeration as well: when EC aggregation ULT or rebuild enumeration ULT yield in iter callback, and VOS aggregation deleted it's parent, EC aggregation or rebuild enumeration will resume to work in an deleted sub-tree, right?

I'm wondering if we need some sort of refcount mechanism to avoid the race of sub-tree deletion and sub-tree iteration (which could yield)?

src/vos/vos_pool_scrub.c
```diff
@@ -359,6 +358,16 @@ sc_handle_corruption(struct scrub_ctx *ctx)
 {
 	int rc;
 
+	/** It's ok if we do the checksum calculation after a yield, hoping for the best, but we
+	 *  absolutely must check before modifying data at the current iterator position.  If the
+	 *  entry has been deleted, we can ignore any corruption we found and move on.
+	 */
```
NiuYawei, line 364: Indeed, if an iter callback yield then continue to access the tree namespace, seems it has to explicitly call some validate API to see if the tree or parent tree has been deleted.

ryon-jensen (approved):
These changes look good to me. Do rebuild and EC aggregation also need to use vos_iter_validate?

src/include/daos_srv/vos.h
jolivier23: I don't think it can happen with EC aggregation because it operates on a different epoch range.  We can't do VOS aggregation on a given range until EC aggregation is complete so I don't think we can ever hit this situation where we delete something that is being operated on.

Rebuild might be different.  @wangdi1 said that it's also not possible but I'm not entirely convinced because unlike EC aggregation, it's working at a lower epoch range than vos aggregation.

NiuYawei (approved):

PR #11458, opened 2023-02-15
DAOS-8772 vos: space reserve for rebuild

NiuYawei:
When the pool property of DAOS_PROP_PO_SPACE_RB is configured, VOS internally will reserve certain amount of space for rebuild by rejecting client update request when the free space is lower than reserved space.

Caveat: The NVMe reserved space for rebuild could be used up by aggregation since the aggregation might migrate data on SCM to NVMe regardless of the reserving for rebuild.

Required-githooks: true

### Gatekeeper:


NiuYawei:
@wangdi1 , could you take a look?

src/vos/vos_space.c
```diff
@@ -303,19 +303,42 @@ vos_space_hold(struct vos_pool *pool, uint64_t flags, daos_key_t *dkey,
...
+		nvme_left -= NVME_SYS(&vps);
+		/* 'NVMe held' has already been excluded from 'NVMe free' */
 
-	if (nvme_left < space_est[DAOS_MEDIA_NVME])
-		goto error;
+		if (nvme_left < space_est[DAOS_MEDIA_NVME])
+			goto error;
+	}
+
+	/* Check space reserve for rebuild */
```
wangdi1, line 320: This is to make sure the non-rebuild IO will not use the space to reserve for rebuild? I assume other background ULT will need reserve the space as well? 
So maybe use general name here? (I mean VOS_OF_REBUILD and vp_space_rb).

src/vos/vos_pool.c
```diff
@@ -988,6 +988,17 @@ vos_pool_ctl(daos_handle_t poh, enum vos_pool_opc opc, void *param)
...
+	case VOS_PO_CTL_SET_SPACE_RB:
+		if (param == NULL)
+			return -DER_INVAL;
+
+		i = *((unsigned int *)param);
+		if (i >= 100 || i < 0) {
+			D_ERROR("Invalid space reserve ratio for rebuild. %d\n", i);
+			return -DER_INVAL;
+		}
+		pool->vp_space_rb = i;
```
wangdi1, line 1000: Just curious, what is the default % for vp_space_rb?

src/vos/vos_space.c
```diff
@@ -303,19 +303,42 @@ vos_space_hold(struct vos_pool *pool, uint64_t flags, daos_key_t *dkey,
...
+				      space_est[DAOS_MEDIA_SCM])) {
+			D_ERROR("Insufficient rebuild reserved SCM space for Pool:"DF_UUID".\n",
+				DP_UUID(pool->vp_id));
+			goto error;
+		}
+
+		if (pool->vp_vea_info == NULL || space_est[DAOS_MEDIA_NVME] == 0)
+			goto success;
+
+		rb_reserve = NVME_TOTAL(&vps) * pool->vp_space_rb / 100;
```
wangdi1, line 334: Do you think we need different ratio for nvme and scm?

wangdi1 (dismissed):
probably add a test case.

src/vos/vos_pool.c
NiuYawei: default value is 0, which means no reserve for rebuild.

src/vos/vos_space.c
NiuYawei: This check is for rebuild only and this vos_space_hold() is called by update request processing only.

src/vos/vos_space.c
NiuYawei: Looks not necessary to me, please see my commit comment, we actually even can't honor NVMe reserve in some cases, but I don't know how to improve it so far.

src/vos/vos_space.c
wangdi1: I mean it can be 

if (!(flags & VOS_USE_RESERVE_SPC) {

}. 
Since these seems very general checking, nothing related with the rebuild except the flag?

And also the comments is a bit confused, since this is actually for checking no-rebuild case? These are not logical fault.

jolivier23 (dismissed):

src/vos/vos_pool.c
kccain: (not sure what if anything to change, so just discussion)

specifying 100 or a very high percentage would probably be crazy and make the pool inoperable for IO, right? 
I guess vos_space_hold() could have a large volume of D_ERROR for things like "Insufficient rebuild reserved SCM space for Pool" (or the similar NVMe-related ERR message) - for example if the pool property value is changed while there is a lot of IO happening, and the percentage is too high?

NiuYawei: right, if the value is set to 100 or a very high percentage, there could lots of I/O failures with the error messages you mentioned.

src/vos/vos_space.c
NiuYawei: This isn't the general checking for "sys reserve", this is the special checking for "rebuild reserve".

"sys reserve" is to reserve space for truncate/destroy/aggregation etc. It's for ensuring the space reclaim can still work even when free space is tight, "sys reserve" is mandatory and not configurable.

"rebuild reserve" is to reserve space for potential rebuild purpose, it's configurable. A user who doesn't use redundant objects should obviously configure the "rebuild reserve" as zero for better space utilization.

src/vos/vos_space.c
```diff
@@ -303,19 +303,42 @@ vos_space_hold(struct vos_pool *pool, uint64_t flags, daos_key_t *dkey,
...
+			goto error;
+	}
+
+	/* Check space reserve for rebuild */
+	if (!(flags & VOS_OF_REBUILD) && pool->vp_space_rb != 0) {
+		rb_reserve = SCM_TOTAL(&vps) * pool->vp_space_rb / 100;
+
+		if (SCM_FREE(&vps) < (rb_reserve + POOL_SCM_HELD(pool) +
+				      space_est[DAOS_MEDIA_SCM])) {
+			D_ERROR("Insufficient rebuild reserved SCM space for Pool:"DF_UUID".\n",
```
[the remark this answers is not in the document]
NiuYawei: sure, thanks.

kccain:
> (discussion only)
> 
> The other thing that I thought about is whether the space set aside for rebuild could be included as part of the calculation of the system reserved space (POOL_SCM_SYS, POOL_NVME_SYS). So that the logic in vos_space_hold() could be roughly the same as it is now. i.e., would it make sense for rebuild, when the DAOS_PROP_PO_SPACE_RB property is set to calculate the number of bytes and then call vos_space_sys_set() to add it to the "system reserved' calculation for the vos pool?
> 
> For example rdb uses vos_space_sys_set() to add more to the system reserved space in a vos pool with its own additional requirements (leave some space for raft INSTALLSNAPSHOT message handling in a logical "staging" area of the vos pool). And, when performing these rdb updates it uses the VOS_OF_CRIT flag bit to specify that vos_space_hold() should not fail the update due to free space calculations.

Chatted with Niu on this idea. rebuild reserved needs to be separate from the system reserved calculation.
- both normal IO and rebuild IO should fail if free space falls below system reserved
- then, normal IO should fail if free space falls below rebuild reserved

In that other ticket I will be re-evaluating the approach for the rdb use case, the part of it that needs to logically set aside > 50% of vos pool space for raft installsnapshot (should it be needed).

NiuYawei:
rebase to pick up fix from DAOS-12709

kccain (approved):

wangdi1 (approved):

johannlombardi:
We need to talk with test team to add some specific test case for that.

PR #11478, opened 2023-02-17
DAOS-12390 test: Bump test timeouts with UCX.

ashleypittman:
These tests run slower under UCX so bump the timeout accordingly.

Test-tag: DaosBuild
Test-provider-hw-medium: ucx+dc_x

Required-githooks: true


daltonbohning (commented):
The bumped timeout also applies to the tests left in daily_regression. Which could be a problem if those had issues for some reason. Should we just move all of these tests to weekly?

ashleypittman:
> The bumped timeout also applies to the tests left in daily_regression. Which could be a problem if those had issues for some reason. Should we just move all of these tests to weekly?

I don't think it does as the timeout is in a conditional based on the arg passed in.  We always want the in-test timeout to be shorter than the yaml timeout as if the in-test one fails it does some diagnostics, the yaml timeout is really just to prevent stuck jobs.

I'd like to run at least one of the four combination daily so we can monitor for breakage still.

daltonbohning:
> I don't think it does as the timeout is in a conditional based on the arg passed in. We always want the in-test timeout to be shorter than the yaml timeout as if the in-test one fails it does some diagnostics, the yaml timeout is really just to prevent stuck jobs.
> 
> I'd like to run at least one of the four combination daily so we can monitor for breakage still.

E.g. `test_dfuse_daos_build_wt_il` runs in mode `writethrough` which does use `cache_time = '2d'`. Also, the `timeout: 36000` in the config applies to every test in the python file

ashleypittman:
DaosBuild tests all passed with ucx, results at https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-11478/3/testReport/FTEST_dfuse/DaosBuild/

daltonbohning (approved):

jamesanunez (approved):

PR #11480, opened 2023-02-17
Revert "DAOS-12314 build: change ipmctl from v2 to v3 (#11404)"

phender:
Reverts daos-stack/daos#11431


ashleypittman (changes requested):
I don't believe you can do a simple revert of packaging changes, you have to bump the version number and add further changelogs whilst applying (reverting) the other changes you want.

PR #11505, opened 2023-02-22
DAOS-12612 pool: fixes for checkpoint

wangshilong:
1. @chkpt_ctx need be set and cleared properly.
2. ABT_eventual_set() accept ABT_eventual rather than ABT_eventual *, refine
to reuse ABT_eventual

Required-githooks: true

Signed-off-by: Niu Yawei <yawei.niu@intel.com>


NiuYawei (approved):

NiuYawei:
I'm afraid this defect could be the root cause of pool destroy timeout failures, since the checkpoint ULT is waiting in wait_cb() and never be woken up.

gnailzenh (approved):

PR #11539, opened 2023-02-23
DAOS-11178 container: move IV update out of svc lock

wangdi1:
Move conainer prop and snapsht IV sync out of svc_lock.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/container/srv_container.c
+++ b/src/container/srv_container.c
@@ -4562,6 +4562,53 @@ out:
 	return rc;
 }
 
+void
+ds_cont_prop_iv_update(struct cont_svc *svc, uuid_t cont_uuid)
+{
+	struct rdb_tx	tx;
+	struct cont	*cont = NULL;
+	daos_prop_t	*prop = NULL;
+	int		rc;
+
+	/* Only happens on xstream 0 */
+	D_ASSERT(dss_get_module_info()->dmi_xs_id == 0);
+	rc = rdb_tx_begin(svc->cs_rsvc->s_db, svc->cs_rsvc->s_term, &tx);
+	if (rc != 0) {
+		D_ERROR(DF_UUID": Failed to start rdb tx: %d\n",
+			DP_UUID(svc->cs_pool_uuid), rc);
+		return;
+	}
+
+	ABT_rwlock_rdlock(svc->cs_lock);
+	rc = cont_lookup(&tx, svc, cont_uuid, &cont);
+	if (rc != 0) {
+		D_ERROR(DF_CONT": Failed to look container: %d\n",
+			DP_CONT(svc->cs_pool_uuid, cont_uuid), rc);
+		D_GOTO(out_lock, rc);
+	}
+
+	rc = cont_prop_read(&tx, cont, DAOS_CO_QUERY_PROP_ALL, &prop, true);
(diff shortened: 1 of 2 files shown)


src/container/srv_container.c
```diff
@@ -4649,8 +4696,12 @@ cont_op_with_svc(struct ds_pool_hdl *pool_hdl, struct cont_svc *svc,
...
 out:
 	/* Propagate new snapshot list by IV */
-	if (rc == 0 && (opc == CONT_SNAP_CREATE || opc == CONT_SNAP_DESTROY))
-		ds_cont_update_snap_iv(svc, in->ci_uuid);
+	if (rc == 0) {
+		if (opc == CONT_SNAP_CREATE || opc == CONT_SNAP_DESTROY)
+			ds_cont_update_snap_iv(svc, in->ci_uuid);
+		else if (opc == CONT_PROP_SET)
+			ds_cont_prop_iv_update(svc, in->ci_uuid);
+	}
```
liw, line 4704: [Question] In addition to adding this new IV update call, does this patch also intend to delete any existing IV update calls? For instance, the one in set_prop?

src/container/srv_epoch.c
```diff
@@ -503,14 +503,18 @@ ds_cont_update_snap_iv(struct cont_svc *svc, uuid_t cont_uuid)
 		goto out_lock;
 	}
 
-	rc = cont_iv_snapshots_update(svc->cs_pool->sp_iv_ns, cont_uuid,
-				      snapshots, snap_count);
-	if (rc != 0)
-		D_ERROR(DF_CONT": Failed to update snapshots IV: %d\n",
-			DP_CONT(svc->cs_pool_uuid, cont_uuid), rc);
-
-	D_FREE(snapshots);
```
liw, line 512: [Question] If we don't touch the IV "state" (apologies for not knowing the correct IV terminology) before releasing the DB lock, then what will happen if an IV_FETCH request comes in after we release the DB lock but before we update the IV "state"? Will this IV_FETCH see stale IV values?

wangdi1: Yes, the IV value on other different target might be stale. But they will be synchronized eventually, if there is need to make them synchronize, there should be other mechanism. Not sure we have such use case for the moment.

IMHO, it is improper to use a simple local lock to implement global synchronization.

liw: I'm only considering the IV "state" in the local engine who is the root of the IV.

Is it possible that the RDB TX commits and a local IV fetch before the IV update doesn't see the committed changes? This depends on a couple of IV questions:

- Does the IV update guarantee that it will never yield?
- Does the IV fetch algorithm at the IV root only looks at the IV "state" or does it also checks with RDB? If it looks at the IV "state" only, then we could consider invalidate the IV locally (i.e., no RPCs involved) while holding the DB lock.

src/container/srv_container.c
wangdi1: Ah, yes, that IV update in set_prop should be removed. But this will be conflicted with PR-11467, actually that patch change "clear status" process, let me merge the patch altogether.

wangdi1:
merge this with PR-11467.

PR #11545, opened 2023-02-24
DAOS-12180 dfs: extend fs tool to support fixing entry types

mchaarawi:
In some scenarios, an entry might get corrupted (mode bits, mtime, ctime, etc.). Most of the these are correctable using regular linux commands (chmod, set mtime/ctime, etc.), but others are not, including a corrupted type bit. This PR extend the daos fs tool to add a fix-type option would allows a user to fix the type. The fix-chunk-size command was removed in favor of this new option that allows also to fix-entry and specify a chunk-size as something to fix. This option can be extended further for other entry options.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfs/dfs.c
+++ b/src/client/dfs/dfs.c
@@ -6452,7 +6452,6 @@ fetch_mark_oids(daos_handle_t coh, daos_obj_id_t oid, daos_key_desc_t *kds, char
 	d_iov_t		iov, sg_iov;
 	daos_recx_t	recx;
 	uint32_t	nr;
-	char		*ptr;
 	daos_iod_t	iod;
 	d_iov_t		marker;
 	bool		mark_data = true;
@@ -6470,7 +6469,6 @@ fetch_mark_oids(daos_handle_t coh, daos_obj_id_t oid, daos_key_desc_t *kds, char
 	sgl.sg_nr_out = 0;
 	d_iov_set(&iov, enum_buf, DFS_ITER_ENTRY_BUF);
 	sgl.sg_iovs = &iov;
-	ptr = enum_buf;
 
 	/** set sgl for fetch */
 	entry_sgl.sg_nr = 1;
@@ -6488,6 +6486,7 @@ fetch_mark_oids(daos_handle_t coh, daos_obj_id_t oid, daos_key_desc_t *kds, char
 	d_iov_set(&marker, &mark_data, sizeof(mark_data));
 	while (!daos_anchor_is_eof(&anchor)) {
 		uint32_t	i;
+		char		*ptr;
 
 		nr = DFS_ITER_NR;
 		rc = daos_obj_list_dkey(oh, DAOS_TX_NONE, &nr, kds, &sgl, &anchor, NULL);
@@ -6497,7 +6496,7 @@ fetch_mark_oids(daos_handle_t coh, daos_obj_id_t oid, daos_key_desc_t *kds, char
 		}
 
 		/** for every entry, fetch its oid and mark it in the oit table */
-		for (i = 0; i < nr; i++) {
+		for (ptr = enum_buf, i = 0; i < nr; i++) {
 			daos_obj_id_t	entry_oid;
 			daos_key_t	dkey;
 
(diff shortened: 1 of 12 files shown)


src/client/dfs/dfs.c
```diff
@@ -7200,3 +7199,81 @@ dfs_relink_root(daos_handle_t coh)
...
+	iod.iod_type	= DAOS_IOD_ARRAY;
+	iod.iod_size	= 1;
+	d_iov_set(&sg_iov, &mode, sizeof(mode_t));
+	sgl.sg_nr	= 1;
+	sgl.sg_nr_out	= 0;
+	sgl.sg_iovs	= &sg_iov;
+	rc = daos_obj_update(parent->oh, DAOS_TX_NONE, DAOS_COND_DKEY_UPDATE, &dkey, 1, &iod, &sgl,
+			     NULL);
+	if (rc) {
+		D_ERROR("Failed to update object type ("DF_RC")\n", DP_RC(rc));
```
daltonbohning, line 7271: Should this be? To match others
```suggestion
		D_ERROR("Failed to update object type "DF_RC"\n", DP_RC(rc));
```

src/tests/daos_mw_fi.c
```diff
@@ -122,11 +169,16 @@ fi_dfs(daos_handle_t poh, daos_handle_t coh, int op, const char *path, daos_prop
...
 		fname = basename(file);
 		dirp = dirname(dir);
-		printf("punching %s from %s\n", fname, dirp);
-	} else {
+		if (op == PUNCH_ENTRY)
+			printf("punching %s from %s\n", fname, dirp);
+		else
+			printf("corrupting %s in %s\n", fname, dirp);
+	} else if (op == PUNCH_OBJ) {
 		D_ASSERT(op == PUNCH_OBJ);
```
daltonbohning, line 177: Probably don't need the assert now

utils/node_local_test.py
```diff
@@ -3231,6 +3233,144 @@ def test_daos_fs_check(self):
...
+                            env=cmd_env, check=False, cwd=dfuse.dir)
+        assert rc.returncode == 0
+
+        # try to read from corrupted entries. all should fail
+        try:
+            fd = os.open(fname1, 'r')
+            os.close(fd)
+            assert False
+        except TypeError:
+            pass
```
daltonbohning, line 3296: `open` is expected to fail, right? And `close` is just in case it doesn't? You might consider this approach, which will handling closing for you. 
```suggestion
        try:
            with os.open(fname1, 'r'):
                assert False
        except TypeError:
            pass
```

utils/node_local_test.py
```diff
@@ -3231,6 +3233,144 @@ def test_daos_fs_check(self):
...
+            fd = os.open(fname3, 'r')
+            os.close(fd)
+            assert False
+        except TypeError:
+            pass
+
+        try:
+            dir_list = os.listdir(dirname2)
+            assert False
+        except OSError:
```
ashleypittman, line 3308: FYI - OSError is a generic "the kernel gave an errno" error, you may be able to trap on a subtype of this to make it more resilient.
```suggestion
        except OSError as error:
            assert error.errno == errno.???
```
Search for errno. elsewhere in this file to see what I mean.

utils/node_local_test.py
daltonbohning: If we don't need a reference to the handle, then we don't need `as fd`

mchaarawi: > What I'm wondering about here though is why TypeError? That seems like a python error and I'd expect a OSError of some kind.

im not sure.. don't have time on my side to dig into python specifics, but it does report the appropriate error from the C API side.

daltonbohning: Ah, this isn't right. Consider this:
```
>>> import os
>>> fd = open('foo', 'r')
>>> fd
<_io.TextIOWrapper name='foo' mode='r' encoding='UTF-8'>
>>> os.close(fd)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: an integer is required (got type _io.TextIOWrapper)
```
What you want is `fd.close()`, which `with open...` does for you. But if you're getting a `TypeError` is probably means the `open` is succeeding here.

daltonbohning: To use `os.close`, you would need to also use `os.open`

mchaarawi: python --

mchaarawi: but thanks @daltonbohning I think i have it right this time :-)

mchaarawi:
any other feedback @ashleypittman @daltonbohning ?

utils/node_local_test.py
```diff
@@ -3231,6 +3233,142 @@ def test_daos_fs_check(self):
...
+
+        pool = self.pool.id()
+        cont = self.container
+
+        dfuse = DFuse(self.server,
+                      self.conf,
+                      pool=self.pool.id(),
+                      container=self.container,
+                      caching=False)
+        dfuse.start(v_hint='fs_fix_test', log_mask='CRIT')
```
ashleypittman, line 3247: This is going to have the effect of disabling the log analysis leak checking which isn't desireable however the code still runs under valgrind so that should still catch any errors.

utils/node_local_test.py
```diff
@@ -1112,7 +1112,7 @@ def __str__(self):
 
         return f'DFuse instance at {self.dir} ({running})'
 
-    def start(self, v_hint=None, single_threaded=False, use_oopt=False):
+    def start(self, v_hint=None, single_threaded=False, use_oopt=False, log_mask=None):
```
ashleypittman, line 1115: If it's an object attribute you can just set it rather than pass it in here, you don't need to do both.  If you're just passing it in here then you don't need to set a value on L1099.

My preference is probably to use object attributes rather than method parameters.

ashleypittman (approved):

utils/node_local_test.py
mchaarawi: ok ill clean it up in a follow on patch that tests only NLT to avoid re-pushing this for a full run

mchaarawi: fixed here https://github.com/daos-stack/daos/pull/11602

PR #11559, opened 2023-02-27
DAOS-12709 object: Do not reset sgl size for EC data recovery

wangdi1:
During EC data degraded fetch, it will reset the sgl buffer size by replied data size from 1st degraded parity fetch, but it might be shorter than the real size, since ioc_trim_tail_holes() might trim the tail size(see akey_fetch).

So let's set the sgl size in recovery fill back process, instead of relying on the data size.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/object/cli_ec.c
+++ b/src/object/cli_ec.c
@@ -1918,10 +1918,20 @@ tgt_check:
 				tail_hole_size += recx_size;
 				continue;
 			}
+
 			D_ASSERT(data_size <= recx_size);
 			tail_hole_size += recx_size - data_size;
 			D_ASSERT(tail_hole_size <= size_in_iod);
-			dc_sgl_out_set(usgl, size_in_iod - tail_hole_size);
+
+			/**
+			 * During EC data recovery, data size might be shorter than
+			 * the real data size, because the tail degraded sgl might
+			 * be truncated by ioc_trim_tail_holes. so the sgl buf size
+			 * is set by obj_ec_recov_fill_back().
+			 */
+			if (!reasb_req->orr_recov_data ||
+			    (size_in_iod - tail_hole_size) > daos_sgl_data_len(usgl))
+				dc_sgl_out_set(usgl, size_in_iod - tail_hole_size);
 
 			return;
 		}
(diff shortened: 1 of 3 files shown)


src/object/cli_ec.c
```diff
@@ -1918,10 +1918,20 @@ obj_ec_fetch_set_sgl(struct dc_object *obj, struct obj_reasb_req *reasb_req,
...
 			D_ASSERT(tail_hole_size <= size_in_iod);
-			dc_sgl_out_set(usgl, size_in_iod - tail_hole_size);
+
+			/**
+			 * During EC data recovery, data size might be shorter than
+			 * the real data size, because the tail degraded sgl might
+			 * be truncated by ioc_trim_tail_holes. so the sgl buf size
+			 * is set by obj_ec_recov_fill_back().
+			 */
+			if (!reasb_req->orr_recov_data ||
```
liuxuezhao, line 1932: "reasb_req->orr_recov_data" looks need not to add such a new flag, can pass the "data_recov" flag in obj_ec_comp_cb() to obj_ec_fetch_set_sgl().

src/object/cli_ec.c
```diff
@@ -2610,17 +2620,35 @@ struct oes_copy_arg {
...
+		sgl->sg_iovs[idx->iov_idx - 1].iov_len =
+		     sgl->sg_iovs[idx->iov_idx - 1].iov_buf_len;
+	} else {
+		sgl->sg_iovs[idx->iov_idx].iov_len =
+			max(sgl->sg_iovs[idx->iov_idx].iov_len, idx->iov_offset);
+		D_ASSERTF(sgl->sg_iovs[idx->iov_idx].iov_len <=
+			  sgl->sg_iovs[idx->iov_idx].iov_buf_len,
+			  "iov_idx %u %p/%p offset %zd + len %zd > buf_len %zd",
+			  idx->iov_idx, sgl, sgl->sg_iovs[idx->iov_idx].iov_buf,
+			  idx->iov_offset, len, sgl->sg_iovs[idx->iov_idx].iov_buf_len);
```
liuxuezhao, line 2649: this code looks confuse that why must needed, because at the end the obj_ec_fetch_set_sgl() -> dc_sgl_out_set() will set each sg_iovs[]'s iov_len and sgl->sg_nr_out. 
do you know why need such extra code here? or any case that may miss the calling of obj_ec_fetch_set_sgl()?

not sure if can with a reproducer for the problem that may be helpful to know what's the exact issue...

src/object/obj_internal.h
```diff
@@ -189,7 +189,9 @@ struct obj_reasb_req {
 	/* the flag of IOM re-allocable (used for EC IOM merge) */
 					 orr_iom_realloc:1,
 	/* orr_fail allocated flag, recovery task's orr_fail is inherited */
-					 orr_fail_alloc:1;
+					 orr_fail_alloc:1,
+	/* The fetch data/sgl is rebuilt by EC parity rebuild */
+					 orr_recov_data:1;
```
liuxuezhao, line 194: probably better to avoid add this new flag if possible (as comment above), thanks.

src/object/cli_ec.c
wangdi1: orr_recov means the shard needs to be rebuilt from parity, Orr_recov_data means the data is actually rebuilt from the parity shard. 

And also Orr_recov is set by the recover task, but this is original task, I thought they are using different obj_auxi and rehabs_req. no?

src/object/cli_ec.c
wangdi1: So during EC fetch,  each shard fetch the data_size from the server, then we use these data_sizes from different shard to reset the input sql buffer in obj_ec_fetch_set_sgl().

But if one shard is missing, then the fetch is redirected to the parity shard, and on the parity shard, the "data size" actually depends on the partial data on the replicate space, not the real data size, i.e. the "data size" can not be trusted in obj_ec_fetch_set_sgl(). So in this case, we will set the buffer size in obj_ec_recov_data()->obj_ec_recov_fill_back()->oes_copy(),  instead of obj_ec_fetch_set_sgl().

src/object/cli_ec.c
liuxuezhao: I mean can reuse/pass in obj_ec_comp_cb()'s "bool data_recov" flag as they should be same purpose, can be refined later. will give a +1 soon and let's rethink sgl set related things later, thx.

liuxuezhao (approved):

jolivier23 (approved):

PR #11583, opened 2023-02-28
DAOS-12710 build: Update location and include paths for protoc headers.

ashleypittman:


mchaarawi:
It's a bad idea IMO to put any private header directly under src/include. it sounds the original issue is the lesser evil?

ashleypittman:
> It's a bad idea IMO to put any private header directly under src/include.

I agree, this is still WIP.  The issue there is svc.pb-c.c includes svc.pb-c.h without any path prefix so I can't put it under src/include/daos without then adding that to -I as well.

PR #11590, opened 2023-03-01
DAOS-12745 bio: Enable v2.0 daos_nvme.conf with v2.2 install

tanabarr:
Enable a DAOS 2.0 daos_nvme.conf to be used in a 2.2 installation.
Allow a missing daos_data key in the JSON config to avoid config
parse errors on daos_engine start-up.

### Gatekeeper:


wangshilong (approved):
Thanks for fixing this up Tom!

NiuYawei (approved):

tanabarr:
CI failures are expected DAOS-12631 & DAOS-12722 (Rebuild EC)

mjmac (commented):
The code changes look fine to me. How has this been tested? What happens when a 2.2 server loads an older configuration? Is the config auto-updated with the new format, or is the admin expected to do something to force the migration?

If the config is auto-updated to the 2.2 format, what happens if the admin then rolls back to 2.0?

tanabarr:
> The code changes look fine to me. How has this been tested? What happens when a 2.2 server loads an older configuration? Is the config auto-updated with the new format, or is the admin expected to do something to force the migration?

I tested by removing the daos_data top-level key from the JSON nvme config (which is what is produced by v2.0 server on format). With unchanged v2.2, engine fails to start, with this patch applied, engine starts.
The PR relaxes engine validation of daos_nvme.conf so that if the daos_data key is missing a warning is printed but the engine still loads. v2.2 only reads the hotplug bus-ID range from daos_data currently and this is not a much used feature. Migration to the new format from a v2.0 generated config file will only happen on reformat.

> If the config is auto-updated to the 2.2 format, what happens if the admin then rolls back to 2.0?

No modification of the config file is performed so the old format config file will work when rollback to v2.0 happens. If the config file is regenerated during a v2.2 reformat then actually it should work fine with v2.0 because the extra top-level daos_data key will just be ignored by the v2.0 engine. Whilst I'm confident in this assertion I haven't actually verified it in practice.

mjmac:
> The PR relaxes engine validation of daos_nvme.conf so that if the daos_data key is missing a warning is printed but the engine still loads. v2.2 only reads the hotplug bus-ID range from daos_data currently and this is not a much used feature. Migration to the new format from a v2.0 generated config file will only happen on reformat.

Hmm. OK. I'm not overly worried about this specific PR because 2.2 is on its way out of support. For future situations like this, though, it doesn't seem ideal that the admin would get a warning every time they (re-)start their servers, and we can't ask them to reformat their storage after an upgrade in order to get rid of the warning.

We don't need to do it for this situation specifically, but is there a technical reason that we couldn't just upgrade the config file in-place when we detect that it's an older format?

tanabarr:
> > The PR relaxes engine validation of daos_nvme.conf so that if the daos_data key is missing a warning is printed but the engine still loads. v2.2 only reads the hotplug bus-ID range from daos_data currently and this is not a much used feature. Migration to the new format from a v2.0 generated config file will only happen on reformat.
> 
> Hmm. OK. I'm not overly worried about this specific PR because 2.2 is on its way out of support. For future situations like this, though, it doesn't seem ideal that the admin would get a warning every time they (re-)start their servers, and we can't ask them to reformat their storage after an upgrade in order to get rid of the warning.
>
> We don't need to do it for this situation specifically, but is there a technical reason that we couldn't just upgrade the config file in-place when we detect that it's an older format?

There is no technical reason for not updating the config, I didn't think it was necessary in this case. We can definitely do that for any future cases. The top-level daos_data key should now be present for all v2.2+ versions.

mjmac (approved):

sylviachanoiyee:
Liang (in gatekeeping channel) thinks the rebuild ec test failure is unrelated.

PR #11593, opened 2023-03-01
DAOS-12734 common: Fix for failure/crash on container destroy

sherintg:
This fix addresses a bug within the bmem allocator, where the initial capacity for redo ops ulog was set incorrectly. Without this fix, if the number of redo log entries within a tx is between 40 and 64, then only the first 40 entries are getting committed.

### Gatekeeper:

### DIFF
--- a/src/common/dav/memops.c
+++ b/src/common/dav/memops.c
@@ -66,7 +66,7 @@ struct operation_context {
 	struct ulog *ulog_curr; /* current persistent log */
 	size_t total_logged; /* total amount of buffer stores in the logs */
 
-	struct ulog *ulog; /* pointer to the persistent ulog log */
+	struct ulog *ulog; /* pointer to the ulog used by context for undo ops */
 	size_t ulog_base_nbytes; /* available bytes in initial ulog log */
 	size_t ulog_capacity; /* sum of capacity, incl all next ulog logs */
 	int ulog_auto_reserve; /* allow or do not to auto ulog reservation */
@@ -75,7 +75,7 @@ struct operation_context {
 
 	enum operation_state state; /* operation sanity check */
 
-	struct operation_log pshadow_ops; /* shadow copy of persistent ulog */
+	struct operation_log pshadow_ops; /* used by context for redo ops */
 	struct operation_log transient_ops; /* log of transient changes */
 
 	/* collection used to look for potential merge candidates */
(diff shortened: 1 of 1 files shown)


src/common/dav/memops.c
```diff
@@ -128,7 +128,7 @@ operation_log_persistent_init(struct operation_log *log,
 	}
 
 	/* initialize underlying redo log structure */
-	src->capacity = ulog_base_nbytes;
+	src->capacity = ULOG_BASE_SIZE;
```
NiuYawei, line 131: This looks really confusing to me. In dav_create_clogs(), we call ulog_construct_new() with LANE_REDO_EXTERNL_SIZE, but here we specify it as ULOG_BASE_SIZE?

It looks to me all the 'lane' stuff is unused? should we remove these LANE_REDO/UNDO size macros and always use ULOG_BASE_SIZE for creating & extending?

src/common/dav/memops.c
```diff
@@ -531,7 +531,7 @@ operation_process_persistent_redo(struct operation_context *ctx)
 int
 operation_reserve(struct operation_context *ctx, size_t new_capacity)
 {
-	if (new_capacity > ctx->ulog_capacity) {
+	if ((ctx->type == LOG_TYPE_UNDO) && (new_capacity > ctx->ulog_capacity)) {
```
NiuYawei, line 534: I don't see why redo log doesn't need extending, could you explain it a bit?

src/common/dav/memops.c
sherintg: The lane stuff in pmdk is to persist the redo and undo logs. The redo ops are initially logged into a transient memory logs (pshadow_ops and transient_ops) and later persisted to lane pointed by ctx->ulog before committing the changes. The undo logs are written directly to ulog as and when it happens.

With BMEM, there is no need to persist these logs and redo ops in pshadow_ops are converted to wal redo ops and not written to ctx->ulog. Hence the ulog_construct_new() with LANE_REDO_EXTERNL_SIZE can be cleaned up. The cleanup requires some more changes as certain functions expects ctx->ulog to be not NULL. @ivadmitry is looking at doing the cleanup with changes to the functions that operates on ctx->ulog.

src/common/dav/memops.c
sherintg: As mentioned in the previous comment, the redo ops do not update ctx->ulog. Instead they get appended to the log within ctx->pshadow_ops.  The log in pshadow_ops is extended using realloc within the function operation_add_typed_entry().

NiuYawei (approved):
Good catch! and thanks for the explanation.

gnailzenh (approved):

PR #11596, opened 2023-03-02
DAOS-12689 control: Don't return an error if dlsym succeeded

kjacque:
dlerror() may not be thread-safe, depending on the implementation. It is possible that after clearing dlerror(), we could encounter a race with another thread.

In the DAOS control plane context, we always expect dlsym to return a non-NULL value if it succeeded. This avoids detecting an error set by another thread, when the current thread's dlsym() succeeded.

Features: control

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/lib/dlopen/dlopen.go
+++ b/src/control/lib/dlopen/dlopen.go
@@ -54,7 +54,7 @@ func (l *LibHandle) GetSymbolPointer(symbol string) (unsafe.Pointer, error) {
 	C.dlerror()
 	p := C.dlsym(l.Handle, sym)
 	e := C.dlerror()
-	if e != nil {
+	if p == nil && e != nil {
 		return nil, fmt.Errorf("error resolving symbol %q: %v", symbol, errors.New(C.GoString(e)))
 	}
 


src/control/lib/dlopen/dlopen.go
```diff
@@ -54,7 +54,7 @@ func (l *LibHandle) GetSymbolPointer(symbol string) (unsafe.Pointer, error) {
 	C.dlerror()
 	p := C.dlsym(l.Handle, sym)
 	e := C.dlerror()
-	if e != nil {
+	if p == nil && e != nil {
```
knard-intel, line 57: It seems more safe for me to directly test if p is nil and then return the error message according to the output of the C.dlerror() function call.

tanabarr: I think the intention is to avoid situations where there is a false error reported due to a thread race, therefore the first check for a valid symbol should be made before checking the error to reduce the chances of false positives in the case of dlerror.

tanabarr (approved):

mjmac (approved):

src/control/lib/dlopen/dlopen.go
knard-intel: I was thinking that we could eventually miss errors as the code is currently organized.
For example, do you think that it could be possible C.dlsym() return NULL and a thread race makes C.dlerror() return NULL ?  From side, I am not able to say if such scenario is possible or not.
If not, then it is OK for me as the only issue that I see could be to eventually have an error message not relevant.  From my side, I do not see how this last issue could be avoided without lock.

tanabarr: I imagine there is a remote possibility that this could occur, it sounds like we cannot easily guard against all remote possibilities and the intent of this PR seems to be to reduce the potential for incorrect operation. I will let @kjacque comment on whether a Lock solution is possible in order to rule out the issue altogether.

kjacque: In theory another thread could clear the error before we got a chance to see it. I wanted to avoid locking if possible. One thing that is worth noting, we do check if the value of dlsym is nil in the callers, as nil is a valid return value for dlsym (just not for our particular use cases).

PR #11599, opened 2023-03-02
DAOS-12577 dfuse: Update NLT to test cache on lookup.

ashleypittman:
Add a reproducer for DAOS-12577

Required-githooks: true


utils/node_local_test.py
```diff
@@ -2735,18 +2763,18 @@ def test_uns_basic(self):
         dfuse.start('uns-1')
 
         # List the root container.
-        print(os.listdir(join(dfuse.dir, pool, container)))
+        print(os.listdir(join(dfuse.dir, pool, container.uuid)))
```
daltonbohning, line 2766: `uuid` or `id()`?

utils/node_local_test.py
```diff
@@ -2991,7 +3014,7 @@ def test_cont_clone(self):
                '--src',
                src_dir.name,
                '--dst',
-               f'daos://{self.pool.uuid}/{self.container}']
+               f'daos://{self.pool.uuid}/{self.container.id()}']
```
daltonbohning, line 3017:
```suggestion
               f'daos://{self.pool.id()}/{self.container.id()}']
```

utils/node_local_test.py
```diff
@@ -3005,7 +3028,7 @@ def test_cont_clone(self):
         cmd = ['container',
                'clone',
                '--src',
-               f'daos://{self.pool.uuid}/{self.container}',
+               f'daos://{self.pool.uuid}/{self.container.id()}',
```
daltonbohning, line 3031:
```suggestion
               f'daos://{self.pool.id()}/{self.container.id()}',
```

utils/node_local_test.py
```diff
@@ -3556,11 +3652,11 @@ def run_dfuse(server, conf):
     stat_and_check(dfuse, pre_stat)
     check_no_file(dfuse)
 
-    pool_stat = os.stat(join(dfuse.dir, pool))
+    pool_stat = os.stat(join(dfuse.dir, pool.uuid))
```
daltonbohning, line 3655: `uuid` or `id()`?

utils/node_local_test.py
```diff
@@ -470,11 +470,47 @@ def fetch_containers(self):
...
+
+        Args:
+            valgrind (bool, optional): Run the command under valgrind. Defaults to True.
+            log_check (bool, optional): Run log analysis. Defaults to True.
+
+        Raises:
+            NLTestFail: If Pool was not provided when object created.
+        """
+        if not isinstance(self.pool, DaosPool):
+            raise NLTestFail('No pool passed when creating container')
```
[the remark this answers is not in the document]
ashleypittman: This is intentional.  If a DaosPool object is passed when DaosCont is created then the DaosCont object can save extra data in the object which means destroy can work.  If a string is passed in for pool then it can't and this method will not work.  Future work might be to only accept a DaosPool when creating which would simplify this.

utils/node_local_test.py
ashleypittman: uuid.  This is used for dfuse path lookup which is uuid based.

utils/node_local_test.py
```diff
@@ -470,11 +470,47 @@ def fetch_containers(self):
...
+
+        Args:
+            attrs (dict): Dictionary of attributes to set.
+        """
+        if not isinstance(self.pool, DaosPool):
+            raise NLTestFail('No pool passed when creating container')
+        for key, value in attrs.items():
+            rc = run_daos_cmd(self.pool.conf, ['container', 'set-attr', self.pool.id(), self.uuid,
+                                               f'{key}:{value}'],
+                              show_stdout=True)
```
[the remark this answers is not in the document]
ashleypittman: Are you sure about that, it doesn't appear to.

utils/node_local_test.py
ashleypittman: As above this needs to be uuid.

utils/node_local_test.py
ashleypittman: No.  We've discussed it but probably not going to add it either as there are very few use-cases where dfuse isn't passed a container on the command line.

utils/node_local_test.py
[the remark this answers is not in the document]
ashleypittman: Ok, I did this.  There are a couple of cases where DaosCont is created through a path so picks up the pool from there and therefore there is no DaosPool object to provide, but those tests don't actually destroy the container afterwards so they work just fine without this check.

daltonbohning (approved):

kprantis:
This is a 2.4 blocker, if reviews could please be prioritized.

ashleypittman:
> I wonder if all these changes were really required to implement what was necessary to address DAOS-12577 issue ?

No, this started off being a attempt at a reproducer and ended up as a code cleanup and reproducer in one, once we had a reproducer it was easy to fix but given the size of this PR and the severity of the issue we landed the fix in https://github.com/daos-stack/daos/pull/11741. I'd be OK to file a new ticket for the cleanups in this PR which need not be a blocker but I do think we should land this anyway so the distinction is academic in that case.

I should have managed this ticket better in Jira and will update it now.

@daltonbohning the reproducer in this PR identified the issue but it's actually a complex issue that affected lots of use-cases, many of them much simpler than the process via which we discovered it, can we have a discussion about easier ways to test core functionality.

bfaccini (approved):

PR #11601, opened 2023-03-02
DAOS-11203 dtx: reduce DTX refresh RPC timeout

Nasf-Fan:
If related DTX is committable, then it will be committed
within DTX_COMMIT_THRESHOLD_AGE seconds. So if need the DTX
refresh to wait for longer, then just let related client
to retry via a shorter timeout.

More debug messge for IO request timeout.

### Gatekeeper:

### DIFF
--- a/src/dtx/dtx_rpc.c
+++ b/src/dtx/dtx_rpc.c
@@ -234,8 +234,16 @@ dtx_req_send(struct dtx_req_rec *drr, daos_epoch_t epoch)
 			din->di_flags.ca_arrays = NULL;
 		}
 
-		if (dra->dra_opc == DTX_REFRESH && DAOS_FAIL_CHECK(DAOS_DTX_RESYNC_DELAY)) {
-			rc = crt_req_set_timeout(req, 3);
+		if (dra->dra_opc == DTX_REFRESH) {
+			if (DAOS_FAIL_CHECK(DAOS_DTX_RESYNC_DELAY))
+				rc = crt_req_set_timeout(req, 3);
+			else
+				/*
+				 * If related DTX is committable, then it will be committed
+				 * within DTX_COMMIT_THRESHOLD_AGE time. So if need to wait
+				 * for longer, then just let related client to retry.
+				 */
+				rc = crt_req_set_timeout(req, DTX_COMMIT_THRESHOLD_AGE);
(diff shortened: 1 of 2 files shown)


src/dtx/dtx_rpc.c
```diff
@@ -234,8 +234,16 @@ dtx_req_send(struct dtx_req_rec *drr, daos_epoch_t epoch)
...
+		if (dra->dra_opc == DTX_REFRESH) {
+			if (DAOS_FAIL_CHECK(DAOS_DTX_RESYNC_DELAY))
+				rc = crt_req_set_timeout(req, 3);
+			else
+				/*
+				 * If related DTX is committable, then it will be committed
+				 * within DTX_COMMIT_THRESHOLD_AGE time. So if need to wait
+				 * for longer, then just let related client to retry.
+				 */
+				rc = crt_req_set_timeout(req, DTX_COMMIT_THRESHOLD_AGE);
```
wangdi1, line 246: min (rpc_timeout, DTX_COMMIT_THRESHOLD_AGE) ?

src/object/cli_shard.c
```diff
@@ -896,6 +896,9 @@ dc_rw_cb(tse_task_t *task, void *arg)
 		D_GOTO(out, rc = -DER_HG);
 	}
 
+	reasb_req = rw_args->shard_args->reasb_req;
+	is_ec_obj = reasb_req != NULL && daos_oclass_is_ec(reasb_req->orr_oca);
```
wangdi1, line 900: probably not this patch issue, but looks like reasb_req != NULL is enough here?

wangdi1 (approved):
I am a bit concern about re-set the timeout, which is bigger than the system rpc timeout.

src/dtx/dtx_rpc.c
Nasf-Fan: In theory, you are right. I will consider to change it in subsequent patch.

src/object/cli_shard.c
Nasf-Fan: Just move the logic from other place, but it seems yes.

liuxuezhao: "reasb_req != NULL" looks enough, and probably better to use obj_is_ec(obj) and assert(reasb_req != NULL) for ec obj. the old original code probably just not sure if need the reasb_req for replica at that time.

PR #11627, opened 2023-03-07
DAOS-12824 dtx: reindex committed DTX entries

Nasf-Fan:
Change the process for committed DTX entries reindex as self-traced by the vos container inside VOS. Then even if related reindex ULT is stopped and restarted again and again (for container open/close, DTX resync/check), the committed DTX entries reindex process will always resume from former paused point instead of from scratch.

There is race between DTX reindex ULT start and stop. If someone is trying to stop the reindex process, then next starting reindex need to wait instead of return directly.

### Gatekeeper:

### DIFF
--- a/src/dtx/dtx_common.c
+++ b/src/dtx/dtx_common.c
@@ -1589,14 +1589,13 @@ dtx_reindex_ult(void *arg)
 {
 	struct ds_cont_child		*cont	= arg;
 	struct dss_module_info		*dmi	= dss_get_module_info();
-	uint64_t			 hint	= 0;
 	int				 rc	= 0;
 
 	D_INFO(DF_CONT": starting DTX reindex ULT on xstream %d, ver %u\n",
 	       DP_CONT(NULL, cont->sc_uuid), dmi->dmi_tgt_id, dtx_cont2ver(cont));
 
 	while (!cont->sc_dtx_reindex_abort && !dss_xstream_exiting(dmi->dmi_xstream)) {
-		rc = vos_dtx_cmt_reindex(cont->sc_hdl, &hint);
+		rc = vos_dtx_cmt_reindex(cont->sc_hdl);
 		if (rc != 0)
 			break;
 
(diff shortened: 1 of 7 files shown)


Nasf-Fan:
Hit DAOS-10377, to be retested.

kccain:
container/srv_target.c debug log change looks fine. Not familiar enough with the other parts of the code to provide a good review.

src/dtx/dtx_common.c
```diff
@@ -1618,7 +1617,11 @@ start_dtx_reindex_ult(struct ds_cont_child *cont)
 
 	D_ASSERT(cont != NULL);
 
-	if (cont->sc_dtx_reindex || cont->sc_dtx_reindex_abort)
+	/* Someone is trying to stop former DTX reindex ULT, wait until its done. */
+	while (cont->sc_dtx_reindex_abort)
+		ABT_thread_yield();
```
wangdi1, line 1622: please use dss_sleep to help the scheduler, instead of using yield directly.

src/dtx/dtx_common.c
```diff
@@ -1637,7 +1640,12 @@ start_dtx_reindex_ult(struct ds_cont_child *cont)
 void
 stop_dtx_reindex_ult(struct ds_cont_child *cont)
 {
-	if (!cont->sc_dtx_reindex || dtx_cont_opened(cont))
+	/* DTX reindex has been done or not has not been started. */
+	if (!cont->sc_dtx_reindex)
+		return;
+
+	/* Do not stop DTX reindex if the container is still opened. */
+	if (dtx_cont_opened(cont))
```
wangdi1, line 1648: I think these worth a D_INFO to help us understand some scenarios.

src/dtx/dtx_common.c
wangdi1: probably use cond signal here?

wangdi1 (approved):

src/dtx/dtx_common.c
Nasf-Fan: This is relative rare race and the race window is very small, it is unnecessary to add new condition member for that.

Nasf-Fan: > please use dss_sleep to help the scheduler, instead of using yield directly.

Hm, the race window is very small, on the other hand, related logic maybe triggered when open container, I do not want to block it for too long time. So I do not know how long sleep is suitable, maybe just use "zero" as the input parameter, that is equal to ABT_thread_yield() but without more unnecessary CPU cycles?

Nasf-Fan:
@liuxuezhao @jolivier23, please help to review the patch, thanks!

PR #11628, opened 2023-03-07
Revert "DAOS-12833 build: Avoid updating to SCons 4.5"

ashleypittman:
Issue is now fixed upstream.

Reverts daos-stack/daos#11620

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,6 @@ ninja
 pyelftools
 pyxattr
 pyyaml
-scons<4.5        # 4.5 has some serious issues
+scons        # Works around a bug in scons on EL 8.
 tabulate
 wheel


johannlombardi:
Build failure on ARM is unrelated?

brianjmurrell (approved):

ashleypittman:
> Build failure on ARM is unrelated?

I was wondering about this, I don't think we use a cachebust on arm so we're probably picking up a old container image that has the incorrect scons version.  Scons wasn't downloaded from the build logs so we can't tell the version however we can also be sure it's not 4.5.1.

Do you know how/when the images there are ever pruned or do we need to setup a cache-bust on self-hosted runners?  I was considering filing a ticket.

PR #11631, opened 2023-03-07
DAOS-12311 control: Remove tmp file during ipmctl show region exec

tanabarr:
When retrieving ipmctl region details in XML format, a zero-length
file called output.tmp is created in the current directory. Issue
seems to be specific to shell command execution in daos_server_helper
binary after the request has been forwarded from the scm provider
running in the server process.

Remove the file if it has been created during command execution.

Features: control
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/server/storage/scm/ipmctl.go
+++ b/src/control/server/storage/scm/ipmctl.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2019-2022 Intel Corporation.
+// (C) Copyright 2019-2023 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
--- a/src/control/server/storage/scm/ipmctl_region.go
+++ b/src/control/server/storage/scm/ipmctl_region.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2022 Intel Corporation.
+// (C) Copyright 2022-2023 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -9,6 +9,7 @@ package scm
 import (
 	"encoding/xml"
 	"fmt"
+	"os"
 	"regexp"
 	"sort"
 	"strconv"
@@ -20,6 +21,13 @@ import (
 	"github.com/daos-stack/daos/src/control/lib/ipmctl"
 	"github.com/daos-stack/daos/src/control/logging"
 	"github.com/daos-stack/daos/src/control/server/storage"
+)
+
+const tmpOutFileName = "output.tmp"
+
+var (
+	osStat   = os.Stat
+	osRemove = os.Remove
 )
 
 type (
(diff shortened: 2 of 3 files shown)


kjacque (approved):

tanabarr:
CI failure in functional VM EL8 is DAOS-12825

mjmac:
The solution seems fine... It's unfortunate, though, that we don't understand the source of these temp files. Is this a recent problem, or has it always been there and we've just never noticed? If it turns out that it's just a transient issue caused by a specific release of ipmctl, I think we should really think about whether or not we want to commit a long-term fix for it.

tanabarr:
> The solution seems fine... It's unfortunate, though, that we don't understand the source of these temp files. Is this a recent problem, or has it always been there and we've just never noticed? If it turns out that it's just a transient issue caused by a specific release of ipmctl, I think we should really think about whether or not we want to commit a long-term fix for it.

I agree that it would be good to understand why it's appearing, I've performed a significant amount of investigation to try to isolate and identify the cause but it seems that the effort/time spent is out of proportion with the impact of the problem and therefore I opted to add the mitigation code to identify and remove the temporary file. Here are some summary results of the investigation:
- The issue is consistent and has probably existed for a long time, there is no indication that it's caused by recent version of ipmctl. The issue was reported by customer.
- The file is only created by the command `ipmctl show -region -o nvmxml`, other commands don't exhibit the issue.
- The file is only created if the xml output specifier is used.
- The file is not created if the command is issued by a single go executable, only when forwarding is used between daos_server and daos_server_helper.
- Handling stdout and sterr explicitly in go code doesn't seem to affect the issue (attempting to rule out a file handler being used by default if stderr handler doesn't exist).
- When running a single go executable (main program that just executes the command to attempt to reproduce the issue), an error is written to temporary file in the case that permissions do not exist to run the program (e.g. "ipmctl requires root permissions" message gets written to `output.tmp`).

I can continue investigation if you feel it warrants further effort.

mjmac:
> * The file is only created by the command `bash -c "ipmctl show -region -o nvmxml"`, other ipmctl commands don't seem to exhibit the issue, I haven't managed to try without the `bash -c` yet during forwarding but that could be another path to investigate.

Why are we using `bash -c` with `cmd.Exec()` instead of invoking ipmctl directly?

> I can continue investigation if you feel it warrants further effort.

You're the engineer working on the problem. I'll leave it up to you to decide whether or not it's worth further investigation in order to understand why the code isn't behaving as expected.

tanabarr:
> > * The file is only created by the command `bash -c "ipmctl show -region -o nvmxml"`, other ipmctl commands don't seem to exhibit the issue, I haven't managed to try without the `bash -c` yet during forwarding but that could be another path to investigate.
> 
> Why are we using `bash -c` with `cmd.Exec()` instead of invoking ipmctl directly?

I can't remember the reason off the top of my head I will have to investigate.
 
> > I can continue investigation if you feel it warrants further effort.
> 
> You're the engineer working on the problem. I'll leave it up to you to decide whether or not it's worth further investigation in order to understand why the code isn't behaving as expected.

If I am understanding your intent behind this comment correctly then it sounds like you want me to continue on this, so I will.

tanabarr:
When I run without "bash -c" I get: "ERROR: /usr/bin/daos_server_helper ERROR: pbin: code = 5 description = "cmd \"ndctl list -N -v\": exec: \"ndctl list -N -v\": executable file not found in $PATH: stdout: ".

@mjmac would you be okay if I landed this PR as is? tbh I think we have bigger fish to fry.

tanabarr:
https://daosio.atlassian.net/browse/DAOS-12311 fix resolves the creation of temporary file making this PR obsolete.

PR #11647, opened 2023-03-08
DAOS-12826 control: Handle empty storage tier in config unmarshal

tanabarr:
Avoid segfault when control-plane parses a server config file with an
empty storage tier yaml list entry by ignoring nil when iterating
tiers during custom unmarshal.

### Gatekeeper:

### DIFF
--- a/src/control/server/storage/config.go
+++ b/src/control/server/storage/config.go
@@ -285,12 +285,19 @@ func (tcs *TierConfigs) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 
+	tid := 0
+	tmp2 := make([]*TierConfig, 0, len(tmp))
 	for i := range tmp {
-		if tmp[i].Tier == 0 {
-			tmp[i].Tier = i
+		if tmp[i] == nil {
+			continue
 		}
+		if tmp[i].Tier == 0 {
+			tmp[i].Tier = tid
+		}
+		tmp2 = append(tmp2, tmp[i])
+		tid++
 	}
-	*tcs = tmp
+	*tcs = tmp2
 
 	return nil
 }
--- a/src/control/server/storage/config.go
+++ b/src/control/server/storage/config.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2019-2022 Intel Corporation.
+// (C) Copyright 2019-2023 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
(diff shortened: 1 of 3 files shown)


src/control/server/storage/config.go
```diff
@@ -285,12 +285,19 @@ func (tcs *TierConfigs) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		}
 		if tmp[i].Tier == 0 {
-			tmp[i].Tier = i
+			tmp[i].Tier = tid
 		}
+		tmp2 = append(tmp2, tmp[i])
+		tid++
 	}
-	*tcs = tmp
+	*tcs = tmp2
```
mjmac, line 300: I think it would be a lot clearer to just delete the nil tier entries, e.g.:
```go
	tid := 0
	for i, tier := range tmp {
		if tier == nil {
			tmp = append(tmp[:i], tmp[i+1:]...)
			continue
		}
		tier.Tier = tid
		tid++
	}

	*tcs = tmp
	return nil
```

Also, I think it would be worthwhile to add unit tests specifically for this method to `config_test.go`, and in particular there should be cases for unmarshaling with nil tiers in the first and second position.

tanabarr: done

kjacque (dismissed):

src/control/server/storage/config.go
knard-intel: From what I tested there is an issue with the code proposed by @mjmac .
For example if we remove two elements and one of them is the last, then we will have an out of range error.
`package main

import "fmt"

func main() {
	foo := []string{"Hello", "stupid", "World"}
	for index, bar := range foo {
		if bar == "stupid" || bar == "World" {
			foo = append(foo[:index], foo[index+1:]...)
			continue
		}

	}
	fmt.Printf("%q", foo)
}`

tanabarr: Good catch @knard-intel . I have added a test case for this which does indeed fail with the proposed code and have reverted to the original version which doesn't. Thanks!

knard-intel (approved):
LGTM

src/control/server/storage/config.go
mjmac: Hmm. Oops! Good catch, not sure what I was thinking... Glad that we've got more unit test coverage here now.

mjmac (approved):

kjacque (approved):

tanabarr:
CI failure unrelated and a known issue: https://daosio.atlassian.net/browse/DAOS-12857

PR #11653, opened 2023-03-09
DAOS-12224 control: With md_on_ssd, remount tmpfs on engine restart

sherintg:
If md_on_ssd is enabled and tmpfs is not mounted upon engine restart then, daos_server automatically mounts tmpfs and populates it with the pool layout as stored in the bdev meta storage backend.

### Gatekeeper:

### DIFF
--- a/src/bio/smd/smd_pool.c
+++ b/src/bio/smd/smd_pool.c
@@ -303,6 +303,29 @@ smd_rdb_get_blob(uuid_t pool_id, uint32_t tgt_id, enum smd_dev_type st, uint64_t
 	return pool_get_blob(pool_id, tgt_id, TABLE_RDBS[st], blob_id);
 }
 
+int
+smd_rdb_get_blob_sz(uuid_t pool_id, uint64_t *blob_sz)
+{
+	struct smd_pool	pool;
+	struct d_uuid	id;
+	int		rc = 0;
+
+	uuid_copy(id.uuid, pool_id);
+
+	smd_db_lock();
+	rc = smd_db_fetch(TABLE_RDBS[SMD_DEV_TYPE_META], &id, sizeof(id), &pool, sizeof(pool));
+	if (rc) {
+		D_CDEBUG(rc != -DER_NONEXIST, DLOG_ERR, DB_MGMT,
+			 "Fetch pool "DF_UUID" failed. "DF_RC"\n",
+			 DP_UUID(&id.uuid), DP_RC(rc));
+		goto out;
+	}
+	*blob_sz = pool.sp_blob_sz;
+out:
+	smd_db_unlock();
+	return rc;
+}
+
 static int
 smd_pool_list_cb(struct sys_db *db, char *table, d_iov_t *key, void *args)
 {
--- a/src/bio/smd/smd_pool.c
+++ b/src/bio/smd/smd_pool.c
@@ -29,7 +29,8 @@ char *TABLE_RDBS[SMD_DEV_TYPE_MAX] = {
 
 struct smd_pool {
 	uint64_t	sp_blob_sz;
-	uint32_t	sp_tgt_cnt;
+	uint16_t	sp_flags;
+	uint16_t	sp_tgt_cnt;
 	uint32_t	sp_tgts[SMD_MAX_TGT_CNT];
 	uint64_t	sp_blobs[SMD_MAX_TGT_CNT];
 };
@@ -86,12 +87,17 @@ pool_add_tgt(uuid_t pool_id, uint32_t tgt_id, uint64_t blob_id, char *table_name
 		pool.sp_tgts[pool.sp_tgt_cnt] = tgt_id;
 		pool.sp_blobs[pool.sp_tgt_cnt] = blob_id;
 		pool.sp_tgt_cnt += 1;
+		if (!strcmp(table_name, TABLE_POOLS[SMD_DEV_TYPE_META]))
+			pool.sp_flags |= SMD_POOL_IN_CREATION;
 
 	} else if (rc == -DER_NONEXIST) {
 		pool.sp_tgts[0]	 = tgt_id;
 		pool.sp_blobs[0] = blob_id;
 		pool.sp_tgt_cnt	 = 1;
 		pool.sp_blob_sz  = blob_sz;
+		pool.sp_flags = 0;
+		if (!strcmp(table_name, TABLE_POOLS[SMD_DEV_TYPE_META]))
+			pool.sp_flags |= SMD_POOL_IN_CREATION;
 
 	} else {
 		D_ERROR("Fetch pool "DF_UUID" failed. "DF_RC"\n",
@@ -213,6 +219,7 @@ smd_pool_alloc_info(struct d_uuid *id, struct smd_pool *pools)
 		}
 		info->spi_blob_sz[st] = pools[st].sp_blob_sz;
 		info->spi_tgt_cnt[st] = pools[st].sp_tgt_cnt;
+		info->spi_flags[st] = pools[st].sp_flags;
 
 		for (i = 0; i < pools[st].sp_tgt_cnt; i++) {
 			info->spi_tgts[st][i] = pools[st].sp_tgts[i];
@@ -429,5 +436,37 @@ smd_pool_replace_blobs_locked(struct smd_pool_info *info, int tgt_cnt,
 			DP_UUID(&id.uuid), DP_RC(rc));
 		return rc;
 	}
+	return rc;
+}
+
+int
+smd_pool_mark_ready(uuid_t pool_id)
+{
+	struct smd_pool	pool;
+	struct d_uuid	id;
+	int		rc;
+
+	uuid_copy(id.uuid, pool_id);
+
+	smd_db_lock();
+	rc = smd_db_fetch(TABLE_POOLS[SMD_DEV_TYPE_META], &id, sizeof(id), &pool, sizeof(pool));
+	if (rc == -DER_NONEXIST) {
+		D_GOTO(out, rc = 0);
+	} else if (rc) {
+		D_ERROR("Failed to fetch smd entry of the meta blob for "DF_UUID". "DF_RC"\n",
+			DP_UUID(&id.uuid), DP_RC(rc));
+		goto out;
+	}
+
+	pool.sp_flags &= ~SMD_POOL_IN_CREATION;
+
+	rc = smd_db_upsert(TABLE_POOLS[SMD_DEV_TYPE_META], &id, sizeof(id), &pool, sizeof(pool));
+	if (rc) {
+		D_ERROR("Failed to make pool "DF_UUID" as ready in smd. "DF_RC"\n",
+			DP_UUID(&id.uuid), DP_RC(rc));
+		goto out;
+	}
+out:
+	smd_db_unlock();
 	return rc;
 }
--- a/src/bio/smd/smd_pool.c
+++ b/src/bio/smd/smd_pool.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2018-2022 Intel Corporation.
+ * (C) Copyright 2018-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -9,19 +9,19 @@
 #include <daos/dtx.h>
 #include "smd_internal.h"
 
-char *TABLE_TGTS[SMD_DEV_TYPE_MAX] = {
+char TABLE_TGTS[SMD_DEV_TYPE_MAX][SMD_DEV_NAME_MAX] = {
 	"target",	/* compatible with old version */
 	"meta_target",
 	"wal_target",
 };
 
-char *TABLE_POOLS[SMD_DEV_TYPE_MAX] = {
+char TABLE_POOLS[SMD_DEV_TYPE_MAX][SMD_DEV_NAME_MAX] = {
 	"pool",
 	"meta_pool",
 	"wal_pool",
 };
 
-char *TABLE_RDBS[SMD_DEV_TYPE_MAX] = {
+char TABLE_RDBS[SMD_DEV_TYPE_MAX][SMD_DEV_NAME_MAX] = {
 	"rdb_data",
 	"rdb_meta",
 	"rdb_wal",
@@ -87,7 +87,7 @@ pool_add_tgt(uuid_t pool_id, uint32_t tgt_id, uint64_t blob_id, char *table_name
 		pool.sp_tgts[pool.sp_tgt_cnt] = tgt_id;
 		pool.sp_blobs[pool.sp_tgt_cnt] = blob_id;
 		pool.sp_tgt_cnt += 1;
-		if (!strcmp(table_name, TABLE_POOLS[SMD_DEV_TYPE_META]))
+		if (!strncmp(table_name, TABLE_POOLS[SMD_DEV_TYPE_META], SMD_DEV_NAME_MAX))
 			pool.sp_flags |= SMD_POOL_IN_CREATION;
 
 	} else if (rc == -DER_NONEXIST) {
@@ -96,7 +96,7 @@ pool_add_tgt(uuid_t pool_id, uint32_t tgt_id, uint64_t blob_id, char *table_name
 		pool.sp_tgt_cnt	 = 1;
 		pool.sp_blob_sz  = blob_sz;
 		pool.sp_flags = 0;
-		if (!strcmp(table_name, TABLE_POOLS[SMD_DEV_TYPE_META]))
+		if (!strncmp(table_name, TABLE_POOLS[SMD_DEV_TYPE_META], SMD_DEV_NAME_MAX))
 			pool.sp_flags |= SMD_POOL_IN_CREATION;
 
 	} else {
--- a/src/control/server/instance_storage.go
+++ b/src/control/server/instance_storage.go
@@ -99,7 +99,7 @@ func (ei *EngineInstance) awaitStorageReady(ctx context.Context, skipMissingSupe
 	}
 	ei.log.Debugf("needsScmFormat: %t", needsScmFormat)
 
-	if !needsMetaFormat && ei.storage.ControlMetadataHasPath() && needsScmFormat {
+	if !needsMetaFormat && ei.storage.ControlMetadataPathConfigured() {
 		cfg, err := ei.storage.GetScmConfig()
 		if err != nil {
 			return err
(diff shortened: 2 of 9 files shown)


src/control/server/instance_storage.go
```diff
@@ -99,6 +99,23 @@ func (ei *EngineInstance) awaitStorageReady(ctx context.Context, skipMissingSupe
...
+			return err
+		}
+		if (cfg.Class == "ram") && ei.storage.BdevRoleMetaConfigured() {
+			ei.log.Debugf("scm class is ram and bdev role meta configured")
+			err := ei.storage.FormatScm(true)
+			if err != nil {
+				ei.log.Errorf("instance %d: failed to setup the scm: %s", idx, err)
+			} else {
+				ei.log.Debugf("remounted scm")
+				needsScmFormat = false
```
tanabarr, line 114: You may want to explicitly mount the Metadata here as `ControlMetadataNeedsFormat()` looks like it will mount and unmount  during `NeedsFormat(req)` call and Metadata won't have been mounted elsewhere in this flow/scenario.

tanabarr (approved):

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
 	return 0;
 }
 
+static int
+recreate_pool_layout()
```
NiuYawei, line 80: The function name looks bit confusing to me, maybe recreate_pool_files() is a better name.

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
...
+{
+	struct smd_pool_info    *pool_info = NULL, *tmp;
+	d_list_t                 pool_list;
+	char                     uuid_str[DAOS_UUID_STR_SIZE];
+	int			 rc = 0, pool_list_cnt, i, fd;
+	char			*path = NULL, *pool_dir = NULL;
+	uint64_t		 blob_sz;
+
+	D_INIT_LIST_HEAD(&pool_list);
+	rc = smd_pool_list(&pool_list, &pool_list_cnt);
```
NiuYawei, line 90: As we discussed offline, the pools in newborns/zombies will be mistakenly re-created as normal pools. We might need to adjust the pool creation/destroy logic to not rely on the pool files but rely on SMD for newborns/zombies cleanup?

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
...
+					path, DP_RC(rc));
+				goto out;
+			}
+			rc = ftruncate(fd, pool_info->spi_blob_sz[SMD_DEV_TYPE_META]);
+			if (fd < 0) {
+				rc = daos_errno2der(errno);
+				D_ERROR("ftruncate on vos file %s failed:"DF_RC"\n",
+					path, DP_RC(rc));
+				goto out;
+			}
```
NiuYawei, line 129: We might need to do fallocate() like what we did on pool creation?

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
...
+		}
+		path = pool_svc_rdb_path(pool_info->spi_id);
+		if (path == NULL) {
+			D_ERROR("Cannot retrieve rdb file info associated with pool %s",
+				uuid_str);
+			rc = -DER_NONEXIST;
+			goto out;
+		}
+		rc = smd_rdb_get_blob_sz(pool_info->spi_id, &blob_sz);
+		if (rc) {
```
NiuYawei, line 140: I suppose RDB pool exists only on ranks having pool service? So we might need to tolerate the NONEXIST error, or we might need to introduce a smd_rdb_pool_list() instead of smd_rdb_get_blob_sz()?

src/control/server/storage/provider.go
```diff
@@ -99,6 +99,17 @@ func (p *Provider) ControlMetadataNeedsFormat() (bool, error) {
 	return p.metadata.NeedsFormat(req)
 }
 
+// ControlMetadataHasPath checks whether metadata section is defined
+func (p *Provider) ControlMetadataHasPath() bool {
```
kjacque, line 103: Naming suggestion - how about `ControlMetadataConfigured` to more accurately describe the intent of this method.

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
...
 
+static int
+recreate_pool_layout()
+{
+	struct smd_pool_info    *pool_info = NULL, *tmp;
+	d_list_t                 pool_list;
+	char                     uuid_str[DAOS_UUID_STR_SIZE];
+	int			 rc = 0, pool_list_cnt, i, fd;
+	char			*path = NULL, *pool_dir = NULL;
+	uint64_t		 blob_sz;
```
kjacque, line 87: Just a note on the DAOS coding style: it is recommended to put each variable on its own line for readability, even when there are multiple variables of the same type.

Here is the C style guide, for reference: https://daosio.atlassian.net/wiki/spaces/DC/pages/4836655703/C+Style+Guide

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
...
+			goto out;
+		}
+		fd = open(path, O_RDWR|O_CREAT, S_IRWXU);
+		if (fd < 0) {
+			rc = daos_errno2der(errno);
+			D_ERROR("failed to create/open the vos file %s:"DF_RC"\n", path, DP_RC(rc));
+			goto out;
+		}
+		rc = ftruncate(fd, blob_sz);
+		if (fd < 0) {
```
kjacque, line 151: Did you mean to check rc here?

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
...
+			}
+			fd = open(path, O_RDWR|O_CREAT, S_IRWXU);
+			if (fd < 0) {
+				rc = daos_errno2der(errno);
+				D_ERROR("failed to create/open the vos file %s:"DF_RC"\n",
+					path, DP_RC(rc));
+				goto out;
+			}
+			rc = ftruncate(fd, pool_info->spi_blob_sz[SMD_DEV_TYPE_META]);
+			if (fd < 0) {
```
kjacque, line 124: should be rc?

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
...
+				uuid_str);
+			rc = -DER_NONEXIST;
+			goto out;
+		}
+		rc = smd_rdb_get_blob_sz(pool_info->spi_id, &blob_sz);
+		if (rc) {
+			D_ERROR("Failed to extract the size of rdb file for pool %s", uuid_str);
+			goto out;
+		}
+		fd = open(path, O_RDWR|O_CREAT, S_IRWXU);
```
kjacque, line 144: Maybe I'm missing something but isn't a call to `close(fd)` needed after it's used?

src/pool/srv.c
```diff
@@ -72,11 +76,107 @@ fini(void)
...
+			D_GOTO(out, rc = daos_errno2der(errno));
+		}
+		for (i = 0; i < pool_info->spi_tgt_cnt[SMD_DEV_TYPE_META]; i++) {
+			D_ASPRINTF(path, "%s/%s%d", pool_dir, VOS_FILE,
+				   pool_info->spi_tgts[SMD_DEV_TYPE_META][i]);
+			if (path == NULL) {
+				D_ERROR("Unable to create target file layout for %s", uuid_str);
+				D_GOTO(out, rc = -DER_NOMEM);
+			}
+			fd = open(path, O_RDWR|O_CREAT, S_IRWXU);
```
kjacque, line 116: I don't see a corresponding `close(fd)`. Is it not needed?

src/control/server/instance_storage.go
```diff
@@ -99,6 +99,23 @@ func (ei *EngineInstance) awaitStorageReady(ctx context.Context, skipMissingSupe
 	}
 	ei.log.Debugf("needsScmFormat: %t", needsScmFormat)
 
+	if !needsMetaFormat && ei.storage.ControlMetadataHasPath() && needsScmFormat {
```
kjacque, line 102: I'd like to understand what this use case actually means. What is the situation where a user unmounts their tmpfs but doesn't want to reformat the whole system?

kjacque (changes requested):

src/control/server/instance_storage.go
tanabarr: When control Metadata is intact (on dedicated storage) after a reboot which has cleared tmpfs.

src/pool/srv.c
sherintg: changed the name to recreate_pooltgts()

src/pool/srv.c
sherintg: Taken care of in the new commit. I have added a new flag SMD_POOL_IN_CREATION in the smd table entry for SMD_DEV_TYPE_META. This flag will be set when the table entry for TABLE_POOLS[SMD_DEV_TYPE_META] is updated and cleared after the directory is renamed. This flag is checked again upon restart. If set it considers the pool as newborn. Tested the same using instrumentation.

src/pool/srv.c
sherintg: Taken care of in the new commit. Now the code is also moved to mgmt setup from pool setup to facilitate code reuse better.

src/pool/srv.c
sherintg: Added the NONEXIST check here.

src/pool/srv.c
sherintg: Corrected the code to follow the guidelines.

src/pool/srv.c
sherintg: Copy paste issue. Fixed it.

src/pool/srv.c
sherintg: Copy paste issue. Fixed it.

src/pool/srv.c
sherintg: Fixed it.

src/pool/srv.c
sherintg: Fixed it.

src/control/server/instance_storage.go
sherintg: There is a slight change in the logic now:
1. If metadata is hosted on the same tmpfs mount then , daos_server restart will reuse the mount and starts the engines.
2. If metadata is hosted in a different path (recommended for md_on_ssd), then tmpfs is force remounted every time the daos_server is restarted. The force remount will cleanup leftover files and directories in the previous run.
3. If the metadata is hosted on a different path and tmpfs is unmounted (node restart), then daos_server will mount the tmpfs and continue with the engine start. This case is covered by (2).

kjacque: Thanks for the explanations, @sherintg and @tanabarr. If the control metadata is on the same tmpfs, and the tmpfs isn't mounted, will that still trigger a need to format? In my view there is value in keeping that path alive (specifically in allowing us to test that flow without SCM).

src/mgmt/srv_target.c
```diff
@@ -336,6 +339,159 @@ cleanup_leftover_pools(bool zombie_only)
...
+	D_FREE(pool_newborn_path);
+	D_FREE(pool_path);
+
+	return rc;
+}
+
+static int
+recreate_pooltgts()
+{
+	struct smd_pool_info    *pool_info = NULL, *tmp;
```
kjacque, line 444: style - should define tmp on a separate line.

src/control/server/instance_storage.go
```diff
@@ -99,7 +99,7 @@ func (ei *EngineInstance) awaitStorageReady(ctx context.Context, skipMissingSupe
 	}
 	ei.log.Debugf("needsScmFormat: %t", needsScmFormat)
 
-	if !needsMetaFormat && ei.storage.ControlMetadataHasPath() && needsScmFormat {
+	if !needsMetaFormat && ei.storage.ControlMetadataPathConfigured() {
```
kjacque, line 102: Is there a permitted case where there is no SCM tier/no tmpfs? That case would fail getting the SCM config below.

src/bio/smd/smd_pool.c
```diff
@@ -86,12 +87,17 @@ pool_add_tgt(uuid_t pool_id, uint32_t tgt_id, uint64_t blob_id, char *table_name
 		pool.sp_tgts[pool.sp_tgt_cnt] = tgt_id;
 		pool.sp_blobs[pool.sp_tgt_cnt] = blob_id;
 		pool.sp_tgt_cnt += 1;
+		if (!strcmp(table_name, TABLE_POOLS[SMD_DEV_TYPE_META]))
```
kjacque, line 90: To avoid possible bugs with unterminated strings, we should be using the `n` versions of these string functions, e.g. `strncmp`, with some reasonable maximum length.

src/control/server/instance_storage.go
```diff
@@ -99,6 +99,23 @@ func (ei *EngineInstance) awaitStorageReady(ctx context.Context, skipMissingSupe
...
+		cfg, err := ei.storage.GetScmConfig()
+		if err != nil {
+			return err
+		}
+		if (cfg.Class == "ram") && ei.storage.BdevRoleMetaConfigured() {
+			ei.log.Debugf("scm class is ram and bdev role meta configured")
+			err := ei.storage.FormatScm(true)
+			if err != nil {
+				ei.log.Errorf("instance %d: failed to setup the scm: %s", idx, err)
```
kjacque, line 111: Why wouldn't this be fatal? Any reason to believe trying it again would work?

kjacque (commented):
A few more little things. Thanks.

src/mgmt/srv_target.c
sherintg: Thanks fixed it.

src/control/server/instance_storage.go
sherintg: SCM tier is required

src/bio/smd/smd_pool.c
sherintg: Both the arguments are static strings used with the module and are of variable sizes. Hence used strcmp().

src/control/server/instance_storage.go
sherintg: Instead of handling the failure here, I allow the code to fall through and take the default code path where it requires user to explicitly invoke format command.

src/control/server/instance_storage.go
sherintg: Yes. If tmpfs is not mounted then you lose all of the control metadata if it was earlier hosted on tmpfs.

src/mgmt/srv_target.c
```diff
@@ -336,6 +339,160 @@ cleanup_leftover_pools(bool zombie_only)
...
+	D_ASSERT(bio_nvme_configured(SMD_DEV_TYPE_META));
+	D_INIT_LIST_HEAD(&pool_list);
+	rc = smd_pool_list(&pool_list, &pool_list_cnt);
+	if (rc != 0) {
+		D_ERROR("Failed to get pool info list from SMD\n");
+		return rc;
+	}
+
+	d_list_for_each_entry_safe(pool_info, tmp, &pool_list, spi_link) {
+		/* Cleanup Newborns */
```
NiuYawei, line 461: Usually the blobs for pool in ZOMBIES should have been destroyed, but if the pool is still referenced on destroy, the blobs destroy will be deferred to ZOMBIES cleanup on next restart (or next pool creation). See tgt_destroy() -> tgt_kill_pool() -> vos_pool_kill(). But that was implemented long time ago, I think today we've ensured the pool is not referenced on destroy?

Anyway, let's keep it in mind, we could address it later if we really hit the issue.

src/bio/smd/smd_pool.c
```diff
@@ -408,3 +438,35 @@ smd_pool_replace_blobs_locked(struct smd_pool_info *info, int tgt_cnt,
...
+	rc = smd_db_fetch(TABLE_POOLS[SMD_DEV_TYPE_META], &id, sizeof(id), &pool, sizeof(pool));
+	if (rc == -DER_NONEXIST) {
+		D_GOTO(out, rc = 0);
+	} else if (rc) {
+		D_ERROR("Failed to fetch smd entry of the meta blob for "DF_UUID". "DF_RC"\n",
+			DP_UUID(&id.uuid), DP_RC(rc));
+		goto out;
+	}
+
+	pool.sp_flags &= ~SMD_POOL_IN_CREATION;
```
NiuYawei, line 461: Minor: Instead of clearing 'in_create" flag, if you set a 'ready' flag here, the code for marking 'in_create' in pool_add_tgt() could be avoided.

NiuYawei (dismissed):
The C code changes looks good to me.

src/bio/smd/smd_pool.c
sherintg: I coded the above way to handle scenario where in if one upgrades the daos bits with smd from previous bits, it continue to work fine. Else the new server would assume that the pools are newborn.

tanabarr (dismissed):
Go changes look good to me

src/control/server/instance_storage.go
kjacque: But we've already attempted the format here, and it failed. I guess I'm wondering what value there is in trying again, instead of having the server just error and quit.

src/control/server/instance_storage.go
kjacque: Hm, okay. I was under the impression MD on SSD mode would eventually stop requiring this tier.

kjacque (approved):
OK with landing. My remaining comments are not urgent.

src/bio/smd/smd_pool.c
sherintg: Fixed this to use strncmp as per suggestion from @jolivier23 as well.

NiuYawei (approved):

tanabarr (approved):

src/include/daos_srv/smd.h
```diff
@@ -40,11 +40,14 @@ struct smd_pool_info {
 	d_list_t	 spi_link;
 	uuid_t		 spi_id;
 	uint64_t	 spi_blob_sz[SMD_DEV_TYPE_MAX];
-	uint32_t	 spi_tgt_cnt[SMD_DEV_TYPE_MAX];
+	uint16_t	 spi_flags[SMD_DEV_TYPE_MAX];
+	uint16_t	 spi_tgt_cnt[SMD_DEV_TYPE_MAX];
```
gnailzenh, line 44: this will break interoperability right? I don't think we can change it. If you really need to add status for each target, I'm afraid you have to create another table.

gnailzenh: ah, nm, this is a data structure in DRAM, sorry for the noise

gnailzenh (approved):

johannlombardi:
@phender Could you please advise whether those test failures are known?

PR #11666, opened 2023-03-10
DAOS-12612 bio: Enable default WAL size

NiuYawei:
- Use a reasoanble default WAL size since checkpoint is ready.
- Add missed wakeup in bio_wal_checkpoint().

Required-githooks: true

### Gatekeeper:


src/bio/bio_context.c
```diff
@@ -623,17 +623,22 @@ __bio_ioctxt_open(struct bio_io_context **pctxt, struct bio_xs_context *xs_ctxt,
...
+	if (meta_sz <= (256ULL << 20) /* 256MB */)
+		return meta_sz * 2;
 
-	return (50ULL << 30);
-#endif
+	/*
+	 * Calculate a reasonable WAL size based on following assumptions:
+	 * - Single target update IOPS can be upto 65k;
+	 * - Each TX consumes 2 WAL blocks in average;
+	 * - Checkpointing interval is 5 seconds, and the WAL should have at least
```
wangshilong, line 636: checkpoint interval now is controllable by pool property, should we take that into account? we might define a min and max allowed interval, then use max allowed to calculate here to avoid any race?

In the long term, it might be nice to allow to specify WAL size/ration when creating pool.

wangshilong (dismissed):

src/bio/bio_context.c
NiuYawei: oh, I wasn't aware that checkpoint interval is configurable now. Yes, I agree that a min/max interval should be defined, and invalid value needs be rejected.

Specifying a reasonable WAL size requires deep knowledge on WAL, I don't think we'd expose it to user. Given that the "meta size" and "checkpoint interval" are already specified by user, and WAL size is highly correlated with these two values, why don't we internally calculate a reasonable WAL size? But we do should allow developer to specify arbitrary WAL size in tests (bio_mc_create() accepts a non-zero 'wal_sz' parameter, which is targeted for developer).

NiuYawei: Actually, making the internal "checkpoint interval" configurable by user looks not a good idea to me as well.

sherintg: It will be good to have WAL size configurable and also make checkpoint ult run based on the pressure on WAL rather than the interval. There will be scenarios where customer would prefer longer startup time replaying the WAL data rather than running checkpoint frequently.

sherintg (dismissed):

src/bio/bio_context.c
jolivier23: The main purpose of the configurable interval is so we can figure out what to set it to.  We can adjust later or remove it, if necessary.  I am working on a patch to use a threshold either instead of or in conjunction with the interval setting but I've been sidetracked trying to make sure the storage estimator is working as expected.

ivadmitry (changes requested):
When nvme size (= meta size) is less than 6GB GC vos test GC05 would hung forever because there is no checkpointing in vos_tests. Suggested change is:
        wal_sz = (6ULL << 30);  /* 6GB */

        return meta_sz > wal_sz ? wal_sz : meta_sz*2;

src/bio/bio_context.c
NiuYawei: yeah, maybe configuring checkpoint interval or WAL size is useful for the scenarios mentioned by @sherintg , so maybe we'd ask user to configure 'WAL size' or 'checkpoint interval', but not both.

wangshilong (approved):

ivadmitry (approved):
The change in default_wal_sz() answers my comment; vos_tests should be able to run on 26 GB ramfs file (data+meta+WAL=10G+10G+6G in VOS815) now.

jolivier23 (approved):

PR #11681, opened 2023-03-13
DAOS-9548 placement: add PDA (performance domain affinity) support

liuxuezhao:
Add PO_COMP_TP_GRP component type to pool map.
Add pda parameter (omd_pda) to struct daos_obj_md to generate layout.
Add pda selection and constrain to jump_map placement code.
Add unit test - "placement/tests/jump_map_pda.c".
Add a test tool to test layout distribution -
    jump_pl_map -d <-n num_objs> <-l obj_class_name>
for example - ./jump_pl_map -d -n 1024 -l EC_8P2G1
to test contiguous OIDs' layout distribution on all targets.

Control-plane side change (from Kris) -
- Add support for a second layer in fault_path in config yaml.
- Update pool map parsing to understand the second layer, if it
   exists, as PO_COMP_TP_GRP.
- Remove previous hacks to add group components.
- Fix the component count used as idx for group case

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/common/pool_map.c
+++ b/src/common/pool_map.c
@@ -1436,11 +1436,10 @@ pool_map_merge(struct pool_map *map, uint32_t version,
 }
 
 static void
-fill_domain_comp(struct pool_map *map, struct d_fd_node *domain, int idx, int map_version,
-		 uint8_t new_status, struct pool_component *map_comp)
+fill_domain_comp(struct pool_map *map, uint8_t type, struct d_fd_node *domain, int idx,
+		 int map_version, uint8_t new_status, struct pool_component *map_comp)
 {
-	/* TODO DAOS-6353: Use the layer number as type */
-	map_comp->co_type = PO_COMP_TP_NODE;
+	map_comp->co_type = type;
 	map_comp->co_status = new_status;
 	map_comp->co_index = idx;
 	map_comp->co_id = domain->fdn_val.dom->fd_id;
@@ -1474,8 +1473,11 @@ add_domain_tree_to_pool_buf(struct pool_map *map, struct pool_buf *map_buf,
 {
 	int			rc;
 	uint32_t		num_dom_comps;
+	uint32_t		num_grp_comps;
 	uint32_t		num_rank_comps;
+	uint32_t		num_comps;
 	uint8_t			new_status;
+	uint8_t			dom_type;
 	struct d_fd_tree	tree = {0};
 	struct d_fd_node	node = {0};
 	bool			updated = false;
@@ -1486,11 +1488,14 @@ add_domain_tree_to_pool_buf(struct pool_map *map, struct pool_buf *map_buf,
 		new_status = PO_COMP_ST_NEW;
 		num_dom_comps = pool_map_find_domain(map, PO_COMP_TP_NODE,
 						     PO_COMP_ID_ALL, NULL);
+		num_grp_comps = pool_map_find_domain(map, PO_COMP_TP_GRP,
+						     PO_COMP_ID_ALL, NULL);
 		num_rank_comps = pool_map_find_domain(map, PO_COMP_TP_RANK,
 						      PO_COMP_ID_ALL, NULL);
 	} else {
 		new_status = PO_COMP_ST_UPIN;
 		num_dom_comps = 0;
+		num_grp_comps = 0;
 		num_rank_comps = 0;
 	}
 
@@ -1513,22 +1518,34 @@ add_domain_tree_to_pool_buf(struct pool_map *map, struct pool_buf *map_buf,
 
 		switch (node.fdn_type) {
 		case D_FD_NODE_TYPE_DOMAIN:
-			fill_domain_comp(map, &node, num_dom_comps, map_version, new_status,
-					 &map_comp);
+			/* TODO DAOS-6353: Use the layer number as type */
+			if (d_fd_node_is_group(&node)) {
+				dom_type = PO_COMP_TP_GRP;
+				num_comps = num_grp_comps;
+			} else {
+				dom_type = PO_COMP_TP_NODE;
+				num_comps = num_dom_comps;
+			}
+			fill_domain_comp(map, dom_type, &node, num_comps, map_version,
+					 new_status, &map_comp);
 			if (map != NULL) {
 				struct pool_domain	*current;
 				int			already_in_map;
 
-				already_in_map = pool_map_find_domain(map, PO_COMP_TP_NODE,
+				already_in_map = pool_map_find_domain(map, dom_type,
 								      map_comp.co_id, &current);
 				if (already_in_map > 0) {
 					D_DEBUG(DB_TRACE, "domain %d/%u already in map\n",
 						map_comp.co_id, current->do_comp.co_status);
 					map_comp.co_status = current->do_comp.co_status;
 					map_comp.co_index = current->do_comp.co_index;
+				} else if (dom_type == PO_COMP_TP_GRP) {
+					num_grp_comps++;
 				} else {
 					num_dom_comps++;
 				}
+			} else if (dom_type == PO_COMP_TP_GRP) {
+				num_grp_comps++;
 			} else {
 				num_dom_comps++;
 			}
@@ -1603,7 +1620,6 @@ gen_pool_buf(struct pool_map *map, struct pool_buf **map_buf_out, int map_versio
 	uint32_t		num_comps;
 	uint8_t			new_status;
 	int			i, rc;
-	unsigned		num_pd = 0;
 	uint32_t		num_domain_comps = 0;
 	d_rank_list_t		*ordered_ranks;
 
@@ -1618,21 +1634,6 @@ gen_pool_buf(struct pool_map *map, struct pool_buf **map_buf_out, int map_versio
 	D_ASSERT(num_domain_comps > 0);
 	num_domain_comps--; /* remove the root domain - allocated separately */
 
-	if (map == NULL) {
-		/* Temporary hack to pass in the performance domain info before control plane
-		 * support available.
-		 */
-		d_getenv_int("DAOS_PD_NUM", &num_pd);
-		if (num_pd < 2)
-			num_pd = 0;
-		if (num_pd > num_domain_comps) {
-			D_ERROR("invalid num_perf_dom %d > num_fault_dom %d\n",
-				num_pd, num_domain_comps);
-			return -DER_INVAL;
-		}
-		num_domain_comps += num_pd;
-	}
-
 	ordered_ranks = d_rank_list_alloc(nnodes);
 	if (ordered_ranks == NULL)
 		return -DER_NOMEM;
@@ -1641,36 +1642,6 @@ gen_pool_buf(struct pool_map *map, struct pool_buf **map_buf_out, int map_versio
 	if (map_buf == NULL)
 		D_GOTO(out_ranks, rc = -DER_NOMEM);
 
-	/* firstly add performance domain */
-	if (num_pd > 0) {
-		uint32_t	num_fdom, fdom_per_pd;
-
-		num_fdom = num_domain_comps - num_pd;
-		fdom_per_pd = num_fdom / num_pd;
-		for (i = 0; i < num_pd; i++) {
-			map_comp.co_type = PO_COMP_TP_GRP;
-			map_comp.co_status = PO_COMP_ST_UPIN;
-			map_comp.co_padding = 0;
-			map_comp.co_id = i;
-			map_comp.co_rank = i;
-			map_comp.co_ver = map_version;
-			map_comp.co_in_ver = map_version;
-			map_comp.co_fseq = 1;
-			map_comp.co_flags = PO_COMPF_NONE;
-			map_comp.co_nr = fdom_per_pd;
-			if ((num_fdom % num_pd) != 0 && i < (num_fdom % num_pd))
-				map_comp.co_nr++;
-
-			rc = pool_buf_attach(map_buf, &map_comp, 1 /* comp_nr */);
-			if (rc != 0) {
-				D_ERROR("failed to attach to pool buf, "DF_RC"\n",
-					DP_RC(rc));
-				D_GOTO(out_map_buf, rc);
-			}
-		}
-	}
-
-	/* then add fault domain */
 	rc = add_domain_tree_to_pool_buf(map, map_buf, map_version, dss_tgt_nr, ndomains,
 					 domains, ordered_ranks);
 	if (rc != 0) {
--- a/src/common/pool_map.c
+++ b/src/common/pool_map.c
@@ -137,6 +137,11 @@ static struct pool_comp_type_dict comp_type_dict[] = {
 		.td_name	= "node",
 	},
 	{
+		.td_type	= PO_COMP_TP_GRP,
+		.td_abbr	= 'g',
+		.td_name	= "group",
+	},
+	{
 		.td_type	= PO_COMP_TP_ROOT,
 		.td_abbr	= 'o',
 		.td_name	= "root",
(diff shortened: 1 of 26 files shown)


src/include/daos/cont_props.h
```diff
@@ -158,4 +158,10 @@ daos_cont_prop2global_version(daos_prop_t *prop);
 uint32_t
 daos_cont_prop2obj_version(daos_prop_t *prop);
 
+static inline uint32_t
+daos_cont_props2pda(struct cont_props *props, bool is_ec_obj)
+{
+	return is_ec_obj ? props->dcp_ec_pda : props->dcp_rp_pda;
```
wangdi1, line 164: Just curious, what is the difference between ec_pda and rp_pda?

src/placement/jump_map_versions.c
```diff
@@ -304,6 +308,17 @@ __get_target_v1(struct pool_domain *curr_dom, struct pool_target **target,
...
 				if (top == -1) {
+					if (curr_pd != root_pos) {
+						/* all domains within the PD are full, ignore the
+						 * PD restrict.
+						 */
+						D_DEBUG(DB_PL, "PD[%d] all doms are full, weak the "
+							"PD restrict\n",
+							(int)(curr_dom - root_pos));
+						curr_pd = root_pos;
+						curr_dom = curr_pd;
```
wangdi1, line 319: hmm, I thought you need re-generate the pd, instead of weak the pd? since there might be other PD yet? or I miss sth?

Btw: I thought __get_target_v1 () can handle multiple domain layer anyway, why separate the pd in a different function, so you may see problem like this?

wangdi1 (commented):
Sorry, some questions for pd in get_layout_v1. maybe we can discuss a bit.

src/common/pool_map.c
```diff
@@ -451,23 +456,29 @@ pool_buf_unpack(struct pool_buf *buf)
...
-								buf->pb_nr) {
+	    buf->pb_domain_nr + buf->pb_node_nr + buf->pb_target_nr != buf->pb_nr) {
 		D_DEBUG(DB_MGMT, "Invalid number of components: %d/%d/%d/%d\n",
 			buf->pb_nr, buf->pb_domain_nr, buf->pb_node_nr,
 			buf->pb_target_nr);
 		return -DER_INVAL;
 	}
+	if (buf->pb_domain_nr != 0 && buf->pb_comps[0].co_type != PO_COMP_TP_GRP &&
+	    buf->pb_comps[0].co_type != PO_COMP_TP_NODE) {
+		D_DEBUG(DB_MGMT, "Invalid co_type %s[%d] for first domain.\n",
```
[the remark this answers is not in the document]
liuxuezhao: ok, just because original code at above L471 use debug loglevel so I also use debug here. Can change it here later if need to refresh. But here is for invalid arg check for very cornel case.

src/include/daos/cont_props.h
liuxuezhao: it is just for EC object and replica obj. -
141 /**                                                                                                 
142  * the placement algorithm should place two-way and three-way                                       
143  * replication object within a PD; for those object classes with                                    
144  * more than 3 replicas, DAOS will place three replicas within a PD                                 
145  * and switch to another PD.                                                                        
146  */                                                                                                 
147 #define DAOS_PROP_PO_RP_PDA_DEFAULT>    3                                                           
148 /**                                                                                                 
149  * the placement algorithm always tries to scatter shards of EC                                     
150  * object to different PDs.                                                                         
151  */                                                                                                 
152 #define DAOS_PROP_PO_EC_PDA_DEFAULT>    1

src/placement/jump_map_versions.c
liuxuezhao: for weak the PD, it is only for the case that there are small number of PD and in each PD with small number of fault doms. There is a test case of "pd_use_all_dom", with 2 PD, 4 fdom in each PD and 1 tgt in each fdom (8 tgts totally), to place OC_RP_4G2 obj with pda 3, without the code here that test case cannot pass. In this case If there are 3 PDs in the system it will not get in this code block.

"I thought __get_target_v1 () can handle multiple domain layer anyway, why separate the pd in a different function, so you may see problem like this"
the current way in this patch seems with less code change, and I tried to avoid adding another bitmap for the PD. we can discuss more details later.

kjacque (approved):

src/include/daos/cont_props.h
wangdi1: >the placement algorithm should place two-way and three-way
143 * replication object within a PD;

So this will favor write? not read?

>the placement algorithm always tries to scatter shards of EC
150 * object to different PDs.

But this will favor read? not write?  Or I miss sth?

src/placement/jump_map.c
```diff
@@ -35,32 +35,114 @@ enum PL_OP_TYPE {
...
+	uint32_t		 selected_pd, pd_id;
+	uint32_t		 pd_nr = jmap->jmp_pd_nr;
+	uint32_t		 dom_nr = jmap->jmp_domain_nr;
+	uint32_t		 shard_nr = jmop->jmop_grp_size * jmop->jmop_grp_nr;
+	uint32_t		 doms_per_pd;
+	uint32_t		 pd_grp_size, pd_grp_nr;
+	int			 i, rc;
+
+	jmop->jmop_root = root;
+	if (md->omd_pda == 0)
```
wangdi1, line 66: it is 0 by default?  will this cause any incompatibility issue?

src/placement/jump_map.c
```diff
@@ -35,32 +35,114 @@ enum PL_OP_TYPE {
...
+		pd_nr = 0;
+	jmop->jmop_pd_nr = pd_nr;
+	if (jmop->jmop_pd_nr == 0) {
+		jmop->jmop_pd_grp_size = 0;
+		return 0;
+	}
+
```
wangdi1, line 73: Probably we need a env setting for 2.4? only enable this if specific env are setting, to avoid it interfering sth for 2.4 release?

wangdi1 (dismissed):
Please adding an environment setting or FAIL, if this is not intend to 2.4 by default? or it will not enable by default?

src/include/daos/cont_props.h
liuxuezhao: > > the placement algorithm should place two-way and three-way
> > 143 * replication object within a PD;
> 
> So this will favor write? not read?

Write could be faster as IO forward is faster, for read it is not bad right? as any replica can be read.
Rebuild could be better also as fetch within PD.

> 
> > the placement algorithm always tries to scatter shards of EC
> > 150 * object to different PDs.
> 
> But this will favor read? not write? Or I miss sth?

EC possibly need to read from all shards, so spread to multiple PD?
anyway it is not a problem of this patch.

src/placement/jump_map.c
liuxuezhao: Now the pool map is without PD, so will not cause incompatibility now.

src/placement/jump_map.c
liuxuezhao: now pool map only will add PD when with special fault_path config, so by default the PD code will not affect the placement generation.

src/placement/jump_map.c
```diff
@@ -560,7 +651,8 @@ get_object_layout(struct pl_jump_map *jmap, uint32_t layout_ver, struct pl_obj_l
 				setbit(tgts_used, target->ta_comp.co_id);
 				setbit(dom_cur_grp_used, domain - root);
 			} else {
-				get_target(root, layout_ver, &target, key, dom_used,
+				curr_pd = jm_obj_shard_pd(jmop, k);
```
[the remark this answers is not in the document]
liuxuezhao: "pd_idx = (shard / jmop->jmop_pd_grp_size) % jmop->jmop_pd_nr;" in jm_obj_shard_pd().
so if pda >= 3 and the 3 replica will be in same PD, with test cases in the UT.

liuxuezhao:
> Please adding an environment setting or FAIL, if this is not intend to 2.4 by default? or it will not enable by default?

the code does not affect the default placement generation right?

src/include/daos/cont_props.h
wangdi1: >Write could be faster as IO forward is faster, for read it is not bad right? as any replica can be read.
Rebuild could be better also as fetch within PD.

Then any replica read will all comes from 1 PD, i.e. using bandwidth from single PD.

> EC possibly need to read from all shards, so spread to multiple PD?

So EC  multiples shards will spread to multiple PD for read? so why not for replica objects?  Anyway I do not see how it is different between EC and replica?

src/placement/jump_map.c
wangdi1: I see. thanks

src/placement/jump_map.c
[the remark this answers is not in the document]
wangdi1: I see. thanks

src/include/daos/cont_props.h
liuxuezhao: > > Write could be faster as IO forward is faster, for read it is not bad right? as any replica can be read.
> > Rebuild could be better also as fetch within PD.
> 
> Then any replica read will all comes from 1 PD, i.e. using bandwidth from single PD.
> 
> > EC possibly need to read from all shards, so spread to multiple PD?
> 
> So EC multiples shards will spread to multiple PD for read? so why not for replica objects? Anyway I do not see how it is different between EC and replica?

below is quoted from liang's HLD -
Besides adaptive routing during regular I/O, another thing that should be considered for object placement is switch-to-switch hops of server-side message forwarding for two-phase commit write, and data transferring for rebuilding objects.
Distributed I/O of DAOS is primary-slave model, where the client only sends RPC to the primary server of a redundancy group, and then the primary server forwards the RPC and RDMA descriptor to other servers. For a large EC object, avoiding one or few objects consume all bandwidth of a performance domain is the main concern, so spreading shards to many groups can avoid bandwidth hogging and benefit overall performance. However, for an object with a small redundancy group size (e.g., three-way replicated object), bandwidth consumption is already limited by shards (targets) number. Instead of spreading shards to different performance domains, placing shards of this type of object in the same performance domain is the optimal option because it has a few benefits:
• Avoid extra network-hops of forwarding RPCs between servers and reduce latency of regular I/O.
• Reduce data transfer between DAOS groups during rebuild.

gnailzenh: > > Write could be faster as IO forward is faster, for read it is not bad right? as any replica can be read.
> > Rebuild could be better also as fetch within PD.
> 
> Then any replica read will all comes from 1 PD, i.e. using bandwidth from single PD.
> 
> > EC possibly need to read from all shards, so spread to multiple PD?
> 
> So EC multiples shards will spread to multiple PD for read? so why not for replica objects? Anyway I do not see how it is different between EC and replica?

Replication is mostly for small objects, so it's latency sensitive, whereas EC is for large object which want more bandwidth

src/include/daos/object.h
```diff
@@ -135,6 +135,8 @@ struct daos_obj_md {
 	 * use pl_map's default value PL_DEFAULT_DOMAIN (PO_COMP_TP_RANK).
 	 */
 	uint32_t		omd_fdom_lvl;
+	/* Performance domain affinity */
+	uint32_t		omd_pda;
```
gnailzenh, line 139: not issue of this patch, but the comment "object metadata stored in the global OI table of container" is out-of-date, it's not in OI table

gnailzenh (dismissed):

src/include/daos/cont_props.h
wangdi1: Well, multiple groups(GX, SX) also intends for scalability and bandwidth, which can have smaller replication number.  So we should also choose multiple PD for this case as well?

@liuxuezhao  how do we choose PD for this case?

wangdi1: Indeed, though EC aggregation and rebuild still need cross the PD.

src/include/daos/object.h
liuxuezhao: right, I was confused when read that comment too. can remove it later in following patch.

liuxuezhao:
@gnailzenh or @jolivier23 the PDA code in this patch does not affect default layout generation (it only affect it when with special pool map which need to be enabled by special fault_path set), could it be landed please? thanks.

liuxuezhao:
@wangdi1 @gnailzenh a rebase and pushed a few bug fixing change, the commit 3 is what changed compared to your last time review. thanks.

src/placement/jump_map_versions.c
```diff
@@ -317,15 +319,19 @@ __get_target_v1(struct pool_domain *root_pos, struct pool_domain *curr_pd,
 						      start_dom, end_dom, exclude_new);
 			if (range_set) {
 				if (top == -1) {
-					if (curr_pd != root_pos) {
-						/* all domains within the PD are full, ignore the
-						 * PD restrict.
+					if (curr_pd != root_pos && grp_size > 1) {
```
wangdi1, line 322: let's say if there are 2 failure domain under one PD, and PD_number = 2,  then creating the object RP2_G2, so this should be created under 1 PD or multiple PD?

liuxuezhao: I just tested the case -
With 2 server groups (PDs),
     2 servers for each group,
     1 engines for each server,
     1 targets for each engine,
     4 targets totalty (each server group with 2 targets).

layout of OC_RP_2G2 object, PDA 2 ---
[  0=>  3_PD1   1=>  2_PD1][  2=>  1_PD0   3=>  0_PD0]

layout of OC_RP_2G2 object, PDA 1 -
[  0=>  3_PD1   1=>  1_PD0][  2=>  2_PD1   3=>  0_PD0]
the number means -
[shard_idx => tgt_id_PD_idx]

wangdi1: I mean if there are 2 targets per engine, then RP2_G2 will be placed to 2PD, instead of 1PD? is this expected for PD_number = 2.

liuxuezhao: yes, will be placed to 2 PDs, the test result is -

With 2 server groups (PDs),
     2 servers for each group,
     1 engines for each server,
     2 targets for each engine,
     8 targets totalty (each server group with 4 targets).

layout of OC_RP_2G2 object, PDA 2 ...
[  0=>  7_PD1   1=>  5_PD1][  2=>  2_PD0   3=>  0_PD0]

layout of OC_RP_2G2 object, PDA 1 ...
[  0=>  7_PD1   1=>  2_PD0][  2=>  5_PD1   3=>  0_PD0]

looks that is good? or what kind of other layout you expect?

wangdi1: Oh, I thought PD_number = 2 means the it will put 2 replicas in one PD, so in this case, I thought all shards will be put to 1 PD? or I misunderstand the PD_number ?

liuxuezhao: no, PD_number is the number of PD( group), PDA (PD affinity) is the number of shards in one PD, I think the "PD number" in your mind is PDA.
Even for PDA = 2 case, for RP2G2, the 2 shards in the grp 0 will be put in one PD, but the 2 shards in grp 1 should be put in another PD.

wangdi1: >Even for PDA = 2 case, for RP2G2, the 2 shards in the grp 0 will be put in one PD, but the 2 shards in grp 1 should be put in another PD.

Ah. Sure, thanks.

> If you want to put RP2G2's 4 shards in one PD, you should set PDA = 4
but a special thing is we only have 2 fault doms in one PD, the actual PDA was changed to 2 (see jmop->jmop_pd_grp_size init code in jm_obj_pd_init()).

Hmm, I thought in this case if PDA=4, then all shards should be in 1 PD? since different group can be put to the same domain anyway?

wangdi1 (dismissed):

src/placement/jump_map_versions.c
liuxuezhao: > > If you want to put RP2G2's 4 shards in one PD, you should set PDA = 4
> > but a special thing is we only have 2 fault doms in one PD, the actual PDA was changed to 2 (see jmop->jmop_pd_grp_size init code in jm_obj_pd_init()).
> 
> Hmm, I thought in this case if PDA=4, then all shards should be in 1 PD? since different group can be put to the same domain anyway?

for RP4G1, should put to 2 PDs in that case. so for RP2G2, now did not differentiate and also put to 2 PDs

liuxuezhao:
@wangdi1 @gnailzenh just added a small refining at commit 4.

wangdi1 (approved):

gnailzenh (approved):

PR #11684, opened 2023-03-13
DAOS-12854 control: Allow system query to retry in PoolCreate

mjmac:
When the pool create size is specified as a percentage, the
PoolCreate handler needs to scan available storage across
the servers in order to determine how much space can be
allocated for the pool. A previous commit added a pre-scan
SystemQuery in order to avoid running the scan before the
MS was up and running, but that change did not allow the
query to be retried, leading to problems when creating pools
on larger systems.

This commit enables the normal retry mechanism for the
pre-scan SystemQuery so that it behaves like most other
RPCs. In particular, it will properly handle an ErrNotReplica
and retry with the replica hosts set in the error.

Required-githooks: true


mjmac:
Received verification that this patch fixes the issue. Rebuild failure is known and unrelated.

tanabarr (approved):

knard-intel (approved):
LGTM

kjacque (approved):

PR #11690, opened 2023-03-14
DAOS-12146 cq: Update daos_pylint for newer VS Code options.

ashleypittman:
Update for newer VS code versions.

The --version flag was not working without --import... so make that automatic
where required.
VS Code is now supplying a new option to pylint which wasn't supported
so add that as well.

Required-githooks: true


utils/cq/daos_pylint.py
```diff
@@ -17,8 +17,21 @@
...
+        try:
+            from pylint.lint import Run
+            from pylint.reporters.collecting_reporter import CollectingReporter
+            from pylint.constants import full_version
+        except ImportError:
+            print('detected venv unusable, install pylint to enable this check')
+            sys.exit(0)
+    else:
+        print('install pylint to enable this check')
+        sys.exit(0)
```
daltonbohning, line 34: There are no other paths
```suggestion
    print('install pylint to enable this check')
    sys.exit(0)
```

ashleypittman: I don't think that would work as if the import works on L28 then it would catch this and we don't want it to.

jolivier23 (approved):

PR #11703, opened 2023-03-15
DAOS-12595 dtx: support CPU yield during local TX commit

Nasf-Fan:
Under original PMDK based local TX model, we assume that current ULT will not yield inside the local PMDK TX. But such assumption is not guaranteed under other backend local TX model, such as on SSD based system.

The patch introduces three temporary status for active DTX entry in DRAM: 'preparing', 'committing' and 'aborting'. Base on these, It adjusts related DTX logic to allow CPU yield during DTX status changes, including:

1. 'init' => 'preparing' => 'prepared'
2. 'init' => 'committing' => 'committed'
3. 'prepared' => 'committing' => 'committed'
4. 'prepared' => 'aborting' => 'aborted'

It works for both PMDK based local TX and metadata on SSD model. It does not change on-wire RPC protocol. As for the on-disk data layout, it is also compatible with old release. In the patch, we reuse the DTX entry local index (a 32-bits integer) in the data record (vos_irec_df::ir_dtx, evt_desc::dc_dtx, ilog_id::id_tx_id) to specially indicate whether related DTX is for solo transaction (single modification against single replicated object) or not via the highest (31th) one bit. For old release, it will not find any DTX entry if the DTX local index with such bit. Then according to old logic, related DTX will be regarded as 'committed', that just is correct and expected.

### Gatekeeper:

### DIFF
--- a/src/vos/vos_internal.h
+++ b/src/vos/vos_internal.h
@@ -95,6 +95,14 @@ enum {
 	DTX_LID_RESERVED,
 };
 
+/**
+ * If the highest bit (31th) of the DTX entry offset is set, then it is for
+ * solo (single modification against single replicated object) transaction.
+ */
+#define DTX_LID_SOLO_BITS	31
+#define DTX_LID_SOLO_FLAG	(1UL << DTX_LID_SOLO_BITS)
+#define DTX_LID_SOLO_MASK	(DTX_LID_SOLO_FLAG - 1)
+
 /*
  * When aggregate merge window reaches this size threshold, it will stop
  * growing and trigger window flush immediately.
@@ -284,6 +292,10 @@ struct vos_container {
 	uint64_t		vc_io_nospc_ts;
 	/* The (next) position for committed DTX entries reindex. */
 	umem_off_t		vc_cmt_dtx_reindex_pos;
+	/* The epoch for the latest committed solo DTX. Any solo
+	 * * transaction with older epoch must have been committed.
+	 */
+	daos_epoch_t		vc_solo_dtx_epoch;
 	/* Various flags */
 	unsigned int		vc_in_aggregation:1,
 				vc_in_discard:1,
@@ -328,11 +340,14 @@ struct vos_dtx_act_ent {
 	struct dtx_handle		*dae_dth;
 
 	unsigned int			 dae_committable:1,
+					 dae_committing:1,
 					 dae_committed:1,
+					 dae_aborting:1,
 					 dae_aborted:1,
 					 dae_maybe_shared:1,
 					 /* Need validation on leader before commit/committable. */
 					 dae_need_validation:1,
+					 dae_preparing:1,
 					 dae_prepared:1;
 };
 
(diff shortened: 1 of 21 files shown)


src/vos/vos_internal.h
```diff
@@ -284,6 +292,10 @@ struct vos_container {
 	uint64_t		vc_io_nospc_ts;
 	/* The (next) position for committed DTX entries reindex. */
 	umem_off_t		vc_cmt_dtx_reindex_pos;
+	/* The epoch for the latest committed solo DTX. Any solo
+	 * * transaction with older epoch must has been committed.
+	 */
+	daos_epoch_t		vc_solo_dtx_epoch;
```
NiuYawei, line 298: It looks to me this 'vc_solo_dtx_epoch' should be per pool. (in vos_pool but not vos_container)

NiuYawei (dismissed):

src/vos/vos_internal.h
Nasf-Fan: Per pool also work, but current DTX table is per container based, so the vc_solo_dtx_epoch keeps the same.

src/vos/vos_internal.h
```diff
@@ -284,6 +292,10 @@ struct vos_container {
 	uint64_t		vc_io_nospc_ts;
 	/* The (next) position for committed DTX entries reindex. */
 	umem_off_t		vc_cmt_dtx_reindex_pos;
+	/* The epoch for the latest committed solo DTX. Any solo
+	 * * transaction with older epoch must has been committed.
```
jolivier23, line 296: s/must has/must have

jolivier23 (approved):

Nasf-Fan:
target_auto_eviction failed for DAOS-12958.
osa_online_drain_oclass failed for DAOS-12965.
rebuild_ec failed for DAOS-12420.
aggregate_ec failed for DAOS-12883.
Not related with the patch.

NiuYawei (approved):

jolivier23 (approved):

Nasf-Fan:
target_auto_eviction failed for DAOS-12958.
create_pool_quantity failed for DAOS-12899.
rebuild_ec failed for DAOS-12883.
Not related with the patch.

jolivier23 (changes requested):
See my comment on the other PR. I don't think we should land this to master branch but rather should land to feature branch only and let it land to master when the feature branch changes lands.  Otherwise, there will be merge conflicts and I don't see a great reason to land this to master branch first.

Nasf-Fan:
> See my comment on the other PR. I don't think we should land this to master branch but rather should land to feature branch only and let it land to master when the feature branch changes lands. Otherwise, there will be merge conflicts and I don't see a great reason to land this to master branch first.

I would suggest to land the patch to master firstly. This patch changes a lot of DTX related code points. If we only land it to vos_on_blob currently, then most of further DTX related changes on master will cause the merging on vos_on_blob to hit conflict, that will be more trouble.

After landing to master, then we can resolve the merge conflict for once. The merge conflict solution can reference to the PR https://github.com/daos-stack/daos/pull/11722.

jolivier23:
> > See my comment on the other PR. I don't think we should land this to master branch but rather should land to feature branch only and let it land to master when the feature branch changes lands. Otherwise, there will be merge conflicts and I don't see a great reason to land this to master branch first.
> 
> I would suggest to land the patch to master firstly. This patch changes a lot of DTX related code points. If we only land it to vos_on_blob currently, then most of further DTX related changes on master will cause the merging on vos_on_blob to hit conflict, that will be more trouble.
> 
> After landing to master, then we can resolve the merge conflict for once. The merge conflict solution can reference to the PR #11722.

to see what would happen, I tried to take your branch against master, merge it with latest master (which is up-to-date with feature/vos_on_blob) and then I created a branch based on feature/vos_on_blob and tried to merge.  It has a fair number of merge conflicts.  My question here is why is it needed on master branch at all?   The feature is supposed to land soon to master before 2.4 anyway.  So the window for any major changes to dtx is potentially pretty small. I believe this is the last major thing we need to be tech preview feature complete.

Nasf-Fan:
> > > See my comment on the other PR. I don't think we should land this to master branch but rather should land to feature branch only and let it land to master when the feature branch changes lands. Otherwise, there will be merge conflicts and I don't see a great reason to land this to master branch first.
> > 
> > 
> > I would suggest to land the patch to master firstly. This patch changes a lot of DTX related code points. If we only land it to vos_on_blob currently, then most of further DTX related changes on master will cause the merging on vos_on_blob to hit conflict, that will be more trouble.
> > After landing to master, then we can resolve the merge conflict for once. The merge conflict solution can reference to the PR #11722.
> 
> to see what would happen, I tried to take your branch against master, merge it with latest master (which is up-to-date with feature/vos_on_blob) and then I created a branch based on feature/vos_on_blob and tried to merge. It has a fair number of merge conflicts. My question here is why is it needed on master branch at all? The feature is supposed to land soon to master before 2.4 anyway. So the window for any major changes to dtx is potentially pretty small. I believe this is the last major thing we need to be tech preview feature complete. Also, I already merge pretty frequently so master shouldn't get too badly out-of-date with respect to the changes on the branch.

I do not think we can land vos_on_blob to master in a short time, because I have tested the latest vos_on_blob (with "Test-nvme: auto_md_on_ssd"), even if with my patch, there are still tens of CI test failures. As my understand, most of them are feature branch special. We may have to spend sometime on resolving them before landing (the vos_on_blob branch) to master. Under such situation, landing this patch to master may be more easy, and then merge master into vos_on_blob will be more operable.

jolivier23 (approved):

PR #11723, opened 2023-03-16
DAOS-12911 dtx: handle stale pool map when prepare TX commit

Nasf-Fan:
dc_tx_commit_prepare() may return -DER_STALE because of stale pool map. Under such case, we need to refresh the pool map to avoid subsequent TX restart falling into dead loop.

Some other code cleanup related with task completion.

### Gatekeeper:

### DIFF
--- a/src/object/obj_tx.c
+++ b/src/object/obj_tx.c
@@ -1305,8 +1305,12 @@ dc_tx_classify_common(struct dc_tx *tx, struct daos_cpd_sub_req *dcsr,
 			continue;
 		}
 
-		if (rc != 0)
+		if (rc != 0) {
+			if (rc == -DER_STALE)
+				D_WARN("Used stale pool map %u/%u for "DF_DTI"\n",
+				       obj->cob_version, tx->tx_pm_ver, DP_DTI(&tx->tx_id));
 			goto out;
+		}
 
 		if (reasb_req != NULL && reasb_req->tgt_bitmap != NIL_BITMAP &&
 		    isclr(reasb_req->tgt_bitmap, shard->do_shard % daos_oclass_grp_size(oca))) {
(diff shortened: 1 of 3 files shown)


Nasf-Fan:
Hit DAOS-12410, to be retested.

Nasf-Fan:
Rebuild_26 failed for DAOS-12420 again, not related with the patch.

liw (approved):

src/object/obj_tx.c
```diff
@@ -2200,13 +2203,34 @@ dc_tx_commit_trigger(tse_task_t *task, struct dc_tx *tx, daos_tx_commit_t *args)
...
+			 */
+			D_ERROR("Failed to refresh pool map for "DF_DTI": "DF_RC"\n",
+				DP_DTI(&tx->tx_id), DP_RC(rc));
+		} else {
+			rc = -DER_TX_RESTART;
+			tx->tx_status = TX_FAILED;
+			/* Unlock for dc_tx_commit held. */
+			D_MUTEX_UNLOCK(&tx->tx_lock);
+			locked = false;
+			tse_task_schedule(pool_task, true);
```
liuxuezhao, line 2221: need not set dependency between task and pool_task? do you need to ensure pool map refresh finished before re-execute task?

Nasf-Fan: Normally, if the original task will be retried automatically, it needs to depend on the pool_task. But for DTX case, we will fail current TX via DER_TX_RESTART. As for how the upper later caller to handle DER_TX_RETART, that depends on the caller, it may restart the TX with re-executing related operations and then commit the new TX (with new epoch), or it can fail the TX directly. So here, we kickoff the pool task directly. If it is completed before TX restarted, then it is fine; otherwise, it may cause another pool query task if the TX restarted so quickly, but that is not fatal.

liuxuezhao (approved):

PR #11724, opened 2023-03-16
DAOS-12561 control: Consistently format pool tier ratios

mjmac:
Use a common helper when printing tier ratio values to ensure
consistent output. Also adds more comprehensive unit testing
for the tierRatioFlag type.

Required-githooks: true

### DIFF
--- a/src/control/cmd/dmg/pool_test.go
+++ b/src/control/cmd/dmg/pool_test.go
@@ -19,11 +19,12 @@ import (
 
 	"github.com/dustin/go-humanize"
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/pkg/errors"
 
+	"github.com/daos-stack/daos/src/control/cmd/dmg/pretty"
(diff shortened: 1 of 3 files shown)


src/control/cmd/dmg/pool_test.go
```diff
@@ -19,21 +19,116 @@ import (
...
+			input:     "45.3",
+			expRatios: []float64{0.453, 0.547},
+			expString: "45.30%,54.70%",
+		},
+		"valid two tiers": {
+			input:     "0.3,99.7",
+			expRatios: []float64{0.003, 0.997},
+			expString: "0.30%,99.70%",
+		},
+		"valid three tiers": {
```
knard-intel, line 100: For future support of level 3 storage ?

knard-intel (approved):
LGTM

src/control/cmd/dmg/pool_test.go
mjmac: Maybe? More just verifying that the type doesn't impose arbitrary constraints on the number of tiers. Those constraints (if any) would be the responsibility of the application code, not the input handling code.

tanabarr (approved):

PR #11741, opened 2023-03-20
DAOS-12577 dfuse: Update dfuse cache on lookup.

ashleypittman:
When dfuse replies to a lookup request then save the data it passes
to the kernel.  Previously only getattr requests were updating the
cache so it was possible to populate the cache, have a remote update
and then have loopup mark the cache as valid without updating it's
contents.  This was particuarly visable for writeback cache where
the kernel queries dfuse to check the file-size before read and dfuse
could return known-stale cache data.

Test-tag: dfuse

Required-githooks: true


ashleypittman:
Tests to follow in https://github.com/daos-stack/daos/pull/11599 but wanted to land the fix as as small a PR as possible.

mchaarawi (approved):

PR #11758, opened 2023-03-21
DAOS-12708 cart: Mark HG_FAULT as DER_INVAL.

ashleypittman:
This avoids a retry on non-recoverable errors and therefore
an infinite loop.

Required-githooks: true


src/cart/crt_hg.h
```diff
@@ -154,6 +154,8 @@ crt_hgret_2_der(int hg_ret)
 		return -DER_NOMEM;
 	case HG_CANCELED:
 		return -DER_CANCELED;
+	case HG_FAULT:
```
frostedcmos, line 157: I would group this with line 148 to be consistent with how other errors are handled, but otherwise I don't see an issue with this. It might not fix the root cause of 12708 but should at least eliminate incorrect repeat logic caused by -DER_HG default return.
@soumagne @jgmoore-or Any objections?

soumagne (approved):

src/cart/crt_hg.h
jgmoore-or: I'm fine with this. UCX NA layer doesn't use the FAULT return code, so no impact there.

jgmoore-or (approved):

PR #11766, opened 2023-03-22
DAOS-12910 tools: Move daos cont list-snap logic to Go

mjmac:
In order to support JSON output for this command, the
listing logic needed to be ported to Go.

Required-githooks: true

### DIFF
--- a/src/control/cmd/daos/snapshot.go
+++ b/src/control/cmd/daos/snapshot.go
@@ -7,7 +7,11 @@
 package main
 
 import (
+	"unsafe"
+
 	"github.com/pkg/errors"
+
+	"github.com/daos-stack/daos/src/control/lib/daos"
 )
 
 /*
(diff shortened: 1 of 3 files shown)


daltonbohning (commented):
FYI, the functional test `test_create_list_delete` uses `daos cont list-snap`. But it shouldn't need to be ran unless the command output with non-json changes.

mjmac:
> FYI, the functional test `test_create_list_delete` uses `daos cont list-snap`. But it shouldn't need to be ran unless the command output with non-json changes.

Yeah, I deliberately left the non-JSON output format the same (even though it looks a bit weird to me), e.g.
```
$ daos cont list-snap tank cont
Container's snapshots :
0xf8cc6fbe1480000
0xf8cc701ee280000
```

kjacque (approved):

src/control/cmd/daos/snapshot.go
```diff
@@ -129,11 +133,71 @@ func (cmd *containerSnapListCmd) Execute(args []string) error {
...
+	C.memset(unsafe.Pointer(&anchor), 0, C.sizeof_daos_anchor_t)
+	if snapCount > 0 {
+		rc = C.daos_cont_list_snap(ap.cont, &snapCount, &epochs[0], &names[0], &anchor, nil)
+		if err := daosError(rc); err != nil {
+			return errors.Wrapf(err, "failed to list snapshots for container %s", cmd.ContainerID())
+		}
+	}
+
+	if expectedSnaps < snapCount {
+		return errors.New("snapshot list has been truncated (size changed)")
```
knard-intel, line 165: Nit, the message is a little bit confusing for me, as the list of snapshots was not truncated, but expanded with new snaps. However, I agree that some snapshot information are missing in epochs and names structure.

knard-intel (approved):
Mostly OK for me, just one minor remark.

PR #11803, opened 2023-03-27
DAOS-0000 dfuse: Add an option to disable readdir plus.

ashleypittman:
Required-githooks: true


ashleypittman:
Having looked into DAOS-12971 I don't think this is worth pursuing now.

PR #11841, opened 2023-03-30
DAOS-13043: vos_tests: Sleep in WAL tests setup when having WAL blob …

ivadmitry:
…on AIO bdev

Initially written as a workaround for DAOS-13040 this test do sleep(1) at WAL tests setup. Though DAOS-13043 fix also solves DAOS-13040 issue, this fix only creates some randomization but does not guarantee uniqueness. The cost of adding the sleep to WAL vos_tests is about 15% more run time which is reasonable. The benefit is that the test remains robust and it would have no false positives when WAL blob created on AIO device.

Required-githooks: true

### Gatekeeper:


NiuYawei (commented):
The PR-11828 is already landed, and the unit tests will be properly executed on every PR since now on, and if we don't hit the issue anymore, we could regard the unit test is robust enough, otherwise, we should consider to add back the 'sleep 1' for AIO device.

What do you think?

src/vos/tests/vts_wal.c
```diff
@@ -526,7 +526,6 @@ wal_kv_large(void **state)
 static void
 wal_args_reset(struct io_test_args *args)
 {
-	sleep(1); /** See DAOS-13040 */
```
NiuYawei, line 529: I thought I've removed these two sleeps in my PR, but seems the landed one is bit different with what I pushed.

NiuYawei (approved):

src/vos/tests/vts_wal.c
jolivier23: Sorry, if that's the case.   It didn't rebase automatically when I landed my patch so I downloaded yours, did a merge with feature/vos_on_blob and pushed it.  So perhaps git removed them for some reason

jolivier23 (approved):

PR #11848, opened 2023-03-31
DAOS-12628 control: Add non-debug event logging for daos_server

mjmac:
When a rank exits unexpectedly, a RAS event is raised and should
be logged at ERROR level. When a pool's service ranks are updated,
the event should be logged at INFO level.

Required-githooks: true

### DIFF
--- a/src/control/system/membership.go
+++ b/src/control/system/membership.go
@@ -547,18 +547,25 @@ func (m *Membership) handleEngineFailure(evt *events.RASEvent) {
 	//
 	// e.g. if member.Addr.IP.Equal(net.ResolveIPAddr(evt.Hostname))
 
-	ns := MemberStateErrored
-	if member.State.isTransitionIllegal(ns) {
-		m.log.Debugf("skipping %s", msgBadStateTransition(member, ns))
+	newState := MemberStateErrored
+	if member.State.isTransitionIllegal(newState) {
+		m.log.Debugf("skipping %s", msgBadStateTransition(member, newState))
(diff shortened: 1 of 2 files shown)


mjmac:
One unrelated failure (this PR only updates logging, no logic changes): [DAOS-13054](https://daosio.atlassian.net/browse/DAOS-13054)

kjacque (approved):

src/control/system/membership.go
```diff
@@ -547,18 +547,25 @@ func (m *Membership) handleEngineFailure(evt *events.RASEvent) {
 	//
 	// e.g. if member.Addr.IP.Equal(net.ResolveIPAddr(evt.Hostname))
 
-	ns := MemberStateErrored
-	if member.State.isTransitionIllegal(ns) {
-		m.log.Debugf("skipping %s", msgBadStateTransition(member, ns))
+	newState := MemberStateErrored
+	if member.State.isTransitionIllegal(newState) {
+		m.log.Debugf("skipping %s", msgBadStateTransition(member, newState))
```
knard-intel, line 552: For an unexpected state machine transition, I would expected to have log at error level ?

tanabarr (approved):

src/control/system/membership.go
knard-intel: OK understood. Thanks for the explanation.

knard-intel (approved):
LGTM thanks to @tanabarr explanation :)

PR #11875, opened 2023-04-05
DAOS-13094 control: Add enable_tls alias for allow_insecure

mjmac:
If the config specifies enable_tls: false, it is the equivalent
to setting allow_insecure: true.

Features: control

Required-githooks: true


mjmac:
Parent ticket was closed as "Won't Do".

PR #11881, opened 2023-04-06
DAOS-13625 dfuse: Merge the info and projection_info structs.

ashleypittman:
There was duplication here from when we could support multiple
remotes so the distinction was artificial and there would only
ever be one of each struct, so merge the two.

Keep the old struct name as an alias to reduce the number of
lines touched by this cahnge, and heep the _iet entry with
the old prefix.

Test-tag: dfuse

Required-githooks: true


jolivier23 (approved):

daltonbohning (approved):
Does this have any API compatibility concerns? For example: middleware / third-party compiled against daos, daos updated, middleware / third-party _**not**_ recompiled.

ashleypittman:
> Does this have any API compatibility concerns? For example: middleware / third-party compiled against daos, daos updated, middleware / third-party _**not**_ recompiled.

No, this is all internal to dfuse

PR #11893, opened 2023-04-07
DAOS-13114 control: Add batching support for PoolEvict requests

mjmac:
The PoolEvict gRPC handler may be invoked from N agents in the
event of a large-scale application crash. This would result in
N dRPC requests from the MS leader to its local engine in order
to evict the pool handles in those requests. With this commit,
a batching mechanism is added to improve performance in these
large-scale events so that N agent requests are consolidated
into a smaller number of dRPC requests containing multiple
handles.

Also improves agent logging of evicted handles due to application
crash by including the process name and logging a single line
at INFO level instead of DEBUG.

Features: pool control
Required-githooks: true

### DIFF
--- a/src/control/common/proc_utils.go
+++ b/src/control/common/proc_utils.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2022 Intel Corporation.
+// (C) Copyright 2022-2023 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -15,22 +15,34 @@ import (
 	"github.com/pkg/errors"
 )
 
-func getProcPids(procDir, procName string) (pids []int, _ error) {
+func readProcName(procPath string) (string, error) {
+	data, err := os.ReadFile(procPath)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to read %q", procPath)
+	}
+	if len(data) == 0 {
+		return "", errors.Wrapf(err, "%q was empty", procPath)
+	}
+	return filepath.Base(strings.Split(string(data), "\x00")[0]), nil
+}
+
+func getProcPids(procDir, searchName string) (pids []int, _ error) {
+	if searchName == "" {
+		return nil, nil
+	}
+
 	allProcs, err := filepath.Glob(filepath.Join(procDir, "*", "cmdline"))
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to read process list")
 	}
 
 	for _, proc := range allProcs {
-		data, err := os.ReadFile(proc)
+		procName, err := readProcName(proc)
 		if err != nil {
 			continue
 		}
-		if len(data) == 0 {
-			continue
-		}
-		comps := strings.Split(string(data), "\x00")
-		if !(filepath.Base(comps[0]) == procName) {
+
+		if procName != searchName {
 			continue
 		}
 
@@ -50,11 +62,12 @@ func GetProcPids(procName string) ([]int, error) {
 }
 
 func getProcName(pid int, procDir string) (string, error) {
-	exe, err := os.Readlink(filepath.Join(procDir, strconv.Itoa(pid), "exe"))
-	if err != nil {
-		return "", errors.Wrap(err, "failed to read executable path")
-	}
-	return filepath.Base(exe), nil
+	return readProcName(filepath.Join(procDir, strconv.Itoa(pid), "cmdline"))
+}
+
+// GetProcName returns the name of the process with the given pid.
+func GetProcName(pid int) (string, error) {
+	return getProcName(pid, "/proc")
 }
 
 func checkDupeProcess(pid int, procDir string) error {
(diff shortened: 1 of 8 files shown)


mjmac:
One unrelated failure: [DAOS-12899](https://daosio.atlassian.net/browse/DAOS-12899)

mjmac:
Follow-on logging improvements in #11914

kjacque (approved):

src/control/server/mgmt_svc.go
```diff
@@ -96,3 +133,136 @@ func (svc *mgmtSvc) checkReplicaRequest(req proto.Message) error {
...
+		msg, ok := req.msg.(*mgmtpb.PoolEvictReq)
+		if !ok {
+			svc.log.Errorf("unexpected message type %T", req.msg)
+			continue
+		}
+		poolReqs[msg.Id] = append(poolReqs[msg.Id], req)
+	}
+
+	for poolID, reqs := range poolReqs {
+		if len(reqs) == 0 {
```
tanabarr, line 176: this can't happen, as key won't exist unless a valid request has been inserted into the slice

src/control/server/mgmt_svc.go
```diff
@@ -96,3 +133,136 @@ func (svc *mgmtSvc) checkReplicaRequest(req proto.Message) error {
...
+		case bprChan <- bpr:
+		}
+		return
+	}
+
+	switch msgName {
+	case string(proto.MessageName(new(mgmtpb.PoolEvictReq))):
+		bpr.retryableReqs = svc.processBatchPoolEvictions(ctx, bprChan, reqs)
+	default:
+		svc.log.Errorf("no batch handler for message type %s", msgName)
```
tanabarr, line 217: is the reason we are allowing for this related to compatibility reasons with different software versions that might support different message types?

src/control/server/mgmt_svc.go
```diff
@@ -96,3 +133,136 @@ func (svc *mgmtSvc) checkReplicaRequest(req proto.Message) error {
...
+			bprChan := make(batchProcessRespChan, batchedMsgNr)
+			for msgName, reqs := range batchedMsgReqs {
+				svc.log.Debugf("processing %d %s requests", len(reqs), msgName)
+				go svc.processBatchedMsgRequests(parent, bprChan, msgName, reqs)
+			}
+
+			for i := 0; i < batchedMsgNr; i++ {
+				bpr := <-bprChan
+				if len(bpr.retryableReqs) > 0 {
+					batchedMsgReqs[bpr.msgName] = bpr.retryableReqs
```
tanabarr, line 261: so here the reqs are reset rather than appended for a given message name key because receiving the process response indicates that all requests for that key have been handled so append is not necessary. do I have that right?

tanabarr (approved):
nice solution!

src/control/server/mgmt_svc.go
```diff
@@ -23,6 +25,39 @@ import (
 	"github.com/daos-stack/daos/src/control/system/raft"
 )
 
+type (
+	batchRequest struct {
+		msg    proto.Message
+		ctx    context.Context
```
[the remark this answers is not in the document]
mjmac: As noted in the link, it's guidance, not a rule. And the reasons for it make sense -- passing a context as an argument encourages good design that allows for easy context scoping and reasoning about a system. In this case, the reason we are storing a context is for one purpose: To avoid a hang in the batch handler that is trying to send a response to a gRPC request that has been canceled ([code](https://github.com/daos-stack/daos/pull/11893/files#diff-c7ceef5c2c3cbaa758519a519c7c7cd76a3209b7d466b296a95db792749df421R55)).

We can't pass the gRPC request's context as a standalone argument because we're sending the request via a channel to the batch processing loop, so the only option we have is to add it to the `batchRequest` type, which can be thought of as a structured argument. Doing it that way is in keeping with the spirit of the guidance, in that we're not really storing the context in a long-lived struct, we're just using the struct to convey the context over a channel with some other related information.

src/control/server/mgmt_svc.go
mjmac: Yeah, would be better to put this up at the top of the method instead. I think I added it here before refactoring to add the coalescing bit.

src/control/server/mgmt_svc.go
mjmac: More that the other option is a `panic()`, and that seems overly harsh. I suppose the argument could be made that a panic like that could only happen in development, though, so maybe it's fine.

src/control/server/mgmt_svc.go
mjmac: The requests in the `bpr.retryableReqs` slice are requests that weren't successfully processed but also didn't result in a clear failure, so they can be tried again on the next turn of the batch processing loop. In most cases this won't happen, but I wanted to allow the design to accommodate that situation.

src/control/common/proc_utils.go
```diff
@@ -15,22 +15,34 @@ import (
...
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to read %q", procPath)
+	}
+	if len(data) == 0 {
+		return "", errors.Wrapf(err, "%q was empty", procPath)
+	}
+	return filepath.Base(strings.Split(string(data), "\x00")[0]), nil
+}
+
+func getProcPids(procDir, searchName string) (pids []int, _ error) {
```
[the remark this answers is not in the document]
mjmac: Hmm, thanks. I'll add a testcase for that in a follow-up.

PR #11898, opened 2023-04-10
DAOS-13121 container: read not set properties for IV

wangshilong:
Currently IV update need all properties be set, don't ignore not set properties(for upgrading case).

Required-githooks: true


kccain (approved):
The ticket mentions a specific "daos cont list-objects" command that fails when updating the daos software from 2.2 to master. Does this issue affect many container operations that require opening the container? e.g., container query? Or is it just the list-objects command for which the issue was seen?

bfaccini (approved):

jolivier23 (approved):

PR #11913, opened 2023-04-11
DAOS-13399 gurt: Memset the buffer on free.

ashleypittman:
Try and protect against use-after-free errors by poisoning allocations on free.

Use the D_FREE() hook to memset data before freeing it back to libc.
Use compiler hints on the daos allocation macros so that the compiler knows
the size of buffers being allocated.

This assists with FORTIFY_SOURCE as well by giving the compiler better
knowledge of behaviour.

Allocations are only memset for non-release builds for now.


src/include/gurt/common.h
```diff
@@ -269,6 +271,8 @@ char *d_realpath(const char *path, char *resolved_path);
 #define D_FREE(ptr)                                                                                \
 	do {                                                                                       \
 		if ((ptr) != NULL) {                                                               \
+			size_t _frs = malloc_usable_size(ptr);                                     \
+			memset(ptr, 0x42, _frs);                                                   \
```
jolivier23, line 275: Should we put this into a DAOS_BUILD_RELEASE ifndef ?

ashleypittman: Yeah, I don't think we should land this I just wanted to see if it would work - which it all does apart from one test.

src/tests/suite/daos_degraded.c
```diff
@@ -1,16 +1,14 @@
...
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
 /**
  * This file is part of daos
  *
- * tests/suite/daos_replicated.c
- *
  * Replication tests need external interaction, to
```
frostedcmos, line 9: Degraded tests?

frostedcmos (approved):

src/tests/suite/daos_degraded.c
ashleypittman: Could be.  I removed the filename because it was obviously wrong but don't want to get into correcting a comment.

src/gurt/misc.c
```diff
@@ -81,6 +81,25 @@ d_asprintf(char **strp, const char *fmt, ...)
 	return rc;
 }
 
+char *
+d_asprintf2(int *_rc, const char *fmt, ...)
```
daltonbohning, line 85: Should the old `d_asprintf` be removed?

src/tests/suite/daos_obj.c
```diff
@@ -3748,8 +3748,8 @@ split_sgl_internal(void **state, int size)
 	rc = daos_obj_open(arg->coh, oid, DAOS_OO_RW, &oh, NULL);
 	assert_rc_equal(rc, 0);
 
-	sbuf1 = calloc(size/2, 1);
-	sbuf2 = calloc(size/2, 1);
+	D_ALLOC(sbuf1, size / 2);
+	D_ALLOC(sbuf2, size / 2);
```
daltonbohning, line 3752: It's test code, but do these need assert_non_null?

src/include/gurt/common.h
```diff
@@ -285,6 +306,44 @@ char *d_realpath(const char *path, char *resolved_path);
 		}                                                                                  \
 	} while (0)
 
+#else
+
+/* Developer/debug version, poison memory on free.
+ * This tries several ways to access the buffer size however none of them are perfect so for now
+ * this is no in release builds.
+ */
```
daltonbohning, line 314: Could this cause issues / weird behavior for devs?

src/gurt/misc.c
ashleypittman: I wasn't sure if there was any middleware using it.

src/tests/suite/daos_obj.c
ashleypittman: In theory, but it'll be easy to tell from any stacktrace if they fail.  The only reason for the change here is the compiler is now warning on mismatched alloc/free calls.

src/gurt/misc.c
daltonbohning: Right, that's public. Then we shouldn't remove. If anything, maybe the old should be a wrapper for the new? But maybe overkill

daltonbohning (approved):

PR #11914, opened 2023-04-11
DAOS-12002 control: General logging cleanup

mjmac:
Improve density and relevance of agent logging for client
process monitoring. Important messages should be logged
at INFO or ERROR, and DEBUG messages should avoid excess
verbosity.

Standardizes a number of debug messages to make grepping
easier and improves log density/relevance. Adds client
process info to attachInfo logs. Removes some unused symbols.

Adds a new TRACE log level for highly-verbose output that
does not normally need to be displayed at DEBUG level.

Fixes log configuration for daos_server, daos_agent, dmg,
and daos to properly handle logging NOTICE/TRACE to
logfiles. If the --debug flag is used, set loglevel to
TRACE, as the intention in this case is usually to get
maximum logging.

### DIFF
--- a/src/control/cmd/daos/main.go
+++ b/src/control/cmd/daos/main.go
@@ -150,7 +150,7 @@ or query/manage an object inside a container.`
 		}
 
 		if opts.Debug {
-			log.WithLogLevel(logging.LogLevelDebug)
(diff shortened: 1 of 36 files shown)


daltonbohning (commented):
ftest LGTM

PR #11936, opened 2023-04-13
DAOS-13088 control: Fix provider query with multiprovider

kjacque:
Some additional work was necessary to make specified provider
query work with multiprovider.


kjacque:
> Looks fine to me. Is this at all related to the crash reported in [DAOS-13148](https://daosio.atlassian.net/browse/DAOS-13148)?

This does look like the same issue. I've created PR #11941 on master as well.

tanabarr (approved):

PR #11949, opened 2023-04-13
DAOS-13164 API: to add parameter flag to handle O_NOFOLLOW in dfs_access

wiliamhuang:
faccessat() in libc has a flag to handle AT_SYMLINK_NOFOLLOW. We need a similar parameter for DFS too.
Also replace EPERM with EACCESS in check_access() to be consistent with access() in glibc.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfs/dfs.c
+++ b/src/client/dfs/dfs.c
@@ -4880,7 +4880,7 @@ out:
 }
 
 int
-dfs_access(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask)
+dfs_access(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask, int flag)
 {
 	daos_handle_t		oh;
 	bool			exists;
@@ -4922,7 +4922,7 @@ dfs_access(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask)
 	if (!exists)
 		return ENOENT;
 
-	if (!S_ISLNK(entry.mode)) {
+	if (!S_ISLNK(entry.mode) || (flag & O_NOFOLLOW)) {
 		if (mask == F_OK)
 			return 0;
 
--- a/src/client/dfs/dfs.c
+++ b/src/client/dfs/dfs.c
@@ -1152,12 +1152,8 @@ check_access(uid_t c_uid, gid_t c_gid, uid_t uid, gid_t gid, mode_t mode, int ma
 {
 	mode_t	base_mask;
 
-	/** Root can access everything */
-	if (uid == 0)
-		return 0;
-
 	if (mode == 0)
-		return EPERM;
+		return EACCES;
 
 	/** set base_mask to others at first step */
 	base_mask = S_IRWXO;
@@ -1174,17 +1170,17 @@ check_access(uid_t c_uid, gid_t c_gid, uid_t uid, gid_t gid, mode_t mode, int ma
 	/** Execute check */
 	if (X_OK == (mask & X_OK))
 		if (0 == (mode & (S_IXUSR | S_IXGRP | S_IXOTH)))
-			return EPERM;
+			return EACCES;
 
 	/** Write check */
 	if (W_OK == (mask & W_OK))
 		if (0 == (mode & (S_IWUSR | S_IWGRP | S_IWOTH)))
-			return EPERM;
+			return EACCES;
 
 	/** Read check */
 	if (R_OK == (mask & R_OK))
 		if (0 == (mode & (S_IRUSR | S_IRGRP | S_IROTH)))
-			return EPERM;
+			return EACCES;
 
 	/** TODO - check ACL, attributes (immutable, append) etc. */
 	return 0;
@@ -4879,8 +4875,8 @@ out:
 	return rc;
 }
 
-int
-dfs_access_wflag(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask, int flag)
+static int
+access_int(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask, int flag)
 {
 	daos_handle_t		oh;
 	bool			exists;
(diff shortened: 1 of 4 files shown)


mchaarawi:
the other option would be to add a new API:
dfs_access2() or dfs_access_wflag() or something like that if you need that.

wiliamhuang:
> that's a no no :-) can't just change a public API like that. can you use something like lookup/release instead for your use case?

ok. I will use workaround then in my case. Thank you! I made a similar change for dfs_chmod() in another PR too. I need to revert it back too.

mchaarawi:
> > that's a no no :-) can't just change a public API like that. can you use something like lookup/release instead for your use case?
> 
> ok. I will use workaround then in my case. Thank you! I made a similar change for dfs_chmod() in another PR too. I need to revert it back too.

or you could name it as the libc routine which makes more sense:
dfs_accessat
dfs_chmodat
which would be good to have i guess.
please add a unit test too if you make that change.

wiliamhuang:
> or you could name it as the libc routine which makes more sense: dfs_accessat dfs_chmodat which would be good to have i guess. please add a unit test too if you make that change.

Sounds good. Thank you!

src/include/daos_fs.h
```diff
@@ -1021,6 +1021,24 @@ dfs_osetattr(dfs_t *dfs, dfs_obj_t *obj, struct stat *stbuf, int flags);
...
+ *			which means operation will be on root object.
+ * \param[in]	mask	accessibility check(s) to be performed.
+ *			It should be either the value F_OK, or a mask with
+ *			bitwise OR of one or more of R_OK, W_OK, and X_OK.
+ * \param[in]	flag	bitwise flag. Handles O_NOFOLLOW.
+ *
+ * \return		0 on success, errno code on failure.
+ */
+int
+dfs_access_wflag(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask, int flag);
```
mchaarawi, line 1040: i know i suggested this, but is this better or to reuse the libc version, like dfs_accessat ?

wiliamhuang: In glibc,
int faccessat(int dirfd, const char *pathname, int mode, int flags);

int dfs_access(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask)
DAOS does not adopt the way of using dirfd generally. If you do prefer dfs_accessat, I can change it.

mchaarawi: ok no problem. this is fine as it is

mchaarawi:
it sounds like the dfs tests are failing. probably better to run those locally and make sure they pass before pushing a PR and consuming CI time.

src/client/dfs/dfs.c
```diff
@@ -4958,7 +4958,13 @@ dfs_access(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask)
 }
 
 int
-dfs_chmod(dfs_t *dfs, dfs_obj_t *parent, const char *name, mode_t mode)
+dfs_access(dfs_t *dfs, dfs_obj_t *parent, const char *name, int mask)
+{
+	return dfs_access_wflag(dfs, parent, name, mask, 0);
```
mchaarawi, line 4963: generally it's not great for an API to call another API. in some cases it's unavoidable, but in this case, it is and it would be better to do something like:

access_int(.., int flags)

dfs_access() {access_int(..., 0);}
dfs_access_wflags() {access_int(..., flags);}

wiliamhuang:
> it sounds like the new dfs tests you added are failing? probably better to run those locally and make sure they pass before pushing a PR and consuming CI time.
Thank you! I did run unit tests locally previously but not hardware tests. I will debug the failed hardware tests locally before updating this PR.

src/client/dfs/dfs.c
wiliamhuang: Sounds good. I will fix it. Thank you!

mchaarawi (approved):

src/client/dfs/dfs.c
```diff
@@ -5082,13 +5090,25 @@ dfs_chmod(dfs_t *dfs, dfs_obj_t *parent, const char *name, mode_t mode)
...
+{
+	return chmod_int(dfs, parent, name, mode, 0);
+}
+
+int
+dfs_chmod_wflag(dfs_t *dfs, dfs_obj_t *parent, const char *name, mode_t mode, int flag)
+{
+	return chmod_int(dfs, parent, name, mode, flag);
+}
+
```
daltonbohning, line 5111: Not sure how this affects your work, but I want to point out that chmod on a symlink itself is not always supported.
1. For example, from https://man7.org/linux/man-pages/man2/chmod.2.html:
`AT_SYMLINK_NOFOLLOW` isn't even implemented for `chmod`.

2. And this discussion: https://unix.stackexchange.com/questions/293763/how-to-change-the-symlink-permission

3. When DFS creates a symlink, it sets 777: https://github.com/daos-stack/daos/blob/d439c2fb7fa0d4a2ac39cab1791d727942b775b0/src/client/dfs/dfs.c#L1473

4. And for datamover tools I ran into this. Hence: https://github.com/daos-stack/daos/blob/d439c2fb7fa0d4a2ac39cab1791d727942b775b0/src/client/dfs/dfs_sys.c#L667-L669

src/tests/suite/dfs_unit_test.c
```diff
@@ -499,6 +503,30 @@ dfs_test_syml_follow_hlpr(const char *name, mode_t mode)
...
+			&actual_mode, NULL);
+	assert_int_equal(rc, 0);
+	rc = dfs_release(obj);
+	assert_int_equal(rc, 0);
+	assert_int_equal(actual_mode, mode_new);
+
+	/* Check with dfs_access_wflag() */
+	rc = dfs_access_wflag(dfs_mt, NULL, name, W_OK, O_NOFOLLOW);
+	print_message("dfs_test_syml_follow_access_wflag(\"%s\") = %d\n", name, rc);
+	assert_int_not_equal(rc, 0);
```
daltonbohning, line 529: For sanity, should we run the same, but DO follow?
```
	rc = dfs_access_wflag(dfs_mt, NULL, name, W_OK, 0);
	assert_int_equal(rc, 0);
```

src/client/dfs/dfs.c
wiliamhuang: @daltonbohning You are right. "AT_SYMLINK_NOFOLLOW" was not implemented in glibc under most linux. This is OS dependent. 

@mchaarawi Since only a few systems have implemented fchmodat() and faccessat(). Rocky 8 does not support them. Shall we simply return
EOPNOTSUPP ("Operation not supported") in the interception library? If so, it may be not necessary to add dfs_chmod_wflag() and dfs_access_wflag(). Do I need to check other OS we commonly use for test?

PR #11825, opened 2023-03-29
DAOS-13006 test: resolve/silence all pylint warnings

daltonbohning:
Test-tag: pr daily_regression SameKeyDifferentValue TestWithTelemetryIOLatency ManagementServiceResilience ManagementServiceFailover soak_smoke HarnessCoreFilesTest test_container_boundary test_pool_management_race test_metadata_fillup test_telemetry_list test_container_telemetry TestWithTelemetryNvme TestHarnessSkipsBase test_scrubber_csum_fault test_scrubber_ssd_auto_eviction test_scrubber_disabled_during_pool_creation test_scrubber_enabled_during_pool_creation test_scrubber_ssd_auto_eviction test_list_pools test_used_imbalance test_fields_basic test_dfuse_find test_get_acl_valid DaosAggregationIOSmall IoSysAdmin container_acl DaosPerfLarge DaosPerf CartRpcOneNodeSwimNotificationOnRankEvictionTest EcodRunIoConf DaosRunIoConf RbldRunIoConf daos_racer DaosServerDumpTest,-manual OverwriteContainerACLTest UpdateContainerACLTest cart TestHarnessSkipsSkipped TestHarnessSkipsRun ParallelIo osa test_snapshot_negativecases test_snapshots test_single_object_validation container test_pool_management_race test_ioaggregation DaosCoreTestNvme DmvrPosixLargeFile TargetFailure EcodCellSizeProperty EcodCellSize EcodDisabledRebuildSingle IorHard OSADmgNegativeTest OSAOnlineDrain PoolAutotestTest
Allow-unstable-test: true

Resolve or silence all remaining pylint issues.

Required-githooks: true

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate watchers.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
[...]

PR #11952, opened 2023-04-13
DAOS-10930 pydaos: Adding option to pass the object class for dictionary as an argument.

ravalsam:
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/docs/user/python.md
+++ b/docs/user/python.md
@@ -75,6 +75,10 @@ stadium
 ```
 
 User can pass the predefined object class id during dict() method call.
+This is optional and it can be RP or EC or S and has to satisfy the rf property of container.
+By default, it will be OC_UNKNOWN (0) object class to `daos_obj_generate_oid()`
+It will allow DAOS to automatically select an object class based on the container properties.
+
 
 ```
 >>> dd_oid = dcont.dict("stadium_2024", {"France" : "Stade de France"}, "OC_RP_2G1")
--- a/docs/user/python.md
+++ b/docs/user/python.md
@@ -74,6 +74,17 @@ stadium
 2
 ```
 
+User can pass the predefined object class id during dict() method call.
+
+```
+>>> dd_oid = dcont.dict("stadium_2024", {"France" : "Stade de France"}, "OC_RP_2G1")
+>>> print(dd_oid)
+stadium_2024
+>>> print(len(dd_oid))
+1
+>>>
+```
+
 This creates a new persistent object named "stadium" and initializes it with two
 key-value pairs.
 
--- a/src/client/pydaos/__init__.py
+++ b/src/client/pydaos/__init__.py
@@ -11,7 +11,7 @@ import atexit
 from . import pydaos_shim
 # pylint: enable=relative-beyond-top-level
 
-DAOS_MAGIC = 0x7A89
+DAOS_MAGIC = 0x7A8A
 
 
 # Define the PyDError class here before doing the pydaos_core import so that
--- a/src/client/pydaos/__init__.py
+++ b/src/client/pydaos/__init__.py
@@ -1,15 +1,14 @@
-# (C) Copyright 2019-2022 Intel Corporation.
+# (C) Copyright 2019-2023 Intel Corporation.
 #
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 #
+# pylint: disable=consider-using-f-string
 """
 PyDAOS Module allowing global access to the DAOS containers and objects.
 """
 
 import atexit
-# pylint: disable=relative-beyond-top-level
-from . import pydaos_shim
-# pylint: enable=relative-beyond-top-level
+from . import pydaos_shim  # pylint: disable=relative-beyond-top-level,import-self
 
 DAOS_MAGIC = 0x7A8A
 
@@ -23,7 +22,7 @@ class PyDError(Exception):
     # possible.  There is an odd effect with daos_init() errors that
     # pydaos_shim is valid during __init__ but None during __str__ so format
     # the string early and just report it later on.
-    def __init__(self, message, rc):
+    def __init__(self, message, rc):  # pylint: disable=super-init-not-called
         err = pydaos_shim.err_to_str(DAOS_MAGIC, rc)
         if err:
             self.message = '{}: {}'.format(message, err)
@@ -36,6 +35,7 @@ class PyDError(Exception):
 
 class DaosClient():
     # pylint: disable=too-few-public-methods
+    # pylint: disable=attribute-defined-outside-init
     """
     DaosClient is responsible for handling DAOS init/fini.
 
--- a/src/client/pydaos/pydaos_core.py
+++ b/src/client/pydaos/pydaos_core.py
@@ -11,6 +11,7 @@ import enum
 # pylint: disable=relative-beyond-top-level
 from . import pydaos_shim
 # pylint: enable=relative-beyond-top-level
+# pylint: disable=invalid-name
 
 from . import DAOS_MAGIC
 from . import PyDError
@@ -62,7 +63,6 @@ class DCont():
     array(name, kwargs):
         Create new DArray object.
     """
-    # pylint: disable=invalid-name
 
     def __init__(self, pool=None, cont=None, path=None):
         self._dc = DaosClient()
--- a/src/client/pydaos/pydaos_core.py
+++ b/src/client/pydaos/pydaos_core.py
@@ -22,6 +22,7 @@ ObjClassID = enum.Enum(
     {key: value for key, value in list(pydaos_shim.__dict__.items())
      if key.startswith("OC_")})
 
+
 def _get_object_id(cid):
     """ Get the existing DAOS object class ID based on name.  """
 
@@ -29,6 +30,7 @@ def _get_object_id(cid):
     if cid == "0":
         return 0
     return ObjClassID[cid].value
+
 
 class DObjNotFound(Exception):
     """Raised by get if name associated with DAOS object not found """
--- a/src/client/pydaos/pydaos_core.py
+++ b/src/client/pydaos/pydaos_core.py
@@ -62,6 +62,7 @@ class DCont():
     array(name, kwargs):
         Create new DArray object.
     """
+    # pylint: disable=invalid-name
 
     def __init__(self, pool=None, cont=None, path=None):
         self._dc = DaosClient()
@@ -113,8 +114,7 @@ class DCont():
         # Default to OC_UNKNOWN (0), which will automatically select an object class.
         if cid == "0":
             return 0
-        if cid != "0":
-            return ObjClassID[cid].value
+        return ObjClassID[cid].value
 
     def dict(self, name, v: dict = None, cid="0"):
         """ Create new DDict object """
--- a/src/client/pydaos/pydaos_core.py
+++ b/src/client/pydaos/pydaos_core.py
@@ -107,13 +107,24 @@ class DCont():
     def __getitem__(self, name):
         return self.get(name)
 
-    def dict(self, name, v: dict = None, cid="OC_SX"):
+    def _get_object_id(self, cid):
+        """ Get the existing DAOS object class ID based on name.  """
+
+        # Default to OC_UNKNOWN (0), which will automatically select an object class.
+        if cid == "0":
+            return 0
+        if cid != "0":
+            return ObjClassID[cid].value
+
+    def dict(self, name, v: dict = None, cid="0"):
         """ Create new DDict object """
+
+        # Get the existing class ID based on class name given. Default to 0
+        objId = self._get_object_id(cid)
 
         # Insert name into root kv and get back an object ID
         (ret, hi, lo) = pydaos_shim.cont_newobj(DAOS_MAGIC, self._hdl, name,
-                                                ObjClassID[cid].value,
-                                                pydaos_shim.PYDAOS_DICT)
+                                                objId, pydaos_shim.PYDAOS_DICT)
         if ret != pydaos_shim.DER_SUCCESS:
             raise PyDError("failed to create DAOS dict", ret)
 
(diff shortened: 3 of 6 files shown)


docs/user/python.md
```diff
@@ -74,6 +74,17 @@ stadium
 2
 ```
 
+User can pass the predefined object class id during dict() method call.
```
mchaarawi, line 77: it would be good to expand this and mention passing the oclass is optional; can be RP or EC or S; and has to satisfy the rf property on the container. Also one could create the container with RF setting indicating that the oclass used automatically would use replication.
- in your example it might encourage using EC for dictionaries, which we should not do ;-); so please use RP instead.

src/client/pydaos/pydaos_core.py
```diff
@@ -107,11 +107,12 @@ def get(self, name):
     def __getitem__(self, name):
         return self.get(name)
 
-    def dict(self, name, v: dict = None):
+    def dict(self, name, v: dict = None, cid="OC_SX"):
```
mchaarawi, line 110: why is this SX? is that default?
if yes, it should be 0

src/client/pydaos/pydaos_core.py
```diff
@@ -124,12 +125,13 @@ def dict(self, name, v: dict = None):
 
         return dd
 
-    def array(self, name, v: list = None):
+    def array(self, name, v: list = None, cid="OC_SX"):
```
mchaarawi, line 128: same comment as dict

mchaarawi (changes requested):
- please add some regression test for this (possibly update the pydaos test in NLT).
- to verify we probably need a get_info on object to query the object class and print for the user to make sure of the oclass being used. same as like daos fs get-attr for posix.

docs/user/python.md
ravalsam: I will extend the comments. Example has OC_RP_2G1 not EC, so it should be ok.

src/client/pydaos/pydaos_core.py
ravalsam: Will use 0 as default

src/client/pydaos/pydaos_core.py
```diff
@@ -107,12 +107,24 @@ def get(self, name):
...
 
-    def dict(self, name, v: dict = None):
+    def _get_object_id(self, cid):
+        """ Get the existing DAOS object class ID based on name.  """
+
+        # Default to OC_UNKNOWN (0), which will automatically select an object class.
+        if cid == "0":
+            return 0
+        if cid != "0":
+            return ObjClassID[cid].value
```
daltonbohning, line 117: This will simplify and resolve the pylint error
```suggestion
        if cid == "0":
            return 0
        return ObjClassID[cid].value
```

src/client/pydaos/pydaos_core.py
```diff
@@ -107,12 +107,24 @@ def get(self, name):
...
+
+    def dict(self, name, v: dict = None, cid="0"):
```
daltonbohning, line 119: This adds a pylint error
`invalid-name, Argument name "v" doesn't conform to snake_case naming style`
This one is easy to fix, but some others like `hi`, `lo`, etc. are more difficult. To appease the pylint GHA check, you can add this to the top of the file (under the module docstring)
```
# pylint: disable=invalid-name
```

ashleypittman: For new code please use valid names and only suppress warnings for methods/attributes that are part of the published API.

ravalsam: dict() is an existing code, I just moved the function below.

ashleypittman: For what it's worth https://github.com/daos-stack/daos/pull/11825 fixes the pylint issues in this file, it might be an idea to cherry-pick the changes to this file in that PR into this PR.

ravalsam: Done

src/client/pydaos/pydaos_core.py
```diff
@@ -22,6 +22,13 @@
     {key: value for key, value in list(pydaos_shim.__dict__.items())
      if key.startswith("OC_")})
 
+def _get_object_id(cid):
```
daltonbohning, line 25: flake wants 2 blank lines between top-level definitions
```suggestion

def _get_object_id(cid):
```

src/client/pydaos/pydaos_core.py
```diff
@@ -22,6 +22,13 @@
...
+    """ Get the existing DAOS object class ID based on name.  """
+
+    # Default to OC_UNKNOWN (0), which will automatically select an object class.
+    if cid == "0":
+        return 0
+    return ObjClassID[cid].value
```
daltonbohning, line 31:
```suggestion
    return ObjClassID[cid].value

```

utils/node_local_test.py
```diff
@@ -3988,6 +3988,76 @@ def test_pydaos_kv(server, conf):
...
+
+def test_pydaos_kv_obj_class(server, conf):
+    """Test the predefined object class works with KV"""
+    # pylint: disable=consider-using-with
+
+    pydaos_log_file = tempfile.NamedTemporaryFile(prefix='kv_objclass_pydaos_',
+                                                  suffix='.log',
+                                                  delete=False)
+
+    os.environ['D_LOG_FILE'] = pydaos_log_file.name
```
daltonbohning, line 4000: Pre-existing in other places in this file, so out of scope - question for @ashleypittman If we just need the name, and `delete=False`, then shouldn't this work?
```suggestion
    with tempfile.NamedTemporaryFile(prefix='kv_objclass_pydaos_',
                                     suffix='.log',
                                     delete=False) as tmp_file:
        os.environ['D_LOG_FILE'] = tmp_file.name
```

daltonbohning (changes requested):
Just the introduced flake errors

src/client/pydaos/pydaos_core.py
```diff
@@ -124,13 +136,16 @@ def dict(self, name, v: dict = None):
 
         return dd
 
-    def array(self, name, v: list = None):
+    def array(self, name, v: list = None, cid="0"):
         # pylint: disable=unused-argument
         """ Create new DArray object """
 
+        # Get the existing class ID based on class name given. Default to 0
+        objId = _get_object_id(cid)
```
mchaarawi, line 144: just to verify this works if cid == 0?
is objID in this case (0)?
could be something like:
if (cid != "0")
  objId = _get_object_id(cid)
else
  objId = 0

utils/node_local_test.py
```diff
@@ -3988,6 +3988,76 @@ def test_pydaos_kv(server, conf):
...
+    kv1 = container.dict('object1', {"Monday": "1"}, "OC_S2")
+    if len(kv1) != 1:
+        failed = True
+        print(f'Expected length of kv object is 1 but got {len(kv1)}')
+
+    # Write kv2 dictionary with OC_SX object type
+    kv2 = container.dict('object2', {"Monday": "1", "Tuesday": "2"}, "OC_SX")
+    if len(kv2) != 2:
+        failed = True
+        print(f'Expected length of kv object is 2 but got {len(kv2)}')
```
mchaarawi, line 4019: maybe add 1 more dictionary that does not set the oclass. expected oclass should be S1 in this case.

src/client/pydaos/pydaos_core.py
ravalsam: That check is in _get_object_id() function above.

utils/node_local_test.py
ravalsam: will do

mchaarawi (approved):

mchaarawi:
for future reference, you should add
Skip-func-test: true
since you are not modifying any core code or functional tests.

daltonbohning (dismissed):

mchaarawi (dismissed):

daltonbohning (approved):

src/client/pydaos/pydaos_shim.c
ashleypittman, line 1: If you're changing any of the entry points in this file then I think we should be changing the PY_SHIM_MAGIC_NUMBER value as well.

src/client/pydaos/raw/README.md
```diff
@@ -23,7 +23,6 @@ pool.connect(1 << 1)
 # query the pool
 pool_info = pool.pool_query()
 print("Pool has {0} storage targets".format(pool_info.pi_ntargets))
-print("Pool created with {0} permissions".format(pool_info.pi_mode))
```
ashleypittman, line 26: Is this change related?  pydaos.raw is not related to pydaos.

utils/node_local_test.py
ashleypittman: Yes, that would be preferable, and then the "disable=consider-using-with" pragma can be removed as well.

It's one fewer local, one fewer suppression and the code is clearer in it's intent.

utils/node_local_test.py
```diff
@@ -3988,6 +3988,77 @@ def test_pydaos_kv(server, conf):
...
+                                                  delete=False)
+
+    os.environ['D_LOG_FILE'] = pydaos_log_file.name
+    daos = import_daos(server, conf)
+
+    pool = server.get_test_pool_obj()
+
+    cont = create_cont(conf, pool, ctype="PYTHON")
+
+    container = daos.DCont(pool.uuid, cont.uuid)
```
ashleypittman, line 4007: I think pool.label or pool.id() would get you a label here, but I'm not sure if daos.DCont() accepts labels.
```suggestion
    container = daos.DCont(pool.label, cont.uuid)
```

utils/node_local_test.py
```diff
@@ -3988,6 +3988,77 @@ def test_pydaos_kv(server, conf):
...
+    pydaos_log_file = tempfile.NamedTemporaryFile(prefix='kv_objclass_pydaos_',
+                                                  suffix='.log',
+                                                  delete=False)
+
+    os.environ['D_LOG_FILE'] = pydaos_log_file.name
+    daos = import_daos(server, conf)
+
+    pool = server.get_test_pool_obj()
+
+    cont = create_cont(conf, pool, ctype="PYTHON")
```
ashleypittman, line 4005: It would be good to pass in a label here.

utils/node_local_test.py
```diff
@@ -3988,6 +3988,77 @@ def test_pydaos_kv(server, conf):
...
+
+    # Write kv2 dictionary without any object type,
+    # so in this case we have 4 targets so default object type should be S4
+    kv2 = container.dict('object2', {"Monday": "1", "Tuesday": "2"})
+    if len(kv2) != 2:
+        failed = True
+        print(f'Expected length of kv object is 2 but got {len(kv2)}')
+
+    # Run a command to list the objects
+    cmd = ['cont', 'list-objects', pool.uuid, cont.uuid]
```
ashleypittman, line 4023: This should definitely be using labels.

utils/node_local_test.py
```diff
@@ -3988,6 +3988,77 @@ def test_pydaos_kv(server, conf):
...
+    data = rc.json
+    assert data['status'] == 0, rc
+    assert data['error'] is None, rc
+    assert data['response'] is not None, rc
+
+    # Run a command to get the object layout
+    print('query the object layout')
+    actual_obj_layout = []
+    for obj in data['response']:
+        cmd = ['object', 'query', pool.uuid, cont.uuid, obj]
```
ashleypittman, line 4036: As should this.

utils/node_local_test.py
```diff
@@ -3988,6 +3988,77 @@ def test_pydaos_kv(server, conf):
...
+
+    # Verify the object has the correct layout used during kv dictionary creation.
+    expected_obj_layout = ['S2', 'S4']
+    for obj in expected_obj_layout:
+        if obj not in actual_obj_layout:
+            failed = True
+            print(f'Expected obj {obj} not found in all {actual_obj_layout}')
+
+    if failed:
+        print("Test step failed in test_pydaos_kv_obj_class()")
```
ashleypittman, line 4053: This will only print to stdout but not actually fail anything.  You need something like the below to fail NLT.

`conf.wf.add_test_case('pydaos object test', failure='test failed')`

ashleypittman (commented):
Given how much effort is going into removing the use of uuids from existing test cases I'd rather we added new ones using labels only.

utils/node_local_test.py
ravalsam: I will update.

utils/node_local_test.py
ravalsam: I will update

ashleypittman: Thinking about this more you want to create the test case anyway, rather than just on failure so you want something like:
`if failed:
   conf.wf.add_test_case('pydaos object test', failure='test failed')
else:
   conf.wf.add_test_case('pydaos object test')
`
and then you'll be able to tell from the test results that this code ran.

I note that the existing test_pydaos_kv() function does not do this, it long predates the junit reporting of results so it does warrant updating.  If you don't want to do it here that's fine, I can push a PR to bring that one up to spec once this PR is compete.

ravalsam: I have not updated all the review comments, so I can incorporate this as well. It's a good way to confirm. 
I used the dummy prints just to make sure it ran in Ci. 
I am not sure how this NLT results are populating in Jenkins, so anything gives the test confirmation will on console will be idle.

src/client/pydaos/__init__.py
ashleypittman, line 1: Are you missing some updates from this file?  I thought you'd pulled in the pylint suppression changes?

ravalsam: I fixed this at the last min and forgot to fix pylint issue. I will update the PR.

utils/node_local_test.py
```diff
@@ -3988,6 +3988,79 @@ def test_pydaos_kv(server, conf):
...
+    expected_obj_layout = ['S2', 'S4']
+    for obj in expected_obj_layout:
+        if obj not in actual_obj_layout:
+            failed = True
+            print(f'Expected obj {obj} not found in all {actual_obj_layout}')
+
+    if failed:
+        conf.wf.add_test_case('pydaos kv object test', failure='test failed')
+    else:
+        conf.wf.add_test_case('pydaos kv object test')
```
ashleypittman, line 4055: There's no stdout from this test but the result is registered at https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-11952/30/testReport/NLT/core/Unit_Tests___NLT_on_EL_8___pydaos_kv_object_test/

ashleypittman (approved):

mchaarawi (approved):

PR #11956, opened 2023-04-14
DAOS-12945 vos: add memory metrics

wangdi1:
Add memory metrics for DTX and VOS memory usage.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/cart/README.env
+++ b/src/cart/README.env
@@ -80,6 +80,11 @@ This file lists the environment variables used in CaRT.
    by default, and can be specified using DD_SUBSYS, for example:
    "DD_SUBSYS=RPC,BULK,CORPC,GRP,LM,HG,PMIX,ST,IV" or also "DD_SUBSYS=all".
 
+ . D_MEMORY_TRACK
+   User can enable memory track for daos engine by D_MEMORY_TRACK=1. With the
+   environment, all of allocations inside DAOS (by D_ALLOC) will be tracked, and
+   total allocated bytes per xstream can be shown through metrics.
+
  . CRT_TIMEOUT
    Set it as integer in the range of (0, 3600] to set the global timeout value
    of all RPCs (second). Without setting it or set it as any other value will
--- a/src/engine/init.c
+++ b/src/engine/init.c
@@ -678,7 +678,6 @@ server_init(int argc, char *argv[])
 		       DP_RC(rc));
 
 	metrics = &dss_engine_metrics;
-
 	/** Report timestamp when engine was started */
 	d_tm_record_timestamp(metrics->started_time);
 
--- a/src/engine/srv.c
+++ b/src/engine/srv.c
@@ -26,6 +26,7 @@
 #include <daos_srv/smd.h>
 #include <daos_srv/vos.h>
 #include <gurt/list.h>
+#include <gurt/telemetry_producer.h>
 #include "drpc_internal.h"
 #include "srv_internal.h"
 
@@ -353,6 +354,7 @@ wait_all_exited(struct dss_xstream *dx, struct dss_module_info *dmi)
 	D_DEBUG(DB_TRACE, "XS(%d) drained ULTs.\n", dx->dx_xs_id);
 }
 
+#define D_MEMORY_TRACK_ENV "D_MEMORY_TRACK"
 /*
  * The server handler ULT first sets CPU affinity, initialize the per-xstream
  * TLS, CRT(comm) context, NVMe context, creates the long-run ULTs (GC & NVMe
@@ -366,11 +368,17 @@ dss_srv_handler(void *arg)
 	struct dss_thread_local_storage	*dtc;
 	struct dss_module_info		*dmi;
 	int				 rc;
+	bool				 track_mem = false;
 	bool				 signal_caller = true;
 
 	rc = dss_xstream_set_affinity(dx);
 	if (rc)
 		goto signal;
+
+	d_getenv_bool(D_MEMORY_TRACK_ENV, &track_mem);
+	if (unlikely(track_mem))
+		d_set_alloc_track_cb(dss_mem_total_alloc_track, dss_mem_total_free_track,
+				     &dx->dx_mem_stats);
 
 	/* initialize xstream-local storage */
 	dtc = dss_tls_init(dx->dx_tag, dx->dx_xs_id, dx->dx_tgt_id);
@@ -643,6 +651,46 @@ dss_xstream_free(struct dss_xstream *dx)
 	D_FREE(dx);
 }
 
+static void
+dss_mem_stats_init(struct mem_stats *stats, int xs_id)
+{
+	int rc;
+
+	rc = d_tm_add_metric(&stats->ms_total_usage, D_TM_GAUGE,
+			     "Total memory usage", "byte", "mem/total_mem/xs_%u", xs_id);
+	if (rc)
+		D_WARN("Failed to create memory telemetry: "DF_RC"\n", DP_RC(rc));
+
+	rc = d_tm_add_metric(&stats->ms_mallinfo, D_TM_MEMINFO,
+			     "Total memory arena", "", "mem/meminfo/xs_%u", xs_id);
+	if (rc)
+		D_WARN("Failed to create memory telemetry: "DF_RC"\n", DP_RC(rc));
+	stats->ms_current = 0;
+}
+
+void
+dss_mem_total_alloc_track(void *arg, daos_size_t bytes)
+{
+	struct mem_stats *stats = arg;
+
+	D_ASSERT(arg != NULL);
+
+	d_tm_inc_gauge(stats->ms_total_usage, bytes);
+	/* Only retrieve mallocinfo every 10 allocation */
+	if ((stats->ms_current++ % 10) == 0)
+		d_tm_record_meminfo(stats->ms_mallinfo);
+}
+
+void
+dss_mem_total_free_track(void *arg, daos_size_t bytes)
+{
+	struct mem_stats *stats = arg;
+
+	D_ASSERT(arg != NULL);
+
+	d_tm_dec_gauge(stats->ms_total_usage, bytes);
+}
+
 /**
  * Start one xstream.
  *
@@ -734,6 +782,8 @@ dss_start_one_xstream(hwloc_cpuset_t cpus, int tag, int xs_id)
 		D_ERROR("create scheduler fails: "DF_RC"\n", DP_RC(rc));
 		D_GOTO(out_dx, rc);
 	}
+
+	dss_mem_stats_init(&dx->dx_mem_stats, xs_id);
 
 	/** start XS, ABT rank 0 is reserved for the primary xstream */
 	rc = ABT_xstream_create_with_rank(dx->dx_sched, xs_id + 1,
--- a/src/gurt/misc.c
+++ b/src/gurt/misc.c
@@ -11,6 +11,7 @@
 
 #include <stdarg.h>
 #include <math.h>
+#include <malloc.h>
 #include <string.h>
 #include <stdlib.h>
 #include <errno.h>
@@ -24,6 +25,20 @@
 
 /* state buffer for DAOS rand and srand calls, NOT thread safe */
 static struct drand48_data randBuffer = {0};
+
+d_alloc_track_cb_t d_alloc_track_cb;
+d_alloc_track_cb_t d_free_track_cb;
+static __thread void *track_arg;
+
+void
+d_set_alloc_track_cb(d_alloc_track_cb_t alloc_cb, d_alloc_track_cb_t free_cb, void *arg)
+{
+	d_alloc_track_cb = alloc_cb;
+	d_free_track_cb = free_cb;
+	track_arg = arg;
+
+	D_INFO("memory track is enabled for the engine.\n");
+}
 
 void
 d_srand(long int seedval)
@@ -49,6 +64,12 @@ d_rand()
 void
 d_free(void *ptr)
 {
+	if (unlikely(track_arg != NULL)) {
+		size_t size = malloc_usable_size(ptr);
+
+		d_free_track_cb(track_arg, size);
+	}
+
 	free(ptr);
 }
 
@@ -87,25 +108,62 @@ d_free(void *ptr)
 void *
 d_calloc(size_t count, size_t eltsize)
 {
-	return calloc(count, eltsize);
+	void *ptr;
+
+	ptr = calloc(count, eltsize);
+	if (unlikely(track_arg != NULL)) {
+		if (ptr != NULL)
+			d_alloc_track_cb(track_arg, malloc_usable_size(ptr));
+	}
+
+	return ptr;
 }
 
 void *
 d_malloc(size_t size)
 {
-	return malloc(size);
+	void *ptr;
+
+	ptr = malloc(size);
+	if (unlikely(track_arg != NULL)) {
+		if (ptr != NULL)
+			d_alloc_track_cb(track_arg, size);
+	}
+
+	return ptr;
 }
 
 void *
 d_realloc(void *ptr, size_t size)
 {
-	return realloc(ptr, size);
+	void *new_ptr;
+
+	if (unlikely(track_arg != NULL)) {
+		size_t old_size = malloc_usable_size(ptr);
+
+		new_ptr = realloc(ptr, size);
+		if (new_ptr != NULL) {
+			d_free_track_cb(track_arg, old_size);
+			d_alloc_track_cb(track_arg, size);
+		}
+	} else {
+		new_ptr = realloc(ptr, size);
+	}
+	return new_ptr;
 }
 
 char *
 d_strndup(const char *s, size_t n)
 {
-	return strndup(s, n);
+	char *ptr;
+
+	ptr = strndup(s, n);
+	if (unlikely(track_arg != NULL)) {
+		if (ptr != NULL)
+			d_alloc_track_cb(track_arg, malloc_usable_size(ptr));
+	}
+
+	return ptr;
 }
 
 int
@@ -117,6 +175,11 @@ d_asprintf(char **strp, const char *fmt, ...)
 	va_start(ap, fmt);
 	rc = vasprintf(strp, fmt, ap);
 	va_end(ap);
+
+	if (unlikely(track_arg != NULL)) {
+		if (rc > 0 && *strp != NULL)
+			d_alloc_track_cb(track_arg, (size_t)rc);
+	}
 
 	return rc;
 }
@@ -143,16 +206,31 @@ d_asprintf2(int *_rc, const char *fmt, ...)
 char *
 d_realpath(const char *path, char *resolved_path)
 {
-	return realpath(path, resolved_path);
+	char *ptr;
+
+	ptr = realpath(path, resolved_path);
+	if (unlikely(track_arg != NULL)) {
+		if (ptr != NULL)
+			d_alloc_track_cb(track_arg, malloc_usable_size(ptr));
+	}
+
+	return ptr;
 }
 
 void *
 d_aligned_alloc(size_t alignment, size_t size, bool zero)
 {
-	void *buf = aligned_alloc(alignment, size);
+	void *buf;
+
+	buf = aligned_alloc(alignment, size);
+	if (unlikely(track_arg != NULL)) {
+		if (buf != NULL)
+			d_alloc_track_cb(track_arg, size);
+	}
 
 	if (!zero || buf == NULL)
 		return buf;
+
 	memset(buf, 0, size);
 	return buf;
 }
--- a/src/gurt/telemetry.c
+++ b/src/gurt/telemetry.c
@@ -12,6 +12,7 @@
 #include <math.h>
 #include <float.h>
 #include <pthread.h>
+#include <malloc.h>
 #include <gurt/common.h>
 #include <gurt/list.h>
 #include <sys/shm.h>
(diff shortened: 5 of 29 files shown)


daltonbohning (commented):
ftest LGTM

jolivier23 (approved):

src/vos/lru_array.c
```diff
@@ -271,6 +279,12 @@ lrua_array_alloc(struct lru_array **arrayp, uint32_t nr_ent, uint32_t nr_arrays,
 	if (array == NULL)
 		return -DER_NOMEM;
 
+	if (flags & LRU_FLAG_TRACK_ALLOC) {
+		tls = vos_tls_get();
+		d_tm_inc_counter(tls->vtl_lru_alloc_size,
```
jolivier23, line 284: minor nit, we could potentially pass in the metric pointer rather than using a flag.  This would allow us to track more than one usage.  Currently, we use it for timestamp cache and DTX local id assignment.

src/vos/vos_common.c
```diff
@@ -479,6 +479,28 @@ vos_tls_init(int xs_id, int tgt_id)
...
+
+	rc = d_tm_add_metric(&tls->vtl_obj_cnt, D_TM_COUNTER,
+			     "Number of cached vos object", "entry",
+			     "mem/vos/vos_obj_%u/tgt_%u",
+			     sizeof(struct vos_object), tgt_id);
+	if (rc)
+		D_WARN("Failed to create vos obj cnt: "DF_RC"\n", DP_RC(rc));
+
+	rc = d_tm_add_metric(&tls->vtl_lru_alloc_size, D_TM_COUNTER,
+			     "Number of LRU cache size", "byte",
```
jolivier23, line 498: If you were to change the above, this could change to be more specific.  It would be more accurate to say something like "active DTX table LRU cache".

jolivier23 (dismissed):

src/vos/lru_array.c
wangdi1: sure thanks

src/vos/vos_common.c
wangdi1: sure

src/gurt/telemetry.c
```diff
@@ -1699,6 +1699,29 @@ d_tm_set_counter(struct d_tm_node_t *metric, uint64_t value)
 	d_tm_node_unlock(metric);
 }
 
+/**
+ * Increment the given counter by the specified \a value
```
Nasf-Fan, line 1703: increment/decrement/s

src/vos/vos_container.c
```diff
@@ -391,7 +391,7 @@ vos_cont_open(daos_handle_t poh, uuid_t co_uuid, daos_handle_t *coh)
 
 	rc = lrua_array_alloc(&cont->vc_dtx_array, DTX_ARRAY_LEN, DTX_ARRAY_NR,
 			      sizeof(struct vos_dtx_act_ent),
-			      LRU_FLAG_REUSE_UNIQUE,
+			      LRU_FLAG_REUSE_UNIQUE | LRU_FLAG_TRACK_ALLOC,
```
Nasf-Fan, line 394: We also need this flag in another lrua_array_alloc() instance in vos_dtx_cache_reset().

src/vos/vos_dtx.c
```diff
@@ -820,6 +821,7 @@ vos_dtx_commit_one(struct vos_container *cont, struct dtx_id *dti, daos_epoch_t
 	if (dce == NULL)
 		D_GOTO(out, rc = -DER_NOMEM);
 
+	d_tm_inc_counter(tls->vtl_dtx_cmt_ent_cnt, 1);
```
Nasf-Fan, line 824: It should be moved after L843.

src/tests/ftest/util/telemetry_utils.py
```diff
@@ -511,6 +511,12 @@ class TelemetryUtils():
         ENGINE_NVME_RELIABILITY_METRICS +\
         ENGINE_NVME_CRIT_WARN_METRICS +\
         ENGINE_NVME_INTEL_VENDOR_METRICS
+    ENGINE_MEM_USAGE_METRICS = [
+        "engine_mem_vos_dtx_cmt_ent_48",
+        "engine_mem_vos_vos_obj_360",
+        "engine_mem_vos_vos_lru_size",
+        "engine_mem_dtx_dtx_leader_handle_336",
+        "engine_mem_dtx_dtx_entry_40"]
```
Nasf-Fan, line 519: What are the postfix digits used for?

Nasf-Fan (changes requested):

src/vos/vos_dtx.c
wangdi1: ah, sure

src/gurt/telemetry.c
wangdi1: ah, right

src/vos/vos_container.c
wangdi1: ah, yes, thanks

wangdi1:
@jolivier23 @Nasf-Fan  @liuxuezhao please review. thanks

src/gurt/telemetry.c
```diff
@@ -1699,6 +1699,29 @@ d_tm_set_counter(struct d_tm_node_t *metric, uint64_t value)
...
+{
+	if (unlikely(metric == NULL))
+		return;
+
+	if (unlikely(metric->dtn_type != D_TM_COUNTER)) {
+		D_ERROR("Failed to set counter [%s] on item not a "
+			"counter.\n", metric->dtn_name);
+		return;
+	}
+
```
liuxuezhao, line 1719: may D_ASSERT(metric->dtn_metric->dtm_data.value >= value).

src/vos/lru_array.c
```diff
@@ -323,13 +350,14 @@ lrua_array_free(struct lru_array *array)
...
 		return;
 
-
 	for (i = 0; i < array->la_array_nr; i++) {
 		sub = &array->la_sub[i];
 		if (sub->ls_table != NULL)
 			array_free_one(array, sub);
 	}
 
+	alloc_cb(array, sizeof(*array) + sizeof(array->la_sub[0]) * array->la_array_nr);
```
liuxuezhao, line 359: why call alloc_cb for _free()?

src/gurt/telemetry.c
wangdi1: sure.

src/vos/lru_array.c
wangdi1: ah, yes

wangdi1:
failure due to DAOS-12883

src/gurt/misc.c
```diff
@@ -35,31 +53,61 @@ d_rand()
...
 
 void *
 d_calloc(size_t count, size_t eltsize)
 {
-	return calloc(count, eltsize);
+	void *ptr;
+
+	ptr = calloc(count, eltsize);
+	if (ptr != NULL && d_alloc_track_cb != NULL && track_arg != NULL)
+		d_alloc_track_cb(track_arg, eltsize);
```
Nasf-Fan, line 71: The real allocated size may be greater than the required "eltsize" because of alignment. So the "eltsize" maybe not equal to the size you will get via malloc_usable_size() when free the ptr. That will confuse the metrics. I think that we need to always use malloc_usable_size() in spite of when alloc or free for metrics purpose. Similar for other cases.

wangdi1: Ah, right it should be be eltsize * count. 

Actually "The value returned by malloc_usable_size() may be greater than the requested size of the allocation because of alignment and minimum size constraints." 
mostly it does not need to be precisely, but malloc_usable_size() should be good enough anyway.  so we can know the about memory usage on the current system.

And also there is overhead for malloc_usable_size() I assume, so I have to avoid to use that too often as much as possible.  Though I do not have other choice for d_free().

Nasf-Fan (approved):

src/tests/ftest/util/telemetry_utils.py
jolivier23: If we are going to use the size like this, it may be better to have a macro

#define metric_from_struct(name) "engine_mem_" #name "_" stringify(sizeof(struct name))

Or some such so it's always right.

My preference would be to leave the size out and add _cnt or something at end.

src/engine/init.c
```diff
@@ -687,7 +687,7 @@ server_init(int argc, char *argv[])
 		       DP_RC(rc));
 
 	metrics = &dss_engine_metrics;
-
+	d_set_alloc_track_cb(dss_mem_total_alloc_track, dss_mem_total_free_track);
```
jolivier23, line 690: Do we have any notion of how expensive this is?

wangdi1:
pass all of tests. ping

Nasf-Fan (dismissed):

src/gurt/misc.c
```diff
@@ -11,12 +11,32 @@
...
 
+d_alloc_track_cb_t d_alloc_track_cb;
+d_alloc_track_cb_t d_free_track_cb;
+static __thread void *track_arg;
+
+void
+d_set_alloc_track_cb(d_alloc_track_cb_t alloc_cb, d_alloc_track_cb_t free_cb)
+{
+#ifndef DAOS_BUILD_RELEASE
+	d_alloc_track_cb = alloc_cb;
```
jolivier23, line 29: do we have numbers on how much overhead this adds?  A lot of our benchmarking is done on 'dev' build.  Maybe we should make this debug by default?  Unless it's cheap.

jolivier23 (commented):
Did you try the malloc hook solution?

src/gurt/misc.c
wangdi1: yeah, according to Dalton's test, this adds about 2-4% overhead.  So Johann requires to move this to dev build. 
Hmm,  the patch is needed to monitor the memory usage, and debug OOM issue IMHO.

>A lot of our benchmarking is done on 'dev' build. 

Hmm, we probably need be aware of that, if we really need some "good" number IMHO.

Nasf-Fan (dismissed):

src/gurt/misc.c
```diff
@@ -11,12 +11,36 @@
...
+d_alloc_track_cb_t d_alloc_track_cb;
+d_alloc_track_cb_t d_free_track_cb;
+static __thread void *track_arg;
+
+void
+d_set_alloc_track_cb(d_alloc_track_cb_t alloc_cb, d_alloc_track_cb_t free_cb)
+{
+	char	*env;
+
+	env = getenv(DAOS_MEM_TRACK_ENV);
```
jolivier23, line 30: We should add an entry to src/cart/README.env (perhaps it should be D_ for consistency with gurt).

We should probably also add a section to docs/debugging.md

jolivier23 (approved):

src/gurt/misc.c
ashleypittman: There's a d_getenv_bool() function that could be used here isn't there?

src/gurt/telemetry.c
```diff
@@ -1699,6 +1699,32 @@ d_tm_set_counter(struct d_tm_node_t *metric, uint64_t value)
...
 }
 
+/**
+ * Decrement the given counter by the specified \a value
+ *
+ * \param[in]	metric	Pointer to the metric
+ * \param[in]	value	Increments the counter by this \a value
+ */
+void
+d_tm_dec_counter(struct d_tm_node_t *metric, uint64_t value)
```
mjmac, line 1709: NB: This was not an accidental omission in the telemetry library. By convention, telemetry counter types are supposed to always increase. Downstream consumers are likely to be confused by a counter that decreases, as most consumers will interpret this as a counter rollover.

If you want to track a value that can fluctuate up or down, then a gauge is more appropriate.

mjmac (changes requested):
There seems to be a little confusion here about the difference between a counter and a gauge. We own the implementation here, so we can do whatever we like, but you should be aware that there are long-established conventions for how these types of metrics work. A counter should not decrement except in the special case where it has rolled over.

For a good overview, see the following: https://prometheus.io/docs/concepts/metric_types/

wangdi1:
> There seems to be a little confusion here about the difference between a counter and a gauge. We own the implementation here, so we can do whatever we like, but you should be aware that there are long-established conventions for how these types of metrics work. A counter should not decrement except in the special case where it has rolled over.
> 
> For a good overview, see the following: https://prometheus.io/docs/concepts/metric_types/

Sure, I will update the patch. Thanks

ashleypittman:
> There seems to be a little confusion here about the difference between a counter and a gauge. We own the implementation here, so we can do whatever we like, but you should be aware that there are long-established conventions for how these types of metrics work. A counter should not decrement except in the special case where it has rolled over.
> 
> For a good overview, see the following: https://prometheus.io/docs/concepts/metric_types/

I've also used systems for this kind of tracking where instead of tracking `memory currently allocated` you have a counter for `total memory allocated` and a second one for `total memory freed` which leads to two incrementing counters and allows you to do more things like see the churn rate of allocations.  It does depend on having the consumer setup to display the combined metric of one vs the other however - and it's more subject to wrapping.

src/cart/README.env
```diff
@@ -80,6 +80,11 @@ This file lists the environment variables used in CaRT.
    by default, and can be specified using DD_SUBSYS, for example:
    "DD_SUBSYS=RPC,BULK,CORPC,GRP,LM,HG,PMIX,ST,IV" or also "DD_SUBSYS=all".
 
+ . D_MEMORY_TRACK
+   User can enable memory track by D_MEMORY_TRACK=1. With the environment, all
+   of allocations inside DAOS (by D_ALLOC) will be tracked, and total allocated
+   bytes per xstream can be shown through metrics.
+
```
ashleypittman, line 87: This is server-only from looking at the code?  The implementation is certainly only in server code and I think metrics are server-only as well currently?

src/engine/srv.c
```diff
@@ -644,6 +647,37 @@ dss_xstream_free(struct dss_xstream *dx)
 	D_FREE(dx);
 }
 
+static void
+dss_mem_stats_init(struct mem_stats *stats, int xs_id)
+{
+	int rc;
+
+	rc = d_tm_add_metric(&stats->ms_total_usage, D_TM_STATS_GAUGE,
+			     "Total memory usage", "byte", "mem/total_mem/xs_%u", xs_id);
```
ashleypittman, line 656: Just the byte count, not the actual alloc count so there's no chance to monitor average size?

src/gurt/misc.c
```diff
@@ -11,12 +11,38 @@
...
+void
+d_set_alloc_track_cb(d_alloc_track_cb_t alloc_cb, d_alloc_track_cb_t free_cb)
+{
+	bool	track_mem = false;
+
+	d_getenv_bool(D_MEMORY_TRACK_ENV, &track_mem);
+	if (unlikely(track_mem)) {
+		d_alloc_track_cb = alloc_cb;
+		d_free_track_cb = free_cb;
+		D_INFO("memory track is enabled.\n");
```
ashleypittman, line 36: d_getenv_bool() should have logged already but it's OK to do it a second time.

src/gurt/misc.c
```diff
@@ -11,12 +11,38 @@
...
+#include <malloc.h>
 #include <string.h>
 #include <gurt/common.h>
 
 /* state buffer for DAOS rand and srand calls, NOT thread safe */
 static struct drand48_data randBuffer = {0};
 
+d_alloc_track_cb_t d_alloc_track_cb;
+d_alloc_track_cb_t d_free_track_cb;
+static __thread void *track_arg;
```
ashleypittman, line 23: I'm a bit confused by this and I wonder if there's a layering violation here?

There is a cb register function that works across the whole program, then each pthread may also provide a handle which would mean that posix thread invokes the callback?  As an interface that's fairly horrible to work with if we want to use it anywhere else in future - and dumping memory stats on exit would be a great thing for the client to do at some point.

src/gurt/misc.c
```diff
@@ -35,31 +61,72 @@ d_rand()
...
 d_calloc(size_t count, size_t eltsize)
 {
-	return calloc(count, eltsize);
+	void *ptr;
+
+	ptr = calloc(count, eltsize);
+	if (unlikely(d_alloc_track_cb && track_arg != NULL)) {
+		if (ptr != NULL)
+			d_alloc_track_cb(track_arg, eltsize);
+	}
```
ashleypittman, line 82: You could probably get away with only checking track_arg and not d_alloc_track_cb here as well - but you might have to update d_set_alloc_track_arg to check if the callbacks have been set.
```suggestion
	if (unlikely(d_alloc_track_cb && track_arg != NULL && prt != NULL))
		d_alloc_track_cb(track_arg, eltsize);
```

src/gurt/misc.c
```diff
@@ -35,31 +61,72 @@ d_rand()
...
 {
-	return realloc(ptr, size);
+	void *new_ptr;
+
+	if (unlikely(d_alloc_track_cb && track_arg != NULL)) {
+		size_t old_size = malloc_usable_size(ptr);
+
+		new_ptr = realloc(ptr, size);
+		if (new_ptr != NULL)
+			d_alloc_track_cb(track_arg, size - old_size);
```
ashleypittman, line 111: This can be a negative number, I assume d_tm_inc_gauge() is OK with that?

src/gurt/misc.c
```diff
@@ -35,31 +61,72 @@ d_rand()
...
 char *
 d_strndup(const char *s, size_t n)
 {
-	return strndup(s, n);
+	char *ptr;
+
+	ptr = strndup(s, n);
+	if (unlikely(d_alloc_track_cb && track_arg != NULL)) {
+		if (ptr != NULL)
+			d_alloc_track_cb(track_arg, n);
```
ashleypittman, line 126: n is not correct here.

src/cart/README.env
wangdi1: yes, server only, since metrics are currently only supported on the server side.

src/gurt/misc.c
wangdi1: If we want to use this on the client side, yes, the current per xstream structure needs to be re-worked.

This patch is only for monitoring the memory usage for the server engine.

src/gurt/misc.c
ashleypittman: That seems an oversight.

src/gurt/misc.c
ashleypittman: How does this compare with just calling mallinfo?

wangdi1: mallinfo will only show the total memory usage of the node, and the patch is for DAOS engine memory usage?

ashleypittman: mallinfo tells you the current process, I'm not sure if that's what you want.

wangdi1: yes, mallinfo will provide the memory information per process. Though the patch is for tracking the memory per xstream. Hmm those these mallinfo could be useful indeed. I probably can export them through system xstream, but should be done through a different patch.

wangdi1:
ping

wangdi1:
ping

Nasf-Fan (approved):

wangdi1:
ping reviewers.

src/gurt/misc.c
```diff
@@ -11,12 +11,41 @@
...
+	d_getenv_bool(D_MEMORY_TRACK_ENV, &track_mem);
+	if (unlikely(track_mem)) {
+		d_alloc_track_cb = alloc_cb;
+		d_free_track_cb = free_cb;
+		D_INFO("memory track is enabled for the engine.\n");
+	}
+}
+
+void
+d_set_alloc_track_arg(void *arg)
```
NiuYawei, line 41: Is there any reason to export two separated APIs for callback & callback arg registering? Why don't we just register both callback & callback arg by single register API?

Making caller to separately register a callback and a per-thread callback arg looks weird to me. As for this PR, I think we don't need to register callback arg, since we could get the per-thread 'mem_stats' in dss_mem_total_alloc_track(), right?

src/gurt/misc.c
```diff
@@ -35,31 +64,74 @@ d_rand()
...
 void *
 d_calloc(size_t count, size_t eltsize)
 {
-	return calloc(count, eltsize);
+	void *ptr;
+
+	ptr = calloc(count, eltsize);
+	if (unlikely(track_arg != NULL)) {
+		if (ptr != NULL)
+			d_alloc_track_cb(track_arg, malloc_usable_size(ptr));
```
NiuYawei, line 84: What if d_alloc_track_cb() isn't registered?

src/gurt/misc.c
```diff
@@ -11,12 +11,41 @@
...
+static __thread void *track_arg;
+
+#define D_MEMORY_TRACK_ENV "D_MEMORY_TRACK"
+
+void
+d_set_alloc_track_cb(d_alloc_track_cb_t alloc_cb, d_alloc_track_cb_t free_cb)
+{
+	bool	track_mem = false;
+
+	d_getenv_bool(D_MEMORY_TRACK_ENV, &track_mem);
```
NiuYawei, line 32: You can't read an engine specific env var in common gurt code. Please move this check to engine and also make sure all the related metrics are only created when this "D_MEMORY_TRAC" is enabled.

NiuYawei (changes requested):

src/gurt/misc.c
liuxuezhao: seems it assumes the track_arg is registered after track_cb, but agreed that seems better to not register track_arg separately if possible.

src/gurt/misc.c
```diff
@@ -35,31 +64,74 @@ d_rand()
...
 char *
 d_strndup(const char *s, size_t n)
 {
-	return strndup(s, n);
+	char *ptr;
+
+	ptr = strndup(s, n);
+	if (unlikely(track_arg != NULL)) {
+		if (ptr != NULL)
+			d_alloc_track_cb(track_arg, malloc_usable_size(ptr));
```
liuxuezhao, line 131: just curious that why here use "malloc_usable_size(ptr)", and other places dmalloc/d_realloc/d_asprintf use requested size?

src/gurt/misc.c
wangdi1: sure, it makes sense.

src/gurt/misc.c
wangdi1: sure

src/gurt/misc.c
wangdi1: because n is max size of s in strndup, i.e. it might allocate less size buffer (< n), since it use malloc_usable_size() to track free buffer size, so let's use malloc_usable_size for allocating size here.

wangdi1:
ping

Nasf-Fan (approved):

liuxuezhao (approved):

gnailzenh (approved):

PR #11971, opened 2023-04-15
DAOS-7983 control: Add daos system query command

kjacque:
This patch adds a mechanism for DAOS clients to collect basic information about the DAOS system that the daos_agent is facilitating their access to.

- Add daos_system_info_alloc() and daos_system_info_free() functions to the public DAOS API to collect and free the system information.
- Add a daos_test subtest for the API functions.
- Add a daos system query command to the DAOS tool.
- Add a functional test for the happy path for the daos tool command.
- Include system name in GetAttachInfo payload.

Features: daos_cmd

Required-githooks: true

### Gatekeeper:


src/tests/ftest/control/daos_system_query.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    def test_system_query(self):
+        """JIRA ID: DAOS-7983
+        Test Description: Test daos system query.
+
+        :avocado: tags=all,full_regression
+        :avocado: tags=vm
+        :avocado: tags=control,daos_cmd
+        :avocado: tags=daos_system_query,test_daos_system_query
```
daltonbohning, line 25: We've been adding the class name as a tag as well
```suggestion
        :avocado: tags=DaosSystemQuery,daos_system_query,test_daos_system_query
```

src/tests/ftest/control/daos_system_query.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        daos_cmd = DaosCommand(self.bin)
```
daltonbohning, line 27: We have a helper function for this
```suggestion
        daos_cmd = self.get_daos_command()
```

src/tests/ftest/control/daos_system_query.py
```diff
@@ -0,0 +1,47 @@
+"""
+  (C) Copyright 2023 Intel Corporation.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+"""
+from apricot import TestWithServers
+from daos_utils import DaosCommand
```
daltonbohning, line 7: Then this isn't needed
```suggestion
```

src/tests/ftest/control/daos_system_query.py
```diff
@@ -0,0 +1,47 @@
...
+        engines_per_host = self.params.get("engines_per_host", "/run/server_config/*")
+        exp_num_ranks = num_servers * engines_per_host
+
+        exp_sys_name = self.server_managers[0].get_config_value("name")
+        exp_provider = self.server_managers[0].get_config_value("provider")
+
+        query_output = daos_cmd.system_query()["response"]
+
+        if query_output["system_name"] != exp_sys_name:
+            self.fail("expected '{}', got '{}'".format(exp_sys_name, query_output["system_name"]))
```
daltonbohning, line 39: I'd recommend a more descriptive message like (similar for others). That way Triage knows what the failure is without digging through logs
```suggestion
            self.fail("expected system_name='{}', got '{}'".format(exp_sys_name, query_output["system_name"]))
```

src/tests/ftest/control/daos_system_query.yaml
```diff
@@ -0,0 +1,27 @@
+hosts:
+  test_servers: 1
```
daltonbohning, line 2: Technically, ftest runs `daos` on a client. That currently defaults to localhost, but eventually will not. So we should be explicit here
```suggestion
hosts:
  test_servers: 1
  test_clients: 1
```

src/tests/ftest/control/daos_system_query.yaml
```diff
@@ -0,0 +1,27 @@
...
+timeout: 80
+server_config:
+  name: daos_server
+  engines_per_host: 2
+  engines:
+    0:
+      targets: 8
```
daltonbohning, line 9: Unless this test specifically needs to specify `targets`, I'd recommend removing this. The default in ftest is 8
```suggestion
```

src/tests/ftest/control/daos_system_query.yaml
```diff
@@ -0,0 +1,27 @@
...
+      pinned_numa_node: 0
+      nr_xs_helpers: 1
+      fabric_iface_port: 31416
+      log_file: daos_server0.log
+      log_mask: DEBUG,MEM=ERR
+      env_vars:
+        - DD_MASK=mgmt
+      storage: auto
+    1:
+      targets: 8
```
daltonbohning, line 19:
```suggestion
```

daltonbohning (commented):
Just ftest aesthetics / new style. FWIW, I don't think people outside the test team would know the current style (existing code is non-conformant), so I'm happy to review ftest code early if it saves devs CI cycles / time.

src/tests/ftest/control/daos_system_query.yaml
daltonbohning: Actually, ignore this one. We currently always run `daos` commands on localhost

src/include/daos_mgmt.h
```diff
@@ -20,6 +20,29 @@ extern "C" {
...
+ *		-DER_AGENT_COMM	Unable to communicate with DAOS agent
+ *		-DER_NO_PERM	No access to agent communications socket
+ *		-DER_MISC	Unexpected error
+ */
+int
+daos_mgmt_sys_info_alloc(struct daos_sys_info **info);
+
+/**
+ * Free the system info structure.
+ *
```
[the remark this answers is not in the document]
kjacque: Fixed.

src/tests/ftest/control/daos_system_query.py
kjacque: Added, thanks.

src/tests/ftest/control/daos_system_query.py
kjacque: Fixed

src/tests/ftest/control/daos_system_query.py
kjacque: Removed

src/tests/ftest/control/daos_system_query.py
kjacque: Fixed for this and the other checks. This also triggered a pylint complaint about line length, so I fixed that, too.

src/tests/ftest/control/daos_system_query.yaml
kjacque: Removed.

src/tests/ftest/control/daos_system_query.yaml
kjacque: I removed the second engine. It caused issues with scm formatting for some reason, but a second engine isn't really necessary anyway.

src/control/cmd/daos/system.go
```diff
@@ -0,0 +1,76 @@
...
+		Provider: C.GoString(&cSysInfo.dsi_provider[0]),
+	}
+
+	for i := C.uint32_t(0); i < cSysInfo.dsi_nr_ms_ranks; i++ {
+		cRankURI := C.get_rank_uri_at_idx(cSysInfo.dsi_ms_ranks, i)
+		sysInfo.RankURIs = append(sysInfo.RankURIs, &rankURI{
+			Rank: int(cRankURI.dru_rank),
+			URI:  C.GoString(cRankURI.dru_uri),
+		})
+	}
```
mjmac, line 62: This is fine, but for the record you could make a slice to iterate over the C array using [unsafe.Slice](https://pkg.go.dev/unsafe#Slice). I think something like the following should work:
```go
for _, cRank := range unsafe.Slice(cSysInfo.dsi_ms_ranks, int(cSysInfo.dso_nr_ms_ranks)) {
```

tanabarr:
I'm starting by replacing all ftest yaml scm storage tiers with class: ram and scm_size: 4 or 16 as these seem to be the defaults for VM tests, I'm replacing them with storage: auto and system_ram_reserved: 1 . I can iterate on the others once I get that working.

src/control/cmd/daos/system.go
```diff
@@ -0,0 +1,76 @@
...
+	Query systemQueryCmd `command:"query" description:"query DAOS system via the daos_agent"`
+}
+
+type systemQueryCmd struct {
+	daosCmd
+}
+
+func (cmd *systemQueryCmd) Execute(_ []string) error {
+	var cSysInfo *C.struct_daos_sys_info
+	rc := C.daos_mgmt_sys_info_alloc(&cSysInfo)
```
mjmac, line 45: Hmm. It seems a little strange to me to be doing a `get_attach_info()` in a function called `daos_mgmt_sys_info_alloc()`. Based on convention, I would expect something named `*_alloc()` to just be responsible for setting up a complex structure for use elsewhere. I wouldn't expect that helper function to be performing network operations. Maybe this ties into a wider conversation that I've been having with @liw over in [DAOS-13151](https://daosio.atlassian.net/browse/DAOS-13151)... I think that we may need some way to cache the `dc_mgmt_sys_info` data for the lifetime of the process.

The other thing to consider about this work in light of that other conversation is that the design of libdaos was originally intended to accommodate connections to multiple systems, as I understand it. Over time, though, assumptions about only connecting to a single system have been creeping in, so I wonder if we need to either fix those assumptions or else just formally abandon support for connections to multiple systems.

src/control/cmd/daos/system.go
kjacque: Good catch, thanks. Since I'll be making changes anyway, I'll change this, too.

liw (dismissed):
Looks good to me. While I don't always agree with GitHub Actions, I do agree in this case that we should document at least `dsi_provider` and the fields including "ms" (what is "ms", or what is "provider", a user may ask) _in the DAOS API_. (For internal APIs I would not nit-picking like this.) If you agree to add such documentation, it can be done in a doc-only follow-on PR so that extra CI runs won't be required.

src/mgmt/cli_mgmt.c
```diff
@@ -160,6 +160,16 @@ fill_sys_info(Mgmt__GetAttachInfoResp *resp, struct dc_mgmt_sys_info *info)
 		return -DER_INVAL;
 	}
 
+	if (strnlen(resp->sys, sizeof(info->system_name)) > 0) {
```
knard-intel, line 163: If an invalid system name could be defined, it could be useful to test if the result of` strnlen()` is equal to` sizeof(info->system_name)` to detect it.

kjacque: Good idea, I'll do that.

tanabarr (dismissed):
LGTM notwithstanding other reviewers comments

src/mgmt/cli_mgmt.c
kjacque: Actually it looks like the `copy_str` macro is covering that case.

knard-intel: OK thanks for the info and sorry for the disturbance.

knard-intel (dismissed):
LGTM

src/tests/ftest/util/daos_utils.py
```diff
@@ -24,6 +24,20 @@ class DaosCommand(DaosCommandBase):
             r"Highest Aggregated Epoch:\s+(\d+)",
     }
 
+    def system_query(self):
+        """Query the DAOS system for client endpoint information.
+
+        Args:
+            None
+
```
daltonbohning, line 32: minor - Only if you need to push again. Not necessary to be explicit
```suggestion
```

daltonbohning (commented):
ftest LGTM. thanks!

kjacque:
I pulled out the refresh option after having a conversation with Mike. We came to the conclusion that this shouldn't be something a client can hit on demand, and the rate limiting idea could be confusing to users. We have some other ideas for how to refresh the agent cache, which I'm addressing in another PR.

knard-intel (approved):
LGTM

liw (dismissed):

src/control/cmd/daos/system.go
```diff
@@ -0,0 +1,71 @@
...
+import "C"
+
+import (
+	"unsafe"
+
+	"github.com/pkg/errors"
+)
+
+type rankURI struct {
+	Rank int    `json:"rank"`
```
mjmac, line 21: minor: We use `uint32` for ranks everywhere else in the codebase. This struct isn't exported outside of the daos tool, so it's not a big deal, but in general we should try to be consistent and also avoid type conversions unless necessary.

mjmac (approved):
LGTM

src/control/cmd/daos/system.go
kjacque: I'll address this since I need to restart the CI job anyway.

liw (approved):

knard-intel (approved):

kjacque:
The test failures are known issues:
- DaosCoreTest.test_daos_extend_simple: https://daosio.atlassian.net/browse/DAOS-13200
- IOR failure: https://daosio.atlassian.net/browse/DAOS-12725

PR #12014, opened 2023-04-21
DAOS-13187 client: interoperability between 2.4 client vs 2.2 server

wangshilong:
1. container operation should query server registered protocal version before register RPC.

2. if server could not handle RPC, client should not retry forever, let it quit immediately.

Required-githooks: true

### Gatekeeper:


wangshilong:
ping reviewers

src/pool/cli.c
```diff
@@ -30,132 +30,35 @@ struct rsvc_client_state {
...
-		}
 	} else if (dc_pool_proto_version == DAOS_POOL_VERSION) {
 		rc = daos_rpc_register(&pool_proto_fmt_v5, POOL_PROTO_CLI_COUNT, NULL,
 				       DAOS_POOL_MODULE);
-		if (rc) {
-			D_ERROR("failed to register pool RPCs: "DF_RC"\n", DP_RC(rc));
-			D_GOTO(out_rsvc, rc);
-		}
+	} else {
+		D_ASSERT(0);
```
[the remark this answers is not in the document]
wangshilong: daos_rpc_proto_query() will return error before this, so if we arrive here, it is unexpected, so i assert here.

mchaarawi (dismissed):

wiliamhuang (approved):

src/container/cli.c
```diff
@@ -23,21 +23,35 @@
...
+		return rc;
+
+	if (dc_cont_proto_version == DAOS_CONT_VERSION - 1) {
+		rc = daos_rpc_register(&cont_proto_fmt_v6, CONT_PROTO_CLI_COUNT,
+				       NULL, DAOS_CONT_MODULE);
+	} else if (dc_cont_proto_version == DAOS_CONT_VERSION) {
+		rc = daos_rpc_register(&cont_proto_fmt_v7, CONT_PROTO_CLI_COUNT,
+				       NULL, DAOS_CONT_MODULE);
+	} else {
+		D_ASSERT(0);
```
mjmac, line 50: Is `D_ASSERT()` the correct action here? What does this look like from the user perspective if they try to run with a version outside of these parameters? Seems like it would be better to log an error and return `-DER_PROTO` or perhaps a new `-DER_VER_INCOMPAT`, right?

wangshilong: If run with a version outside of these parameters, daos_rpc_proto_query() will return error and won't come here. I could modify to return -DER_PROTO instead of assert as you wish.

wangshilong: please see in daos_rpc_proto_query():
       for (i = 0; i < count; i++)
                if (ver_array[i] == rproto->version)
                        break;

        if (i == count) {
                D_ERROR("Invalid RPC protocol version %d\n", rproto->version);
                rc = -DER_PROTO;
        } else {
                *ret_ver = rproto->version;
        }

mjmac: > If run with a version outside of these parameters, daos_rpc_proto_query() will return error and won't come here. I could modify to return -DER_PROTO instead of assert as you wish.

I just noticed that @mchaarawi made the same point earlier. I think it's best to return an error instead of asserting, in case the usage of the code changes in the future. And I see that you pushed an update to do this, thanks.

mjmac (dismissed):

src/container/cli.c
```diff
@@ -956,6 +975,10 @@ dc_cont_open(tse_task_t *task)
 		dc_task_set_priv(task, cont);
 	}
 
+	/* DAOS_COO_RO_MDSTATS introduced since 2.4 */
```
kccain, line 978:
```suggestion
	/* DAOS_COO_RO_MDSTATS introduced since 2.4 (protocol version 7) */
```

very minor, only if you need to update the patch for some other reason

src/object/cli_mod.c
```diff
@@ -19,132 +19,43 @@
...
 	rc = obj_class_init();
 	if (rc)
 		D_GOTO(out_utils, rc);
 
 	dc_obj_proto_version = 0;
-
-	rc = dc_mgmt_sys_attach(NULL, &sys);
-	if (rc != 0) {
-		D_ERROR("failed to attach to grp rc "DF_RC"\n", DP_RC(rc));
+	rc = daos_rpc_proto_query(obj_proto_fmt_0.cpf_base, ver_array, 2, &dc_obj_proto_version);
```
kccain, line 40: I would not suggest changes in this patch, but maybe in the future object, pool, and container can all have some convention for naming these crt_proto_format structures/variable names to represent whatever the current and current-1 versions are without including the actual version number in the variable name. object is using fmt_1 and fmt_0 to represent (DAOS_OBJ_VERSION and DAOS_OBJ_VERSION-1 that are currently 9 and 8 respectively). So this is a pretty good model. But pool and container are using structure names like fmt_5 and fmt_4, or fmt_7 and fmt_6, with the actual protocol version number in the structure/variable name.

kccain (approved):

src/object/cli_mod.c
wangshilong: So do you think fmt_1 and fmt_0 is good practice? i remembered pool/containers fmt_5 xx followed @liw suggestions. i am good with either way, but yup, it is better to have same convention.

kccain: Yeah, would need to think about it in a little more detail to be sure. fmt_1/0 or perhaps fmt_cur/prev might be reasonable.

kccain (approved):

mchaarawi (approved):

wangshilong:
@daos-stack/daos-gatekeeper failed test are duplicated as DAOS-12883, please consider to land this pr.

Thanks

PR #12015, opened 2023-04-21
DAOS-13216 dfuse: Add a pre-read feature for non-cached files.

ashleypittman:
When the kernel cache is in use but a file is not cached
then pre-read the file on open.

Required-githooks: true


mchaarawi:
have not dug into the code yet in this PR and have just some high level questions based on the comment of pre-read you added:
- if we are doing this ls -R or recursive readdir like resnet does initially from every thread before reading any data, are we going to trigger this pre-read? (that would be disastrous). i don't think the file is opened in this case, but not sure.
- in the case of a pre-read during the open, do we read 1 mib or up to the size of the file? is this querying the file size, or does fuse or dfuse already have queried the file size before issuing the open?
- is there a practical application this is aimed for?

ashleypittman:
> have not dug into the code yet in this PR and have just some high level questions based on the comment of pre-read you added:
> 
> * if we are doing this ls -R or recursive readdir like resnet does initially from every thread before reading any data, are we going to trigger this pre-read? (that would be disastrous). i don't think the file is opened in this case, but not sure.

ls -R will not call open so no, there would be no difference to that code path.

> * in the case of a pre-read during the open, do we read 1 mib or up to the size of the file? is this querying the file size, or does fuse or dfuse already have queried the file size before issuing the open?

The file size is known during open and returned to the kernel.  This is then used for the read size.  The open performance should be unaffected (and there's no I/O changes during open).

> * is there a practical application this is aimed for?

This was tested with the Resnet training stage and helped, it should be useful for any dataset ingest which is dominated by read of smaller files.

mchaarawi (dismissed):
looks good to me. im not clear though what happens in the case if the file was written to after the pre-read happens.  Is the pre-read buffer dropped?

ashleypittman:
> looks good to me. im not clear though what happens in the case if the file was written to after the pre-read happens. Is the pre-read buffer dropped?

If the pre-read buffer is used then once it's exhausted (which might take several reads depending on application I/O sizes) then it will be released.  If there is other I/O, either writes or reads from elsewhere in the file then the buffer will not be used, but it won't be released until close.

src/client/dfuse/ops/read.c
```diff
@@ -51,6 +51,55 @@ dfuse_cb_read_complete(struct dfuse_event *ev)
...
+		return true;
+	}
+
+	if (!oh->doh_linear_read || oh->doh_readahead->dra_ev == NULL) {
+		DFUSE_TRA_DEBUG(oh, "Readahead disabled");
+		return false;
+	}
+
+	if (oh->doh_linear_read_pos != position) {
+		DFUSE_TRA_DEBUG(oh, "disabling readahead");
```
daltonbohning, line 70: So "linear read" is synonymous with "readahead"?

daltonbohning (approved):

src/client/dfuse/ops/read.c
ashleypittman: Apparently yes.  This PR started are read-ahead after the first read before I realised that the speculative read could happen on open.  This is debugging so I think it's Ok though.

daltonbohning: I was thinking from the perspective of the variable names, etc., but not asking you to change this

ashleypittman: https://github.com/daos-stack/daos/blob/master/src/client/dfuse/dfuse.h#L117-L126

`linear read` means the file was opened, read from start to end without seeking and closed.  We already track this to short-circuit the last zero-length read so we use that data to know if to use the pre-read buffer.

jolivier23 (approved):

PR #12017, opened 2023-04-21
DAOS-12002 control: General logging cleanup

mjmac:
Improve density and relevance of agent logging for client
process monitoring. Important messages should be logged
at INFO or ERROR, and DEBUG messages should avoid excess
verbosity.

Standardizes a number of debug messages to make grepping
easier and improves log density/relevance. Adds client
process info to attachInfo logs. Removes some unused symbols.

Adds a new TRACE log level for highly-verbose output that
does not normally need to be displayed at DEBUG level.

Fixes log configuration for daos_server, daos_agent, dmg,
and daos to properly handle logging NOTICE/TRACE to
logfiles. If the --debug flag is used, set loglevel to
TRACE, as the intention in this case is usually to get
maximum logging.

Required-githooks: true


daltonbohning (commented):
ftest LGTM

src/control/lib/hardware/sysfs/provider.go
```diff
@@ -484,7 +484,7 @@ func (s *Provider) ibStateToNetDevState(stateStr string) hardware.NetDevState {
 	stateSubstrs := strings.Split(string(stateStr), ": ")
 	stateNum, err := strconv.Atoi(stateSubstrs[0])
 	if err != nil {
-		s.log.Debugf("unable to parse IB state %q: %s", stateStr, err.Error())
+		s.log.Tracef("unable to parse IB state %q: %s", stateStr, err.Error())
```
kjacque, line 487: This might be worth pushing up to a higher notification level. I would not expect this to happen.

src/control/cmd/daos_agent/mgmt_rpc.go
```diff
@@ -109,13 +111,20 @@ func (mod *mgmtModule) handleGetAttachInfo(ctx context.Context, reqb []byte, pid
...
 
-	mod.log.Debugf("GetAttachInfo req from client: %+v", pbReq)
+	client := &procInfo{
+		pid: pid,
+	}
+	procName, err := common.GetProcName(int(pid))
+	if err == nil {
+		client.name = procName
+	}
+	mod.log.Tracef("%s: %s", client, pblog.Debug(pbReq))
```
kjacque, line 121: IMO this should stay at debug.

src/control/cmd/daos_agent/mgmt_rpc.go
```diff
@@ -143,8 +149,12 @@ func (mod *mgmtModule) handleGetAttachInfo(ctx context.Context, reqb []byte, pid
...
 	}
 
-	mod.log.Debugf("GetAttachInfoResp: %+v", resp)
-
+	if resp.ClientNetHint != nil {
+		mod.log.Infof("%s: numa:%d iface:%s dom:%s prov:%s srx:%d", client, numaNode,
+			resp.ClientNetHint.Interface, resp.ClientNetHint.Domain,
+			resp.ClientNetHint.Provider, resp.ClientNetHint.SrvSrxSet)
+	}
+	mod.log.Tracef("%s: %s", client, pblog.Debug(resp))
```
kjacque, line 157: I'd be inclined to keep this at debug as well.

src/control/cmd/daos_agent/mgmt_rpc.go
```diff
@@ -125,12 +134,9 @@ func (mod *mgmtModule) handleGetAttachInfo(ctx context.Context, reqb []byte, pid
 
 	numaNode, err := mod.getNUMANode(ctx, pid)
 	if err != nil {
-		mod.log.Errorf("unable to get NUMA node: %s", err.Error())
+		mod.log.Errorf("%s: unable to get NUMA node: %s", client, err)
 		return nil, err
 	}
-
-	mod.log.Debugf("client process NUMA node %d", numaNode)
```
kjacque, line 132: It is interesting to know this in the failure case as well. Keep it at trace level?

src/tests/ftest/control/dmg_network_scan.yaml
```diff
@@ -6,7 +6,7 @@ timeout: 180
 server_config:
   name: daos_server
   port: 10001
-  control_log_mask: INFO
+  control_log_mask: TRACE
```
kjacque, line 9: Is this the only test to change?

utils/node_local_test.py
```diff
@@ -748,7 +748,10 @@ def start(self):
 
         agent_config = join(self.agent_dir, 'nlt_agent.yaml')
         with open(agent_config, 'w') as fd:
-            agent_data = {'access_points': scyaml['access_points']}
+            agent_data = {
+                'access_points': scyaml['access_points'],
+                'control_log_mask': 'NOTICE',  # INFO logs every client process connection
```
kjacque, line 753: Was NLT using INFO level before? Too spammy now?

src/control/cmd/daos_agent/procmon.go
```diff
@@ -272,8 +283,6 @@ func (p *procMon) cleanupLeakedHandles(ctx context.Context, info *procInfo) {
 
 func (p *procMon) handleNotifyExit(ctx context.Context, request *procMonRequest) {
 	info, found := p.procs[request.pid]
-
-	p.log.Debugf("Received request to exit pid:%d\n", request.pid)
```
kjacque, line 276: Might be useful to keep at trace

src/control/cmd/daos_agent/main.go
```diff
@@ -202,8 +202,10 @@ func parseOpts(args []string, opts *cliOptions, invoker control.Invoker, log *lo
 			// Create an additional set of loggers which append everything
 			// to the specified file.
 			log.WithErrorLogger(logging.NewErrorLogger("agent", f)).
+				WithNoticeLogger(logging.NewNoticeLogger("agent", f)).
 				WithInfoLogger(logging.NewInfoLogger("agent", f)).
-				WithDebugLogger(logging.NewDebugLogger(f))
+				WithDebugLogger(logging.NewDebugLogger(f)).
+				WithTraceLogger(logging.NewTraceLogger(f))
```
knard-intel, line 208: The new log level  TRACE and NOTICE should be added in the comment section of the `control_log_mask` key of the file _utils/config/daos_agent.yml_

src/control/cmd/daos_agent/procmon.go
```diff
@@ -213,22 +221,19 @@ func (p *procMon) handleNotifyPoolConnect(ctx context.Context, request *procMonR
 		go info.monitorProcess(child)
 	}
 
+	p.log.Debugf("%s, connect %s/%s", info, dbgId(request.poolUUID), dbgId(request.poolHandleUUID))
 	info.handles.add(request.poolUUID, request.poolHandleUUID)
 }
 
 func (p *procMon) handleNotifyPoolDisconnect(request *procMonRequest) {
-	p.log.Debugf("Received request to disconnect pool:%s with handle %s, for pid:%d", request.poolUUID, request.poolHandleUUID, request.pid)
```
knard-intel, line 220: At first glance seems to be a relevant debugging information. Could we keep it at trace level ?

src/control/cmd/daos_server/start.go
```diff
@@ -91,9 +91,15 @@ func (cmd *startCmd) configureLogging() error {
 	// unless it was explicitly set to debug via CLI flag.
 	applyLogConfig := func() error {
 		switch logging.LogLevel(cmd.config.ControlLogMask) {
+		case logging.LogLevelTrace:
```
knard-intel, line 94: The new log level TRACE and NOTICE should be added in the comment section of the control_log_mask key of the file utils/config/daos_server.yml

src/control/cmd/dmg/main.go
```diff
@@ -262,12 +262,14 @@ and access control settings, along with system wide operations.`
...
+				WithNoticeLogger(logging.NewNoticeLogger("dmg", f)).
 				WithInfoLogger(logging.NewInfoLogger("dmg", f)).
-				WithDebugLogger(logging.NewDebugLogger(f))
+				WithDebugLogger(logging.NewDebugLogger(f)).
+				WithTraceLogger(logging.NewTraceLogger(f))
 		}
 
 		if opts.Debug {
-			log.WithLogLevel(logging.LogLevelDebug)
+			log.SetLevel(logging.LogLevelTrace)
```
knard-intel, line 272: Probably a stupid question, could we still choose the LogLevelDebug when calling the dmg command ?

knard-intel (commented):
Mostly OK for me, just some documentation should probably be updated.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -59,17 +59,18 @@ func (m *SecurityModule) getCredential(session *drpc.Session) ([]byte, error) {
 
 	signingKey, err := m.config.PrivateKey()
 	if err != nil {
-		m.log.Error(err.Error())
+		m.log.Errorf("%s: failed to get signing key: %s", info, err)
```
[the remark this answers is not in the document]
mjmac: I'm not sure how uid/gid is sensitive info. My thought was that including it could help an admin with tracking down a problem with a particular user's application, etc. Worth noting that this is not a raw/debug representation of the DomainInfo struct; I added a [new stringer](https://github.com/daos-stack/daos/pull/12017/files#diff-af690e912abd119c3eef7575b9d02915494ff41e2782990571582747444fd04dR44) that doesn't include the credential. If you think this needs more consideration, though, I'm happy to pull it out into a separate PR for extra design review.

src/control/lib/hardware/sysfs/provider.go
mjmac: Should it be an ERROR-level message?

src/control/cmd/daos_agent/mgmt_rpc.go
mjmac: Please remember that these will be happening for every single application connection. The balance that I'm trying to strike with these agent logs in particular is allowing DEBUG to be turned on at scale without causing a complete avalanche of log entries. The cached `GetAttachInfoResp` is logged once near the beginning of the log and doesn't change, so I'm not sure why we would need to log it for every application connection in the normal debug case.

Please compare the agent logs for a "busy" test, if you haven't already:
- [IOR small (new)](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-12017/4/artifact/Functional%20Hardware%20Medium/ior/small.py/daos_logs.wolf-253/test_ior_small_daos_agent.log)
- [IOR small (old)](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/master/8427/artifact/Functional%20Hardware%20Medium/ior/small.py/daos_logs.wolf-25/test_ior_small_daos_agent.log)

The new-style agent logs should be even more compact (but still very useful!) once [DAOS-13151](https://daosio.atlassian.net/browse/DAOS-13151) has been addressed.

src/control/cmd/daos_agent/mgmt_rpc.go
mjmac: Seems like maybe it'd be better in `mod.getAttachInfo()`? I can look at adding a log entry there.

src/control/cmd/daos_agent/procmon.go
mjmac: We are still logging it, just in a more compact and greppable format (from https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-12017/4/artifact/Functional%20Hardware%20Medium/ior/small.py/daos_logs.wolf-253/test_ior_small_daos_agent.log):

```
DEBUG 18:31:55.041370 procmon.go:224: pid:121756 (python3.6), connect 6a926447/18bc792d
DEBUG 18:31:55.042952 procmon.go:236: pid:121756 (python3.6), disconnect 6a926447/18bc792d
```

src/control/cmd/daos_server/start.go
mjmac: Ah, yes. I'll add those, thanks.

src/control/cmd/dmg/main.go
mjmac: I could be convinced otherwise, but my thought was that if the admin has been asked to run any of the utilities with the `--debug` flag, it is because the developers are trying to gather as much information as possible, and therefore we are OK with the extra verbosity of TRACE logging in the control plane components. This is separate from setting the log level in a config file, as it may be appropriate/useful to run with DEBUG during system validation, but TRACE is too verbose to leave enabled except for short periods of time.

utils/node_local_test.py
mjmac: As we're now logging each application connection at INFO level, yes it did become pretty spammy without any benefit IMO. NLT is kind of a special case in that it's doing fault injection and making client connections as fast as it can (many many times per second).

src/tests/ftest/control/dmg_network_scan.yaml
mjmac: I'm thinking that the vast majority of tests don't need to run at TRACE level because that stuff doesn't really change from run to run. I'd like to see how things go with the proposed logging updates and we can continue to adjust going forward. I'm absolutely willing to accept that this PR may have gone too far in the opposite direction, but I think we won't really know until we try it.

src/control/cmd/daos_agent/mgmt_rpc.go
mjmac: Just added a trace-level log here, as it's simplest and has access to the new client info.

src/control/cmd/daos_server/start.go
mjmac: Done.

src/control/cmd/daos_agent/security_rpc.go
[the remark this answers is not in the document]
kjacque: I'm okay with leaving it as-is, then.

src/control/lib/hardware/sysfs/provider.go
kjacque: Notice or error. It's not fatal, so maybe Notice would be a better fit.

src/tests/ftest/control/dmg_network_scan.yaml
kjacque: My instinct is to run with TRACE as a default, since the main use of test logs is debugging failures (which may be hard to reproduce).

utils/node_local_test.py
kjacque: Hmm, I do wonder if that's a sign that logging every client connection at INFO may make INFO too noisy. I guess we'll find out.

src/control/cmd/daos_agent/mgmt_rpc.go
kjacque: Fair point, as long as it's logged somewhere without turning on Trace.

utils/node_local_test.py
mjmac: That change came per a comment in slack by Kevin that they want to be able to see which interface was chosen for a given client process.

src/tests/ftest/control/dmg_network_scan.yaml
mjmac: I get that, but the way I'm looking at it is that running at DEBUG by default in CI forces us to be mindful of what gets logged. If we're running at TRACE all the time for all tests even though it's just noise 99% of the time, it doesn't force us to make those decisions. Honestly, I'd kind of like to run at INFO for more tests, because it will again force us to think more carefully about what should be logged in production systems.

That having been said, I'm totally in favor of finding some useful subset of tests where we should run with TRACE because those tests are exercising the network probing stuff or are otherwise specifically testing agent/hardware logic and having the extra logging is likely to yield critical details.

src/control/lib/hardware/sysfs/provider.go
mjmac: Changed it to Notice.

utils/node_local_test.py
kjacque: Ah, yes, I remember that discussion. I had added a log entry in the client logs for that as well.

kjacque (approved):

knard-intel (approved):

tanabarr (approved):

PR #12026, opened 2023-04-24
DAOS-11599 client: middleware consistency branch

mchaarawi:
Add a new API and tool extensions for invoking a checker on POSIX and PyDAOS containers. This checker scans through a POSIX/PyDAOS container and marks all OIDs as seen in an OID table, then scans the OID table to check for unmarked objects that are were not seen in the middleware namespace and print, punch, or relink those leaked objects to lost+found.

### Gatekeeper:


src/tests/SConscript
```diff
@@ -73,6 +73,12 @@ def build_tests(prereqs, env):
     # rpc tests
     SConscript('rpc/SConscript')
 
+    # daos middleware consistency corruption tool for testing
+    libs_client += ['dfs']
+    denv.AppendUnique(RPATH_FULL=['$PREFIX/lib64'])
+    daos_mw = denv.d_program('daos_mw_fi', 'daos_mw_fi.c', LIBS=libs_client, RPATH='$PREFIX/lib64')
```
jolivier23, line 79: You shouldn't need RPATH= setting here
RPATH_FULL is supposed to add any RPATH

mchaarawi: it's been a while, but it doesn't seem that i would have added that to start with :-)
i am pretty sure that i added it because it was needed..

mchaarawi: ah i see what you are saying with RPATH_FULL vs RPATH.. yea im not sure why i had that. i'll test removing the rpath. but should i commit now on the branch or should i fix this on master after the branch lands?

jolivier23: I'm fine either way.  If someone makes you update, push it in an update

jolivier23 (approved):

ci/unit/test_nlt_node.sh
```diff
@@ -21,7 +21,7 @@ sudo bash -c ". ./utils/sl/setup_local.sh; ./utils/setup_daos_server_helper.sh"
 
 # NLT will mount /mnt/daos itself.
 # TODO: Enable this for DAOS-10905
-# ./utils/node_local_test.py --max-log-size 800MiB --dfuse-dir /localhome/jenkins/ \
+# ./utils/node_local_test.py --max-log-size 1500MiB --dfuse-dir /localhome/jenkins/ \
 #			   --server-valgrind all
 
-./utils/node_local_test.py --max-log-size 900MiB --dfuse-dir /localhome/jenkins/ all
+./utils/node_local_test.py --max-log-size 1500MiB --dfuse-dir /localhome/jenkins/ all
```
ashleypittman, line 27: I would prefer not to make so big a change here for a couple of reasons.  Parsing the log-file takes some time so this will extend the test time in CI, but also most of NLT is unit testing on the client so will hit various server code paths but typically not all that many times.

NLT will start the server twice, initially with full debugging on where it runs client unit tests, then it will re-start it without debugging to test restart and at that point it runs client-heavy testing.

If this branch causes the server logfile to be so much bigger then the test should be moved after the server restart.

src/client/pydaos/pydaos_shim.c
```diff
@@ -446,6 +446,273 @@ __shim_handle__cont_close(PyObject *self, PyObject *args)
...
+	daos_size_t		buf_size = ITER_NR * 256;
+	struct pydaos_df	entry;
+	size_t			size = sizeof(entry);
+	d_iov_t			marker;
+	bool			mark_data = true;
+	int			rc = 0;
+
+	D_ALLOC(buf, buf_size);
+	if (buf == NULL)
+		goto out;
```
ashleypittman, line 468: This needs to set rc.

src/client/pydaos/pydaos_shim.c
```diff
@@ -446,6 +446,273 @@ __shim_handle__cont_close(PyObject *self, PyObject *args)
...
+
+	rc = daos_cont_query(coh, NULL, prop, NULL);
+	if (rc)
+		goto out;
+
+	/** Verify that this is a python container */
+	entry = daos_prop_entry_get(prop, DAOS_PROP_CO_LAYOUT_TYPE);
+	if (entry == NULL || entry->dpe_val != DAOS_PROP_CO_LAYOUT_PYTHON) {
+		D_ERROR("Container is not a python container\n");
+		rc = -DER_INVAL;
```
ashleypittman, line 579: D_ERROR calls should include DF_RC of the rc value.

src/client/pydaos/pydaos_shim.c
```diff
@@ -446,6 +446,273 @@ __shim_handle__cont_close(PyObject *self, PyObject *args)
...
+		goto out;
+	}
+
+	/** Fetch root object ID */
+	entry = daos_prop_entry_get(prop, DAOS_PROP_CO_ROOTS);
+	if (entry == NULL) {
+		D_ERROR("Invalid entry in properties for root object ID\n");
+		rc = -DER_INVAL;
```
ashleypittman, line 587: As above.  There seem to be many like this, I won't comment on all of them.

src/utils/daos_dfs_hdlr.c
```diff
@@ -199,3 +200,109 @@ fs_dfs_get_attr_hdlr(struct cmd_args_s *ap, dfs_obj_info_t *attrs)
...
+{
+	dfs_t		*dfs;
+	char            *name = NULL;
+	char            *dir_name = NULL;
+	int		rc, rc2;
+
+	rc = dfs_mount(ap->pool, ap->cont, O_RDWR, &dfs);
+	if (rc) {
+		fprintf(ap->errstream, "failed to mount container %s: %s (%d)\n",
+			ap->cont_str, strerror(rc), rc);
```
ashleypittman, line 215: Theres a macro that should be used here rather than fprintf.

utils/node_local_test.py
```diff
@@ -3114,6 +3120,438 @@ def test_dfuse_perms(self):
...
+        if line[-1] != 'Number of Leaked OIDs in Namespace = 0':
+            raise NLTestFail('Wrong number of Leaked OIDs')
+
+        # remount dfuse
+        dfuse = DFuse(self.server,
+                      self.conf,
+                      pool=self.pool.id(),
+                      container=self.container,
+                      caching=False)
+        dfuse.log_mask = 'CRIT'
```
ashleypittman, line 3289: What is the need to set log_mask here, are there issues with dfuse or is it to limit run-times?

I remember some discussion about this but I can't recall what the issue wasl.

ci/unit/test_nlt_node.sh
mchaarawi: the simple matter of the fact here is that the mwc adds a lot of server side debug logs. more tests = more logs.
so there is no way around this here.

src/client/pydaos/pydaos_shim.c
mchaarawi: this really should have been caught and change requested before the change landed to the feature branch. i don't see a need to repush this for that. will create a ticket to address those later.

src/client/pydaos/pydaos_shim.c
mchaarawi: will address this with your other comments in a follow on patch.

ci/unit/test_nlt_node.sh
mchaarawi: again if this needs to be changed, we can do that later.. those tests were already approved and landed to the feature branch.

mchaarawi: > If this branch causes the server logfile to be so much bigger then the test should be moved after the server restart.

IMO just rearranging a test because of server side log files size is large is just a bad approach. ideally every NLT test should have its own logs, so that is not a good solution to the issue. 1g is already large enough to parse. 2g or 3g is not going to be different..

utils/node_local_test.py
mchaarawi: correction, it was a skype discussion.
the test adds some faults and expects ERR messages later on daos commands, but NLT causes the stage to fail if it sees those error messages. so for this test, the logging is set to CRIT (which you had suggested) to avoid the NLT stage to fail, but still verifies the output from the daos tool is an error.

ci/unit/test_nlt_node.sh
ashleypittman: Overall this change adds a couple of minutes to NLT but that's probably something we have to live with.  I didn't realise this PR had server side changes.

Most tests are run in parallel so you can't distinguish the server logs between tests, the pydaos tests however are run in serial, partly for historic reasons but as they also call daos_init() it would be a challenge to make them run in parallel and maintain any ability to debug the test.

The log size with this PR is reported as 1.3GiB so this is reasonable if you don't want to run this new test without server debug and think the increased run-time is worth it.

utils/node_local_test.py
ashleypittman: Yes, I couldn't remember the details.

I've got a better idea here where we leave debugging enabled which would capture any memory leaks but install a log handler that would ignore error messages just for the specific log files and only if they had the "correct" DF_RC suffix, whatever the code is reporting in these cases.

PR #12030, opened 2023-04-24
DAOS-13170 chk: dmg check enable should not restart ranks

mjmac:
In order to avoid unexpected behavior, the `dmg check enable`
command now requires the system to be stopped first. Likewise,
the `dmg check disable` command just disables the checker
and stops the ranks. The admin will need to use `dmg system start`
again in order to restart the ranks in normal mode.

Features: recovery

Required-githooks: true


daltonbohning (commented):
ftest LGTM

PR #12032, opened 2023-04-24
DAOS-13006 cq: Remove pylint options for checking diffs.

ashleypittman:
Remove diff code from daos_pylint, we no longer support files
with known errors.

Required-githooks: true


daltonbohning (approved):

utils/cq/daos_pylint.py
```diff
@@ -641,46 +567,16 @@ def main():
...
-        # strict for files which have changed.
-        if args.git:
-            all_files = get_git_files()
-        else:
-            all_files = FileTypeList()
-        regions = None
-        file = None
-        lineno = 0
+    if args.files_from_stdin:
+        assert not args.git, 'No longer supported'
```
phender, line 571: Should this be more explicit?  Something like `The --git and --files_from_stdin argument combination is no longer supported`.

phender (approved):

ashleypittman:
Master has been pylint-clean for a couple of weeks now so there should be on in-flight PRs with issues - now would be a good time to land this.

daltonbohning:
I agree we should land this. @phender and @jolivier23 agree?

PR #12040, opened 2023-04-25
DAOS-623 test: Tidy up fault injection testing code.

ashleypittman:
Required-githooks: true

### DIFF
--- a/src/utils/daos_hdlr.c
+++ b/src/utils/daos_hdlr.c
@@ -1648,7 +1648,7 @@ dm_connect(struct cmd_args_s *ap,
 		rc = daos_cont_open(ca->src_poh, ca->src_cont, DAOS_COO_RW, &ca->src_coh,
 				    src_cont_info, NULL);
 		if (rc != 0) {
-			DH_PERROR_DER(ap, rc, "failed to open source container\n");
+			DH_PERROR_DER(ap, rc, "failed to open source containerz\n");
 			D_GOTO(err, rc);
 		}
 		if (is_posix_copy) {
--- a/utils/node_local_test.py
+++ b/utils/node_local_test.py
@@ -462,7 +462,7 @@ class DaosPool():
             return []
 
         containers = []
-        for cont in data['test']:
+        for cont in data['response']:
             containers.append(DaosCont(cont['uuid'], cont['label'], pool=self))
         return containers
 
@@ -3704,7 +3704,7 @@ def run_in_fg(server, conf, args):
             break
 
     if not container:
-        cont = create_cont(conf, pool, label=label, ctype="POSIX")
+        container = create_cont(conf, pool, label=label, ctype="POSIX")
 
         # Only set the container cache attributes when the container is initially created so they
         # can be modified later.
@@ -3716,7 +3716,7 @@ def run_in_fg(server, conf, args):
         cont_attrs['dfuse-direct-io-disable'] = False
 
         container.set_attrs(cont_attrs)
-        cont = container.id()
+        container = container.uuid
 
     dfuse = DFuse(server,
                   conf,
@@ -4016,8 +4016,6 @@ class AllocFailTestRun():
 
     def __init__(self, aft, cmd, env, loc):
 
-        # pylint: disable=consider-using-with
-
         # The subprocess handle
         self._sp = None
         # The valgrind handle
@@ -4040,11 +4038,12 @@ class AllocFailTestRun():
             prefix = f'dnt_{loc:04d}_'
         else:
             prefix = 'dnt_reference_'
-        self.log_file = tempfile.NamedTemporaryFile(prefix=prefix,
-                                                    suffix='.log',
-                                                    dir=self.aft.log_dir,
-                                                    delete=False).name
-        self.env['D_LOG_FILE'] = self.log_file
+        with tempfile.NamedTemporaryFile(prefix=prefix,
+                                         suffix='.log',
+                                         dir=self.aft.log_dir,
+                                         delete=False) as log_file:
+            self.log_file = log_file.name
+            self.env['D_LOG_FILE'] = self.log_file
 
     def __str__(self):
         cmd_text = ' '.join(self.cmd)
@@ -4173,9 +4172,9 @@ class AllocFailTestRun():
         if not self.fault_injected:
             _explain()
             return
-        if not self.aft.check_stderr:
-            _explain()
-            return
+        # if not self.aft.check_stderr:
+        #    _explain()
+        #    return
 
         # Check stderr from a daos command.
         # These should mostly be from the DH_PERROR_SYS or DH_PERROR_DER macros so check for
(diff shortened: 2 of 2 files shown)


src/client/dfs/dfs.c
```diff
@@ -1622,25 +1622,30 @@ open_sb(daos_handle_t coh, bool create, bool punch, int omode, daos_obj_id_t sup
...
+		D_GOTO(err, rc);
 	}
 
 	if (iods[LAYOUT_VER_IDX].iod_size != sizeof(layout_ver) ||
 	    layout_ver > DFS_LAYOUT_VERSION) {
-		D_ERROR("Incompatible DFS Layout version: %d\n", layout_ver);
-		D_GOTO(err, rc = EINVAL);
+		rc = EINVAL;
+		D_ERROR("Incompatible DFS Layout version: %d %d (%s)\n", layout_ver, rc,
+			strerror(rc));
```
ashleypittman, line 1647: This is wrong.

src/client/dfs/dfs.c
```diff
@@ -1622,25 +1627,30 @@ open_sb(daos_handle_t coh, bool create, bool punch, int omode, daos_obj_id_t sup
 
 	/** check if SB info exists */
 	if (iods[MAGIC_IDX].iod_size == 0) {
-		D_ERROR("SB does not exist.\n");
-		D_GOTO(err, rc = ENOENT);
+		rc = ENOENT;
```
ashleypittman, line 1630: I'm not sure if ENOENT is the correct error to return here.  If you create a posix container without a SB then you get here and ENOTENT could be mistaken for the container doesn't exist - whereas in this case the container does but it's not usable so EIO might be more suitable.

src/client/dfs/dfs.c
```diff
@@ -1249,7 +1249,12 @@ open_file(dfs_t *dfs, dfs_obj_t *parent, int flags, daos_oclass_id_t cid,
 			/** just try fetching entry to open the file */
 			daos_array_close(file->oh, NULL);
 		} else if (rc) {
-			daos_array_close(file->oh, NULL);
+			int rc2;
+
+			rc2 = daos_array_close(file->oh, NULL);
```
mchaarawi, line 1254: i really hate these changes in the code just to support fault injection. i think it's time we figure some other solution rather than polluting the code with these changes.

src/client/dfs/dfs.c
mchaarawi: i can make an argument with or against each of those error codes

src/client/dfs/dfs.c
```diff
@@ -1622,25 +1627,30 @@ open_sb(daos_handle_t coh, bool create, bool punch, int omode, daos_obj_id_t sup
...
-		D_GOTO(err, rc = ENOENT);
+		rc = ENOENT;
+		D_ERROR("SB does not exist: %d (%s)\n", rc, strerror(rc));
+		D_GOTO(err, rc);
 	}
 
 	if (magic != DFS_SB_MAGIC) {
-		D_ERROR("SB MAGIC verification failed.\n");
-		D_GOTO(err, rc = EINVAL);
+		rc = EINVAL;
```
mchaarawi, line 1636: i understand you want to have a unified error reporting format, but this is making the error message more complicated / verbose that it needs to be.. no strong objection though

src/client/dfs/dfs_sys.c
```diff
@@ -824,7 +825,10 @@ dfs_sys_stat(dfs_sys_t *dfs_sys, const char *path, int flags,
 			sys_path.name, rc);
 		D_GOTO(out_free_path, rc);
 	}
-	dfs_release(obj);
+
+	rc2 = dfs_release(obj);
+	if (rc2 == ENOMEM)
```
mchaarawi, line 830: sigh ...

src/common/prop.c
```diff
@@ -303,7 +313,11 @@ daos_prop_valid(daos_prop_t *prop, bool pool, bool input)
 		case DAOS_PROP_PO_ACL:
 		case DAOS_PROP_CO_ACL:
 			acl_ptr = prop->dpp_entries[i].dpe_val_ptr;
-			if (daos_acl_validate(acl_ptr) != 0)
+			/* This can fail with out of memory errors */
+			rc = daos_acl_validate(acl_ptr);
+			if (rc == -DER_NOMEM)
```
mchaarawi, line 318: sigh

src/include/daos_prop.h
```diff
@@ -648,6 +648,19 @@ daos_prop_free(daos_prop_t *prop);
...
+ *
+ * \param[in]	old_prop	Existing set of properties
+ * \param[in]	new_prop	New properties - may override old entries
+ * \param[out]	out_prop	New properties - may override old entries
+ *
+ * \return		0		Success
+ *			-DER_NOMEM	Out of memory
+ */
+int
+daos_prop_merge2(daos_prop_t *old_prop, daos_prop_t *new_prop, daos_prop_t **out_prop);
```
mchaarawi, line 662: so we are adding a new API just to support this fault injection framework?
i agree generally this new API looks better; but if that is the only cause it sounds unnecessary to start versioning in the API just because of that.

src/client/dfs/dfs.c
ashleypittman: Separately, I thought daos_array_close() was a local operation, does it need to be as complicated as it is?

src/include/daos_prop.h
mchaarawi: i don't think this API was used anywhere other than internally. i wonder if we can just get away by breaking the API here. probably not a good suggestion :-)

ashleypittman: I think what I've done is a good way forward, this is not a major issue.  Maintaining an API over multiple versions can be a pain.

mchaarawi (approved):

src/client/array/dc_array.c
```diff
@@ -2129,10 +2130,9 @@ punch_extent(daos_handle_t oh, daos_handle_t th, daos_size_t dkey_val, daos_off_
 	io_arg->iods	= iod;
 	io_arg->sgls	= sgl;
 
-	rc = tse_task_register_comp_cb(io_task, free_io_params_cb, &params,
-				       sizeof(params));
+	rc = tse_task_register_comp_cb(io_task, free_io_params_cb, &params, sizeof(params));
 	if (rc)
-		D_GOTO(err, rc);
+		D_GOTO(free_reqs, rc);
```
daltonbohning, line 2135: It appears `D_GOTO(err, rc);`  below should also be `D_GOTO(free_reqs, rc);`

daltonbohning (changes requested):

src/client/array/dc_array.c
ashleypittman: No, because there the free callback has been registered so that would result in a double-free.

src/include/daos_prop.h
ashleypittman: And it should be unaffected by this PR.

src/client/array/dc_array.c
daltonbohning: > the free callback has been registered

Where?

ashleypittman: Do you mean on L2139?  This line we're commenting on is registering the free callback to free params so they'll be freed as part of tse_task_complete.

As the code stands in master I was seeing a double-free although this PR does not exercise this code, as part of this work I added https://daosio.atlassian.net/browse/DAOS-13397 to create a new test for copying to an existing target which will extend coverage to this function.

daltonbohning (approved):

PR #12054, opened 2023-04-26
DAOS-12828 object: bug fix for multiple EC akeys update in one IO

liuxuezhao:
1. Fix a bug of EC multiple singv update
    When there are multiple singv update (different akey) in one IO request,
     if the first singv distributed to all shards, it possibly cause other short singv
     being written to all shards. This patch fixed it by skip those update.
2. Fix a bug of EC multiple array update in obj_get_iods_offs_by_oid().

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/object/srv_obj.c
+++ b/src/object/srv_obj.c
@@ -496,9 +496,11 @@ obj_bulk_transfer(crt_rpc_t *rpc, crt_bulk_op_t bulk_op, bool bulk_bind, crt_bul
 		  uint64_t *remote_offs, uint8_t *skips, daos_handle_t ioh, d_sg_list_t **sgls,
 		  int sgl_nr, struct obj_bulk_args *p_arg, struct ds_cont_hdl *coh)
 {
+	struct obj_rw_in	*orw = crt_req_get(rpc);
 	struct obj_bulk_args	arg = { 0 };
 	int			i, rc, *status, ret;
 	int			skip_nr = 0;
+	int			bulk_nr;
 	bool			async = true;
 	uint64_t		time = daos_get_ntime();
 
@@ -507,6 +509,7 @@ obj_bulk_transfer(crt_rpc_t *rpc, crt_bulk_op_t bulk_op, bool bulk_bind, crt_bul
 		return -DER_INVAL;
 	}
 
+	bulk_nr = orw->orw_bulks.ca_count;
 	if (p_arg == NULL) {
 		p_arg = &arg;
 		async = false;
@@ -542,9 +545,12 @@ obj_bulk_transfer(crt_rpc_t *rpc, crt_bulk_op_t bulk_op, bool bulk_bind, crt_bul
 		 * because we always create bulk handle if EC object IO targeted for multiple
 		 * data shards.
 		 */
-		if (skips != NULL && isset(skips, i))
+		while (skips != NULL && isset(skips, i + skip_nr))
 			skip_nr++;
 
+		if (bulk_nr > 0)
+			D_ASSERTF(i + skip_nr < bulk_nr, "i %d, skip_nr %d, bulk_nr %d\n",
+				  i, skip_nr, bulk_nr);
 		if (remote_bulks[i + skip_nr] == NULL)
 			continue;
 
@@ -568,8 +574,11 @@ obj_bulk_transfer(crt_rpc_t *rpc, crt_bulk_op_t bulk_op, bool bulk_bind, crt_bul
 				       bulk_op, bulk_bind, sgl, i, p_arg);
 		if (sgls == NULL)
 			d_sgl_fini(sgl, false);
-		if (rc)
+		if (rc) {
+			D_ERROR("bulk_transfer_sgl i %d, skip_nr %d failed, "DF_RC"\n",
+				i, skip_nr, DP_RC(rc));
 			break;
+		}
 	}
 done:
 	if (--(p_arg->bulks_inflight) == 0)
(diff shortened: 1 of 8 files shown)


src/vos/vos_io.c
```diff
@@ -1759,6 +1760,13 @@ akey_update(struct vos_io_context *ioc, uint32_t pm_ver, daos_handle_t ak_toh,
 			return rc;
 		}
 		flags |= SUBTR_EVT;
+	} else if (ioc->ic_ec && iod->iod_recxs != NULL) {
```
jolivier23, line 1763: why do we put this akey in the list at all?  Is it to avoid refactoring the upper layer to remove an iod from the list ?

Would it be clearer if we set a iod_flag or some such?   At the very least, I think some more comments in the VOS code around the usage of gsize would be useful.

liuxuezhao: I tested to remove/skip this akey in the iod list, but a problem is client-side bulk handle array are in original iod/akey order, if change like that, need another data structure to store which iod/akey is skipped before doing bulk transferring, that will need add some other trivial code.
Using "iod->iod_recxs" to pass singv gsize was already in current master code, this patch just add another special VOS_EC_SINGV_UPDATE_SKIP value for the skip case.
I'll add some comments, please check if it is fine with you, thanks.

src/object/srv_obj.c
```diff
@@ -1003,7 +1003,18 @@ obj_singv_ec_rw_filter(daos_unit_oid_t oid, struct daos_oclass_attr *oca,
...
+				/* VOS_EC_SINGV_UPDATE_SKIP for the case for example update 2 akeys
+				 * singv in one IO, first long singv distributed to all shards,
+				 * second short singv only stored on one data shard and all parity
+				 * shard, should skip the second update on some data shards.
+				 */
+				if (nr > 1 && tgt_off != OBJ_EC_SHORT_SINGV_IDX &&
+				    is_ec_data_shard_by_tgt_off(tgt_off, oca) &&
+				    iod->iod_size <=
+				    OBJ_EC_SINGV_EVENDIST_SZ(obj_ec_data_tgt_nr(oca)))
+					iod->iod_recxs = (void *)VOS_EC_SINGV_UPDATE_SKIP;
```
wangdi1, line 1015: Can we set iod_recxs to NULL to skip this iod? instead of a new flag?

Btw: I thought obj_get_iods_offs_by_oid() should skip this IOD anyway? or there is a bug there? Maybe fix there?

liuxuezhao: "Can we set iod_recxs to NULL to skip this iod?"
normal replica singv fetch with NULL iod_recxs, so cannot use NULL as an indicator.

"I thought obj_get_iods_offs_by_oid() should skip this IOD anyway? or there is a bug there? Maybe fix there?"
yes, I tested to skip iod/akey in obj_get_iods_offs_by_oid(), but a following problem is when doing bulk transferring, the client-side bulk handles array (one bulk handle for each iod/akey) was still same as original order, so we need another data structure to store the info of which set of iod/akey skipped and then skip corresponding client bulk handles when doing bulk transferring, that need some extra trivial code and seems worse than this patch's method.

src/object/srv_obj.c
```diff
@@ -1744,8 +1755,12 @@ obj_get_iods_offs_by_oid(daos_unit_oid_t uoid, struct obj_iod_array *iod_array,
 		oiod = &iod_array->oia_oiods[i];
 		if (iod_parent->iod_type == DAOS_IOD_ARRAY) {
 			siod = obj_shard_iod_get(oiod, local_tgt);
-			if (siod == NULL)
+			if (siod == NULL) {
+				/* will be skipped in vos_io akey_update() */
+				(*iods)[idx].iod_nr = 0;
+				idx++;
 				continue;
```
wangdi1, line 1762: Just curious, why can not we do the same thing for single value to skip the bulk transfer? or I miss sth?

wangdi1 (dismissed):

src/object/srv_obj.c
liuxuezhao: it actually is same as singv that will skip the akey update (by akey_update(), current master code already with the code to skip array's akey update for that case), then need not do bulk transferring in update_end().

liuxuezhao: I re-checked the code, the original code to skip/bypass one akey's update is actually incorrect and possibly cause space leak. I have changed that part of code by remove the akey/iod before calling into VOS, that can address @jolivier23 's concern as well.

src/object/srv_obj.c
```diff
@@ -1744,8 +1753,12 @@ obj_get_iods_offs_by_oid(daos_unit_oid_t uoid, struct obj_iod_array *iod_array,
 		oiod = &iod_array->oia_oiods[i];
 		if (iod_parent->iod_type == DAOS_IOD_ARRAY) {
 			siod = obj_shard_iod_get(oiod, local_tgt);
-			if (siod == NULL)
+			if (siod == NULL) {
+				D_DEBUG(DB_IO, "akey[%d] "DF_KEY" array skipped.\n",
+					i, DP_KEY(&iod_parent->iod_name));
+				setbit(*skips, i);
```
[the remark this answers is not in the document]
liuxuezhao: as discussed offline, as we want to "remove" the skipped iod/akey before calling in to vos, that cause the result iod array possibly mismatch with client-side bulk handle array so added the skip bitmap.

wangdi1 (approved):

src/object/srv_obj.c
```diff
@@ -533,7 +533,19 @@ obj_bulk_transfer(crt_rpc_t *rpc, crt_bulk_op_t bulk_op, bool bulk_bind,
...
+		 * every akey, but some akeys possibly be skipped in obj_get_iods_offs (number of
+		 * akeys possibly reduced). So here need to find the corresponding bulk handle
+		 * for non-skipped akey. The akey skip case only possible for EC object and targeted
+		 * for multiple data shards.
+		 * For RPC inline (non-bulk) case (bio_iod_copy) need not consider the akey skip
+		 * because we always create bulk handle if EC object IO targeted for multiple
+		 * data shards.
+		 */
+		if (skips != NULL && isset(skips, i))
+			skip_nr++;
```
Nasf-Fan, line 546: Some confused, why not directly "continue" for the skipped case? The "sgl_nr" is equal to the length of remote_bulks array or not?

liuxuezhao: cannot continue here, the sgl_nr (returned by obj_get_iods_offs) possible not equal to the length of remote_bulks (client created and in RPC).
this patch with a few issues and cannot pass one test, still need a few debug

Nasf-Fan: Then need to guarantee that the "i +skip_nr" will exceeds the remote_bulks array boundary.

liuxuezhao: sure, will add assertion check, thx

wangdi1 (approved):

jolivier23 (approved):

PR #12076, opened 2023-05-01
DAOS-13249 test: Use test.Context() instead of TODO/Background

mjmac:
Leverage the t.Cleanup() method to automatically cancel test
contexts in order to avoid leaking goroutines during test runs.

Reorganizes some tests and production code to properly clean
up when the test context is canceled.

Required-githooks: true


src/control/server/harness_test.go
```diff
@@ -97,8 +100,11 @@ func TestServer_Harness_Start(t *testing.T) {
 		},
 		"startup/shutdown with preset ranks": {
 			trc: &engine.TestRunnerConfig{
-				RunnerExitInfoCb: func() *engine.RunnerExitInfo {
-					time.Sleep(testLongTimeout)
+				RunnerExitInfoCb: func(ctx context.Context) *engine.RunnerExitInfo {
```
tanabarr, line 103: this could probably be extracted as a function variable in the test to avoid the repetition

src/control/server/instance_test.go
```diff
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2019-2022 Intel Corporation.
+// (C) Copyright 2019-2023 Intel Corporation.
```
tanabarr, line 2: there are a few copyright changes without relevant updates, presumably this is because they've been modified elsewhere since 2022?

tanabarr (approved):

kjacque (approved):

knard-intel (approved):

PR #12084, opened 2023-05-02
DAOS-11599 client: Update test to run leak checking on mwc tests.

ashleypittman:
Required-githooks: true


ashleypittman:
@mchaarawi there are only actually three errors with this PR which is fewer than I expected.  We could in theory land as-is and the warnings-ng plugin should use the correct reference builds and not cause any problems with PRs but that hasn't been our experiences in the past so I'll tidy up these error messages and add suppressions for them to the code where they're passed.

mchaarawi:
> @mchaarawi there are only actually three errors with this PR which is fewer than I expected. We could in theory land as-is and the warnings-ng plugin should use the correct reference builds and not cause any problems with PRs but that hasn't been our experiences in the past so I'll tidy up these error messages and add suppressions for them to the code where they're passed.

we don't want to add suppressions for those. they are legitimate errors that need to be flagged outside of this test. i think this is a case where you should just leave it alone :-)

ashleypittman:
I checked all the dfuse and dfs using of stderror as a result of this so the code should be a lot more consistent now.

src/client/dfuse/dfuse_main.c
```diff
@@ -557,8 +557,8 @@ main(int argc, char **argv)
 	 */
 	duns_attr.da_flags = DUNS_NO_REVERSE_LOOKUP;
 	rc = duns_resolve_path(dfuse_info->di_mountpoint, &duns_attr);
-	DFUSE_TRA_INFO(dfuse_info, "duns_resolve_path() on mountpoint returned %d %s",
-		       rc, strerror(rc));
+	DFUSE_TRA_INFO(dfuse_info, "duns_resolve_path() on mountpoint failed: %d (%s)", rc,
```
daltonbohning, line 560: This isn't necessarily a failure, right?

src/tests/ftest/cart/util/cart_logtest.py
```diff
@@ -418,11 +419,16 @@ def _check_pid_from_log_file(self, pid, abort_on_warning, leak_wf, show_memleaks
...
-                        if line.level < cart_logparse.LOG_LEVELS['ERR']:
-                            show_line(line, 'HIGH', 'error in strict mode')
-                        else:
-                            show_line(line, 'NORMAL', 'warning in strict mode')
-                        warnings_mode = True
+                        show_this = True
+                        for ignores in self.skip_suffixes:
+                            if line.get_msg().endswith(ignores):
+                                show_this = False
+                        if show_this:
```
daltonbohning, line 426: FYI - for a one-liner that also early-exits (`map` returns a generator)
```suggestion
                        if not any(map(line.get_msg().endswith, self.skip_suffixes)):
```

daltonbohning (approved):

src/client/dfuse/dfuse_main.c
ashleypittman: True, in fact I've just seen this is a INFO message not an ERROR.  I still think the change is valid however, just for consistency.

mchaarawi: how is it valid that a return value of 0 would log that it failed?

src/client/dfuse/ops/lookup.c
```diff
@@ -207,13 +207,13 @@ check_for_uns_ep(struct dfuse_projection_info *fs_handle,
 
 	rc = dfs_release(ie->ie_obj);
 	if (rc) {
-		DFUSE_TRA_ERROR(dfs, "dfs_release() failed: (%s)", strerror(rc));
+		DFUSE_TRA_ERROR(dfs, "dfs_release() returned: %d (%s)", rc, strerror(rc));
```
mchaarawi, line 210: here it's the opposite. failed is more appropriate than returned.

src/client/dfuse/dfuse_main.c
ashleypittman: Sorry, I didn't look closely enough at the change in light of Daltons comment.  Consistency with use of strerror is good but you're correct I should not have changed this to failed, I've reverted that part of the change.

src/client/dfuse/ops/lookup.c
ashleypittman: I've reverted this working change as well.

src/tests/ftest/cart/util/cart_logtest.py
ashleypittman: This PR is going to conflict with https://github.com/daos-stack/daos/pull/12112 whatever I do here so I'll leave this for now and fix up once that has landed.

mchaarawi (approved):

daltonbohning (approved):

PR #12098, opened 2023-05-05
DAOS-13198 control: Improve error for insecure incompatibility

mjmac:
When running without certificates, the interoperability checks
cannot verify the peer component, and therefore inter-version
compatibility cannot be determined. The current error message
is a little confusing, so this commit adds a new fault with
a clearer description of the problem and proposed solution.

e.g.
$ dmg storage format
ERROR: dmg: server: code = 617 description = "versions 2.3.106 and 1.2.3 are not compatible in insecure mode"
ERROR: dmg: server: code = 617 resolution = "enable certificates or use identical component versions"

Required-githooks: true

### DIFF
--- a/src/control/build/interop.go
+++ b/src/control/build/interop.go
@@ -53,7 +53,7 @@ func NewVersionedComponent(comp Component, version string) (*VersionedComponent,
 	case ComponentServer, ComponentAdmin, ComponentAgent, ComponentAny:
 		return &VersionedComponent{
 			Component: comp,
-			Version:   *v,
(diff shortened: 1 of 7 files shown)


kjacque (approved):
Nice clarification in the messaging.

knard-intel (approved):

mjmac:
@daos-stack/daos-gatekeeper: This PR passed everything except for the doxygen issue that was introduced previously.

tanabarr (approved):

PR #12099, opened 2023-05-05
amd/readdir docs

ashleypittman:
- DAOS-8722 dfuse: Document new readdir caching.
- Move the readdir buffer from alloc-every-call to saved within the readdir handle itself.


src/client/dfuse/dfuse.h
```diff
@@ -138,6 +138,75 @@ struct dfuse_obj_hdl {
 	bool                      doh_kreaddir_finished;
 };
 
+/* Readdir support.
```
daltonbohning, line 141: Should this be documented in the docs somewhere instead of buried in a `.h`? (that most people won't read)

ashleypittman: It's only of interest to people working on the code I think.  There's no public API here.

daltonbohning (commented):
This is pretty dense, so I think I'll need another read. But non-doc looks okay

src/client/dfuse/dfuse.h
```diff
@@ -138,6 +138,75 @@ struct dfuse_obj_hdl {
...
+ *
+ * Readdir is by far the most complicated component of dfuse as a result of the kernel interfaces,
+ * the dfs interface and the lack of kernel caching for concurrent operations.
+ *
+ * The kernel interface to readdir is to make a callback into dfuse to request a number of dentries
+ * which are then populated in a buffer which is returned to the kernel, each entry in the buffer
+ * contains the name of the entry, the type and some other metadata.  The kernel requests a buffer
+ * size and the dfuse can reply with less if it choses - 0 is taken as end-of-directory.  The
+ * length of the filenames affects the number of entries that will fit in the buffer.
```
daltonbohning, line 150: > the dfuse can reply with less if it choses

choses -> chooses
But does it choose to? If so, when?

> The length of the filenames affects the number of entries that will fit in the buffer.

maybe be more direct? Something like:
"The buffer is populated with as many filenames as fit."
or "The filenames are stored sequentially in the buffer, up to the buffer size."

src/client/dfuse/dfuse.h
```diff
@@ -138,6 +138,75 @@ struct dfuse_obj_hdl {
...
+ *
+ * The dfs interface to reading entries is to call dfs_iterate() which then calls a dfuse callback
+ * with the name of each entry, dfuse then has to perform a lookup to get any metadata for the
+ * entry which it does after the iterate has completed.  DFS takes in a buffer size and max count
+ * which can be much larger than the 4k buffer the kernel uses.
```
daltonbohning, line 155: > dfuse then has to perform a lookup to get any metadata for the entry which it does after the iterate has completed

Nit - Recommend wording to correspond to linear progression. E.g.
After the iterate completes, dfuse performs a lookup to get any ...

> can be much larger than the 4k buffer the kernel uses

So what does this mean for readdir caching? Do we only pass in a 4K buffer to DFS or do we pass in a larger buffer and cache the results?

src/client/dfuse/dfuse.h
```diff
@@ -138,6 +138,75 @@ struct dfuse_obj_hdl {
...
+ *
+ * The kernel uses "auto readdir plus" to switch between two types of readdir, the plus calls
+ * return full stat information for each file including size (which is expensive to read) and
+ * it takes a reference for each entry returned so requires a hash table reference for each entry.
+ * The non-plus call just takes the name and mode for each entry so can do a lighter weight
+ * dfs lookup and does not need to do any per dentry hash table operations.
+ * For any directory the first call will be a plus type, subsequent entries will depend on if the
+ * application is doing stat calls on the dentries, "ls -l" will result in readdir plus being used
+ * throughout, "/bin/ls" will result in only the first call being plus.
```
daltonbohning, line 164: Using "ls -l" and then "/bin/ls" almost implies that using the full pathname results in different behavior

src/client/dfuse/dfuse.h
```diff
@@ -138,6 +138,75 @@ struct dfuse_obj_hdl {
...
+ *
+ * In all cases the kernel holds an inode lock on readdir however this does not extend to closedir
+ * so list management is needed to protect shared readddir handles and the readdir handle pointer
+ * in the inode against concurrent readdir and closedir calls for different directory handles on the
+ * same inode.
+ *
+ * DFuse has inode handles (struct dfuse_inode_entry), open directory handles
+ * (struct dfuse_obj_hdl) as well as readdir handles (struct dfuse_readdir_hdl) which may be shared
+ * across directory handles and maybe linked from the inode handle.
```
daltonbohning, line 173:
```suggestion
 * across directory handles and may be linked from the inode handle.
```

src/client/dfuse/dfuse.h
```diff
@@ -138,6 +138,75 @@ struct dfuse_obj_hdl {
...
+ * readdir handle if it's not already.  In this way shared readdir handles never seek.
+ *
+ * Entries that have been read by dfs_iterate but not passed to the kernel or put in the cache
+ * are kept in the drh_dre entries in the readdir handle, as calls progress through the directory
+ * then these are processed and added to the reply buffer and put into the cache.  When out of
+ * entries in the array a new dfs_iterate call is made to repopulate the array.
+ *
+ * The cache is kept as a list in dfh_cache_list on the readdir handle which is a standard d_list_t
+ * however the directory handle also save a pointer to the appropriate entry for that caller.  When
+ * the front of the list is reached then new entries are consumed from the dre entry arracy and
```
daltonbohning, line 195:
```suggestion
 * the front of the list is reached then new entries are consumed from the dre entry array and
```

mchaarawi (commented):
some typos as Dalton indicated. but it would be great to split the doc only change and the code changes in separate PR.
note that the doc only change can skip all testing.

src/client/dfuse/dfuse.h
ashleypittman: I don't think that's true.  If the readdir call had to fetch more data from dfs then it'll call iterate but if it runs our again then it might return with a partially empty buffer.  You will also probable have a partially empty buffer on end-of-directory.

src/client/dfuse/dfuse.h
ashleypittman: The buffers are packed differently so 4K != 4K in terms of capacity.  dfs_iterate takes a count and a size so dfuse passes in count * (max_size + 1) so the buffer will be big enough for count no matter the file length.

I've updated this comment to say that dfuse will fetch up to 1024 entries from dfs at a time.

src/client/dfuse/dfuse.h
ashleypittman: it does.  "ls" is almost always aliased to "ls --color" or similar which actually does a stat to fetch the mode.  A common recommendation on HPC systems it to change the default shell setup to unalias ls for preciously this reason.

ashleypittman:
> some typos as Dalton indicated. but it would be great to split the doc only change and the code changes in separate PR. note that the doc only change can skip all testing.

I cherry-picked the logic change into https://github.com/daos-stack/daos/pull/12122 for review separately - I'm a lot less certain that we want that change as it may not be a benefit.

For the struct member rename, bug-fix and logging level changes I kept then here however as we definitely want them.

ashleypittman:
Sorry about the force-push, I edited a commit out of the history of this PR without thinking about the implications.

src/client/dfuse/dfuse.h
```diff
@@ -138,6 +138,76 @@ struct dfuse_obj_hdl {
...
+ *
+ * The kernel interface to readdir is to make a callback into dfuse to request a number of dentries
+ * which are then populated in a buffer which is returned to the kernel, each entry in the buffer
+ * contains the name of the entry, the type and some other metadata.  The kernel requests a buffer
+ * size and the dfuse can reply with less if it chooses - 0 is taken as end-of-directory.  The
+ * length of the filenames affects the number of entries that will fit in the buffer.
+ *
+ * The dfs interface to reading entries is to call dfs_iterate() which then calls a dfuse callback
+ * with the name of each entry, after the iterate completes dfuse then has to perform a lookup to
+ * get any metadata for the entry which it does after the iterate has completed.  DFS takes in a
```
daltonbohning, line 154: Duplicate phrase
```suggestion
 * with the name of each entry, after the iterate completes dfuse then has to perform a lookup to
 * get any metadata for the entry.  DFS takes in a
```

src/client/dfuse/dfuse.h
ashleypittman: that's an internal libfuse detail.

src/client/dfuse/dfuse.h
daltonbohning: Should the doc be explicit then? There are two differences, and reading the current wording, I (not an expert) don't know why the readdir behavior is different
- `ls -l` 
  - non-absolute path
  - argument `-l`
- `/bin/ls`
  - absolute path
  - no argument

mchaarawi (approved):

ashleypittman:
@daos-stack/daos-gatekeeper the last full build of this was green and I've only made comment changes since.

PR #12106, opened 2023-05-06
DAOS-13288 test: don't ignore pool creation failure

wangshilong:
It is not easy to know if error really comes from not enough ranks, we hit some real bugs, but test was not marked as failure.

Required-githooks: true

### DIFF
--- a/src/tests/suite/daos_rebuild_ec.c
+++ b/src/tests/suite/daos_rebuild_ec.c
@@ -267,7 +267,7 @@ rebuild_ec_setup(void  **state, int number, uint32_t rf)
 		print_message("It can not create the pool with %d ranks"
 			      " probably due to not enough ranks %d\n",
 			      number, rc);
-		return 0;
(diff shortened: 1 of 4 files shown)


wangshilong:
@wangdi1 The best way might revisit every sub rebuild test and check required ranks/targets before test and skip if there are not enough. but i am not sure every test details, i return failure for now.

src/tests/suite/daos_rebuild_ec.c
```diff
@@ -267,7 +267,7 @@ rebuild_ec_setup(void  **state, int number, uint32_t rf)
 		print_message("It can not create the pool with %d ranks"
 			      " probably due to not enough ranks %d\n",
 			      number, rc);
-		return 0;
+		return rc;
```
kccain, line 270: Did you intend to also remove the comment above "Let's skip for this case, ..."?

PR #12140, opened 2023-05-16
DAOS-13410 bio: fix smd_pool_list() without data blobs

wangshilong:
smd_pool_list() only iterate Data type, it will miss pools which were created with SCM only under MD-on-SSD case.

Required-githooks: true
Test-nvme: auto_md_on_ssd

### DIFF
--- a/src/bio/smd/smd_pool.c
+++ b/src/bio/smd/smd_pool.c
@@ -342,19 +342,27 @@ smd_pool_list_cb(struct sys_db *db, char *table, d_iov_t *key, void *args)
 	enum smd_dev_type	 st;
 	struct d_uuid            id;
 	int                      rc;
-	int			 count = 0;
+	bool			 meta_pool = false;
+
+	if (!strncmp(table, TABLE_POOLS[SMD_DEV_TYPE_META], SMD_DEV_NAME_MAX))
+		meta_pool = true;
 
 	D_ASSERT(key->iov_len == sizeof(id));
 	id = *(struct d_uuid *)key->iov_buf;
 	memset(pools, 0, sizeof(struct smd_pool) * SMD_DEV_TYPE_MAX);
 	for (st = SMD_DEV_TYPE_DATA; st < SMD_DEV_TYPE_MAX; st++) {
 		rc = smd_db_fetch(TABLE_POOLS[st], &id, sizeof(id), &pools[st], sizeof(pools[st]));
-		if (rc && rc != -DER_NONEXIST)
+		/* PMDK case, only check data blob */
+		if (!meta_pool) {
+			if (rc)
+				return rc;
+			break;
+		}
+
+		/* MD-ON-SSD case, data blob might not exist */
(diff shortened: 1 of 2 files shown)


src/bio/smd/smd_pool.c
```diff
@@ -342,16 +342,20 @@ smd_pool_list_cb(struct sys_db *db, char *table, d_iov_t *key, void *args)
...
 	int                      rc;
+	int			 count = 0;
 
 	D_ASSERT(key->iov_len == sizeof(id));
 	id = *(struct d_uuid *)key->iov_buf;
 	memset(pools, 0, sizeof(struct smd_pool) * SMD_DEV_TYPE_MAX);
 	for (st = SMD_DEV_TYPE_DATA; st < SMD_DEV_TYPE_MAX; st++) {
 		rc = smd_db_fetch(TABLE_POOLS[st], &id, sizeof(id), &pools[st], sizeof(pools[st]));
-		if (rc && !(rc == -DER_NONEXIST && st > SMD_DEV_TYPE_DATA))
+		if (rc && rc != -DER_NONEXIST)
```
NiuYawei, line 352: This looks bug-prone, It's better to cleanup the iterate callback a bit.

jolivier23 (approved):

NiuYawei (approved):

wangshilong:
@daos-stack/daos-gatekeeper https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-12140/5/artifact/Functional%20Hardware%20Medium/daos_test/rebuild.py/results.html rebuild test passed, other failures should be not related to this PR.

src/bio/smd/smd_pool.c
```diff
@@ -342,13 +342,25 @@ smd_pool_list_cb(struct sys_db *db, char *table, d_iov_t *key, void *args)
...
+	bool			 meta_pool = false;
+
+	if (!strncmp(table, TABLE_POOLS[SMD_DEV_TYPE_META], SMD_DEV_NAME_MAX))
+		meta_pool = true;
 
 	D_ASSERT(key->iov_len == sizeof(id));
 	id = *(struct d_uuid *)key->iov_buf;
 	memset(pools, 0, sizeof(struct smd_pool) * SMD_DEV_TYPE_MAX);
 	for (st = SMD_DEV_TYPE_DATA; st < SMD_DEV_TYPE_MAX; st++) {
 		rc = smd_db_fetch(TABLE_POOLS[st], &id, sizeof(id), &pools[st], sizeof(pools[st]));
```
gnailzenh, line 354: why do we fetch from all TABLE_POOLS, instead of just the current "table", since you have already tried both META and DATA in smd_pool_list()

wangshilong: So for MD-on-SSD case, if we fetch succeed from META firstly, we won't try DATA.
For PMDK case, we don't have Meta, we will try to fetch something from data.

actually META or Data is acutally iterated, this is somehow for back-compatible reason.

gnailzenh (approved):

PR #12150, opened 2023-05-17
DAOS-7029 control: Add refresh methods for agent cache

kjacque:
- Add agent config option for a cache refresh interval. The agent will perform a lazy cache refresh when the configured interval elapses.
- Allow users to force-refresh the caches with SIGUSR2. 

### Gatekeeper:


src/control/lib/cache/cache.go
```diff
@@ -0,0 +1,225 @@
+//
+// (C) Copyright 2023 Intel Corporation.
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+//
+
+package cache
```
tanabarr, line 7: is https://pkg.go.dev/istio.io/pkg/cache unsuitable? do we need to roll our own?

tanabarr (approved):
don't see any problems with the PR, only question is whether we could use an existing (mature) cache library rather than rolling our own?

src/control/lib/cache/cache.go
kjacque: The cache lib ends up being pretty simple, especially with some changes suggested by Mike. I think rolling our own makes sense here, as the functionality in the linked package is also a little different from what we're looking for.

src/control/lib/cache/cache.go
```diff
@@ -0,0 +1,196 @@
...
+	ic.mutex.RLock()
+	defer ic.mutex.RUnlock()
+
+	item, err := ic.get(key)
+	if err != nil {
+		return nil, noopRelease, err
+	}
+
+	if item.NeedsRefresh() {
+		if err := item.Refresh(ctx); err != nil {
```
mjmac, line 163: My only hesitation with this pattern (and the reason I removed `NeedsRefresh()` from the interface) is that it's potentially racy. The `NeedsRefresh()` check and `Refresh()` actions are done under different locks. It's not unsafe, as far as I can tell, but could result in a refresh being done unnecessarily if more than one goroutine does the `NeedsRefresh()` check nearly simultaneously.

kjacque: Good point, I'll revisit this.

src/control/cmd/daos_agent/config_test.go
```diff
@@ -175,3 +178,28 @@ transport_config:
 		})
 	}
 }
+
+/*func TestAgent_Config_CacheRefreshInterval(t *testing.T) {
```
tanabarr, line 182: should this be removed or uncommented?

tanabarr (dismissed):

src/control/cmd/daos_agent/config_test.go
kjacque: Ah, good catch, I forgot to remove it since it was no longer necessary.

mjmac (approved):

PR #12165, opened 2023-05-18
DAOS-13441 test: Add a test for readdir correctness after cache expire.

ashleypittman:
Add a new test that readdir caching will return stale data before
the cache has expired and correct data when it has.

Required-githooks: true


utils/node_local_test.py
```diff
@@ -1752,9 +1752,20 @@ class PrintStat():
...
         # Setup the object, and maybe add some data to it.
         self._stats = []
+        self.count = 0
         if filename:
             self.add(filename)
 
+    def dir_add(self, dirname):
+        """Add a directory contents
+
+        This differs from .add(dirname, show_dir=True) as it does not add the dir itself can the
```
jamesanunez, line 1762: "can" should be "add"?

utils/node_local_test.py
```diff
@@ -1995,6 +2010,101 @@ def test_two_mounts(self):
...
+
+        stat_log_oob = PrintStat()
+        stat_log_oob.dir_add(dfuse1.dir)
+        print(stat_log_oob)
+
+        stat_log1 = PrintStat()
+        stat_log1.dir_add(dfuse0.dir)
+        print(stat_log1)
+
+        elapsed = time.time() - start
```
daltonbohning, line 2084: FYI that `time.perf_counter` and `time.process_time` are more reliable. But `time.time` is probably okay here

daltonbohning (approved):

utils/node_local_test.py
```diff
@@ -1995,6 +2010,101 @@ def test_two_mounts(self):
...
+
+        re-read directory contents again, now this should be up to date.
+        """
+        cache_time = 20
+
+        cont_attrs = {}
+        cont_attrs['dfuse-data-cache'] = False
+        cont_attrs['dfuse-attr-time'] = cache_time
+        cont_attrs['dfuse-dentry-time'] = cache_time
+        cont_attrs['dfuse-ndentry-time'] = cache_time
```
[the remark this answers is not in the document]
ashleypittman: This is a cut-and-paste change, yes it would better like that.  Historically it used OrderedDict() to be deterministic however that's no longer required and it didn't get updated full when that was removed.

jamesanunez (approved):

utils/node_local_test.py
ashleypittman: Good to know.  I've used this code in a few places, will update.

daltonbohning: And to elaborate, `time.perf_counter` includes time during sleep, whereas `time.process_time` does not:
https://docs.python.org/3/library/time.html#time.perf_counter

PR #12171, opened 2023-05-19
DAOS-13447 client: Update hash table usage in pil4dfs.

ashleypittman:
Update hash table calls to be thread safe.
Use locking consistently.
Correctly handle the string including tracking the size properly.
Improve the performance of lookup.
Reduce the size of descriptors in the hash table.
Pass in dfs_mt * into lookup_insert_dir() rather than the idx.

Turn on leak checking in NLT tests.
Add a new test but leave disabled due to memory leaks.

Required-githooks: true

### DIFF
--- a/src/client/dfuse/pil4dfs/int_dfs.c
+++ b/src/client/dfuse/pil4dfs/int_dfs.c
@@ -265,11 +265,12 @@ static d_hash_table_ops_t hdl_hash_ops = {.hop_key_cmp    = key_cmp,
  * should be length of the string, not including any \0 termination.
  */
 static int
-lookup_insert_dir(struct dfs_mt *mt, const char *name, size_t len, mode_t *mode, dfs_obj_t **obj)
+lookup_insert_dir(struct dfs_mt *mt, const char *name, size_t len, dfs_obj_t **obj)
 {
 	struct dir_hdl *hdl = NULL;
 	dfs_obj_t      *oh;
 	d_list_t       *rlink;
+	mode_t          mode;
 	int             rc;
 
 	/* TODO: Remove this after testing. */
@@ -282,11 +283,11 @@ lookup_insert_dir(struct dfs_mt *mt, const char *name, size_t len, mode_t *mode,
 		return 0;
 	}
 
-	rc = dfs_lookup(mt->dfs, name, O_RDWR, &oh, mode, NULL);
+	rc = dfs_lookup(mt->dfs, name, O_RDWR, &oh, &mode, NULL);
 	if (rc)
 		return rc;
 
-	if (!S_ISDIR(*mode)) {
+	if (!S_ISDIR(mode)) {
 		*obj = oh;
 		return 0;
 	}
@@ -761,7 +762,6 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
 	   char **parent_dir, char **full_path, struct dfs_mt **dfs_mt)
 {
 	int    pos, len;
-	mode_t mode;
 	bool   with_daos_prefix;
 	char   pool[DAOS_PROP_MAX_LABEL_BUF_LEN + 1];
 	char   cont[DAOS_PROP_MAX_LABEL_BUF_LEN + 1];
@@ -932,7 +932,7 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
 					path_len);
 			}
 			/* look up the dfs object from hash table for the parent dir */
-			rc = lookup_insert_dir(*dfs_mt, *parent_dir, path_len, &mode, parent);
+			rc = lookup_insert_dir(*dfs_mt, *parent_dir, path_len, parent);
 			/* parent dir does not exist or something wrong */
 			if (rc)
 				D_GOTO(out_err, rc);


src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -919,19 +920,29 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
...
 			if (pos == (*dfs_mt)->len_fs_root) {
+				/* TODO: Do we need to call lookup_insert_dir() in this case, as
+				 * there's no ref counting on the hash table it's probably
+				 * sufficient to keep a pointer to the root dfs object and use
+				 * that here.
+				 */
 				*parent       = NULL;
 				(*parent_dir)[0] = '/';
-				(*parent_dir)[1] = '\0';
+				path_len         = 1;
```
ashleypittman, line 933: Do you intend for the DFS root to be in the hash table or not?  I think most DFS options can take NULL for a parent and then assume the root but it would certainly be more consistent if we always passed oh - and that's what the existing code does although the fact you set *parent to NULL here makes me wonder if that was intentional or not.

Assuming we do want / in the hash table then I suspect we could just avoid this conditional and just do the strncpy in either case which would simplify the code.

In terms of the bigger picture I the query_path function is passing out an allocated parent_dir here but actually it could return a pointer to the hash table entry, then the string would not need to be allocated or copied (or freed by the callers) but that's a bigger change and I was trying to stick to bug-fixes here.

Finally, I think overall this function would be better if it set `*parent = NULL` on the way in and then only overwrite it if there is a valid value for it.  Right now it's set to null in most (or all cases?) and sometimes set to NULL and then overwritten.  If you did this then you could remove the `*ojb = NULL` in lookup_insert_dir() as well and the code would be safer overall.

utils/node_local_test.py
```diff
@@ -3715,6 +3716,12 @@ def test_pil4dfs(self):
         file3 = join(path, 'file3')
         self.server.run_daos_client_cmd_pil4dfs(['touch', file3])
 
+        # cat a filename where a directory in the path is a file, should fail.  For now this does
+        # work but leaks memory.
+        # nop_file = join(file3, 'new_file which will not exist...')
+        # rc = self.server.run_daos_client_cmd_pil4dfs(['cat', nop_file], check=False)
+        # assert rc.returncode == 1, rc
+
```
ashleypittman, line 3724: For this one I propose landing this PR with the test added-but-disabled and then you can look at the memory leak.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -241,79 +247,76 @@ rec_hash(struct d_hash_table *htable, d_list_t *rlink)
...
 		return 0;
 	}
 
-	D_ALLOC_PTR(hdl);
+	/* Allocate struct and string in a single buffer.  This includes a extra byte so name will
+	 * be \0 terminiated however that is not required.
+	 */
+	D_ALLOC(hdl, sizeof(*hdl) + len + 1);
 	if (hdl == NULL) {
+		D_GOTO(out_release, rc = ENOMEM);
```
wiliamhuang, line 300: Need to call "dfs_release(oh);" before "D_GOTO(out_release, rc = ENOMEM);".

utils/node_local_test.py
wiliamhuang: I created a ticket for the memory leak issue. I will enable this test after the issue is fixed.

wiliamhuang: https://daosio.atlassian.net/browse/DAOS-13481

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: Thank you.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -241,79 +247,76 @@ rec_hash(struct d_hash_table *htable, d_list_t *rlink)
...
 		dfs_release(oh);
 		return ENOMEM;
 	}
 
-	/* "len + 1" is used to avoid compiling issue on Ubuntu. */
-	strncpy(hdl->name, name, len + 1);
-	hdl->oh = oh;
+	hdl->name_size = len;
+	hdl->oh        = oh;
+	strncpy(hdl->name, name, len);
```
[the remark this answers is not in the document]
ashleypittman: `hdl->name` is a zero sized array so after having allocated the struct you can just copy to it.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: > Regarding "return a pointer to the hash table entry", are the location of each entry fixed or they could be re-positioned or even evicted?

They will be fixed and without reference counting then they won't be evicted either.  I think the best here would be for lookup_insert_dir and query_path to return a hash table entry with reference held and have the caller call decref() when done but that's a bigger change.

ashleypittman: > I do not remember exactly now. Yes. I agree that we can return after "path_len = 1;" and simplify else {}.

I think it would be better for this function to consistently return a `dfs_obt_t *` for now even if it may not be strictly needed.

src/client/dfuse/pil4dfs/int_dfs.c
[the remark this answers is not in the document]
ashleypittman: I disagree.  Please look closer at what is happening here.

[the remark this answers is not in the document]
wiliamhuang: You are right. I missed it! Thank you!

wiliamhuang:
It looks good to me. @mchaarawi Could you please review? Thank you!

wiliamhuang:
The failures of "Functional Hardware Medium Verbs Provider" seem interim recently in multiple PRs. I think they are not related, but I could not find existing tickets.

mchaarawi (approved):

wiliamhuang:
The failed case in build 4 is existing issue. Ticket, https://daosio.atlassian.net/browse/DAOS-13488

PR #12173, opened 2023-05-19
DAOS-13485 client: fix several coverity issues in libpil4dfs

wiliamhuang:
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfuse/pil4dfs/hook.c
+++ b/src/client/dfuse/pil4dfs/hook.c
@@ -104,6 +104,7 @@ static char     *path_libpthread;
 static int
 read_map_file(char **buf, const char str_target[], char **pos)
 {
+	bool buffer_full;
 	int max_read_size, complete = 0, read_size;
 	FILE *fIn;
 
@@ -112,6 +113,7 @@ read_map_file(char **buf, const char str_target[], char **pos)
 	/* There is NO way to know the size of /proc/self/maps without reading the full file */
 	/* Keep reading the file until finish the whole file. Increase the buffer size if needed. */
 	while (complete == 0) {
+		buffer_full = false;
 		*buf = malloc(max_read_size + 1);
 		if (*buf == NULL) {
 			printf("Failed allocate memory (%d bytes).\nQuit\n", max_read_size + 1);
@@ -135,8 +137,8 @@ read_map_file(char **buf, const char str_target[], char **pos)
 			exit(1);
 		} else if (read_size == max_read_size) {
 			/* need to increase the buffer and try again */
-			free(*buf);
 			max_read_size *= 3;
+			buffer_full = true;
 		} else {
 			/* reached the end of the file */
 			complete = 1;
@@ -152,6 +154,8 @@ read_map_file(char **buf, const char str_target[], char **pos)
 				break;
 			}
 		}
+		if (buffer_full)
+			free(*buf);
 		if (max_read_size >= MAP_SIZE_LIMIT) {
 			printf("/proc/self/maps is TOO large!\nQuit.\n");
 			exit(1);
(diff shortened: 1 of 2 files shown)


wiliamhuang:
All CI failures are existing issues.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -4667,8 +4670,6 @@ dup2(int oldfd, int newfd)
 			return fd;
 		else
 			return idx;
-	} else {
-		next_dup2(oldfd, newfd);
```
wangshilong, line 4671: just a minder this is on purpose right?

src/client/dfuse/pil4dfs/hook.c
```diff
@@ -914,14 +918,14 @@ register_a_hook(const char *module_name, const char *func_name, const void *new_
 		get_module_maps_inited = 1;
 	}
 
-	if (strcmp(module_name, "ld") == 0)
-		strcpy(module_name_local, path_ld);
-	else if (strcmp(module_name, "libc") == 0)
-		strcpy(module_name_local, path_libc);
-	else if (strcmp(module_name, "libpthread") == 0)
-		strcpy(module_name_local, path_libpthread);
+	if (strncmp(module_name, "ld", 3) == 0)
```
mchaarawi, line 921: i think strncmp should not include the NULL terminator, so it should be in this case strlen("ld") instead of 3.
strncpy should be strlen("ld")+1

same comment applies to the others.

src/client/dfuse/pil4dfs/hook.c
```diff
@@ -112,6 +113,7 @@ read_map_file(char **buf, const char str_target[], char **pos)
 	/* There is NO way to know the size of /proc/self/maps without reading the full file */
 	/* Keep reading the file until finish the whole file. Increase the buffer size if needed. */
 	while (complete == 0) {
+		buffer_full = false;
```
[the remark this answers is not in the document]
wiliamhuang: "/proc/self/maps" is special. stat() does not work to get file size. st_size is always zero. One has to read the file to find out the file size.

src/client/dfuse/pil4dfs/hook.c
wiliamhuang: The NULL terminator is included on purpose, so exact full string matching is used. It is not only matching the initial sub-string.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3480,19 +3480,22 @@ rename(const char *old_name, const char *new_name)
 		}
 		/* New_name was removed, now create a new one from the old one */
 		switch (type_old) {
+			ssize_t link_len_libc;
```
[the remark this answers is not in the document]
wiliamhuang: I used "daos_size_t      link_len;" to store the read in string length of a symbol link with both dfs_get_symlink_value() and readlink(). "daos_size_t" is a unsigned int64. The return type of readlink() is ssize_t, which has a sign.
Coverity complained comparing a unsigned number with zero in "link_len < 0", so I need a new variable for the return value of readlink() here.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Coverity will still complain dead code here. I will set ignoring this issue in coverity at this moment.

mchaarawi: ah i see. then nevermind adding the assert

mchaarawi (approved):

wiliamhuang:
Force landing is added since there are non-related existing failure in CI tests.

mchaarawi:
> Force landing is added since there are non-related existing failure in CI tests.

can you please paste the ticket here where those failures are being tracked and increment the occurence count in Jira?

src/client/dfuse/pil4dfs/hook.c
```diff
@@ -152,6 +154,8 @@ read_map_file(char **buf, const char str_target[], char **pos)
 				break;
 			}
 		}
+		if (buffer_full)
+			free(*buf);
 		if (max_read_size >= MAP_SIZE_LIMIT) {
```
jolivier23, line 159: If we have this hard limit on the size of the file, why not just allocate that and avoid the complication.   Probably better for fragmentation to do one large malloc/free than several of differing sizes.

wiliamhuang: Normally the size of "/proc/self/maps" should be small, less than 100 K bytes. I would expect that the adaptively increasing buffer size only will occur in very rare cases.

wiliamhuang: @jolivier23 free() and re-allocation should be not likely triggered since the default buffer size 512K is quite large already. Normally, "/proc/self/maps" during application initialization phase is far smaller.

wiliamhuang:
@jolivier23 How do you want to proceed? allocate the limit size large buffer for reading "/proc/self/maps" or keeping the current form?

jolivier23:
> @jolivier23 How do you want to proceed? allocate the limit size large buffer for reading "/proc/self/maps" or keeping the current form?

I'm ok with current PR since it doesn't actually change the code but I think we should address my concerns.   I think my personal preference would be to handle it via parsing rather than reallocation

jolivier23:
whatever we do, we should have a test that covers the exceptions.

wiliamhuang:
Jeff and I discussed offline. I will revise the code later to couple reading and parsing with a small fixed-size buffer to be safer.

src/client/dfuse/pil4dfs/hook.c
ashleypittman: I was thinking the same, unexercised code is something we want to avoid.

Perhaps this would be a case for realloc() in a loop?

PR #12183, opened 2023-05-22
DAOS-13447 pil4dfs: Add NLT test for running without dfuse.

ashleypittman:
Required-githooks: true


ci/jira_query.py
```diff
@@ -27,8 +27,8 @@
 # Expected components from the commit message, and directory in src/, src/client or utils/ is also
 # valid.  We've never checked/enforced these before so there have been a lot of values used in the
 # past.
-VALID_COMPONENTS = ('agent', 'build', 'ci', 'csum', 'doc', 'gha', 'il', 'md', 'mercury', 'swim',
-                    'test', 'tools')
+VALID_COMPONENTS = ('agent', 'build', 'ci', 'csum', 'doc', 'gha', 'il', 'md', 'mercury', 'pil4dfs',
+                    'swim', 'test', 'tools')
```
daltonbohning, line 31: There's nothing formal about this, right? It's somewhat arbitrary what we allow?

utils/node_local_test.py
```diff
@@ -1072,10 +1079,21 @@ def run_daos_client_cmd_pil4dfs(self, cmd, check=True):
         cmd_env['DAOS_AGENT_DRPC_DIR'] = self.conf.agent_dir
         cmd_env['D_IL_REPORT'] = '1'
         cmd_env['LD_PRELOAD'] = join(self.conf['PREFIX'], 'lib64', 'libpil4dfs.so')
+        if container is not None:
+            # Create a temporary directory for the mount point, this will be removed as it goes out
+            # scope so keep as a local for the rest of the function.
+            # pylint: disable-next=consider-using-with
+            tmp_dir = tempfile.TemporaryDirectory(prefix='pil4dfs_mount')
+            cwd = tmp_dir.name
```
daltonbohning, line 1087: You might consider `tempfile.mkdtemp`, which I think pylint doesn't complain about
https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp
Because:
> The user of [mkdtemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) is responsible for deleting the temporary directory and its contents when done with it.

Though I haven't verified

daltonbohning (approved):

wiliamhuang (approved):

ci/jira_query.py
ashleypittman: Traditionally there was nothing formal about it and all kinds of values was used, this is the only place it's checked and we do seem to be sticking to the known components now.  Note that it'll also accept the names of directories in src or src/client, this list isn't all the acceptable values.

We don't yet use this value for anything else like selecting testing which I think we could probably start doing now, and we don't make any attempt to verify that the component is broadly correct.

utils/node_local_test.py
ashleypittman: Well, we do want the directory removed.  The only reason we need to suppress this is because doing it properly would require re-indenting the rest of the function and more testing, something that would make sense when the code was written but for now it's easier to move forward with the suppression.

PR #12193, opened 2023-05-23
DAOS-13227 dfuse: Use 16 threads by default.

ashleypittman:
Default to 16 threads rather than 1-per-core.
If a cpuset is defined then use 1 thread per core still,
however if nothing is defined then limit to 16.

It is still possible to over-commit by using the command line.

Switch to using hwloc to read the cpusets as the standard Linux
functions only tell you the allowed size, not the total
size.

Required-githooks: true


mchaarawi (commented):
was there any performance evaluation done with this?

ashleypittman:
> was there any performance evaluation done with this?

No, but we've previously found that the default was too high for resource consumption (even though threads will spend their time sleeping).  The recommendation is always to pick the cores you want dfuse to run on and bind to that - dfuse will use one thread per available core, this is more about having sensible defaults in cases where that't not done.

I believe most of the performance testing we've done is with 16 threads as that's standard for our test configs but to my knowledge we've not put much effort into refining that.

mchaarawi (approved):

jolivier23 (approved):

PR #12199, opened 2023-05-24
DAOS-13239 cart: Fix send failure error mapping..

jgmoore-or:
### Before requesting gatekeeper

### Gatekeeper:


src/include/daos_errno.h
```diff
@@ -161,7 +161,10 @@ extern "C" {
 	       incompatible user or group permissions)			\
 	/** Fatal (non-retry-able) transport layer mercury error */	\
 	ACTION(DER_HG_FATAL,		(DER_ERR_GURT_BASE + 45),	\
-	       Fatal transport layer mercury error)
+	       Fatal transport layer mercury error)			\
+	/** Failed to send rpc */                                       \
+	ACTION(DER_HG_SEND_FAILED,      (DER_ERR_GURT_BASE + 46),       \
+		Failed to send RPC)
```
daltonbohning, line 167: Can we just remove this? Wouldn't that cause compatibility issues?

jgmoore-or: What do you want removed?  I think the added lines of code are needed.

daltonbohning: Never mind. The GitHub diff was showing that the above DER_HG_FATAL was removed entirely. But now it looks like you just added a new one, which is okay.

daltonbohning (commented):
is raft meant to be updated?

frostedcmos (commented):
cart changes lgtm, raft needs to be removed from submission

frostedcmos (approved):

src/cart/crt_hg.c
```diff
@@ -1309,7 +1309,17 @@ crt_hg_req_send_cb(const struct hg_cb_info *hg_cbinfo)
 		break;
 	default:
 		rpc_priv->crp_state = RPC_STATE_COMPLETED;
-		rc = crt_hgret_2_der(hg_cbinfo->ret);
+
+		if (crt_is_service()) {
```
liw, line 1313: [Question] Why do we differentiate "service" and "non-service" cases? I'm curious because part of libdaos code is also used in a daos_engine, where `crt_is_service` returns true. (Is that right, @wangdi1?)

frostedcmos: the idea here is that client will just get all errors squashed as 'couldnt send', as a client shouldnt really care about why the destination was not reachable, while server might need an exact error

liw: I see. Just wanted to remind that the `daos` "client" code executes both in "service" and "non-service" cases.

jgmoore-or: The engines essentially do the old client behavior (receive and evaluate detailed results from the send) whereas the the clients now do a simplified handling (retrying on all failures). This makes sense conceptually, and it allows all of the daos_test suite to pass (without the crt_is_service check, DTX and rebuild tests were failing).

PR #12205, opened 2023-05-25
DAOS-13510 chk: proper message for inconsistency report

Nasf-Fan:
To avoid confusing under dryrun mode.

### Gatekeeper:

### DIFF
--- a/src/chk/chk_engine.c
+++ b/src/chk/chk_engine.c
@@ -337,8 +337,9 @@ report:
 	cru.cru_pool = (uuid_t *)&cpr->cpr_uuid;
 	cru.cru_pool_label = cpr->cpr_label;
 	snprintf(msg, CHK_MSG_BUFLEN - 1,
-		 "Check engine detects orphan %s entry in pool map for "
+		 "Check engine %s orphan %s entry in pool map for "
 		 DF_UUIDF", rank %u, index %d\n",
+		 prop->cp_flags & CHK__CHECK_FLAG__CF_DRYRUN ? "dryrun-detects" : "detects",
 		 index < 0 ? "rank" : "target", DP_UUID(cpr->cpr_uuid), rank, index);
 	cru.cru_msg = msg;
 	cru.cru_options = options;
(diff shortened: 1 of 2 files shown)


Nasf-Fan:
The patch only changes the log message, not related with the CI test failures.

Nasf-Fan:
Ping @mjmac @liw , thanks!

liw (approved):
[Not requesting changes] Generally speaking, we should have two kinds of messages:

1. "Detected fault X."
2. "Repaired fault X."

Then in a dry run, we could print the first kind of messages but not the second kind of messages---conveying a clearer meaning to users. (Or we could print both kinds of messages but change "repaired" to "skipped repairing.")

Nasf-Fan:
> [Not requesting changes] Generally speaking, we should have two kinds of messages:
> 
> 1. "Detected fault X."
> 2. "Repaired fault X."
> 
> Then in a dry run, we could print the first kind of messages but not the second kind of messages---conveying a clearer meaning to users. (Or we could print both kinds of messages but change "repaired" to "skipped repairing.")

Here is example of our current inconsistency report:

Inconsistency Reports:
  ID:         0x10e25057f98c0002
  Class:      POOL_NONEXIST_ON_ENGINE
  Message:    Check leader detects dangling pool.
  Pool:       mypool
  Resolution: IGNORE: Ignore the pool finding

The "Message" section describes what the inconsistency is; the "Resolution" section describes how the inconsistency is handled. You expect to change the "Resolution" instead of the "Message". That is possible. But "ignore" is already used for other purpose. We need to define new resolution type and transfer it from engine to MS, then MS can covert such new type to some suitable human readable string. That will change the proto between engine and MS. That is why I made current simple PR.

mjmac (commented):
Hmm. From a UX standpoint, it seems a little awkward to me that we would expose a new verb (dryrun-detect) to the user and make all of these changes to user-visible strings. In other tools that I have used, the convention seems to be that dryrun mode is announced via a single line printed before the rest of the output.

If the concern is that we could have mixed reports in dryrun-on/dryrun-off mode, then perhaps the best way to handle this is to add a Dryrun bool to the report message, then we can properly handle this out in the display logic.

Nasf-Fan:
> Hmm. From a UX standpoint, it seems a little awkward to me that we would expose a new verb (dryrun-detect) to the user and make all of these changes to user-visible strings. In other tools that I have used, the convention seems to be that dryrun mode is announced via a single line printed before the rest of the output.

I am fine for this way if you can improve the "check query" output with the expected dryrun information to avoid confusing. Thanks!

Nasf-Fan:
The issue will be handled via control plane logic (by @mjmac ), do not need this patch any longer.

PR #12217, opened 2023-05-26
DAOS-13516 object: shrink the group number of GX object

wangdi1:
Shrink the group number of GX object to avoid putting multiple shards in the same domain during rebuild, which might break RF setting.

GX group number will be shrinked from target_num/domain_nr to target_num/domain_nr - RF.

Required-githooks: true

### Gatekeeper:


wangdi1:
passed all tests here.

wangshilong (approved):

wangshilong (dismissed):

liuxuezhao (approved):
not very sure can the layout upgrade deal with such layout change?

src/object/obj_class.c
```diff
@@ -270,9 +270,17 @@ daos_oclass_fit_max(daos_oclass_id_t oc_id, int domain_nr, int target_nr,
...
+	if (ca.ca_grp_nr == DAOS_OBJ_GRP_MAX) {
 		ca.ca_grp_nr = max(1, (target_nr / grp_size));
 
+		/* To honor RF setting during failure cases, let's reserve RF
+		 * groups, so if some targets fail, there will be enough replacement
+		 * targets to rebuild, so to avoid putting multiple shards in the same
+		 * domain, which may break the RF setting.
+		 */
+		if (ca.ca_grp_nr > rf_factor)
+			ca.ca_grp_nr -= rf_factor;
```
mchaarawi, line 282: what if the RF==0 but the user is using oclass for an object like EC_*GX?
then group number is not shrunk or adjusted. the container or group RF is just a minimum setting, but objects can have higher level of redundancy themselves.

wangdi1: If RF = 0, then it means the user does not "care" the data lose, i.e. it is ok to assign multiple shards in the same domain, IMHO.  

And, if the user really want to use all targets for performance reason,  then he can choose RF = 0 .

mchaarawi: no that's not the case. one can set rf=0 and individually do redundancy on some object and others not. so we cannot assume: "If RF = 0, then it means the user does not "care" the data lose"

mchaarawi: anyway this is still better than before at least. let's check for @gnailzenh input on this.

wangdi1: "If RF = 0, then it means the user does not "care" the data lose"

Oh, sorry for confusion, "not care data lose" may not be the right word here.  I mean RF = 0, any target exclusion will cause container to be inaccessible, i.e. the user get notification. So it does not care if it will put multiple shards in the same domain.

wangdi1: >I also have the same question as Mohamad, if RF=0, I think we can still check ca.ca_resil_degree correct? Or you want to keep this as an interface to try all targets for EC & replicaiton?

Hmm, this patch is for honoring the RF setting, so
1. the user can get notification if there are potential data loss.
2. the data in the container can keep complete as long as the number of failure targets is less than the RF setting.

So if RF is 0, then the user will get notification if any targets excluding happened, and also probably the user want to get maximum performance anyway. IMHO, we do not need reserve the target anyway.

> The last question is, if user sets RC=2 and use EC_16P2_GX, we need to reserve 36 targets, but if user chooses EC_4P2_GX, then this function reserves 12 targets, I'm wondering if we can set an upper bound for the number, so we don't have to reserve too many targets.

Indeed, probably we need set some boundary here.

liuxuezhao (approved):

wangshilong (approved):

wangshilong (approved):

wangshilong (approved):

wangdi1:
Sorry, need increase drain timeout value.

daltonbohning (commented):
ftest LGTM

wangshilong (approved):

wangshilong (approved):

wangdi1:
passed all of tests.

src/object/obj_class.c
```diff
@@ -234,9 +234,24 @@ daos_oclass_grp_nr(struct daos_oclass_attr *oc_attr, struct daos_obj_md *md)
 	return oc_attr->ca_grp_nr;
 }
 
+/**
+ * To honor RF setting during failure cases, let's reserve RF
+ * groups, so if some targets fail, there will be enough replacement
+ * targets to rebuild, so to avoid putting multiple shards in the same
+ * domain, which may break the RF setting.
+ *
+ * Though let's keep reserve targets to be less than 30% of the total
```
mchaarawi, line 243: can you please add some comments here as to why 30% max?
what happens if rf > 30% (probably unlikely though to run into this) ?

wangdi1: 30% is just a rough number to avoid reserving too much targets here.   So we will reserve at most 30% of the total targets, if rf > 30%, then it will not reserve too much targets for the objects, i.e. RF promise might be broken even before RF targets are excluded.

Anyway,  since this patch is just the temporary solution for this RF dilemma for 2.4, so we are are trying to balance the performance and RF.

src/object/obj_class.c
```diff
@@ -234,9 +234,24 @@ daos_oclass_grp_nr(struct daos_oclass_attr *oc_attr, struct daos_obj_md *md)
...
+ * targets.
+ */
+static uint32_t
+reserve_grp_by_rf(uint32_t target_nr, uint32_t grp_size, uint32_t rf)
+{
+	return min(((target_nr * 3) / 10) / grp_size, rf);
```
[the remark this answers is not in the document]
wangdi1: > im still not convinced that an object with EC or RP, in a container with RF0 is not supported here..
> it sounds like we are telling users that containers with RF0 do not support any object with redundancy, however it should.

I think this is what RF means, once the failure reach the RF, the container will turns to inaccessible.  There are no way for user to know if there are data lost in their container, unless we do object scan etc, but this is not supported in the current version.

So object class have to work together with RF to achieve the data redundancy in the current version.


> same case applies for an object with EC_xP2 or RP3 in a container with RF1 .. we don't reserve enough in this case here either right?

Yes, reservation is by RF,  which is the only one to be used for checking the data lost for now.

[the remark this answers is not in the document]
mchaarawi: i agree that this is at least better than what we have now. let's document this properly though for users to understand the limitation of using redundant object class but with RF that is at a lower level of redundancy (this can be a common usecase fwiw).

wangdi1:
ping

gnailzenh (approved):

PR #12260, opened 2023-06-02
DAOS-13089 control: Fix debounce evt unit test race

tanabarr:
The TestEvents_PubSub_Debounce_Cooldown go unit test is intermittently
failing due to sporadic VM multi-millisecond lockups causing cooldown
timeout expiry. The test debug shows events having >10ms latency very
occasionally resulting in cooldown period being exceeded and more than
expected events being recorded. To ensure that the test is reliable in
CI this change applies the following approach:

- Increase cooldown time period to 25ms from 10ms to reduce the chance
  of false expiry.
- Validate that single events are recorded from each outer-loop
  iteration (where event incarnation is increased) and incarnation
  value matches what is expected whilst all other events are ignored
  through debounce logic
- If test loop gets stalled for more or equal length of time to the
  cooldown period, restart the test.

In the remote eventuality that the test continues to fail and is
repeatedly restarted, the outer 10m timeout of the test harness will
be triggered.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/events/pubsub_test.go
+++ b/src/control/events/pubsub_test.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2022 Intel Corporation.
+// (C) Copyright 2020-2023 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -203,6 +203,7 @@ func TestEvents_PubSub_Debounce_NoCooldown(t *testing.T) {
 	defer test.ShowBufferOnFailure(t, buf)
 
 	ps := NewPubSub(test.Context(t), log)
+	defer ps.Close()
 
 	evt1 := mockSwimRankDeadEvt(1, 1)
 	debounceType := evt1.ID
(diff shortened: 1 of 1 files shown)


src/control/events/pubsub_test.go
```diff
@@ -35,7 +35,9 @@ func (tly *tally) OnEvent(_ context.Context, evt *RASEvent) {
 	tly.Lock()
 	defer tly.Unlock()
 
-	tly.rx = append(tly.rx, evt.String())
+	if len(tly.rx) < tly.expectedRx {
+		tly.rx = append(tly.rx, evt.String())
+	}
```
mjmac, line 40: Isn't this just changing the test to make it pass vs. verifying that the `OnEvent()` method is only called the expected number of times?

tanabarr: Yes you are right, I will have to study and figure out the debounce code bug, (I didn't write that part)

kjacque (changes requested):
Agreed with Mike's comment. I think we need changes to address the underlying issue.

src/control/events/pubsub_test.go
```diff
@@ -234,32 +235,64 @@ func TestEvents_PubSub_Debounce_NoCooldown(t *testing.T) {
...
+			log.Debugf("start loop %d", i+1)
+
+			e := mockSwimRankDeadEvt(1, uint32(i+1))
+			t := time.Now()
+			for j := 0; j < 16; j++ {
+				l := time.Since(t)
+				if l >= debounceCooldown {
+					// Test loop was stalled, print warning
+					log.Noticef("test loop stalled for %s, restart iteration", l)
+					return true
```
knard-intel, line 268: Not sure to understand why you return true.
Maybe continue could be more suited ?

tanabarr: The "test" function returns true to indicate a restart which is detected on L294 and the test counter is decremented so the whole test runs again (resulting in teardown and setup). When a false expiry is triggered by the VM lockup all the test results are invalid so it should restart from the beginning. This is an acceptable solution because the false expiry seems to occur only once in every thousand test iterations or less.

knard-intel: Thanks for the clarification :)

knard-intel (approved):
LGTM

src/control/events/pubsub_test.go
```diff
@@ -234,32 +235,64 @@ func TestEvents_PubSub_Debounce_NoCooldown(t *testing.T) {
 }
 
 func TestEvents_PubSub_Debounce_Cooldown(t *testing.T) {
-	log, buf := logging.NewTestLogger(t.Name())
-	defer test.ShowBufferOnFailure(t, buf)
+	test := func(t *testing.T, iter int) bool {
```
mjmac, line 238: minor: Avoid variable names that could conflict with package names. We already have an imported test package. This obviously works OK as-is because the compiler figures it out, but as a general rule we should try to avoid confusion.

mjmac (approved):
Looks good overall. Only suggestion I have is that it would probably be good to add a comment in the test explaining the need for the retry logic.

kjacque (approved):

kjacque (approved):

src/control/events/pubsub_test.go
tanabarr: @mjmac I will address these comments in a follow-on.

PR #12242, opened 2023-05-31
DAOS-13047 chk: properly handle start options

Nasf-Fan:
Test-tag-hw-medium: Pass1Test

The start options for new instance should properly overwrite old ones. It also enhances CR pass 1 functionality tests:

1. Verify DAOS checker under dryrun mode.

2. Verify DAOS checker under auto mode.

3. Verify that the DAOS checker can detect and repair the inconsistency under regular mode.

4. Fix some typo for the test tags.

5. More accurate comment for test flow.

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate watchers.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
  * [ ] Quick-build or Quick-functional is not used.
* [ ] Fix the commit message upon landing. Check the standard [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments). Edit it to create a single commit. If necessary, ask submitter for a new summary.

PR #12294, opened 2023-06-07
DAOS-13510 chk: Query output should include dry-run status

mjmac:
If the checker is running in dry-run mode, notify the user.

Required-githooks: true


Nasf-Fan:
> @Nasf-Fan: Please take a look to see if this is what you had in mind. The only issue that I can see is that there does not appear to be a way to get out of dry-run mode. If I run `dmg check start --dry-run`, then `dmg check start --reset`, the checker seems to stay in dry-run mode. I have also tried running `dmg check stop && dmg check start`, but the flag is persistent in the checker properties, it seems.
> 
> I think probably we need to clear the property flag if the user did not specify it with the start command. I tried to implement that, but ran into some trouble. Perhaps you can suggest a solution that I can include with this patch, thanks.

Does MS store the "dryrun" flag somewhere? If not, then when you start checker next time without "--dry-run", then the former "dryrun" will be automatically removed. If it does not work, then it is daos engine bug, and I need to fix it. Would you please to verify the "dryrun" behavior against the patch https://github.com/daos-stack/daos/pull/12242 ? In such patch, we start checker with "dryrun" firstly, and then without "dryrun" in subsequent "check start". According to the CI test result, it works.

Nasf-Fan:
As my understand, MS can have more simple way to know whether current check instance is running under "dryrun" mode or not. For example, when user input "dmg check query", then MS can handle that in two steps:
1. Firstly, trigger DRPC_METHOD_MGMT_CHK_PROP to DAOS engine, then it will get current checker start options, that includes the "dryrun" if have.
2. Then, trigger DRPC_METHOD_MGMT_CHK_QUERY to DAOS engine as you orignal "dmg check query" does.

With the firstly step checker property information, you can know whether current checker is running under dryrun mode or not. That will much simplify the patch. How do you think?

tanabarr (approved):

Nasf-Fan (changes requested):
@kjacque, as my understand, MS can have more simple way to know whether current check instance is running under "dryrun" mode instead of changing the API just for the dryrun flag. For example, when user input "dmg check query", then MS can handle that in two steps:

Firstly, trigger DRPC_METHOD_MGMT_CHK_PROP to DAOS engine, then it will get current checker start options, that includes the "dryrun" flag if have.

Then, trigger DRPC_METHOD_MGMT_CHK_QUERY to DAOS engine as you orignal "dmg check query" does.

With the firstly step checker property information, you can know whether current checker is running under dryrun mode or not. That will much simplify current patch.

PR #12300, opened 2023-06-07
DAOS-12541 ddb: Improved List Output

ryon-jensen:
Replaced the "vos path builder" with "indexed tree path" structure. It is used for parsing and printing vos paths. The 'ls' command is updated to print the paths now to make it possible to copy/paste the paths. The indexed_tree_path better supports dkey and key values by using a syntax for specifying type (example: '{uint32: 0x123abc}') and length of a string (example: 'akey{5}'). Can also escape special characters using the '\' in keys.

Required-githooks: true

### Gatekeeper:


ryon-jensen:
Functional test is still failing, but I want to start getting some eyes on this.

shimizukko (approved):

src/tests/ftest/recovery/ddb.py
```diff
@@ -23,10 +23,15 @@ class DdbTest(TestWithServers):
...
         super().__init__(*args, **kwargs)
+        # how many objects and keys to insert/expect
+        self.object_count = 5
+        self.dkey_count = 2
+        self.akey_count = 1
         # Generate random keys and data to insert into the object.
         self.random_dkey = get_random_string(10)
         self.random_akey = get_random_string(10)
         self.random_data = get_random_string(10)
+        self.standalone = False  # enables running test in a "stand alone" mode
```
daltonbohning, line 34: Is this for future work? It seems always `False`

src/tests/ftest/recovery/ddb.py
```diff
@@ -83,26 +88,35 @@ def test_recovery_ddb_ls(self):
...
         ls_out = "\n".join(cmd_result[0]["stdout"])
-        match = re.search(r"\[\d+\]\s+([a-f0-9\-]+)", ls_out)
+        # Matches the container uuid
+        uuid_regex = r"([0-f]{8}-[0-f]{4}-[0-f]{4}-[0-f]{4}-[0-f]{12})"
+        match = re.search(uuid_regex, ls_out)
+        if match is None:
+            self.log.info("##### Errors #####")
+            report_errors(test=self, errors=["Unexpected output from command, unable to parse."])
+            self.log.info("##################")
+            return  # don't continue
```
daltonbohning, line 119: Suggestion - `report_errors` will call `test.fail`, which ends the test immediately, so `return` isn't necessary. Also, since there is only one error, I'd recommend simplifying to something like
```suggestion
        if match is None:
            self.fail("Unexpected output from ddb command, unable to parse.")
```

src/tests/ftest/recovery/ddb.py
```diff
@@ -143,28 +157,16 @@ def test_recovery_ddb_ls(self):
...
-                       "Actual = {}").format(obj_index, self.random_dkey, actual_dkey_2)
-
-            # Verify the dkey size field.
-            for dkey in match:
-                dkey_string = dkey[1]
-                # int(dkey[2]) is the dkey size; avoid pylint error.
-                if len(dkey_string) != int(dkey[2]):
-                    msg = ("Wrong dkey size! obj_index = {}. String = {}; "
-                           "Size = {}").format(obj_index, dkey_string, int(dkey[2]))
+            for i in range(self.dkey_count):
```
daltonbohning, line 165: > GitHub Actions / Pylint check
>
> invalid-name, Variable name "i" doesn't conform to snake_case naming style

Variable names have to be at least 3 characters long
```suggestion
            for idx in range(self.dkey_count):
```

daltonbohning (dismissed):
-1 for the introduced pylint

In case you aren't familiar, you can see these either inline on the "Files changed" tab, or you can see the checks that failed in the "Checks" tab

src/tests/ftest/recovery/ddb.py
ryon-jensen: This was so I could iterate over the test on an existing vos file without test needing to create it each time, So in a temporary script I could do something like:  
```
test = DdbTest(methodName='test_recovery_ddb_ls')
test.standalone = True
(...)
test.test_recovery_ddb_ls()
```

I'm open to suggestions or can just remove it all together.

src/tests/ftest/recovery/ddb.py
ryon-jensen: Thanks ... will update.

src/tests/ftest/recovery/ddb.py
daltonbohning: That's an interesting use case - one which we don't consider in the framework. The problem with using `self.standalone` like this is that it won't get covered by most test changes, so will eventually break. I.e. I don't imagine anyone is going to test that when making changes. I'd be in favor removing it, if not too inconvenient.
To support your use case while not introducing unused code paths, I think the simplest approach would be to break the test method into helper functions such that you can call the helper functions as needed. But that's a pretty big re-org of this test.

daltonbohning (commented):
ftest LGTM, other than the recommendation to remove `self.standalone`

shimizukko (approved):

src/tests/ftest/recovery/ddb.py
ryon-jensen: I'll remove "standalone". Thanks.

src/control/cmd/ddb/main.go
```diff
@@ -122,7 +122,33 @@ a file in the VOS format. It offers both a command line and interactive
 shell mode. If neither a single command or '-f' option is provided, then
 the tool will run in interactive mode. In order to modify the VOS file,
 the '-w' option must be included. If supplied, the VOS file supplied in
-the first positional parameter will be opened before commands are executed.`
+the first positional parameter will be opened before commands are executed.
+
+Many of the commands take a vos tree path. The format for this path
+is [cont]/[obj]/[dkey]/[akey]/[extent].
```
kjacque, line 128: Just wondering how a user would know these paths. I think this is fine for the time being, but I'm curious about the use cases for the tool.

kjacque (approved):
Go code looks good. I didn't perform an in-depth review of the C changes.

tanabarr (approved):
Go changes LGTM, C changes look okay but I'm unable to give a thorough review as it's an unfamiliar area for me.

src/control/cmd/ddb/main.go
ryon-jensen: One of the commands is "ls" which lists the paths. It would be used to determine which paths the other commands might want to be run on.

ryon-jensen:
I went through the clang-format changes proposed and they're mostly changes from tabs to spaces which would make the changed lines inconsistent with the other lines around it.

daltonbohning:
Removing gatekeeper until this is passing / ready for landing. Please re-add when needed

src/ddb/ddb.c
```diff
@@ -754,7 +758,9 @@ ddb_run_cmd(struct ddb_ctx *ctx, const char *cmd_str, bool write_mode)
 	struct argv_parsed	 parse_args = {0};
 	struct ddb_cmd_info	 info = {0};
 	int			 rc;
-	char			*cmd_copy = strdup(cmd_str);
+	char                    *cmd_copy;
+
+	D_STRNDUP(cmd_copy, cmd_str, strlen(cmd_str));
```
ashleypittman, line 763: You need to check the result of this.

ryon-jensen:
Test failure is: https://daosio.atlassian.net/browse/DAOS-13998

PR #12311, opened 2023-06-08
DAOS-13627 build: Enable and fix issues with -Wextra.

ashleypittman:
Make use of the static analysis that modern compiers can
provide and fix issues in the codebase.
Leave a couple of warnings disabled that result in multiple errors

Required-githooks: true


src/container/srv_container.c
```diff
@@ -4217,7 +4217,7 @@ cont_filter_match(struct rdb_tx *tx, struct cont *cont, daos_pool_cont_filter_t
 	int		rc = 0;
 
 	/* logical OR combining: start with false result, transition to true on first match */
-	if ((filt->pcf_parts > 0) && (combine_op == PCF_COMBINE_LOGICAL_OR))
+	if ((filt->pcf_parts != 0) && (combine_op == PCF_COMBINE_LOGICAL_OR))
```
ashleypittman, line 4220: This one I'm not sure about but pcf_parts appears to be a pointer, in which case I think this is the correct change.

src/include/daos_obj_class.h
ashleypittman, line 1: This file was interesting, we have a couple of functions that take `enum daos_obj_redun` as an option and they only accepted the core types as the complex ones were a different, anonymous enum.  This change makes all values the same type of enum.

.github/workflows/landing-builds.yml
```diff
@@ -18,6 +18,7 @@ on:
 
 jobs:
 
+
```
ashleypittman, line 21: Whitespace change to trigger landings testing for this PR to get full compiler coverage.  I can remove before landing.

src/container/srv_container.c
daltonbohning: If it's a pointer, then should it check NULL instead of 0?

daltonbohning (dismissed):

src/container/srv_container.c
ashleypittman: Will change (and the same change in dfs) when I re-push to remove the change to landings.yaml.

ashleypittman:
This just fixes the issues for now but doesn't enable the flag, that is blocked by the changes @daltonbohning is making to fix some coverity issues.

jolivier23 (dismissed):
Why is -Wextra not added anywhere?  Or is this just making it possible to use it?

jolivier23:
> Why is -Wextra not added anywhere? Or is this just making it possible to use it?

nevermind, just saw the comment above.

.github/workflows/landing-builds.yml
jolivier23: Does it not trigger testing if you just do --allow-empty commit ?

ashleypittman: GitHub actions only tests the full matrix of compilers for landings builds, not PRs.

ashleypittman:
https://github.com/daos-stack/daos/pull/12319 is the other PR that's needed to enable the flag.

daltonbohning:
> #12319 is the other PR that's needed to enable the flag.

Just merged

jolivier23 (dismissed):

daltonbohning (dismissed):

daltonbohning (approved):

jolivier23 (approved):

PR #12312, opened 2023-06-08
DAOS-623 gurt: Make daos error numbers a named enum and remove dead code.

ashleypittman:
Newer versions of compilers can perform type checking of enums
so put daos error numbers in a named enum.

Remove some dead code for adding error numbers dynamically.

Required-githooks: true

### DIFF
--- a/src/include/daos_errno.h
+++ b/src/include/daos_errno.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -271,6 +271,7 @@ extern "C" {
 	DER_ERR_##name##_BASE = (base),                                                            \
 	D_FOREACH_##name##_ERR(D_DEFINE_ERRNO) DER_ERR_##name##_LIMIT,
 
+/** The actual error codes */
 enum daos_error_number {
 	/** Return value representing success */
 	DER_SUCCESS = 0,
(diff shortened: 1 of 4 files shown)


jolivier23 (approved):
Typo in the title (s/nunbers/numbers)

daltonbohning (approved):

soumagne (approved):

src/include/daos_errno.h
```diff
@@ -267,79 +267,37 @@ extern "C" {
...
-/** Macro to deregister a range defined using D_DEFINE_RANGE macros */
-#define D_DEREGISTER_RANGE(name)			\
-	d_errno_deregister_range(DER_ERR_##name##_BASE)
-
-/** Return value representing success */
-#define DER_SUCCESS	0
-/** Unknown error */
-#define DER_UNKNOWN	(DER_ERR_GURT_BASE + 500000)
+/** The actual error codes */
+enum daos_error_number {
```
soumagne, line 275: minor detail but was there a reason for not naming it `daos_errno` ?

src/include/daos_errno.h
```diff
@@ -267,79 +267,37 @@ extern "C" {
...
- */
-int d_errno_register_range(int start, int end,
-			   const char * const *error_strings,
-			   const char * const *strerror);
-
-/** De-register error codes with gurt.  Use D_DEREGISTER_RANGE.
- *
- * \param[in]	start	Start of error range
- */
-void d_errno_deregister_range(int start);
```
mchaarawi, line 333: that's a major API version change, no?

ashleypittman: Technically yes, or at least it would be if it was daos rather than gurt.

These functions were added to cart at a time when daos and gurt were separate, I don't think there's ever been a user of them since the two were merged.

mchaarawi: i don't think there is either. but at some point we have to stop assuming that.

jolivier23: Probably still can add a patch to deprecate them in 2.4

PR #12331, opened 2023-06-09
DAOS-12751 control: Add a daos filesystem evict command.

ashleypittman:
Add a command to tell dfuse and the kernel to evict a
path in dfuse.  This is implemented via a ioctl that
sets a flag on the file handle, when the file is then
closed the kernel is then told the dentry for the file
is no longer valid which should cause the kernel to
forget the entry and any contents if the entry is
a directory.

Required-githooks: true


src/client/dfuse/ops/open.c
```diff
@@ -152,5 +153,15 @@ dfuse_cb_release(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
...
 		DFUSE_REPLY_ERR_RAW(oh, req, rc);
+
+	if (oh->doh_evict_on_close) {
+		rc = fuse_lowlevel_notify_inval_entry(fs_handle->dpi_info->di_session,
+						      oh->doh_ie->ie_parent, oh->doh_ie->ie_name,
+						      strnlen(oh->doh_ie->ie_name, NAME_MAX));
+
+		if (rc != 0)
+			DFUSE_TRA_ERROR(oh->doh_ie, "inval_entry() returned: %d (%s)", rc,
+					strerror(-rc));
```
ashleypittman, line 164: I think this is actually unsafe, once DFUSE_REPLY_ENTRY is called there's no longer a reference on the inode so I will take copies of this before landing.

src/control/cmd/daos/filesystem.go
```diff
@@ -37,6 +37,7 @@ type fsCmd struct {
 	ResetAttr      fsResetAttrCmd      `command:"reset-attr" description:"reset fs attributes"`
 	ResetChunkSize fsResetChunkSizeCmd `command:"reset-chunk-size" description:"reset fs chunk size"`
 	ResetObjClass  fsResetOclassCmd    `command:"reset-oclass" description:"reset fs obj class"`
+	DfuseEvict     fsDfuseEvictCmd     `command:"evict" description:"Evict object from dfuse"`
```
daltonbohning, line 40: If this is ONLY for dfuse - not DFS - maybe `evict` -> `dfuse-evict`?
Or maybe it's okay if it only makes sense for dfuse anyway

daltonbohning (commented):
Maybe I misunderstand the ioctl stuff, but how would this work in the case of evicting a _**pool**_ or _**container**_ handle? For example, to pick up ACL changes

src/control/cmd/daos/filesystem.go
ashleypittman: I think the only thing al these commands have in common is they're for POSIX containers and therefore DFS at some level.  `evict` doesn't mean anything in dfs directly that I'm aware of.

src/control/cmd/daos/filesystem.go
```diff
@@ -448,31 +451,109 @@ func (cmd *fsDfuseQueryCmd) Execute(_ []string) error {
...
+	cmd.Infof(" Containers: %d", ap.dfuse_mem.container_count)
+	cmd.Infof("     Inodes: %d", ap.dfuse_mem.inode_count)
+	cmd.Infof(" Open files: %d", ap.dfuse_mem.fh_count)
+	if cmd.Ino != 0 {
+		if ap.dfuse_mem.found {
+			cmd.Infof(" Inode %#lx known", cmd.Ino)
+		} else {
+			cmd.Infof(" Inode %#lx not known", cmd.Ino)
+		}
+	}
```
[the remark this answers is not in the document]
ashleypittman: This is not an error either way, it's in effect polling for the operation to have completed.  Maybe the text in this output is badly worded but neither of these is an error.  See the changes to filesystem.md in this PR for a full explanation.

src/control/cmd/daos/filesystem.go
mchaarawi: one thing i was thinking about here is that the daos fs options are on the container (server side), but these dfuse commands apply only on the client side (just the node they are running on). so they are just local operations and only executed on the node where this runs.
i don't know if that counts as grounds that we need to split this off into 2 tools or just have an extension to the dfuse command, or just keep it as it is. what do you think?

ashleypittman: We discussed this extensively in the working group calls, this was your suggestion.

I'd be fine with making a new subcommand for the `daos` command other than `daos filesystem` but I don't want to do it in this PR.

ashleypittman: I agree though, there's a clear distinction between commands that have an effect on the container itself vs commands that change local in-memory state.

docs/user/filesystem.md
```diff
@@ -701,6 +701,30 @@ These are two command line options to control the DFuse process itself.
 
 These will affect all containers accessed via DFuse, regardless of any container attributes.
 
+### Managing memory usage and disconnecting from containers
+
+DFuse can be instructed to evict paths from local memory which drops and open handles on containers
+or pools as well as reducing the working set size and memory consumption.  This is an asynchronous
+operation and there is no automatic way to tell if it's completed, in addition any lookup of the
+path specified in the eviction call will cause a new lookup and prevent the eviction from
+completing.
```
daltonbohning, line 710:
```suggestion
DFuse can be instructed to evict paths from local memory which drops any open handles on containers
or pools as well as reducing the working set size and memory consumption.  This is an asynchronous
operation and there is no automatic way to tell if it's completed. In addition, any lookup of the
path specified in the eviction call will cause a new lookup and prevent the eviction from
completing.
```

docs/user/filesystem.md
```diff
@@ -701,6 +701,30 @@ These are two command line options to control the DFuse process itself.
...
+
+Paths can be requested for eviction from dfuse using the `daos filesystem evict` command, this does
+not change any data that is stored in DAOS in any way but rather releases local resources.  This
+command will return the inode number of the path as well as key dfuse metrics.
```
daltonbohning, line 714:
```suggestion
Paths can be requested for eviction from dfuse using the `daos filesystem evict` command. This does
not change any data that is stored in DAOS in any way but rather releases local resources.  This
command will return the inode number of the path as well as key dfuse metrics.
```

docs/user/filesystem.md
```diff
@@ -701,6 +701,30 @@ These are two command line options to control the DFuse process itself.
...
+
+DFuse metrics can be queried with the `daos filesystem query` command which takes an optional
+`--inode` parameter.  This will return information on the number of inodes held in memory, the
+number of open files as well as the number of pools and containers that DFuse is connected to.  If
+the `--inode` option is given then this command will also report if the inode is in memory or not.
+
+Together these two commands can be used to request eviction of a path and to poll for it's release,
+although lookups from other processes might block the eviction process.
```
daltonbohning, line 722:
```suggestion
Together these two commands can be used to request eviction of a path and to poll for its release,
although lookups from other processes might block the eviction process.
```

docs/user/filesystem.md
```diff
@@ -701,6 +701,30 @@ These are two command line options to control the DFuse process itself.
...
+
+If `daos filesystem evict` is passed the root of the DFuse mount then the path itself cannot be
+evicted, in this case all top-level entries in the directory are evicted instead and no inode
+number is returned.
```
daltonbohning, line 726:
```suggestion
If `daos filesystem evict` is passed the root of the DFuse mount then the path itself cannot be
evicted - in this case all top-level entries in the directory are evicted instead and no inode
number is returned.
```

src/utils/daos_hdlr.c
```diff
@@ -2442,13 +2439,47 @@ cont_clone_hdlr(struct cmd_args_s *ap)
...
+	}
+
+	query.ino = ap->dfuse_mem.ino;
+
+	rc = ioctl(fd, DFUSE_IOCTL_COUNT_QUERY, &query);
+	if (rc < 0) {
+		rc = daos_errno2der(errno);
+		DH_PERROR_DER(ap, rc, "ioctl failed");
+		goto close;
+	}
```
daltonbohning: Should these error prints be consistent? The first prints an `errno` and the second prints a `D_ERR`

utils/node_local_test.py
```diff
@@ -1449,6 +1451,35 @@ def check_usage(self, inodes=None, open_files=None, pools=None, containers=None)
...
+        for inode in inodes:
+            found = True
+            while found:
+                rc = self.check_usage(inode)
+                print(rc)
+                found = rc['resident']
+                if not found:
+                    time.sleep(1)
+                    sleeps += 1
+                    assert sleeps < 10, 'Path still present 10 seconds after eviction'
```
daltonbohning, line 1481: 1. Shouldn't sleeps be reset to 0 for each inode?
2. Maybe move the assert before `time.sleep`? Otherwise we sleep for 1 second, don't check the path again, and then assert

ashleypittman: 1, no, this is a lazy way of doing a timeout.
2, yes, the actual sleep should be last here.  Will fix.

src/utils/daos_hdlr.c
ashleypittman: Yes they should.

ashleypittman: Somehow you're commenting on an old commit within the PR but yes, I've updated all these to be consistent and fixed the `query` command as well.

utils/node_local_test.py
daltonbohning: 1. So ~10 seconds total for all paths?

daltonbohning (approved):

wiliamhuang:
Minor issues,
https://github.com/daos-stack/daos/blob/amd/dfuse-evict/src/client/dfuse/ops/ioctl.c#L329
"struct dfuse_mem_query        query      = {};" can be removed.

https://github.com/daos-stack/daos/blob/amd/dfuse-evict/src/client/dfuse/ops/ioctl.c#L333
"if (in_bufsz != sizeof(query))" --> "if (in_bufsz != sizeof(struct dfuse_mem_query))"

ashleypittman:
> Minor issues, https://github.com/daos-stack/daos/blob/amd/dfuse-evict/src/client/dfuse/ops/ioctl.c#L329 "struct dfuse_mem_query query = {};" can be removed.
> 
> https://github.com/daos-stack/daos/blob/amd/dfuse-evict/src/client/dfuse/ops/ioctl.c#L333 "if (in_bufsz != sizeof(query))" --> "if (in_bufsz != sizeof(struct dfuse_mem_query))"

Yes, you're right.  I've got more PRs to follow in this series, I can fix this up later.

wiliamhuang (approved):

PR #12346, opened 2023-06-09
DAOS-13595 utils: Check return codes

kjacque:
Check return codes for retried functions in daos_hdlr disconnect code.

Required-githooks: true

### Gatekeeper:


ashleypittman (dismissed):
I wish there was a better way to do this, these are only retried for the fault injection testing.

Checking return code and logging on failure helps but it doesn't change the fundamentals that the return code doesn't change the code paths taken.

mchaarawi:
> I wish there was a better way to do this, these are only retried for the fault injection testing.
> 
> Checking return code and logging on failure helps but it doesn't change the fundamentals that the return code doesn't change the code paths taken.

mchaarawi (approved):
sorry i accidentally closed the PR but reopened it.
lgtm. although this is really just a result of the flaw of how FI testing is done.

knard-intel (dismissed):
LGTM

kjacque:
Added some fixes for the NLT failures. The issue seemed to be one where pool disconnect reported failure when the actual disconnect succeeded, but the notification to the agent failed. Since that error is supposed to be nonfatal, I don't think we should return a failure for it.

kjacque:
Since this code is well-exercised by NLT, I don't think any Feature tag is required for this PR.

src/pool/cli.c
```diff
@@ -770,6 +770,7 @@ pool_disconnect_cp(tse_task_t *task, void *data)
 		 */
 		D_ERROR("failed to notify agent of pool disconnect: "DF_RC"\n",
 			DP_RC(rc));
+		rc = 0;
```
mchaarawi, line 773: it doesn't make sense to me to log an error that is not an error on the operation. so maybe change to warn?
maybe another question here, are all errors non-fatal or some? should we fail on some errors, but ignore others? not sure if it's a good idea to ignore all errors..

src/client/dfuse/dfuse_core.c
```diff
@@ -270,18 +270,6 @@ _ph_free(struct dfuse_pool *dfp)
...
 	if (daos_handle_is_valid(dfp->dfp_poh)) {
 		rc = daos_pool_disconnect(dfp->dfp_poh, NULL);
-		/* Hook for fault injection testing, if the disconnect fails with out of memory
-		 * then simply try it again, only what might have happened is that the first
-		 * call might have disconnected, but then failed to notify about the disconnect,
-		 * in which case the subsequent call will return -DER_NO_HDL, if that's the case
-		 * then this is expected, if odd, behavior so silence that case and just return
-		 * success.
-		 */
-		if (rc == -DER_NOMEM) {
```
mchaarawi, line 280: wouldn't this change break FI testing?

mchaarawi (changes requested):

src/pool/cli.c
ashleypittman, line 773: we typically use a 'int rc2' variable for cases like this.

src/client/dfuse/dfuse_core.c
```diff
@@ -270,18 +270,6 @@ _ph_free(struct dfuse_pool *dfp)
...
-			rc = daos_pool_disconnect(dfp->dfp_poh, NULL);
```
ashleypittman, line 281: I think you need this bit but not the next two lines.

src/client/dfuse/dfuse_core.c
kjacque: No, because in this case the underlying behavior issue is fixed.

src/pool/cli.c
kjacque: I could change to warn, although I do think it's useful to log at error, since it is a sign something is wrong with the daos_agent or with the client machine more generally (e.g. out of memory).

What happened here is the pool disconnect in fact succeeded. After we successfully disconnect a pool handle, we also send a notification to the daos_agent to update its local process monitor. That is the operation that fails here. At this point, the pool handle is gone on the server side, too, so the operation succeeded regardless of why the agent notify operation failed. Retrying the daos_pool_disconnect call will report -DER_NO_HDL.

For that reason I think it's okay to log the errors here and then ignore them. The side effect, as noted in the code comment, is that the agent will contact the server when the client process terminates, asking the server to clean up the already-gone handles that the agent thinks are still around. For the server this is a little inefficiency, as there will be nothing left to clean up. I guess in theory this could impact performance, which may be a good reason to log at error level for the client.

src/client/dfuse/dfuse_core.c
kjacque: As Ashley pointed out below, you are right after all. I thought its intent was only to fix this specific odd case with the agent notification.

src/client/dfuse/dfuse_core.c
ashleypittman: Actually I'm not sure myself now as this code didn't appear to actually call pool_disconnect a second time, just adjust the rc value.  Lets see what testing returns.

kjacque: Looks like fault injection passed as-is.

mchaarawi (approved):

ashleypittman (approved):

tanabarr (approved):

kjacque:
FI failed again. Sorry for the confusion before -- seems it is only marked as failed after all tests in the Test stage have finished, not when the FI test itself finishes. I'll add the fix we discussed in the comments.

tanabarr (approved):

ashleypittman (approved):
Looks good so far but we need to see the CI results.

knard-intel (approved):

kjacque:
@daos-stack/daos-gatekeeper Can we get this force landed please? This is a required 2.4 issue. I'd like to get started on backporting.

mchaarawi (approved):

PR #12367, opened 2023-06-12
DAOS-13658 dfuse: Add filesystem query command.

ashleypittman:
Use atomics to track number of pool and containers open, as well
as inodes and file handles.

Required-githooks: true


src/control/cmd/daos/filesystem.go
```diff
@@ -428,3 +429,49 @@ func (cmd *fsFixRootCmd) Execute(_ []string) error {
...
+	defer freeString(ap.path)
+	defer deallocCmdArgs()
+
+	rc := C.dfuse_count_query(ap)
+	if err := daosError(rc); err != nil {
+		return errors.Wrapf(err, "failed to query %s", cmd.Path)
+	}
+
+	if cmd.jsonOutputEnabled() {
+
```
tanabarr, line 455: nit: unnecessary newline

tanabarr (approved):
go changes look good to me

tanabarr (commented):
removing approval for others who can review the whole PR

src/control/cmd/daos/filesystem.go
ashleypittman: Will fix on next push.

src/client/dfuse/ops/ioctl.c
```diff
@@ -297,6 +297,33 @@ handle_dooh_ioctl(struct dfuse_obj_hdl *oh, size_t size, fuse_req_t req)
...
+static void
+handle_cont_query_ioctl(fuse_req_t req, const void *in_buf, size_t in_bufsz)
+{
+	struct dfuse_info            *dfuse_info = fuse_req_userdata(req);
+	struct dfuse_mem_query        query      = {};
+	const struct dfuse_mem_query *in_query   = in_buf;
+	int                           rc;
+
+	if (in_bufsz != sizeof(query))
+		D_GOTO(err, rc = EIO);
```
daltonbohning, line 309: Should this print a DEBUG? If it doesn't fail here, I imagine it would be tricky to track down

src/client/dfuse/ops/ioctl.c
```diff
@@ -297,6 +297,33 @@ handle_dooh_ioctl(struct dfuse_obj_hdl *oh, size_t size, fuse_req_t req)
...
+
+	/* Not supported yet, future-proofing for DAOS-12751 */
+	if (in_query->ino != 0)
+		D_GOTO(err, rc = EIO);
```
daltonbohning, line 313: same

src/control/cmd/daos/filesystem.go
```diff
@@ -428,3 +429,49 @@ func (cmd *fsFixRootCmd) Execute(_ []string) error {
 
 	return nil
 }
+
+type fsDfuseQueryCmd struct {
+	daosCmd
+
+	Path string `long:"path" description:"Path to evict from dfuse" required:"1"`
```
daltonbohning, line 436: The command is `memquery`, but this says `Path to evict`?

src/include/dfuse_ioctl.h
```diff
@@ -60,6 +59,15 @@ struct dfuse_user_reply {
...
 };
 
+struct dfuse_mem_query {
+	uint64_t inode_count;
+	uint64_t fh_count;
+	uint64_t pool_count;
+	uint64_t container_count;
+	ino_t    ino;
+	bool     found;
+};
```
daltonbohning, line 69: Should we future-proof ourselves by adding a `version` to this struct? Just thinking about what would happen if middlware / third-party compiled with this version of the struct, then recompiled a new version of DAOS without recompiling middleware. @mchaarawi I remember we ran into something similar with DFS before. Thoughts?

utils/node_local_test.py
```diff
@@ -2563,6 +2563,13 @@ def test_uns_create(self):
         print(stbuf)
         assert stbuf.st_ino < 100
         print(os.listdir(path))
+        rc = run_daos_cmd(self.conf, ['filesystem', 'memquery', '--path', self.dfuse.dir])
```
daltonbohning, line 2566: Should we support positional?
```
filesystem memquery <path>
```

src/client/dfuse/ops/ioctl.c
ashleypittman: DFUSE_REPLY_ERR_RAW() will do that.

src/control/cmd/daos/filesystem.go
ashleypittman: Will update.

src/include/dfuse_ioctl.h
ashleypittman: It would be possible to but I think we'd just switch to a new cmd number if we did that, ioctls are integer commands so don't have the same issues as c functions which are resolved by name.

utils/node_local_test.py
ashleypittman: Probably, I'll speak to @tanabarr about how to add that.

ashleypittman:
Here's the help text with the lastest push to this PR.
`
(wolf-63) [08:51] [0] ampittma@wolf-63:~/coral/daos $ daos filesystem memquery --help
Usage:
  daos [OPTIONS] filesystem memquery [path]

Application Options:
      --debug     enable debug output
      --verbose   enable verbose output (when applicable)
  -j, --json      enable JSON output

Help Options:
  -h, --help      Show this help message

[memquery command arguments]
  path:           DFuse path to query
(wolf-63) [08:53] [0] ampittma@wolf-63:~/coral/daos $ daos filesystem memquery
ERROR: daos: the required argument `path` was not provided
(wolf-63) [08:53] [1] ampittma@wolf-63:~/coral/daos $`

utils/node_local_test.py
```diff
@@ -2563,6 +2563,12 @@ def test_uns_create(self):
         print(stbuf)
         assert stbuf.st_ino < 100
         print(os.listdir(path))
+        rc = run_daos_cmd(self.conf, ['filesystem', 'memquery', self.dfuse.dir])
+        print(rc)
+        assert rc.returncode == 0
+        rc = run_daos_cmd(self.conf, ['filesystem', 'memquery', self.dfuse.dir], use_json=True)
+        print(rc)
+        assert rc.returncode == 0
```
mchaarawi, line 2571: i'd like to see a real test here that actually verifies the collected query stats.

src/client/dfuse/dfuse.h
```diff
@@ -56,6 +56,11 @@ struct dfuse_info {
 	/* Array of dfuse_eq */
 	struct dfuse_eq     *di_eqt;
 	ATOMIC uint64_t      di_eqt_idx;
+
+	ATOMIC uint64_t      di_inode_count;
+	ATOMIC uint64_t      di_fh_count;
+	ATOMIC uint64_t      di_pool_count;
+	ATOMIC uint64_t      di_container_count;
```
mchaarawi, line 63: i don't think memquery really associate with these counters. a user would think this tool is actually printing memory usage but those counters. so the tool name probably makes more sense to be called stat or probably just query

mchaarawi (approved):

daltonbohning (approved):

PR #12370, opened 2023-06-12
DAOS-7029 control: Add refresh methods for agent cache (feature/multiprovider)

kjacque:
These two commits required some significant changes to successfully merge from master. 

### Gatekeeper:

### DIFF
--- a/src/control/cmd/daos_agent/infocache.go
+++ b/src/control/cmd/daos_agent/infocache.go
@@ -8,6 +8,7 @@ package main
 
 import (
 	"context"
+	"net"
 	"strings"
 	"sync"
 	"time"
@@ -19,6 +20,7 @@ import (
 	"github.com/daos-stack/daos/src/control/lib/atm"
 	"github.com/daos-stack/daos/src/control/lib/cache"
 	"github.com/daos-stack/daos/src/control/lib/control"
+	"github.com/daos-stack/daos/src/control/lib/hardware"
 	"github.com/daos-stack/daos/src/control/lib/hardware/hwprov"
 	"github.com/daos-stack/daos/src/control/logging"
 )
@@ -34,19 +36,15 @@ type fabricScanFn func(ctx context.Context, providers ...string) (*NUMAFabric, e
 // NewInfoCache creates a new InfoCache with appropriate parameters set.
 func NewInfoCache(ctx context.Context, log logging.Logger, client control.UnaryInvoker, cfg *Config) *InfoCache {
 	ic := &InfoCache{
-		log:           log,
-		ignoreIfaces:  cfg.ExcludeFabricIfaces,
-		client:        client,
-		cache:         cache.NewItemCache(log),
-		getAttachInfo: control.GetAttachInfo,
-		fabricScan: func(ctx context.Context, provs ...string) (*NUMAFabric, error) {
-			scanner := hwprov.DefaultFabricScanner(log)
-			fis, err := scanner.Scan(ctx, provs...)
-			if err != nil {
-				return nil, err
-			}
-			return NUMAFabricFromScan(ctx, log, fis).WithIgnoredDevices(cfg.ExcludeFabricIfaces), nil
-		},
+		log:            log,
+		ignoreIfaces:   cfg.ExcludeFabricIfaces,
+		client:         client,
+		cache:          cache.NewItemCache(log),
+		getAttachInfo:  control.GetAttachInfo,
+		fabricScan:     getFabricScanFn(log, cfg, hwprov.DefaultFabricScanner(log)),
+		netIfaces:      net.Interfaces,
+		devClassGetter: hwprov.DefaultNetDevClassProvider(log),
+		devStateGetter: hwprov.DefaultNetDevStateProvider(log),
 	}
 
 	if cfg.DisableCache {
@@ -64,6 +62,16 @@ func NewInfoCache(ctx context.Context, log logging.Logger, client control.UnaryI
 	}
 
 	return ic
+}
+
+func getFabricScanFn(log logging.Logger, cfg *Config, scanner *hardware.FabricScanner) fabricScanFn {
+	return func(ctx context.Context, provs ...string) (*NUMAFabric, error) {
+		fis, err := scanner.Scan(ctx, provs...)
+		if err != nil {
+			return nil, err
+		}
+		return NUMAFabricFromScan(ctx, log, fis).WithIgnoredDevices(cfg.ExcludeFabricIfaces), nil
+	}
 }
 
 type cacheItem struct {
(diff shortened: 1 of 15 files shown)


src/control/cmd/daos_agent/infocache.go
```diff
@@ -8,176 +8,445 @@ package main
...
+	if sys == "" {
+		sys = build.DefaultSystemName
+	}
+	createItem := func() (cache.Item, error) {
+		c.log.Debugf("cache miss for %s", sysAttachInfoKey(sys))
+		return newCachedAttachInfo(c.attachInfoRefresh, sys, c.client, c.getAttachInfo), nil
+	}
+
+	item, release, err := c.cache.GetOrCreate(ctx, sysAttachInfoKey(sys), createItem)
+	defer release()
```
knard-intel, line 321: From what I understand, the item lock is needed to safely copy its content at line 326.
However, as it is not a deep copy, the content of cai.lastResponse could be shared by a different processes after the function return ?
I was expecting to use value instead of pointer to avoid this issue.
However, I am probably missing something as I am not a GO expert.

src/control/cmd/daos_agent/infocache.go
```diff
@@ -8,176 +8,445 @@ package main
...
+	createItem := func() (cache.Item, error) {
+		c.log.Debug("NUMAFabric cache miss")
+		if err := c.waitFabricReady(ctx, netDevClass); err != nil {
+			return nil, err
+		}
+		return newCachedFabricInfo(c.log, c.fabricScan), nil
+	}
+
+	item, release, err := c.cache.GetOrCreate(ctx, fabricKey, createItem)
+	defer release()
```
knard-intel, line 391: Same question as for line 321.

src/control/cmd/daos_agent/start.go
```diff
@@ -166,6 +156,9 @@ func (cmd *startCmd) Execute(_ []string) error {
 			case syscall.SIGUSR1:
 				cmd.Infof("Signal received.  Caught %s; flushing open pool handles", sig)
 				procmon.FlushAllHandles(ctx)
+			case syscall.SIGUSR2:
```
knard-intel, line 159: This management of this POSIX signal should be probably documented somewhere such as in the new section of the daos_agent.yml file.

src/control/lib/cache/cache.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// Keys returns a sorted list of all keys in the cache.
+func (ic *ItemCache) Keys() []string {
+	if ic == nil {
+		return nil
+	}
+
+	keys := []string{}
+	for k := range ic.items {
```
knard-intel, line 98: From what I know range is not thread safe and thus a lock (or using a channel to access keys) should be needed ?

knard-intel (commented):
Mostly OK for me.
Just some small points to clarify.

src/control/lib/cache/cache.go
tanabarr: +1, looks like it needs RLock like `Has()` method above.

src/control/lib/cache/cache.go
```diff
@@ -0,0 +1,220 @@
...
+		ic.log.Debugf("failed to get item for key %q: %s", key, err.Error())
+		item, err = missFn()
+		if err != nil {
+			return nil, noopRelease, errors.Wrapf(err, "create item for %q", key)
+		}
+		ic.log.Debugf("created item for key %q", key)
+		ic.set(item)
+	}
+
+	if item.NeedsRefresh() {
```
tanabarr, line 147: don't we need to lock the item before calling refresh to avoid concurrent access when refreshing the item?

tanabarr: I was thinking that responsibility could be delegated to the refresh implementation but the underlying implementation wouldn't be aware of the Item interface's Lock() method. I would have thought that NeedsRefresh() requires a RLock and Refresh requires a Lock.

src/control/cmd/daos_agent/infocache.go
```diff
@@ -8,176 +8,445 @@ package main
...
+	return fabricKey
+}
+
+// NeedsRefresh indicates that the fabric information does not need to be refreshed unless it has
+// never been populated.
+func (cfi *cachedFabricInfo) NeedsRefresh() bool {
+	if cfi == nil {
+		return false
 	}
+	return !cfi.isCached()
```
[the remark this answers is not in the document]
kjacque: This was a discussion I had with @mjmac when we were first discussing this feature for the master branch. In practice it is unlikely that the local fabric devices will change once we've initially detected them. The main issue the cache refresh is trying to address is when GetAttachInfo has changed (could mean ranks have gone down and some of the URIs returned are no longer good). I allowed the force-refresh to impact fabric scans, in the event that there is some hotplug case, for example, but didn't allow it to be auto-refreshed. It's almost always unnecessary, and the fabric scans can chew up some cycles.

src/control/cmd/daos_agent/infocache.go
kjacque: Hm, at first I was thinking that this concern wasn't relevant, since there is a single user of the InfoCache. But as I consider it, there are potentially issues with several clients requesting GetAttachInfo at the same time. Depending on the system hardware, each of those clients might actually be assigned a different network interface. With the pointer passed back, this could be a data race. I'd like to address this in a PR to master, which will eventually trickle back to this branch.

src/control/lib/cache/cache.go
kjacque: Good catch.

src/control/lib/cache/cache.go
kjacque: Currently the Item implementations are taking care of this under the covers. Agreed that it feels confusing at the higher level, though. Since there are other issues I'd like to revisit, I'll look at this at the same time.

kjacque:
Since this was intended as a simple backport to allow me to merge master into feature/multiprovider, I've created a ticket to address the issues noted, since they are also present in master: https://daosio.atlassian.net/browse/DAOS-13696

@tanabarr @knard-intel Are you okay with me landing this to the feature/multiprovider branch as-is? This will unblock me from merging master into the branch.

tanabarr (approved):

src/control/cmd/daos_agent/infocache.go
knard-intel: OK for me

src/control/cmd/daos_agent/start.go
knard-intel: @Michael-Hennecke could probably have some good advices for the better location.
However, the admin guide seems a good location for me.

PR #12373, opened 2023-06-13
DAOS-13569 gurt: limit log size more accurate

wangshilong:
if we merge stderr to log file which is not
calculated by log_size, to avoid exceeding threshold too much, log_size will be updated with fstat if log size increased by 2% of max size every time.

Test-tags: daos_test
Test-nvme: auto_md_on_ssd
Required-githooks: true

### DIFF
--- a/src/gurt/dlog.c
+++ b/src/gurt/dlog.c
@@ -68,6 +68,8 @@ struct d_log_state {
 	int		 log_old_fd;
 	/** current size of log file */
 	uint64_t	 log_size;
+	/** log size of last time check */
+	uint64_t	 log_last_check_size;
 	/** max size of log file */
 	uint64_t	 log_size_max;
 	/** Callback to get thread id and ULT id */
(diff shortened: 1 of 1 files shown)


wiliamhuang (approved):

bfaccini (approved):

wangshilong:
@daos-stack/daos-gatekeeper failures should be not related to this PR.

src/gurt/dlog.c
```diff
@@ -405,7 +434,7 @@ d_log_write(char *msg, int len, bool flush)
 	if (mst.log_buf_nob == 0)
 		return 0; /* nothing to write */
 
-	if (mst.log_size + mst.log_buf_nob >= mst.log_size_max) {
+	if (log_exceed_threshold()) {
```
jolivier23, line 437: minor nit, s/exceed/exceeds.   As it's written, it sounds like "make log exceed threshold" whereas using exceeds sounds like the check it's doing.

jolivier23 (approved):

PR #12405, opened 2023-06-15
DAOS-13380 engine: refine tgt_nr check

liuxuezhao:
1. for non-DAOS_TARGET_OVERSUBSCRIBE case fail to start engine if #cores is not enough
2. for DAOS_TARGET_OVERSUBSCRIBE case allow to force start engine.

Required-githooks: true

Co-authored-by: Tom Nabarro <tom.nabarro@intel.com>

### Gatekeeper:


tanabarr:
I think there are some associated control-plane changes related to this, as we update the number of targets somewhere in the control plane if the number used is different from the number requested. @liuxuezhao do you mind if I push to this PR or should I create a new one?

tanabarr (approved):

liuxuezhao:
> I think there are some associated control-plane changes related to this, as we update the number of targets somewhere in the control plane if the number used is different from the number requested. @liuxuezhao do you mind if I push to this PR or should I create a new one?

@tanabarr thanks for updating the go part. there are a few test failures in src/control/run_go_tests.sh (log in unit_test_logs/src-control-run_go_tests.sh_61/output.log) I am not quite familiar, could you please check at convenient time?

tanabarr:
I’ve commented, thanks. It is passed test now

tanabarr:
go unit test failures for previous CI runs seem unrelated, I have created a ticket for them as they seem to be a new issue: https://daosio.atlassian.net/browse/DAOS-13771

src/engine/init.c
```diff
@@ -246,56 +245,61 @@ modules_load(void)
...
-			"use (%d) target XS\n", nr, ncores, tgt_nr);
+	if (ncores_needed(tgt_nr, dss_tgt_offload_xs_nr) > ncores) {
+		if (ncores < DAOS_TGT0_OFFSET + tgt_nr) {
+			D_ERROR("cannot start engine with %d targets on %d cores, may try with "
+				"DAOS_TARGET_OVERSUBSCRIBE=1\n",
+				tgt_nr, ncores);
+			return -DER_INVAL;
+		}
+		dss_tgt_offload_xs_nr = ncores - DAOS_TGT0_OFFSET - tgt_nr;
+		D_PRINT("Start engine with %d targets on %d cores, #nr_xs_helpers set as %d.\n",
```
knard-intel, line 294: The message should also indicate that the number of helpers have been forced.

I was expecting that updating this value could only be done with oversubscribe is true.

src/engine/init.c
```diff
@@ -246,56 +245,61 @@ modules_load(void)
...
+			tgt_nr, ncores, dss_tgt_offload_xs_nr);
+	}
 
+out:
 	if (dss_tgt_offload_xs_nr % tgt_nr != 0)
 		dss_helper_pool = true;
```
knard-intel, line 300: If there is some performance impact, it could be useful indicate that the helper threads will be shared.

src/engine/init.c
```diff
@@ -246,56 +245,61 @@ modules_load(void)
...
+
 /**
- * Get the appropriate number of main XS based on the number of cores and
- * passed in preferred number of threads.
+ * Check if the #targets and #nr_xs_helpers is valid to start server, the #nr_xs_helpers possibly
+ * be reduced.
  */
 static int
-dss_tgt_nr_get(unsigned int ncores, unsigned int nr, bool oversubscribe)
+dss_tgt_nr_check(unsigned int ncores, unsigned int tgt_nr, bool oversubscribe)
```
knard-intel, line 259: I am not used to the engine code, but it seems a little bit odd that a check function change the value of the global variables dss_tgt_offload_xs_nr and dss_helper_pool.

knard-intel (commented):
There is some points which are not clear for me, probably as I am not used to work on the engine code.

tanabarr:
@liuxuezhao I've resolved the conflict, could you please address @knard-intel 's review comments. @NiuYawei when you get a chance please review, thanks in advance.

tanabarr (dismissed):

src/engine/init.c
NiuYawei: Yeah, I'm wondering if we should treat nr_offload in the same of how we treat nr_target? I mean should we silently 'correct' user's invalid config or should we explicitly reject the invalid config? Should we apply 'oversubscribe' to nr_offload too?
@Michael-Hennecke , @tanabarr , any thoughts?

src/engine/init.c
liuxuezhao: helper threads is not must needed to start daos engine, and start more #helpers than #cores could hurt performance, so I select to reduce #helpers if #cores is not enough but #cores are enough to start the #targets.
@NiuYawei any thought with this?

src/engine/init.c
liuxuezhao: whether or not helper XS is shared by targets is based on the #helpers and #targets. for example if with 8 tgts and 4 helpers, it will be shared.

src/engine/init.c
liuxuezhao: ah, just see this comment. 
helper threads is not must needed to start daos engine, and start more #helpers than #cores could hurt performance, that is different with #targets, so I select to reduce #helpers if #cores is not enough but #cores are enough to start with #targets.
For example even if DAOS_TARGET_OVERSUBSCRIBE not specified, on a node with 16 cores, user configured with #target=12 and #helpers=12, I think the #target=12 is mandatory but #helpers=12 is not mandatory, so why not just start server with #targets=12 and #helpers=2 in this case? (another 2 cores used for system XS).

NiuYawei (approved):

src/engine/init.c
tanabarr: I understand the argument above but I would be favour of being explicit and refusing to start with #helpers=12 if there are not enough cores (assuming oversubscribe is not set). Printing explicit error that is easy to understand and forces the user to reduce #helpers and restart the server. This should make behaviour easier to understand and avoid the situation where the user is running an engine unaware that the number of actual helper XSs isn't what they requested in the config.

liuxuezhao: I tend to think that the "nr_xs_helpers" is not a mandatory requirement, that is different with "targets". Imagine the case  that #cores is not enough for helpers but #cores are enough to start with #targets, if fails the engine starting it just add some trouble for user to change yaml file and restart. For a system with different nodes that with different #cores, user have to provide different yml files for different node.
we may need to change document (admin/deployment.md) to explain the "nr_xs_helpers" is not a mandatory requirement and possibly be reduced if the hardware core resource is not enough.
How do you think? @tanabarr @Michael-Hennecke

tanabarr: @liuxuezhao sorry for delay in response, I'm on leave. I think that explanation makes sense to me so let's leave it as it is for the moment with the loose requirement special case for `nr_xs_helpers`. We should get this PR landed.

src/engine/init.c
knard-intel: In fact, I was just wondering that it could be useful for the end user and the support to know if the helper XS is shared or not.
Indeed, this configuration is not explicitly managed by the end user, and thus if there is some performance impact, this information could be useful.

src/engine/init.c
knard-intel: Fair enough for me.

knard-intel (dismissed):
Small remarks not preventing landing from my side.

liuxuezhao:
please don't merge, need a few other changes.

liuxuezhao:
just a rebase/merge with latest master, and add a few changes at the last commit.

liuxuezhao:
the only below test failure is due to https://daosio.atlassian.net/browse/DAOS-13914
ASSERT: result != -DER_TX_RESTART
src/object/obj_tx.c:3633: error: Failure!Test teardown failed

src/engine/init.c
liuxuezhao: > @liuxuezhao sorry for delay in response, I'm on leave. I think that explanation makes sense to me so let's leave it as it is for the moment with the loose requirement special case for `nr_xs_helpers`. We should get this PR landed.

Hi no problem at all, I just refreshed it to add a few other small changes.

NiuYawei (dismissed):

knard-intel (dismissed):
LGTM

tanabarr (dismissed):

liuxuezhao:
@daos-stack/daos-gatekeeper ping? thx

mchaarawi:
> @daos-stack/daos-gatekeeper ping? thx

just waiting for a confirmation from @Michael-Hennecke

Michael-Hennecke (commented):
I would prefer that the #xs_helpers are treated the same way as the #targets: If DAOS_TARGET_OVERSUBSCRIBE is not set, and the sum of targets and helpers exceeds the number of cores on the socket, we should throw an error rather than silently adjusting either #targets or #xs_helpers.

it is important to notify the administrator that something is wrong in the requested server config, and it's easy for the admin to adjust the YML file once they understand the requirements. but silently changing the requested values may lead to obscure and hard-to-find performance behaviour so we should really avoid doing that.

tanabarr (dismissed):

tanabarr (approved):

knard-intel (approved):

liuxuezhao:
> I would prefer that the #xs_helpers are treated the same way as the #targets: If DAOS_TARGET_OVERSUBSCRIBE is not set, and the sum of targets and helpers exceeds the number of cores on the socket, we should throw an error rather than silently adjusting either #targets or #xs_helpers.
> 
> it is important to notify the administrator that something is wrong in the requested server config, and it's easy for the admin to adjust the YML file once they understand the requirements. but silently changing the requested values may lead to obscure and hard-to-find performance behaviour so we should really avoid doing that.

OK, changed as the suggestion. thanks

gnailzenh (approved):

PR #12423, opened 2023-06-16
DAOS-623 gurt: Further improve daos_errno enum

ashleypittman:
Cleanup defenition macros.

Required-githooks: true


src/vos/tests/vts_dtx.c
```diff
@@ -833,7 +833,7 @@ dtx_18(void **state)
 
 	for (i = 0; i < 10; i++) {
 		rc = vos_dtx_check(args->ctx.tc_co_hdl, &xid[i], NULL, NULL, NULL, NULL, false);
-		assert_rc_equal(rc, DTX_ST_COMMITTED);
+		assert_int_equal(rc, DTX_ST_COMMITTED);
```
jolivier23, line 836: Perhaps we should have assert_rc_equal handle such internally?

It is asserting that rc == a positive value.  But if rc is negative, we actually want to print with DF_RC

So perhaps we could change the macro to something like

if (expected > 0) {
   if (rc <= 0)
       print rc with DF_RC and expected without
   else
      print both rc and expected without DF_RC
} else {
   if (rc > 0)
      print rc with without DF_RC and expected with DF_RC
  else
     print both with DF_RC    
}

jolivier23 (approved):

src/vos/tests/vts_dtx.c
ashleypittman: This is a corner case IMO.  We should not use interfaces where negative values mean one thing and positive ones another.

soumagne (approved):

PR #12446, opened 2023-06-19
DAOS-13686 test: reduce pool NVME size for small_file_count

wangshilong:
There is a possibility that our CI NVME only have about ~370GiB capacity which is not enough to store both 350GiB Data + 40GiB Meta on this SSD, reduce pool NVME size a bit to 300GiB. (test required 30G x 2 for IOR and 50K x 4KIB for mdtest), so it should be fine.

Test-tag: SmallFileCount,test_smallfilecount
Test-nvme: auto_md_on_ssd
Required-githooks: true

### DIFF
--- a/src/tests/ftest/io/small_file_count.yaml
+++ b/src/tests/ftest/io/small_file_count.yaml
@@ -22,7 +22,7 @@ server_config:
       storage: auto
 pool:
   scm_size: 40G
-  nvme_size: 350G
+  nvme_size: 300G
 container:
   type: POSIX
   control_method: daos


NiuYawei (approved):

wangshilong:
Test Passed:
https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-12446/2/artifact/Functional%20Hardware%20Medium/io/small_file_count.py/

src/tests/ftest/io/small_file_count.yaml
```diff
@@ -22,7 +22,7 @@ server_config:
       storage: auto
 pool:
   scm_size: 40G
-  nvme_size: 350G
+  nvme_size: 300G
```
daltonbohning, line 25: I'm not sure the intention of this test, but maybe `size: 95%` would work

daltonbohning (approved):

src/tests/ftest/io/small_file_count.yaml
wangshilong: Yup, that could be better ideally, but now we have known timeout issues with ratio > 90% which is still in-progress fixing, i'd like to land this as it for now.

PR #12448, opened 2023-06-19
DAOS-13214 control: Check engines are ready before fetching usage

tanabarr:
Engines send a notify ready message to indicate dRPC server is up and
running. Refuse to service a storage scan request with usage flag set
when a host's engines are not ready to enforce deterministic behavior.
Differentiate between a dmg storage query usage command request and
dmg storage scan --nvme-meta as the latter should be able to be called
when engines are off-line (in which case cached data will be returned).

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/server/ctl_storage_rpc.go
+++ b/src/control/server/ctl_storage_rpc.go
@@ -506,12 +506,38 @@ func (c *ControlService) adjustScmSize(resp *ctlpb.ScanScmResp) {
 	}
 }
 
+func checkEnginesReady(instances []Engine) error {
+	for _, inst := range instances {
+		if !inst.IsReady() {
+			var err error = FaultDataPlaneNotStarted
+			if inst.IsStarted() {
+				err = errInstanceNotReady
+			}
+
+			return errors.Wrapf(err, "instance %d", inst.Index())
+		}
+	}
+
+	return nil
+}
+
 // StorageScan discovers non-volatile storage hardware on node.
 func (c *ControlService) StorageScan(ctx context.Context, req *ctlpb.StorageScanReq) (*ctlpb.StorageScanResp, error) {
 	if req == nil {
 		return nil, errors.New("nil request")
 	}
 	resp := new(ctlpb.StorageScanResp)
+
+	// In the case that usage stats are being requested, relevant flags for both SCM and NVMe
+	// will be set and so fail if engines are not ready for comms. This restriction should not
+	// be applied if only the Meta flag is set in the NVMe component of the request to continue
+	// to support off-line storage scan functionality which uses cached stats (e.g. dmg storage
+	// scan --nvme-meta).
+	if req.Scm.Usage && req.Nvme.Meta {
+		if err := checkEnginesReady(c.harness.Instances()); err != nil {
+			return nil, err
+		}
+	}
 
 	respScm, err := c.scanScm(ctx, req.Scm)
 	if err != nil {
--- a/src/control/server/ctl_storage_rpc_test.go
+++ b/src/control/server/ctl_storage_rpc_test.go
@@ -73,6 +73,7 @@ func TestServer_CtlSvc_StorageScan_PreEngineStart(t *testing.T) {
 		smbc        *scm.MockBackendConfig
 		tierCfgs    storage.TierConfigs
 		expResp     *ctlpb.StorageScanResp
+		expErr      error
 	}{
 		"successful scan; scm namespaces": {
 			bmbc: &bdev.MockBackendConfig{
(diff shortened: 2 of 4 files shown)


tanabarr:
@NiuYawei I tried to but this is currently sent before modules are initialized,

NiuYawei:
> @NiuYawei I tried to switch the engine init order to send notify ready before modules are initialized but the engine would not start so I reverted the change in [aeed750](https://github.com/daos-stack/daos/commit/aeed75088bac705a481e22da9f1e4ae68dcf303f) in this PR, That probably explains why it has been left like that. Unless you feel strongly I probably won't pursue that in this PR, it seems to address the current problem even without that change.

Yes, I talked with LiWei. The dRPC ready has to be called before setup modules, since some module setup require the self-rank sent from daos server. What Liwei suggested is:
1. Move the VOS files re-creation to ds_mgmt_drpc_set_up() which is called on handling the MGMT_SET_UP dRPC;
2. Make sure control plane not sending request to engine before getting the response of this MGMT_SET_UP dRPC;

This suggestion looks reasonable to me, but I don't know why your current patch could solve the issue, maybe it just shrunk the race window and make the issue unlikely to happen?

NiuYawei (dismissed):
I'm fine with the PR if i's able to solve the issue for this moment. If we hit the issue again, we'd consider the long-term solution I mentioned in the comment.

knard-intel (dismissed):
As indicated by @NiuYawei , I also do not well understand how this patch solve the original issue.

tanabarr:
added small change to avoid another possibility of calling over dRPC before ready

src/control/server/ctl_storage_rpc.go
```diff
@@ -506,13 +506,39 @@ func (c *ControlService) adjustScmSize(resp *ctlpb.ScanScmResp) {
...
 		return nil, errors.New("nil request")
 	}
 	resp := new(ctlpb.StorageScanResp)
 
+	// In the case that usage stats are being requested, relevant flags for both SCM and NVMe
+	// will be set and so fail if engines are not ready for comms. This restriction should not
+	// be applied if only the Meta flag is set in the NVMe component of the request to continue
+	// to support off-line storage scan functionality which uses cached stats (e.g. dmg storage
+	// scan --nvme-meta).
+	if req.Scm.Usage && req.Nvme.Meta {
```
kjacque, line 536: Based on the comment, I'm wondering, is checking `req.Nvme.Meta` necessary at all? Is it possible to query only SCM usage without NVMe?

tanabarr: it's already a bit of a hack extrapolating intent from these derived flags, relying on just the Scm.Usage flag seems like a worse hack. Seems like depending on both would prevent strange behaviour if someone introduced Scm only usage scan request.

kjacque: Hm, okay. I would have thought any request for the usage stats could result in this issue.

Maybe it would be best to make this check at the point where we actually need something from the running engines? We wouldn't abort as quickly, but maybe it would feel less hacky.

NiuYawei (approved):

src/control/server/ctl_storage_rpc.go
tanabarr: We do make "engine ready" checks elsewhere further down the call stack but the reason I want to do it here is to avoid performing an SCM usage scan before realising dRPC isn't ready. I think it's best to check that first and return early.

knard-intel (approved):
LGTM,

tanabarr:
clang-format expected to fail for golang protobuf auto compiled code

ashleypittman:
> clang-format expected to fail for golang protobuf auto compiled code

No it's not, today should be ignored for the check. There was an issue that we only fixed yesterday though that meant the check was running against the wrong target and that looks like what happened here.

tanabarr:
> > clang-format expected to fail for golang protobuf auto compiled code
> 
> No it's not, auto generated code should be ignored for the check. There was an issue that we only fixed yesterday though that meant the check was running against the wrong target and that looks like what happened here.

Can this PR be landed then please @daos-stack/daos-gatekeeper ?

src/control/server/ctl_storage_rpc.go
kjacque: All right. I do agree it's good to pre-empt it if we have good signals, but my issue is just what you mentioned above: "extrapolating intent from these derived flags." I'm okay with landing as-is for now, but maybe we should revisit the messages we use for StorageScan if they're not describing the user's intent clearly enough in this case.

PR #12462, opened 2023-06-21
DAOS-13678 mgmt: use large RPC timeout for pool creation

wangshilong:
Test-nvme: auto_md_on_ssd
Test-tag: ec_online_rebuild_array,ec_online_rebuild_mdtest
Required-githooks: true


src/mgmt/srv_pool.c
```diff
@@ -58,6 +58,23 @@ ds_mgmt_tgt_pool_destroy_ranks(uuid_t pool_uuid, d_rank_list_t *filter_ranks)
...
+static uint32_t
+pool_create_rpc_timeout(crt_rpc_t *tc_req, size_t scm_size)
+{
+	uint32_t	timeout;
+	uint32_t	extra_timeout = 0;
+	int		rc;
+
+	rc = crt_req_get_timeout(tc_req, &timeout);
+	D_ASSERTF(rc == 0, "crt_req_get_timeout: "DF_RC"\n", DP_RC(rc));
+	if (bio_nvme_configured(SMD_DEV_TYPE_META)) {
```
NiuYawei, line 70: Maybe it's not necessary to check this, tmpfs in legacy mode has the exactly same issue. Probably we'd just simply bump pool create timeout no matter if it's md-on-ssd mode.

src/mgmt/srv_pool.c
```diff
@@ -58,6 +58,23 @@ ds_mgmt_tgt_pool_destroy_ranks(uuid_t pool_uuid, d_rank_list_t *filter_ranks)
...
+		extra_timeout = scm_size / ((size_t)1024 * 1024 * 1024 * 5);
+		extra_timeout = min(60, extra_timeout);
+	}
+
+	return timeout + extra_timeout;
```
NiuYawei, line 75: I think a proper way is to estimate timeout based on pool size, then return the max(default, estimated).

knard-intel:
Hello @wangshilong ,

FYI I work on the same kind of issue with my ticket [#13582](https://daosio.atlassian.net/browse/DAOS-13582), and I was not able to change the RPC timeout with using your patch: calling the function `crt_context_set_timeout()`
At the end, I have found a solution which is working locally.
I have pushed the draft PR #12578 for checking if it is also working with the CI.
I am going to continue to find a cleaner solution.

frostedcmos:
> I was not able to change the RPC timeout with using your patch: calling the function `crt_context_set_timeout()`.

One thing to keep in mind is that if the individual RPC timeout is set via crt_req_set_timeout() then it will override the timeout set for the context using crt_context_set_timeout().

knard-intel:
> > I was not able to change the RPC timeout with using your patch: calling the function `crt_context_set_timeout()`.
> 
> One thing to keep in mind is that if the individual RPC timeout is set via crt_req_set_timeout() then it will override the timeout set for the context using crt_context_set_timeout().

I have tried to use `crt_req_set_timeout()` as it was done in the patch proposed by shilong and the new timeout was not taken into account: still equal to 10s.  I am going to investigate why it is not working as expected.

wangshilong:
> > > I was not able to change the RPC timeout with using your patch: calling the function `crt_context_set_timeout()`.
> > 
> > 
> > One thing to keep in mind is that if the individual RPC timeout is set via crt_req_set_timeout() then it will override the timeout set for the context using crt_context_set_timeout().
> 
> @frostedcmos, I have tried to use `crt_req_set_timeout()` function as it was done in the patch proposed by @wangshilong and the new timeout was not taken into account: still equal to 10s. I am going to investigate why it is not working as expected. @frostedcmos, could it be related to the ticket [DAOS-13776](https://daosio.atlassian.net/browse/DAOS-13776) ?

Hi @knard-intel , yes, i think this PR was blocked by DAOS-13776

frostedcmos:
> > > > I was not able to change the RPC timeout with using your patch: calling the function `crt_context_set_timeout()`.
> > > 
> > > 
> > > One thing to keep in mind is that if the individual RPC timeout is set via crt_req_set_timeout() then it will override the timeout set for the context using crt_context_set_timeout().
> > 
> > 
> > @frostedcmos, I have tried to use `crt_req_set_timeout()` function as it was done in the patch proposed by @wangshilong and the new timeout was not taken into account: still equal to 10s. I am going to investigate why it is not working as expected. @frostedcmos, could it be related to the ticket [DAOS-13776](https://daosio.atlassian.net/browse/DAOS-13776) ?
> 
> Hi @knard-intel , yes, i think this PR was blocked by DAOS-13776

If this is a collective RPC then yes, it will be a problem until DAOS-13776 is implemented. 

Collective RPCs are sent to more than 1 engine at a time and internally look like this (assuming spanning factor of 2):

Client/Sender --RPC--> Server A --- internal rpc ---> Server B,C --- internal rpc ---> Servers D, E. etc...

The timeout you specify will control only the timeout between a Client/Sender and Server A, but will not impact internal timeouts of the  RPC between servers. When any step of the collective RPC fails or times out, the overall operation gets marked as failed.

Timeouts between servers will be controlled by server side crt_context_set_timeout() or a global CRT_TIMEOUT used by that server. If your collective RPC is being handled on the dedicated engine-side context, then crt_context_set_timeout() would be the best solution for now. Otherwise you would need to bump CRT_TIMEOUT value on the engine so that all RPCs including internal ones use this timeout.

knard-intel:
> > > > > I was not able to change the RPC timeout with using your patch: calling the function `crt_context_set_timeout()`.
> > > > 
> > > > 
> > > > One thing to keep in mind is that if the individual RPC timeout is set via crt_req_set_timeout() then it will override the timeout set for the context using crt_context_set_timeout().
> > > 
> > > 
> > > @frostedcmos, I have tried to use `crt_req_set_timeout()` function as it was done in the patch proposed by @wangshilong and the new timeout was not taken into account: still equal to 10s. I am going to investigate why it is not working as expected. @frostedcmos, could it be related to the ticket [DAOS-13776](https://daosio.atlassian.net/browse/DAOS-13776) ?
> > 
> > 
> > Hi @knard-intel , yes, i think this PR was blocked by DAOS-13776
> 
> If this is a collective RPC then yes, it will be a problem until DAOS-13776 is implemented.
> 
> Collective RPCs are sent to more than 1 engine at a time and internally look like this (assuming spanning factor of 2):
> 
> Client/Sender --RPC--> Server A --- internal rpc ---> Server B,C --- internal rpc ---> Servers D, E. etc...
> 
> The timeout you specify will control only the timeout between a Client/Sender and Server A, but will not impact internal timeouts of the RPC between servers. When any step of the collective RPC fails or times out, the overall operation gets marked as failed.
> 
> Timeouts between servers will be controlled by server side crt_context_set_timeout() or a global CRT_TIMEOUT used by that server. If your collective RPC is being handled on the dedicated engine-side context, then crt_context_set_timeout() would be the best solution for now. Otherwise you would need to bump CRT_TIMEOUT value on the engine so that all RPCs including internal ones use this timeout.

@frostedcmos, FYI `CRT_TIMEOUT` seems to not working in our case: the cart request is constructed with the function `crt_corpc_req_create()` which internal use the `crt_context_set_timeout()` to fix the timeout of the request. From what I understood this last function call supersede the `CRT_TIMEOUT` environment variable.
As for @wangshilong , I will wait after the [DAOS-13776](https://daosio.atlassian.net/browse/DAOS-13776) to fix my own ticket.
Thanks for your explanation.

frostedcmos:
> > > > > > I was not able to change the RPC timeout with using your patch: calling the function `crt_context_set_timeout()`.
> > > > > 
> > > > > 
> > > > > One thing to keep in mind is that if the individual RPC timeout is set via crt_req_set_timeout() then it will override the timeout set for the context using crt_context_set_timeout().
> > > > 
> > > > 
> > > > @frostedcmos, I have tried to use `crt_req_set_timeout()` function as it was done in the patch proposed by @wangshilong and the new timeout was not taken into account: still equal to 10s. I am going to investigate why it is not working as expected. @frostedcmos, could it be related to the ticket [DAOS-13776](https://daosio.atlassian.net/browse/DAOS-13776) ?
> > > 
> > > 
> > > Hi @knard-intel , yes, i think this PR was blocked by DAOS-13776
> > 
> > 
> > If this is a collective RPC then yes, it will be a problem until DAOS-13776 is implemented.
> > Collective RPCs are sent to more than 1 engine at a time and internally look like this (assuming spanning factor of 2):
> > Client/Sender --RPC--> Server A --- internal rpc ---> Server B,C --- internal rpc ---> Servers D, E. etc...
> > The timeout you specify will control only the timeout between a Client/Sender and Server A, but will not impact internal timeouts of the RPC between servers. When any step of the collective RPC fails or times out, the overall operation gets marked as failed.
> > Timeouts between servers will be controlled by server side crt_context_set_timeout() or a global CRT_TIMEOUT used by that server. If your collective RPC is being handled on the dedicated engine-side context, then crt_context_set_timeout() would be the best solution for now. Otherwise you would need to bump CRT_TIMEOUT value on the engine so that all RPCs including internal ones use this timeout.
> 
> @frostedcmos, FYI `CRT_TIMEOUT` seems to not working in our case: the cart request is constructed with the function `crt_corpc_req_create()` which internally use the `crt_context_set_timeout()` to fix the timeout of the request. From what I understood this last function call supersede the `CRT_TIMEOUT` environment variable. As for @wangshilong , I will wait after the [DAOS-13776](https://daosio.atlassian.net/browse/DAOS-13776) to fix my own ticket. Thanks for your explanation.

This PR (https://github.com/daos-stack/daos/pull/12585) is currently in CI for testing and it makes CORPC children inherit the original timeout of a parent.

I've only done limited testing of it on my corpc samples, but it appears to work. This only addresses CORPCs, but does not touch IV timeouts. 
`Results from local testing on test_corpc_prefwd:
...
SRV [rank=4 pid=1161108]        Pre-forward called
SRV [rank=3 pid=1161107]        Pre-forward called
SRV [rank=3 pid=1161107]        Handler called
SRV [rank=3 pid=1161107]        Post-reply called
crt_corpc_req_hdlr() child_rpc=0x7f34640833b0 inherits 60 timeout from parent_rpc=0x7f3464082eb0
crt_corpc_req_hdlr() child_rpc=0x7f20dc083e00 inherits 60 timeout from parent_rpc=0x7f20dc082eb0
...`

wangshilong:
@knard-intel Since you also have a PR, please let me know if you want me to drop mine.

knard-intel:
> @knard-intel Since you also have a PR, please let me know if you want me to drop mine.

Your PR is perfectly fine for me, and it will be more clean for the GIT tree to use yours as it is linked to the original ticket.

knard-intel (approved):

wangshilong:
Rebased to include Alex's collective RPC fix.

NiuYawei (approved):

wangshilong:
@daos-stack/daos-gatekeeper hardware large test passed(which included ec_rebuild_mdtest), there are still two failures for MD-on-SSD, i checked log, it should not be related to this PR, one is OOM happend(DAOS-13147)during test, and another one is we might need increase timeout for extend rebuild simple test.

knard-intel:
Hello @wangshilong ,
Your test failure with _pool/create.py:PoolCreateTests_ should be related to the ticket [DAOS-13582](https://daosio.atlassian.net/browse/DAOS-13582).
For the test failure with _daos_test/suite.py:DaosCoreTest_, I also have several unrelated PRs which are failing on the same way and the   [md-on-ssd daily](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/md-on-ssd-testing/47/testReport/FTEST_daos_test/DaosCoreTest/Test___Functional_Hardware_Medium_Verbs_Provider___20___daos_test_suite_py_DaosCoreTest_test_daos_extend_simple_run_agent_config_transport_config_daos_tests_args_daos_test_num_clients_pools_created_scalable_endpoint_stopped_ranks_test_name_dmg_hosts_pool_server_config_control_metadata_engines_0_storage_0_1_timeouts_b1bc/) jenkins job is also failing on this test.
Sadly, I was not yet able to find a JIRA ticket reporting this error.

PR #12464, opened 2023-06-21
DAOS-13146 tests: increase system ram reserved for ior smoke

wangshilong:
Test-tag: ec,ec_smoke,ior
Test-repeat: 10
Required-githooks: true
Test-nvme: auto_md_on_ssd


daltonbohning (approved):

daltonbohning (commented):
We'll need to wait for the build+test to pass before landing

PR #12484, opened 2023-06-22
DAOS-13214 control,engine: Notify ready flow improvements

tanabarr:
Move IsReady() check into CallDrpc() function and adjust rank setup
routine as appropriate. This eliminates the possibility of calling
into an engine from the control-plane before SetRank() and Setup()
have both returned successfully.

On the engine side, call pool targets recreation logic and setup all
modules from within the dRPC setup handler to ensure this is performed
before the ready state is set in the control-plane.

Required-githooks: true

### Gatekeeper:


tanabarr:
@liw @wangshilong @NiuYawei do the C changes look appropriate regarding recreating targets on dRPC setup (and before control-plane updates to ready state)? The changes have some build repercussions and now because srv_drpc.c is calling ds_mgmt_tgt_init() (separated out from _setup) srv_drpc_tests is failing to build:
```
[2023-06-22T10:42:17.388Z] /bin/ld: build/dev/gcc/src/mgmt/rpc.os: undefined reference to symbol 'crt_proc_d_rank_list_t'
[2023-06-22T10:42:17.389Z] build/dev/gcc/src/cart/libcart.so.4: error adding symbols: DSO missing from command line
[2023-06-22T10:42:17.389Z] collect2: error: ld returned 1 exit status
[2023-06-22T10:42:17.389Z] scons: *** [build/dev/gcc/src/mgmt/tests/srv_drpc_tests] Error 1
```
Should I pull all the requisites into mgmt/tests/SConscript or reorganising another way to avoid this?

liw:
`NotifyReady` means that the engine is notifying the server that it is ready to receive dRPC requests from dRPC's point of view. The server then sends `SetRank` and `SetUp` dRPCs to the engine to continue the startup procedure. In other words, `NotifyReady` is not intended to say that the engine were ready to process any kind of dRPC requests. Hence I think we should not change this part.

In my discussion with Niu earlier this week, I mentioned that we should move `dss_module_setup_all` into `ds_mgmt_drpc_set_up` so that the server would know that when it received the `SetUp` response, the engine must have completed `dss_module_setup_all` and become capable of processing more kinds of dRPC requests (e.g., those that depend on `dss_module_setup_all`).

src/control/server/instance.go
```diff
@@ -273,16 +273,22 @@ func (ei *EngineInstance) handleReady(ctx context.Context, ready *srvpb.NotifyRe
 	return ei.SetupRank(ctx, r, mapVersion)
 }
 
-func (ei *EngineInstance) SetupRank(ctx context.Context, rank ranklist.Rank, map_version uint32) error {
-	if err := ei.callSetRank(ctx, rank, map_version); err != nil {
+func (ei *EngineInstance) SetupRank(ctx context.Context, rank ranklist.Rank, map_version uint32) (err error) {
+	ei.ready.SetTrue()
```
kjacque, line 277: Doesn't this potentially create a gap where ready is set to true, but the setup hasn't happened yet?

kjacque (commented):
Just a couple questions. Overall looks okay.

liw:
Hmm, I mean the following daos_server logic:

- Upon receiving the `NotifyReady` request, this engine enters **state X**. (In state X, the engine is listening to dRPCs and can handle `SetRank`, `SetUp`, etc.)
- Upon receiving the `SetUp` response, this engine moves from state X to **state Y**. (In state Y, the engine is not only listening to dRPCs but has finished setting up and can now handle more kinds of dRPCs.)

I think `CallDrpc` can check if the engine is in state X, but it should not check if the engine is in state Y. It seems like part of the patch wants to use the "ready" flag to indicate state X, while the other part wants to use it to indicate state Y. That probably leads to Kris's comment on `SetupRank` (which I share).

If we move `dss_module_setup_all` to `ds_mgmt_drpc_set_up`, then we don't need to change `ds_mgmt_tgt_setup`, as it is called from `dss_module_setup_all`.

Am I overlooking something? :)

src/control/server/harness.go
```diff
@@ -186,22 +186,10 @@ func (h *EngineHarness) CallDrpc(ctx context.Context, method drpc.Method, body p
...
-		if !i.IsReady() {
-			if i.IsStarted() {
-				if err == nil {
-					err = errInstanceNotReady
-				}
-			} else {
-				if err == nil {
-					err = FaultDataPlaneNotStarted
-				}
-			}
```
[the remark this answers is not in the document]
tanabarr: I removed it because I don't think there is significant benefit in performing the check, CallDrpc now will only return EngineNotReady or DRPCNotReady, returning FaultDataPlaneNotStarted is less valuable because the description is "not started or not responding" which a less specific error. Is "not started" vs just "not ready" something that you feel is particularly useful in this context?

tanabarr:
As discussed with @mjmac I'm splitting the PR into two, one for the engine changes and one for the control-plane. The changes for each don't seem to be interdependent.

https://github.com/daos-stack/daos/pull/12503 concerns the control-plane changes and commit https://github.com/daos-stack/daos/pull/12503/commits/a27be8d3273fd38d3d31b27a969b119e5d995f28 attempts to address the review comments from @kjacque and @liw whilst trying to maintain the intention of ensuring that CallDrpc() can only be called when ready/after SetUp() returns successfully. Does this change satisfies the problems described?

tanabarr:
This PR has been replaced by https://github.com/daos-stack/daos/pull/12503 and https://github.com/daos-stack/daos/pull/12504.

src/control/server/harness.go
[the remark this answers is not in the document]
kjacque: I don't feel that strongly about it, but it *could* tell you whether the engines have crashed, or if they just haven't finished coming up yet.

wangshilong:
PR needs push with Test-nvme: auto_md_on_ssd as it changed MD-on-SSD codes.

src/control/server/instance.go
liw: I share this concern. @tanabarr, since this PR changes the meaning of "ready" from state X to state Y (as defined by my previous PR comment), I wonder if we could skip the "ready" check in EngineInstance.CallDrpc for `SetRank` and `SetUp` dRPCs? E.g.,

    if method != SetRank && method != SetUp && !ei.IsReady() {
        return nil, errEngineNotReady
    }

Then we would no longer need the change here in `EngineInstance.SetupRank`. Is that right?

src/mgmt/srv_target.c
```diff
@@ -494,10 +494,10 @@ recreate_pooltgts()
 }
 
 int
-ds_mgmt_tgt_setup(void)
+ds_mgmt_tgt_init(void)
```
liw, line 497: I believe we do not need to change anything in the file if we move `dss_module_setup_all` to `ds_mgmt_drpc_set_up`. (Apologies for the repeated comment. Could you elaborate a bit why we still need to split `ds_mgmt_tgt_setup`, please? Perhaps I have overlooked something.)

NiuYawei: Indeed, if we move the whole dss_module_setup_all() into ds_mgmt_drpc_set_up(), no other changes required for this ds_mgmt_tgt_setup() anymore.

tanabarr:
replaced by PR-12503 and PR-12504

PR #12502, opened 2023-06-23
DAOS-13397 test: Change fault injection testing to ignore leaks on busy.

ashleypittman:
Update the fault injection testing to disable the leak check
for individual tests if the test opts in ahte Busy is logged
to stderr.

Skip-unit-test: true
Skip-unit-test-memcheck: true
Skip-build-el8-rpm: true
Skip-build-leap15-rpm: true
Skip-build-leap15-icc: true
Skip-build-ubuntu20-rpm: true
Skip-python-bandit: true
Skip-scan-rpms: rpms
Skip-scan-el8-rpms: true
Skip-scan-leap15-rpms: true
Skip-checkpatch: true

Required-githooks: true


src/utils/daos_hdlr.c
```diff
@@ -49,12 +49,18 @@ struct file_dfs {
 };
 
 /* Report an error with a system error number using a standard output format */
-#define DH_PERROR_SYS(AP, RC, STR, ...)					\
-	fprintf((AP)->errstream, STR ": %s (%d)\n", ## __VA_ARGS__, strerror(RC), (RC))
+#define DH_PERROR_SYS(AP, RC, STR, ...)                                                            \
+	do {                                                                                       \
+		fprintf((AP)->errstream, STR ": %s (%d)\n", ##__VA_ARGS__, strerror(RC), (RC));    \
+		fflush((AP)->errstream);                                                           \
+	} while (0)
```
ashleypittman, line 56: @tanabarr @mjmac without this flush I was seeing that sometimes this output was being lost.  This PR introduces changes to check certain things are present in the output which is now I found this.

That makes this part of the PR a bug fix and therefore something we want on 2.4.

src/utils/daos_hdlr.c
```diff
@@ -1668,57 +1673,44 @@ dm_disconnect(struct cmd_args_s *ap,
 	if (src_file_dfs->type == DAOS) {
 		if (is_posix_copy) {
 			rc = dfs_sys_umount(src_file_dfs->dfs_sys);
-			if (rc != 0) {
-				DH_PERROR_DER(ap, daos_errno2der(rc), "failed to unmount source");
-				rc = dfs_sys_umount(src_file_dfs->dfs_sys);
-				if (rc != 0)
-					DH_PERROR_DER(ap, daos_errno2der(rc),
-						      "failed to unmount source on retry");
```
kjacque, line 1676: So these retries weren't needed after all? I think it's an improvement to remove them if we can, but I don't fully understand why they were here to begin with.

kjacque (approved):
LGTM

src/utils/daos_hdlr.c
daltonbohning: Maybe related to
https://daosio.atlassian.net/browse/DAOS-11551
And also maybe part of the fix needed for https://daosio.atlassian.net/browse/DAOS-13499

utils/node_local_test.py
```diff
@@ -5340,10 +5350,64 @@ def get_cmd(cont_id):
...
+    test_cmd.ignore_busy = True
 
     return test_cmd.launch()
 
 
+# Hack for test_alloc_fail_copy_trunc.  Variable scoping means that the get_cmd() function can read
+# but not modify variables in enclosing scope so keeping a counter has to be global.  As python
+# expects global variables to be constants pylint then checks this is upper case so disable that
+# warning.
+aftc_idx = 0  # pylint: disable=invalid-name
```
daltonbohning, line 5363: Why not just make it uppercase?

src/utils/daos_hdlr.c
ashleypittman: This PR removes the need for them.

Rather than always checking for memory leaks the test in question now disables the memory leak test if the process outputs "DER_BUSY".  It makes the check less robust but does mean that the retries in the code aren't needed - it's the desire to not do this that I created the PR for initially.

daltonbohning (commented):
These changes are probably okay, but will you please run extra datamover test just in case? Especially with master instability lately:
```
Features: daos_fs_copy
```

daltonbohning:
> These changes are probably okay, but will you please run extra datamover test just in case? Especially with master instability lately:
> 
> ```
> Features: daos_fs_copy
> ```

Actually, scratch that. I forgot `test_dm_preserve_props_fs_copy_posix_dfs` is a `pr` test

utils/node_local_test.py
```diff
@@ -5340,10 +5350,64 @@ def get_cmd(cont_id):
...
+    # however too many will consume extra resources.
+    files_needed = 4000
+
+    def get_cmd(_):
+        global aftc_idx
+        cmd = [join(conf['PREFIX'], 'bin', 'daos'), 'filesystem', 'copy', '--src', src_file.name,
+               '--dst', f'daos://{pool.id()}/aftc/new_dir/file.{aftc_idx}']
+        aftc_idx += 1
+        assert aftc_idx <= files_needed
+        return cmd
```
daltonbohning, line 5384: This also works:
```suggestion
    def get_cmd(_):
        cmd = [join(conf['PREFIX'], 'bin', 'daos'), 'filesystem', 'copy', '--src', src_file.name,
               '--dst', f'daos://{pool.id()}/aftc/new_dir/file.{get_cmd.aftc_idx}']
        get_cmd.aftc_idx += 1
        assert get_cmd.aftc_idx <= files_needed
        return cmd
    get_cmd.aftc_idx = 0
```


Simplified example:
```
def fun():
    def helper():
        print(helper.val)
        helper.val += 1
    helper.val = 0

    for _ in range(5):
        helper()
fun()
```

daltonbohning (dismissed):

utils/node_local_test.py
ashleypittman: I'm not familiar with the helper.val notation but I got it to work using similar but couldn't resolve the pylint or runtime warnings.

ashleypittman: Will update this PR, thank you.

utils/node_local_test.py
```diff
@@ -4994,14 +4998,22 @@ def _explain():
             show_memleaks = False
             fi_signal = -rc
 
+        if self._aft.ignore_busy and self._aft.check_daos_stderr:
+            stderr = self._stderr.decode('utf-8').rstrip()
+            for line in stderr.splitlines():
+                if line.endswith(': Device or resource busy (-1012)'):
```
tanabarr, line 5004: nit: you probably only need to check for the bracketed error code here rather than a text representation as well

utils/node_local_test.py
```diff
@@ -5062,6 +5074,9 @@ def _explain():
                 if line.endswith(': Cannot allocate memory (12)'):
                     continue
 
+                if self._aft.ignore_busy and line.endswith(': Device or resource busy (-1012)'):
```
tanabarr, line 5077: as above

utils/node_local_test.py
ashleypittman: We expect output to be well-formed so if the test fails because something prints -1012 but not the full text then that's probably something we want to fail.

kjacque (approved):

PR #12503, opened 2023-06-23
DAOS-13214 control: Disallow dRPCs if engine is not ready

tanabarr:
Move IsReady() check into CallDrpc() function and adjust rank setup
routine as appropriate. This eliminates the possibility of calling
into an engine from the control-plane before SetRank() and Setup()
have both returned successfully.

Required-githooks: true

### Gatekeeper:


kjacque (dismissed):

src/control/server/instance.go
```diff
@@ -287,7 +287,7 @@ func (ei *EngineInstance) SetupRank(ctx context.Context, rank ranklist.Rank, map
 }
 
 func (ei *EngineInstance) callSetRank(ctx context.Context, rank ranklist.Rank, map_version uint32) error {
-	dresp, err := ei.CallDrpc(ctx, drpc.MethodSetRank, &mgmtpb.SetRankReq{Rank: rank.Uint32(), MapVersion: map_version})
+	dresp, err := ei.callDrpc(ctx, drpc.MethodSetRank, &mgmtpb.SetRankReq{Rank: rank.Uint32(), MapVersion: map_version})
```
liw, line 290: Ah, this PR supersedes the previous one? I think this is a good solution too.

liw (dismissed):

knard-intel (dismissed):

src/control/server/harness.go
```diff
@@ -186,22 +186,10 @@ func (h *EngineHarness) CallDrpc(ctx context.Context, method drpc.Method, body p
...
-					err = FaultDataPlaneNotStarted
-				}
-			}
-			continue
-		}
 		resp, err = i.CallDrpc(ctx, method, body)
 
 		switch errors.Cause(err) {
-		case errDRPCNotReady, FaultDataPlaneNotStarted:
+		case errEngineNotReady, errDRPCNotReady:
```
mjmac, line 192: I think that this subtle change is what's causing the MS Failover test problem. The logic in this method was intended to invoke the dRPC failure handlers when the MS Leader has no ranks available to service a request. Looking at the control logs, we can see that the MS Leader's rank is killed, but it never resigns leadership, so the PoolCreate RPC is just retried over and over until timeout. What's supposed to happen is that the PoolCreate goes to the current MS Leader after its rank is killed, the RPC fails and the dRPC failure handlers are invoked (including a ResignLeadership() call), and the PoolCreate is then serviced by the new MS Leader.

I'm not sure what the best approach here is -- basically you're going to want to distinguish between the following scenarios:

- Engine wasn't started yet (retryable)
- Engine was started, but there was a race and the dRPC connection wasn't ready yet (retryable)
- Engine was started, but the dRPC call failed (not retryable, return the error to the caller)

I suspect that the changes in this PR have eliminated detection of the last scenario, with the result that the MS Leader never fails over.

tanabarr: Thanks Mike, let me try to fix

tanabarr: So should harness.CallDrpc() specifically avoid checking if the instances are ready before calling instance.callDrpc()?

mjmac (approved):

kjacque (approved):

src/control/server/harness.go
tanabarr: I've updated implementation

liw (approved):

knard-intel (approved):

PR #12523, opened 2023-06-26
DAOS-13801 client: Add asserts to tse to check for errors.

ashleypittman:
When accessing a tse structure verify it's valid before use.
Add asserts for both used-to-be-valid an not-valid to catch
both use-after-free and memory corruption errors.

Required-githooks: true


src/common/tse.c
```diff
@@ -249,6 +249,8 @@ tse_task_decref(tse_task_t *task)
 	D_ASSERT(d_list_empty(&dtp->dtp_dep_list));
 	D_ASSERT(d_list_empty(&dtp->dtp_comp_cb_list));
 
+	dtp->dtp_magic = TASK_BAD_MAGIC;
+
 	/*
 	 * MSC - since we require user to allocate task, maybe we should have
```
ashleypittman, line 255: This comment looks outdated, all task are allocated in this tse_task_create() function in this file.

src/common/tse_internal.h
ashleypittman, line 1: As a general comment on this file it's only included from tse.c so my preference would be to move the code there and just remove the header file as it's not adding anything.  I've tried to keep this PR small however.

src/common/tse_internal.h
```diff
@@ -53,7 +53,9 @@ struct tse_task_private {
 					 dtp_no_propagate:1,
 					 dtp_dep_cnt:28;
 	/* refcount of the task */
-	uint32_t			 dtp_refcnt;
+	uint16_t                         dtp_refcnt;
```
ashleypittman, line 56: If uint16_t isn't deemed big enough for the ref count we could set the refcnt itself to MAD_MAGIC on free and still detect the use-after-free case which is the bulk of the benefit in this PR.

mchaarawi (dismissed):

src/common/tse_internal.h
liuxuezhao: this could be a debug-purpose patch, but looks not a general/nice solution for product code. I think it is not worth to reduce refcnt to 16 bits just to reserve 16 bits for magic number.
similar issue with RPC refcount etc.

ashleypittman: Yes, I was also thinking that.  65535 should be more than enough for ref count (but we could add an assert to verify it doesn't wrap).

To make this conditionally compiled we'd need to add something to the SConstruct file to set TSE_DEBUG and that would mean we'd want to move the struct from the header file where it could be accessed from elsewhere in the source to the c file where it couldn't (as accessing the header without the define would be a bug).

src/common/tse.c
```diff
@@ -8,18 +8,166 @@
...
+	    /* Don't propagate err-code from dependent tasks */
+	    dtp_no_propagate : 1, dtp_dep_cnt : 28;
+
+#if TSE_DEBUG
+	/* refcount of the task */
+	uint16_t dtp_refcnt;
+
+	uint16_t dtp_magic;
+#else
+	uint32_t dtp_refcnt;
```
mchaarawi, line 62: i don't like this difference in types for ref cnt for debug vs release builds.
although the ref count is never that high.

ashleypittman: The size of the struct is tightly controlled and @liuxuezhao wanted the full 32 bits for release builds so unless we can find a different member to adjust this is the best I can do.

I do think that 16 bits is enough for a ref count though.

jolivier23 (approved):
Not sure why dev build watchers is added but the change looks fine to me

ashleypittman:
Is this still of use?

Can we either close it or land it?

mchaarawi:
> Is this still of use?
> 
> Can we either close it or land it?

I'd rather not land this for now as it has not captured any issues in any testing i did, and it will conflict with some other changes that im making to fix some thread sanitizer warnings.

PR #12560, opened 2023-06-29
DAOS-623 test: Check for use of DER_UNKNOWN in logs.

ashleypittman:
Required-githooks: true

### DIFF
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -6205,9 +6205,8 @@ obj_list_shards_get(struct obj_auxi_args *obj_auxi, unsigned int map_ver,
 		grp_idx, *shard, *shard_cnt);
 
 out:
-	D_DEBUG(DB_IO, DF_OID" list on shard %u leader %s: "DF_RC"\n",
-		DP_OID(obj->cob_md.omd_id), *shard,
-		obj_auxi->to_leader ? "yes" : "no", DP_RC(rc));
+	D_DEBUG(DB_IO, DF_OID " list on shard %u leader %s: %d\n", DP_OID(obj->cob_md.omd_id),
+		*shard, obj_auxi->to_leader ? "yes" : "no", rc);
 	return rc;
 }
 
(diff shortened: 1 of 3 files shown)


src/object/cli_obj.c
```diff
@@ -6205,9 +6205,8 @@ obj_list_shards_get(struct obj_auxi_args *obj_auxi, unsigned int map_ver,
 		grp_idx, *shard, *shard_cnt);
 
 out:
-	D_DEBUG(DB_IO, DF_OID" list on shard %u leader %s: "DF_RC"\n",
-		DP_OID(obj->cob_md.omd_id), *shard,
-		obj_auxi->to_leader ? "yes" : "no", DP_RC(rc));
+	D_DEBUG(DB_IO, DF_OID " list on shard %u leader %s: %d\n", DP_OID(obj->cob_md.omd_id),
+		*shard, obj_auxi->to_leader ? "yes" : "no", rc);
```
daltonbohning, line 6209: why not DP_RC?

daltonbohning (approved):
Would this need to go into 2.4 as well? If so, we'll probably need a standalone ticket

ashleypittman:
> Would this need to go into 2.4 as well? If so, we'll probably need a standalone ticket

I don't think so, this is just clearing the way for further checking of the logs.

> If so, we'll probably need a standalone ticket

Agree on this side.

src/object/cli_obj.c
ashleypittman: This is one of those functions where negative is a error code and positive has another meaning.

jolivier23 (approved):

jolivier23:
My only concern is that I'd rather see DER_UNKNOWN(1) on occasion if it means I'll get a proper DER_ when it is an error.  Perhaps the way to handle such cases is to do something like 

If (D_LOG_ENABLED(DB_IO) {
   if (rc > 0)
       D_DEBUG(..., "%d", rc);
   else
      D_DEBUG(..., DF_RC, DP_RC(rc));
}

Perhaps we need a D_DEBUG_RC(DB_IO, rc, .....) essentially does the above?

ashleypittman:
> My only concern is that I'd rather see DER_UNKNOWN(1) on occasion if it means I'll get a proper DER_ when it is an error. Perhaps the way to handle such cases is to do something like

We shouldn't be using code where negative is an error code and positive is a actual value.  Outside of things like iterators where we can only return one value then it's adding risk for little benefit.  We also don't typically use DF_RC for DEBUG messages.

PR #12568, opened 2023-06-30
DAOS-13213 control: Display exclude,rebuild for default self_heal

mjmac:
The default value for the self_heal pool property is Exclude|Rebuild,
so the output should reflect this.

Required-githooks: true

### DIFF
--- a/src/control/lib/daos/pool_property.go
+++ b/src/control/lib/daos/pool_property.go
@@ -46,24 +46,11 @@ func PoolProperties() PoolPropertyMap {
 			Property: PoolProperty{
 				Number:      PoolPropertySelfHealing,
 				Description: "Self-healing policy",
-				valueStringer: func(v *PoolPropertyValue) string {
-					n, err := v.GetNumber()
-					if err != nil {
-						return "not set"
-					}
-					switch {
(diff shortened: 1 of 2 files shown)


src/control/lib/daos/pool_property.go
```diff
@@ -46,24 +46,11 @@ func PoolProperties() PoolPropertyMap {
...
-						return "unknown"
-					}
-				},
 			},
 			values: map[string]uint64{
-				"exclude": PoolSelfHealingAutoExclude,
-				"rebuild": PoolSelfHealingAutoRebuild,
+				"exclude":         PoolSelfHealingAutoExclude,
+				"rebuild":         PoolSelfHealingAutoRebuild,
+				"exclude,rebuild": PoolSelfHealingAutoExclude | PoolSelfHealingAutoRebuild,
```
knard-intel, line 53: `rebuild,exclude` shoudl also probably also be valid values ?

mjmac: Yes, you're right. Fixed, thanks.

kjacque (approved):

knard-intel (approved):

PR #12509, opened 2023-06-23
DAOS-13670 control: Add system name validation on startup

mjmac:
Explicitly check the configured system name and fail if
it does not meet requirements.

Test-tag: pr test_daos_server_config_basic

Required-githooks: true

Signed-off-by: Michael MacDonald <mjmac.macdonald@intel.com>

PR #12508, opened 2023-06-23
DAOS-13471 control: Add JSON-format version to utilities

mjmac:
Allow consumers to grab structured build/version information.
Centralizes JSON output logic to a single implementation in
the cmdutil package. Includes updates to control/version.py
to use the structured version information instead of scraping
stdout.

Signed-off-by: Michael MacDonald <mjmac.macdonald@intel.com>

PR #12570, opened 2023-07-03
DAOS-13886 test: Fix system name validation

knard-intel:
# Description

Fix test regression with system name introduce with PR #12509:

Fix regression introduce with PR #12508: issue with _dmg_ and configuration files parsing.

Miscellaneous small improvements of the _src/tests/ftest/control/version.py_ test.

# Validation

Test-tag is used to limit the tests to the functional ones updated by this PR.

# Before requesting gatekeeper:

# Gatekeeper:

### DIFF
--- a/src/tests/ftest/control/daos_control_config.yaml
+++ b/src/tests/ftest/control/daos_control_config.yaml
@@ -33,7 +33,7 @@ control_config_val: !mux
     config_val:
       - "name"
       - "! @#$%^&*()_+{}|:<>?-=[];',./"
-      - "PASS"
+      - "FAIL"
   name_numeric:
     config_val:
       - "name"
--- a/src/tests/ftest/control/daos_control_config.yaml
+++ b/src/tests/ftest/control/daos_control_config.yaml
@@ -32,6 +32,11 @@ control_config_val: !mux
   name_special_chars:
     config_val:
       - "name"
+      - "! @#$%^&*()_+{}"
+      - "PASS"
+  name_exceeds_max_length:
+    config_val:
+      - "name"
       - "! @#$%^&*()_+{}|:<>?-=[];',./"
       - "FAIL"
   name_numeric:
--- a/src/tests/ftest/control/version.py
+++ b/src/tests/ftest/control/version.py
@@ -78,8 +78,7 @@ class DAOSVersion(TestWithServers):
         self.log.info("RPM version = %s", rpm_version)
 
         # Remove configuration files
-        cleanup_cmd = "sudo rm -frv /etc/daos/daos_control.yml /etc/daos/daos_server.yml"
-        cleanup_cmd += " /etc/daos/daos_agent.yml /etc/daos/certs"
+        cleanup_cmd = "sudo find /etc/daos -type f -delete -print"
         run_pcmd(hosts=self.hostlist_servers, command=cleanup_cmd)
 
         # Get dmg version.
--- a/src/tests/ftest/control/version.py
+++ b/src/tests/ftest/control/version.py
@@ -22,9 +22,30 @@ class DAOSVersion(TestWithServers):
     def __init__(self, *args, **kwargs):
         """Initialize a DAOSVersion object."""
         super().__init__(*args, **kwargs)
+
         # Don't waste time starting servers and agents.
         self.setup_start_servers = False
         self.setup_start_agents = False
+
+        self.errors = []
+
+    def report_result(self):
+        """Helper printing test result"""
+        self.log.info("###### Test Result ######")
+        report_errors(test=self, errors=self.errors)
+        self.log.info("#########################")
+
+    def append_error(self, title, details):
+        """Helper adding an error to the list of errors
+
+        Args:
+            title (str): Error message title
+            details (list): List of string of the error details
+        """
+        msg = title
+        if details:
+            msg += "\n\t" + "\n\t".join(details)
+        self.errors.append(msg)
 
     def test_version(self):
         """Verify version number for dmg, daos, daos_server, and daos_agent against RPM.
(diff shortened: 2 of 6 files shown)


src/tests/ftest/control/version.py
```diff
@@ -22,32 +22,56 @@ class DAOSVersion(TestWithServers):
     def __init__(self, *args, **kwargs):
         """Initialize a DAOSVersion object."""
         super().__init__(*args, **kwargs)
-        # Don't waste time starting servers and agents.
-        self.setup_start_servers = False
-        self.setup_start_agents = False
```
mjmac, line 27: Was this intentional? I added those lines because I noticed that the test was wasting a lot of time in setting up and then starting the server and agent for no reason. The test doesn't need anything to be running in order to check the version of the utilities.

knard-intel: In fact, the configuration files and certificates  seems to not be properly generated when the service are not started.
From what I have understand, the test should not be run by the CI with your PR (as it is not in the pr tag) and locally it should work if these files where previously generated.

mjmac: OK, but my point is that the config/certs shouldn't need to be generated in order to run the version commands. If there is an error when running e.g. 'daos version` without the agent or configuration, then that is a bug that needs to be fixed.

knard-intel: In fact, it is dmg which is complaining.

knard-intel: The agent could probably be not started, however I suspect that the server has to be started for properly generating the daos_control.yml file.

knard-intel: I will make more experiments tomorrow to try to find the best compromise.

mjmac: Ah, then that is a bug that needs to be fixed. None of the commands should ever require a configuration in order to print the version. I would prefer to keep the behavior I added so that we find this more easily if it regresses in the future.

knard-intel: OK, I will try to fix in this PR as it should not be so complex.
Is it OK for you ?

mjmac: Sounds good to me, thanks.

knard-intel: Fixed with commit a233d3fd050f28a749f98427f7290dbacb66f0b5

knard-intel: @mjmac my apologies for the rebase.
I was still working as a draft PR.
Following is the difference between the version that you have reviewed and the latest version.
https://github.com/daos-stack/daos/compare/30411573abd590510ad9b4675dfa122df7865ba7..a233d3fd050f28a749f98427f7290dbacb66f0b5

mjmac (dismissed):

src/tests/ftest/control/version.py
```diff
@@ -22,32 +22,64 @@ class DAOSVersion(TestWithServers):
...
+                break
         if not result:
-            errors.append("RPM version is not in the output! {}".format(output))
-        else:
-            rpm_version = result[0]
-            self.log.info("RPM version = %s", rpm_version)
+            msg = "RPM version could not be defined"
+            self.append_error(msg, stdout)
+            self.report_result()
+        self.log.info("RPM version = %s", rpm_version)
```
shimizukko, line 78: I believe your code works, but I feel more comfortable moving this to line 73 because if you didn't know that `self.report_result()` on line 77 would fail the test, then you would think that this line will print `rpm_version` that is `None`.

shimizukko (approved):

src/tests/ftest/control/version.py
knard-intel: In fact, it was on purpose that it is after the `self.report_result()` to not print an informative message with an inconsistent RPM version.
However, if you think that it will be helpful to have this message, I have no issue to move it at line 73.

shimizukko: I see, no problem. It's okay as it is.

knard-intel:
Restart CI build as it was stuck in functional HW test with the wolf network issue.
However, the functional tests were all passed :)

shimizukko (approved):

mjmac (approved):

src/tests/ftest/control/daos_control_config.yaml
```diff
@@ -33,7 +33,7 @@ control_config_val: !mux
     config_val:
       - "name"
       - "! @#$%^&*()_+{}|:<>?-=[];',./"
-      - "PASS"
+      - "FAIL"
```
tanabarr, line 36: We should expand the coverage to cover special characters and over length, the reason this now fails is because there is a new length requirement introduced by 0b5cdc07a409c525b03dd632626d4fd8753982c0 which limits to 15 chars.

I suggest reducing the value in this test case to 15 chars (which should pass even with special characters) and leaving the expectation as "PASS". Then add another `name_links_exceeded` test case with the original value which should result in a failure.

tanabarr (approved):

shimizukko (approved):

src/tests/ftest/control/daos_control_config.yaml
ashleypittman: I agree that would be better.

knard-intel: > I agree that would be better.

@tanabarr pushed the commit ce831918db33ef78ba682f06a50fbcabba7949b1 to update the unit tests in this way 😸

src/tests/ftest/control/version.py
```diff
@@ -22,32 +22,69 @@ class DAOSVersion(TestWithServers):
...
         """Initialize a DAOSVersion object."""
         super().__init__(*args, **kwargs)
+
         # Don't waste time starting servers and agents.
         self.setup_start_servers = False
         self.setup_start_agents = False
 
+        self.errors = []
+
+    def report_result(self):
```
phender, line 32: The `tearDown()` and `setUp()` methods provide markers to help find the beginning and end of the test - a sequence of 80 `=` characters.  We typically just end each `test_*` method with `self.log.info("Test passed")` to indicate test passing (in addition to avocado's own messaging) and rely on `self.fail()` or assertions to indicate the test failure before calling `tearDown()`.  If we need more than this it should be standardized so we can use the same search string for all test job.log results.   That said, it can be addressed in a different PR.

src/tests/ftest/control/version.py
```diff
@@ -22,32 +22,69 @@ class DAOSVersion(TestWithServers):
...
+        self.log.info("RPM version = %s", rpm_version)
+
+        # Remove configuration files
+        cleanup_cmds = [
+            "sudo find /etc/daos/certs -type f -delete -print",
+            "sudo rm -fv /etc/daos/daos_server.yml /etc/daos/daos_control.yml"
+            "     /etc/daos/daos_agent.yml",
+        ]
+        for cmd in cleanup_cmds:
+            run_pcmd(hosts=self.hostlist_servers, command=cmd)
```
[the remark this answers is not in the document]
knard-intel: I will add these config parameters in a follow up PR as this one has been already merged.
Otherwise, I would prefer to keep the cleanup as it seems that the _/etc/daos_ is not properly cleanned between test runs.

- [ ] Add insecure yaml config

src/tests/ftest/control/version.py
knard-intel: In fact, I have done it this way, to keep the same behavior as it was originally done in the PR.
However, if there is some standard to respect I will fix this in a follow up PR as this one has been already merged.

- [ ] Use standardized output message

knard-intel: Fixed with PR #12615

PR #12571, opened 2023-07-03
DAOS-13400 test: Add a test for faults injection in the server.

ashleypittman:
Add the ability to run the server with fault injection of allocation calls enabled.
Use NLT to start the server normally, enable fault-injection at run-time, perform
some I/O, disable fault injection again and then  stop the server.

Fix a number of warning messages not reporting rc properly.

This test is not enabled or run anywhere yet and due to the random nature
of faults injected is non-deterministic in it's results.


ashleypittman:
This does not change CI but I've run it locally and filed DAOS-13902 DAOS-13903 DAOS-13904 DAOS-13905 and DAOS-13906 this morning with issues this has uncovered as well as DAOS-13900 and DAOS-13901 in getting the test to work.

src/engine/server_iv.c
```diff
@@ -609,7 +615,7 @@ ivc_on_get(crt_iv_namespace_t ivns, crt_iv_key_t *iv_key,
 	D_ALLOC_PTR(priv_entry);
 	if (priv_entry == NULL) {
 		class->iv_class_ops->ivc_ent_put(entry, entry_priv_val);
-		D_GOTO(out, rc);
+		D_GOTO(out, rc = -DER_NOMEM);
```
ashleypittman, line 618: This has it's own ticket and is also included in https://github.com/daos-stack/daos/pull/12635 which should land first.

utils/node_local_test.py
```diff
@@ -917,7 +933,7 @@ def stop(self, wf):
         self.conf.compress_file(self.control_log.name)
 
         for log in self.server_logs:
-            log_test(self.conf, log.name, leak_wf=wf)
+            log_test(self.conf, log.name, leak_wf=wf, skip_fi=self._fi)
```
daltonbohning, line 936: Is this backward? In `__init__`: `self._fi = enable_fi`. So `skip_fi=not self._fi` ?

utils/node_local_test.py
```diff
@@ -1117,6 +1140,70 @@ def run_daos_client_cmd_pil4dfs(self, cmd, check=True, container=None):
...
+                rc = subprocess.run(exec_cmd,
+                                    env=cmd_env,
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE,
+                                    check=False)
+                print(rc)
+                valgrind_hdl.convert_xml()
+                log_test(self.conf, log_file.name)
+                # Do not check returncode here, fault injection might prevent it from working.
+                return False  # fatal_errors
```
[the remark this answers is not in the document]
ashleypittman: This could probably be removed, I'll check.  I'm not sure anything checks the return code here and the comment is wrong, I'm also surprised pylint allows it.

utils/node_local_test.py
```diff
@@ -5772,6 +5800,48 @@ def test_alloc_fail(server, conf):
...
+
+        for idx in range(100):
+            server.run_daos_client_cmd_pil4dfs(
+                ['touch', f'file.{idx}'], container=cont, check=False, report=False)
+            server.run_daos_client_cmd_pil4dfs(
+                ['rm', '-f', f'file.{idx}'], container=cont, check=False, report=False)
+
+        # Turn off fault injection again to assist in server shutdown.
+        server.set_fi(probability=0)
+        server.set_fi(probability=0)
```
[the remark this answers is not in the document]
ashleypittman: We probably don't, we might want to always do it when it's been enabled so it could go in the DaosServer() logic however at some point we should start checking allocation failures on shutdown so better to keep it here and not complicate the DaosServer() object I think.

utils/node_local_test.py
ashleypittman: I think the skip_fi is probably named backwards.  `skip_fi` means do not warn on warnings in the file related to fault injection, therefore if fault injection is on and therefore expected we should not warn if it happens.  I have more work on this (not least finding and reporting issues) so I can re-name or re-work this at a later date.

src/cart/crt_iv.c
```diff
@@ -3426,11 +3427,11 @@ crt_iv_update(crt_iv_namespace_t ivns, uint32_t class_id,
 	* All other checks are performed inside of crt_iv_update_interna.
 	*/
 	if (iv_value == NULL) {
-		D_ERROR("iv_value is NULL\n");
-
 		rc = -DER_INVAL;
-		update_comp_cb(ivns, class_id, iv_key, NULL, iv_value,
-			       rc, cb_arg);
+
+		D_ERROR("iv_value is NULL " DF_RC "\n", DP_RC(rc));
```
frostedcmos, line 3432: probably should be either "iv_value is NULL; rc: " DF_RC  or just drop printing of rc

frostedcmos (approved):
cart changes look good, one minor comment inline

src/cart/crt_iv.c
daltonbohning: > "iv_value is NULL; rc: "

Why though? That wouldn't format the rc like DF_RC does.

> just drop printing of rc

I think it's better to print the rc for debugging

utils/node_local_test.py
```diff
@@ -1101,12 +1121,20 @@ def run_daos_client_cmd_pil4dfs(self, cmd, check=True, container=None):
             print('Stdout from command')
             print(rc.stdout.decode('utf-8').strip())
 
+        # if cwd and os.listdir(tmp_dir.name):
+        #    print('Temporary directory is not empty')
+        #    print(os.listdir(tmp_dir.name))
+        #    assert False, 'Files left in tmp dir by pil4dfs'
```
daltonbohning, line 1127: Remove?

daltonbohning (approved):

src/cart/crt_iv.c
ashleypittman: Including the rc if available is in our coding standard.

utils/node_local_test.py
ashleypittman: Oh yeah, forgot about this.  If `pil4dfs` is working properly then the directory should be empty after use as all I/O is redirected to the container.  What is actually happening for this test is dfs_mount is failing and then pil4dfs is falling back to POSIX I/O and running the program against the backing POSIX file but unless there's a dfuse mount there then it should be rejecting the open call.  I'll file a bug Monday.

PR #12572, opened 2023-07-03
DAOS-13929 client: Add logging macros for direntries.

ashleypittman:
Add DF_DE/DP_DE marcos for logging dentries.  These macros
print as strings for debug builds but hide the contents for release
builds.


src/object/cli_ec.c
```diff
@@ -29,6 +29,59 @@
 #define EC_TRACE(fmt, ...)
 #endif
 
+/* Move to next iov, it's caller's responsibility to ensure the idx boundary */
+#define daos_sgl_next_iov(iov_idx, iov_off)                                                        \
```
daltonbohning, line 33: Doesn't moving these out of `src/include` break backwards compatibility?

src/common/debug.c
```diff
@@ -303,3 +303,26 @@ daos_key2str(daos_key_t *key)
...
+
+	for (i = 0; i < NAME_MAX; i++) {
+		if (de[i] == '\0')
+			return (char *)de;
+		if (de[i] == '/')
+			return "<contains slash>";
+		if (!isprint(de[i]) || de[i] == '\'')
+			return "<not printable>";
+	}
+	return "<entry too long>";
```
daltonbohning, line 327: So the rules are
1. If `null` print `<NULL>`
2. If contains a `/` print `<contains slash>`
   Why?
4. If contains a `'` print `<not printable>`
   Why?
6. If too long print `<entry too long>`
7. Otherwise print the name.

ashleypittman: For non-release builds yes, for release builds just print something else.  I might have missed that part of the PR.

The thinking is that for non-release builds we print `de'<name>'` and can then choose to do post-analysis on the logs to remove them.  Having ' in the logs would cause problems with that.  It's the reason I changed the key2str function above not to print ' as well.

src/object/cli_ec.c
ashleypittman: Yes.

jolivier23 (dismissed):

src/include/daos/common.h
```diff
@@ -149,15 +149,27 @@ char *DP_UUID(const void *uuid);
 #define DF_CONTF		DF_UUIDF"/"DF_UUIDF
 
 #ifdef DAOS_BUILD_RELEASE
-#define DF_KEY			"[%d]"
-#define DP_KEY(key)		(int)((key)->iov_len)
+
+#define DF_KEY       "key[%d]"
```
mchaarawi, line 153: this change is not in the PR description. can you explain why you are changing the logging for DF_KEY?
i'd rather not.

src/include/daos/common.h
```diff
@@ -149,15 +149,27 @@ char *DP_UUID(const void *uuid);
...
+#define DP_KEY(_key) (int)((_key)->iov_len)
+
+#define DF_DE        "de[%zi]"
+#define DP_DE(_de)   strnlen(_de, NAME_MAX)
```
mchaarawi, line 157: perhaps it's best of those macros are added in dfuse layer and not in common as they are dfuse specific?

mchaarawi (changes requested):

src/include/daos/common.h
ashleypittman: They will also be needed for dfs, so unless there's a 'client' header file then here is the best spot as it's the same as the other DF_DP macros.

src/include/daos/common.h
ashleypittman: I thought I'd replied to this earlier but can't see it now, the details are all in https://daosio.atlassian.net/wiki/spaces/~60357e57e2020c0070b4d71f/pages/11403198500/Keeping+information+out+of+log+files.

mchaarawi: this does not address my question, although i was not aware of this proposal. it might be a good topic to discuss at the PDT. also i will mention again that the pr title and description don't mention anything about ppi cleanup and is only about dentries. so this PPI cleanup for keys should be in a different PR.

aside from that, for this change you are adding this "key" logging to DF_KEY. this is not going to be repetitive in most logs i would assume, as the log message can have dkey or akey followed by DF_KEY. this is my requested change for this PR.

mchaarawi (changes requested):

src/include/daos/common.h
ashleypittman: I didn't think that three additional characters in some log lines was a big enough change in itself to warrant an individual ticket/PR for.

A part of this overall change is having different types of strings in the log file represented in a consistent way, so updating the existing macros (of which we currently only have keys) at the same time as adding a second pair of macros made sense to me.

I can update the description to include this detail.

mchaarawi: well im not sure adding the type is a good idea. as i mentioned for keys, the type can be a dkey or akey and the log message usually have that already/

ashleypittman: Looking at the existing use of `DF_KEY` in the source you're right, a lot of it already does include an `akey` or `dkey` prefix already but it's very inconsistent with the exact syntax.  There are only ~100 uses of DF_KEY in the source which is much less than I thought, will revert this part of the PR and handle this afterwards.

mchaarawi (approved):

jolivier23 (approved):

PR #12575, opened 2023-07-04
DAOS-13895 cq: Add checking for log formatting.

ashleypittman:
Add a python script to scrape the source and identify any logging
issues.

Required-githooks: true


.github/workflows/spelling.yml
```diff
@@ -16,3 +16,5 @@ jobs:
           skip: ./src/control/vendor,./.git
           ignore_words_file: ci/codespell.ignores
           builtin: clear,rare,informal,names,en-GB_to_en-US
+      - name: Check DAOS logging macro use.
+        run: ./utils/cq/d_error_check.py --github src
```
ashleypittman, line 20: I will remove this prior to landing.

In theory this should probably go in it's own action however it's easy enough to add it to an existing one. That can be discussed when we're ready to enable this.

daltonbohning: Since we keep adding GHA, and want to add more, it probably makes sense to group similar checks.

daltonbohning (dismissed):
> Check warning on line 616 in src/cart/crt_self_test_client.c
>
> GitHub Actions
> / Codespell
> newline-check, Line contains too many newlines

For this line:
>		D_WARN("Got invalid session id = %ld from endpoint %u:%u -\n"
>		       " removing it from the list of endpoints\n",

Do we really not allow multiple newlines? Just trying to make sure we don't complain about something that IS allowed

ashleypittman:
> 



> > Check warning on line 616 in src/cart/crt_self_test_client.c
> > GitHub Actions
> > / Codespell
> > newline-check, Line contains too many newlines
> 
> For this line:
> 
> > ```
> >   D_WARN("Got invalid session id = %ld from endpoint %u:%u -\n"
> >          " removing it from the list of endpoints\n",
> > ```
> 
> Do we really not allow multiple newlines? Just trying to make sure we don't complain about something that IS allowed

It's not clear.  We have very few places that do and I don't think we should.  It messes up the logs for multi-threaded programs and won't play nicely with things like putting warnings into syslog so no, I don't think we should.

jolivier23 (approved):

src/utils/daos_hdlr.c
```diff
@@ -1490,10 +1490,8 @@ dm_connect(struct cmd_args_s *ap,
 				dattr.da_props = props;
 			rc = duns_create_path(ca->dst_poh, path, &dattr);
 			if (rc != 0) {
-				rc = daos_errno2der(rc);
-				DH_PERROR_DER(ap, rc, "provide a destination pool or UNS path "
-					      "of the form:\n\t --dst </$pool> | </path/to/uns>");
-				D_GOTO(err, rc);
+				DH_PERROR_SYS(ap, rc, "failed to create destination UNS path");
+				D_GOTO(err, rc = daos_errno2der(rc));
```
daltonbohning, line 1494: Looks like the previous and next errors in this same function use DH_PERROR_DER, instead of DH_PERROR_SYS. Should we be consistent?

utils/cq/d_logging_check.py
```diff
@@ -115,12 +119,13 @@ def __next__(self):
...
-# Logging macros that expect a new-line.
+# Logging macros where the new-line is added if missing.
 PREFIXES = ['D_ERROR', 'D_WARN', 'D_INFO', 'D_NOTE', 'D_ALERT', 'D_CRIT', 'D_FATAT', 'D_EMIT',
             'D_TRACE_INFO', 'D_TRACE_NOTE', 'D_TRACE_WARN', 'D_TRACE_ERROR', 'D_TRACE_ALERT',
-            'D_TRACE_CRIT', 'D_TRACE_FATAL', 'D_TRACE_EMIT', 'RPC_TRACE', 'RPC_ERROR']
+            'D_TRACE_CRIT', 'D_TRACE_FATAL', 'D_TRACE_EMIT', 'RPC_TRACE', 'RPC_ERROR',
+            'VOS_TX_LOG_FAIL', 'VOS_TX_TRACE_FAIL', 'D_DEBUG']
 
-# Logging macros that do not expect a new-line.
+# Logging macros where a new-line is always added.
```
daltonbohning: Do we plan to eventually update these macros to have the same behavior with newlines?

utils/cq/d_logging_check.py
```diff
@@ -189,23 +209,107 @@ def check_quote(self, line):
...
+        """Check there is no . after DF_RC"""
+        code = line.raw()
+        count = code.count('DF_RC')
+        if count == 0:
+            return
+        code = code.replace('DF_RC ', 'DF_RC')
+        if 'DF_RC".\\n"' not in code:
+            return
+        code = code.replace('DF_RC".\\n"', 'DF_RC"\\n"')
+        line.fix(code)
```
daltonbohning, line 237: Looks like this would convert
`" DF_RC ".\n"` to
`" DF_RC"\n"`
Where the space _**before**_ remains but the space _**after**_ is removed.

Also, you might consider `re.sub` here. (regular expression substitute). For example:
```
>>> import re
>>> re.sub(r'DF_RC *"\.\\n', 'DF_RC"\\n', 'plain string')
'plain string'
>>> re.sub(r'DF_RC *"\.\\n', 'DF_RC"\\n', 'some error " DF_RC ".\\n')
'some error " DF_RC"\n'
```

utils/cq/d_logging_check.py
```diff
@@ -189,23 +209,107 @@ def check_quote(self, line):
...
+            line.note('Unable to check DF_RC')
+            return
+
+        # Remove any spaces around macros as these may or may not be present.  This updated input
+        # is used for the update check at the end so white-space differences here will not cause
+        # code to be re-written.
+        code = code.replace('DF_RC ', 'DF_RC')
+        code = code.replace(' DF_RC', 'DF_RC')
+        code = code.replace('DP_RC ', 'DP_RC')
+        code = code.replace(' DP_RC ', 'DP_RC')
```
daltonbohning, line 266: FYI
```suggestion
        code = code.replace('DF_RC ', 'DF_RC')
        code = re.sub(r' *DF_RC *', 'DF_RC', code)
```

utils/cq/d_logging_check.py
```diff
@@ -189,23 +209,107 @@ def check_quote(self, line):
...
+        # Check what comes before DF_RC in the message, and strip any trailing punctuation.
+        imsg = None
+        while imsg != msg:
+            imsg = msg
+            if any(map(msg.endswith, [' ', '=', '.', ',', ':', ';'])):
+                msg = msg[:-1]
+            if msg.endswith(var_name):
+                msg = msg[:-len(var_name)]
+            if msg.endswith('rc'):
+                msg = msg[:-2]
```
daltonbohning, line 306: I think a lot of this could be replaced with regular expression matching, but I don't have it off the top of my head

utils/cq/d_logging_check.py
ashleypittman: Yes it would.  The c formatting is very messed up after using this script to apply fixes or corrections, there's an assumption (and I think it tells you to) that the user should run git-clang-format afterwards.

src/utils/daos_hdlr.c
daltonbohning: I agree we shouldn't update them all on this PR, but it was consistently DH_PERROR_DER before. Now this PR changed just one line to DH_PERROR_SYS. So it introduced inconsistency

utils/cq/d_logging_check.py
daltonbohning: But if we're already fixing it, shouldn't we fix it correctly? Or instead of fixing at all, just error?

daltonbohning (approved):

PR #12582, opened 2023-07-05
DAOS-13213 control: Display exclude,rebuild for default self_heal

mjmac:
The default value for the self_heal pool property is Exclude|Rebuild,
so the output should reflect this.

Required-githooks: true


knard-intel (approved):
LGTM

kjacque (approved):

tanabarr (approved):

PR #12612, opened 2023-07-12
DAOS-13904 pool: don't assert ENOMEM for daos_acl_validate()

wangshilong:
daos_acl_validate() might fail because of out of memory (e.g.fault injection), don't assert them in the pool and container IV for this case.

Required-githooks: true

### DIFF
--- a/src/container/container_iv.c
+++ b/src/container/container_iv.c
@@ -1276,9 +1276,8 @@ cont_iv_prop_g2l(struct cont_iv_prop *iv_prop, daos_prop_t **prop_out)
 		acl = &iv_prop->cip_acl;
 		if (acl->dal_ver != 0) {
 			rc = daos_acl_validate(acl);
-			if (rc == -DER_NOMEM)
+			if (rc != -DER_SUCCESS)
 				D_GOTO(out, rc);
-			D_ASSERT(rc == 0);
 			prop_entry->dpe_val_ptr = daos_acl_dup(acl);
 			if (prop_entry->dpe_val_ptr == NULL)
 				D_GOTO(out, rc = -DER_NOMEM);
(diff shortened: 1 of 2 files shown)


src/container/container_iv.c
```diff
@@ -1275,7 +1275,10 @@ cont_iv_prop_g2l(struct cont_iv_prop *iv_prop, daos_prop_t **prop_out)
 		prop_entry = &prop->dpp_entries[i++];
 		acl = &iv_prop->cip_acl;
 		if (acl->dal_ver != 0) {
-			D_ASSERT(daos_acl_validate(acl) == 0);
+			rc = daos_acl_validate(acl);
+			if (rc == -DER_NOMEM)
+				D_GOTO(out, rc);
+			D_ASSERT(rc == 0);
```
ashleypittman, line 1281: For some reason I thought daos_acl_validate() returned a boolean but I see now it's an int.

This function is validating input from the client, if the client sends a invalid acl then the server should not fail so I wonder if this patch could be even simpler.
```suggestion
			rc = daos_acl_validate(acl);
			if (rc != -DER_SUCCESS)
				D_GOTO(out, rc);
```

wiliamhuang (approved):

ashleypittman (approved):
Do you agree this is something where bad client input could have crashed the server?  If so then this is a bug-fix and needs to go on 2.4.

wangshilong:
> Do you agree this is something where bad client input could have crashed the server? If so then this is a bug-fix and needs to go on 2.4.

I thought this IV update and fetch only happen across the server side, that is why i still keep assertion except ENOMEM, but yeah, i agreed that your comments to simplify code is good idea as well.

bfaccini (approved):

ashleypittman:
> > Do you agree this is something where bad client input could have crashed the server? If so then this is a bug-fix and needs to go on 2.4.
> 
> I thought this IV update and fetch only happen across the server side, that is why i still keep assertion except ENOMEM, but yeah, i agreed that your comments to simplify code is good idea as well.

Ah, I hadn't looked into what RPC it was or considered that it might be server to server.

In general though ASSERT() is not suitable for input validation so we do want this PR as is.

PR #12627, opened 2023-07-13
DAOS-11551 control: Flush c output buffers.

ashleypittman:
Ensure test written from C subroutines in go is printed.

Required-githooks: true

### DIFF
--- a/src/control/cmd/daos/util.go
+++ b/src/control/cmd/daos/util.go
@@ -154,15 +154,20 @@ func createWriteStream(ctx context.Context, printLn func(line string)) (*C.FILE,
 
 		rdr := bufio.NewReader(r)
 		for {
-			line, err := rdr.ReadString('\n')
-			if err != nil {
-				if !(errors.Is(err, io.EOF) || errors.Is(err, os.ErrClosed)) {
-					printLn(fmt.Sprintf("read err: %s", err))
-				}
-				r.Close()
+			select {
+			case <-ctx.Done():
 				return
+			default:
+				line, err := rdr.ReadString('\n')
+				if err != nil {
+					if !(errors.Is(err, io.EOF) || errors.Is(err, os.ErrClosed)) {
+						printLn(fmt.Sprintf("read err: %s", err))
+					}
+					r.Close()
+					return
+				}
+				printLn(line)
 			}
-			printLn(line)
 		}
 	}(ctx)
 
--- a/src/control/cmd/daos/util.go
+++ b/src/control/cmd/daos/util.go
@@ -138,9 +138,8 @@ func freeString(str *C.char) {
 }
 
 func createWriteStream(ctx context.Context, printLn func(line string)) (*C.FILE, func(), error) {
-	// Create a FILE object for the handler to use for
-	// printing output or errors, and call the callback
-	// for each line.
+	// Create a FILE object for the handler to use for printing output or errors, and call the
+	// callback for each line.
 	r, w, err := os.Pipe()
 	if err != nil {
 		return nil, nil, err
@@ -169,7 +168,6 @@ func createWriteStream(ctx context.Context, printLn func(line string)) (*C.FILE,
 
 	return stream, func() {
 		C.fclose(stream)
-		w.Sync()
 		w.Close()
 	}, nil
 }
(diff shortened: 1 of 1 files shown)


src/control/cmd/daos/util.go
```diff
@@ -153,33 +152,36 @@ func createWriteStream(ctx context.Context, prefix string, printLn func(line str
...
+			if prefix == "" {
+				printLn(fmt.Sprintf("%s", line))
+			} else {
+				printLn(fmt.Sprintf("%s: %s", prefix, line))
 			}
 		}
 	}(ctx, prefix)
 
 	return stream, func() {
+		w.WriteString("close\n")
```
daltonbohning, line 179: This seems like a workaround. Is there not a more direct approach?

src/control/cmd/daos/util.go
```diff
@@ -153,33 +152,36 @@ func createWriteStream(ctx context.Context, prefix string, printLn func(line str
...
+		w.Sync()
+		C.fflush(stream)
+		C.fflush(stream)
 		C.fflush(stream)
```
daltonbohning, line 183: 3 flushes on the same stream?

ashleypittman: Old School.  Will revise.

src/control/cmd/daos/util.go
ashleypittman: Perhaps.  There are two things going on here, getting the consumer not the producer to shut the pipe but also to ensure that the pipe is not just flushed but also empty, and the only way I know how to do that is to send some kind of tag through it.

tanabarr: this seems okay as a temporary workaround IMO, we can follow-up with a more concrete fix

tanabarr (approved):

src/control/cmd/daos/util.go
```diff
@@ -153,33 +152,33 @@ func createWriteStream(ctx context.Context, prefix string, printLn func(line str
 	}
 
 	go func(ctx context.Context, prefix string) {
-		if prefix != "" {
-			prefix = ": "
-		}
```
ashleypittman, line 158: This is also a (different) bug fix, this code was obviously wrong.

src/control/cmd/daos/util.go
ashleypittman: I don't know what a more concrete fix would be, even using a semaphore to message from the producer to the consumer that it should exit there's still no way to flush a buffered pipe that I'm aware of.

ashleypittman:
I wasn't very happy with this so I've tweaked it a bit further.

src/control/cmd/daos/util.go
```diff
@@ -152,33 +150,24 @@ func createWriteStream(ctx context.Context, prefix string, printLn func(line str
...
+	go func(ctx context.Context) {
 
 		rdr := bufio.NewReader(r)
 		for {
-			select {
-			case <-ctx.Done():
-				return
```
kjacque, line 164: I'm not sure about removing this. Is this what's cutting off the output? The purpose of these lines is to exit this goroutine in the event that the user exits the program.

src/control/cmd/daos/util.go
```diff
@@ -152,33 +150,24 @@ func createWriteStream(ctx context.Context, prefix string, printLn func(line str
...
+				r.Close()
+				return
 			}
+			printLn(line)
 		}
-	}(ctx, prefix)
+	}(ctx)
 
 	return stream, func() {
-		C.fflush(stream)
```
kjacque, line 179: Don't need to force-flush anymore?

src/control/cmd/daos/util.go
ashleypittman: Will try re-adding, thank you for the review.

src/control/cmd/daos/util.go
ashleypittman: No.

tanabarr (approved):

kjacque (approved):

ashleypittman:
There's a typo in the commit message and I'm not at my laptop to fix it now.

ashleypittman:
Do not land this, further work needed.

ashleypittman:
This only partially worked so dropping in favour of https://github.com/daos-stack/daos/pull/12632 which is similar but uses a channel to communicate back success.

PR #12632, opened 2023-07-14
DAOS-11551 control: Flush c output buffer.

ashleypittman:
Update go code to flush and wait for completion on shutdown.

This avoids loosing any output from the c helper code.

### DIFF
--- a/src/control/cmd/daos/util.go
+++ b/src/control/cmd/daos/util.go
@@ -144,7 +144,7 @@ func createWriteStream(ctx context.Context, printLn func(line string)) (*C.FILE,
 	if err != nil {
 		return nil, nil, err
 	}
-	done := make(chan int)
+	done := make(chan bool)
 
 	stream, err := fd2FILE(w.Fd(), "w")
 	if err != nil {
@@ -152,7 +152,7 @@ func createWriteStream(ctx context.Context, printLn func(line string)) (*C.FILE,
 	}
 
 	go func(ctx context.Context) {
-		defer func() { done <- 1 }()
+		defer func() { done <- true }()
 
 		rdr := bufio.NewReader(r)
 		for {
--- a/src/control/cmd/daos/util.go
+++ b/src/control/cmd/daos/util.go
@@ -12,7 +12,7 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"time"
+	"sync"
 	"unsafe"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


src/control/cmd/daos/util.go
```diff
@@ -165,21 +164,20 @@ func createWriteStream(ctx context.Context, prefix string, printLn func(line str
 			default:
 				line, err := rdr.ReadString('\n')
 				if err != nil {
-					if !(errors.Is(err, io.EOF) || errors.Is(err, os.ErrClosed)) {
+					if !(errors.Is(err, io.EOF)) {
```
tanabarr, line 167: nit: unnecessary params

src/control/cmd/daos/util.go
```diff
@@ -165,21 +164,20 @@ func createWriteStream(ctx context.Context, prefix string, printLn func(line str
...
 		}
-	}(ctx, prefix)
+	}(ctx)
 
 	return stream, func() {
-		C.fflush(stream)
 		C.fclose(stream)
-		r.Close()
 		w.Close()
+		done.Lock()
```
tanabarr, line 180: I'm a bit confused by why you would lock the mutex within a "outCleanup" function. seems like an odd pattern.

src/control/cmd/daos/util.go
```diff
@@ -138,24 +138,23 @@ func freeString(str *C.char) {
...
 	if err != nil {
 		return nil, nil, err
 	}
 
-	go func(ctx context.Context, prefix string) {
-		if prefix != "" {
-			prefix = ": "
-		}
+	go func(ctx context.Context) {
+		defer done.Unlock()
```
tanabarr, line 157: normal pattern is to lock before deferring unlock, with this don't you risk unlocking multiple times?

ashleypittman: The lock is locked as it's created, and remains locked until this function (thread?) returns so I don't see how it would be unlocked multiple times?

src/control/cmd/daos/util.go
ashleypittman: I'm new to go.

src/control/cmd/daos/util.go
ashleypittman: It is an odd pattern, really in c a semaphore would be better.  I want to block the outCleanup() function from returning until the helper thread has completed draining the pipe which it won't do until it receives EOF.

The overall process is the lock is created and locked, when it comes to cleanup then the cleanup function closes the writer end of the pipe, the worker function keeps reading the pipe until it gets EOF at which point it returns, the lock is unlocked and cleanup completes.

This guarantees the pipe is empty before the context is torn down.

src/control/cmd/daos/util.go
```diff
@@ -138,24 +137,22 @@ func freeString(str *C.char) {
...
-	// printing output or errors, and call the callback
-	// for each line.
+func createWriteStream(ctx context.Context, printLn func(line string)) (*C.FILE, func(), error) {
+	// Create a FILE object for the handler to use for printing output or errors, and call the
+	// callback for each line.
 	r, w, err := os.Pipe()
 	if err != nil {
 		return nil, nil, err
 	}
+	done := make(chan bool)
```
tanabarr, line 147: So if you want to use a channel as a mutex, use a buffered struct{} one: `make(chan struct{}, 1)` as it is memory efficient and deterministic.

src/control/cmd/daos/util.go
```diff
@@ -138,24 +137,22 @@ func freeString(str *C.char) {
...
 	if err != nil {
 		return nil, nil, err
 	}
 
-	go func(ctx context.Context, prefix string) {
-		if prefix != "" {
-			prefix = ": "
-		}
+	go func(ctx context.Context) {
+		defer func() { done <- true }()
```
tanabarr, line 155: `done <- struct{}{}`, you can also just close the channel `close(done)`

tanabarr (dismissed):
looks good, just a minor suggestion

src/control/cmd/daos/util.go
ashleypittman: I'm not sure of the implication of that but I don't want buffering at this point.

ashleypittman: Ok, I'm reading up on this now.  It looks like this code as-is the write will not happen until the read is called so this could potentially deadlock if the fd is closed for any other reason and I should make the buffer at least one deep in case EOF is seen before the teardown function is called.

tanabarr (approved):

kjacque (approved):

ashleypittman:
Can this one be landed now please?

mchaarawi:
> Can this one be landed now please?

it has not completed testing

PR #12633, opened 2023-07-14
DAOS-12751 dfuse: Improve evict command.

ashleypittman:
Improve error cases where path is not backed by dfuse.
Lock forget commands against inode query.
Print and accept inodes as integers only.

Evict the build tree as part of the DaosBuild test.

Required-githooks: true

### DIFF
--- a/src/client/dfuse/ops/forget.c
+++ b/src/client/dfuse/ops/forget.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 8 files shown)


src/utils/daos_hdlr.c
```diff
@@ -2448,8 +2449,12 @@ dfuse_count_query(struct cmd_args_s *ap)
 	rc = ioctl(fd, DFUSE_IOCTL_COUNT_QUERY, &query);
 	if (rc < 0) {
 		rc = errno;
-		DH_PERROR_SYS(ap, rc, "ioctl failed");
-		rc = daos_errno2der(errno);
+		if (rc == ENOTTY) {
+			rc = -DER_MISC;
```
ashleypittman, line 2453: This doesn't' change the value of rc but it does remove the extra output.  I've filed https://daosio.atlassian.net/browse/DAOS-13977 to remove -DER_MISC across this command.

src/client/dfuse/ops/forget.c
```diff
@@ -8,58 +8,51 @@
...
-	int i;
+	struct dfuse_info *dfuse_info = fuse_req_userdata(req);
+	int                i;
 
 	fuse_reply_none(req);
 
-	DFUSE_TRA_DEBUG(fs_handle, "Forgetting %zi", count);
+	DFUSE_TRA_DEBUG(dfuse_info, "Forgetting %zi", count);
+
+	D_RWLOCK_RDLOCK(&dfuse_info->di_forget_lock);
```
daltonbohning, line 52: D_RWLOCK_RDLOCK, not D_RWLOCK_WRLOCK?
Is this because d_hash_rec_ndecref has its own lock?

src/tests/ftest/dfuse/daos_build.py
```diff
@@ -203,6 +203,9 @@ def run_build_test(self, cache_mode, intercept=False, dfuse_namespace=None):
                 'python3 -m pip install pip --upgrade',
                 'python3 -m pip install -r {}/requirements.txt'.format(build_dir),
                 'scons -C {} --jobs {} --build-deps=only'.format(build_dir, build_jobs),
+                'daos filesystem query {}'.format(mount_dir),
+                'daos filesystem evict {}'.format(build_dir),
+                'daos filesystem query {}'.format(mount_dir),
```
daltonbohning, line 208: Is this a fix or a workaround?

daltonbohning (approved):

src/client/dfuse/ops/forget.c
ashleypittman: I want forgets to happen in parallel as they do now so forget takes a reader lock (although it does modify - this is protected by the hash table) but I also want the ioctl to block whilst there are pending forget operations so that takes a writer lock.  It's the only way I could think of for the ioctl to pause which pending operations are happening.  The kernel seems to send a small number of forgetmany operation, each with a large number of forgets so this won't guarantee that the kernel has completed but it does at least pause the reader for any in-flight to complete.

src/tests/ftest/dfuse/daos_build.py
ashleypittman: Neither, this is to see how long the forget takes and see how effective it is.  Plus check that it doesn't deadlock of course.

daltonbohning (approved):

src/tests/ftest/dfuse/daos_build.py
ashleypittman: First results are in, all 3 commands took less than 0.00 seconds so the evict is fast, the first two commands reported 45065 inode in memory, the last one 4951.  I don't think this test should be putting anything other than the build_dir so it would be interesting to see if nodes does drop to 1, if we could run the command in json mode remotely and get the evicted inode number then we could wait for it but I think that's outside the scope of this PR.

ashleypittman: All tests had exactly the same values which shouldn't be a surprise, the files that aren't being evicted are the venv files which makes sense.

wiliamhuang (approved):

mchaarawi:
the results are missing for the hw medium stage. it seems to have been skipped in the CI run?

ashleypittman:
> the results are missing for the hw medium stage. it seems to have been skipped in the CI run?

It ran with Test-tag: dfuse

mchaarawi:
> > the results are missing for the hw medium stage. it seems to have been skipped in the CI run?
> 
> It ran with Test-tag: dfuse

shouldn't this be:
Feature: dfuse
since this is not a test only change?

ashleypittman:
> > > the results are missing for the hw medium stage. it seems to have been skipped in the CI run?
> > 
> > 
> > It ran with Test-tag: dfuse
> 
> shouldn't this be: Feature: dfuse since this is not a test only change?

`Features` rather than `Feature` but yes, I think that would have done more testing and be enough to have at least one test in every stage.  I'll update my processes.

mchaarawi:
> > > > the results are missing for the hw medium stage. it seems to have been skipped in the CI run?
> > > 
> > > 
> > > It ran with Test-tag: dfuse
> > 
> > 
> > shouldn't this be: Feature: dfuse since this is not a test only change?
> 
> `Features` rather than `Feature` but yes, I think that would have done more testing and be enough to have at least one test in every stage. I'll update my processes.

thanks. please re-add gatekeeper when the testing completes

PR #12635, opened 2023-07-14
DAOS-13906 engine: Check for memory allocation properly to avoid crash.

ashleypittman:
Fix some error handling code.

Required-githooks: true

### DIFF
--- a/src/cart/crt_iv.c
+++ b/src/cart/crt_iv.c
@@ -3305,8 +3305,11 @@ crt_iv_update_internal(crt_iv_namespace_t ivns, uint32_t class_id,
 		D_GOTO(exit, rc);
 	}
 
-	rc = iv_ops->ivo_on_get(ivns, iv_key,
-				0, CRT_IV_PERM_WRITE, NULL, &priv);
+	rc = iv_ops->ivo_on_get(ivns, iv_key, 0, CRT_IV_PERM_WRITE, NULL, &priv);
+	if (rc != 0) {
+		D_ERROR("ivo_on_get(): " DF_RC, DP_RC(rc));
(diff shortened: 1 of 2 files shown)


src/cart/crt_iv.c
```diff
@@ -3305,8 +3305,11 @@ crt_iv_update_internal(crt_iv_namespace_t ivns, uint32_t class_id,
 		D_GOTO(exit, rc);
 	}
 
-	rc = iv_ops->ivo_on_get(ivns, iv_key,
-				0, CRT_IV_PERM_WRITE, NULL, &priv);
+	rc = iv_ops->ivo_on_get(ivns, iv_key, 0, CRT_IV_PERM_WRITE, NULL, &priv);
+	if (rc != 0) {
+		D_ERROR("ivo_on_get(): " DF_RC, DP_RC(rc));
+		D_GOTO(exit, rc);
+	}
```
ashleypittman, line 3312: This wasn't something I could reproduce but it's also the only place that on_get() is called that isn't checked.

frostedcmos: I think it happened to work (or maybe work someitmes) due to:
- if call on 3308 fails then 'priv' wont be set (iv_value = NULL for this case, so no allocation is requested)
- calls on 3312 (ivo_on_update()) or 3316 (iov_on_refresh) are then passed that priv. Most likely those implementations check for priv being valid and would return appropriate rc, which is then checked on 3319.

The problem is that priv inst initialized to anything, so its still good to have this check (or alternatively set priv to NULL at init).

frostedcmos (approved):

soumagne (approved):

PR #12641, opened 2023-07-14
DAOS-13295 rebuild: reverse recx enumeration order for rebuild

wangdi1:
1. reverse recx enumeration order, so data from parity rebuild will be rebuilt first, then if any replicate data being left after EC aggregation, they will be ignored during rebuild.

2. Exclude EC aggregation with the whole rebuild process to avoid enumeration might enumerate any aggregating replicate recx.

Required-githooks: true
Features: rebuild

### Gatekeeper:

### DIFF
--- a/src/object/srv_obj.c
+++ b/src/object/srv_obj.c
@@ -2294,7 +2294,7 @@ obj_inflight_io_check(struct ds_cont_child *child, uint32_t opc, uint32_t flags)
 {
 	if (opc == DAOS_OBJ_RPC_ENUMERATE && flags & ORF_FOR_MIGRATION) {
 		if (child->sc_ec_agg_active) {
-			D_ERROR(DF_UUID" ec aggreate still active\n",
+			D_ERROR(DF_UUID" ec aggregate still active\n",
 				DP_UUID(child->sc_pool->spc_uuid));
 			return -DER_UPDATE_AGAIN;
 		}
--- a/src/object/srv_obj.c
+++ b/src/object/srv_obj.c
@@ -2292,6 +2292,14 @@ obj_ioc_init_oca(struct obj_io_context *ioc, daos_obj_id_t oid)
 static int
 obj_inflight_io_check(struct ds_cont_child *child, uint32_t opc, uint32_t flags)
 {
+	if (opc == DAOS_OBJ_RPC_ENUMERATE && flags & ORF_FOR_MIGRATION) {
+		if (child->sc_ec_agg_active) {
+			D_ERROR(DF_UUID" ec aggreate still active\n",
+				DP_UUID(child->sc_pool->spc_uuid));
+			return -DER_UPDATE_AGAIN;
+		}
+	}
+
 	if (!obj_is_modification_opc(opc))
 		return 0;
 
@@ -3137,6 +3145,9 @@ obj_local_enum(struct obj_io_context *ioc, crt_rpc_t *rpc,
 			param.ip_epc_expr = VOS_IT_EPC_RE;
 		}
 		recursive = true;
+
+		if (oei->oei_flags & ORF_DESCENDING_ORDER)
+			param.ip_flags |= VOS_IT_RECX_REVERSE;
 
 		if (daos_oclass_is_ec(&ioc->ioc_oca))
 			enum_arg->ec_cell_sz = ioc->ioc_oca.u.ec.e_len;
--- a/src/object/srv_obj_migrate.c
+++ b/src/object/srv_obj_migrate.c
@@ -1381,7 +1381,7 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
 						 min_eph,
 						 DIOF_FOR_MIGRATION | DIOF_EC_RECOV_FROM_PARITY,
 						 ds_cont);
-		if (rc < 0)
+		if (rc != 0)
 			D_GOTO(out, rc);
 	}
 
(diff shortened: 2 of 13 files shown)


src/object/srv_obj_migrate.c
```diff
@@ -1401,8 +1400,22 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
...
+					/* In some cases, EC aggregation failed to delete the
+					 * replicate recx, then during rebuild these replicate
+					 * recx might be rebuilt again. Since the data rebuilt
+					 * from parity will be rebuilt first. so let's ingore
+					 * this replicate recx for now.
+					 */
+					D_WARN(DF_UOID" "DF_RECX"/"DF_X64" already rebuilt\n",
+					       DP_UOID(mrone->mo_oid), DP_RECX(iod.iod_recxs[0]),
+					       mrone->mo_iods_update_ephs[i][j]);
+					rc = 0;
```
liuxuezhao, line 1414: since now rebuild and ec agg cannot happen concurrently, do we still need this handling?

NiuYawei (commented):
I think our design was to allow rebuild while aggregation is running, could you explain more why we have to change this design now?

src/object/srv_obj_migrate.c
```diff
@@ -1382,7 +1381,7 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
 						 min_eph,
 						 DIOF_FOR_MIGRATION | DIOF_EC_RECOV_FROM_PARITY,
 						 ds_cont);
-		if (rc > 0)
+		if (rc < 0)
```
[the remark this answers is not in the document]
wangdi1: yeah, rc != 0 should be fine.

src/object/srv_obj_migrate.c
wangdi1: hmm, if replica extent is left there for some reasons, then it might still see this issue, so we still need handling in this case.

wangdi1:
> I think our design was to allow rebuild while aggregation is running, could you explain more why we have to change this design now?

During EC rebuild,  once data shard is failed,  it needs to to rebuilt the data from partial extent,  and also rebuilt from the parity extent(data).   Even though their extents might be overlapped each other, but both of them needs to be rebuilt individually.

At the same time, EC aggregation might merge the partial (replicate) extent and parity extent if we do not exclude them, which might cause failure.

src/object/srv_obj_migrate.c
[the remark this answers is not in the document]
wangdi1: > yeah, rc != 0 should be fine.

can I refine this in the later patch? since the current patch already pass the test. and all failure code are < 0 anyway.

src/object/srv_obj.c
```diff
@@ -2270,6 +2270,14 @@ obj_ioc_init_oca(struct obj_io_context *ioc, daos_obj_id_t oid)
 static int
 obj_inflight_io_check(struct ds_cont_child *child, uint32_t opc, uint32_t flags)
 {
+	if (opc == DAOS_OBJ_RPC_ENUMERATE && flags & ORF_FOR_MIGRATION) {
+		if (child->sc_ec_agg_active) {
+			D_ERROR(DF_UUID" ec aggreate still active\n",
+				DP_UUID(child->sc_pool->spc_uuid));
+			return -DER_UPDATE_AGAIN;
```
liuxuezhao, line 2277: the err code UPDATE_AGAIN seems a bit confuse as it is for obj list operation.

src/object/srv_obj_migrate.c
```diff
@@ -2683,6 +2696,13 @@ migrate_one_epoch_object(daos_epoch_range_t *epr, struct migrate_pool_tls *tls,
 					DP_UOID(arg->oid));
 			}
 			continue;
+		} else if (rc == -DER_UPDATE_AGAIN) {
+			/* -DER_UPDATE_AGAIN means the remote target does not parse EC
+			 * aggregation yet, so let's retry.
+			 */
+			D_DEBUG(DB_REBUILD, DF_UOID "retry with %d \n", DP_UOID(arg->oid), rc);
+			rc = 0;
+			continue;
```
liuxuezhao, line 2705: my dss_sleep(10)?

liuxuezhao (approved):

src/object/srv_obj.c
wangdi1: hmm, just want to reuse the ERROR code a bit. though it is a confuse indeed. probably need redefine

src/object/srv_obj_migrate.c
wangdi1: EC aggregation should be aborted soon. and also this needs resend the new RPC anyway. since rebuild is a higher priority job, so it will need retry immediately.

src/object/srv_obj_migrate.c
liuxuezhao: @wangdi1 sorry you mean for the case that  for example during EC agg, it will 1) generate new parity extend, and 2) remove old replica extend. But somehow the step 2) is not done and rebuild cut in and here try to rebuild the replica extend? the build's local vos update epoch == the exist replica extends (left as step 2 did not remove it) and extend range overlapped with each other?

will the EC agg exit executing (after finish step 2) and then rebuild should not hit that case?

liuxuezhao (approved):

PR #12650, opened 2023-07-17
DAOS-11551 control: Flush c output buffer. (#12632)

ashleypittman:
Update go code to flush and wait for completion on shutdown.

This avoids loosing any output from the c helper code.

Required-githooks: true


ashleypittman:
Unclean as the master PR had a change to flush the file handles in daos_hdlr.c which was a previous failed attempt to resolve the issue, that change is not on 2.4 so does not need undoing.

kjacque (approved):

Issue #11, opened 2018-07-17
A couple issues with quickstart.md


TheOtherDave:
1) Dunno if this change would break anything, but the Quickstart file links to protobuf 3.5.1; 3.6.0 is the latest release.
2) The link to the forums is broken. Not sure if it'd be better to fix or remove, though, since the forums just say that everything's being moved to GitHub anyway (which is why I'm mentioning these here instead of there).

Issue #10, opened 2018-07-02
'daos_server' insta-crashes, and the offending line of code shouldn't even compile


TheOtherDave:
Hi, I'm trying to track down what has got to be the oddest thing I've come across... If I run the 'daos_server' command, I immediately get this error:
`DAOS I/O server exited with error: exit status 1`
Tracing through the source code, error is generated on line 199 of daos/src/iosrv/init.c:
```
	rc = crt_init(server_group_id,
		      CRT_FLAG_BIT_SERVER | CRT_FLAG_BIT_LM_DISABLE);
	if (rc)
		D_GOTO(exit_mod_init, rc);
```
'rc' gets set to -1003 here instead of 0. Grepping through the daos source tree shows that the only place it's defined is in daos/src/client/event.c, which defines it like this:
`#define crt_init(a,b,c)			({0;})`

Notice anything strange? 'crt_init' takes _three_ arguments, not the two arguments that are provided at the call site. Furthermore, if I paste that macro into a test program and assign its result to an int, I observe two other interesting things (both of which seem obvious, but I wanted to confirm). 1) it evaluates to 0, not -1003, and 2) its value doesn't depend on its arguments.

How is this code even compiling (in fact, it won't compile if I copy/paste it into a different test program until I add a third argument), let alone giving the value of -1003?

johannlombardi:
The crt_init() macro you found is only used for unit tests.
In fact, DAOS uses CaRT as the underlying RPC layer and crt_init() is defined here.
https://github.com/daos-stack/cart/blob/master/src/include/cart/api.h#L91
https://github.com/daos-stack/cart/blob/master/src/cart/crt_init.c#L182

As for the error codes, there are defined here:
https://github.com/daos-stack/cart/blob/master/src/include/gurt/errno.h#L62
1003 is DER_INVAL or invalid parameter.

Did you run the daos_server with orterun as specified in the quickstart guide?
https://github.com/daos-stack/daos/blob/master/doc/quickstart.md#running-daos

TheOtherDave:
Ah, ok, thanks. I wonder why VS Code thought it was a macro? That's gotta just be an issue on my machine.

WRT orterun, I've tried it both with and without... no change as far as this error is concerned. Orterun itself generates an error (something about not being able to register memory), but the error message indicates that it can deal with it. I'll get my laptop out and edit this to add the exact error.

TheOtherDave:
[...]

PR #12664, opened 2023-07-18
DAOS-13548 client: enable one NLT fault injection test for libpil4dfs

wiliamhuang:

* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for 

### Gatekeeper:


src/client/dfs/dfs.c
```diff
@@ -1063,6 +1063,9 @@ entry_stat(dfs_t *dfs, daos_handle_t th, daos_handle_t oh, const char *name, siz
 		} else {
 			daos_handle_t	file_oh;
 
+			/* Added to avoid issues in daos_hhash_link_lookup due to uninitialized
+			 * file_oh in NLT fault injection test */
+			memset(&file_oh, 0, sizeof(daos_handle_t));
```
ashleypittman, line 1068: There are several issues here:

1. Testing is to find bugs, when you fix one you're not "making the test pass", you're fixing a bug so you should reference the bug, not the test in any comment.
2. This comment includes technical information on a different part of the code-base, that might be appropriate for a commit message but it doesn't help future readers of this function understand the code in front of them.
3. Many daos functions, `daos_array_open_with_attr` included return a rc value and pass out the result via a arg, functions like this should always set the arg `oh` on the successful case and not use it on the unsuccessful case.  They should not expect the arg to be zeroed before being called.
4. If you did want to set file_oh to zeros then the correct way would be to change the declaration to be `daos_handle_t file_oh = {0}` or `daos_handle_t file_oh = {}`, I think the compiler would expect both.
5. Further to the 3. above `daos_array_open_with_attr` defines `of` as a param of type `out` so it's value at input should not matter.

What I think is happening here is that there's likely a bug in the error handling of array_open() which I think is exactly the type of bug this test is looking for.

Can you look at this again please and revert this change.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -274,9 +274,7 @@ rec_free(struct d_hash_table *htable, d_list_t *rlink)
 
 	rc = dfs_release(hdl->oh);
 	if (rc == ENOMEM)
-		rc = dfs_release(hdl->oh);
-	if (rc)
-		D_ERROR("dfs_release() failed: %d (%s)\n", rc, strerror(rc));
+		dfs_release(hdl->oh);
```
ashleypittman, line 277: Was there a problem with the code here before?

It's always a good idea to review logging and see if it's appropriate so if you're removing logging as you think it improves the end result that's fine.  The presence of logging itself however will not cause failures in NLT, as long as the messages are well-formed which this one looked to be.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -619,8 +621,10 @@ discover_daos_mount_with_env(void)
 		D_FATAL("DAOS_CONTAINER is not set.\n");
 		D_GOTO(out, rc = EINVAL);
 	}
-
 	D_STRNDUP(dfs_list[num_dfs].fs_root, fs_root, len_fs_root);
+	/* Added to avoid issues due to fault injection. Try again in case of ENOMEM */
+	if (dfs_list[num_dfs].fs_root == NULL)
+		D_STRNDUP(dfs_list[num_dfs].fs_root, fs_root, len_fs_root);
```
ashleypittman, line 627: This change also isn't right. The fault injection test in NLT doesn't check that a use-case or function works, but rather than if it doesn't then there's no undesired behaviour.  There's no need to put in hacks for this in order to make the test complete.

What happens if you simply return the error here, as the code did before?  As long as there's no crash and log lines are well-formed then that's enough to pass.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -680,6 +684,8 @@ discover_dfuse_mounts(void)
 			atomic_init(&pt_dfs_mt->inited, 0);
 			pt_dfs_mt->pool         = NULL;
 			D_STRNDUP(pt_dfs_mt->fs_root, fs_entry->mnt_dir, pt_dfs_mt->len_fs_root);
+			if (pt_dfs_mt->fs_root == NULL)
+				D_STRNDUP(pt_dfs_mt->fs_root, fs_entry->mnt_dir, pt_dfs_mt->len_fs_root);
```
ashleypittman, line 688: Same comment as above, although in this case the error handling in the function will also be called.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5492,13 +5500,17 @@ finalize_dfs(void)
 			continue;
 		}
 		rc = daos_cont_close(dfs_list[i].coh, NULL);
+		if (rc == -DER_NOMEM)
+			rc = daos_cont_close(dfs_list[i].coh, NULL);
```
ashleypittman, line 5504: This isn't desirable.  Since initially writing this test I've added logic to cover this case so that if DER_BUSY is logged then leak checking is disabled.  If we enable the new flag for this test then what would happen here is that dfs_umount() would fail with -DER_NOMEM, then daos_cont_close() and daos_pool_disconnect() would fail with -DER_BUSY and NLT would not run the leak checker in this case so the test would pass without this change.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5492,13 +5500,17 @@ finalize_dfs(void)
...
 		if (rc != 0) {
 			D_ERROR("error in daos_cont_close(%s): " DF_RC "\n", dfs_list[i].fs_root,
 				DP_RC(rc));
 			continue;
 		}
 		rc = daos_pool_disconnect(dfs_list[i].poh, NULL);
-		if (rc != 0) {
+		if (rc == -DER_NOMEM)
+			rc = daos_pool_disconnect(dfs_list[i].poh, NULL);
+		if (rc != -DER_SUCCESS) {
```
ashleypittman, line 5513: As above.

src/client/dfs/dfs.c
```diff
@@ -1078,6 +1081,10 @@ entry_stat(dfs_t *dfs, daos_handle_t th, daos_handle_t oh, const char *name, siz
 			}
 
 			rc = daos_array_close(file_oh, NULL);
+			/* Try again in case of out of memory. Needed to avoid memory leak in
+			 * NLT fault injection */
+			if (rc == -DER_NOMEM)
+				rc = daos_array_close(file_oh, NULL);
```
ashleypittman, line 1087: This again is an example of a bug in dfs that ideally we'd fix.  I'm not keen on this style of hack but there is a bug here that we should look at.

ashleypittman (commented):
Try adding `test_cmd.ignore_busy` to the `test_alloc_pil4dfs_ls` function to silence the shutdown errors.

For the dfs_release() calls I think maybe we should try another approach, for this PR for now can you keep moving forward and fixing the issues but perhaps when it comes to landing we'll just take the fixes, leave the test disabled and remove the `dfs_release` retries.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: The log here does not cause any failure. I will put it back. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: In this case, returning error "ENOMEM" from discover_daos_mount_with_env() leads to calling abort(). NLT fault injection will fail.

src/client/dfs/dfs.c
wiliamhuang: https://github.com/daos-stack/daos/blob/master/src/client/api/array.c#L118-L120
When the fault injection causes dc_task_create() returns -DER_NOMEM, dc_task_schedule() is not called. The leaked memory was allocated by "D_ALLOC_PTR(array);" at line 91 in array_alloc() in dc_array.c,
https://github.com/daos-stack/daos/blob/master/src/client/array/dc_array.c#L91

Shall we retry "dc_task_create(dc_array_close, NULL, ev, &task);" in case it returns -DER_NOMEM inside daos_array_close()?

wiliamhuang: If we retry "dc_task_create(dc_array_close, NULL, ev, &task);" in case it returns -DER_NOMEM inside daos_array_close(), we need to retry "dc_task_schedule(task, true)" too. How do you want to proceed? Thank you!

src/client/dfs/dfs.c
wiliamhuang: Thank you! Initialize oh.cookie with an invalid handle in dc_obj_open() now.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: I removed the retry of D_STRNDUP() here.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Removed as suggested. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Removed as suggested. Thank you!

src/client/dfs/dfs.c
wiliamhuang: ok. What should we do about the caused memory leak due to fault injection?

ashleypittman: > such hacks are really getting out of hand in the code. let's not add more please.

I agree, however this is a bug and it's impossible to address it at a higher level than dfs as the context has been lost by that point.  Outside the scope of this PR however.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: The correct thing for pil4dfs to do in case of startup errors (of which this is an example) is to fail with an abort() and a helpful error message.  If you need to then update NLT to detect that error message and not check the exit code in that case.  We typically still try and resolve all memory leaks however.

If you take the approach I've described then it means the product behaves in a defined way, and the testing is verifying that.  Adding this kind of hack is making the product worst to appease a test which is the wrong approach.

src/object/cli_obj.c
```diff
@@ -1414,6 +1414,8 @@ dc_obj_open(tse_task_t *task)
 
 	args = dc_task_get_args(task);
 	D_ASSERTF(args != NULL, "Task Argument OPC does not match DC OPC\n");
+	/* initialize with an invalid handle */
+	(*args->oh).cookie = 0;
```
ashleypittman, line 1418: Much better.  I still can't see where this was being used undefined, can you point it out to me?

wiliamhuang: Here is the calling stack when the handle is accessed in dc_obj_close(). 

#0  d_hhash_link_lookup (hhash=0x55b881a7e4d0, key=<optimized out>) at src/gurt/hash.c:1269
#1  0x00007fa691f77923 in daos_hhash_link_lookup (key=<optimized out>) at src/common/misc.c:604
#2  0x00007fa6922dc6e0 in obj_hdl2ptr (oh=...) at src/object/cli_obj.c:171
#3  0x00007fa6922f06bc in dc_obj_close (task=0x55b881a7b940) at src/object/cli_obj.c:1496
#4  0x00007fa691fbcedc in tse_task_schedule_with_delay (task=0x55b881a7b940, instant=instant@entry=true, delay=delay@entry=0) at src/common/tse.c:1056
#5  0x00007fa691fbd2e9 in tse_task_schedule (task=<optimized out>, instant=instant@entry=true) at src/common/tse.c:1070
#6  0x00007fa692393342 in open_handle_cb (task=0x55b882040be0, data=<optimized out>) at src/client/array/dc_array.c:675
#7  0x00007fa691fb6745 in tse_task_complete_callback (task=task@entry=0x55b882040be0) at src/common/tse.c:520
#8  0x00007fa691fb7583 in tse_task_post_process (task=task@entry=0x55b8820536f0) at src/common/tse.c:675
#9  0x00007fa691fbae62 in tse_sched_process_complete (dsp=dsp@entry=0x7fa6925fb350 <daos_sched_g+16>) at src/common/tse.c:727
#10 0x00007fa691fbc5ff in tse_task_complete (task=task@entry=0x55b8820536f0, ret=ret@entry=-1009) at src/common/tse.c:885
#11 0x00007fa6922f11fd in dc_obj_open (task=0x55b8820536f0) at src/object/cli_obj.c:1467
#12 0x00007fa691fbcedc in tse_task_schedule_with_delay (task=0x55b8820536f0, instant=instant@entry=true, delay=delay@entry=0) at src/common/tse.c:1056
#13 0x00007fa691fbd2e9 in tse_task_schedule (task=<optimized out>, instant=instant@entry=true) at src/common/tse.c:1070
#14 0x00007fa69239471b in dc_array_open (task=0x55b882040be0) at src/client/array/dc_array.c:768
#15 0x00007fa691fbcedc in tse_task_schedule_with_delay (task=task@entry=0x55b882040be0, instant=instant@entry=true, delay=delay@entry=0) at src/common/tse.c:1056
#16 0x00007fa691fbd2e9 in tse_task_schedule (task=task@entry=0x55b882040be0, instant=instant@entry=true) at src/common/tse.c:1070
#17 0x00007fa6922547ac in dc_task_schedule (task=0x55b882040be0, instant=instant@entry=true) at src/client/api/task.c:114
#18 0x00007fa6922369c2 in daos_array_open_with_attr (coh=..., oid=..., th=th@entry=..., mode=mode@entry=2, cell_size=<optimized out>, cell_size@entry=1, chunk_size=<optimized out>, oh=0x7ffe3f702c18, ev=0x0) at src/client/api/array.c:95
#19 0x00007fa691d1972d in entry_stat (dfs=dfs@entry=0x55b882033df0, th=th@entry=..., oh=..., name=name@entry=0x7ffe3f702d70 "file.4", len=<optimized out>, obj=obj@entry=0x0, get_size=<optimized out>, stbuf=<optimized out>, obj_hlc=<optimized out>)
    at src/client/dfs/dfs.c:1069

ashleypittman: That looks to be at https://github.com/daos-stack/daos/blob/master/src/client/array/dc_array.c#L675 which is in error handling code, do you know what allocation failed to get to that point?  Perhaps https://github.com/daos-stack/daos/blob/master/src/client/array/dc_array.c#L618 should just return instead?

wiliamhuang: The allocation target of fault injection is,
https://github.com/daos-stack/daos/blob/master/src/object/cli_obj.c#L143

#1  0x00007f85ce6d8234 in d_should_fail (fault_attr=0x55585b488230) at src/gurt/fault_inject.c:792
#2  0x00007f85d080ff11 in obj_alloc () at src/object/cli_obj.c:143
#3  0x00007f85d081bb52 in dc_obj_open (task=0x55585ba71bd0) at src/object/cli_obj.c:1420
#4  0x00007f85d04eb3f4 in tse_task_schedule_with_delay (task=0x55585ba71bd0, instant=true, delay=0)
    at src/common/tse.c:1056
#5  0x00007f85d04eb445 in tse_task_schedule (task=0x55585ba71bd0, instant=true) at src/common/tse.c:1070
#6  0x00007f85d08ecd5e in dc_array_open (task=0x55585ba5ef90) at src/client/array/dc_array.c:768
#7  0x00007f85d04eb3f4 in tse_task_schedule_with_delay (task=0x55585ba5ef90, instant=true, delay=0)
    at src/common/tse.c:1056
#8  0x00007f85d04eb445 in tse_task_schedule (task=0x55585ba5ef90, instant=true) at src/common/tse.c:1070
#9  0x00007f85d07855c3 in dc_task_schedule (task=0x55585ba5ef90, instant=true) at src/client/api/task.c:114
#10 0x00007f85d0764ed1 in daos_array_open_with_attr (coh=..., oid=..., th=..., mode=2, cell_size=1, chunk_size=1048576,
    oh=0x7ffdf170ba60, ev=0x0) at src/client/api/array.c:95
#11 0x00007f85d0242de2 in entry_stat (dfs=0x55585ba52000, th=..., oh=..., name=0x7ffdf170bbe0 "file.4", len=6, obj=0x0,
    get_size=true, stbuf=0x7ffdf170bd40, obj_hlc=0x0) at src/client/dfs/dfs.c:1066
#12 0x00007f85d025f5b1 in dfs_stat (dfs=0x55585ba52000, parent=0x55585ba71a80, name=0x7ffdf170bbe0 "file.4",
    stbuf=0x7ffdf170bd40) at src/client/dfs/dfs.c:4918
#13 0x00007f85d1580dfa in new_lxstat (ver=1, path=0x55585b4996e0 "/tmp/pil4dfs_mountsx0ils0g/new_dir/file.4",
    stat_buf=0x7ffdf170bd40) at src/client/dfuse/pil4dfs/int_dfs.c:2242
#14 0x00007f85d1581485 in statx (dirfd=-100, path=0x7ffdf170bf10 "new_dir/file.4", flags=256, mask=606,
    statx_buf=0x7ffdf170bdf0) at src/client/dfuse/pil4dfs/int_dfs.c:2357
#15 0x000055585a03183b in do_statx ()
#16 0x000055585a0356f5 in gobble_file.constprop ()
#17 0x000055585a0368af in print_dir ()
#18 0x000055585a03019d in main ()

wiliamhuang: https://github.com/daos-stack/daos/blob/master/src/client/array/dc_array.c#L661

open_handle_cb() is triggered to call dc_obj_close(). In case of uninitialized "oh", the content is non-zero. "daos_handle_is_valid(*args->oh)" returns true.

ashleypittman: is this from the goto on L618?  If so change that to a return?

wiliamhuang: Right. Sounds good!

wiliamhuang: @ashleypittman Is there other possible types of errors other than "out of memory" which do need to call "DAOS_OPC_OBJ_CLOSE" (having a valid handle)? If yes, initialization oh could be safer.

ashleypittman: The error type isn't relevant here is it?

wiliamhuang: If a non-zero rc means the handle is invalid, then the error type is not relevant.

ashleypittman: Or rather, there's been an error so the handles not valid so it doesn't need closing.  What the error is is not a factor.

wiliamhuang: Since I do not know how long time this PR can be landed, I created a separate ticket DAOS-14030 and PR https://github.com/daos-stack/daos/pull/12691 to address this issue.

ashleypittman: Can you add it to this PR to see if it passes the test?  Happy to land it separately but I'm like to see it tested here.

wiliamhuang: ok. I will add it to this PR soon. It does help to avoid the issue encountered in close in my test on boro.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: @ashleypittman @mchaarawi 
		rc = daos_cont_close(dfs_list[i].coh, NULL);
		if (rc != -DER_SUCCESS && rc != -DER_NONEXIST)
			rc = daos_cont_close(dfs_list[i].coh, NULL);

		rc = daos_pool_disconnect(dfs_list[i].poh, NULL);
		if (rc != -DER_SUCCESS && rc != -DER_NONEXIST)
			rc = daos_pool_disconnect(dfs_list[i].poh, NULL);

Shall we retry daos_cont_close() and daos_pool_disconnect when they fail? Is there resource leak on server side without retry on client side?

The fault injections causes client-side memory leak are in tse_task_create(), 
https://github.com/daos-stack/daos/blob/master/src/common/tse.c#L961

Calling stacks,
daos_cont_close() failed
DBG> 0, /scratch/leihuan1/tickets/13548/daos/install/lib64/libgurt.so.4(print_trace+0x67) [0x7f84b1ea0886]
DBG> 1, /scratch/leihuan1/tickets/13548/daos/install/lib64/libgurt.so.4(d_should_fail+0x574) [0x7f84b1ea0ee7]
DBG> 2, /scratch/leihuan1/tickets/13548/daos/install/lib64/libdaos_common.so(tse_task_create+0x5a) [0x7f84b3cb389e]
DBG> 3, /scratch/leihuan1/tickets/13548/daos/install/lib64/libdaos.so.2(dc_task_create+0x80) [0x7f84b3f4e4cd]
DBG> 4, /scratch/leihuan1/tickets/13548/daos/install/lib64/libdaos.so.2(daos_cont_close+0x1dd) [0x7f84b3f2f9df]
DBG> 5, /scratch/leihuan1/tickets/13548/daos/install/lib64/libpil4dfs.so(+0x250ba) [0x7f84b4d5f0ba]
DBG> 6, /scratch/leihuan1/tickets/13548/daos/install/lib64/libpil4dfs.so(+0x23992) [0x7f84b4d5d992]
DBG> 7, /lib64/ld-linux-x86-64.so.2(+0x151e6) [0x7f84b4f8e1e6]
DBG> 8, /lib64/libc.so.6(+0x5123c) [0x7f84b459423c]
DBG> 9, /lib64/libc.so.6(on_exit+0) [0x7f84b4594370]
DBG> 10, /scratch/leihuan1/tickets/13548/daos/install/lib64/libpil4dfs.so(+0x22015) [0x7f84b4d5c015]
DBG> 11, /lib64/libc.so.6(__libc_start_main+0xfa) [0x7f84b457dcfa]
DBG> 12, ls(+0x5e7e) [0x561f33b92e7e]

daos_pool_disconnect() failed
DBG> 0, /scratch/leihuan1/tickets/13548/daos/install/lib64/libgurt.so.4(print_trace+0x67) [0x7f315e085886]
DBG> 1, /scratch/leihuan1/tickets/13548/daos/install/lib64/libgurt.so.4(d_should_fail+0x574) [0x7f315e085ee7]
DBG> 2, /scratch/leihuan1/tickets/13548/daos/install/lib64/libdaos_common.so(tse_task_create+0x5a) [0x7f315fe9889e]
DBG> 3, /scratch/leihuan1/tickets/13548/daos/install/lib64/libdaos.so.2(dc_task_create+0x80) [0x7f31601334cd]
DBG> 4, /scratch/leihuan1/tickets/13548/daos/install/lib64/libdaos.so.2(daos_pool_disconnect+0x1dd) [0x7f316012f3ae]
DBG> 5, /scratch/leihuan1/tickets/13548/daos/install/lib64/libpil4dfs.so(+0x25269) [0x7f3160f44269]
DBG> 6, /scratch/leihuan1/tickets/13548/daos/install/lib64/libpil4dfs.so(+0x23992) [0x7f3160f42992]
DBG> 7, /lib64/ld-linux-x86-64.so.2(+0x151e6) [0x7f31611731e6]
DBG> 8, /lib64/libc.so.6(+0x5123c) [0x7f316077923c]
DBG> 9, /lib64/libc.so.6(on_exit+0) [0x7f3160779370]
DBG> 10, /scratch/leihuan1/tickets/13548/daos/install/lib64/libpil4dfs.so(+0x22015) [0x7f3160f41015]
DBG> 11, /lib64/libc.so.6(__libc_start_main+0xfa) [0x7f3160762cfa]
DBG> 12, ls(+0x5e7e) [0x55b1934ace7e]

wiliamhuang: @ashleypittman @mchaarawi RPC is needed in daos_cont_close() and daos_pool_disconnect(), right? How we should proceed here? It looks retries exist in several other places. Shall I also adopt retry? Thank you!

mchaarawi: i cannot do a similar "fix" that i did for object close since as you said those are not local operations and require a tse task to be attached to the operation / RPC. 
I also do not think we should add more retries / hacks like this and should work on removing them. I'm not sure what can be done here other than not do fault injection on this path.

wiliamhuang: @mchaarawi I see. Thank you very much for your comments!

PR #12677, opened 2023-07-20
DAOS-0000 dfs: Keep a handle on dfs objects if release fails.

ashleypittman:
Rather than return an error code from release keep if the operation
fails then hang it on a list from the dfs object itself.
This removes the need to check the return code from dfs_release
and tidies up a lot of code.

The only places that were doing anything with the return code
were simply re-trying in order to pass the NLT test.

This PR highlights an issue with dfuse that inodes are not
closed in an orderly manner so therefore dfs can be unmounted
before all the inodes are releaseed.

Required-githooks: true


ashleypittman:
Closing in favour of https://github.com/daos-stack/daos/pull/12687

PR #12683, opened 2023-07-21
DAOS-13996 pool: Fix invalid D_FREE in pool_glance

liw:
pool_glance should only free map_buf if ds_pool_svc_load returns zero. The segfault might be triggered because ds_pool_svc_load returned DER_UNINIT; it is unknown that what actually happened. This patch fixes the D_FREE logic.

ds_pool_svc_load does not need to free map_buf in any case. It is likely a merge error. This patch removes the unnecessary D_FREE in ds_pool_svc_load.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -1540,8 +1540,6 @@ check_map:
 	*global_version_out = global_version;
 	*map_buf_out = map_buf;
 	*map_version_out = map_version;
-	if (rc != 0)
-		D_FREE(map_buf);
 out:
 	return rc;
 }
--- a/src/pool/srv_pool_check.c
+++ b/src/pool/srv_pool_check.c
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2022 Intel Corporation.
+ * (C) Copyright 2022-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -84,7 +84,9 @@ pool_glance(uuid_t uuid, char *path, struct ds_pool_clue *clue_out)
 	}
 
 	rc = ds_pool_svc_load(&tx, uuid, &root, &global_version, &map_buf, &clue.psc_map_version);
-	if (rc == DER_UNINIT) {
+	if (rc == 0) {
+		D_FREE(map_buf);
+	} else if (rc == DER_UNINIT) {
 		clue.psc_map_version = 0;
 		rc = 0;
 	} else if (rc != 0) {
@@ -92,7 +94,6 @@ pool_glance(uuid_t uuid, char *path, struct ds_pool_clue *clue_out)
 	}
 
 	memcpy(clue_out->pc_svc_clue, &clue, sizeof(clue));
-	D_FREE(map_buf);
 out_label:
 	if (rc != 0) {
 		D_FREE(clue_out->pc_label);


src/pool/srv_pool.c
```diff
@@ -1540,8 +1540,6 @@ ds_pool_svc_load(struct rdb_tx *tx, uuid_t uuid, rdb_path_t *root, uint32_t *glo
 	*global_version_out = global_version;
 	*map_buf_out = map_buf;
 	*map_version_out = map_version;
-	if (rc != 0)
-		D_FREE(map_buf);
```
liw, line 1544: [Note] This is probably caused by previous merged with master.

src/pool/srv_pool_check.c
```diff
@@ -84,15 +84,16 @@ pool_glance(uuid_t uuid, char *path, struct ds_pool_clue *clue_out)
 	}
 
 	rc = ds_pool_svc_load(&tx, uuid, &root, &global_version, &map_buf, &clue.psc_map_version);
-	if (rc == DER_UNINIT) {
+	if (rc == 0) {
+		D_FREE(map_buf);
+	} else if (rc == DER_UNINIT) {
```
Nasf-Fan, line 89: Just want to know why ds_pool_svc_load() return positive number for empty pool map instead of "-DER_UNINIT"?

liw: Because that is a important, non-error scenario that will almost 100% happen during the creation of a PS, and I was afraid that functions called by ds_pool_svc_load may return `-DER_UNINIT` for other unrelated scenarios. Hence, I chose a positive return value many years ago.

If I were to make the same choice again today, I might prefer defining an `enum` with a positive value.

Nasf-Fan: > Because that is a important, non-error scenario that will almost 100% happen during the creation of a PS, and I was afraid that functions called by ds_pool_svc_load may return `-DER_UNINIT` for other unrelated scenarios. Hence, I chose a positive return value many years ago.
> 
> If I were to make the same choice again today, I might prefer defining an `enum` with a positive value.

I think you can do that in master independently.

Nasf-Fan (approved):

src/pool/srv_pool.c
liuxuezhao: or set "*map_version_out = NULL" at above L1519 (rc != 0) case, seems in that case the map_buf ptr in stack possibly not init.

liuxuezhao (approved):

src/pool/srv_pool.c
liw: @liuxuezhao, it is intentional that `ds_pool_svc_load` changes `*map_version_out` **only when returning zero**. Callers shall only expect an output in `*map_version_out` if `ds_pool_svc_load` returns zero.

liw:
Thanks to the reviewers for the quick responses!

Nasf-Fan:
@liw , what is the next step for this patch? Do you need more reviewer or any further investigation? The patch will be helpful for CR tests on CI. Thanks!

liw:
@Nasf-Fan, this PR is hitting CR and non-CR test failures. If those test failures don't go away, how does this PR proceed?

Nasf-Fan:
> @Nasf-Fan, this PR is hitting CR and non-CR test failures. If those test failures don't go away, how does this PR proceed?

I did not want to push the landing with break regular land process.
I hit DAOS-13996 corruption many times during both CR CI test and manual test, so just to make sure the gatekeeper can be aware of the land requirement in time since quite busy recently.

liw:
The rebuild_simple failure is unrelated to this PR; the pool_list_consolidation failure involves multiple problems that are likely unrelated to this PR. We need to investigate the latter.

liw:
Discussed with Fan Yong the situation. We think the pool_list_consolidation is first a test issue (a ticket will be created in a minute) and unrelated to this PR.

gnailzenh (approved):

PR #12702, opened 2023-07-24
DAOS-14032 log: Add facility error counters

mjmac:
Expose per-facility error counters in each engine's
telemetry. When an error or higher log entry is made,
the associated counter for that facility will be
incremented. This can be used by alerting solutions
to detect events based on error rates.

Required-githooks: true


src/gurt/dlog.c
```diff
@@ -143,6 +147,43 @@ static const char *clog_pristr(int pri)
...
+
+	clog_lock();
+
+	if (!d_log_xst.dlog_facs)
+		D_GOTO(unlock, -DER_UNINIT);
+
+	if (d_log_xst.fac_cnt == 0)
+		D_GOTO(unlock, 0);
+
+	nctrs = calloc(1, d_log_xst.fac_cnt * sizeof(*nctrs));
```
ashleypittman, line 169: This memory is presumably lost?  Probably not a problem in practice but NLT tests that no memory is lost.

src/gurt/dlog.c
```diff
@@ -631,6 +672,15 @@ void d_vlog(int flags, const char *fmt, va_list ap)
 		clog_unlock();
 		return;
 	}
+	/* 
+	 * If logging an error or higher, increment the error counter for
+	 * the facility. This can be used by monitoring solutions to detect
+	 * increased error rates.
+	 * 
+	 * TODO: Count by errno too? Might be too expensive.
```
ashleypittman, line 680: It wouldn't be possible to count by error no just yet.  I have in mind to re-write the logging macros to take "rc" as a parameter but for now logging is just printf so extracting the errno won't be possible.   Re-writing every logging call (or at least D_ERROR + D_WARN) is a big operation however so I want to get it right.

I've started on https://github.com/daos-stack/daos/pull/12575 which is the first step to allow us to make the change programatically but it's early days.  Once we've defined the new interface and got the tools then we've got the task of migrating the code and that'll be disruptive - hence the desire to do it slowly and correctly the first time.

ashleypittman:
> Keeping this as a draft for the moment. This may be a terrible idea, but it seems to work fine. I was able to add a Grafana panel to display error rates, and I can see it being useful for other monitoring solutions.

NLT already displays this data for a given log file after the run, one issue is that different areas of the code log with different frequencies, vos for example is very "chatty" so `frequency` of errors as a detection tool will miss many cases.

mjmac:
> > Keeping this as a draft for the moment. This may be a terrible idea, but it seems to work fine. I was able to add a Grafana panel to display error rates, and I can see it being useful for other monitoring solutions.
> 
> NLT already displays this data for a given log file after the run, one issue is that different areas of the code log with different frequencies, vos for example is very "chatty" so `frequency` of errors as a detection tool will miss many cases.

The idea here is to allow realtime monitoring via telemetry (e.g. Prometheus+Grafana, or LDMS, etc) so that an admin can set up automated alerts without having to deal with log scraping. The approach in this PR would be a crude tool, but I think it's probably better than nothing. The updated logging macros with rc captured as a parameter would allow for finer-grained control.

jolivier23 (approved):

PR #12708, opened 2023-07-25
DAOS-13919 control: Fix dmg pool get-prop compat with v2.2 pools

tanabarr:
JSON marshalling fails when attempting to print PoolGetProps()
control-API response as some of the returned properties have no
associated value set. This can occur when dmg pool get-prop is called
from a >= v2.4 dmg/daos_server using a pool that was created with v2.2.
This occurs because the old pool has a reduced set of properties and so
those that don't have an associated value in the RPC response end up
as PoolProperty objects with no associated value (which is not allowed
in the JSON marshal implementation for the type).

Solution is to remove the missing properties from the response before
sending to dmg, the control-API consumer. This does mean that
get-props calls can return a different property list depending on
which DAOS version was used to create the pool.

Features: control
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/lib/control/pool.go
+++ b/src/control/lib/control/pool.go
@@ -833,7 +833,6 @@ func PoolGetProp(ctx context.Context, rpcClient UnaryInvoker, req *PoolGetPropRe
 		return nil, errors.New("unable to extract PoolGetPropResp from MS response")
 	}
 
-	resp := req.Properties
 	pbMap := make(map[uint32]*mgmtpb.PoolProperty)
 	for _, prop := range pbResp.GetProperties() {
 		if _, found := pbMap[prop.GetNumber()]; found {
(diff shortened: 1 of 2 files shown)


src/control/lib/control/pool.go
```diff
@@ -842,10 +841,12 @@ func PoolGetProp(ctx context.Context, rpcClient UnaryInvoker, req *PoolGetPropRe
 		pbMap[prop.GetNumber()] = prop
 	}
 
-	for _, prop := range resp {
+	resp := make([]*daos.PoolProperty, 0, len(req.Properties))
+	for _, prop := range req.Properties {
 		pbProp, found := pbMap[prop.Number]
 		if !found {
-			rpcClient.Debugf("DAOS-11418: Unable to find prop %d (%s) in resp", prop.Number, prop.Name)
+			// Properties can be missing due to DAOS-11418 and DAOS-13919
```
tanabarr, line 848: @knard-intel please can you have a look to see if this fix will disrupt the solution for DAOS-11418 as the change will remove properties that are missing from the response.

src/control/lib/control/pool_test.go
```diff
@@ -1244,6 +1282,94 @@ func TestPoolGetProp(t *testing.T) {
...
+		},
+		"missing props; v2_2 pool": {
```
mjmac, line 1333: I'm a little confused by this test case. Is the error expected because in the normal case, the properties with no value should be filtered out from the response?

tanabarr: yes this is the error that caused the issue we are trying to solve, this verifies that Unmarshal doesn't allow missing values for existing properties

mjmac (approved):

src/control/lib/control/pool.go
knard-intel: From my understanding, this should not disrupt the functional tests as we are not using non homogeneous version of DAOS in the test case: creating pool with 22. versions and then using it with a 2.4 versions of dmg.
Moreover, if a end-user is facing such use case, I prefer that the fields which have been renamed do not appear rather than having invalid values.

knard-intel (approved):

tanabarr:
Manual verification of this fix has been performed and documented in [the DAOS-13919 JIRA ticket](https://daosio.atlassian.net/browse/DAOS-13919).

PR #12717, opened 2023-07-26
DAOS-14064 control: INFO logging for ramdisk/hugepages

mjmac:
Log the following at INFO level so that they are visible
in production configurations:
  * Auto-calculated per-engine ramdisk size
  * Auto-caculated hugepage count

Required-githooks: true

### DIFF
--- a/src/control/server/config/server.go
+++ b/src/control/server/config/server.go
@@ -448,6 +448,10 @@ func getAccessPointAddrWithPort(log logging.Logger, addr string, portDefault int
 	return addr, nil
 }
 
+func hugePageBytes(hpNr, hpSz int) uint64 {
+	return uint64(hpNr*hpSz) * humanize.KiByte
+}
+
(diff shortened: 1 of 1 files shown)


tanabarr (approved):

src/control/server/config/server.go
```diff
@@ -590,6 +596,8 @@ func (cfg *Server) SetRamdiskSize(log logging.Logger, mi *common.MemInfo) error
 			// Apply calculated size in config as not already set.
 			log.Debugf("%s: auto-sized ram-disk in engine-%d config", msg, idx)
 			scs[0].WithScmRamdiskSize(uint(maxRamdiskSize / humanize.GiByte))
+			log.Infof("engine-%d: ramdisk size automatically set to %s", idx,
```
knard-intel, line 599: Nit, for sustaining task support it could be useful to have the size in bytes as at line 500.

knard-intel (approved):

PR #12719, opened 2023-07-26
DAOS-13988 control: MemberResult state should reflect error

tanabarr:
Change addresses the issue where dmg system stop fails if some ranks
have already been previously stopped. This is due to a couple of bugs.
First a situation exists where the state conveyed in a MemberResult
doesn't match the fact that the result was a failure. The consumer of
the results checks this and raises an error in the case that they don't
match. Secondly the PrepShutdown flow which is triggered as stage-one
of dmg system stop returns a failed result for an already stopped rank.

Fixes made:
- Return successful results when calling PrepShutdown on stopped ranks
- Ensure errored MemberResults convey a matching error state

### Gatekeeper:

### DIFF
--- a/src/control/system/member.go
+++ b/src/control/system/member.go
@@ -139,15 +139,6 @@ func (ms MemberState) isTransitionIllegal(to MemberState) bool {
 		MemberStateJoined: {
 			MemberStateReady: true,
 		},
-		MemberStateStopping: {
-			MemberStateReady: true,
-		},
-		MemberStateExcluded: {
-			MemberStateReady:    true,
-			MemberStateJoined:   true,
-			MemberStateStopping: true,
-			MemberStateErrored:  true,
-		},
 		MemberStateErrored: {
 			MemberStateReady:    true,
 			MemberStateJoined:   true,
@@ -346,6 +337,10 @@ func NewMemberResult(rank ranklist.Rank, err error, state MemberState, action ..
 	if err != nil {
 		result.Errored = true
 		result.Msg = err.Error()
+		// Any error should result in either an unresponsive or errored state.
+		if state != MemberStateUnresponsive {
+			result.State = MemberStateErrored
+		}
 	}
 	if len(action) > 0 {
 		result.Action = action[0]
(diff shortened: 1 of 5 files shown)


knard-intel (dismissed):
Seems good from what I understood of the ticket discussion on the state machine transition.

tanabarr:
investigating issues with PR found during manual verification

tanabarr:
CI failure is unrelated and identified as https://daosio.atlassian.net/browse/DAOS-14072, requesting forced landing

tanabarr:
> The bugfix looks fine to me. While I am generally in favor of code cleanup, adding it to a PR that is going to be backported to 2.4 increases the risk. It would be safer to just isolate the bugfix into a smaller chunk and then land the cleanup to master in a subsequent PR.

@mjmac the only code modification not relevant to the fix is removing unused code, I don't think that incurs any risk, do you?

mjmac:
> @mjmac the only code modification not relevant to the fix is removing unused code, I don't think that incurs any risk, do you?

If there's no way that a functional test could somehow still be using that endpoint, then I guess there's very little risk. I don't know for certain, though, and that's my point. For patches going into release branches, especially at the last moment, good discipline would dictate that the changes are kept to the minimum necessary to fix the problem. I'll give this a +1 because I agree that it's probably fine and the fix should get into 2.4, but in the future please refrain from making nonessential changes on release patches.

mjmac (dismissed):

src/control/system/member.go
```diff
@@ -139,15 +139,6 @@ func (ms MemberState) isTransitionIllegal(to MemberState) bool {
...
 		},
-		MemberStateStopping: {
-			MemberStateReady: true,
-		},
-		MemberStateExcluded: {
-			MemberStateReady:    true,
-			MemberStateJoined:   true,
-			MemberStateStopping: true,
-			MemberStateErrored:  true,
-		},
```
kjacque, line 150: Aren't some of these still illegal transitions? I think the only one that needs to be removed is MemberStateExcluded -> MemberStateStopping. Unless I've misunderstood the bug.

knard-intel (approved):
Seems good from what I understand.

src/control/system/member.go
tanabarr: These are unnecessary restrictions and while investigating the bug I've seen instances where these transitions are valid but prevented. I introduced these restrictions a while ago and the state machine has evolved since. There is no benefit in restricting transition from excluded or stopping states and allowing them offers greater usability.

kjacque: Okay, that makes sense. Thanks for explaining.

kjacque (approved):

PR #12722, opened 2023-07-27
DAOS-14064 control: INFO logging for ramdisk/hugepages (#12717)

mjmac:
Log the following at INFO level so that they are visible
in production configurations:
  * Auto-calculated per-engine ramdisk size
  * Auto-calculated hugepage count

### DIFF
--- a/src/control/server/config/server.go
+++ b/src/control/server/config/server.go
@@ -448,6 +448,10 @@ func getAccessPointAddrWithPort(log logging.Logger, addr string, portDefault int
 	return addr, nil
 }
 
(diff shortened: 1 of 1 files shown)


kjacque (approved):

tanabarr (approved):

mjmac:
@daos-stack/daos-gatekeeper: This is a tiny logging-only change to improve log output in production configurations. I can't check the hw-med-verbs failure, but it seems reasonable to assume that it's not related to this patch.

daltonbohning:
Failure looks to be https://daosio.atlassian.net/browse/DAOS-14027

PR #12757, opened 2023-08-02
DAOS-13812 container: fix destroy vs lookup

wangshilong:
If container is going to be dead, set @sc_stopping firstly, and later lookup will fail, then wait existed container services exit.

Required-githooks: true


src/rebuild/scan.c
```diff
@@ -839,35 +840,45 @@ rebuild_container_scan_cb(daos_handle_t ih, vos_iter_entry_t *entry,
 	}
 
 	rc = vos_cont_open(iter_param->ip_hdl, entry->ie_couuid, &coh);
```
NiuYawei, line 842: It's not from this patch, but I don't see why we need to call vos_cont_open() here, won't following ds_cont_child_lookup() open vos container? @wangdi1 ?

src/container/srv_target.c
```diff
@@ -1154,6 +1166,14 @@ cont_child_destroy_one(void *vin)
 				       &cont);
 		if (rc == -DER_NONEXIST)
 			break;
+
+		if (cont->sc_stopping) {
+			rc = -DER_SHUTDOWN;
+			D_ERROR(DF_CONT": container is in stopping "DF_RC"\n",
+				DP_CONT(cont->sc_pool->spc_uuid, cont->sc_uuid), DP_RC(rc));
+			cont_child_put(tls->dt_cont_cache, cont);
+		}
```
NiuYawei, line 1175: This looks incorrect to me, why should container destroy return error when it's stopping?

src/container/srv_target.c
```diff
@@ -1178,7 +1198,7 @@ cont_child_destroy_one(void *vin)
 		ABT_mutex_unlock(cont->sc_mutex);
 
 		/* Give chance to DTX reindex ULT for exit. */
-		if (unlikely(cont->sc_dtx_reindex))
+		while (unlikely(cont->sc_dtx_reindex))
 			ABT_thread_yield();
```
NiuYawei, line 1202: @Nasf-Fan , reindex is supposed to be very quick? Is this busy yield loop ok?

src/container/srv_target.c
wangshilong: Yup, this should be removed.

src/container/srv_target.c
```diff
@@ -772,16 +789,11 @@ cont_child_lookup(struct daos_lru_cache *cache, const uuid_t co_uuid,
 		return rc;
 	}
 
-	*cont = cont_child_obj(llink);
+	ds_cont = cont_child_obj(llink);
+	*cont = ds_cont;
```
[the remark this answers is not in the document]
wangshilong: Sure

NiuYawei (approved):

src/container/srv_target.c
Nasf-Fan: Reindex may take long time. But when close the container, reindex ULT will exit soon when detects the "stopping" flag.

Nasf-Fan (approved):

PR #12819, opened 2023-08-10
DAOS-14411 dfuse: Add per-container statistics.

ashleypittman:
Add dfuse operation statistics so it's possible to see what a
workload is performing.  Keep the statistics per container,
log them at INFO on umount, and export them via the daos
filesystem query command.


src/include/dfuse_ioctl.h
```diff
@@ -66,9 +67,15 @@ struct dfuse_mem_query {
...
 	ino_t    ino;
+	uint32_t stat_count;
 	bool     found;
 };
 
+struct dfuse_stat {
+	uint64_t value;
+	char     name[16];
+};
+
```
daltonbohning, line 78: Since this is in `src/include`, just checking if we forsee any need to extend this struct in the future? Is there a need to version or set aside extra space just in case?

src/utils/daos_hdlr.c
```diff
@@ -2464,8 +2467,35 @@ dfuse_count_query(struct cmd_args_s *ap)
 	ap->dfuse_mem.container_count = query.container_count;
 	ap->dfuse_mem.found           = query.found;
 
+	D_ALLOC_ARRAY(stat, query.stat_count);
+	if (stat == NULL)
+		goto close;
```
daltonbohning, line 2472: Do we need this?
```
rc = daos_errno2der(errno);
```

src/include/dfuse_ioctl.h
ashleypittman: There's already DFUSE_IOCTL_VERSION and we can also change the ioctl `cmd` number if we need to change the interface.  This is all new now so we can still change it up until we branch for 2.6 - I expect to continue working on the json output whilst we still have that freedom.

src/utils/daos_hdlr.c
ashleypittman: Actually no, if this fails then the ioctl will still work but only include the other metrics, not the stats.  I can document or revise this behaviour later if you prefer though.

daltonbohning: If we fail to allocate something, we're out of memory, right? So isn't that sort of "fatal"?

ashleypittman: The command will fetch two things, the inode counts which it already does and the per-container statistics which this PR adds, if this allocation fails then the inode counts will still be reported but the per-container stats won't be.

This code would therefore work as-is, but you're right, it would probably be better to return an error here rather than half the data.

ashleypittman: Fixed.

daltonbohning (approved):

jolivier23 (approved):

PR #12828, opened 2023-08-13
DAOS-14150 object: restart distributed TX if hit unavailable target

Nasf-Fan:
If CPD RPC gets -DER_HG or -DER_EXCLUDED error, it spite of it is from client local network or remote server, it means that related target may be unavailable. Under such case, related distributd TX should restarted instead of directly resent. That will avoid miss TX conflict detection on related unavailable target. It will also generate new targets list after TX restart.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -1713,21 +1713,21 @@ dc_obj_layout_refresh(daos_handle_t oh)
 	return rc;
 }
 
-static inline uint32_t
-dc_obj_retry_delay(struct obj_auxi_args *obj_auxi, int err)
+uint32_t
+dc_obj_retry_delay(tse_task_t *task, int err, uint16_t *retry_cnt, uint16_t *inprogress_cnt)
 {
 	uint32_t	delay = 0;
 
-	/* Randomly delay 5 - 36 us if it is not the first retry for
-	 * -DER_INPROGRESS && -DER_UPDATE_AGAIN case.
-	 **/
+	/*
+	 * Randomly delay 5 - 68 us if it is not the first retry for
+	 * -DER_INPROGRESS || -DER_UPDATE_AGAIN cases.
+	 */
+	++(*retry_cnt);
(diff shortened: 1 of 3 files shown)


src/object/cli_obj.c
```diff
@@ -1713,21 +1713,21 @@ dc_obj_layout_refresh(daos_handle_t oh)
...
+uint32_t
+dc_obj_retry_delay(tse_task_t *task, int err, uint16_t *retry_cnt, uint16_t *inprogress_cnt)
 {
 	uint32_t	delay = 0;
 
-	/* Randomly delay 5 - 36 us if it is not the first retry for
-	 * -DER_INPROGRESS && -DER_UPDATE_AGAIN case.
-	 **/
+	/*
+	 * Randomly delay 5 - 68 us if it is not the first retry for
```
NiuYawei, line 1722: 5 to 68 us delay looks too short to me, could you explain a bit why we choose such short delay time? And why we only delay for -DER_INPROGRESS & -DER_UPDATE_AGAIN? Shouldn't we delay on every retry?

Nasf-Fan: Retry may be caused by many reasons. In fact, if the retry is caused by normal in-processing conflict update, 5-68 us delay is equal to 1 ~ N cycles yield, that may be enough for the conflict update. If we increase the retry interval, from application perspective, it may increase related read(s) latency too much.

> Shouldn't we delay on every retry?
I do not think so. For example, DER_TIMEOUT will cause related RPC to be retry, but we do not need to introduce delay for such case. Currently, if we are sure the target is still alive well, then we will add delay before retry.

Nasf-Fan: For the failure in DAOS-14150, the delay is not added here, instead, after TX restart, related task will be delay scheduled after N (0 - 1^20) us. If without the patch, the TX will be resent directly without restart.

NiuYawei: I see, thanks.

NiuYawei (approved):

wangdi1 (approved):

PR #12837, opened 2023-08-14
DAOS-6854 dfs: add dfs_file_ostat with async support

mchaarawi:
add an async version of ostat that schedules the fetch and stat in parallel.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfs/dfs.c
+++ b/src/client/dfs/dfs.c
@@ -4976,8 +4976,15 @@ ostatx_cb(tse_task_t *task, void *data)
 	if (rc)
 		D_GOTO(out, rc);
 
-	args->stbuf->st_size = op_args->array_stbuf.st_size;
-	args->stbuf->st_blocks = (args->stbuf->st_size + (1 << 9) - 1) >> 9;
+	if (S_ISREG(args->obj->mode)) {
+		args->stbuf->st_size = op_args->array_stbuf.st_size;
+		args->stbuf->st_blocks = (args->stbuf->st_size + (1 << 9) - 1) >> 9;
+	} else if (S_ISDIR(args->obj->mode)) {
+		args->stbuf->st_size = sizeof(op_args->entry);
+	} else if (S_ISLNK(args->obj->mode)) {
+		args->stbuf->st_size = op_args->entry.value_len;
+	}
+
 	args->stbuf->st_nlink = 1;
 	args->stbuf->st_mode = op_args->entry.mode;
 	args->stbuf->st_uid = op_args->entry.uid;
@@ -5004,14 +5011,16 @@ statx_task(tse_task_t *task)
 	tse_task_t		*fetch_task;
 	daos_obj_fetch_t	*fetch_arg;
 	tse_task_t		*stat_task;
-	daos_array_stat_t	*stat_arg;
 	tse_sched_t		*sched = tse_task2sched(task);
+	bool			need_stat = false;
 	int			i;
 	int			rc;
 
 	D_ALLOC_PTR(op_args);
-	if (op_args == NULL)
+	if (op_args == NULL) {
+		daos_obj_close(args->parent_oh, NULL);
 		return ENOMEM;
+	}
 
 	/** Create task to fetch entry. */
 	rc = daos_task_create(DAOS_OPC_OBJ_FETCH, sched, 0, NULL, &fetch_task);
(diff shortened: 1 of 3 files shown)


src/client/dfs/dfs.c
```diff
@@ -4939,6 +4939,239 @@ dfs_ostat(dfs_t *dfs, dfs_obj_t *obj, struct stat *stbuf)
...
+	} else if (S_ISDIR(args->obj->mode)) {
+		daos_obj_query_key_t *stat_arg;
+
+		rc = daos_task_create(DAOS_OPC_OBJ_QUERY_KEY, sched, 0, NULL, &stat_task);
+		if (rc != 0) {
+			D_ERROR("daos_task_create() failed: "DF_RC"\n", DP_RC(rc));
+			D_GOTO(err2_out, rc);
+		}
+
+		/** set array_stat parameters */
```
wiliamhuang, line 5092: "/** set array_stat parameters */" from copy-paste. Need revision?

wiliamhuang (dismissed):

src/client/dfs/dfs.c
```diff
@@ -4939,6 +4939,239 @@ dfs_ostat(dfs_t *dfs, dfs_obj_t *obj, struct stat *stbuf)
...
+ostatx_cb(tse_task_t *task, void *data)
+{
+	struct dfs_statx_args	*args = daos_task_get_args(task);
+	struct statx_op_args	*op_args = *((struct statx_op_args **)data);
+	int			rc2, rc = daos_der2errno(task->dt_result);
+
+	if (rc != 0) {
+		D_ERROR("Failed to stat file "DF_RC"\n", DP_RC(rc));
+		D_GOTO(out, rc);
+	}
```
ashleypittman, line 4969: I think this is wrong, I expect rc is a daos errno here and it needs to be to use DF_RC however this function returns a system error number in other cases.  That said, I'm not sure where the RC of this function is used/passed to?

ashleypittman (commented):
Needs an updated description and Jira ticket filed.

src/client/dfs/dfs_internal.h
```diff
@@ -127,6 +127,10 @@ dfs_recreate_sb(daos_handle_t coh, dfs_attr_t *attr);
 int
 dfs_relink_root(daos_handle_t coh);
 
+/** Internal routine for async ostat. Today this only works for files. */
```
ashleypittman, line 130: Comment needs updating.

src/client/dfs/dfs.c
mchaarawi: well i had that right the first time around (for the printf) until you recommended the change in the dev channel :-/

otherwise this is using a system errno which is set in the event error completion or returned as the function return code if ev is NULL

mchaarawi: OK should be fixed now. what was missing is setting the task result from der to errno in this case because even if the CB return errno, tse will use the original task error from DAOS over the one from CB since it keeps the original failure code (in case there is a failure).
I added 2 test cases, one that exercises this code (error from DAOS, and the second that exercises the error next in this CB, to verify it returns appropriately the error.

wiliamhuang (approved):

src/client/dfs/dfs.c
mchaarawi: @ashleypittman any more comments on this?

src/client/dfs/dfs.c
```diff
@@ -4939,6 +4939,243 @@ dfs_ostat(dfs_t *dfs, dfs_obj_t *obj, struct stat *stbuf)
...
+
+	if (args->obj->oid.hi != op_args->entry.oid.hi ||
+	    args->obj->oid.lo != op_args->entry.oid.lo)
+		D_GOTO(out, rc = ENOENT);
+
+	rc = update_stbuf_times(op_args->entry, op_args->array_stbuf.st_max_epoch, args->stbuf, NULL);
+	if (rc)
+		D_GOTO(out, rc);
+
```
ashleypittman, line 4981: At least in theory if the entry has been replaced by something else the the oid should have changed and this will be caught above I think it would be worth adding an assert that `S_IFMT & args->obj->mode == S_IFMT & op_args.entry.mode`

ashleypittman (commented):
The existing stat code has checks for container/dfs layout version.  This is fine if we don't support that but if that's the case then we should add a check in dfs_mount() and remove the other code for it as it's likely that without this PR it would probably still work and with this PR it would fail.

src/client/dfs/dfs.c
mchaarawi: i don't think this assert is necessary here.

mchaarawi:
> The existing stat code has checks for container/dfs layout version. This is fine if we don't support that but if that's the case then we should add a check in dfs_mount() and remove the other code for it as it's likely that without this PR it would probably still work and with this PR it would fail.

although im confident we do not have 2.0 datasets lying around. anyway, we can add the layout support code for this stat.

ashleypittman:
> > The existing stat code has checks for container/dfs layout version. This is fine if we don't support that but if that's the case then we should add a check in dfs_mount() and remove the other code for it as it's likely that without this PR it would probably still work and with this PR it would fail.
> 
> although im confident we do not have 2.0 datasets lying around. anyway, we can add the layout support code for this stat.

I think just a check in dfs_mount() and removing the logic to support it would be fine.  Doesn't need to happen in this PR.

mchaarawi:
> > > The existing stat code has checks for container/dfs layout version. This is fine if we don't support that but if that's the case then we should add a check in dfs_mount() and remove the other code for it as it's likely that without this PR it would probably still work and with this PR it would fail.
> > 
> > 
> > although im confident we do not have 2.0 datasets lying around. anyway, we can add the layout support code for this stat.
> 
> I think just a check in dfs_mount() and removing the logic to support it would be fine. Doesn't need to happen in this PR.

ok that would be easier than to keep support for 2.0 containers and can cleanup the code a bit. ill push a follow on pr.

PR #12840, opened 2023-08-14
jgm/daos-13239-2.4 cart: Refactor RPC send error code mapping.

jgmoore-or:

### Gatekeeper:

### DIFF
--- a/src/cart/crt_hg.c
+++ b/src/cart/crt_hg.c
@@ -1311,7 +1311,17 @@ crt_hg_req_send_cb(const struct hg_cb_info *hg_cbinfo)
 		break;
 	default:
 		rpc_priv->crp_state = RPC_STATE_COMPLETED;
-		rc = crt_hgret_2_der(hg_cbinfo->ret);
+
+		if (crt_is_service()) {
+			rc = crt_hgret_2_der(hg_cbinfo->ret);
+		} else {
+			if (hg_cbinfo->ret) {
+				rc = -DER_HG_SEND_FAILED;
+			} else {
+				rc = DER_SUCCESS;
+			}
+		}
+
 		hg_ret = hg_cbinfo->ret;
 		RPC_TRACE(DB_NET, rpc_priv, "hg_cbinfo->ret: " DF_HG_RC ".\n",
 			  DP_HG_RC(hg_cbinfo->ret));
--- a/src/include/daos_errno.h
+++ b/src/include/daos_errno.h
@@ -162,8 +162,8 @@ extern "C" {
 	/** Fatal (non-retry-able) transport layer mercury error */	\
 	ACTION(DER_HG_FATAL,		(DER_ERR_GURT_BASE + 45),	\
 	       Fatal transport layer mercury error)                     \
-        /** Failed to send rpc */                                       \
-        ACTION(DER_HG_SEND_FAILED,       (DER_ERR_GURT_BASE + 46),      \
+	/** Failed to send rpc */                                       \
+	ACTION(DER_HG_SEND_FAILED,       (DER_ERR_GURT_BASE + 46),      \
 	       Failed to send rpc)
 	/** TODO: add more error numbers */
 
--- a/src/include/daos_errno.h
+++ b/src/include/daos_errno.h
@@ -161,7 +161,10 @@ extern "C" {
 	       incompatible user or group permissions)			\
 	/** Fatal (non-retry-able) transport layer mercury error */	\
 	ACTION(DER_HG_FATAL,		(DER_ERR_GURT_BASE + 45),	\
-	       Fatal transport layer mercury error)
+	       Fatal transport layer mercury error)                     \
+        /** Failed to send rpc */                                       \
+        ACTION(DER_HG_SEND_FAILED,       (DER_ERR_GURT_BASE + 46),      \
+	       Failed to send rpc)
 	/** TODO: add more error numbers */
 
 /** Preprocessor macro defining DAOS errno values and internal definition of d_errstr */
--- a/src/gurt/tests/test_gurt.c
+++ b/src/gurt/tests/test_gurt.c
@@ -207,6 +207,16 @@ void test_d_errstr(void **state)
 #else
 	test_d_errstr_v2(state);
 #endif
+
+	/* Check the boundary at the end of the GURT error numbers, this will need updating if
+	 * additional error numbers are added.
+	 */
+	value = d_errstr(-DER_HG_SEND_FAILED);
+	assert_string_equal(value, "DER_HG_SEND_FAILED");
+	value = d_errstr(-1046);
+	assert_string_equal(value, "DER_HG_SEND_FAILED");
+	value = d_errstr(-(DER_HG_SEND_FAILED + 1));
+	assert_string_equal(value, "DER_UNKNOWN");
 }
 
 void test_d_errdesc(void **state)
--- a/src/include/daos/common.h
+++ b/src/include/daos/common.h
@@ -621,7 +621,7 @@ static inline bool
 daos_crt_network_error(int err)
 {
 	return err == -DER_HG || err == -DER_UNREACH || err == -DER_CANCELED ||
-	       err == -DER_NOREPLY || err == -DER_OOG;
+	       err == -DER_NOREPLY || err == -DER_OOG || err == -DER_HG_SEND_FAILED;
 }
 
 /** See crt_quiet_error. */


src/include/daos_errno.h
```diff
@@ -161,7 +161,10 @@ extern "C" {
 	       incompatible user or group permissions)			\
 	/** Fatal (non-retry-able) transport layer mercury error */	\
 	ACTION(DER_HG_FATAL,		(DER_ERR_GURT_BASE + 45),	\
-	       Fatal transport layer mercury error)
+	       Fatal transport layer mercury error)                     \
+        /** Failed to send rpc */                                       \
+        ACTION(DER_HG_SEND_FAILED,       (DER_ERR_GURT_BASE + 46),      \
```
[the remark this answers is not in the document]
jgmoore-or: Fixed.

soumagne (approved):

frostedcmos (approved):

src/cart/crt_hg.c
```diff
@@ -1311,7 +1311,17 @@ crt_hg_req_send_cb(const struct hg_cb_info *hg_cbinfo)
 		break;
 	default:
 		rpc_priv->crp_state = RPC_STATE_COMPLETED;
-		rc = crt_hgret_2_der(hg_cbinfo->ret);
+
+		if (crt_is_service()) {
```
mchaarawi, line 1315: why is the server treated differently?
if those errors from mercury are retryable on the client side, why would they not be on the server side?
@wangdi1 @liuxuezhao  might need to weigh in here also on what is expected.

frostedcmos: to be clear this is not really about retriable though it can affect it, this is whether rpc errors get squashed into a single 'rpc send failed' (which is sufficient for clients and is now added to the list of retriable errors) vs split into 'it failed because of reason X' which might be needed by engines. if engine-to-engine doesnt (or shouldnt) care about why rpc send failed then we should be able to remove condition here and squash them for both the client and the server.

frostedcmos: to add some context:

the goal of this change to differentiate when rpc failed due to a 'send failure' vs any other parts, such as local handle allocation failure. if both parts today fail with the same rc then we cant distinguish between whether the operation should be retried or not. 

if rpc failed to send due to a send failure then it is probably safe to retry to another engine. 

if it failed to send for other reasons then we should abort or handle it gracefully.

What happens today is that sometimes we can get a floor of retries due to something else failing, filling logs etc..

A reverse issue also happens today when send fails due to a new error and then we end up having to add it to the list of retriable ones.

The reason why you cant just squash all errors on the engines is there are some server code parts that rely on the specific rc:
e.g. part in iv added by @bfaccini recently
```
2717         /* handle locally generated errors during IV operations synchronously to ensure unregister
2718          * of bulk buffer will occur before freeing it, just in case peer will finally make it
2719          * unexpectedly
2720          */
2721         if (cb_info->cci_rc == -DER_TIMEDOUT || cb_info->cci_rc == -DER_EXCLUDED ||
2722             cb_info->cci_rc == -DER_CANCELED)
2723                 goto callback;
```
On a side issue now reading this code, I remember we had a discussion about it at some point, is there a reason why we dont do this for != DER_SUCCESS @bfaccini ? seems if we get any other rc we will not free bulk buffers how we should which could read to a lot of weird problems.

Back to the topic -- this above is not the only occurrence in the server-side code where a specific rc is expected, so I wouldn't feel comfortable enabling this error squashing on engine-to-engine rpc sends at this point.

I think ideally, the client should eventually change to always retry on this new 'send failed' rc only, and on engines we can keep adding to the list of retry errors, until we can enable this change on engines too.

jgmoore-or: When the server-to-server handling of send errors was set to be the same as for (non crt_is_service) clients, there were failure for DTX tests and some degraded/EC tests  These failures were fixed by restoring the prior mapping of the errors for the engine handling.

wangdi1: > if rpc failed to send due to a send failure then it is probably safe to retry to another engine.

if it failed to send for other reasons then we should abort or handle it gracefully.

What happens today is that sometimes we can get a floor of retries due to something else failing, filling logs etc..

I thought daos client can tell these codes anyway, why squash them? If the error code can not tell if this is due to the logging or real network failure, etc. then the error code should be fixed. Though I am not how much efforts it needs here.  I probably miss sth here.

> if engine-to-engine doesnt (or shouldnt) care about why rpc send failed then we should be able to remove condition here and squash them for both the client and the server.

engine to engine does treat RPC failure differently as client to engine, ec aggregation, iv etc, though it is DAOS Server stack internal logic, and it needs specific error code anyway.

mchaarawi: maybe i should have phrased my question / concern differently. i am not asking why we are not squashing all the errors for the server. Do we think that the client should retry on any error from mercury?

wangdi1: Probably not, client will only retry for "temporary" error, 
1. the error from mercury/cart can cause SWIM to exclude the rank from pool map.
2. the error can be "fixed" by itself, i.e. by daos engine itself, dtx/connection etc.
3. network hiccup error,  though this probably should be controlled by SWIM.

frostedcmos: > I thought daos client can tell these codes anyway, why squash them? If the error code can not tell if this is due to the logging or real network failure, etc. then the error code should be fixed. Though I am not how much efforts it needs here. I probably miss sth here.

No, because we have cases when same codes can be returned both from the rpc send path and from other code paths, and we only want to be retrying the send path. For example if send path and other path leading to send both return DER_NO_MEM, then by the time you decide whether to retry or not you won't know it based on rc alone.

frostedcmos: I think we should have clients retry on any error from mercury if it originates from HG_Forward() (rpc send path). The destination peer might not be available for any number of reason, but in each of those client should probably retry, picking new target is possible. 

Relying on specific mercury error codes will be messy as different providers can return different sets of errors.

Any other mercury/cart error either before or after the rpc send is likely to be more fatal where a retry will not produce any different outcome. For example, failure to create rpc handle, endpoint, bulk creation/mapping errors and so on.

wangdi1: Thanks for clarification. Though I feel a bit layer violation to do any specific squashing or checking for the purpose of "retry", and I also agree relay on error code might be a "messy" and complicate indeed.

Maybe client retry protocol could be improved a bit here, so if client get any error from cart/mercury, it should ask PS leader, what is the status of that rank, 

If leader made decision to exclude the rank, then client will refresh the map anyway.

Otherwise client will only retry(keep asking) until SWIM timeout, or some certain threshold to make sure SWIM can make the decision.

wangdi1 (approved):

src/cart/crt_hg.c
liuxuezhao: sorry, saw those comments but still feel confuse what is this PR really to fix.

is there a detailed case that original code cannot handle but with this PR can "fix"?

liuxuezhao: sorry, saw those comments but still feel confuse what is this PR really try to fix.

is there a detailed case that original code cannot handle but with this PR can "fix"?

And TBH I don't know why should retry all cases of HG_Forward() reported failure by its complete CB? for example even if got HG_NOMEM/HG_FAULT/HG_PROTOCOL_ERROR/HG_MSGSIZE/HG_ADDRNOTAVAIL/HG_CHECKSUM_ERROR etc, you mean it is impossible to get those failure from mercury or daos should retry all those errs?
I agree that it is better for daos if need not to understand those detailed HG err code, but looks also not a good idea to make daos always retry despite of what err code returned by HG.

I think we probably should understand what this PR try to fix first, which detailed case cannot work without this PR?

liuxuezhao (changes requested):
I think we probably should understand what this PR try to fix first, which detailed case cannot work without this PR?

src/cart/crt_hg.c
frostedcmos: The master version of the PR as mentioned above tries to differentiate send failures from all others for purposes of deciding if the error should be retriable or not. It is meant as an improvement to the current method.

It is not ideal ofc, but provides benefits over the current method of adding new rc to the daos_crt_network_error() each time a mercury returns a new error code or the error mapping changes. Practically most mercury errors end up being mapped to DER_HG, which 

Error mapping is involved because daos_crt_netowrk_error() checks cart-level error codes, not mercury errors.

In this ticket DAOS-13239 the HG_PROTOCOL_ERROR is returned during online reintegration in ior, but HG_PROTOCOL_ERROR was mapped to -DER_PROTO which is not retriable; only generic DER_HG is retriable. (We should also probably remove DER_HG from retriable errors now as its too broad).

If you think some rcs should not be retriable then we can exclude them from being squashed (and map them to something not in retriable list), but i am not sure i myself know what rcs those would be or what to do in those cases

So an alternative to this PR, is to either change HG_PROTOCOL_ERROR to DER_HG (since DER_HG is in daos_crt_network_error() list), or add -DER_PROTO to that list.

liuxuezhao: Thanks for explanations. I don't want to be nitpicking but this PR looks with some confuse so just want to clarify.

"the HG_PROTOCOL_ERROR is returned during online reintegration in ior"
do we know which module and which part of code reported HG_PROTOCOL_ERROR? is it the mercury UCX plugin code? can we consider to change that part of code if possible?
seems HG_PROTOCOL_ERROR code from HG all belong to un-retryable case like unreasonable serialize bulk size, bad poll fd etc. So probably cannot map HG_PROTOCOL_ERROR to DER_HG.
If only for the HG_PROTOCOL_ERROR case, we may see whether or not could with other solution? which code report that err code in the first place?

frostedcmos: @soumagne can you provide details as to when we can expect HG_PROTOCOL_ERROR from mercury?

jgmoore-or: HG_PROTOCOL_ERROR was remapped to not be retry-able, due to its failed return from a call to allocate a bulk handle, which caused an infinite retry:
https://github.com/daos-stack/daos/commit/15f9dfd1ebd0a0419b859094f2cc49e5bb6f140e
This PR does not change that resolution (since the above failure did not occur directly on an RPC send).

We are currently incurring a failure on Aurora (https://daosio.atlassian.net/browse/DAOS-14161) wherein an HG_PROTOCOL_ERROR is returned, but a retry is the required handling. Since in this case the failure is occurring on an RPC send, it is anticipated that a retry in this case will resolve the failure (we're currently awaiting results from the Aurora test team).

There is no evidence from DAOS testing that indicates the policy of retrying on all (non crt_is_service) client RPC sends results in any issues (the passage of this PR for all CI tests supports this assessment). We have an issue with RPC send failing for UCX due a transient memory issue (UCX layer returns NA_NOMEM via na_ucx.c).  In this case a retry resolves the failure, and hence this UCX defect is resolved by this PR.

The CaRT is proposing that we go ahead with the solution implemented in this PR, but remain prepared to address any issues with it, if and when such issues are encountered.

mchaarawi: we have just seen a PR in triage where a protocol failure should not have been retried:
```
08/15-11:15:59.13 wolf-233vm3 DAOS[420799/420799/0] external WARN # [36136.009190] mercury->op: [warning] /builddir/build/BUILD/mercury-2.3.0/src/na/na_ofi.c:5778
 # na_ofi_cq_readerr(): fi_cq_readerr() got err: 88 (Socket operation on non-socket), prov_errno: 88 (Socket operation on non-socket)
 ```
but with this patch which is landed on master the test ended up timing out with huge logs of retries. so i don't think it's a good idea to retry on any error.

also just because a PR passes CI tests does not mean it has no issues ;-)
please note that I'm not saying your pr is wrong, but just that this statement is (with a lot of historical evidence) wrong :-)

jgmoore-or: A "proof" is always defeated by a single counter-example. BTW, I just said it was supportive. We have a problem when handling errors on a per-error-type basis when we have the same error code requiring retry in one instance and no retry in another. A comprehensive, "provably" correct handling of each individual error at each layer of the protocol stack would perhaps resolve this, but this solution has been elusive.  @mchaarawi: If you have time, please add a post with a link to the PR with the (retried) failure -- quoted above. Thanks.

mchaarawi: https://build.hpdd.intel.com/job/daos-stack/job/daos/job/master/8940/artifact/Functional%20on%20Leap%2015.4/dfuse/daos_build.py/daos_logs.wolf-106vm3/

liuxuezhao: besides the above failure mentioned by @mchaarawi , I think we need to figure out where and why mercury returned "HG_PROTOCOL_ERROR".
" We are currently incurring a failure on Aurora (https://daosio.atlassian.net/browse/DAOS-14161) wherein an HG_PROTOCOL_ERROR is returned, but a retry is the required handling."
Seems it is unclear yet at which case and code why mercury returned HG_PROTOCOL_ERROR, and why the retry can resolve it. 
In HG code, as far as I can see, these cases with HG_PROTOCOL_ERROR set/returned -
1. hg_bulk_register(), serialize_size == 0
2. hg_bulk_transfer_segments_na(), 
3. hg_core_context_create() pool fd error,
4. hg_core_poll_wait, hg_poll_wait failure
5. hg_core_progress_loopback_notify()
6. hg_core_header_request_verify() rpc header error

do we know which case in the test reported that err? And all cases of HG's HG_PROTOCOL_ERROR looks un-ertryable err, why retry can help to resolve it. Seems the root cause of the issue looks still unclear yet.

bfaccini: > The reason why you cant just squash all errors on the engines is there are some server code parts that rely on the 
> specific rc:
> 
> e.g. part in iv added by @bfaccini recently
> 
> 2717         /* handle locally generated errors during IV operations synchronously to ensure unregister
> 2718          * of bulk buffer will occur before freeing it, just in case peer will finally make it
> 2719          * unexpectedly
> 2720          */
> 2721         if (cb_info->cci_rc == -DER_TIMEDOUT || cb_info->cci_rc == -DER_EXCLUDED ||
> 2722             cb_info->cci_rc == -DER_CANCELED)
> 2723                 goto callback;
> On a side issue now reading this code, I remember we had a discussion about it at some point, is there a reason why
> we dont do this for != DER_SUCCESS @bfaccini ? seems if we get any other rc we will not free bulk buffers how we
> should which could read to a lot of weird problems.

@frostedcmos , bulk will be freed asynchronously in handle_response_internal() in case of other/remotely-generated RCs or success. This is ok since transfer should not occur anymore in all these cases.

ashleypittman:
For what it's worth when this code landed on master we started seeing timeouts in some if the interception library tests as some I/O paths which were expected to fail are not re-trying indefinetly.

I think we need to revert this on master and continue this discussion in a Jira ticket.

src/cart/crt_hg.c
frostedcmos: @liuxuezhao 
> do we know which case in the test reported that err? And all cases of HG's HG_PROTOCOL_ERROR looks un-ertryable err, why retry can help to resolve it. Seems the root cause of the issue looks still unclear yet.

Jerome is on a vacation until next week, and we will need him to give his input on this.

Given @ashleypittman report about time outs on master, a protocol error that should not be retried example form @mchaarawi  as well as unresolved questions here, it is reasonable to unroll this patch from the master, and continue the discussion and mercury side investigations once @soumagne is back next week.

PR #12865, opened 2023-08-18
DAOS-14202 object: check failed targets during retry

wangdi1:
Failed targets should be kept and checked during retry, otherwise same target might be counted duplicately, and cause data lost.

Required-githooks: true

### Gatekeeper:


src/object/cli_obj.c
```diff
@@ -3728,9 +3728,17 @@ obj_shard_comp_cb(tse_task_t *task, struct shard_auxi_args *shard_auxi,
 				shard_auxi->shard);
 		}
 	} else if (obj_retry_error(ret)) {
+		int rc;
+
 		D_DEBUG(DB_IO, "shard %d ret %d.\n", shard_auxi->shard, ret);
 		if (obj_auxi->result == 0)
 			obj_auxi->result = ret;
+		rc = obj_auxi_add_failed_tgt(obj_auxi, shard_auxi->target);
```
liuxuezhao, line 3736: for temporary timeout failure, may need not to add it to failed tgt? 

BTW, for the issue this PR try to resolve "Failed targets should be kept and checked during retry, otherwise same target might be counted duplicately, and cause data lost."
In code obj_ec_fail_info_insert() we already with code to check duplicated tgt -
1384 >       if (obj_ec_tgt_in_err(fail_info->efi_tgt_list, fail_info->efi_ntgts, fail_tgt))             
1385 >       >       return 0;  

So even if we insert same tgt multiple times, it won't be counted more than one time. no?

wangdi1: > So even if we insert same tgt multiple times, it won't be counted more than one time. no?

NVME IO retry will retry all shards initially, so if we do not remember the failure tgts, then innocent shards might be added to the failure list, then IO might be sent to the "wrong" server again.

src/object/cli_obj.c
```diff
@@ -5117,12 +5142,21 @@ obj_ec_valid_shard_get(struct obj_auxi_args *obj_auxi, uint8_t *tgt_bitmap,
...
+	if (need_retry_redundancy(obj_auxi)) {
+		struct obj_auxi_tgt_list *failed_list;
+		uint32_t tgt_id;
+
+		D_ASSERT(obj_auxi->failed_tgt_list != NULL);
+		failed_list = obj_auxi->failed_tgt_list;
+		tgt_id = obj->cob_shards->do_shards[shard_idx].do_target_id;
+
+		if (tgt_in_failed_tgts_list(tgt_id, failed_list))
+			force_ec_degrade = true;
```
liuxuezhao, line 5158: OK, i see what this PR try to fix.
But the logic of "nvme_io_err" flag seems still not fully correct, the below code when exit this func looks not right to me -
	if (ec_degrade) {
		obj_auxi->tx_uncertain = 0;
		obj_auxi->csum_retry = 0;
		obj_auxi->nvme_io_err = 0;
	}

why clear those flags right after the first degraded shard fetch?
For example if we have EC4P2G1 obj, get -DER_NVME_IO from both shard 0 and shard 1's fetch, when retry the fetch, after force degraded fetch for shard 0, the next shard 1 fetch also need to force degraded right? but the obj_auxi->nvme_io_err was cleared when force degraded for shard 0, then shard 1 still will get -DER_NVME_IO after retry.

I think we need to fix it if I am not wrong, but could be in a following patch.
BTW, those flags and retry logic really confuse, we may need to refine it later at appropriate time.

liuxuezhao (approved):

src/object/cli_obj.c
NiuYawei: yeah, the retry logic needs be refined. To be honest, I didn't fully follow the client retry code when I introduce the nvme io error retry, I just followed the csum error retry code.

NiuYawei (approved):
@Nasf-Fan , could you take a look too? Thanks.

src/object/cli_obj.c
wangdi1: >why clear those flags right after the first degraded shard fetch?
For example if we have EC4P2G1 obj, get -DER_NVME_IO from both shard 0 and shard 1's fetch, when retry the fetch, after force degraded fetch for shard 0, the next shard 1 fetch also need to force degraded right? but the obj_auxi->nvme_io_err was cleared when force degraded for shard 0, then shard 1 still will get -DER_NVME_IO after retry.

Hmm, let me update the patch, make this clearer. thanks

src/object/cli_obj.c
```diff
@@ -5107,28 +5129,53 @@ obj_retry_next_shard(struct dc_object *obj, struct obj_auxi_args *obj_auxi,
...
+static inline bool
+shard_was_fail(struct obj_auxi_args *obj_auxi, uint32_t shard_idx)
+{
+	struct obj_auxi_tgt_list *failed_list;
+	uint32_t tgt_id;
+
+	if (obj_auxi->force_degraded) {
+		D_DEBUG(DB_IO, DF_OID" fail idx %u\n",
+			DP_OID(obj_auxi->obj->cob_md.omd_id), shard_idx);
+		obj_auxi->force_degraded = 0;
```
liuxuezhao, line 5148: seems not must to clear "obj_auxi->force_degraded" flag?
now DIOF_FOR_FORCE_DEGRADE flag only used for rebuild code and will set obj_auxi->force_degraded if carried  DIOF_FOR_FORCE_DEGRADE flag

liuxuezhao (approved):

Nasf-Fan (approved):

PR #12875, opened 2023-08-21
Revert "DAOS-13239 cart: Fix send failure error mapping.."

ashleypittman:
Reverts daos-stack/daos#12199

Reverting to test DAOS-14182.

Test-tag: dfuse


frostedcmos (approved):

jgmoore-or (approved):

ashleypittman:
This PR failed the dfuse/DaosBuild test but the point is that it failed in 11 minutes rather than times out in six hours so although it's a test failure it's actually reverting a regression.

PR #12880, opened 2023-08-21
DAOS-14155 gurt: Add DL_CDEBUG macro.

ashleypittman:
Update usage in lots of places.

Required-githooks: true

### DIFF
--- a/src/include/gurt/debug.h
+++ b/src/include/gurt/debug.h
@@ -152,6 +152,14 @@ extern void (*d_alt_assert)(const int, const char*, const char*, const int);
 			D_DEBUG(flag_false, __VA_ARGS__);	\
 	} while (0)
 
+#define DL_CDEBUG(cond, flag_true, flag_false, _rc, _fmt, ...)                                     \
+	do {                                                                                       \
+		if (cond)                                                                          \
+			D_DEBUG(flag_true, _fmt ": " DF_RC " \n", ##__VA_ARGS__, DP_RC(_rc));      \
+		else                                                                               \
+			D_DEBUG(flag_false, _fmt ": " DF_RC "\n", ##__VA_ARGS__, DP_RC(_rc));      \
+	} while (0)
+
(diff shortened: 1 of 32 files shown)


src/include/gurt/debug.h
```diff
@@ -152,6 +152,14 @@ extern void (*d_alt_assert)(const int, const char*, const char*, const int);
 			D_DEBUG(flag_false, __VA_ARGS__);	\
 	} while (0)
 
+#define DL_CDEBUG(cond, flag_true, flag_false, _rc, _fmt, ...)                                     \
+	do {                                                                                       \
+		if (cond)                                                                          \
+			D_DEBUG(flag_true, _fmt ": " DF_RC " \n", ##__VA_ARGS__, DP_RC(_rc));      \
```
jolivier23, line 158: Unrelated but seeing this made me think of VOS_TX_LOG_FAIL and daos_rpc_retryable_rc among others where we are checking many error codes to make a decision.  Perhaps it's not a big deal since they are only error paths but it seems like we could make it more efficient somehow, like via a bit mask or such.

jolivier23 (approved):
Only issue is whether this will cause merge conflicts on 2.4

src/include/gurt/debug.h
ashleypittman: I just found out about the `VOS_TX_LOG_FAIL` macro today, I'll add it to the ones we check.

More eyes on this code, adding checkers and more consistency will make it easier to do more things with these macros.

ashleypittman:
> Only issue is whether this will cause merge conflicts on 2.4

It might do, but they'll be easy to fix.

For the previous changes I just added the macros as it had to go through a longer review and I wanted to minimise conflicts, now we've established the principal of a `error with rc` then I'm hoping this can be landed faster without conflicts, then i grepped the source and thought `that's not many locations to change` but it did take me longer than expected.

liw (approved):

PR #12883, opened 2023-08-22
DAOS-14188 object: set bitmaps if all data shards are choosen

wangdi1:
Bitmap should be set if all data shards are choosen in obj_ec_get_parity_or_alldata_shard().

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -6168,6 +6168,7 @@ obj_ec_get_parity_or_alldata_shard(struct obj_auxi_args *obj_auxi, unsigned int 
 	int			shard;
 	unsigned int		first;
 
+	grp_start = grp_idx * obj_get_grp_size(obj);
 	oca = obj_get_oca(obj);
 	if (dkey == NULL && obj_ec_parity_rotate_enabled(obj)) {
 		int fail_cnt = 0;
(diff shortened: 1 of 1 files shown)


src/object/cli_obj.c
```diff
@@ -6208,10 +6208,16 @@ obj_ec_get_parity_or_alldata_shard(struct obj_auxi_args *obj_auxi, unsigned int
 			if (shard < 0)
 				D_GOTO(out, shard);
 
-			if (is_ec_data_shard(obj_auxi->obj, obj_auxi->dkey_hash, shard))
-				*shard_cnt = obj_ec_data_tgt_nr(oca);
-			if (bitmaps != NULL)
-				setbit(*bitmaps, shard % obj_get_grp_size(obj));
+			if (is_ec_data_shard(obj_auxi->obj, obj_auxi->dkey_hash, shard)) {
```
liuxuezhao, line 6211: it seems strange that with flag "obj_auxi->to_leader" set but also need to send to all data shards, do you know any reasonable case for that? thx

liuxuezhao (approved):

src/object/cli_obj.c
wangdi1: Hmm, I thought I just follow the previous implementation here. Indeed, can not think any one either. @Nasf-Fan  any idea here?

Nasf-Fan: The scenario is like this: the client sends fetch RPC to one data shard, but gets -DER_INPROGRESS, then it is expected to re-send fetch RPC to the leader shard. For EC case, normally, the leader shard is one of the parities, but if all parity shards are lost, then one of data shard will be used as leader. Under such degrade mode, the fetch RPC needs to be sent to all data shards for degraded fetch.

Nasf-Fan (approved):

src/object/cli_obj.c
wangdi1: Thanks for clarification. Hmm, in this case, probably to_leader should not be set at all. since EC degraded fetch(no parity) does not care the leader at all.

PR #12891, opened 2023-08-22
DAOS-000 vos: Change return of vos re-index function

ashleypittman:
Required-githooks: true


ashleypittman:
@jolivier23 what do you think about work like this?  It would make the error handling a lot easier but there's also a lot of code to change - probably 5-6 vox functions and an unknown number of raft ones.

PR #12892, opened 2023-08-22
DAOS-623 build: Make scons quiet quiet.

ashleypittman:
Update build so there's less output of quiet is used.

Required-githooks: true


jolivier23 (approved):

ashleypittman:
@brianjmurrell @phender can you review this please?

site_scons/env_modules.py
```diff
@@ -56,7 +57,8 @@ def _module_func(self, command, *arguments):  # pylint: disable=no-self-use
 
         # pylint: disable=consider-using-with
         try:
-            print(f"Going to run {cmd}")
+            if not self._silent:
+                print(' '.join(cmd))
```
daltonbohning, line 61: Maybe define a `self.print` so we don't need this check everywhere?

src/tests/suite/SConscript
```diff
@@ -97,7 +29,7 @@ def scons():
                                c_files + daos_test_tgt,
                                LIBS=['daos_common'] + libraries)
 
-    newenv = configure_cmocka(denv.Clone())
```
daltonbohning, line 100: Why isn't this needed anymore?

site_scons/env_modules.py
ashleypittman: I considered it but there are only two calls.

src/tests/suite/SConscript
ashleypittman: I tested - none of the distros we have in CI use a version of cmocka with the bug we're testing for.

phender (approved):
It would be nice if we could define `def _print(msg):` in one common file and import it in the other ones, but it does not look we already have a common shared daos scons python file.

ashleypittman:
> It would be nice if we could define `def _print(msg):` in one common file and import it in the other ones, but it does not look we already have something like that set up already that would be easy to leverage.

I looked at this again after merging, we can add methods to envs but that doesn't help as much as you think it would as most of these prints don't have a env object available when the print is done.

PR #12898, opened 2023-08-22
DAOS-13396 dfs: Add a fi test for mwc code.

ashleypittman:
Add a fault injection test for dfs checking.
Create a new DS_ERROR macro for errors with system error codes.

Required-githooks: true


mchaarawi (changes requested):
this PR needs to be split into two. one for the FI test and the other for the new macro.
As for the DFS file update, -1 for that until we release 2.4 to avoid having merge issues and unclean cherry picks if needed. although i would prefer such mass updates to the code are not done like this and are more gradual to be done be each owner of the part of the code.

ashleypittman:
> this PR needs to be split into two. one for the FI test and the other for the new macro. As for the DFS file update, -1 for that until we release 2.4 to avoid having merge issues and unclean cherry picks if needed. although i would prefer such mass updates to the code are not done like this and are more gradual to be done be each owner of the part of the code.

This did end up being a lot bigger than I expected, I'm not sure how to carve it up smaller.

The new macro changes would be a tiny PR, probably just touching 2-3 files.  There are two actual bug-fixes to dfs although neither serious enough to warrant a 2.4 backport, the rest is all logging improvements.

I can push as a number of smaller PRs to take the edge of it but this might have to sit in the queue until 2.4 is stabilized.

ashleypittman:
Created https://github.com/daos-stack/daos/pull/12911 which contains just the new test and the non-logging fixes only.

ashleypittman:
New logging macros being added in https://github.com/daos-stack/daos/pull/12913 so setting this to draft for now.

PR #12918, opened 2023-08-23
DAOS-13399 gurt: Do not include malloc.h in public header file.

ashleypittman:
Mpich does not like this.

Required-githooks: true


mchaarawi (approved):

knard-intel (approved):
OK for me also.

PR #12925, opened 2023-08-24
DAOS-14158 pool: do not check iv valid for entry invalid

wangdi1:
It does not need to check invalid entry in pool_iv_ent_invalid(), because disconnection will remove entry locally on all nodes, so the entry is not really invalid, otherwise, disconnect handles will be left in the IV cache, then during extend, these handles might be fetched by the new rank incorrectly.

Move iv_valid check into the callback, because in some cases, if invalidate only invalid a single item under one entry, for example disconnect handle, it does not need set the entry to be invalid.

Add tests to verify it.

Required-githooks: true
Features: rebuild ec pr
Allow-unstable-test: true

### Gatekeeper:


liuxuezhao:
@wangdi1 need rebase?

wangdi1:
> @wangdi1 need rebase?

ah, sure

src/object/cli_obj.c
```diff
@@ -832,7 +815,7 @@ obj_dkey2grpidx(struct dc_object *obj, uint64_t hash, unsigned int map_ver)
 	D_ASSERT(grp_size > 0);
 
 	D_RWLOCK_RDLOCK(&obj->cob_lock);
-	if (obj->cob_version != map_ver || map_ver < pool_map_ver) {
+	if (obj->cob_version != pool_map_ver) {
```
liuxuezhao, line 818: is this change intended?

looks one difference is for example the pool map refreshed and obj layout refreshed by another thread (so both obj->cob_version and pool_map_ver refreshed), but this thread's IO handling with stale map_ver, this case may need return ESTALE as previous code?

PR #12929, opened 2023-08-24
DAOS-13564 object: add obj flatten APIs

liuxuezhao:
1. add daos_cont_set_ro() API definition
2. add DAOS_PROP_CO_READONLY container status bit
3. rename daos_cont_status_clear() to daos_cont_set_healthy()
    add daos_cont_status_clear alias to keep back-compatibility
4. define DAOS_PROP_CSF_HEALTHY/DAOS_PROP_CSF_READONLY internal flag bit
    cleanup DAOS_PROP_CO_CLEAR as no usage in code.
5. add daos_obj_set_ro() API definition

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/include/daos_obj.h
+++ b/src/include/daos_obj.h
@@ -287,6 +287,23 @@ enum {
 	DAOS_COND_MASK = ((1 << IO_FLAGS_COND_BITS) - 1),
 };
 
+/** DAOS I/O flag bits, first 8 bits reserved for conditional checks */
+#define DAOS_IOF_COND_PUNCH		(DAOS_COND_PUNCH)
+#define DAOS_IOF_COND_DKEY_INSERT	(DAOS_COND_DKEY_INSERT)
+#define DAOS_IOF_COND_DKEY_UPDATE	(DAOS_COND_DKEY_UPDATE)
+#define DAOS_IOF_COND_DKEY_FETCH	(DAOS_COND_DKEY_FETCH)
+#define DAOS_IOF_COND_AKEY_INSERT	(DAOS_COND_AKEY_INSERT)
+#define DAOS_IOF_COND_AKEY_UPDATE	(DAOS_COND_AKEY_UPDATE)
+#define DAOS_IOF_COND_AKEY_FETCH	(DAOS_COND_AKEY_FETCH)
+#define DAOS_IOF_COND_PER_AKEY		(DAOS_COND_AKEY_FETCH)
+
+/**
+ * Object flattening flag, this flag carried daos_obj_update() means this update is the last update
+ * request, the object will be immutable after this update's finish. DAOS can start flattening
+ * this object.
+ */
+#define DAOS_IOF_FLATTEN		(1ULL << 63)
+
 /**
  * Object attributes (metadata).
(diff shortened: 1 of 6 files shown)


src/include/daos_obj.h
```diff
@@ -781,7 +798,7 @@ daos_obj_fetch(daos_handle_t oh, daos_handle_t th, uint64_t flags,
  * \param[in]	th	Optional transaction handle to update with.
  *			Use DAOS_TX_NONE for an independent transaction.
  *
- * \param[in]	flags	Update flags (conditional ops).
+ * \param[in]	flags	Update flags (conditional ops or DAOS_IOF_FLATTEN).
```
liuxuezhao, line 801: add DAOS_IOF_FLATTEN flag for daos_obj_update() because in wiki page it says -
"DAOS can provide API to allow user to indicate immutability of WORM objects:
   Application/middleware calls write(obj) with “DONE” flag."

But adding this flag to normal update request will make the obj_update() be more complex - for example request targets are different (normal update only to one dkey/group, but flatten to all groups' shards) this will make some detailed code not very convenient to handle, and possibly cause update RPC interoperability issue.

How about remove this flag's usage, only provide daos_obj_flatten() API?

mchaarawi: yes this would even be worse coming in from the dfs or array api where an "update" can turn into multiple daos_obj_update(). so just a single API call is better here.
that being said, the name flatten is ambiguous and has several meanings from a user perspective. there is no indication here that flatten means make immutable.
so i would suggest to rename flatten to some more informative for the user like daos_obj_set_immutable() or set_ro(). that being said, what if we want to add more properties later on the object?
maybe better to just add a new API:
daos_obj_set_prop(oid, flags);
and DAOS_OBJ_RO being one of the flags (only one for now)?

liuxuezhao: Ok, let's only add a single API if no objection from others.

I think probably no plan to add object level property for now.
So maybe just one API - daos_obj_set_ro () is good...

mchaarawi: sure, until there becomes a plan to add more ;-)
so IMO better to plan ahead always for APIs rather than make API changes later.

NiuYawei: I agree that single API like daos_obj_set_prop(DAOS_OBJ_PROP_RO) looks better.

liuxuezhao: daos_obj_set_prop() looks will mislead user think daos will provide object level property, I think that is not true because object level property will cause too much metadata , not we only have property for pool and container level.

mchaarawi: i see. anyway no strong opinion on this one. both ways sound OK to me.
i'd like to discuss though how this applies in a more generic sense, like "flattening" an entire container. we cannot expect to drive this all from the client side per object. i think better to discuss this on a meeting for now.

liuxuezhao: in last week's md on ssd meeting, @gnailzenh suggested to start with container level flattening (flaten all objects for the container).

liuxuezhao: @mchaarawi added container property related change for it, please check if it is fine wit you, thanks.

PR #12942, opened 2023-08-25
DAOS-14203 test: Bump system_ram_reserved for aggregation basic

tanabarr:
Increase system_ram_reserved to 64gib to prevent available memory
check failures caused by high memory resource usage.

Test-tag: test_basic_aggregation
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/tests/ftest/aggregation/basic.yaml
+++ b/src/tests/ftest/aggregation/basic.yaml
@@ -9,6 +9,7 @@ server_config:
     0:
       log_mask: INFO
       storage: auto
+  system_ram_reserved: 64
 pool:
   scm_size: 50000000000
   nvme_size: 500000000000


knard-intel (approved):

src/tests/ftest/aggregation/basic.yaml
```diff
@@ -9,6 +9,7 @@ server_config:
     0:
       log_mask: INFO
       storage: auto
+  system_ram_reserved: 64
```
daltonbohning, line 12: Why does just this test need this? Wouldn't others need the same?

kjacque (commented):
I have the same question as Dalton. This value seems rather high -- is there something special about this test?

src/tests/ftest/aggregation/basic.yaml
tanabarr: I'm not sure why the usage of memory on a single host was so high in the specific instance of the test running in CI, I'm not sure whether that's to do with the test itself. As mentioned in one of the above comments it's not expected that a host running the test would have other applications using so much memory during the test so that's why such a large buffer is required to take this odd scenario into account. As observed in the ticket comment: https://daosio.atlassian.net/browse/DAOS-14203?focusedCommentId=119749 it's strange that one host is reporting 163 available RAM and the other 123 (which causes the failure) when they both have the same amount of installed RAM. There is one other test yaml that has the value set to 64, think it's one of the DMG tests. This scenario where 64/187 gib of RAM is being used elsewhere on the host when trying to run DAOS wouldn't be realistically expected in production, hence why the check is failing and why adjustment of the tunable is required.

kjacque (approved):
I'll approve as I don't think it's a harmful thing to change, but it would be nice to know why one of the machines has much less RAM available than the other. If there's some issue with the configuration of those nodes, it would be good to know.

daltonbohning (changes requested):
My concern is that, according to the ticket, other tests are having the same issue:
https://daosio.atlassian.net/browse/DAOS-14203?focusedCommentId=119428
At a minimum, we need to update those tests. But storage format occurs before the actual test steps, and in some example configs, there isn't anything special. I.e. if `storage: auto` generates a config that requests too much memory, then pretty much any HW test could have similar issues.
- aggregation/punching.yaml
```
server_config:
  name: daos_server
  engines_per_host: 1
  engines:
    0:
      log_mask: INFO
      storage: auto
```
- pool/create_query.yaml
```
server_config:
  name: daos_server
  engines_per_host: 1
  engines:
    0:
      storage: auto
```

knard-intel:
@daltonbohning and @kjacque,
For my understanding, could you confirm that this issue is appearing as there is some unexpected discrepancies on the resources of the CI VMs ?
From what I understand, this patch is kind of workaround to deal with this.
Thus, for the long time should it not be managed at the CI level with providing some features allowing to define the minimal resources and/or load of a VM used for running a test ?

phender:
> @daltonbohning and @kjacque, For my understanding, could you confirm that this issue is appearing as there is some unexpected discrepancies on the resources of the CI VMs ? From what I understand, this patch is kind of workaround to deal with this. Thus, for the long time should it not be managed at the CI level with providing some features allowing to define the minimal resources and/or load of a VM used for running a test ?

@knard-intel the tests failing in this ticket are HW tests.

kjacque:
> I think that if we don't want to use the `system_ram_reserved` to cater for this imbalance we should investigate why the memory resource usage disparities exist and fix those.

I agree with this. There is clearly something odd happening with the CI systems in the moments where this bug occurs. I think it's important to know the actual root cause and fix the issue in CI.

knard-intel:
> > @daltonbohning and @kjacque, For my understanding, could you confirm that this issue is appearing as there is some unexpected discrepancies on the resources of the CI VMs ? From what I understand, this patch is kind of workaround to deal with this. Thus, for the long time should it not be managed at the CI level with providing some features allowing to define the minimal resources and/or load of a VM used for running a test ?
> 
> @knard-intel the tests failing in this ticket are HW tests.

Thanks for the clarification and my misunderstanding: I was thinking it was a VM test as for me HW tests were running with homogeneous nodes.
At the end, I am inline with the conclusions from @kjacque and @tanabarr that the fixing this CI is the proper way to this fix issue.

phender:
> > My concern is that, according to the ticket, other tests are having the same issue: https://daosio.atlassian.net/browse/DAOS-14203?focusedCommentId=119428 At a minimum, we need to update those tests. But storage format occurs before the actual test steps, and in some example configs, there isn't anything special. I.e. if `storage: auto` generates a config that requests too much memory, then pretty much any HW test could have similar issues.
> > 
> > * aggregation/punching.yaml
> > 
> > ```
> > server_config:
> >   name: daos_server
> >   engines_per_host: 1
> >   engines:
> >     0:
> >       log_mask: INFO
> >       storage: auto
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * pool/create_query.yaml
> > 
> > ```
> > server_config:
> >   name: daos_server
> >   engines_per_host: 1
> >   engines:
> >     0:
> >       storage: auto
> > ```
> 
> Right, it does seem that this issue is due to mismatch of RAM usage across test hosts as opposed to anything specific with the test. The situation where you have a test that uses 2 hosts and one of those is using 64gib more RAM (in applications outside of DAOS) is something specific to CI and would not be any issue you would experience in a production environment (a cluster administrator would not expect a dedicated DAOS storage server in the cluster to have such a discrepancy). The control-plane defaults are focused on catering for production environments first and offer config tunables for situations outside of that remit. I think that if we don't want to use the `system_ram_reserved` to cater for this imbalance we should investigate why the memory resource usage disparities exist and fix those. @phender @daltonbohning @kjacque @knard-intel @mjmac thoughts?

The `storage: auto` test yaml setting results in us defining `scm_size: 0` for the first/ram storage tier, e.g. https://build.hpdd.intel.com/job/daos-stack/job/daos/job/md-on-ssd-testing/87/artifact/Functional%20Hardware%20Medium/aggregation/basic.py/daos_configs.wolf-201/daos_server.yml/*view*/.  Should this not use the lowest common denominator between the available memory on the servers hosts to set the size?  This would seem to account for the imbalance.

I also have added a [debug PR](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-12960/2/) to attempt to collect memory info from the hosts during the harness server setup in hopes of sheading some light onto why the imbalance occurs in CI.

phender:
> I also have added a [debug PR](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-12960/2/) to attempt to collect memory info from the hosts during the harness server setup in hopes of sheading some light onto why the imbalance occurs in CI.

Of course all the tests passed: https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-12960/2/testReport/

tanabarr:
> Should this not use the lowest common denominator between the available memory on the servers hosts to set the size? This would seem to account for the imbalance.

The reason we don't use the global minima of `mem_available` to avoid confusing situations where, because of a resource usage spike on a single host during start-up, a significantly low value is used across the cluster as the SCM size. https://daosio.atlassian.net/browse/DAOS-12921 describes the implementation of SCM size auto calculation and validation including discussion as to rationale for using `mem_total` to derive the value to be used across the cluster. Other solutions were suggested and discussed but in summary; the total memory was used in conjunction with calculated reservations for both systemwide and engine specific usages to arrive at a value that would be reasonable to use across a cluster without relying on a variable metric (`mem_available`) that would change independently across a cluster and have the potential to drop significantly at start-up time causing an unrepresentative value to be selected resulting in underutilisation of RAM across the cluster. See discussion details in the ticket comments.

tanabarr:
@phender @daltonbohning should we learned this is an interim step whilst we figure out CI discrepancies ?

PR #12948, opened 2023-08-25
DAOS-14163 cart: Do not access rpc after decref

ashleypittman:
Do not log target rank in decref as rpc there is no reference held.

Required-githooks: true


mchaarawi (approved):

ashleypittman:
Adding gatekeeper early as we're going into a long weekend here in the UK.

PR #12949, opened 2023-08-25
DAOS-10942 utils: chmod should ignore unsupported bits

chowes:
The setuid, setgid, and sticky bit can cause fatal errors when the datamover tool sets file permissions after copying a file, since these are not supported by DFS. We can just ignore this bit when calling dfs_chmod.

### Gatekeeper:


src/control/cmd/daos/filesystem.go
```diff
@@ -44,9 +44,10 @@ type fsCmd struct {
 type fsCopyCmd struct {
 	daosCmd
 
-	Source   string `long:"src" short:"s" description:"copy source" required:"1"`
-	Dest     string `long:"dst" short:"d" description:"copy destination" required:"1"`
-	Preserve string `long:"preserve-props" short:"m" description:"preserve container properties, requires HDF5 library" required:"0"`
+	Source          string `long:"src" short:"s" description:"copy source" required:"1"`
+	Dest            string `long:"dst" short:"d" description:"copy destination" required:"1"`
+	Preserve        string `long:"preserve-props" short:"m" description:"preserve container properties, requires HDF5 library" required:"0"`
+	IgnoreNotsupMod bool   `long:"ignore-notsup-mode-bits" description:"squash unsupported special mode bits when copying to DFS" required:"0"`
```
mjmac, line 50: minor: With the assumption that there may be more areas of POSIX noncompliance found, it might make sense to generalize this a little to `IgnoreUnsupported`, with the flag being `--ignore-unsupported`, and the help text being "ignore unsupported filesystem features when copying to DFS". From a maintainability standpoint, I could see this being a little bit of a moving target otherwise. From a UX standpoint, we probably wouldn't want to ask the user to have to figure out which combinations of unsupported things they want to ignore, if we need to add more things that are known not to work but shouldn't be fatal errors.

src/include/daos_fs_sys.h
```diff
@@ -180,15 +180,17 @@ dfs_sys_access(dfs_sys_t *dfs_sys, const char *path, int mask, int flags);
...
- * \param[in]	dfs_sys Pointer to the mounted file system.
- * \param[in]	path	Link path of object.
- * \param[in]	mode	New permission access modes. For now, we don't support
- *			the sticky bit, setuid, and setgid.
+ * \param[in]	dfs_sys 		Pointer to the mounted file system.
+ * \param[in]	path			Link path of object.
+ * \param[in]	mode			New permission access modes. For now,
+ * 					we don't support the sticky bit, setuid,
+ * 					and setgid.
+ * \param[in]	ignore_notsup_mod 	Ignore unsupported file mode bits.
```
mjmac, line 188: Likewise here... We're making a public API change, so it would probably be best to generalize, as I suspect that callers won't really care about having fine-grained control over which unsupported features should be ignored.

src/utils/daos_hdlr.c
```diff
@@ -523,8 +523,12 @@ file_close(struct cmd_args_s *ap, struct file_dfs *file_dfs, const char *file)
...
 
 static int
-file_chmod(struct cmd_args_s *ap, struct file_dfs *file_dfs, const char *path,
-	   mode_t mode)
+file_chmod(struct cmd_args_s *ap,
+	   struct file_dfs *file_dfs,
+	   const char *path,
+	   mode_t mode,
+	   bool ignore_notsup_mod,
+	   uint64_t *num_chmod_enotsup)
```
mjmac, line 531: As this is an internal function, I think it's OK to keep the specific parameters especially in light of the inout param used for reporting.

src/control/cmd/daos/filesystem.go
chowes: Seems reasonable, I feel like that argument was pretty clunky anyway. This should be more usable.

src/include/daos_fs_sys.h
chowes: Done.

ashleypittman (dismissed):

mjmac:
Latest revision LGTM, but I've added @mchaarawi for another client reviewer.

src/include/daos_fs_sys.h
```diff
@@ -180,15 +180,17 @@ dfs_sys_access(dfs_sys_t *dfs_sys, const char *path, int mask, int flags);
...
+ * 				we don't support the sticky bit, setuid,
+ * 				and setgid.
+ * \param[in]	ignore_unsup	Ignore unsupported filesystem features
  *
- * \return		0 on success, errno code on failure.
+ * \return				0 on success, errno code on failure.
  */
 int
-dfs_sys_chmod(dfs_sys_t *dfs_sys, const char *path, mode_t mode);
+dfs_sys_chmod(dfs_sys_t *dfs_sys, const char *path, mode_t mode, bool ignore_unsup);
```
mchaarawi, line 193: this is a major version API change (it breaks the existing API); so not acceptable as is.

mchaarawi (changes requested):
This can't land as is because of the API change.
@daltonbohning mpifileutils uses this API, no?
We would need to update mfu to use this new API but also add proper version checks (which this PR does not bump) to work with the existing API (2.4) and the new API (2.5).

alternatively, it's probably easier to just add a new API instead of breaking the API

src/include/daos_fs_sys.h
daltonbohning: Right. For example, third-party mpifileutils uses this function. Which would require a recompile. And, unlikely but possible, some unknown user application.

mjmac:
> This can't land as is because of the API change.
> 
> alternatively, it's probably easier to just add a new API instead of breaking the API
>
Another option, perhaps, would be to return to the original implementation that didn't change the API and just emitted a warning for unsupported bits. I don't feel strongly either way, but I gather that the API change was driven by some feedback that the behavior should be configurable.

src/utils/daos_hdlr.c
```diff
@@ -538,7 +538,16 @@ file_chmod(struct cmd_args_s *ap, struct file_dfs *file_dfs, const char *path,
...
+		rc = dfs_sys_chmod(file_dfs->dfs_sys, path, mode, ignore_unsup);
+		if (rc == ENOTSUP && !ignore_unsup) {
+			/* If the file has unsupported mode bits, unset them
+			 * and retry the chmod. We track these errors so they
+			 * can be surfaced to the user at the end of the copy
+			 * operation.
+			 */
+			(*num_chmod_enotsup)++;
+			rc = dfs_sys_chmod(file_dfs->dfs_sys, path, mode, true);
+		}
```
daltonbohning, line 550: We could avoid an API change and the need for a new function by handling the bits here instead of within `dfs_sys_chmod`. The benefit to handling here is that we could unset the bits _**before**_ calling `dfs_sys_chmod`, avoiding the extra call.
The downside is that if other applications use `dfs_sys_chmod`, they potentially need similar logic.

@mchaarawi @ashleypittman What do you think about handling here vs in the `dfs_sys` API?

daltonbohning (changes requested):

daltonbohning (commented):
I don't think you have visibility into the Jenkins CI, but will you please include this line in future commit messages so extra tests are ran?
```
Features: daos_fs_copy
```

mchaarawi:
> > This can't land as is because of the API change.
> > alternatively, it's probably easier to just add a new API instead of breaking the API
> 
> Another option, perhaps, would be to return to the original implementation that didn't change the API and just emitted a warning for unsupported bits. I don't feel strongly either way, but I gather that the API change was driven by some feedback that the behavior should be configurable.

right, so just emitting a warning has its downside where it's not always visible to the caller that what they did is not supported. But this does not mean that we can just break the API without any regards to proper versioning and handling for API & ABI compatibility!
This PR requires several things to be acceptable:
1) bump the API major version to X+1
2) patch mpifileutils to have compile time check for the API version (if <= X use old API; if > X use new API). see an example here: https://github.com/hpc/ior/blob/main/src/aiori-DFS.c#L519
3) update RPM packaging for our CI for mpifileutils after 2 lands to MFU.

and yes it's not an easy change to break the API and it shouldn't be..
which is why i suggested the alternative to add a new API function instead of breaking the API:
dfs_sys_chmod_wflags(..., uin32_t flags); where one of the flags can be allow_unsupported bit. and use that function where this functionality is desired.

the other suggestion as Dalton was saying in his comment is to handle the change in where this is called without adding the new API (or breaking the old one).
I do not have a strong opinion on either as long as it's done properly.

chowes:
I agree with Dalton's suggestion, let's just move the error handling outside of dfs_sys_chmod. Other applications would need to update anyway to support the parameter, so I don't think the extra burden of adding the error handling is worth the burden of workout around a breaking API change.

src/include/daos_fs_sys.h
```diff
@@ -180,12 +180,13 @@ dfs_sys_access(dfs_sys_t *dfs_sys, const char *path, int mask, int flags);
...
- * \param[in]	mode	New permission access modes. For now, we don't support
- *			the sticky bit, setuid, and setgid.
+ * \param[in]	dfs_sys 	Pointer to the mounted file system.
+ * \param[in]	path		Link path of object.
+ * \param[in]	mode		New permission access modes. For now,
+ * 				we don't support the sticky bit, setuid,
+ * 				and setgid.
  *
- * \return		0 on success, errno code on failure.
+ * \return				0 on success, errno code on failure.
```
daltonbohning, line 189: To undo these unnecessary format changes, probably can just do
```
git checkout origin/master -- src/include/daos_fs_sys.h
```

src/client/dfs/dfs_sys.c
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2018-2022 Intel Corporation.
+ * (C) Copyright 2018-2023 Intel Corporation.
```
daltonbohning, line 2: Minor, but tricky to backout the githook copyright change. I think you have to
1. Manually change back to 2022
2. `git commit --no-verify` to skip the githooks
3. Continue with other changes
4. Make sure your last commit was NOT ran with `--no-verify` so the githooks take affect for subsequent changes

src/tests/suite/dfs_sys_unit_test.c
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2019-2022 Intel Corporation.
+ * (C) Copyright 2019-2023 Intel Corporation.
```
daltonbohning, line 2: And here...

daltonbohning (commented):
Content LGTM - thanks! Minor comments about backing out the no-longer-needed stuff

daltonbohning (approved):

src/utils/daos_hdlr.c
```diff
@@ -539,6 +543,16 @@ file_chmod(struct cmd_args_s *ap, struct file_dfs *file_dfs, const char *path,
 		}
 	} else if (file_dfs->type == DAOS) {
 		rc = dfs_sys_chmod(file_dfs->dfs_sys, path, mode);
+		if (rc == ENOTSUP && !ignore_unsup) {
```
mchaarawi, line 546: Im nor sure i quite understand why a retry is needed here.
so the user has not requested ignore_unsup, but you still want to ignore_unsup?

chowes: Yeah, the idea is...

with ignore_unsup:
- unset the bits before the call to chmod
- there will be no error, we won't log a warning, and we exit 0

without ignore_unsup:
- the call to chmod fails
- unset the bits and try again
- print a warning and exit non-zero

Some history of the discussion in the bug: https://daosio.atlassian.net/browse/DAOS-10942

daltonbohning: Looks like with and without `--ignore-unsupported`, the unsupported mode bits are removed. By default, `num_chmod_enotsup` is reported at the end. But with `--ignore-unsupported` it's silent.

Maybe to avoid a retry we could do something like this upfront?
```
if (!ignore_unsup && mode & (S_ISVTX | S_ISGID | S_ISUID))
    (*num_chmod_enotsup)++;
mode &= ~(S_ISVTX | S_ISGID | S_ISUID);
```

chowes: That seems reasonable. The only downside I can see is that maybe we eventually support these bits and it's not an error anymore. In that case we'd probably want to update this code anyway though.

daltonbohning: Yeah, I think we'd want to update this anyway if we support in the future, and if at all I expect it would be far off

mchaarawi: if you are going to update this please also merge latest master to your branch as it is lagging behind a bit now.

chowes: Shoot, my attempt to rebase brought in a bunch of random commits. Sorry, I'm not used to github yet.

daltonbohning: The flow is usually
```
git checkout my_branch
git fetch
git merge origin/master
```

chowes: Yeah, I'm talking to Mike and Jeff and they are saying that github doesn't handle rebases well. Okay let me see if I can fix it. If not, I'll create a new PR.

daltonbohning: Did you use `rebase` or `merge`? `rebase` is trickier because it re-writes history (i.e. commits end up with different hashes). It's better to use `merge` after people have started reviewing so the history of commits is preserved.
Another common thing is if you do `git merge master`, your local `master` might not be up-to-date with the remote `master`. So you either have to do this
```
git checkout master
git fetch
git pull
git checkout my_branch
git merge master
```
or (simpler I think)
```
git checkout my_branch
git fetch
git merge origin/master
```

chowes: I did a rebase the first time. I think it should be okay now.

daltonbohning (approved):

ashleypittman (approved):

daltonbohning:
Single test failure in https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-12949/20/testReport/
is existing and unrelated to these changes:
https://daosio.atlassian.net/browse/DAOS-14450

mchaarawi (approved):

src/control/cmd/daos/filesystem.go
```diff
@@ -107,6 +109,10 @@ func (cmd *fsCopyCmd) Execute(_ []string) error {
 	cmd.Infof("    Files:       %d", ap.fs_copy_stats.num_files)
 	cmd.Infof("    Links:       %d", ap.fs_copy_stats.num_links)
 
+	if ap.fs_copy_stats.num_chmod_enotsup > 0 {
+		return errors.New(fmt.Sprintf("Copy completed successfully, but %d files had unsupported mode bits that could not be applied. Run with --ignore-unsupported to suppress this warning.", ap.fs_copy_stats.num_chmod_enotsup))
```
tanabarr, line 113: errors.Errorf("...", fargs) should be used to create error from format string

tanabarr (commented):
PRs touching go code should probably be reviewed by a go developer before landing

daltonbohning:
> PRs touching go code should probably be reviewed by a go developer before landing

It was reviewed by a Go developer, but just not approved by one.

PR #12968, opened 2023-08-29
DAOS-14251 control: Allow premounted empty tmpfs

mjmac:
In the special case where the tmpfs has already been
mounted but is empty, don't skip NVMe format and
configuration. Enables the use case of running
daos_server in a container with an external tmpfs.

Also removes unnecessary recreate-superblocks logic.


src/control/server/server_utils.go
```diff
@@ -540,6 +540,29 @@ func checkEngineTmpfsMem(srv *server, ei *EngineInstance, mi *common.MemInfo) er
...
 
+	// In the event that tmpfs was already mounted, we need to verify that it
+	// is the correct size and that the memory usage still makes sense.
+	if isMounted, err := ei.storage.ScmIsMounted(); err == nil && isMounted {
+		usage, err := ei.storage.GetScmUsage()
+		if err != nil {
+			return errors.Wrap(err, "unable to check tmpfs usage")
+		}
+		if usage.TotalBytes != memRamdisk {
+			// Check that the existing ramdisk is at least 90% of the
```
kjacque, line 551: Wondering what the reasoning is for the 90% allowance.

kjacque (dismissed):
I'm glad to see --recreate-superblocks finally being deprecated. A couple questions, nothing I'd block on. Good work.

src/control/server/server_utils.go
```diff
@@ -540,6 +540,29 @@ func checkEngineTmpfsMem(srv *server, ei *EngineInstance, mi *common.MemInfo) er
...
+			// calculated optimal ramdisk size and not larger.
+			if usage.TotalBytes < memRamdisk*90/100 {
```
tanabarr, line 553: This first check should be replaced by `if err := checkMemForRamdisk(log, memRamdisk, usage.TotalBytes); err != nil { return err } else if { ...` to avoid duplication and "magic" numbers.

src/control/server/server_utils.go
```diff
@@ -540,6 +540,29 @@ func checkEngineTmpfsMem(srv *server, ei *EngineInstance, mi *common.MemInfo) er
...
+				return storage.FaultRamdiskBadSize(usage.TotalBytes, memRamdisk)
+			} else if usage.TotalBytes > memRamdisk {
+				return storage.FaultRamdiskBadSize(usage.TotalBytes, memRamdisk)
+			}
+		}
+		// Looks OK, so update the required amount to match available so that
+		// we don't fail the check below.
+		memRamdisk = memAvail
```
tanabarr, line 561: Rather than cooking the test, we should just return here.

src/control/server/server_utils.go
```diff
@@ -540,6 +540,29 @@ func checkEngineTmpfsMem(srv *server, ei *EngineInstance, mi *common.MemInfo) er
...
 	memAvail := uint64(mi.MemAvailableKiB) * humanize.KiByte
 
+	// In the event that tmpfs was already mounted, we need to verify that it
+	// is the correct size and that the memory usage still makes sense.
+	if isMounted, err := ei.storage.ScmIsMounted(); err == nil && isMounted {
+		usage, err := ei.storage.GetScmUsage()
+		if err != nil {
+			return errors.Wrap(err, "unable to check tmpfs usage")
+		}
+		if usage.TotalBytes != memRamdisk {
```
tanabarr, line 550: why are we checking this, in what case would the RAMdisk equal the entire amount of RAM?

tanabarr (changes requested):

src/control/server/server_utils.go
```diff
@@ -540,6 +540,29 @@ func checkEngineTmpfsMem(srv *server, ei *EngineInstance, mi *common.MemInfo) er
...
+			// Check that the existing ramdisk is at least 90% of the
+			// calculated optimal ramdisk size and not larger.
+			if usage.TotalBytes < memRamdisk*90/100 {
+				return storage.FaultRamdiskBadSize(usage.TotalBytes, memRamdisk)
+			} else if usage.TotalBytes > memRamdisk {
+				return storage.FaultRamdiskBadSize(usage.TotalBytes, memRamdisk)
+			}
```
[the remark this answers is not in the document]
mjmac: Resolved with removal of minimum size check.

src/control/server/server_utils.go
mjmac: Resolved with removal of minimum size check.

src/control/server/server_utils.go
mjmac: Resolved with removal of minimum size check.

src/control/server/server_utils.go
mjmac: Yep, I was thinking that the additional debug logging might be useful, but I agree now that it is probably more confusing than helpful. Just added a debug log indicating that the existing tmpfs will be used and returned early.

src/control/server/server_utils.go
mjmac: Resolved with removal of minimum size check.

kjacque (approved):
LGTM.

tanabarr (approved):

PR #12972, opened 2023-08-30
DAOS-10028 client: Go bindings for Client API

mjmac:
Add Go bindings for DAOS Client API. This is an
initial pass, and will be expanded incrementally.

Adds the following API methods:
  Init()
  Fini()
  PoolConnect()
  PoolQuery()
  PoolDisconnect()
  ContainerCreate()
  ContainerDestroy()
  ContainerOpen()
  ContainerQuery()
  ContainerGetProperties()
  ContinaerSetProperties()
  ContainerClose()

Change-Id: I373a5ce5c4fc64384b7923eb3fba449887fd8b1e
Required-githooks: true


src/control/lib/daos/client/api.go
```diff
@@ -0,0 +1,289 @@
...
+	log logging.DebugLogger = logging.NewDebugLogger(io.Discard)
+)
+
+type (
+	// APIContext is an opaque type used to convey DAOS context across API methods.
+	APIContext struct {
+		sync.RWMutex
+		cancelCtx context.CancelFunc
+		poolConns map[string]*ConnHandle
+		contConns map[string]map[string]*ConnHandle
```
kjacque, line 37: I agree with getting rid of this stuff. I think letting the client deal with the handles directly is probably the best way to go.

I do think the context object is a nice way to handle daos_init and daos_fini.

src/control/lib/daos/client/api.go
```diff
@@ -0,0 +1,289 @@
+package client
```
kjacque, line 1: Do you think it would be crossing too many boundaries to include the client API directly in the "daos" package? In terms of writing applications, users would end up having to import both daos and daos/client every time.

src/control/lib/daos/client/util.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// daosError converts a return code from a DAOS API
+// call to a Go error.
+func daosError(rc C.int) error {
+	if rc == 0 {
+		return nil
+	}
+	return daos.Status(rc)
+}
```
kjacque, line 38: Duplicated from the daos package?

src/control/cmd/daos/acl.go
```diff
@@ -32,92 +23,19 @@ import (
...
+	/*props, cleanup, err := getContAcl(ap.cont)
 	if err != nil {
 		return nil, err
 	}
 	defer cleanup()
 
-	return convertACLProps(props), nil
+	return convertACLProps(props), nil*/
+	return nil, nil
 }
```
kjacque, line 39: Presumably the commented out functions are destined to be removed?

kjacque (commented):
This is a nice division of responsibilities. I know we already spoke about how to represent the API at this level (a fairly literal interpretation of the libdaos interface vs. object oriented approaches). I found myself wanting to build objects on top of this, similar to what pydaos is doing over the "raw" python bindings. Maybe we could consider that as a future extension to these changes. Wrapping the C code fairly tightly seems like a good first approach to me.

PR #12976, opened 2023-08-30
DAOS-623 build: Add option for compiling out D_ASSERT.

ashleypittman:
Add a new D_ABORT macro which always aborts.
Add a ASSERT_ENABLED scons option.
Update build to add compile flags to support this.
Resolve issues in codebase wrt braces.

Required-githooks: true


site_scons/prereq_tools/base.py
```diff
@@ -459,7 +459,7 @@ def __init__(self, env, opts):
                               ['gcc', 'covc', 'clang', 'icc'], ignorecase=2))
         opts.Add(EnumVariable('WARNING_LEVEL', "Set default warning level", 'error',
                               ['warning', 'warn', 'error'], ignorecase=2))
-
+        opts.Add(BoolVariable('ASSERT_ENABLED', 'D_ASSERT macro enabled', False))
```
jolivier23, line 462: Shouldn't the default be true?  Or at least enabled with dev or debug build?

ashleypittman: Yes, but there are a lot of compiler issues on various distros that I'm working through first.  Once this PR is stable as-is (and can pass daily-testing) then I'll change some of the GitHub actions builds to set it to False and change the default to True.

jolivier23 (approved):

src/cart/utils/crt_utils.c
```diff
@@ -375,8 +375,9 @@ crtu_dc_mgmt_net_cfg_rank_add(const char *name, crt_group_t *group,
 				true /* all_ranks */,
 				&crt_net_cfg_info,
 				&crt_net_cfg_resp);
-	if (opts.assert_on_error)
+	if (opts.assert_on_error) {
```
mchaarawi, line 378: even though i prefer this style and long time ago i had advocated for it, but this is not our coding style. If we want to change that, it's better to make the proposal at the pdt meeting before you do the change in everyone's code yourself.

mchaarawi (changes requested):
it looks like this has some bug fixes that need to be separated out and merged separatly.

as for the other style changes, see my comment in there. personally I just don't like these ticky tac changes that touch all the code at once.

src/cart/utils/crt_utils.c
ashleypittman: This is needed to prevent compile warnings, if D_ASSERT() is defined out then this code expands to the below
`if (cond)
    ;
`
which the compiler doesn't like as it's easy to make mistakes that way.  Adding in the braces makes it explicit that the conditional is expected and therefore silences the compiler warning.  It would have been possible to add -Wempty-body to silence these however on reflection I thought that the compiler warning was a reasonable one, hence why I changed the code where necessary.

ashleypittman:
> it looks like this has some bug fixes that need to be separated out and merged separatly.

I don't think there are any bug-fixes here, there are two places where D_ASSERT() had side effects which this PR addresses, but without also having this PR then there's no bug.  One of these only affects logging anyway, and the other if it happens will be a memory leak so I don't see the value in seperate PRs.

src/bio/bio_buffer.c
```diff
@@ -551,6 +551,19 @@ iterate_biov(struct bio_desc *biod,
...
+				 * the compiler knows about what is behind the pointer in
+				 * different cases and the writes here would be a buffer overflow
+				 * for different cb functions.  The compiler does not like this and
+				 * only allows the code to compile based on the presence of the
+				 * assert() - and D_ASSERT() may be disabled so add an additional
+				 * call to assert() itself.  Overall the logic of modifying data
+				 * should be in the cb itself as it's not really a void * depsite
+				 * what it's declared as.
+				 */
+				assert(i < arg->ca_sgl_cnt);
```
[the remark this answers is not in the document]
mchaarawi: in that case can the D_ASSERT be just removed instead of having 2 ?

src/cart/crt_swim.c
```diff
@@ -819,7 +819,7 @@ static swim_id_t crt_swim_get_dping_target(struct swim_context *ctx)
 {
 	struct crt_grp_priv	*grp_priv = crt_gdata.cg_grp->gg_primary_grp;
 	struct crt_swim_membs	*csm = &grp_priv->gp_membs_swim;
-	struct crt_swim_target	 cst;
+	struct crt_swim_target   cst      = {0};
```
mchaarawi, line 822: is that a bug fix? do we rely on cst fields being 0ed?

src/container/cli.c
```diff
@@ -816,15 +817,17 @@ cont_open_complete(tse_task_t *task, void *data)
...
-	/* log metadata times. NB: ctime_r inserts \n so don't add it in the D_DEBUG format */
 	rc = daos_hlc2timestamp(arg->coa_info->ci_md_otime, &otime_sec);
 	if (rc)
 		goto out;
 	rc = daos_hlc2timestamp(arg->coa_info->ci_md_mtime, &mtime_sec);
 	if (rc)
 		goto out;
-	D_ASSERT(otime_str == ctime_r((const time_t *)&otime_sec, otime_str));
-	D_ASSERT(mtime_str == ctime_r((const time_t *)&mtime_sec, mtime_str));
+	out_str = ctime_r((const time_t *)&otime_sec, otime_str);
```
mchaarawi, line 826: why is this change required?

src/container/cli.c
```diff
@@ -1232,15 +1236,16 @@ cont_query_complete(tse_task_t *task, void *data)
...
-	/* log metadata times. NB: ctime_r inserts \n so don't add it in the D_DEBUG format */
 	rc = daos_hlc2timestamp(arg->cqa_info->ci_md_otime, &otime_sec);
 	if (rc)
 		goto out;
 	rc = daos_hlc2timestamp(arg->cqa_info->ci_md_mtime, &mtime_sec);
 	if (rc)
 		goto out;
-	D_ASSERT(otime_str == ctime_r((const time_t *)&otime_sec, otime_str));
-	D_ASSERT(mtime_str == ctime_r((const time_t *)&mtime_sec, mtime_str));
+	out_str = ctime_r((const time_t *)&otime_sec, otime_str);
```
mchaarawi, line 1245: why is this change required?

src/vos/vos_gc.c
```diff
@@ -127,10 +128,15 @@ gc_drain_evt(struct vos_gc *gc, struct vos_pool *pool, daos_handle_t coh,
 
 	D_DEBUG(DB_TRACE, "drain %s evtree, creds=%d\n", gc->gc_name, *credits);
 	rc = evt_drain(toh, credits, empty);
-	D_ASSERT(evt_close(toh) == 0);
+	rc2 = evt_close(toh);
```
mchaarawi, line 131: that sounds like another bug fix?

mchaarawi (commented):
im still not convinced we need this flag. could you explain why this is needed if release builds, which supposedly should disable asserts, set NDEBUG which underneath just makes the assert() a no-op?

src/bio/bio_buffer.c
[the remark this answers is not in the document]
ashleypittman: No, the code won't compile unless there's a real assert here.  See the comment I added.

src/cart/crt_swim.c
ashleypittman: I can't see a code-path where this matters but the compiler thinks there is one, so I think this is a case of making the compiler happy.

src/bio/bio_buffer.c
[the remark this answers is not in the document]
mchaarawi: i saw the comment, and im asking if the D_ASSERT() can be removed and not the assert()

src/container/cli.c
ashleypittman: Without it otime_str is undefined and you get valgrind errors when it's logged on the D_DEBUG below.  This is the case where only debug output is affected by the side-effect of code within the assert.

src/vos/vos_gc.c
ashleypittman: Yes, this is a bug fix but a minor one.

ashleypittman:
> im still not convinced we need this flag. could you explain why this is needed if release builds, which supposedly should disable asserts, set NDEBUG which underneath just makes the assert() a no-op?

I don't know anything about release builds disabling asserts.  This PRs add the build option but doesn't actually change any of the builds to set it, I was hoping to change the GitHub actions builds so they would but that would mean extra docker options which isn't great - I'm still thinking how best to get this into CI.

Other than checking that assert() shouldn't have side-effects (and there are as you see two places where it was) my main aim here is to be able to run coverity with asserts disabled as I think that will give us better quality results, and hopefully catch any places where we're using assert incorrectly for input validation or checking success.

mchaarawi:
> I don't know anything about release builds disabling asserts. This PRs add the build option but doesn't actually change any of the builds to set it, I was hoping to change the GitHub actions builds so they would but that would mean extra docker options which isn't great - I'm still thinking how best to get this into CI.

theoretically any release build (not debugging enabled) would have C assert() calls be no-ops. i hope that is what our release builds already have. so im still questioning why this PR and the special flag is needed.
if you use BUILD_TYPE=release, i would assume assert() in the D_ASSERT() macro is a no-op. if not then that would be a release build bug.

ashleypittman:
> > I don't know anything about release builds disabling asserts. This PRs add the build option but doesn't actually change any of the builds to set it, I was hoping to change the GitHub actions builds so they would but that would mean extra docker options which isn't great - I'm still thinking how best to get this into CI.
> 
> theoretically any release build (not debugging enabled) would have C assert() calls be no-ops. i hope that is what our release builds already have. so im still questioning why this PR and the special flag is needed. if you use BUILD_TYPE=release, i would assume assert() in the D_ASSERT() macro is a no-op. if not then that would be a release build bug.

I don't think that's the case.

mchaarawi:
> I don't think that's the case.

ok yea that would a bug to fix. we need to have something like:
```
diff --git a/src/include/gurt/debug.h b/src/include/gurt/debug.h
index 1dbe4cc9b3..bc17513985 100644
--- a/src/include/gurt/debug.h
+++ b/src/include/gurt/debug.h
@@ -9,6 +9,9 @@
 
 #include <stdio.h>
 #include <stddef.h>
+#ifdef DAOS_BUILD_RELEASE
+#define NDEBUG
+#endif
 #include <assert.h>
 
 #include <gurt/debug_setup.h>
```
+ all the non bracket changes you have in this PR that accounts for assert() being a no-op

src/bio/bio_buffer.c
```diff
@@ -551,6 +551,19 @@ iterate_biov(struct bio_desc *biod,
...
+				 * This function is weird in that it takes a callback and a void *
+				 * but then modifies the void * based on which callback is provided,
+				 * the compiler knows about what is behind the pointer in
+				 * different cases and the writes here would be a buffer overflow
+				 * for different cb functions.  The compiler does not like this and
+				 * only allows the code to compile based on the presence of the
+				 * assert() - and D_ASSERT() may be disabled so add an additional
+				 * call to assert() itself.  Overall the logic of modifying data
+				 * should be in the cb itself as it's not really a void * depsite
+				 * what it's declared as.
```
NiuYawei, line 564: I don't quite follow this comment, if it's about modifying the "void *" argument, why the assert(i < arg->ca_sgl_cnt) would make compiler happy? Maybe compiler just like to ensure there is no overflow issue?

src/include/gurt/debug.h
```diff
@@ -325,6 +325,14 @@ int d_register_alt_assert(void (*alt_assert)(const int, const char*,
 		fflush(stdout);						\
 	} while (0)
 
+#ifdef DAOS_ASSERT_DISABLE
+
+#define D_ASSERT(cond)
+
+#define D_ASSERTF(cond, fmt, ...)
```
NiuYawei, line 332: Instead of changing every D_ASSERT() call to add "{ }", why don't we define it as "#define D_ASSERT(cond) do { } while (0)"? Other ongoing PR may still use the coding style of "if (cond) D_ASSERT()", they would be conflicting with this PR, and such conflicts is hard to be detected.

src/bio/bio_buffer.c
ashleypittman: The compiler knows the size of data (it's on the stack in the parent) from different calls, my assumption is the compiler is working out the smallest size that data can point to and these writes are beyond that.  This code however is safe in that the function pointer is also passed in and these writes only happen for a certain pointer value so in practice it's safe.  The compiler does have a point though, this code is unusual in modifying a void * before calling the cb_fn.

Perhaps the fix is to have the cb_fn take `i` as an arg as well?  I can try pushing a patch for that in a bit.

src/bio/bio_buffer.c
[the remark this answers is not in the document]
ashleypittman: Oh I see, yes the D_ASSERT() could be removed if the assert is present.  Ideally I'd like a better resolution here though, and if we are going to try and release with asserts disabled then we'll need one.

src/include/gurt/debug.h
ashleypittman: I've checked that and it would actually work.  I found the extra `{}` useful though to make it clearer that D_ASSERT() may be a no-op.  I'm not worried about conflicts, as I said this PR just adds a new compile option which works today as I've tested it but no-where does it enable that option, I'll have to address that in a future PR.

ashleypittman:
> > I don't think that's the case.
> 
> ok yea that would be a bug to fix. we need to have something like:
> 
> ```
> diff --git a/src/include/gurt/debug.h b/src/include/gurt/debug.h
> index 1dbe4cc9b3..bc17513985 100644
> --- a/src/include/gurt/debug.h
> +++ b/src/include/gurt/debug.h
> @@ -9,6 +9,9 @@
>  
>  #include <stdio.h>
>  #include <stddef.h>
> +#ifdef DAOS_BUILD_RELEASE
> +#define NDEBUG
> +#endif
>  #include <assert.h>
>  
>  #include <gurt/debug_setup.h>
> ```
> 
> plus all the non bracket changes you have in this PR to account for assert() being a no-op (the ones that fix the "use un-initialized" and "defined but not used" in the tests).

You'd need to file a ticket if you want that behaviour, it's not something that DAOS has ever done although I can see the advantage.  To be honest, given modern CPU speeds and coding styles I wouldn't expect it to have much performance difference however.

This PR is what you need to change in order to be able to support that option, in practice D_ASSERT() is not a wrapper around assert so NDEBUG doesn't apply here, but having a different default for ASSERT_ENABLED based on BUILD_TYPE would be possible in scons.  We'd need to address the bio_buffer.c issue properly if we did that and set NDEBUG as then the code wouldn't compile.

Some of the unit tests also check for an assert failing as part of the test so compiling out asserts leads to unit test failures, that's why I changed the unit test build to enable asserts as part of this PR - whilst that is the default I wanted to be explicit that the tests need it enable.

Finally, for this PR I disabled asserts, then fixed all compiler warnings on all distros and all test failures, one of them was a unitialised read, one was a memory leak plus the unit test failures I've already mentioned.  I would not be at all surprised if there was not more actual code being called inside assert that's not tested with the regular PR test flags so for now I think that enabling this in production might well introduce bugs - the ways to address that would be to perform a full test cycle with asserts disabled and to look into any new compiler warnings or coverity issues when they are - there are 100s of new compiler warnings about unused variables which need to be looked at.

Quite a lot of test code uses D_ASSERT() for error handing as well, this isn't OK in the product obviously, but is widely used in test code - and as yet we don't have a way of using different compiler flags for test vs product code.

I've also found at least one bug recently where assert was used for input validation on the server,  I'm sure we have quite a few more like this that we need to work through - another reason to land this PR and look at it under coverity.

ashleypittman:
> > This PR is what you need to change in order to be able to support that option, in practice D_ASSERT() is not a wrapper around assert so NDEBUG doesn't apply here, but having a different default for ASSERT_ENABLED based on BUILD_TYPE would be possible in scons.
> 
> i disagree. D_ASSERT is a wrapper around assert but just add some extra debugging and lets a user register their own assert. if we just disable assert() for release builds or builds with this new flag it's actually useful now since we do not get assert and we still get some debugging messages though that something is wrong.

I've re-read the macro and I perhaps see what you mean now, you want to keep the D_ASSERT() macro performing the check, log an error but then not exit?  I'm not really sure what the point of that would be to be honest, the cost of asserts is in performing the check so enabling the check but not acting on a negative result seems odd.

> > Some of the unit tests also check for an assert failing as part of the test so compiling out asserts leads to unit test failures, that's why I changed the unit test build to enable asserts as part of this PR - whilst that is the default I wanted to be explicit that the tests need it enable.
> 
> that's a bug in the test. those test should either be doing proper error checking or using the cmocka asserts.

I don't disagree.  Some of it is just an easier way of writing tests but some tests themselves rely on it.

mchaarawi:
> I've re-read the macro and I perhaps see what you mean now, you want to keep the D_ASSERT() macro performing the check, log an error but then not exit? I'm not really sure what the point of that would be to be honest, the cost of asserts is in performing the check so enabling the check but not acting on a negative result seems odd.

From my perspective, disabling assert is not about the cost, but about functionality. In production, an assert should never be seen really. assert() is just a developer tool.
I do agree that extra logging without the assert would be weird and all I'm saying that for a developer might still be valuable when debugging production issues, but we don't have to have that. We can also ifdef that logging out in release / non-assert builds in the D_ASSERT macro and still avoid having to add the {} in cond D_ASSERT usage in the code.

NiuYawei:
> > I've re-read the macro and I perhaps see what you mean now, you want to keep the D_ASSERT() macro performing the check, log an error but then not exit? I'm not really sure what the point of that would be to be honest, the cost of asserts is in performing the check so enabling the check but not acting on a negative result seems odd.
> 
> From my perspective, disabling assert is not about the cost, but about functionality. In production, an assert should never be seen really. assert() is just a developer tool. I do agree that extra logging without the assert would be weird and all I'm saying that for a developer might still be valuable when debugging production issues, but we don't have to have that. We can also ifdef that logging out in release / non-assert builds in the D_ASSERT macro and still avoid having to add the {} in cond D_ASSERT usage in the code.

D_ASSERT() should only be used to caught software bug, so if a D_ASSERT() (bug) is hit in production system, it could abort the engine immediately (with some debug information) to avoid further damaging to user data. I think it's quite similar to the Linux kernel panic or the Windows blue screen. :)

I agree we'd revise existing D_ASSERT() to avoid improper use cases (not for catching bug).

ashleypittman:
> D_ASSERT() should only be used to caught software bug, so if a D_ASSERT() (bug) is hit in production system, it could abort the engine immediately (with some debug information) to avoid further damaging to user data. I think it's quite similar to the Linux kernel panic or the Windows blue screen. :)

Agree.  Typical uses would be a pointer isn't NULL, ref count > 0 or detecting memory corruption.  If a developer can identify any way that an assert could fail that isn't the result of a bug then asserts should not be used.
 
> I agree we'd revise existing D_ASSERT() to avoid improper use cases (not for catching bug).

My suspicion is that in some cases we incorrectly use D_ASSERT() where it shouldn't be, I recently fixed a case where it was used for input validation and there are two minor bug-fixes in this PR where D_ASSERT() itself has side effects (closing a descriptor or fetching debug information).

The aim I have with this work is to be able to run coverity on the source with D_ASSERT() disabled, in theory if D_ASSERT() is used correctly this should have the same results as coverity does today, any check that only passes because of the presence of D_ASSERT() needs looking at and possibly changing not to use D_ASSERT.

It's for this reason that I think adding the `{}` is the right thing to do, D_ASSERT() might be a no-op and for the purpose of correct error handling and flow through the code it should be treated as though it doesn't so putting it after a if statement without braces looks and feels weird - you should be able to logically or practically remove the D_ASSERT() code without it affecting the code around it.

As to the coding standards, they say that if any code block after a if/else clause has braces then all code blocks for that if chain should have braces, so therefore braces are optional for simple if statements.  I know clang-format has an option to remove them but we don't enable that option, I don't know if it has an option to add them or not - but I'm not looking at changing this across the tree or changing either the coding standard itself or checks here, I do think that having explicit braces for D_ASSERT() works to highlight the fact that it may be a no-on and shouldn't have any side effects.

mchaarawi:
I still do not think this patch adds value if assert is not going to be disabled in release builds. Several engineers expressed concern to do that atm, so making the assert a no-op is not going to happen soon as was stated in the PDT. 
just adding an option to do that specifically is not going to do anything or be used anywhere other than your use. if you want to run Coverity, you can just do it on the branch and we do not need to merge this to master.

This is also disabling only D_ASSERT(). To do this properly, you need to disable assert() by defining NDEBUG as i had stated in my patch. we do use assert() plainly in the code instead of D_ASSERT() in several places and those needs to be made no-ops too.

As for the braces, I do understand your point, but i still disagree with making this change across the code base for a couple of reasons:
1) it's unnecessary and it can be avoided by just tweaking the macro as Niu mentioned. 
2) anytime a developer adds such an assert in a conditional without {} they are going to either have to fix in a follow on PR or repush the PR if we build this option by default. just increasing latency because our coding standard does not require braces for 1 line conditional and many will be doing that.

mchaarawi:
> Maintaining a branch in order to run coverity checks on is not good practice, we should be running coverity often and on the master/release branches - and to get the most of the out tool then D_ASSERT and assert both need to be disabled as you say.

Coverity on master and release branches already run often. what you are suggesting is running coverity on a special build that no one is going to use. so there is no difference in running it on a separate branch.
several developers expressed concern on removing assert. so no they cannot be disabled, even if they should IMO. If we do end up disabling them, again it should be in release builds not a separate flag that will not be used anywhere.

> We already compile in CI with a broad spectrum of compilers that have different levels of checking etc, so there are already cases where that happens, gcc on fedora used to be particularly fussy but I think el9 is newer currently. It's rare but this is a small cost that we already bare.

this is unrelated IMO. im just talking about this particular coding style that is very common and is a price we do not have to pay that your change will make us pay. by just adjusting the macro, we can avoid this pain related to this and avoid this mass code change in 90% if files touched in this PR unnecessarily.

PR #13000, opened 2023-09-04
DAOS-14223 control: Add --meta-size to dmg pool create

tanabarr:
Add plumbing for new pool create option to specify size of blob on
meta-role SSDs. In MD-on-SSD phase II this can be significantly larger
than the VOS file in tmpfs in order to to support larger md capacity
with limited memory capacity,

Limit PR scope by only propagating new parameter to dRPC pool create
handler and only supporting the new size when manually specifying SCM
and NVMe storage sizes in dmg pool create call.

Further work required to:
- Propagate meta size throughout create call stack
- Expand meta size param support to auto-sizing create options
- Consolidate and improve control-plane pool create workflow and logic

Features: control
Required-githooks: true

### Gatekeeper:


src/control/cmd/dmg/pool.go
```diff
@@ -236,79 +274,80 @@ func (cmd *PoolCreateCmd) Execute(args []string) error {
...
 		if scmBytes == 0 {
 			return errors.Errorf("Not enough SCM storage available with ratio %s: "+
 				"SCM storage capacity or ratio should be increased", cmd.Size)
 		}
-
-		cmd.updateRequest(req, scmBytes, nvmeBytes)
-
+		req.TierBytes = []uint64{scmBytes, nvmeBytes}
+		req.TierRatio = nil
+		_ = ratio2Percentage(cmd, float64(scmBytes), float64(nvmeBytes))
```
kjacque, line 324: We don't need the returned value?

src/control/server/ctl_storage.go
```diff
@@ -212,6 +212,9 @@ func (cs *ControlService) scanAssignedBdevs(ctx context.Context, nsps []*ctl.Scm
 				continue
 			}
 
+			// NOTE DAOS-14223: This metadata size calculation won't necessarily match
+			//                  the meta blob size on SSD if --md-size is specified in
+			//                  pool create command.
```
kjacque, line 217: What are the ramifications of this?

src/mgmt/srv_pool.c
```diff
@@ -168,16 +168,17 @@ ds_mgmt_pool_svc_create(uuid_t pool_uuid, int ntargets, const char *group, d_ran
 }
 
 int
-ds_mgmt_create_pool(uuid_t pool_uuid, const char *group, char *tgt_dev,
-		    d_rank_list_t *targets, size_t scm_size, size_t nvme_size,
-		    daos_prop_t *prop, d_rank_list_t **svcp,
-		    int domains_nr, uint32_t *domains)
+ds_mgmt_create_pool(uuid_t pool_uuid, const char *group, char *tgt_dev, d_rank_list_t *targets,
+		    size_t scm_size, size_t nvme_size, daos_prop_t *prop, d_rank_list_t **svcp,
+		    int domains_nr, uint32_t *domains, size_t meta_blob_size)
```
kjacque, line 173: Is the engine code that uses the MD size not written yet?

kjacque (commented):
I'd like to know a bit more about the use case for this. Seems pretty fiddly for something an admin user might try to use.

src/control/server/ctl_storage.go
knard-intel: With MD-on-SSD, To compute the maximal storage usable on a NVMe disk to store DATA, you need to know the size that will be used for storing metadata, wall, etc on this NVMe device.
This part of the codes makes the assumption that the metadata size will be at most equal to the size of the VOS file stored into the SCM.  This size is then send to the engine to compute the size of the WAL associated to these VOS files.
With this new parameter, this assumption will not be any more valid and thus this part will need to be redesign.

knard-intel (commented):
Mostly OK for me.
It should be compatible with the create pool percentage as it.
However, the _usable_size_ field of the of a dmg storage scan will probably have to be updated.
The DMG commands could also check that the `--md_size` option is greater or equal than the `--scm-size` one.

src/mgmt/srv_pool.c
tanabarr: it's WIP, this is just the control-plane side

src/control/server/ctl_storage.go
tanabarr: the purpose of this PR is to enable the value to be passed into the engine, but the first iteration only supports manual setting. future iterations will provide auto calculation of the value based on total size or percentage specifications in the create call

src/control/cmd/dmg/pool.go
tanabarr: no, only called for validation purposes

src/proto/mgmt/pool.proto
```diff
@@ -32,6 +32,7 @@ message PoolCreateReq {
 	uint32 numranks = 11; // Number of target ranks to use (auto config)
 	repeated uint32 ranks = 12; // target ranks (manual config)
 	repeated uint64 tierbytes = 13; // Size in bytes of storage tiers (manual config)
+	uint64 meta_blob_size     = 15; // Size in bytes of metadata blob on SSD (manual config)
```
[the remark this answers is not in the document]
tanabarr: fixed

src/control/server/mgmt_pool_test.go
```diff
@@ -393,6 +393,20 @@ func TestServer_MgmtSvc_PoolCreate(t *testing.T) {
...
+				Uuid:         test.MockUUID(1),
+				Tierbytes:    []uint64{100 * humanize.GiByte, 10 * humanize.TByte},
+				MetaBlobSize: 2 * humanize.GiByte,
+				Properties:   testPoolLabelProp(),
+			},
+			expResp: &mgmtpb.PoolCreateResp{
+				//TierBytes: []uint64{100 * humanize.GiByte, 10 * humanize.TByte},
+				//MetaBlobSize: 2 * humanize.GiByte,
+				//TgtRanks: []uint32{0, 1},
+			},
```
kjacque, line 408: A bit of a surprising result, if the response is empty.

kjacque (commented):
Hard to tell the changes after the force push, but looking at the changed files, it looks reasonable. I have a minor comment, in addition to the earlier formatting comment.

src/control/server/mgmt_pool_test.go
tanabarr: This is actually a strange feature of how the test works, I had left the contents of the response commented to remind myself to revisit it, then I subsequently forgot :-) thanks for reminding me.
On L536 the mock drpc client is setup to return tc.expResp, PoolCreate() then simply returns tc.expResp and then on L555 we basically compare tc.expResp with itself.

As it stands I think TestServer_MgmtSvc_PoolCreate doesn't cover much of mgmtSvc.poolCreate() - which has been updated over time to address a number of edge cases and now has multiple code paths not all of which are covered in the unit tests. TestServer_MgmtSvc_calculateCreateStorage, TestServer_MgmtSvc_PoolCreateAlreadyExists and TestServer_MgmtSvc_PoolCreateDownRanks do a good job of covering some of these paths but we should probably look at evaluating what is actually covered by the unit tests and addressing any gaps as a technical debt task: https://daosio.atlassian.net/browse/DAOS-14351.

As for the changes in the scope of this ticket, looks like the main purpose of TestServer_MgmtSvc_PoolCreate is to flush out errors in the request and only 4 of 16 test cases provide tc.expResp. I think it's best to remove tc.expResp and instead just focus on error checking, with 4 of 16 cases passing with no error.

tanabarr:
> Hard to tell the changes after the force push, but looking at the changed files, it looks reasonable. I have a minor comment, in addition to the earlier formatting comment.

I apologise for the force push, I did it in order to save time when splitting out the storage calculation consolidation work into separate PR.

src/mgmt/srv_internal.h
```diff
@@ -64,9 +64,10 @@ void ds_mgmt_svc_put(struct mgmt_svc *svc);
 int ds_mgmt_group_update_handler(struct mgmt_grp_up_in *in);
 
 /** srv_pool.c */
-int ds_mgmt_create_pool(uuid_t pool_uuid, const char *group, char *tgt_dev, d_rank_list_t *targets,
+int
+    ds_mgmt_create_pool(uuid_t pool_uuid, const char *group, char *tgt_dev, d_rank_list_t *targets,
```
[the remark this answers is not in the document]
tanabarr: fixed

src/control/server/mgmt_pool_test.go
tanabarr: In the end I've just separated drpc return value from expected response to clarify intent at the cost of response value duplication in the 4 successful test cases.

tanabarr:
@ashleypittman can we ignore the suggested clang-format change? As observed by @kjacque this looks a bit odd:
```+++ b/src/mgmt/srv_internal.h
@@ -64,7 +64,8 @@ void ds_mgmt_svc_put(struct mgmt_svc *svc);
 int ds_mgmt_group_update_handler(struct mgmt_grp_up_in *in);

 /** srv_pool.c */
-int ds_mgmt_create_pool(uuid_t pool_uuid, const char *group, char *tgt_dev, d_rank_list_t *targets,
+int
+    ds_mgmt_create_pool(uuid_t pool_uuid, const char *group, char *tgt_dev, d_rank_list_t *targets,
                        size_t scm_size, size_t nvme_size, daos_prop_t *prop, d_rank_list_t **svcp,
                        int domains_nr, uint32_t *domains, size_t meta_blob_size);
```

ashleypittman:
> @ashleypittman can we ignore the suggested clang-format change? As observed by @kjacque this looks a bit odd:
> 
> ```
> @@ -64,7 +64,8 @@ void ds_mgmt_svc_put(struct mgmt_svc *svc);
>  int ds_mgmt_group_update_handler(struct mgmt_grp_up_in *in);
> 
>  /** srv_pool.c */
> -int ds_mgmt_create_pool(uuid_t pool_uuid, const char *group, char *tgt_dev, d_rank_list_t *targets,
> +int
> +    ds_mgmt_create_pool(uuid_t pool_uuid, const char *group, char *tgt_dev, d_rank_list_t *targets,
>                         size_t scm_size, size_t nvme_size, daos_prop_t *prop, d_rank_list_t **svcp,
>                         int domains_nr, uint32_t *domains, size_t meta_blob_size);
> ```

It is a bit odd and to me this is a bug in clang-format.  We are supposed to be putting the return types on their own lines and generally in practice that happens for functions but not declarations in header files.  As clang-format is working from a diff and just re-indenting the modified lines then what it's doing is trying to put the return on it's own line, but then (incorrectly) align the function name with the previous function name which is also badly aligned.  The way to make clang-format work and be happy here would be to apply formatting to the whole consecutive block of definitions which is overkill for a PR.

knard-intel (approved):
LGTM

src/control/server/engine/utils.go
```diff
@@ -141,8 +141,8 @@ var (
 		"CRT", "RPC", "BULK", "CORPC", "GRP", "LM", "HG", // CaRT subsystems
 		"EXTERNAL", "ST", "IV", "CTL",
 	}
-	errLogNameAllWithOther = errors.New("'all' identifier cannot be used with any other")
-	errLogNameAllInMasks   = errors.New("'all' identifier cannot be used in log mask level assignments")
+	errLogNameAllWithOther = errors.New("'all' identifier can not be used with any other log identifier")
+	errLogNameAllInMasks   = errors.New("'all' identifier can not be used in log mask level assignments")
```
kjacque, line 145: Spelling nit but I think "cannot" was correct here.

kjacque (approved):
LGTM. My comment needn't be addressed now unless you have other changes to make.

tanabarr:
clang-format warning doesn't make sense so please ignore

tanabarr:
@daos-stack/daos-gatekeeper can this please be landed?

PR #13007, opened 2023-09-05
DAOS-14110 il: create and do io against an EQ in the IL

mchaarawi:
- Use an EQ in the IL for dfs read and write for better performance and support of fork mode in fio()
- add a new env variable D_IL_MAX_EQ to make the number of EQs / network contexts configurable. 64 is the default and the maximum allowed. Asking for more than that will not be an error though and the library will print a warning an set that to 64.
- in case of multiple threads, each thread will create a new EQ till the max are created, then other threads will just round robin use of the EQs that are created.
- When fork is called, the main thread eq is replaced with a new one from the child process (with at_fork child handler) to prevent sharing of any internal network context resources.
- Update the IL VM build test to use less EQs than the default to prevent OOM since the VMs are constrained on memory.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfuse/il/int_posix.c
+++ b/src/client/dfuse/il/int_posix.c
@@ -354,7 +354,6 @@ ioil_fini(void)
 	int               rc;
 	pid_t             tid = syscall(SYS_gettid);
 
-	__real_fprintf(stderr, "[IOIL] Calling fini on tid: %d\n", tid);
 	if (tid != ioil_iog.iog_init_tid) {
 		DFUSE_TRA_INFO(&ioil_iog, "Ignoring destructor from alternate thread");
 		return;
@@ -394,7 +393,8 @@ ioil_fini(void)
 		for (i = 0; i < ioil_iog.iog_eq_count; i++)
 			daos_eq_destroy(ioil_iog.iog_eqs[i], 0);
 		/** destroy main thread eq */
-		daos_eq_destroy(ioil_iog.iog_main_eqh, 0);
+		if (daos_handle_is_valid(ioil_iog.iog_main_eqh))
+			daos_eq_destroy(ioil_iog.iog_main_eqh, 0);
 		daos_fini();
 	}
 	ioil_iog.iog_daos_init = false;
--- a/src/client/dfuse/il/int_posix.c
+++ b/src/client/dfuse/il/int_posix.c
@@ -737,6 +737,8 @@ child_hdlr(void)
 {
 	int rc;
 
+	daos_dti_reset();
+
 	ioil_eqh.cookie = 0;
 	rc = daos_eq_create(&ioil_eqh);
 	if (rc)
--- a/src/client/dfuse/il/int_posix.c
+++ b/src/client/dfuse/il/int_posix.c
@@ -327,6 +327,7 @@ ioil_init(void)
 	if (rc)
 		return;
 
+	d_getenv_bool("D_IL_USE_EQ", &use_eq);
 	ioil_iog.iog_initialized = true;
 }
 
--- a/src/client/dfuse/il/int_posix.c
+++ b/src/client/dfuse/il/int_posix.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2017-2022 Intel Corporation.
+ * (C) Copyright 2017-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -33,6 +33,8 @@
 
 FOREACH_INTERCEPT(IOIL_FORWARD_DECL)
 
+__thread daos_handle_t ioil_eqh = {.cookie = 0};
+
 struct ioil_pool {
 	daos_handle_t	iop_poh;
 	uuid_t		iop_uuid;
@@ -49,7 +51,6 @@ struct ioil_global {
 	bool		iog_daos_init;
 
 	bool		iog_show_summary;	/**< Should a summary be shown at teardown */
-
 	unsigned	iog_report_count;	/**< Number of operations that should be logged */
 
 	ATOMIC uint64_t iog_file_count;  /**< Number of file opens intercepted */
@@ -377,8 +378,10 @@ ioil_fini(void)
 			ioil_shrink_pool(pool);
 	}
 
-	if (ioil_iog.iog_daos_init)
+	if (ioil_iog.iog_daos_init) {
+		daos_eq_destroy(ioil_eqh, 0);
 		daos_fini();
+	}
 	ioil_iog.iog_daos_init = false;
 	daos_debug_fini();
 }
@@ -729,6 +732,17 @@ out:
 	return rcb;
 }
 
+static void
+child_hdlr(void)
+{
+	int rc;
+
+	ioil_eqh.cookie = 0;
+	rc = daos_eq_create(&ioil_eqh);
+	if (rc)
+		DFUSE_LOG_WARNING("daos_eq_create() failed: %d (%s)", rc, strerror(rc));
+}
+
 /* Returns true on success */
 static bool
 check_ioctl_on_open(int fd, struct fd_entry *entry, int flags)
@@ -764,9 +778,20 @@ check_ioctl_on_open(int fd, struct fd_entry *entry, int flags)
 	rc = pthread_mutex_lock(&ioil_iog.iog_lock);
 	D_ASSERT(rc == 0);
 
-	if (!ioil_iog.iog_daos_init)
+	if (!ioil_iog.iog_daos_init) {
 		if (!call_daos_init(fd))
 			goto err;
+
+		rc = daos_eq_create(&ioil_eqh);
+		if (rc) {
+			DFUSE_LOG_WARNING("daos_eq_create() failed: "DF_RC"\n", DP_RC(rc));
+			D_GOTO(err, rc = daos_der2errno(rc));
+		}
+
+		/** register a child fork handler to create a new EQ */
+		rc = pthread_atfork(NULL, NULL, &child_hdlr);
+		D_ASSERT(rc == 0);
+	}
 
 	d_list_for_each_entry(pool, &ioil_iog.iog_pools_head, iop_pools) {
 		if (uuid_compare(pool->iop_uuid, il_reply.fir_pool) != 0)
--- a/src/client/dfuse/il/int_posix.c
+++ b/src/client/dfuse/il/int_posix.c
@@ -52,6 +52,7 @@ struct ioil_global {
 	daos_handle_t	iog_main_eqh;
 	daos_handle_t	iog_eqs[IOIL_MAX_EQ];
 	uint16_t	iog_eq_count;
+	uint16_t	iog_eq_idx;
 	pid_t           iog_init_tid;
 	bool		iog_initialized;
 	bool		iog_no_daos;
@@ -412,21 +413,22 @@ ioil_get_eqh(daos_handle_t *eqh)
 	}
 
 	rc = pthread_mutex_lock(&ioil_iog.iog_lock);
-
+	/** create a new EQ if the EQ pool is not full; otherwise round robin EQ use from pool */
 	if (ioil_iog.iog_eq_count >= IOIL_MAX_EQ) {
-		pthread_mutex_unlock(&ioil_iog.iog_lock);
-		return -1;
-	}
+		*eqh = ioil_iog.iog_eqs[ioil_iog.iog_eq_idx ++];
+		if (ioil_iog.iog_eq_idx == IOIL_MAX_EQ)
+			ioil_iog.iog_eq_idx = 0;
+	} else {
+		rc = daos_eq_create(&ioil_eqh);
+		if (rc) {
+			pthread_mutex_unlock(&ioil_iog.iog_lock); 
+			return -1;
+		}
 
-	rc = daos_eq_create(&ioil_eqh);
-	if (rc) {
-		pthread_mutex_unlock(&ioil_iog.iog_lock); 
-		return -1;
+		*eqh = ioil_eqh;
+		ioil_iog.iog_eqs[ioil_iog.iog_eq_count] = ioil_eqh;
+		ioil_iog.iog_eq_count ++;
 	}
-
-	*eqh = ioil_eqh;
-	ioil_iog.iog_eqs[ioil_iog.iog_eq_count] = ioil_eqh;
-	ioil_iog.iog_eq_count ++;
 	pthread_mutex_unlock(&ioil_iog.iog_lock);
 	return 0;
 }
--- a/src/client/dfuse/il/int_posix.c
+++ b/src/client/dfuse/il/int_posix.c
@@ -33,7 +33,11 @@
 
 FOREACH_INTERCEPT(IOIL_FORWARD_DECL)
 
-__thread daos_handle_t ioil_eqh = {.cookie = 0};
+static __thread daos_handle_t ioil_eqh;
+
+#define IOIL_MAX_EQ 64
+
+bool use_eq = true;
 
 struct ioil_pool {
 	daos_handle_t	iop_poh;
@@ -45,6 +49,9 @@ struct ioil_pool {
 struct ioil_global {
 	pthread_mutex_t	iog_lock;
 	d_list_t	iog_pools_head;
+	daos_handle_t	iog_main_eqh;
+	daos_handle_t	iog_eqs[IOIL_MAX_EQ];
+	uint16_t	iog_eq_count;
 	pid_t           iog_init_tid;
 	bool		iog_initialized;
 	bool		iog_no_daos;
@@ -346,6 +353,7 @@ ioil_fini(void)
 	int               rc;
 	pid_t             tid = syscall(SYS_gettid);
 
+	__real_fprintf(stderr, "[IOIL] Calling fini on tid: %d\n", tid);
 	if (tid != ioil_iog.iog_init_tid) {
 		DFUSE_TRA_INFO(&ioil_iog, "Ignoring destructor from alternate thread");
 		return;
@@ -379,11 +387,47 @@ ioil_fini(void)
 	}
 
 	if (ioil_iog.iog_daos_init) {
-		daos_eq_destroy(ioil_eqh, 0);
+		int i;
+
+		/** destroy EQs created by threads */
+		for (i = 0; i < ioil_iog.iog_eq_count; i++)
+			daos_eq_destroy(ioil_iog.iog_eqs[i], 0);
+		/** destroy main thread eq */
+		daos_eq_destroy(ioil_iog.iog_main_eqh, 0);
 		daos_fini();
 	}
 	ioil_iog.iog_daos_init = false;
 	daos_debug_fini();
+}
+
+int
+ioil_get_eqh(daos_handle_t *eqh)
+{
+	int rc;
+
+	if (daos_handle_is_valid(ioil_eqh)) {
+		*eqh = ioil_eqh;
+		return 0;
+	}
+
+	rc = pthread_mutex_lock(&ioil_iog.iog_lock);
+
+	if (ioil_iog.iog_eq_count >= IOIL_MAX_EQ) {
+		pthread_mutex_unlock(&ioil_iog.iog_lock);
+		return -1;
+	}
+
+	rc = daos_eq_create(&ioil_eqh);
+	if (rc) {
+		pthread_mutex_unlock(&ioil_iog.iog_lock); 
+		return -1;
+	}
+
+	*eqh = ioil_eqh;
+	ioil_iog.iog_eqs[ioil_iog.iog_eq_count] = ioil_eqh;
+	ioil_iog.iog_eq_count ++;
+	pthread_mutex_unlock(&ioil_iog.iog_lock);
+	return 0;
 }
 
 /* Get the object handle for the file itself */
@@ -738,11 +782,12 @@ child_hdlr(void)
 	int rc;
 
 	daos_dti_reset();
-
-	ioil_eqh.cookie = 0;
+	ioil_eqh = DAOS_HDL_INVAL;
 	rc = daos_eq_create(&ioil_eqh);
 	if (rc)
-		DFUSE_LOG_WARNING("daos_eq_create() failed: %d (%s)", rc, strerror(rc));
+		DFUSE_LOG_WARNING("daos_eq_create() failed: "DF_RC, DP_RC(rc));
+	else
+		ioil_iog.iog_main_eqh = ioil_eqh;
 }
 
 /* Returns true on success */
@@ -786,9 +831,10 @@ check_ioctl_on_open(int fd, struct fd_entry *entry, int flags)
 
 		rc = daos_eq_create(&ioil_eqh);
 		if (rc) {
-			DFUSE_LOG_WARNING("daos_eq_create() failed: "DF_RC"\n", DP_RC(rc));
+			DFUSE_LOG_WARNING("daos_eq_create() failed: "DF_RC, DP_RC(rc));
 			D_GOTO(err, rc = daos_der2errno(rc));
 		}
+		ioil_iog.iog_main_eqh = ioil_eqh;
 
 		/** register a child fork handler to create a new EQ */
 		rc = pthread_atfork(NULL, NULL, &child_hdlr);
--- a/src/client/dfuse/il/int_posix.c
+++ b/src/client/dfuse/il/int_posix.c
@@ -780,15 +780,8 @@ out:
 static void
 child_hdlr(void)
 {
-	int rc;
-
 	daos_dti_reset();
 	ioil_eqh = DAOS_HDL_INVAL;
-	rc = daos_eq_create(&ioil_eqh);
-	if (rc)
-		DFUSE_LOG_WARNING("daos_eq_create() failed: "DF_RC, DP_RC(rc));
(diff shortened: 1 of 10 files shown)


src/client/dfuse/il/int_read.c
```diff
@@ -15,22 +15,54 @@
...
 
 	DFUSE_TRA_DEBUG(entry->fd_dfsoh, "%#zx-%#zx", position, position + len - 1);
 
 	sgl.sg_nr = 1;
 	d_iov_set(&iov, (void *)buff, len);
 	sgl.sg_iovs = &iov;
-	rc = dfs_read(entry->fd_cont->ioc_dfs, entry->fd_dfsoh, &sgl, position, &read_size, NULL);
+
+	if (daos_handle_is_inval(ioil_eqh))
+		daos_eq_create(&ioil_eqh);
```
wiliamhuang, line 32: Need to handle errors in case daos_eq_create() fails.

src/client/dfuse/il/int_posix.c
```diff
@@ -377,8 +378,10 @@ ioil_fini(void)
 			ioil_shrink_pool(pool);
 	}
 
-	if (ioil_iog.iog_daos_init)
+	if (ioil_iog.iog_daos_init) {
+		daos_eq_destroy(ioil_eqh, 0);
```
wiliamhuang, line 382: This may be problematic. The destructor function is only executed by the main thread. The event queue created by other child threads are leaked. We need to maintain a list of "ioil_eqh" allocated by all threads and destroy all of them here.

src/client/dfuse/il/int_read.c
mchaarawi: not really, because if it fails, we fall back to the original way of using the internal one.

src/client/dfuse/il/int_write.c
```diff
@@ -15,22 +15,53 @@
...
 
 	DFUSE_TRA_DEBUG(entry->fd_dfsoh, "%#zx-%#zx", position, position + len - 1);
 
 	sgl.sg_nr = 1;
 	d_iov_set(&iov, (void *)buff, len);
 	sgl.sg_iovs = &iov;
 
-	rc = dfs_write(entry->fd_cont->ioc_dfs, entry->fd_dfsoh, &sgl, position, NULL);
+	if (daos_handle_is_inval(ioil_eqh))
+		daos_eq_create(&ioil_eqh);
```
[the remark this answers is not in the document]
mchaarawi: same

src/client/dfuse/il/int_posix.c
mchaarawi: yea im not sure how to handle this. i tried doing atexit handlers, but those don't get executed in case of fork().
in case of threads, yea having a list of eqs might work.

wiliamhuang: For fork(), we could intercept fork(). We can tell we are in parent or child process from the pid returned by fork(). We can zero ioil_eqh in child process. It will be re-created when necessary. destructor function will be called by both parent and child processes.

mchaarawi: hmm i think the destructor function is only called on the parent as i can tell.

wiliamhuang: I was not sure either, so I wrote a mini code to verify. It was confirmed that both parent and child process call the destructor function.

Example code,

#include <stdio.h>
#include <sys/types.h>
#include <unistd.h>

int tid;

static __attribute__((destructor)) void fini()
{
        printf("In fini(). tid = %d\n", tid);
}

void forkexample()
{
        tid = fork();
        // child process because return value zero
        if (tid == 0)
                printf("Hello from Child!\n");

        // parent process because return value non-zero.
        else
                printf("Hello from Parent!\n");
}

int main()
{
        forkexample();
        return 0;
}

mchaarawi: try to call exec in the child

mchaarawi: i have added a print in ioil_fini() and run fio with 32 processes, and i only see one call.
i assume because of this:
"Upon a successful call to one of the [exec(3)](https://man7.org/linux/man-pages/man3/exec.3.html) functions, all registrations are removed."

wiliamhuang: yes. calling exec() would be different. It starts a fresh process. "ioil_eqh" created by the parent still will be destroyed.

wiliamhuang: @mchaarawi I checked fio code again. With your configure file, fio call pthread_create() to spawn new threads. It does not call fork/exec. ioil_fini() is expected to be called only one time since there is only one process in fio.

mchaarawi: im using the non-thread option in fio which does use fork. let's discuss offline

ashleypittman: Have a look at L349-351 above, there are lots of cases where this destructor is being called at least from different threads, I'm not sure if it's called from multiple threads in the same process however.  Without the thread check we were seeing segfaults however.

src/client/dfuse/il/int_posix.c
```diff
@@ -764,10 +780,21 @@ check_ioctl_on_open(int fd, struct fd_entry *entry, int flags)
...
 	D_ASSERT(rc == 0);
 
-	if (!ioil_iog.iog_daos_init)
+	if (!ioil_iog.iog_daos_init) {
 		if (!call_daos_init(fd))
 			goto err;
 
+		rc = daos_eq_create(&ioil_eqh);
+		if (rc) {
+			DFUSE_LOG_WARNING("daos_eq_create() failed: "DF_RC"\n", DP_RC(rc));
```
ashleypittman, line 789: As some history here dfuse predates daos so initially the DFUSE_LOG macros were wrappers around the gurt logging and dfuse has always added a \n in the macro automatically.  When daos came along it didn't do this so D_LOG* etc needed the \n.  At some point several years ago however gurt was updated to add the newline in the c function if it was missing so for the D_LOG macros it's optional.

Including rc at the end of the line has always been in the coding standards but we didn't have DF_RC initially and it's use has been sporadic and inconsistent.

The addition of DL_ERROR and friends is an attempt to get on top of the chaos and it handles rc automatically, putting the DF_RC logic in for you, but that relies on the macro adding text to the end of the error message - something it can't do if there's already a new-line there, hence DL_ERROR needs to not have a new-line even though most developers are used to adding one, an obvious cause of annoying issues, hence the check that macros that should not be provided with a new-line do not provide one.

The end result of all this is use DL_ERROR() here, or DHL() if there's a handle available.  There is a handle for the entry here but that's not actually relevant to any failure so if a handle was used here it would be `&ioil_iog`
```suggestion
			DL_ERROR(rc, "daos_eq_create() failed");
```

ashleypittman, line 789:
```suggestion
			DHL_ERROR(&iol_iog, rc, "daos_eq_create() failed);
```

src/client/dfuse/il/int_posix.c
```diff
@@ -729,6 +732,19 @@ call_daos_init(int fd)
...
 }
 
+static void
+child_hdlr(void)
+{
+	int rc;
+
+	daos_dti_reset();
+
+	ioil_eqh.cookie = 0;
```
ashleypittman, line 742:
```suggestion
	ioil_eqh = DAOS_HDL_INVAL;
```

src/client/dfuse/il/int_read.c
ashleypittman: We do however want to avoid the case where it's attempted, and fails for every I/O.

I'm not sure why this needs to be called for every read and write however, can't this be done in the fork handler or are you thinking of the case where a program calls fork and exec but you don't want to do anything before the exec?  In which so I think a boolean flag or lock indicates that this call needs to be made was appropriate.  Alternatively, you could do in on open, this might miss the case where files where open during the fork and then written to before another file was opened but that's probably an acceptable cost.

src/client/dfuse/il/int_read.c
```diff
@@ -15,22 +15,54 @@
...
+
+	rc = daos_event_init(&ev, ioil_eqh, NULL);
+	if (rc) {
+		DFUSE_TRA_ERROR(entry->fd_dfsoh, "daos_event_init() failed: %d (%s)",
+				rc, strerror(rc));
```
ashleypittman, line 37: I've not started the migration yet but soon all these calls will be changing to this format:
```suggestion
		DHS_ERROR(entry->fd_dfsoh, rc, "daos_event_init() failed");
```

ashleypittman (commented):
This will need more work so it doesn't leak eqs for every thread, even catching this on exit isn't good enough as there will be long-running processes that start and stop helper threads so some cleanup on thread exit or sharing of eqs will be needed.

mchaarawi:
> This will need more work so it doesn't leak eqs for every thread, even catching this on exit isn't good enough as there will be long-running processes that start and stop helper threads so some cleanup on thread exit or sharing of eqs will be needed.

the only thing that we can do here is to cleanup the eqs on exit for all threads. I will add a limit on how many eqs one can create. this is still going to be better than what we have today with the original way of just using sync calls on the internal eq.
if you would like to extend this to be more dynamic, please do so in a separate PR

src/client/dfuse/il/int_posix.c
mchaarawi: yes i saw those. in my testing, the destructor is called only once. i verified this with Lei and it's because fio calls exit on forked processes which does not call this destructor.

src/client/dfuse/il/int_posix.c
mchaarawi: just following what has been done here. im not interested in changing what is here to the new macros atm

src/client/dfuse/il/int_read.c
mchaarawi: again, i don't care for that in this pr

src/client/dfuse/il/int_read.c
ashleypittman: Ah, I'd missed that this was on a thread local variable.

src/client/dfuse/il/int_posix.c
ashleypittman: Then you just need to remove the \n from where you're getting these warnings.

mchaarawi: ok will do that.

src/client/dfuse/il/int_posix.c
wiliamhuang: Right. We need to intercept "_exit()" (not "exit()") and call daos_eq_destroy() for all allocated eqs there too. Destructor function is still executed for processes calling "exit()" in my test.

src/client/dfuse/il/int_posix.c
mchaarawi: there are several other occurences in this file and others where \n is included fwiw..

ashleypittman: They will be calls to D_ERROR or D_INFO which do not add the newline at the macro level.

mchaarawi: sigh.. how is one to remember all this

ashleypittman: This is why I'm trying to ratify it all and add tools to identify mis-use.  See my first comment.

wiliamhuang:
Need to intercept "_exit()" and calls destructor function ioil_fini() there.

mchaarawi:
> Need to intercept "_exit()" and calls destructor function ioil_fini() there.

im not going to do that in this PR. as we discussed, we do not support fork and this is just to get fio to work. Leaking eq in this case is not the only thing we would be doing anyway even without this pr if one uses fork with _exit().

wiliamhuang:
> > Need to intercept "_exit()" and calls destructor function ioil_fini() there.
> 
> im not going to do that in this PR. as we discussed, we do not support fork and this is just to get fio to work. Leaking eq in this case is not the only thing we would be doing anyway even without this pr if one uses fork with _exit().

ok. :)

wiliamhuang:
Compiling daos with IL failed. This is a very complicated test which involves fork() and execve() (and its variants) etc. Current patch does not destroy allocated eq before calling execve(). Leaks in many processes might cause issues.

ashleypittman:
> Compiling daos with IL failed. This is a very complicated test which involves fork() and execve() (and its variants) etc. Current patch does not destroy allocated eq before calling execve(). Leaks in many processes might cause issues.

Looks like it got killed via the oom killer again.

mchaarawi:
> Compiling daos with IL failed. This is a very complicated test which involves fork() and execve() (and its variants) etc. Current patch does not destroy allocated eq before calling execve(). Leaks in many processes might cause issues.

@wiliamhuang fork does not work without this patch. you have your reproducer to prove that. so how would it work before with this test if this is the issue?

mchaarawi:
> Compiling daos with IL failed. This is a very complicated test which involves fork() and execve() (and its variants) etc. Current patch does not destroy allocated eq before calling execve(). Leaks in many processes might cause issues.

so i have realized after trying locally and a few runs in CI that this is not a memory leak, but just using too much memory at once. even with this environment variable to limit the number of context (or even set it to use  the original way), the build test ignores that env variable in different parts of the build. so we might have a problem here with this on vm.

ashleypittman:
> > Compiling daos with IL failed. This is a very complicated test which involves fork() and execve() (and its variants) etc. Current patch does not destroy allocated eq before calling execve(). Leaks in many processes might cause issues.
> 
> so i have realized after trying locally and a few runs in CI that this is not a memory leak, but just using too much memory at once

Yes, that's my assumption.

>  even with this environment variable to limit the number of context (or even set it to use the original way), the build test ignores that env variable in different parts of the build. so we might have a problem here with this particular test on vm.

Take a look at https://github.com/daos-stack/daos/blob/d42b2ab2f088be6296910b2527de3f80b9de39ed/SConstruct#L364-L366

What happens is that scons strips the environment so it's 'clean', then we have to tell it to pass on any that we want to keep.  To get this to work you need to change the code above to add the new env to the list, and then (this is the annoying part) change the test so that it does a `git checkout mscchara/il_eq` command, then have it run the build which should at that point have the env set, then when you're happy this can pass you need to modify this PR again to remove the checkout from the list of commands and then it can land.  The test checks out master (or rather it doesn't, that's the default branch) so once this is landed then it'll all work without the extra checkout step.

mchaarawi:
> > > Compiling daos with IL failed. This is a very complicated test which involves fork() and execve() (and its variants) etc. Current patch does not destroy allocated eq before calling execve(). Leaks in many processes might cause issues.
> > 
> > 
> > so i have realized after trying locally and a few runs in CI that this is not a memory leak, but just using too much memory at once
> 
> Yes, that's my assumption.
> 
> > even with this environment variable to limit the number of context (or even set it to use the original way), the build test ignores that env variable in different parts of the build. so we might have a problem here with this particular test on vm.
> 
> Take a look at
> 
> https://github.com/daos-stack/daos/blob/d42b2ab2f088be6296910b2527de3f80b9de39ed/SConstruct#L364-L366
> 
> What happens is that scons strips the environment so it's 'clean', then we have to tell it to pass on any that we want to keep. To get this to work you need to change the code above to add the new env to the list, and then (this is the annoying part) change the test so that it does a `git checkout mscchara/il_eq` command, then have it run the build which should at that point have the env set, then when you're happy this can pass you need to modify this PR again to remove the checkout from the list of commands and then it can land. The test checks out master (or rather it doesn't, that's the default branch) so once this is landed then it'll all work without the extra checkout step.

thanks for that!
with this change and setting eq count to 6, the test passed here:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-13007/20/pipeline/

i will revert the checkout now and not run this test for PR landing.

src/client/dfuse/il/int_read.c
```diff
@@ -15,20 +15,59 @@
...
 
 	DFUSE_TRA_DEBUG(entry->fd_dfsoh, "%#zx-%#zx", position, position + len - 1);
 
 	sgl.sg_nr = 1;
 	d_iov_set(&iov, (void *)buff, len);
 	sgl.sg_iovs = &iov;
-	rc = dfs_read(entry->fd_cont->ioc_dfs, entry->fd_dfsoh, &sgl, position, &read_size, NULL);
+
+	rc = ioil_get_eqh(&eqh);
+	if (rc == 0) {
```
ashleypittman, line 32: There are several exit paths here where *errcode does not get set, I think you want a label at L69 and then replace L39-40, L48-49 and L57-58 with D_GOTOS.   The rc values at L64 and L66 will be not be propagated with this code as it is now.

src/client/dfuse/il/int_write.c
```diff
@@ -15,20 +15,56 @@
...
 
 	DFUSE_TRA_DEBUG(entry->fd_dfsoh, "%#zx-%#zx", position, position + len - 1);
 
 	sgl.sg_nr = 1;
 	d_iov_set(&iov, (void *)buff, len);
 	sgl.sg_iovs = &iov;
 
-	rc = dfs_write(entry->fd_cont->ioc_dfs, entry->fd_dfsoh, &sgl, position, NULL);
+	rc = ioil_get_eqh(&eqh);
+	if (rc == 0) {
```
ashleypittman, line 31: Same issue as read I think.

src/tests/ftest/dfuse/daos_build.py
```diff
@@ -135,9 +135,11 @@ def run_build_test(self, cache_mode, intercept=False, run_on_vms=False):
 
         # Note that run_on_vms does not tell ftest where to run, this should be set according to
         # the test tags so the test can run with appropriate settings.
+        remote_env = {}
         if run_on_vms:
             dfuse_namespace = dfuse_namespace = "/run/dfuse_vm/*"
             build_jobs = 6 * 2
+            remote_env['D_IL_MAX_EQ'] = '6'
```
ashleypittman, line 142: Could this not be set at L199-203 along with the rest of the intercept options.

src/client/dfuse/il/int_read.c
mchaarawi: ah do you mean when the call actually succeeds that for line 64 and 66 we do not set *errcode?

src/tests/ftest/dfuse/daos_build.py
mchaarawi: i thought we want this only for vms and IL testing; but if you want to do this for all IL testing with this, then im fine to do that.

src/client/dfuse/il/int_read.c
mchaarawi: nevermind i see it

src/tests/ftest/dfuse/daos_build.py
ashleypittman: Ah, I see the difference.  In practice they're the same thing, threre's only one variant of this test that's run on VMs and that's the IL variant.

src/client/dfuse/il/int_posix.c
```diff
@@ -377,12 +398,55 @@ ioil_fini(void)
...
+	}
+
+	/** No EQ support requested */
+	if (ioil_iog.iog_eq_count_max == 0)
+		return -1;
+
+	rc = pthread_mutex_lock(&ioil_iog.iog_lock);
+	/** create a new EQ if the EQ pool is not full; otherwise round robin EQ use from pool */
+	if (ioil_iog.iog_eq_count >= ioil_iog.iog_eq_count_max) {
+		ioil_eqh = ioil_iog.iog_eqs[ioil_iog.iog_eq_idx ++];
```
wiliamhuang, line 433: Shall we move Line 433 "ioil_eqh = ioil_iog.iog_eqs[ioil_iog.iog_eq_idx ++];" after line 435?

mchaarawi: can you explain why that is needed?

wiliamhuang: Line 433 uses the value "ioil_iog.iog_eq_idx" before the initialization in line 435. The code should be fine since static variables "static struct ioil_global ioil_iog" are set zero anyway. If we rely on the default zero value of ioil_iog.iog_eq_idx, there is no need of line 435 then.

mchaarawi: right, that is my assumption. static global variables are already initialized to 0. So this is not needed IMO

wiliamhuang: I see. You can remove line 435 either in this PR or in future PR.

wiliamhuang (approved):

src/client/dfuse/il/int_posix.c
```diff
@@ -729,6 +793,20 @@ call_daos_init(int fd)
...
+{
+	int rc;
+
+	daos_dti_reset();
+	ioil_eqh = DAOS_HDL_INVAL;
+	rc = daos_eq_create(&ioil_eqh);
+	if (rc)
+		DFUSE_LOG_WARNING("daos_eq_create() failed: "DF_RC, DP_RC(rc));
+	else
+		ioil_iog.iog_main_eqh = ioil_eqh;
```
jolivier23, line 807: does fio actually do I/O in the forked child without subsequent exec?  Seems like a lot of stuff can go wrong in that scenario.   Like the log message above.  Do we need to reinitialize the logging?

src/tests/ftest/dfuse/daos_build.py
mchaarawi: yea but i was thinking what if we add more variants on HW later? probably want those to run with default setting for that one.

src/client/dfuse/il/int_posix.c
mchaarawi: yes. actually those forked processes don't get to finalize, so the stat / logging don't even show up / get printed.
anyway for that scenario, my intent is to just get FIO to work and perform well, and not worry about stats or leaked contexts in this case. we really cannot / should not claim we support that scenario today.

mchaarawi: As @wiliamhuang mentioned, we can do more  to have this fio case work better by intercepting _exit() and do the finalize and printing of stats, etc.
i don't want to do that in this PR though and we can create a ticket and fix that in a future PR.

src/client/dfuse/il/int_posix.c
wiliamhuang: Yes. You are right.

src/tests/ftest/dfuse/daos_build.py
mchaarawi: i would suggest to leave it like this for now.

mchaarawi:
@ashleypittman any more feedback?

src/client/dfuse/il/int_posix.c
jolivier23: Have you tried with D_LOG_FILE?  I suspect it has issues with current patch

mchaarawi: no the parent doesn't exit. the children just do their IO job and _exit()
we don't need to re-init logging because logging for the children don't get printed anyway in the FIO case.

i will reiterate again that im not interested in implementing support for fork() in this PR. this PR is at least better than what we end up with today (a segfault in libfabric).

what is the issue with D_LOG_FILE?

mchaarawi: sorry i confused the logging that you are talking about with the stats of the IL. 
but im not getting any logging errors in my case and D_LOG_FILE works fine.

mchaarawi: i just tried a quick run setting D_LOG_FILE and debug mask and it's fine.
as for what to recommend for the user, we should still not recommend fork as beyond this FIO use case, because any usage of dfs metadata calls or any daos calls with NULL event, we get back to the libfabric segfault. so this is not something we can support today anyway, but just a workaround for FIO testing.

jolivier23: Ok, it does just seem to work which is surprising

jeffolivier@jeffolivier1:~/community/daos$ cat foo.c
#include <unistd.h>
#include <gurt/debug.h>

int main(int argc, char **argv)
{
	int rc = d_log_init_adv("DAOS", "/tmp/foo.log", 0, DLOG_INFO, DLOG_CRIT, NULL);
	int pid = getpid();
	int child_pid = fork();
	if (child_pid == 0) {
		D_INFO("Log from child\n");
	}
	D_INFO("Log from both\n");
	d_log_fini();
}

effolivier@jeffolivier1:~/community/daos$ cat /tmp/foo.log 
09/11-21:52:48.29 jeffolivier1 INFO foo.c:15 main() Log from both
09/11-21:52:48.29 jeffolivier1 INFO foo.c:12 main() Log from child
09/11-21:52:48.29 jeffolivier1 INFO foo.c:15 main() Log from both

mchaarawi:
ping!

jolivier23 (approved):

wangshilong (approved):
Looks a good improvement, I got question from people that fio could not work for IL before.
@mchaarawi  Do you have performance numbers for this? eg compared to fio over dfs directly?

mchaarawi:
> Looks a good improvement, I got question from people that fio could not work for IL before. @mchaarawi Do you have performance numbers for this? eg compared to fio over dfs directly?

i have not done a comprehensive perf study with this, but when using the same arguments with psync (which support only a single io_dpeth in flight) and dfs engine, using a 1mib block size and 32-64 threads or procs, performance is about the same now. before this, ioil perf was about 5-10x worse.

PR #13011, opened 2023-09-06
DAOS-14328 cart: Do not print error when exceeding context limit

frostedcmos:
- Do not print an internal layer error when exceeding the context limit, as daos client layer will try to allocate as many contexts as it can in some usage models. Printing an error is not user friendly in such situations.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/cart/crt_hg.c
+++ b/src/cart/crt_hg.c
@@ -617,8 +617,9 @@ crt_provider_get_ctx_idx(bool primary, int provider)
 		}
 	}
 
-	D_ERROR("ctx_num %d, will exceed CRT_SRV_CONTEXT_NUM (%d) if create more context.\n",
-		prov_data->cpg_ctx_num, CRT_SRV_CONTEXT_NUM);
+	D_DEBUG(DB_ALL, "provider:%d allowed context limit = %d exceeded\n",
+		provider, CRT_SRV_CONTEXT_NUM);
+
 	return -1;
 }
 


ashleypittman:
This feels like it should be an INFO so it's emitted by default.  The code will continue to work afterwards but with a work-around, and there are settings that could be applied to mitigate some of the workaround cost?

frostedcmos:
> This feels like it should be an INFO so it's emitted by default. The code will continue to work afterwards but with a work-around, and there are settings that could be applied to mitigate some of the workaround cost?

Customers were confused to see these messages during a normal run, which is why i moved it to the debug. I can change it to info if that is preferred. @mchaarawi ?

I am not really sure what costs you are referring to or which settings.

frostedcmos:
> > > This feels like it should be an INFO so it's emitted by default. The code will continue to work afterwards but with a work-around, and there are settings that could be applied to mitigate some of the workaround cost?
> > 
> > 
> > Customers were confused to see these messages during a normal run, which is why i moved it to the debug. I can change it to info if that is preferred. @mchaarawi ?
> > I am not really sure what costs you are referring to or which settings.
> 
> Oh, I thought `CRT_SRV_CONTEXT_NUM` was an env var, is it just a #define? The cost is I/O being on a shared context rather than a dedicated one isn't it so worse data locality?

It's the define, set to 64. I am not sure what we can do on our end wrt costs of using the same shared context; this is a daos client library decision to use one of the contexts as a shared one (as opposed to round-robbing among available context for example).

mchaarawi:
> > > > This feels like it should be an INFO so it's emitted by default. The code will continue to work afterwards but with a work-around, and there are settings that could be applied to mitigate some of the workaround cost?
> > > 
> > > 
> > > Customers were confused to see these messages during a normal run, which is why i moved it to the debug. I can change it to info if that is preferred. @mchaarawi ?
> > > I am not really sure what costs you are referring to or which settings.
> > 
> > 
> > Oh, I thought `CRT_SRV_CONTEXT_NUM` was an env var, is it just a #define? The cost is I/O being on a shared context rather than a dedicated one isn't it so worse data locality?
> 
> It's the define, set to 64. I am not sure what we can do on our end wrt costs of using the same shared context; this is a daos client library decision to use one of the contexts as a shared one (as opposed to round-robbing among available context for example).

i think debug for this message is fine. the name is a little misleading though for this.
is there any reason why this can't be configurable with an env variable? why is the max 64 for example? why not 128?

jgmoore-or (approved):

soumagne (approved):

mchaarawi (approved):

PR #13018, opened 2023-09-07
DAOS-14332 control: Use gRPC metadata for interop

mjmac:
The interoperability checking code relies on getting the
peer component from the peer certificate. When running
in insecure mode, the peer component information is
unavailable, and the interoperability check defaults
to the most stringent requirements.

This patch adds new component/version headers to the
grpc client request to allow the server to perform
interoperability checks without the peer certificate.

Features: control

Required-githooks: true
Change-Id: I1d9a6d171ecadded3bf041a36e2a156881d93f57

### DIFF
--- a/src/control/lib/control/interceptors.go
+++ b/src/control/lib/control/interceptors.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2021 Intel Corporation.
+// (C) Copyright 2020-2023 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -78,11 +78,15 @@ func unaryErrorInterceptor() grpc.UnaryClientInterceptor {
 
 // unaryVersionedComponentInterceptor appends the component name and version to the
 // outgoing request headers.
-func unaryVersionedComponentInterceptor() grpc.UnaryClientInterceptor {
+func unaryVersionedComponentInterceptor(comp build.Component) grpc.UnaryClientInterceptor {
 	return func(parent context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-		comp, err := security.MethodToComponent(method)
-		if err != nil {
-			return errors.Wrap(err, "unable to determine component from method")
+		// NB: The caller should specify its component, but as a fallback, we
+		// can make a decent guess about the calling component based on the method.
+		if comp == build.ComponentAny {
+			var err error
+			if comp, err = security.MethodToComponent(method); err != nil {
+				return errors.Wrap(err, "unable to determine component from method")
+			}
 		}
 		ctx := metadata.AppendToOutgoingContext(parent,
 			proto.DaosComponentHeader, comp.String(),
--- a/src/control/lib/control/interceptors.go
+++ b/src/control/lib/control/interceptors.go
@@ -12,8 +12,10 @@ import (
 
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
+	"github.com/daos-stack/daos/src/control/build"
 	"github.com/daos-stack/daos/src/control/common/proto"
 	"github.com/daos-stack/daos/src/control/security"
 )
@@ -59,8 +61,8 @@ func streamErrorInterceptor() grpc.DialOption {
 }
 
 // unaryErrorInterceptor calls the specified unary RPC and returns any unwrapped errors.
-func unaryErrorInterceptor() grpc.DialOption {
-	return grpc.WithUnaryInterceptor(func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+func unaryErrorInterceptor() grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 		err := invoker(ctx, method, req, reply, cc, opts...)
 		if err != nil {
 			st := status.Convert(err)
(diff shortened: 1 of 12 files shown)


kjacque (changes requested):
Overall approach looks good. This should resolve various points of confusion about interop in insecure mode.

src/control/server/interceptors_test.go
```diff
@@ -141,10 +144,79 @@ func TestServer_checkVersion(t *testing.T) {
...
+		},
+		"insecure 2.4.1 dmg with 2.4.0 server": {
+			selfVersion: "2.4.0",
+			ctx: metadata.NewIncomingContext(test.Context(t), metadata.Pairs(
+				proto.DaosComponentHeader, build.ComponentAdmin.String(),
+				proto.DaosVersionHeader, "2.4.1",
+			)),
+			nonSysMsg: true,
+		},
+		"insecure 2.6.0 dmg with 2.4.0 server": {
```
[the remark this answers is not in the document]
mjmac: Yep, added a test case.

src/control/server/interceptors.go
```diff
@@ -116,7 +117,32 @@ var selfServerComponent = func() *build.VersionedComponent {
...
+func compVersionFromHeaders(ctx context.Context) *build.VersionedComponent {
+	md, hasMD := metadata.FromIncomingContext(ctx)
+	if !hasMD {
+		return nil
+	}
+	compName, hasName := md[proto.DaosComponentHeader]
+	if !hasName {
+		return nil
+	}
+	comp := build.Component(compName[0])
```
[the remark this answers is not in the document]
mjmac: Hmm. It is checked later, by [build.NewVersionedComponent()](https://github.com/daos-stack/daos/blob/master/src/control/build/interop.go#L46). When that returns an error, the other component is set to "unknown", and the test case I've added for this scenario fails as expected.

That having been said, given that I think the goal should be to move toward using the headers for both secure and insecure mode, I've refactored `compVersionFromHeaders()` a little to do more immediate input validation.

src/control/server/interceptors.go
```diff
@@ -127,33 +153,53 @@ func checkVersion(ctx context.Context, self *build.VersionedComponent, req inter
...
+	// Prefer the new header-based component/version mechanism.
+	// If we are in secure mode, verify that the component presented
+	// in the header matches the certificate's component.
+	fromHeaders := compVersionFromHeaders(ctx)
+	if fromHeaders != nil {
+		otherVersion = fromHeaders.Version
+		if isInsecure {
+			otherComponent = fromHeaders.Component
+		} else if otherComponent != fromHeaders.Component {
+			return status.Errorf(codes.PermissionDenied, "invalid component %q for request", fromHeaders.Component)
```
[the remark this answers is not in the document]
mjmac: Done.

src/control/lib/control/interceptors.go
```diff
@@ -71,5 +73,21 @@ func unaryErrorInterceptor() grpc.DialOption {
...
 		return nil
-	})
+	}
+}
+
+// unaryVersionedComponentInterceptor appends the component name and version to the
+// outgoing request headers.
+func unaryVersionedComponentInterceptor() grpc.UnaryClientInterceptor {
+	return func(parent context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		comp, err := security.MethodToComponent(method)
```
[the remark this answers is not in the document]
mjmac: > If these are the outgoing headers, surely we can determine the component name with no ambiguity from something inside the binary?

Hmm. Currently, no. I thought that I might have added something a while ago, but I think I was remembering something from an experiment that never landed. Added a new control.Client option to specify the component, with the method lookup helper as a fallback.

src/control/security/grpc_authorization.go
```diff
@@ -77,6 +83,16 @@ var methodAuthorizations = map[string][]Component{
 	"/RaftTransport/InstallSnapshot":       {ComponentServer},
 }
 
+// MethodToComponent resolves a gRPC method string to a build.Component.
+func MethodToComponent(method string) (build.Component, error) {
+	comps, found := methodAuthorizations[method]
+	if !found || len(comps) == 0 {
```
knard-intel, line 89: Could it be possible that len(comps) > 1

knard-intel (approved):
LGTM

src/control/security/grpc_authorization.go
kjacque: `len(comps) != 1` may be a better check. It is possible to have multiple components, in which case this method can't tell which one is correct.

kjacque (commented):
I think Cedric's comment is worth addressing. Everything else LGTM.

src/control/security/grpc_authorization.go
mjmac: Yeah, agreed that this helper can be more strict now that the caller can and should explicitly set the calling component. Fixed, and added a unit test for the helper as well.

kjacque (approved):

knard-intel (approved):

src/control/security/grpc_authorization_test.go
```diff
@@ -218,3 +221,56 @@ func TestSecurity_AuthorizedRpcsAreValid(t *testing.T) {
...
+				"/one": {ComponentServer},
+			},
+			expComp: build.ComponentServer,
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			var gotComp build.Component
+			var gotErr error
+
+			if tc.authMap != nil {
```
tanabarr, line 262: nit: rather than calling different functions here I would just set tc.authMap = methodAuthorizations if not specified in the test case

src/control/server/interceptors.go
```diff
@@ -116,7 +121,25 @@ var selfServerComponent = func() *build.VersionedComponent {
...
+		return nil, errNoReqMetadata
+	}
+	compName, hasName := md[proto.DaosComponentHeader]
+	if !hasName {
+		return nil, errNoReqMetadata
+	}
+	comp := build.Component(compName[0])
+	compVersion, hasVersion := md[proto.DaosVersionHeader]
+	if !hasVersion {
+		return nil, errNoReqMetadata
```
tanabarr, line 136: is there value in wrapping these generic errors to tell which part of the meta data is missing? might be helpful in diagnosing problems

tanabarr (approved):
nice solution!

src/control/security/grpc_authorization_test.go
mjmac: I figured that this way would give us full coverage by including the exported wrapper.

PR #13022, opened 2023-09-08
DAOS-14208 rebuild: several fixes for multiple shards in the same target

wangdi1:
Checking rebuilding shard by comparing the old and new layout map, similar as drain/reintegration, since non-failure shard might be moved to other targets due to co-locate and failure domain factors.

Add reclaim phase for rebuild as well.

Add test to verify it.

A few other fixes due to this.

Increase rebuild EC timeout value.

Disable rs_obj_nr and rs_rec_nr check, since
reclaim phase is added after rebuild, rs_obj_nr
and rs_rec_nr might be the objects and record
being reclaimed, rather than rebuilt.

Test-tag: pr rebuild ec
Features: rebuild ec
Allow-unstable-test: true
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/object/srv_obj_migrate.c
+++ b/src/object/srv_obj_migrate.c
@@ -620,7 +620,6 @@ mrone_obj_fetch(struct migrate_one *mrone, daos_handle_t oh, d_sg_list_t *sgls,
 		d_iov_t *csum_iov_fetch)
 {
 	struct migrate_pool_tls	*tls;
-	struct dc_object	*obj;
 	int			rc = 0;
 
 	tls = migrate_pool_tls_lookup(mrone->mo_pool_uuid,
@@ -634,21 +633,7 @@ mrone_obj_fetch(struct migrate_one *mrone, daos_handle_t oh, d_sg_list_t *sgls,
 	if (daos_oclass_grp_size(&mrone->mo_oca) > 1)
 		flags |= DIOF_TO_LEADER;
 
-	/**
-	 * For EC data migration, let's force it to do degraded fetch,
-	 * make sure reintegration will not fetch from the original
-	 * shard, which might cause parity corruption.
-	 */
-	obj = obj_hdl2ptr(oh);
-	if (iods[0].iod_type != DAOS_IOD_SINGLE &&
-	    daos_oclass_is_ec(&mrone->mo_oca) &&
-	    is_ec_data_shard(obj, mrone->mo_dkey_hash, mrone->mo_oid.id_shard) &&
-	    obj_ec_parity_alive(oh, mrone->mo_dkey_hash, NULL))
-		flags |= DIOF_FOR_FORCE_DEGRADE;
-
-	obj_decref(obj);
-
-	rc = dsc_obj_fetch(oh, mrone->mo_epoch, &mrone->mo_dkey,
+	rc = dsc_obj_fetch(oh, eph, &mrone->mo_dkey,
 			   iod_num, iods, sgls, NULL,
 			   flags, NULL, csum_iov_fetch);
 	if (rc != 0)
@@ -669,7 +654,7 @@ mrone_obj_fetch(struct migrate_one *mrone, daos_handle_t oh, d_sg_list_t *sgls,
 		csum_iov_fetch->iov_len = 0;
 		csum_iov_fetch->iov_buf = p;
 
-		rc = dsc_obj_fetch(oh, mrone->mo_epoch, &mrone->mo_dkey, iod_num, iods, sgls,
+		rc = dsc_obj_fetch(oh, eph, &mrone->mo_dkey, iod_num, iods, sgls,
 				   NULL, flags, NULL, csum_iov_fetch);
 	}
 
@@ -1223,7 +1208,8 @@ out:
 
 static int
 __migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
-			    daos_iod_t *iods, int iod_num, daos_epoch_t update_eph,
+			    daos_iod_t *iods, int iod_num, daos_epoch_t fetch_eph,
+			    daos_epoch_t update_eph,
 			    uint32_t flags, struct ds_cont_child *ds_cont)
 {
 	d_sg_list_t		 sgls[OBJ_ENUM_UNPACK_MAX_IODS];
@@ -1282,8 +1268,7 @@ __migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
 		p_csum_iov = &csum_iov;
 	}
 
-	rc = mrone_obj_fetch(mrone, oh, sgls, iods, iod_num, mrone->mo_epoch,
-			     flags, p_csum_iov);
+	rc = mrone_obj_fetch(mrone, oh, sgls, iods, iod_num, fetch_eph, flags, p_csum_iov);
 	if (rc) {
 		D_ERROR("migrate dkey "DF_KEY" failed: "DF_RC"\n",
 			DP_KEY(&mrone->mo_dkey), DP_RC(rc));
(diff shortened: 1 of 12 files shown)


src/object/srv_obj_migrate.c
```diff
@@ -1370,22 +1356,19 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
...
+		for (j = 0; j < mrone->mo_iods_from_parity[i].iod_nr; j++) {
+			daos_iod_t iod = mrone->mo_iods_from_parity[i];
 
-		for (i = 0; i < mrone->mo_iods_num_from_parity; i++) {
-			for (j = 0; j < mrone->mo_iods_from_parity[i].iod_nr; j++)
-				min_eph = min(min_eph,
-					      mrone->mo_iods_update_ephs_from_parity[i][j]);
+			iod.iod_nr = 1;
+			iod.iod_recxs = &mrone->mo_iods_from_parity[i].iod_recxs[j];
+			rc = __migrate_fetch_update_bulk(mrone, oh, &iod, 1,
```
liuxuezhao, line 1365: just confirm that now the parity exts will be rebuild one by one, as previous code the take the min epoch to update multiple parity ext may cause client-side degraded fetch's parity epoch mismatch issue?

src/object/srv_obj_migrate.c
```diff
@@ -634,21 +633,7 @@ mrone_obj_fetch(struct migrate_one *mrone, daos_handle_t oh, d_sg_list_t *sgls,
...
-	obj = obj_hdl2ptr(oh);
-	if (iods[0].iod_type != DAOS_IOD_SINGLE &&
-	    daos_oclass_is_ec(&mrone->mo_oca) &&
-	    is_ec_data_shard(obj, mrone->mo_dkey_hash, mrone->mo_oid.id_shard) &&
-	    obj_ec_parity_alive(oh, mrone->mo_dkey_hash, NULL))
-		flags |= DIOF_FOR_FORCE_DEGRADE;
-
-	obj_decref(obj);
-
-	rc = dsc_obj_fetch(oh, mrone->mo_epoch, &mrone->mo_dkey,
```
liuxuezhao, line 651: just confirm why don't need to force degraded fetch now?
is it because after this PR the "po_rebuilding = 1" for failed tgt is set no need not the FORCE_DEGRADED flag?

src/object/srv_obj_migrate.c
```diff
@@ -2579,7 +2562,7 @@ migrate_one_epoch_object(daos_epoch_range_t *epr, struct migrate_pool_tls *tls,
 	/* Only open with RW flag, reintegrating flag will be set, which is needed
```
liuxuezhao, line 2562: may need to update comment here "RW flag"
and why must change DAOS_OO_RW to DAOS_OO_RO for dsc_obj_list_obj()? to include the DOWN tgt in the layout? (will do degraded enumerate for the DOWN tgt?)

src/placement/tests/jump_map_place_obj.c
```diff
@@ -1607,7 +1607,7 @@ placement_handles_multiple_states(void **state)
 	 */
 	ctx.ver = ver_after_fail;
 	jtc_scan(&ctx);
-	assert_int_equal(ctx.rebuild.out_nr, 1);
+	assert_int_equal(ctx.rebuild.out_nr, 2);
```
liuxuezhao, line 1610: why 2 shards changed? and the comment in L1606 need update.

src/rebuild/srv.c
```diff
@@ -1412,7 +1412,8 @@ rebuild_task_complete_schedule(struct rebuild_task *task, struct ds_pool *pool,
 					 task->dst_new_layout_version, &task->dst_tgts,
 					 retry_opc, 5);
 	} else if (task->dst_rebuild_op == RB_OP_REINT || task->dst_rebuild_op == RB_OP_EXTEND ||
-		   task->dst_rebuild_op == RB_OP_UPGRADE) {
+		   task->dst_rebuild_op == RB_OP_UPGRADE || task->dst_rebuild_op == RB_OP_EXCLUDE ||
+		   task->dst_rebuild_op == RB_OP_DRAIN) {
 		/* Otherwise schedule reclaim for reintegrate/extend/upgrade. */
```
liuxuezhao, line 1417: this comment need update as well.
But why need to schedule the RB_OP_RECLAIM task after rebuild done for RB_OP_EXCLUDE? and DRAIN?

liuxuezhao (commented):
some questions to confirm, thanks.

src/object/srv_obj_migrate.c
gnailzenh: The same question

wangdi1: yes, especially if the shard needs to be relocated to different shard due to exclude/reint etc, and if there are no available shards to reconstruct the shard, so it has to fetch from the original shard, then we have to rebuild one by one, and can not merge the shard, which might ruin the data.

src/object/srv_obj_migrate.c
wangdi1: oh, no. because we will use exact epoch to fetch the data, so it does not need to reconstruct the data from the parity, to make sure the epoch of the data shard will not ruin the data, (see above),  anymore.

src/placement/tests/jump_map_place_obj.c
wangdi1: Oh, because we compare the shard to know how many shard needs to be rebuilt, so the shard might need to be rebuilt even it is healthy, due to satisfy the co-location or RF etc, i.e. there might be more shard needs to be rebuilt in some rare cases.

src/rebuild/srv.c
wangdi1: because some healthy shard might also be rebuilt after exclude/drain, due to co-located and rf issues, that is one of the reason we saw corruption in some cases.   See my test in this PR.

src/object/srv_obj_migrate.c
wangdi1: Because read-only case will only fetch the data, so during reintegratin/rebuilding, the layout might include some rebuilding shard, which might be healthy shard, and also make sure the layout is calculated by the original pool map, so DOWN is allowed status here.

DOWN tgt will be skipped by client read stack.  This is mainly used by rebuild/migration.

src/object/srv_obj_migrate.c
```diff
@@ -1370,22 +1356,19 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
 	 * this data shard.
```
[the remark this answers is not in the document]
wangdi1: will update the comment by later patch. Thanks.

src/rebuild/srv.c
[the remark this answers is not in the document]
wangdi1: yes, will update if the PR needs to refresh.

src/object/srv_obj_migrate.c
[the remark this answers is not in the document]
wangdi1: yes, see above comments

src/object/srv_obj_migrate.c
```diff
@@ -634,21 +633,7 @@ mrone_obj_fetch(struct migrate_one *mrone, daos_handle_t oh, d_sg_list_t *sgls,
...
-	 * For EC data migration, let's force it to do degraded fetch,
-	 * make sure reintegration will not fetch from the original
-	 * shard, which might cause parity corruption.
-	 */
-	obj = obj_hdl2ptr(oh);
-	if (iods[0].iod_type != DAOS_IOD_SINGLE &&
-	    daos_oclass_is_ec(&mrone->mo_oca) &&
-	    is_ec_data_shard(obj, mrone->mo_dkey_hash, mrone->mo_oid.id_shard) &&
-	    obj_ec_parity_alive(oh, mrone->mo_dkey_hash, NULL))
-		flags |= DIOF_FOR_FORCE_DEGRADE;
```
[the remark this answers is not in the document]
wangdi1: see above. Xuezhao has the same question. Thanks.

liuxuezhao (approved):
seems with some CI test failure

gnailzenh (approved):

PR #13027, opened 2023-09-08
DAOS-11552 doc: Document Interoperability Policy

mjmac:
Formally document the policy with a reference table
in the Admin Guide.

### DIFF
--- a/docs/admin/administration.md
+++ b/docs/admin/administration.md
@@ -969,11 +969,11 @@ policies for all major components in a DAOS system.
 
 ||Server<br>(daos_server)|Engine<br>(daos_engine)|Agent<br>(daos_agent)|Client<br>(libdaos)|Admin<br>(dmg)|
 |:---|:---:|:---:|:---:|:---:|:---:|
-|Server|x.y|x.y.z|x.(y±1)|n/a|x.y|
-|Engine||x.y|n/a|x.(y±1)|n/a|
-|Agent|||n/a|x.y.z|n/a|
-|Client||||n/a|n/a|
-|Admin|||||n/a|
+|Server|x.y.z|x.y.z|x.(y±1)|n/a|x.y|
+|Engine|x.y.z|x.y.z|n/a|x.(y±1)|n/a|
+|Agent|x.(y±1)|n/a|n/a|x.y.z|n/a|
+|Client|n/a|x.(y±1)|x.y.z|n/a|n/a|
+|Admin|x.y|n/a|n/a|n/a|n/a|
 
 Key:
   * x.y.z: Major.Minor.Patch must be equal
--- a/docs/admin/administration.md
+++ b/docs/admin/administration.md
@@ -970,10 +970,10 @@ policies for all major components in a DAOS system.
 ||Server<br>(daos_server)|Engine<br>(daos_engine)|Agent<br>(daos_agent)|Client<br>(libdaos)|Admin<br>(dmg)|
 |:---|:---:|:---:|:---:|:---:|:---:|
 |Server|x.y|x.y.z|x.(y±1)|n/a|x.y|
-|Engine|x.y.z|x.y|n/a|x.(y±1)|n/a|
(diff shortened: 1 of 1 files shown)


mjmac:
@Michael-Hennecke, @ipoddubn: Pushed this to open discussion. I think there is a fair amount of uncertainty about the exact details of what is supported, so hopefully we can memorialize the definitive policy as part of the documentation.

ipoddubn:
I'm generally good with the proposal and inclined to approve the suggested format. Let me discuss this internally a bit more before I approve.

docs/admin/administration.md
```diff
@@ -960,3 +960,28 @@ DAOS v2.2 client connections to pools which were created by DAOS v2.4
...
+
+### Interoperability Matrix
+
+The following table is intended to visually depict the interoperability
+policies for all major components in a DAOS system.
+
+
+||Server<br>(daos_server)|Engine<br>(daos_engine)|Agent<br>(daos_agent)|Client<br>(libdaos)|Admin<br>(dmg)|
+|:---|:---:|:---:|:---:|:---:|:---:|
+|Server|x.y|x.y.z|x.(y±1)|n/a|x.y|
```
kccain, line 972: libdaos client <-> engine interoperability in this table shows minor version differences of at most 1.
However in this same document in the Software Upgrade section it states that DAOS v2.0 will maintain RPC protocol and persistent format compatibility for all future DAOS 2.x versions.

> DAOS v2.0 will maintain interoperability for both the wire protocol and persistent layout with any future v2.x versions. That being said, it is required that all engines in the same system run the same DAOS version.

Note: In some upcoming work for 2.6 feature development I will be needing to update the client <-> engine RPC protocol versions, and the pool metadata layout version too. While doing this I will need to know whether back-compatibility all the way back to 2.2 (and 2.0) software is required (I am assuming it is). If my assumption is correct then it may require libdaos/engine managing not just two protocol versions (current and "minus 1"), but also a third version, something not yet done in the development.
That other work is described in https://daosio.atlassian.net/wiki/spaces/DAOS/pages/11411390465/Metadata+Duplicate+Operation+Detection and I have an (admittedly) imperfect description of interop challenges in that document.

kccain: may want feedback from @wangshilong on this

mjmac:
Pushed what I hope is a final update: [v3](https://github.com/daos-stack/daos/blob/32ba0bf72a1e5aac4c3ad793bf73755deb008c43/docs/admin/administration.md#interoperability-matrix)

I decided to repopulate all cells and tightened up the server/server requirements for now. Things can be relaxed later if testing supports it.

Michael-Hennecke (approved):

kjacque (commented):
The one item that stands out to me is agent vs. client required to be identical down to the patch version. Practically speaking, we don't actually compare the versions for agent and client, but I would expect libdaos to work with any agent of the same major.minor version. If that's not true, it seems like a mistake on our part.

mjmac:
> The one item that stands out to me is agent vs. client required to be identical down to the patch version. Practically speaking, we don't actually compare the versions for agent and client, but I would expect libdaos to work with any agent of the same major.minor version. If that's not true, it seems like a mistake on our part.

The documented policy here is not necessarily a reflection of the technical capabilities, but what Intel is willing to test and support. I agree that it's probably technically fine for the agent and client to have the same major.minor versions, but practically speaking, other combinations aren't tested and aren't likely to be seen in production where the agent and client are installed/upgraded together via RPM.

kjacque:
> The documented policy here is not necessarily a reflection of the technical capabilities, but what Intel is willing to test and support. I agree that it's probably technically fine for the agent and client to have the same major.minor versions with different patch versions, but practically speaking, other combinations aren't tested and aren't likely to be seen in production where the agent and client are installed/upgraded together via RPM.

Fair point. I'm fine with approving if this is all we're testing/supporting.

kjacque (approved):

PR #13044, opened 2023-09-12
DAOS-14223 control: Tidy up pool create workflow

tanabarr:
Calculation of component storage values before creating a pool happens
in a number of different places, this change starts the process of
consolidating this logic and improving the clarity of the create flow.

Allow-unstable: true
Features: control
Required-githooks: true

### Gatekeeper:


src/control/server/mgmt_pool.go
```diff
@@ -167,73 +168,76 @@ func (svc *mgmtSvc) calculateCreateStorage(req *mgmtpb.PoolCreateReq) error {
...
+		if nvmeMissing && nvmeBytes > 0 {
+			return errors.Errorf("%s NVMe requested for pool but config has zero bdevs",
+				humanize.Bytes(nvmeBytes))
+		}
+
+	// Pool tier sizes to be populated based on total-size and ratio.
+	case len(req.Tierbytes) == 0 && len(req.Tierratio) == 2 && req.Totalbytes > 0:
+		// If no NVMe, adjust ratio as NVMe hasn't been specifically requested.
+		if nvmeMissing {
+			svc.log.Info("config has zero bdevs; excluding NVMe from pool create " +
```
knard-intel, line 203: From my side, I would increase the log level to notice or warning as we change the input request.

knard-intel (dismissed):
Just one minor comments to take into account if you have to update your PR

knard-intel (approved):
LGTM

src/control/server/mgmt_pool_test.go
```diff
@@ -213,14 +212,13 @@ func TestServer_MgmtSvc_calculateCreateStorage(t *testing.T) {
...
 				Tierbytes: []uint64{defaultScmBytes / 2, defaultNvmeBytes / 2},
-				Tierratio: []float64{0, 0},
 				Ranks:     []uint32{0, 1},
 			},
 		},
 		"auto sizing (not enough SCM)": {
 			in: &mgmtpb.PoolCreateReq{
 				Totalbytes: scmTooSmallTotal,
-				Tierratio:  []float64{scmTooSmallRatio},
+				Tierratio:  []float64{scmTooSmallRatio, 0},
```
kjacque, line 221: what happens if you don't provide the nvme tier?

src/control/server/mgmt_pool_test.go
```diff
@@ -307,16 +331,18 @@ func TestServer_MgmtSvc_PoolCreate(t *testing.T) {
...
-		memberCount   int
-		req           *mgmtpb.PoolCreateReq
-		drpcRet       *mgmtpb.PoolCreateResp
-		expResp       *mgmtpb.PoolCreateResp
-		expErr        error
+		mgmtSvc        *mgmtSvc
+		setupMockDrpc  func(_ *mgmtSvc, _ error)
+		targetCount    int
+		memberCount    int
+		mdonssdEnabled bool
```
kjacque, line 341: It would be better to use bumpy case for the mdOnSSD portion of the name, too.

kjacque (approved):
Overall looks good. Nice cleanup.

src/control/cmd/dmg/pool.go
```diff
@@ -223,6 +223,114 @@ func (cmd *PoolCreateCmd) Execute(args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	availRatio := cmd.Size.availRatio
+	if availRatio < 100 {
+		scmBytes = availRatio * scmBytes / 100
+		nvmeBytes = availRatio * nvmeBytes / 100
+	}
+	req.TierBytes = []uint64{scmBytes, nvmeBytes}
```
mjmac, line 265: Ideally all of this logic should be moved behind the `control.PoolCreate()` API so that non-dmg users can take advantage of it.

mjmac (commented):
This is a good step in the right direction, but I am wondering why it doesn't go further in the case of moving the create-by-size logic into the `PoolCreate()` client implementation. Ideally, dmg should be just a thin wrapper around the APIs, with no "business logic".

src/control/cmd/dmg/pool.go
tanabarr: I tried doing that and ran into an issue which I now cannot remember, I will try again and report back as to why it was a problem.

src/control/server/mgmt_pool_test.go
tanabarr: that's not possible because of the logic in `dmg/pool.go:storageManual()` and validation in `lib/control/pool.go:poolCreateReqChkSizes()`

tanabarr: i.e. the validation in both those functions constricts the possible combinations so missing nvme tier would be handled early. Possible I'm not understanding the context of your question correctly though.

src/control/server/mgmt_pool_test.go
tanabarr: will do if I have to re-push

tanabarr:
> This is a good step in the right direction, but I am wondering why it doesn't go further in the case of moving the create-by-size logic into the `PoolCreate()` client implementation. Ideally, dmg should be just a thin wrapper around the APIs, with no "business logic".

I remember why now, I was trying to do exactly that but found that I couldn't log to info when calling `GetMaxPoolSize()` from within `lib/control/pool.go:PoolCreate()` using the UnaryInvoker client as a debugLogger. I could have reduced all the log levels used within the function to Debug but it seems that some of them really should be Info, Notice or Warn. What is your leaning? should I just reduce them to Debug and call the function from the control API PoolCreate()?

tanabarr:
> > I remember why now, I was trying to do exactly that but found that I couldn't log to info when calling `GetMaxPoolSize()` from within `lib/control/pool.go:PoolCreate()` using the UnaryInvoker client as a debugLogger. I could have reduced all the log levels used within the function to Debug but it seems that some of them really should be Info, Notice or Warn. What is your leaning? should I just reduce them to Debug and call the function from the control API PoolCreate()?
> 
> Oh, bleah. I see that `control.GetMaxPoolSize()` accepts a `logging.Logger` instead of using the debug logger in the client. Guess that one slipped through review... Nothing in the control library should be logging above the Debug level. If there is some critical information that needs to be conveyed back to the caller, then it should be added to the return signature. Looking at the helpers, though, I only see one instance of a Notice-level log message, the rest are all Debug.
> 
> In my opinion, we need to think a bit more about what it means for a SMD device to be unusable during a pool create. If it's really an error, then we should return an error. Otherwise, what is the admin supposed to do with that information?

Do you require changes to this PR? Should I change the function to accept a debugLogger and change the notice to an error return? Otherwise if we want to continue regardless of unusable devices then we can just log to debug. IMO returning an error if we can't use a device is valid.

mjmac:
> Do you require changes to this PR? Should I change the function to accept a debugLogger and change the notice to an error return? Otherwise if we want to continue regardless of unusable devices then we can just log to debug. IMO returning an error if we can't use a device is valid.

IMO it doesn't make sense to land this without moving the call to `control.GetMaxPoolSize()` into `control.PoolCreate()`. You could probably unexport it and make it an internal helper as well. There's no need to pass in a debugLogger as it's provided by the client implementation.

I am not sure about why the logic was added for skipping/logging unusable devices vs returning an error. I believe that may have been added by @knard-intel, so perhaps he can comment. My inclination would be to make it an error, but if it turns out there are clear use cases where we should allow a PoolCreate to proceed even with unusable devices detected, then I think it would be better to make that explicit behavior in a follow-on. For example, allowing the devices to be marked in such a way that they're ignored in scans, or perhaps adding a new PoolCreate flag that allows unusable devices to be skipped. The current implementation isn't ideal because it would be very easy for the unusable device log lines to be overlooked, particularly in a scripted/automated environment.

src/control/cmd/dmg/pool.go
knard-intel: From what I remember, the pool max size calculation was integrated this way for not being too much intrusive in the code.
However, I perfectly OK to move it behind `control.PoolCreate()`: it will probably needs some slight modifications of the `PoolCreateReq` structure.

knard-intel:
> > Do you require changes to this PR? Should I change the function to accept a debugLogger and change the notice to an error return? Otherwise if we want to continue regardless of unusable devices then we can just log to debug. IMO returning an error if we can't use a device is valid.
> 
> IMO it doesn't make sense to land this without moving the call to `control.GetMaxPoolSize()` into `control.PoolCreate()`. You could probably unexport it and make it an internal helper as well. There's no need to pass in a debugLogger as it's provided by the client implementation.
> 
> I am not sure about why the logic was added for skipping/logging unusable devices vs returning an error. I believe that may have been added by @knard-intel, so perhaps he can comment. My inclination would be to make it an error, but if it turns out there are clear use cases where we should allow a PoolCreate to proceed even with unusable devices detected, then I think it would be better to make that explicit behavior in a follow-on. For example, allowing the devices to be marked in such a way that they're ignored in scans, or perhaps adding a new PoolCreate flag that allows unusable devices to be skipped. The current implementation isn't ideal because it would be very easy for the unusable device log lines to be overlooked, particularly in a scripted/automated environment.

I have no issue of moving GetMaxPoolSize

> > Do you require changes to this PR? Should I change the function to accept a debugLogger and change the notice to an error return? Otherwise if we want to continue regardless of unusable devices then we can just log to debug. IMO returning an error if we can't use a device is valid.
> 
> IMO it doesn't make sense to land this without moving the call to `control.GetMaxPoolSize()` into `control.PoolCreate()`. You could probably unexport it and make it an internal helper as well. There's no need to pass in a debugLogger as it's provided by the client implementation.
> 
> I am not sure about why the logic was added for skipping/logging unusable devices vs returning an error. I believe that may have been added by @knard-intel, so perhaps he can comment. My inclination would be to make it an error, but if it turns out there are clear use cases where we should allow a PoolCreate to proceed even with unusable devices detected, then I think it would be better to make that explicit behavior in a follow-on. For example, allowing the devices to be marked in such a way that they're ignored in scans, or perhaps adding a new PoolCreate flag that allows unusable devices to be skipped. The current implementation isn't ideal because it would be very easy for the unusable device log lines to be overlooked, particularly in a scripted/automated environment.

See my answer in the original code comments.
To resume, moving the max pool size calculation is perfectly fine for me.

src/control/lib/control/pool_test.go
```diff
@@ -2803,3 +2910,524 @@ func TestControl_GetMaxPoolSize(t *testing.T) {
...
+				"Invalid size of TierRatio attribute: disabled with manual allocation")
+			//			msg := fmt.Sprintf("Creating DAOS pool with %d%% of all storage",
+			//				int(tc.StorageRatio*100))
+			//			test.AssertTrue(t, strings.Contains(buf.String(), msg),
+			//				fmt.Sprintf("missing success message: %q", msg))
+			//			if tc.ExpectedOutput.WarningMsg != "" {
+			//				test.AssertTrue(t,
+			//					strings.Contains(buf.String(), tc.ExpectedOutput.WarningMsg),
+			//					"missing warning message: "+tc.ExpectedOutput.WarningMsg)
+			//			}
```
kjacque, line 3430: Needs to be removed?

kjacque (commented):
Changes look reasonable to me. Would be best to remove this commented out code before landing.

src/control/lib/control/pool_test.go
tanabarr: thanks, fixed

tanabarr:
pylint and codespell checks seem to be picking up issues with unchanged code

https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-13044/9/tests/ is showing a failure that needs investigating

knard-intel (approved):
LGTM.

kjacque (approved):

tanabarr:
ready to land, codespell and pylint check failures are for codes not changed in this PR

tanabarr:
@daos-stack/daos-gatekeeper can this be landed please

PR #13049, opened 2023-09-13
DAOS-14352 object: fetch epoch should be lower than stable epoch

wangdi1:
fetch epoch should be lower than stable epoch to avoid inflight data being rebuilt as well.

Required-githooks: true
Features: rebuild ec

### Gatekeeper:

### DIFF
--- a/src/object/srv_obj_migrate.c
+++ b/src/object/srv_obj_migrate.c
@@ -1350,22 +1350,38 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
 	/* For EC object, if the migration include both extent from parity rebuild
 	 * and extent from replicate rebuild, let rebuild the extent with parity first,
 	 * then extent from replication.
-	 *
-	 * Since the parity shard epoch should be higher or equal to the data shard epoch,
-	 * so let's use the minimum epochs of all parity shards as the update epoch of
-	 * this data shard.
 	 */
-
 	for (i = 0; i < mrone->mo_iods_num_from_parity; i++) {
 		for (j = 0; j < mrone->mo_iods_from_parity[i].iod_nr; j++) {
 			daos_iod_t iod = mrone->mo_iods_from_parity[i];
+			daos_epoch_t fetch_eph;
 
 			iod.iod_nr = 1;
 			iod.iod_recxs = &mrone->mo_iods_from_parity[i].iod_recxs[j];
-			rc = __migrate_fetch_update_bulk(mrone, oh, &iod, 1,
-							 mrone->mo_iods_update_ephs_from_parity[i][j],
-							 mrone->mo_iods_update_ephs_from_parity[i][j],
-							 DIOF_EC_RECOV_FROM_PARITY, ds_cont);
+
+			/* If the epoch is higher than EC aggregate boundary, then
+			 * it should use stable epoch to fetch the data, since
+			 * the data could be aggregated independently on parity
+			 * and data shard, so using stable epoch could make sure
+			 * the consistency view during rebuild. And also EC aggregation
+			 * should already aggregate the parity, so there should not
+			 * be any partial update on the parity as well.
+			 *
+			 * Otherwise there might be partial update on this rebuilding
+			 * shard, so let's use the epoch from the parity shard to fetch
+			 * the data here, which will make sure partial update will not
+			 * be fetched here. And also EC aggregation is being disabled
+			 * at the moment, so there should not be any vos aggregation
+			 * impact this process as well.
+			 */
+			if (ds_cont->sc_ec_agg_eph_boundary >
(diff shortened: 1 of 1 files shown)


src/object/srv_obj_migrate.c
```diff
@@ -1350,20 +1350,19 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
...
+			daos_epoch_t fetch_eph;
 
 			iod.iod_nr = 1;
 			iod.iod_recxs = &mrone->mo_iods_from_parity[i].iod_recxs[j];
-			rc = __migrate_fetch_update_bulk(mrone, oh, &iod, 1,
-							 mrone->mo_iods_update_ephs_from_parity[i][j],
+			/* fetch epoch can not be higher than rebuild stable epoch(mo_eph),
+			 * to avoid fetching the inflight(new) data.
+			 */
+			fetch_eph = min(mrone->mo_epoch, mrone->mo_iods_update_ephs_from_parity[i][j]);
```
liuxuezhao, line 1364: just confirm that the "mrone->mo_epoch" is the rebuild epoch (stable epoch) and also the dsc_obj_list_obj()'s epoch range's epr_hi?
the mo_iods_update_ephs_from_parity is the enumerated ext's epoch?
then mo_iods_update_ephs_from_parity should not be higher than the list_obj's epoch (rebuild stable epoch) right? thx

wangdi1: >just confirm that the "mrone->mo_epoch" is the rebuild epoch (stable epoch) and also the dsc_obj_list_obj()'s epoch range's epr_hi?

right.  Enumerate epoch is already limited by the stable epoch. And this patch will limit the epoch by stable epoch as well.   This is actually a regression from recent PR.  Sorry I may miss your question here.

liuxuezhao (approved):

src/object/srv_obj_migrate.c
```diff
@@ -1350,22 +1350,38 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
...
+			 * the data here, which will make sure partial update will not
+			 * be fetched here. And also EC aggregation is being disabled
+			 * at the moment, so there should not be any vos aggregation
+			 * impact this process as well.
+			 */
+			if (ds_cont->sc_ec_agg_eph_boundary >
+			    mrone->mo_iods_update_ephs_from_parity[i][j])
+				fetch_eph = mrone->mo_epoch;
+			else
+				fetch_eph = mrone->mo_iods_update_ephs_from_parity[i][j];
```
NiuYawei, line 1381: I don't see the purpose of PR, is it for bug fixing or performance improvement?

When the sc_ec_agg_eph_boundary is less than mo_iods_update_ephs_from_parity[i][j], it still keeps old behavior, so sounds to me this is not a bug fix?

wangdi1:
failed due to DAOS-14320 and 7378

src/object/srv_obj_migrate.c
wangdi1: oh, it is a bug fix. it is just to avoid the impact of vos aggregation on the EC rebuild.

>When the sc_ec_agg_eph_boundary is less than mo_iods_update_ephs_from_parity[i][j], it still keeps old behavior, so sounds to me this is not a bug fix?

So agg_eph_boundary is the upper limit for VOS aggregation, if it is greater than the epoch (got from parity enumeration), then it should use the stable epoch to do logic view fetch to avoid the vos aggregation impacts, similar as replicate object rebuild.


Otherwise, it should use the parity epoch to fetch the data, because 

1. vos aggregation will be blocked by agg_eph_boundary, so no impact here.
2. it can not use stable epoch in this case, otherwise it might ruin the partial(overwrite) recx rebuild.  And for above case, since EC aggregation already processed the partial(overwrite) update, so there should not be any overwrite partial recx left.

src/object/srv_obj_migrate.c
```diff
@@ -1350,22 +1350,38 @@ migrate_fetch_update_bulk(struct migrate_one *mrone, daos_handle_t oh,
...
+			daos_epoch_t fetch_eph;
 
 			iod.iod_nr = 1;
 			iod.iod_recxs = &mrone->mo_iods_from_parity[i].iod_recxs[j];
-			rc = __migrate_fetch_update_bulk(mrone, oh, &iod, 1,
-							 mrone->mo_iods_update_ephs_from_parity[i][j],
-							 mrone->mo_iods_update_ephs_from_parity[i][j],
-							 DIOF_EC_RECOV_FROM_PARITY, ds_cont);
+
+			/* If the epoch is higher than EC aggregate boundary, then
```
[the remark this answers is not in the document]
wangdi1: ah, yes. I can update the comments in the following patch. really do not want go through the CI again. Thanks.

gnailzenh (approved):

PR #13052, opened 2023-09-14
DAOS-14270 il: Move initialization to constructor

jolivier23:
To workaround issue with multiple threads racing on, getenv/setenv move the daos_init call to the constructor for libioil. Push the change so it can run through CI to see if there are issues with doing it this way.

Required-githooks: true

### Gatekeeper:


jolivier23:
Well, it passed so maybe it's ok.  We can potentially do this under an environment variable setting to make it less risky if that's preferred.

src/client/dfuse/il/int_posix.c
```diff
@@ -320,6 +378,15 @@ ioil_init(void)
 		return;
 
 	ioil_iog.iog_initialized = true;
+
+	rc = pthread_mutex_lock(&ioil_iog.iog_lock);
+	D_ASSERT(rc == 0);
+
+	if (!ioil_iog.iog_daos_init)
+		(void)call_daos_init(0);
```
ashleypittman, line 386: What would happen if this call failed?  I think it would be fine wouldn't it, the function cleans up after itself and there's no other state to unwind, and it wouldn't then get called a second time?

jolivier23: yeah, that was my thinking

jolivier23: The other place it's called doesn't print anything either when it fails.  I think any prints would be internal so there is really nothing to do with the return code.

src/client/dfuse/il/int_posix.c
```diff
@@ -271,6 +271,64 @@ init_links(void)
...
+ * calling daos_init() and close them afterwards so that daos itself does not use and of the low
+ * number file descriptors.
+ * The DAOS logging uses fnctl calls to force it's FDs to higher numbers to avoid the same problems.
+ * See DAOS-13381 for more details. Returns true on success
+ */
+
+#define IOIL_MIN_FD 10
+
+static bool
+call_daos_init(int fd)
```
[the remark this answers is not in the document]
jolivier23: To avoid the move?

jolivier23:
> I think we did this initially, at the time we were looking at only having dfuse on compute nodes and launching processes via mpirun, and were then getting errors from the mpirun process itself.
> 
> I'm fine to have an option for this if it solves a problem but it seems overkill for the default, init-on-first-touch shouldn't have more overhead over all, and obviously less in some cases.

I can put it under an environment variable if that is preferred.  Let me upload a new patch (sometime later)

jolivier23:
@wiliamhuang I presume it would be safe to do the same for libpil4dfs?

I also noticed that libpil4dfs doesn't wrap daos_init to avoid the issue with applications expecting contiguous fd's.   Is that not an issue at all for libpil4dfs?

docs/user/filesystem.md
```diff
@@ -799,6 +799,15 @@ $ LD_PRELOAD=/usr/lib64/libioil.so dd if=/dev/zero of=./bar bs=1G count=20
 Alternatively, it's possible to simply link the interception library into the application
 at compile time with the `-lioil` flag.
 
+By default, daos is only initialized in the interception library if the application calls
```
ashleypittman, line 802: It's not if/when it calls open, it's when it detects the first file descriptor backed by dfuse. It does the ioctl after every open and then calls daos_init only if it gets a reply from the ioctl.

wiliamhuang:
My only concern is that function "daos_init()" is very expensive to call (~200 ms or even longer). All processes will call it with this patch even they are not going to access the files on daos.

wiliamhuang:
> @wiliamhuang I presume it would be safe to do the same for libpil4dfs?
> 
> I also noticed that libpil4dfs doesn't wrap daos_init to avoid the issue with applications expecting contiguous fd's. Is that not an issue at all for libpil4dfs? See the definition for call_daos_init and the comment before that definition.

@jolivier23 daos_init() was called in the constructor function too in my original implementation of libpil4dfs. Considering this call is very expensive, I changed it to defer this call until the first time accessing files on daos as in libioil.

jolivier23:
How hard would it be to split out the code that sets environment variables from the test of daos_init?  If we could call something like daos_env_init here that would address the issue in a way that doesn't cause any extra unnecessary initialization

ashleypittman:
incidentally, this has conflicts after the weekend.

jolivier23:
Do mercury and libfabric use setenv?

PR #13054, opened 2023-09-14
DAOS-14347 control: Enable setting of port numbers in config generate

tanabarr:
Allow config generate to produce output that contains custom port
numbers for both the control server and for fabric interfaces rather
than just using hardcoded values.  This new behaviour will work in
both dmg and daos_server config generate commands.

Supplying a custom port number to all of the comma-separated access
point addresses in the commandline option will set the (control server)
port parameter in the output server config file
e.g. --access-points=bob:10002,jeff:10002,ben:10002.

Fabric interface port numbers can be set by specifying a comma
separated list in a new commandline option. There must be enough
values to supply all generated engines e.g. --fabric-ports=12460,13460.

Also:
- Mock API backend call to improve dmg unit test coverage.
- Reduce excessive parameters by passing request.
- Only ERROR and more severe log levels are logged to stderr during
  config generate command invocation unless --debug option specified.
- Allow tmpfs non-MD-on-SSD mode (ephemeral) config generation by
  keying MD-on-SSD mode off the control_metadata cli option.
- Update docs to reflect changes.

Features: control
Required-githooks: true

### Gatekeeper:


src/control/lib/control/auto.go
```diff
@@ -84,6 +96,47 @@ type (
...
+		*Alias
+	}{
+		Alias: (*Alias)(cgr),
+	}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	cgr.AccessPoints = strings.Split(aux.AccessPoints, ",")
```
kjacque, line 115: Do these need to be verified similarly to the other parameters below?

src/control/lib/control/auto.go
```diff
@@ -973,7 +1030,13 @@ func getBdevTiers(log logging.Logger, scmCls storage.Class, ssds *hardware.PCIAd
 
 type newEngineCfgFn func(int) *engine.Config
 
-func genEngineConfigs(log logging.Logger, scmOnly bool, newEngineCfg newEngineCfgFn, nodeSet []int, nd *networkDetails, sd *storageDetails) ([]*engine.Config, error) {
+func genEngineConfigs(req ConfGenerateReq, newEngineCfg newEngineCfgFn, nodeSet []int, nd *networkDetails, sd *storageDetails) ([]*engine.Config, error) {
+	nrFabPorts := len(req.FabricPorts)
+	if nrFabPorts > 0 && nrFabPorts < len(nodeSet) {
+		return nil, errors.Errorf("insufficient fabric ports for nr engines, want %d got %d",
+			len(nodeSet), nrFabPorts)
+	}
```
kjacque, line 1038: I think we should accept the case where the user passes in a single port that will be used for both engines, if the engines are on different interfaces.

kjacque (changes requested):

knard-intel (commented):
Mostly OK for me.
Just have the same concerns as @kjacque regarding the needs to have different ports if the interface are not the same.

src/control/lib/control/auto.go
tanabarr: apparently this should not be allowed according to @Michael-Hennecke

Michael-Hennecke: we have seen failures many times when the port numbers of the two engine interfaces are identical, even when the fabric interfaces are different. i do not understand the details _why_ that is, but i can attest _that_ it happens. 

and more pressing than the changes here would be a clear guidance how big the _range_ of ports used by an engine actually is. saying "A difference of ~1000 is recommended." isn't really a satisfactory answer because we do not explain how we arrive at that estimate. on which other config parameters or cluster size metrics does this depend? rather than a single port number, i would actually prefer an interval of port numbers in the fabric_iface_port statement, and then the code can check how many ports it actually needs and fail if the interval is too small.

tanabarr: I have requested clarification on this guidance and the rational behind it from @frostedcmos . The ask for actual port number validation and interval range in the config is nontrivial and the work would require scheduling so please create a ticket if you want this feature in 2.6 (or another) release @Michael-Hennecke . In terms of what changes you require to this PR specifically is it an expansion of the port number difference statement to include some detail on how to calculate the interval? or should I just remove that statement for the moment and follow up with a subsequent PR after I have found some more definitive explanation?

src/control/server/config/faults.go
```diff
@@ -118,11 +118,11 @@ var (
...
 
-func FaultConfigDuplicateFabric(curIdx, seenIdx int) *fault.Fault {
+func FaultConfigDuplicateFabricPort(curIdx, seenIdx int) *fault.Fault {
 	return serverConfigFault(
-		code.ServerConfigDuplicateFabric,
-		fmt.Sprintf("the fabric configuration in I/O Engine %d is a duplicate of I/O Engine %d", curIdx, seenIdx),
-		"ensure that each I/O Engine has a unique combination of provider,fabric_iface,fabric_iface_port and restart",
+		code.ServerConfigDuplicateFabricPort,
+		fmt.Sprintf("the fabric port in I/O Engine %d is a duplicate of I/O Engine %d", curIdx, seenIdx),
+		"ensure that each I/O Engine has a unique fabric_iface_port (difference of ~1000 is recommended) and restart",
```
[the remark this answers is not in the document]
tanabarr: apparently not as per @Michael-Hennecke 's guidance

utils/config/daos_server.yml
```diff
@@ -348,7 +348,10 @@
 #
 #  # Specify the fabric network interface and interface port that will
 #  # be used by this engine. The fabric_iface_port must be different
-#  # for each engine on a DAOS server.
+#  # for each engine on a DAOS server. A difference of ~1000 is recommended.
+#  # Optionally specify the fabric network interface port that will be used
+#  # by this engine but please only if you have a specific need, this will
+#  # normally be chosen automatically.
```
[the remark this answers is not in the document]
tanabarr: Apparently the port numbers need to be unique even if on different interfaces so I won't change that wording. I have simplified the comment  and with the pool specific information to directly above the port definition as suggested.

utils/config/daos_server.yml
```diff
@@ -519,14 +522,15 @@
 #
 #  bypass_health_chk: true
 #
-#  # Use specific network interface.
-#  # Specify the fabric network interface that will be used by this engine.
+#  # Specify the fabric network interface and interface port that will
+#  # be used by this engine. The fabric_iface_port must be different
+#  # for each engine on a DAOS server. A difference of ~1000 is recommended.
```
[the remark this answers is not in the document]
tanabarr: as above

src/control/lib/control/auto.go
Michael-Hennecke: no change requested from me as this point - yes let's move the "interval" topic to a separate ticket.

Michael-Hennecke (approved):

knard-intel (dismissed):
Thanks for the clarification of the Interface and port numbers

src/control/lib/control/auto.go
tanabarr: I've reverted port validation changes in this PR, @Michael-Hennecke is going to discuss the issue of identical port numbers across engines with different interfaces with @frostedcmos and we will address those changes separately.

kjacque: OK with me. Thanks for the explanation.

kjacque (approved):

knard-intel (approved):
LGTM

docs/admin/deployment.md
```diff
@@ -216,24 +223,29 @@ Each generated engine section will specify a SCM storage tier (PMem or tmpfs) in
...
 - `--net-provider` selects a specific network provider , this can be set to any provider string
 supported on the hosts e.g. ofi+tcp. This option will alter the command's behavior and should only
 be used when normal command operation is not sufficient. Note that specifying a provider will
 prevent the command from selecting the best available.
 
 - `--use-tmpfs-scm` will produce a config specifying RAM-disk (tmpfs) devices in the first (SCM)
-storage tier. The RAM-disk sizes will be calculated based on using 75% of the host's total
-memory (as reported by `/proc/meminfo`).
+storage tier. The RAM-disk sizes will be calculated based on the host's total memory (as reported
+by `/proc/meminfo`). An MD-on-SSD config will be generated if this flag is set.
```
mjmac, line 242: Why does this flag enable MD on SSD? I would think that should be keyed off of the `--control-metadata-path` flag instead, as that's what the server logic does.

tanabarr: I could make that change, currently the behaviour is as described. Would you like to do that in this PR?

mjmac: @kjacque: Is my understanding correct re: the server logic here? If we've got an inconsistency then it seems like it should be resolved...

tanabarr: what's the inconsistency? the config generate flag is explicit and indicates that the user wants to generate a tmpfs config rather than use any present pmem. IMO that's valid and in this case explicit is better than implicit. I'm happy to change if the consensus is strong that we should automate that selection based on the presence of another option but I'm not seeing any obvious inconsistency.

mjmac: > what's the inconsistency?

If I am understanding correctly, based on the documentation here, using the `--use-tmpfs-scm` flag enables MD-on-SSD mode. Whereas when the configuration is generated manually, MD-on-SSD mode is only enabled when the `control_metadata:` parameter is set. I'm just saying that it's inconsistent (and confusing) to have two different ways of enabling MD-on-SSD. Furthermore, doing it this way would mean that there's no way to use tmpfs for emulated pmem in non-MD-on-SSD mode anymore, at least via generated configuration.

tanabarr: ok, so I will remove the --use-tmpfs-scm flag and enable tmpfs-MD-on-SSD config generation when control_metadata commandline option is supplied. does that sound like what you want?

mjmac: I'm not objecting to the `--use-tmpfs-scm` flag itself. I think there's still a use case for emulating pmem in non-MD-on-SSD (aka legacy) mode. I just wanted to point out that the code as implemented currently uses the presence of a populated `control_metadata:` section in the server config to toggle between the modes. So it didn't make sense to me to tie the mode toggle to the use of tmpfs or not.

tanabarr: I will update to allow for ephemeral config and key MD-on-SSD off control_metadata option.

tanabarr: done, please could you review @mjmac

src/control/server/storage/config.go
```diff
@@ -441,16 +441,20 @@ func (tcs TierConfigs) validateBdevRoles() error {
...
 
 	// Require tier-0 to be a SCM tier.
 	if len(scs) != 1 || scs[0].Tier != 0 {
 		return errors.New("first storage tier is not scm")
 	}
-	// No roles should be assigned if scm tier is DCPM.
+	// MD-on-SSD roles can not be assigned if scm tier is DCPM.
 	if scs[0].Class == ClassDcpm {
-		return nil
+		return errors.New("external metadata path invalid with DCPM scom class")
```
knard-intel, line 457: Nit
```suggestion
		return errors.New("external metadata path invalid with DCPM scm class")
```

docs/admin/deployment.md
```diff
@@ -365,10 +370,15 @@ SCM format required on instance 0
...
+startup in the event that externally triggered format is not desirable. When using this option
+the storage format will occur on `daos_server` startup without any user intervention assuming the
+supplied config file is valid.
+
+To manually format the storage and start the engine processes, we run the following on a separate
 terminal window and verify that engine processes (ranks) have registered with the system.
-Note the subsequent system query command may not show ranks started immediately after the
-storage format command returns so leave a short period before invoking.
+Note the subsequent system query command may not show ranks started immediately after the storage
+format command returns so it is recommended to leave a short delay before invoking.
```
knard-intel, line 381: Nit, a minimal value or a rough value of delay could be helpful for the end user.

knard-intel (approved):
LGTM, just some minor remarks if you have to repush

docs/admin/deployment.md
tanabarr: fixed

src/control/server/storage/config.go
tanabarr: fixed

tanabarr:
manual verification of PR:
[x] supply custom port numbers for 2 fabric interfaces
[x] create pmem/dcpm config
[x] create tmpfs/ram config (-t option)
[x] create md-on-ssd config (-t and -m options)
[x] perform all the above with daos_server and dmg tools

knard-intel (approved):

src/control/cmd/daos_server/auto_test.go
```diff
@@ -199,31 +211,46 @@ func TestDaosServer_Auto_confGen(t *testing.T) {
...
 	}
 	// Nr hugepages expected with 18+1 (extra MD-on-SSD sys-xstream) targets * 2 engines * 512
 	// pages-per-target.
-	tmpfsNrHugepages := 19 * 2 * 512
-	tmpfsHugeMemGiB := (humanize.MiByte * 2 * tmpfsNrHugepages) / humanize.GiByte
-	tmpfsEngRsvdGiB := 2 /* calculated based on 18 targets */
-	tmpfsSysRsvdGiB := storage.DefaultSysMemRsvd / humanize.GiByte
-	tmpfsRamdiskGiB := storage.MinRamdiskMem / humanize.GiByte
+	mdonssdNrHugepages := 19 * 2 * 512
+	mdonssdHugeMemGiB := (humanize.MiByte * 2 * mdonssdNrHugepages) / humanize.GiByte
```
kjacque, line 239: mdonssd should be bumpy case: mdOnSSD

kjacque (approved):
New changes LGTM. One tiny comment, not needed to land.

src/control/cmd/daos_server/auto_test.go
tanabarr: will address if I have to rePush

tanabarr:
gatekeeper, has 2 approvals but don't know why Kris's green tick isn't appearing in the top right of the page

Codespell and pylint issues are related to code unchanged in this PR

docs/admin/deployment.md
```diff
@@ -123,6 +123,7 @@ Application Options:
       --allow-proxy                         Allow proxy configuration via environment
   -o, --config=                             Server config file path
   -b, --debug                               Enable debug output
+  -j, --json                                enable JSON output
```
ashleypittman, line 126:
```suggestion
  -j, --json                                Enable JSON output
```

src/control/cmd/daos_server/auto_test.go
ashleypittman: What's the difference between bumpycase and camelcase?

tanabarr: they are the same

docs/admin/deployment.md
tanabarr: Will fix in a follow-up PR

src/control/cmd/daos_server/auto_test.go
tanabarr: Will fix in a follow-up PR

PR #13055, opened 2023-09-14
DAOS-623 bio: Do not access void data in bio code.

ashleypittman:
Rather than deference the void * in iterate_biov pass
in the index to the callback function and let that do
any assignment.  This avoids a compiler warning where
it was detecting a possible buffer overflow based on
incomplete knowedge.


ashleypittman:
This isn't exactly the same as before and I don't claim to have looked into what the code is doing however this would be a more conventional way of handling the additional assignments.

src/bio/bio_buffer.c
```diff
@@ -446,14 +446,21 @@ struct bio_copy_args {
...
 	void			*addr = bio_iov2req_buf(biov);
 	ssize_t			 size = bio_iov2req_len(biov);
 	uint16_t		 media = bio_iov2media(biov);
 
+	D_ASSERT(idx < arg->ca_sgl_cnt);
+	arg->ca_sgl_idx = idx;
+	arg->ca_iov_idx = 0;
+	arg->ca_iov_off = 0;
+	if (biod->bd_type == BIO_IOD_TYPE_FETCH)
+		arg->ca_sgls[idx].sg_nr_out = 0;
```
NiuYawei, line 462: This per SGL stuff should only be initialized once (when switching to a new SGL), but the callback is called for each IOV, so it's not correct to do such initialization work in every callback.

So maybe the callback could be modified to accept both "sgl_idx" and "iov_idx" as parameter, and only perform the per SGL initialization work when "iov_idx == 0"?

NiuYawei (changes requested):

src/bio/bio_buffer.c
ashleypittman: I'm not sure even that would work as what about if bs_nr_out is zero?  The callback isn't invoked at all in that case?

I wonder about having two callbacks, one for sgls and one for iovs?

jolivier23 (commented):
I'll let you sort it out with NIu before a more thorough review.  The title should be "Do not" ?

src/bio/bio_buffer.c
```diff
@@ -536,37 +536,51 @@ copy_one(struct bio_desc *biod, struct bio_iov *biov, void *data)
...
+	int rc                                          = 0;
+
+	if (data != NULL) {
+		if (cb_fn == copy_one)
+			prep_fn = copy_one_setup;
+		else if (cb_fn == bulk_map_one)
+			prep_fn = map_one_setup;
+	}
+
+	for (int i = 0; i < biod->bd_sgl_cnt; i++) {
```
NiuYawei, line 574: This is the new coding style that declares a variable here?

NiuYawei (approved):

src/bio/bio_buffer.c
ashleypittman: it's not really new, the coding standard has always said use the smallest scope that you can for variables.  Including this wasn't entirely deliberate, I just happened to have re-read the document the day before writing this and wasn't sure if it would work or not.

jolivier23 (approved):

PR #13061, opened 2023-09-15
DAOS-14366 client: intercept _exit() in libioil and libpil4dfs

wiliamhuang:
If applications call "_exit()", destructor function is not invoked. We do need to intercept "_exit()" and call destructor functions to finish necessary cleanup work before calling glibc “_exit()“.

Required-githooks: true

### Gatekeeper:


wangshilong (dismissed):

wiliamhuang:
@mchaarawi Could you please review when you have time? Thank you!

src/client/dfuse/il/int_posix.c
```diff
@@ -2886,5 +2886,12 @@ dfuse_get_bypass_status(int fd)
 	return rc;
 }
 
+DFUSE_PUBLIC void __attribute__ ((__noreturn__))
+dfuse__exit(int rc)
+{
+	ioil_fini();
```
mchaarawi, line 2892: have you tested this properly? this can disconnect containers and pools that are on the main process i believe.
If I understand this correctly, this can call pool_disconnect for example on a pool, close it on the server, and when the main thread tries to access the pool, the server complains that the pool is not connected. but im not 100% sure on this.

unfortunately we do not have a lot of IL tests in CI with things like FIO and others.

mchaarawi (commented):
im not sure this patch is correct.
in any case, this should be repushed also with:
Test-tag: pr il test_daos_dfuse_unit_ioil test_dfuse_daos_build_wt_il test_dfuse_mu_perms_ioil test_ior_small

src/client/dfuse/il/int_posix.c
wiliamhuang: https://man7.org/linux/man-pages/man2/exit.2.html
"_exit()" is called to terminate the current process (including all threads). I think it is appropriate to call "ioil_fini()". "pthread_exit()" is the function for thread call to terminate current thread.

mchaarawi: maybe you did not get my point.
you start with 1 process, it calls init, connect to pool A, container B.
fork();
let's say child does nothing and and calls _exit();
child calls ioil_fini() and closes pool handle A and container handle B
now for the client side, those handles are different, but for the server side, they are the same. so those handles are closed on the server.

parent tries to access pool A and gets an error.

wiliamhuang: ah. I see. If this is an issue, we should have same issue if one process (either parent or child process) exits without calling "_exit" since destructor function "ioil_fini" will be invoked.

mchaarawi: no. because of this:
https://github.com/daos-stack/daos/blob/master/src/client/dfuse/il/int_posix.c#L369

your patch is actually a no-op because you also hit that condition and fini from those forked processes are not doing anything :-)

mchaarawi (changes requested):
i tested this locally with fio and libiol and found that with fork, the _exit does call fini which doesn't do anything because of this check:
https://github.com/daos-stack/daos/blob/master/src/client/dfuse/il/int_posix.c#L369

this requires more thought / work
even if we remove that check, the processes will close pool and  container handles that were created by the parent process, which is not supposed to happen.

src/client/dfuse/il/int_posix.c
wiliamhuang: Good to know. Thank you!
If the parent process calls "_exit()", the destructor function will not be called.
ok. I will revise the PR to only call daos_eq_destroy() to clean up EQs.

mchaarawi: that will also not work. because for threads, the eq pool is now shared and even if a thread _exit() the eq it had created will be destroyed by the parent when it exits.
so you will need to be able differentiate a new thread calling _exit(); and a child process from fork() calling _exit() and destroy only on the latter.

ashleypittman:
This is obviously a tricky piece of code to get right, and whilst the DaosBuild test is picky good at finding issues it's not a good way of testing this.

I think the thing to do would be to add specific test cases for all the scenarios we know about, probably to https://github.com/daos-stack/daos/blob/master/src/tests/ftest/daos_test/dfuse.py. It may be beneficial to do that in a different PR and possibly also a different ticket.

wiliamhuang:
> This is obviously a tricky piece of code to get right, and whilst the DaosBuild test is picky and good at finding issues it's not a good way of testing this.
> 
> I think the thing to do would be to add specific test cases for all the scenarios we know about, probably to https://github.com/daos-stack/daos/blob/master/src/tests/ftest/daos_test/dfuse.py. It may be beneficial to do that in a different PR and possibly also a different ticket.

Good idea! I will add tests in a different PR as you suggested. Thank you!

wiliamhuang:
All CI tests passed except a known issue, https://daosio.atlassian.net/browse/DAOS-14450
I will add more tests in a different PR later.

src/client/dfuse/il/int_posix.c
```diff
@@ -364,39 +364,7 @@ ioil_fini(void)
...
-			 */
-			rc = ioil_shrink_cont(cont, false, true);
-			if (rc == -DER_NOMEM)
-				ioil_shrink_cont(cont, false, true);
-		}
-		rc = ioil_shrink_pool(pool);
-		if (rc == -DER_NOMEM)
-			ioil_shrink_pool(pool);
-	}
+	pid_t             pid;
```
mchaarawi, line 367: unfortunately this is still wrong. you are destroying all EQs for threads that are now in a shared pool and can be in-use by other threads!
Im talking about the no-fork case and just using threads.

mchaarawi (changes requested):

src/client/dfuse/il/int_posix.c
wiliamhuang: Thank you! You are right. It would need more work.

wiliamhuang:
Was included in https://github.com/daos-stack/daos/pull/13337 which was landed.

PR #13064, opened 2023-09-18
DAOS-13565 vos: obj flatten basic header file

liuxuezhao:
With some basic format of flattened object.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/vos/vos_layout.h
+++ b/src/vos/vos_layout.h
@@ -363,7 +363,7 @@ struct vos_irec_df {
  */
 struct vos_obj_df {
 	daos_unit_oid_t			vo_id;
-	/** The latest sync epoch */
+	/** The latest sync epoch, is DAOS_EPOCH_MAX for flattened object */
 	daos_epoch_t			vo_sync;
 	/** Offset of known existing dkey */
 	umem_off_t			vo_known_dkey;
@@ -371,8 +371,12 @@ struct vos_obj_df {
 	daos_epoch_t			vo_max_write;
 	/** Incarnation log for the object */
 	struct ilog_df			vo_ilog;
-	/** VOS dkey btree root */
-	struct btr_root			vo_tree;
+	union {
+		/** VOS dkey btree root for hierarchical object */
+		struct btr_root		vo_tree;
+		/** external address for flattened object */
+		bio_addr_t		vo_addr;
+	};
 };
 
 #endif
--- a/src/vos/vos_obj.h
+++ b/src/vos/vos_obj.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -21,6 +21,7 @@
 #include "vos_layout.h"
 #include "vos_ilog.h"
 #include "vos_ts.h"
+#include "vos_obj_flat.h"
 
 #define LRU_CACHE_BITS 16
 
@@ -47,6 +48,8 @@ struct vos_object {
 	struct vos_obj_df		*obj_df;
 	/** backref to container */
 	struct vos_container		*obj_cont;
+	/** flattened object in memory structure */
+	struct vos_obj_flat		*obj_flat;
 	/** nobody should access this object */
 	bool				obj_zombie;
 	/** Object is in discard */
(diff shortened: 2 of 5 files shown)


src/vos/vos_obj_flat.h
```diff
@@ -0,0 +1,128 @@
...
+	uint16_t	ofd_dkey_nr;
+	/** number of vof_item_df in payload */
+	uint16_t	ofd_item_nr;
+	/** number of bytes of payload */
+	uint32_t	ofd_len;
+	/**
+	 * payload's layout -
+	 * struct vof_item_df[ofd_item_nr]
+	 * followed by each key and value's content and checksums.
+	 */
```
gnailzenh, line 96: do we want to have checksum for each individual item? I think one checksum per flattend object is fine, we can verify it while loading. It can save more space and make the code simpler?

src/vos/vos_obj.h
```diff
@@ -238,4 +241,16 @@ vos_obj_discard_hold(struct daos_lru_cache *occ, struct vos_container *cont, dao
...
+static inline bool
+vos_obj_is_empty(struct vos_object *obj)
+{
+	return !obj->obj_df || obj->obj_df->vo_tree.tr_class == 0;
+}
+
+static inline bool
+vos_obj_flattened(struct vos_object *obj)
+{
+	return obj->obj_df && obj->obj_df->vo_sync == DAOS_EPOCH_MAX;
```
gnailzenh, line 253: we don't have use case of epoch_max in vos_object_df today?

src/vos/vos_obj_flat.h
liuxuezhao: now in EV tree it stores csums for each entry (struct evt_entry::en_csum), and for single-value/dkey/akey all with its csums stored. When fetch, those csums can be read back and compare/transfer back to client to compare.
If for flattened obj we changed it to only store a whole csums for the obj_flat_df and verify it while loading, it means -
1) we need to calculate related csums when fetch the data every time to transfer back to client, 
2) seems can broke the end-to-end checksum assumption as the data-corruption between obj loading and csum re-calculate cannot be detected?

src/vos/vos_obj.h
liuxuezhao: now the "vo_sync" only updated by "vos_dtx_mark_sync" and the reasonable value should be [0, current HLC/epoch], the EPOCH_MAX seems impossible to be used in normal case.
@Nasf-Fan could you please confirm if it is reasonable to use EPOCH_MAX to differentiate flattened obj? thanks

src/vos/vos_layout.h
```diff
@@ -363,16 +363,20 @@ struct vos_irec_df {
...
 	daos_epoch_t			vo_max_write;
 	/** Incarnation log for the object */
 	struct ilog_df			vo_ilog;
-	/** VOS dkey btree root */
-	struct btr_root			vo_tree;
+	union {
+		/** VOS dkey btree root for hierarchical object */
+		struct btr_root		vo_tree;
+		/** external address for flattened object */
+		bio_addr_t		vo_addr;
```
NiuYawei, line 378: D_CASSERT(sizeof(bio_addr_t) <= sizeof(struct bio_root));

liuxuezhao: sure, will add when refresh this patch or in next patch, thx

src/vos/vos_obj.h
liuxuezhao: just checked offline with @Nasf-Fan , we don't have use case of EPOCH_MAX in vos_obj_df::vo_sync today.

Nasf-Fan: Yes, sync logic will not use "EPOCH_MAX".

src/vos/vos_obj_flat.h
```diff
@@ -0,0 +1,128 @@
...
+			uint32_t		ofk_singv_gsize;
+			union {
+				/** offset of the single-value */
+				uint32_t		ofk_singv_off;
+				/** inline data (for short single-value and no csum) */
+				uint8_t			ofk_singv[VOF_SINGV_INLINE_SZ];
+			};
+		};
+	};
+} __attribute__ ((aligned(4)));
```
jolivier23, line 69: does this mean we always align to 4 bytes even if it's a 64-bit value in middle of struct?  Do we ever plan to access these directly?

liuxuezhao: I was try to put it to 20B but just realized that the size of record should be stored for each array ext, so will change it to 24B, so can avoid that problem.

src/vos/vos_obj_flat.h
jolivier23: I presume you would just need to minic the behavior.  Like if you had one checksum for the whole flat object but then recalculated the checksum for each checksum chunk you return to client

src/vos/vos_obj_flat.h
```diff
@@ -0,0 +1,131 @@
...
+	 */
+	uint8_t		ofd_payload[0];
+};
+
+struct vof_item;
+
+/** data structure to help binary search */
+struct vof_sorter {
+	/** type (dkey/akey/sing-value/array-ext) */
+	uint16_t		  vs_type;
```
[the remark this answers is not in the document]
liuxuezhao: a flat object can with multiple dkeys, and one dkey can with multiple akey, and same that one akey can with multiple value, that is same as current object. @jolivier23

PR #13065, opened 2023-09-18
DAOS-14511 dfuse: Instruct fuse to invalidate dentries after timeout.

ashleypittman:
Have dfuse instruct the kernel to invalidate old cache-expired inodes.
Create a extra data structure to keep inodes in time order, when inodes
are looked up move them to the end of the list and have a thread checking
the start of the list for expired entries.  When entries are expired then after
some additional time call inval_dentry() on the last known dentry which will
cause the kernel to call forget() on the inode and all it's children.

This means that over time as timeouts expire the size of the working set will
decrease as will memory consumption of both dfuse and the kernel and as
a result dfuse will close open containers and detach from pools.


src/client/dfuse/inval.c
```diff
@@ -0,0 +1,423 @@
...
+ *  ie_close() which is called from forget and some failure paths in readdir()
+ *  lookup() to move entries to the end of this list.
+ *  de_run() to pull items from the front of the list.
+ *
+ * Wakeup: The invalidation thread is woken up when:
+ *  dfuse is exiting.
+ *  something is added to an empty list.
+ *  after a timeout.
+ * Timeouts are chosen based on the entries still on any list, dfuse will sleep as long as it can
+ * but at least 2 seconds and at most 60.
```
daltonbohning, line 69: Maybe I missed it, but I don't see the "at most 60"

src/client/dfuse/inval.c
```diff
@@ -0,0 +1,423 @@
...
+ */
+static bool
+ival_loop(int *sleep_time)
+{
+	struct dfuse_time_entry *dte;
+	struct inode_core        ic[EVICT_COUNT] = {};
+	int                      idx             = 0;
+	double                   sleep           = (60 * 1) - 1;
+
+	D_MUTEX_LOCK(&ival_lock);
```
daltonbohning, line 126: Is there potential for this loop to take a long time and hold this lock for too long? E.g. in the case of needing to evict millions of entries

daltonbohning: And on that note, consider an extreme case
1. Evict time set to 3 seconds
2. Entry A has 0.1 seconds remaining, so not evicted
3. Loop continues for X seconds
4. Sleep for 2 seconds
5. Start next loop
6. Entry A expired 2 + X seconds ago

src/client/dfuse/dfuse_core.c
```diff
@@ -723,6 +723,10 @@ dfuse_cont_open_by_label(struct dfuse_info *dfuse_info, struct dfuse_pool *dfp,
 		} else if (rc != 0) {
 			D_GOTO(err_close, rc);
 		}
+
+		rc = ival_add_cont_buckets(dfc);
+		if (rc)
+			goto err_close;
```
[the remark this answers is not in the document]
ashleypittman: Honestly, D_GOTO doesn't add anything other than the ability to not use brackets if performing one additional statement before the goto, if it's not doing that then I don't think we should be using it.

src/client/dfuse/inval.c
ashleypittman: Line 124.  It starts at 59, then decreases based on any inode with a lower timeout, then rounds up.

src/client/dfuse/inval.c
ashleypittman: L101-104 covers this case, 8 entries are copied onto the stack then the lock is dropped.

Your second comment is covered by L70-71, there is no harm in keeping something longer than required, we already keep something for "X + <something> + <sample time>" intentionally so decreasing the sample time at the cost of CPU usage and lock contention is not desireable.

src/client/dfuse/dfuse_core.c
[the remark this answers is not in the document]
daltonbohning: I don't disagree, but from a consistency perspective, a few lines above this does use D_GOTO

src/client/dfuse/inval.c
daltonbohning: Ah, I missed the `EVICT_COUNT`

src/client/dfuse/dfuse_core.c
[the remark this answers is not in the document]
ashleypittman: Depends if you value consistency or trending in the right direction.  L260 of inval.c however should use D_GOTO according to my logic here.

jolivier23 (approved):

mchaarawi (approved):
no request changes, just some generic questions or feedback that can be addressed after.

src/client/dfuse/inval.c
```diff
@@ -0,0 +1,438 @@
...
+	d_list_for_each_entry(dte, &ival_data.time_entry_list, dte_list) {
+		struct dfuse_inode_entry *inode, *inodep;
+
+		DFUSE_TRA_DEBUG(dte, "Iterating for timeout %lf", dte->time);
+
+		d_list_for_each_entry_safe(inode, inodep, &dte->inode_list, ie_evict_entry) {
+			double timeout;
+
+			if (dfuse_dentry_get_valid(inode, dte->time, &timeout)) {
+				DFUSE_TRA_DEBUG(inode, "Keeping left %lf " DF_DE, timeout,
```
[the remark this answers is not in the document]
ashleypittman: I think this will log at most once per timeout value per thread wakeup which itself is a minimum of two seconds.  Debug logs getting out of control is something than can happen but I don't think this one will be particularly noisy.

src/client/dfuse/dfuse.h
```diff
@@ -782,27 +782,28 @@ struct fuse_lowlevel_ops dfuse_ops;
...
+#define DFUSE_REPLY_CREATE(inode, req, entry, fi)                                                  \
 	do {                                                                                       \
 		int __rc;                                                                          \
-		DFUSE_TRA_DEBUG(_ie, "Returning create");                                          \
-		_Static_assert(IS_IE(_ie), "Param is not inode entry");                            \
-		(_ie) = NULL;                                                                      \
-		__rc  = fuse_reply_create(req, &entry, fi);                                        \
+		DFUSE_TRA_DEBUG(inode, "Returning create");                                        \
+		ival_update_inode(inode, (entry).entry_timeout);                                   \
+		_Static_assert(IS_IE(inode), "Param is not inode entry");                          \
```
ashleypittman, line 790: Actually I don't think this _Static_assert() is needed now as if the wrong type is passed here then ival_update_inode() above will throw a compiler warning anyway.  Will check and remove in another PR if possible

src/client/dfuse/inval.c
```diff
@@ -0,0 +1,438 @@
...
+ * For kernel based filesystems there is a benefit to keeping all this data in memory as it can
+ * simply be re-validated before use however with fuse + DAOS then re-validate is the same cost
+ * as lookup so there is no benefit in keeping this data around.
+ *
+ * Maintain a number of lists for inode timeouts, for each timeout value keep a list of inodes
+ * that are using that value, when a inode is refreshed by the kernel then move the inode to the end
+ * of the correct list.
+ *
+ * Separately have a thread which periodically will walk each list starting at the front and
+ * invalidate any entries where the timeout has expired.
```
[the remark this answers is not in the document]
ashleypittman: Yes it's a dedicated thread.  I can't remember how it interacts with thread code, will check and ensure the documentation is correct in future work.

src/client/dfuse/inval.c
```diff
@@ -0,0 +1,438 @@
...
+ * As this relates to releasing resources there is no additional benefit in finer grained time
+ * control than this.
+ */
+
+/* Grace period before invalidating directories or non-directories.  Needs to be long enough so that
+ * entries in the working set are invalidated but short enough to be meaningful.
+ * Directories that are used as the cwd for processes can cause problems with being invalidated too
+ * early so use a higher value here.
+ */
+#define INVAL_DIRECTORY_GRACE (60 * 30)
```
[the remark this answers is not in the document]
ashleypittman: There's a wider problem that processes can call stat() on the inode of their current working directory but they do not hold a reference on it so if this inode expires and is invalidated then processes can start seeing errors - I'm hoping this will be fixed in later kernels.  See the associated ticket for more details.

utils/node_local_test.py
```diff
@@ -1535,12 +1536,17 @@ def run_query(self, use_json=False, quiet=False):
...
         """Query and verify the dfuse statistics.
 
         Returns the raw numbers in a dict.
         """
-        cmd = ['filesystem', 'query', self.dir]
+        cmd = ['filesystem', 'query']
+        if qpath:
+            cmd.append(qpath)
+        else:
+            cmd.append(self.dir)
```
daltonbohning, line 1549: Just FYI - no need to change
```suggestion
        cmd = ['filesystem', 'query', qpath or self.dir]
```

daltonbohning (approved):

ashleypittman:
@daos-stack/daos-gatekeeper this is a major 2.6 feature and the only test failures are in the unit tests which are also failing on master currently.

PR #13073, opened 2023-09-20
DAOS-14511 dfuse: Avoid using decref on inodes in dfuse.

ashleypittman:
When the kernel performs a fuse operations it already holds
a reference on the inode, so it is not required for dfuse to
take a second one.  DFuse needs to perform a lookup which
will take a reference but rather than holding this reference
until after replying to the kernel and then calling
decref which will take locks etc simply decrease the ref count
after the lookup using atomics.

This is safe as the reference being dropped is never the last
one so locks are not needed.

Assert that the inode is found for fuse operations, if it isn't
then something serious has gone wrong so assert is the correct
check here.

Overall this decreases hash table operations considerably and
removes both the need for locking and also code-paths where
inodes could potentially be closed.

Test-tag: dfuse
Required-githooks: true


src/client/dfuse/ops/open.c
```diff
@@ -111,26 +107,29 @@ dfuse_cb_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
 		}
 	}
 
-	dfuse_inode_decref(dfuse_info, ie);
 	DFUSE_REPLY_OPEN(oh, req, &fi_out);
 
+	/* At this point no reference is held by dfuse on oh, but oh->doh_readahead->dra_lock is
+	 * held and will block any call to release so oh->doh_readahead is still accessible.
+	 */
+
```
ashleypittman, line 115: DAOS-14637 applies here, needs fixing.

mchaarawi (approved):
patch looks good to me. it seems one of those changes though that it would be very easy to miss some ref inc somewhere

src/client/dfuse/ops/open.c
ashleypittman: I updated to PR to fix this issue.

jolivier23 (approved):

ashleypittman:
@daos-stack/daos-gatekeeper can this be landed?  The PR was run with 'Test-tag: dfuse` as recommended and only modifies dfuse code however this means that no verbs tests are run so the PR requires force-landing.

mchaarawi:
> @daos-stack/daos-gatekeeper can this be landed? The PR was run with 'Test-tag: dfuse` as recommended and only modifies dfuse code however this means that no verbs tests are run so the PR requires force-landing.

the tag should be Feature:dfuse not test-tag
I do think there might be several PR tests that are not tagged with dfuse that might be using dfuse.

mchaarawi:
> > > @daos-stack/daos-gatekeeper can this be landed? The PR was run with 'Test-tag: dfuse` as recommended and only modifies dfuse code however this means that no verbs tests are run so the PR requires force-landing.
> > 
> > 
> > the tag should be Feature:dfuse not test-tag I do think there might be several PR tests that are not tagged with dfuse that might be using dfuse.
> 
> I understand what you're saying but you saying this on the `pil4fdfs` PR yesterday is literally the first time I've head this, I wasn't even aware that the Feature tag was working yet.
> 
> I can certainly update my process (although it'll mean dfuse PRs using more test cycles) but I don't think it would make any difference in this case and re-pushing with even an empty commit would loose the approvals and mean at least another 24 hour turn-around time.

The Feature tag has never stopped working really. it just means it tests PR tests AND any tests that are tagged with what you provide, vs test-tag that test only the tests that are tagged with this.
for example if i write a test that is using dfuse, but im testing things like rebuild and i don't add dfuse tag to those tests, this test is not tested even though the code you are changing in dfuse could be triggered by my test that is not run on your PR, but later might fail bec your PR might have introduce something (just a hypothetical situation).

for your process, you can keep using the test-tag as you are working on the PR, but the final push really needs to test all PR tests (including dfuse tagged ones) using the Feature tag.

For this PR, i have not quantified what PR tests that are skipped that might be using dfuse. maybe @daltonbohning can help with that.

ashleypittman:
> The Feature tag has never stopped working really. it just means it tests PR tests AND any tests that are tagged with what you provide, vs test-tag that test only the tests that are tagged with this. for example if i write a test that is using dfuse, but im testing things like rebuild and i don't add dfuse tag to those tests, this test is not tested even though the code you are changing in dfuse could be triggered by my test that is not run on your PR, but later might fail bec your PR might have introduce something (just a hypothetical situation).

I agree that's the ideal but it's not a process that's ever worked in the past and to the best of my recollection I've never seen anyone even try to actually follow it, it seems unnecessarily strict to start enforcing it without wider discussions on this subject.

mchaarawi:
> > The Feature tag has never stopped working really. it just means it tests PR tests AND any tests that are tagged with what you provide, vs test-tag that test only the tests that are tagged with this. for example if i write a test that is using dfuse, but im testing things like rebuild and i don't add dfuse tag to those tests, this test is not tested even though the code you are changing in dfuse could be triggered by my test that is not run on your PR, but later might fail bec your PR might have introduce something (just a hypothetical situation).
> 
> I agree that's the ideal but it's not a process that's ever worked in the past and to the best of my recollection I've never seen anyone even try to actually follow it, it seems unnecessarily strict to start enforcing it without wider discussions on this subject.

im not sure really where you are coming from. i think this has always been working and being utilized. there are some who do not properly use Feature tag, but gatekeepers usually indicate for some features, that the feature tag needs to be added.
what im enforcing here is that you are skipping PR tagged tests! this is always enforced and there are no discussions to be had on requiring that.

daltonbohning:
> For this PR, i have not quantified what PR tests that are skipped that might be using dfuse. maybe @daltonbohning can help with that.

That's not easy to answer, since some dfuse uses could be buried in test utilities. Ideally, tests should be tagged appropriately, but that hasn't historically been the case.


> re-pushing with even an empty commit would loose the approvals and mean at least another 24 hour turn-around time

If there are no code changes, github doesn't dismiss the approvals.

> the tag should be Feature:dfuse not test-tag

** `Features: dfuse`



FWIW, the auto-tag stuff on working on would have effectively chosen `Features: dfuse` for this PR. `pr` because `src/*` was modified, and `dfuse` because dfuse was modified. If tests were tagged appropriately, just `dfuse` should be okay, but, again, that unfortunately isn't the case. So I'd recommend `Features: dfuse` so `pr` is also ran

PR #13089, opened 2023-09-25
DAOS-13929 common: Remove or silence %s in logging.

ashleypittman:
Improve d_log_check checkking of %s usage.
Add ability to suppress warning through annotation in code.
Add list of struct members which are allowed by default.
Resolve issues, or suppress after manual inspection.

This makes some areas of the code 'clean' but still leaves
approx 700 areas to look at.

Required-githooks: true


src/include/gurt/common.h
```diff
@@ -434,7 +434,6 @@ bool d_rank_list_identical(d_rank_list_t *rank_list1,
 bool d_rank_in_rank_list(d_rank_list_t *rank_list, d_rank_t rank);
 int d_idx_in_rank_list(d_rank_list_t *rank_list, d_rank_t rank, uint32_t *idx);
 int d_rank_list_append(d_rank_list_t *rank_list, d_rank_t rank);
-int d_rank_list_dump(d_rank_list_t *rank_list, d_string_t name, int name_len);
```
mchaarawi, line 437: why are you removing this?
gurt/common.h is a public header file

src/cart/crt_context.c
```diff
@@ -182,8 +182,10 @@ crt_context_create_on_provider(crt_context_t *crt_ctx, const char *provider, boo
 
 	provider_idx = crt_str_to_provider(provider);
 	if (provider_idx == -1) {
-		D_ERROR("Invalid requested provider '%s'\n", provider);
-		return -DER_INVAL;
+		int rc = -DER_INVAL;
+		/* d_log_check: disable=print-string */
+		DL_ERROR(rc, "Invalid requested provider '%s'", provider);
+		return rc;
```
mchaarawi, line 188: can't you just add
/* d_log_check: disable=print-string */
and not change the other parts? assigning rc seems unnecessary here IMO

utils/cq/d_logging_check.py
```diff
@@ -141,6 +141,17 @@ def __next__(self):
 PREFIXES_ALL = PREFIXES.copy()
 PREFIXES_ALL.extend(PREFIXES_NNL)
 
+# Variables that are checked for when %s is using in a logging macro.  The number of %s calls
+# is counted, and the number of times any of the strings below is included is also counted and
+# if the two are the same then no warning is issued.  This allows us to skip checks for cases where
+# strings are of known types, for example uris or CaRT group names, or performing enum->string
+# lookup.
+ALLOWED_VARS = ['strerror', 'db_file', 'ii_group_name', 'cg_grpid', 'agg_op2str', 'gc_name',
```
jolivier23, line 149: so what is the purpose of checking variable names?  is this to avoid PII leak?

ashleypittman: > is this to avoid PII leak?

Yes.

> so what is the purpose of checking variable names?

Some variable names are known to contain things like cart group name or URIs so these don't need to be manually checked or annotated.  It's a (crude) attempt to reduce the number of lines we need to audit/modify/annotate.

This is really just doing string matching on the whole line so it's kinda dumb but it does work to catch a lot of cases. We also have quite a number of places and ways where either arrays or functions are used to convert enums to strings and there are covered here as well..

PR #13093, opened 2023-09-26
DAOS-14438 mgmt: Improve error reporting on uuid_parse failure.

ashleypittman:
This function returns -1 rather than a daos errno so treat it
as such.
Do not use the result in DF_RC as it will print DER_UNKNOWN.
Put the call in the if statement itself rather than assigning rc.
Use the new logging macros to log the error proplerly.
Do not log the invalid uuid.

Required-githooks: true

### DIFF
--- a/src/mgmt/srv_drpc.c
+++ b/src/mgmt/srv_drpc.c
@@ -471,11 +471,10 @@ ds_mgmt_drpc_pool_create(Drpc__Call *drpc_req, Drpc__Response *drpc_resp)
 			D_GOTO(out, rc = -DER_NOMEM);
 	}
 
-	rc = uuid_parse(req->uuid, pool_uuid);
-	if (rc != 0) {
-		D_ERROR("Unable to parse pool UUID %s: "DF_RC"\n", req->uuid,
-			DP_RC(rc));
-		D_GOTO(out, rc = -DER_INVAL);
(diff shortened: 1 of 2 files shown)


src/mgmt/srv_drpc.c
```diff
@@ -630,11 +628,10 @@ ds_mgmt_drpc_pool_evict(Drpc__Call *drpc_req, Drpc__Response *drpc_resp)
 
 	D_INFO("Received request to evict pool connections %s\n", req->id);
 
-	rc = uuid_parse(req->id, uuid);
-	if (rc != 0) {
-		D_ERROR("Unable to parse pool UUID %s: "DF_RC"\n", req->id,
-			DP_RC(rc));
-		D_GOTO(out, rc = -DER_INVAL);
+	if (uuid_parse(req->id, uuid) != 0) {
```
tanabarr, line 631: given we repeat this with the same error message so frequently, why not create a pool_uuid_parse helper to return INVAL and print the invalid error?

tanabarr (approved):

src/mgmt/srv_drpc.c
ashleypittman: I did wonder that but getting it into a macro to keep the right line number and crucially function name would be tricky, it was a borderline decision though.  Some of the UUIDs are also not for pools

PR #13094, opened 2023-09-26
DAOS-623 cq: Enable declaration-after-statement checking.

ashleypittman:
Add in the compiler flag and fix minor issues.
Disable again for DAV and json code.

Required-githooks: true


jolivier23 (commented):
This is a big patch.  I expected a compilation error due to dav or such but it looks like it's failing in SCons python code.   I presume we can termporarily use #pragma for removing diagnostic in dav code.

ashleypittman:
> This is a big patch. I expected a compilation error due to dav or such but it looks like it's failing in SCons python code. I presume we can termporarily use #pragma for removing diagnostic in dav code.

It's failing because I conditionally set it based on availability but then unconditionally unset it in the DAV scons file, so I think it's failing on systems where the option doesn't exist.

ashleypittman:
@jolivier23 @gnailzenh is this change worth it or should we drop this?

jolivier23 (dismissed):

src/mgmt/srv.c
```diff
@@ -322,9 +322,8 @@ ds_mgmt_hdlr_svc_rip(crt_rpc_t *rpc)
 	 * the yield below is to workaround an ofi err msg at client-side -
 	 * fi_cq_readerr got err: 5(Input/output error) ..
 	 */
-	int i;
 
-	for (i = 0; i < 200; i++) {
+	for (int i = 0; i < 200; i++) {
```
mchaarawi, line 326: should the declaration of i be move to the beginning rather than in the loop?
to be consistent with other changes in this PR like:
https://github.com/daos-stack/daos/pull/13094/files#diff-f733e5ebd07cfd256e6f6cf2f5e934e5b88ed055b1520d558a71ddb0be10d746R600

src/placement/tests/jump_map_place_obj.c
```diff
@@ -535,19 +536,19 @@ jtc_get_layout_rebuild_count(struct jm_test_ctx *ctx)
...
 	D_ASSERT(ctx != NULL);
 	D_ASSERT(ctx->po_map != NULL);
-	const uint32_t total_targets = pool_map_target_nr(ctx->po_map);
+	total_targets = pool_map_target_nr(ctx->po_map);
 
 	D_ALLOC_ARRAY(target_set, total_targets);
 	D_ASSERT(target_set != NULL);
 
-	for (i = 0; i < ctx->layout->ol_nr; i++) {
+	for (int i = 0; i < ctx->layout->ol_nr; i++) {
```
mchaarawi, line 551: i guess this is going the other way now?

mchaarawi (commented):
looks OK to me, but one of those PRs which will conflict with every PR most probably.
it does need a rebase.

src/placement/tests/jump_map_place_obj.c
ashleypittman: I'm backing out this one, there's no other changes needed in this function.

src/placement/tests/place_obj_common.c
```diff
@@ -396,16 +396,14 @@ plt_set_domain_status(uint32_t id, int status, uint32_t *ver,
 	rc = pool_map_find_domain(po_map, level, id, &domain);
 	D_ASSERT(rc == 1);
 
-	int i;
-
-	for (i = 0; i < domain->do_child_nr; i++) {
+	for (int i = 0; i < domain->do_child_nr; i++) {
```
ashleypittman, line 399: This function however does require changes so I've scoped the variable properly.

ashleypittman:
> looks OK to me, but one of those PRs which will conflict with every PR most probably.

It doesn't seem to conflict much but does tend to need updating as new code is added because of the stricter checking.  The real danger is that it'll cause compile errors in master if it's landed and then a PR with a regression lands, this is unlikely, particularly if we stick to the general rule that PRs need to have been merged with master no more than a week before landing and I'm not aware of any steps we could take to mitigate this.

mchaarawi (dismissed):

daltonbohning (dismissed):

PR #13120, opened 2023-10-04
DAOS-14469 gurt: Add D_LOG_FILE_APPEND_RANK env

frostedcmos:
- Add new D_LOG_FILE_APPEND_RANK environment variable
- When set, will cause first crt_rank_self_set() invocation to rename a log file and append rank to it.

Required-githooks: true

### Gatekeeper:


frostedcmos:
verbs testing failure due to https://daosio.atlassian.net/browse/DAOS-14450

src/cart/README.env
```diff
@@ -26,6 +26,13 @@ This file lists the environment variables used in CaRT.
    in D_LOG_FILE. If not defined or setting to zero value will result in this
    feature being disabled.
 
+ . D_LOG_FILE_APPEND_RANK
+   Set this to a non-zero value to create separate log files based upon the
+   process cart rank. When enabled this feature will rename the existing log file
+   and append rank to the filename upon the first invocation of rank setting.
+   Subsequent settings of the rank within the process are treated as noops.
+   As ranks are only set on the server, this is a server-only feature.
```
ashleypittman, line 34: If only the server is setting this then wouldn't it be better to export this as a c function rather than an environment variable?

ashleypittman (changes requested):

src/cart/README.env
frostedcmos: > If only the server is setting this then wouldn't it be better to export this as a c function rather than an environment variable?
not sure how a c function would help here since it will need to be conditionally called anyways based on either a yaml setting or an env, as not every test might want to enable this log renaming at runtime. 
plus this seemed consistent with all other log-related options that we provide (e.g. D_LOG_FILE_APPEND_PID)

ashleypittman: Is this for testing then?  The comment in the code implies that it's something the server would want to use.

What's going to set this env, and why?

frostedcmos: > Is this for testing then? The comment in the code implies that it's something the server would want to use.
> 
> What's going to set this env, and why?

This is to help with debugging on clusters that have lots of servers. We typically get logs in a form of <group_name>.log.<pid>, but then someone has to go through each log and figure out which rank what log corresponds to.  Instead when setting this envariable one can get rank automatically attached to the filename once the rank is determined.

I've included the example of how log files look like when running with this set.

jgmoore-or (approved):

src/cart/README.env
ashleypittman: So why make this configurable via an env?  Shouldn't the server just set the log file automatically when it knows the pid?

frostedcmos: Some people might not want this and instead would want the log file to be deterministic and be exactly how they specified it in D_LOG_FILE. 
This is no different than any other option we expose, such as whether to attach PID to the log or not (controlled via D_LOG_FILE_ATTACH_PID env), or what to include in logs themselves.

frostedcmos: Ideally we would just support special characters as part of D_LOG_FILE e.g. D_LOG_FILE=~/my_log.%h.%r     or some such to expand into, for example, %h = hostname, %r = rank, etc... but this would be much bigger change.

jolivier23 (approved):

src/cart/README.env
ashleypittman: I can see an env makes sense for cart in isolation but from a daos perspective what we're saying is that there's a server option that has to be set in a config file, and if that option is set then the daos engine will communicate with itself, via an environment variable, that it wasn't this feature enabled.

ashleypittman (commented):
I think I see what you're trying to achieve though.

Should there be some ftest or example server config file updates to go with this?

ashleypittman (approved):

src/tests/ftest/launch.py
```diff
@@ -1083,6 +1083,7 @@ def _set_test_environment(self, servers, clients, list_tests, provider, insecure
                     os.environ["DAOS_TEST_SHARED_DIR"], "daos_test", "apps")
             os.environ["D_LOG_FILE"] = os.path.join(os.environ["DAOS_TEST_LOG_DIR"], "daos.log")
             os.environ["D_LOG_FILE_APPEND_PID"] = "1"
+            os.environ["D_LOG_FILE_APPEND_RANK"] = "1"
```
daltonbohning, line 1086: Adding D_LOG_FILE_APPEND_RANK with existing D_LOG_FILE_APPEND_PID is going to double the number of log files, right? I should point out that it's already notoriously difficult to sort through the large number of logs from just D_LOG_FILE_APPEND_PID.

@phender Thoughts here?

frostedcmos: > Adding D_LOG_FILE_APPEND_RANK with existing D_LOG_FILE_APPEND_PID is going to double the number of log files, right? I should point out that it's already notoriously difficult to sort through the large number of logs from just D_LOG_FILE_APPEND_PID.
> 
> @phender Thoughts here?

no, it will not double the number of log files; it will just cause the existing log files to have both pid (due to D_LOG_FILE_APPEND_PID) and rank present in the resultant log filename.

frostedcmos: > I should point out that it's already notoriously difficult to sort through the large number of logs from just D_LOG_FILE_APPEND_PID.

this is kind of the purpose of this feature, to make it easier to sort through the large number of logs:) instead of opening each log and grepping for 'Setting rank to %d' to determine which engine rank what file corresponds to, instead with this option it will be in the filename itself... e.g.  daos_engine.log.54421.rank=12

daltonbohning: Oh, duh. The PID is unique to each rank :) Ignore my noise...
Yes, I think this will help

daltonbohning (commented):
ftest LGTM but not familiar with cart side

jolivier23 (approved):

soumagne (approved):

frostedcmos:
Examples of new logs being generated in CI after this change:
https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-13120/2/artifact/Functional%20on%20EL%208/cart/corpc_five_node.py/cart_logs.wolf-103vm2/

PR #13138, opened 2023-10-07
DAOS-13703 umem: umem cache APIs for phase II

NiuYawei:
Four sets of umem cache APIs will be exported for md-on-ssd phase II:

1. Cache initialization & finalization
   - umem_cache_alloc()
   - umem_cache_free()

2. Cache map, load and pin
   - umem_cache_map();
   - umem_cache_load();
   - umem_cache_pin();
   - umem_cache_unpin();

3. Offset and memory address converting
   - umem_cache_off2ptr();
   - umem_cache_ptr2off();

4. Misc
   - umem_cache_commit();

Test-nvme: auto_md_on_ssd

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/common/mem.c
+++ b/src/common/mem.c
@@ -1692,7 +1692,6 @@ struct umem_page_info {
 	uint64_t pi_io		: 1, /** Page is being flushed/loaded to/from MD-blob */
 		 pi_copying	: 1, /** Page is being copied. Blocks writes. */
 		 pi_mapped	: 1, /** Page is mapped to a MD page */
-		 pi_evictable	: 1, /** Page is evictable */
 		 pi_sys		: 1, /** Page is brought to cache by system internal access */
 		 pi_loaded	: 1; /** Page is loaded */
 	/** Highest transaction ID checkpointed.  This is set before the page is copied. The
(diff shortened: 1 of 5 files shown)


src/common/mem.c
```diff
@@ -1696,182 +1703,454 @@ struct umem_page_info {
...
+/* Convert MD-blob offset to MD page */
+static inline struct umem_page *
+cache_off2page(struct umem_cache *cache, umem_off_t offset)
 {
-	struct umem_cache *cache;
-	struct umem_page_info *pinfo;
-	uint64_t           num_pages;
-	int                rc = 0;
-	int                idx;
+	uint32_t idx = offset >> cache->ca_page_shift;
```
sherintg, line 1728: The allocator will use the first 4k bytes for heap header and this will not be part of the memory bucket. The actual buckets starts at offset 4096. To handle this can we add one more field cache->ca_base_offset which indicates the offset where zone0 starts and change the above id calculation as (offset - cache->ca_base_offset) >> cache->ca_page_shift. Similar change is required at other places where the offset is used for translation.

src/common/mem.c
```diff
@@ -2195,29 +2491,592 @@ umem_cache_checkpoint(struct umem_store *store, umem_cache_wait_cb_t wait_cb, vo
...
+error:
+	if (old_pages)
+		D_FREE(old_pages);
+	return rc;
+}
+
+#define UMEM_PAGES_ON_STACK	16
+
+int
+umem_cache_map(struct umem_store *store, struct umem_cache_range *ranges, int range_nr)
```
sherintg, line 2964: umem_cache_map() can be called for evictable pages as-well. This can be the case till all of the md blob zones are used. This is not a must, but good to have requirement.

sherintg, line 2964: need_evict() function to reserve an addition free page is called only by umem_cache_pin() and that too when new page has to be loaded. I think we can potentially hit scenarios where umem_cache_pin() tries to pin the same evictable page multiple times and the transactions would require allocator to expand the non-evictable heap more than once.  In such cases we end up hitting DER_BUSY right?

sherintg (changes requested):
The design and code looks well thought out. I have few comments/requirements from allocator perspective. Can you pls address the same?

src/common/mem.c
NiuYawei: You mean the this 4k header is part of bucket 0 but would never be accessed by VOS (The offset returned to VOS starts from 4k)? Then I don't see why umem cache needs this tweak for page ID calculation, I think it just need to ensure never load/flush the first 4k of page 0, right?

If bucket 0 starts from 4k (the 4k header isn't part of bucket 0), then allocator should just use the start of bucket 0 as exported LBA 0, did I miss anything?

src/common/mem.c
NiuYawei: The question is that is allocator able to tell if a bucket is evictable or not before it's loaded? I think I've asked this before, and your answer is that allocator needs to load the bucket first, otherwise it won't tell if the bucket is evictable or not.

I "think" allocator should be able to tell if a bucket is evict-able or not (if it put some necessary global information in non-evictable zone). If umem cache is able to tell if a bucket is evict-able by ID (without loading it), it could simplify current implementation a bit.

NiuYawei: I don't quite follow what you said, but I think current issue is that umem_cache_map() could use the last free page, if there is no other umem_cache_pin/load() call to trigger eviction, next umem_cache_map() would hit DER_BUSY.

I think I might have to re-add the umem_cache_reserve() call to ensure one free page before starting transaction.

sherintg (approved):

PR #13143, opened 2023-10-09
DAOS-623 gurt: Add logging of deadlock to mutex locks.

ashleypittman:
When a process is blocked on a lock then periodically log
at debug level to aid debugging.

This is disabled for release builds, and only logs at debug level.

Required-githooks: true


src/include/gurt/common.h
```diff
@@ -329,13 +329,40 @@ d_realpath(const char *path, char *resolved_path) _dalloc_;
...
+
+#define __D_PTHREAD_TIMED(fn, fn2, x)                                                              \
+	({                                                                                         \
+		int _rc;                                                                           \
+		_rc = fn((x));                                                                     \
+		if (_rc == EBUSY) {                                                                \
+			int             delay1 = 0;                                                \
+			int             delay2 = 1;                                                \
+			int             f      = delay1 + delay2;                                  \
+			int             c      = f;                                                \
```
frostedcmos, line 348: am a bit worried when we use variable names like these in macro, due to a potential conflict/scope shadow. i believe we have stuck with _ prefix for those for that reason.

src/include/gurt/common.h
```diff
@@ -329,13 +329,40 @@ d_realpath(const char *path, char *resolved_path) _dalloc_;
...
+			struct timespec _wait  = {};                                               \
+			clock_gettime(CLOCK_REALTIME, &_wait);                                     \
+			_wait.tv_sec += f;                                                         \
+			D_DEBUG(DB_MEM, #fn "(%p) held, waiting", x);                              \
+			while ((_rc = fn2((x), &_wait)) == ETIMEDOUT) {                            \
+				delay1 = delay2;                                                   \
+				delay2 = f;                                                        \
```
frostedcmos, line 355: why not just call fn2 (timed lock) right away with a 1 second timeout instead of try-locking first? is it to avoid clock_gettime() in a happy path?

src/include/gurt/common.h
ashleypittman: Good point.  The compiler would warn of course but I'll change this.

src/include/gurt/common.h
ashleypittman: Yes, exactly, and avoid any logging overhead.  Uncontended locks should be cheap.  I hadn't thought about having just one function though and it would work just as well so I'll take a look at the code tomorrow to see if that simplifies things.

src/include/gurt/common.h
```diff
@@ -337,29 +337,25 @@ d_realpath(const char *path, char *resolved_path) _dalloc_;
...
-			int             f      = delay1 + delay2;                                  \
-			int             c      = f;                                                \
-			struct timespec _wait  = {};                                               \
+		int             _rc;                                                               \
+		int             _delay1 = 0;                                                       \
+		int             _delay2 = 1;                                                       \
+		int             _f      = _delay1 + _delay2;                                       \
+		int             _c      = _f;                                                      \
+		struct timespec _wait   = {};                                                      \
+		clock_gettime(CLOCK_REALTIME, &_wait);                                             \
```
frostedcmos, line 348: Reading manpage on timed locks i see this:
"Under no  circumstances  shall  the  function  fail  with  a timeout if the lock can be
       acquired immediately. The validity of the abstime parameter need not be
       checked if the lock can be immediately acquired."

Perhaps then in order to avoid clock_gettime() on a happy path, the first time you can call fn() with null wait, and if it fails only then do clock_gettime() and proper time calc.

Having said that, i dont think doing gettime is expensive as it goes through some specially mmapped page instead of doing full system call iirc.

frostedcmos (approved):

src/include/gurt/common.h
ashleypittman: I'm not worried about the cost of clock_gettime(), I just wrote this as I did and hadn't thought through how to simplfy it.  Thank you for the feedback.

jolivier23 (approved):

PR #13175, opened 2023-10-13
DAOS-13563 vos: Implement flat dkey for array and flat KV object types

jolivier23:
Since many use cases, particularly array and KV APIs use a single AKEY, implement an optimization for these object classes to bypass the akey tree altogether.

The interface remains the same with the exception that the akey passed to vos is ignored completely. Iteration adds a fake akey '0' so that various things built on vos_iterate (e.g. rebuild, ec aggregation), will still function the same as before.

Also removed the akey condition from vos.  This
capability is not used and can be replaced entirely using the pipeline API.

Also refactored iterator callbacks to reduce switch statements.  More work can be done here in the future.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/include/daos_obj.h
+++ b/src/include/daos_obj.h
@@ -178,12 +178,12 @@ static inline bool
 daos_is_dkey_uint64_type(enum daos_otype_t type)
 {
 	switch (type) {
+	case DAOS_OT_ARRAY_BYTE:
 	case DAOS_OT_MULTI_UINT64:
 	case DAOS_OT_DKEY_UINT64:
 	case DAOS_OT_KV_UINT64:
 	case DAOS_OT_ARRAY:
 	case DAOS_OT_ARRAY_ATTR:
-	case DAOS_OT_ARRAY_BYTE:
 		return true;
 	default:
 		return false;
@@ -220,9 +220,9 @@ static inline bool
 daos_is_array_type(enum daos_otype_t type)
 {
 	switch (type) {
+	case DAOS_OT_ARRAY_BYTE:
 	case DAOS_OT_ARRAY:
 	case DAOS_OT_ARRAY_ATTR:
-	case DAOS_OT_ARRAY_BYTE:
 		return true;
 	default:
 		return false;
--- a/src/vos/vos_io.c
+++ b/src/vos/vos_io.c
@@ -1188,79 +1188,19 @@ has_uncertainty(const struct vos_io_context *ioc,
 }
 
 static int
-akey_fetch(struct vos_io_context *ioc, daos_handle_t ak_toh)
+fetch_value(struct vos_io_context *ioc, daos_iod_t *iod, daos_handle_t toh,
+	    const daos_epoch_range_t *epr, bool standalone)
 {
-	daos_iod_t		*iod = &ioc->ic_iods[ioc->ic_sgl_at];
-	struct vos_krec_df	*krec = NULL;
-	daos_epoch_range_t	 val_epr = {0};
-	daos_handle_t		 toh = DAOS_HDL_INVAL;
-	int			 i, rc;
-	int			 flags = 0;
-	bool			 is_array = (iod->iod_type == DAOS_IOD_ARRAY);
-	bool			 has_cond = false;
 	struct daos_recx_ep_list *shadow;
-	bool			 standalone = ioc->ic_cont->vc_pool->vp_sysdb;
+	int                       rc = 0;
+	int                       i;
 
-	D_DEBUG(DB_IO, "akey "DF_KEY" fetch %s epr "DF_X64"-"DF_X64"\n",
-		DP_KEY(&iod->iod_name),
-		iod->iod_type == DAOS_IOD_ARRAY ? "array" : "single",
-		ioc->ic_epr.epr_lo, ioc->ic_epr.epr_hi);
-
-	if (is_array) {
-		if (iod->iod_nr == 0 || iod->iod_recxs == NULL) {
-			D_ASSERT(iod->iod_nr == 0 && iod->iod_recxs == NULL);
-			D_DEBUG(DB_TRACE, "akey "DF_KEY" fetch array bypassed - NULL iod_recxs.\n",
-				DP_KEY(&iod->iod_name));
-			return 0;
-		}
-		flags |= SUBTR_EVT;
-	}
-
-	rc = key_tree_prepare(ioc->ic_obj, ak_toh,
-			      VOS_BTR_AKEY, &iod->iod_name, flags,
-			      DAOS_INTENT_DEFAULT, &krec,
-			      (ioc->ic_check_existence || ioc->ic_read_ts_only) ? NULL : &toh,
-			      ioc->ic_ts_set);
-
-	if (stop_check(ioc, VOS_OF_COND_AKEY_FETCH, iod, &rc, true)) {
-		if (rc == 0 && !ioc->ic_read_ts_only)
-			iod_empty_sgl(ioc, ioc->ic_sgl_at);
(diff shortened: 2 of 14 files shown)


src/include/daos_obj.h
```diff
@@ -220,9 +220,9 @@ static inline bool
 daos_is_array_type(enum daos_otype_t type)
 {
 	switch (type) {
+	case DAOS_OT_ARRAY_BYTE:
```
jolivier23, line 223: I can probably revert these, just figured it might be better to put the more likely one on top

jolivier23:
Ping

src/vos/vos_io.c
```diff
@@ -1355,16 +1361,24 @@ dkey_fetch(struct vos_io_context *ioc, daos_key_t *dkey)
...
 
 	rc = obj_tree_init(obj);
 	if (rc != 0)
 		return rc;
 
-	rc = key_tree_prepare(obj, obj->obj_toh, VOS_BTR_DKEY,
-			      dkey, 0, DAOS_INTENT_DEFAULT, &krec,
-			      &toh, ioc->ic_ts_set);
+	if (vos_obj_flat_kv_supported(ioc->ic_cont, obj->obj_id)) {
+		if (ioc->ic_iod_nr == 1) {
```
wangdi1, line 1373: same here.

src/vos/vos_io.c
```diff
@@ -1868,16 +1896,24 @@ dkey_update(struct vos_io_context *ioc, uint32_t pm_ver, daos_key_t *dkey,
...
+			flags |= SUBTR_FLAT;
+			if (ioc->ic_iods[0].iod_type == DAOS_IOD_ARRAY)
+				flags |= SUBTR_EVT;
+		}
+	}
+
+	rc = key_tree_prepare(obj, obj->obj_toh, VOS_BTR_DKEY, dkey, flags, DAOS_INTENT_UPDATE,
+			      &krec, &ak_toh, ioc->ic_ts_set);
```
wangdi1, line 1916: I wonder should we skip dkey as well? i.e. attaching the ev tree to object initially, and only add dkey tree only if there are multiple dkeys being insert.

src/vos/vos_io.c
```diff
@@ -1908,12 +1944,18 @@ dkey_update(struct vos_io_context *ioc, uint32_t pm_ver, daos_key_t *dkey,
 		goto out;
 	}
 
-	for (i = 0; i < ioc->ic_iod_nr; i++) {
-		iod_set_cursor(ioc, i);
+	if (krec->kr_bmap & KREC_BF_FLAT) {
```
wangdi1, line 1947: should we check vp_feat here? in case there are some corruption?

src/vos/vos_obj.c
```diff
@@ -418,6 +430,15 @@ vos_obj_punch(daos_handle_t coh, daos_unit_oid_t oid, daos_epoch_t epoch,
 	if (oid.id_shard % 3 == 1 && DAOS_FAIL_CHECK(DAOS_DTX_FAIL_IO))
 		return -DER_IO;
 
+	cont = vos_hdl2cont(coh);
+
+	if (vos_obj_flat_kv_supported(cont, oid) && dkey != NULL && akeys != NULL) {
+		D_ERROR("Akey punch is not supported with flat object types: " DF_UOID "\n",
+			DP_UOID(oid));
+
+		return -DER_INVAL;
```
wangdi1, line 439: I wonder it should return -INVAL, ENONEXIST or even 0, in case it causing unnecessary failure if the user needs to do group akey punch.

src/vos/vos_io.c
```diff
@@ -1868,16 +1896,24 @@ dkey_update(struct vos_io_context *ioc, uint32_t pm_ver, daos_key_t *dkey,
...
 
 	rc = obj_tree_init(obj);
 	if (rc != 0)
 		return rc;
 
-	rc = key_tree_prepare(obj, obj->obj_toh, VOS_BTR_DKEY, dkey,
-			      SUBTR_CREATE, DAOS_INTENT_UPDATE, &krec, &ak_toh,
-			      ioc->ic_ts_set);
+	if (vos_obj_flat_kv_supported(ioc->ic_cont, obj->obj_id)) {
+		if (ioc->ic_iod_nr == 1) {
```
wangdi1, line 1908: should the flag depend on the request?  what will happen if the array object being accessed by daos_object_xxx with multiple iod?

wangdi1 (approved):
Just some questions and minor suggestion.

src/vos/vos_io.c
jolivier23: In vos_update_begin, vos_check_akeys will fail the update if there is more than one iod for the same akey.   This is a simplifying assumption for things like conditional operations.

So for no akey case, this should never happen.  A flat dkey should always have 1 entry.

Perhaps, though, we should check this in vos_check_akeys.

src/vos/vos_obj.c
jolivier23: Sorry, what is the use case for that?

src/vos/vos_io.c
jolivier23: I guess we could but it would just be to assert.

src/vos/vos_io.c
jolivier23: That's tricky because we need to store some dkey specific information such as the incarnation log and the actual key.  The most natural place for that is in the actual dkey.

I think DAOS-13560 gets us most of the way to what your are saying in that it would avoid creating all of the tree nodes in such a case.

src/vos/vos_io.c
jolivier23: Hmm, for fetch case, I guess we don't prevent multiple iod.  I should probably fix this then.

src/vos/vos_obj.c
wangdi1: I assume some background ULT might need punch all dkey/akey/recxs for consistency check or fixing object purpose.  Anyway my point is this is trying to punch akey on an non-akey object, it probably should return -ENOEXIST or 0. Though on the other hand, the caller might check the object(FLAT) before punch,  so return -EINVAL might make sense as well. hmm.

wangdi1 (approved):
Thanks!

src/vos/vos_obj.c
```diff
@@ -1960,70 +2199,160 @@ vos_obj_iter_fini(struct vos_iterator *iter)
...
+
+		it_entry->ie_epoch     = oiter->it_epr.epr_hi;
+		it_entry->ie_punch     = oiter->it_ilog_info.ii_next_punch;
+		it_entry->ie_obj_punch = oiter->it_obj->obj_ilog_info.ii_next_punch;
+		/** Use the dkey for this */
+		vos_ilog_last_update(&oiter->it_dkey_krec->kr_ilog, VOS_TS_TYPE_DKEY,
+				     &it_entry->ie_last_update, !!oiter->it_iter.it_for_sysdb);
+		d_iov_set(&it_entry->ie_key, &oiter->it_fake_akey, sizeof(oiter->it_fake_akey));
+
+		return 0;
```
NiuYawei, line 2334: I don't quite follow this, when this '0' fake akey will be fetched by iterator and how will the iterator caller deal with it?

jolivier23: It just treats it like any other iterator value.  The intention was to not really change the interface. For example, rebuild will still see this fake akey and will use it to fetch from remote and to update locally.

At some point, I may revisit this and remove it altogether but I didn't want to have to change a lot of code outside of vos to support this change.

NiuYawei (approved):

mchaarawi:
I would rather not land this before doing some more performance testing with this.

mchaarawi:
> I would rather not land this before doing some more performance testing with this.

ok .. or not..

PR #13182, opened 2023-10-16
DAOS-14063 engine: add RPC throttling

wangshilong:
Resources on DAOS server is limited, if service threads keep creating
ULT for incoming RPC, its processing capability will not be able to
keep up with receiving speed at some point, DAOS engine will be killed by OOM killer.

In order to control resource consumption on engine side: DAOS engine
should track number of incoming RPCs and created ULTs, these counters are per-xstream.

ULT counter: RPC being processed
queued RPC counter: RPC queued (no ULT yet)

RPCs will be rejected if they could be not processed timely
a new error code(DER_OVERLOAD_RETRY) is returned to client. and RPC enqueue's
timestamp is returned as well.

In the client side, RPC will be dealyed randomly from 0 to max delay(currently
it is RPC timeout). and resend RPC will use first time's enqueue timestamp.

In the server side, retried RPC will be inserted to a heap(sorted by enqueue timestamp).
while processing RPCs, always tries to pick up current min timestamp RPC from FIFO list and heap.

Required-githooks: true


src/engine/sched.c
```diff
@@ -1086,16 +1145,75 @@ policy_fifo_enqueue(struct dss_xstream *dx, struct sched_request *req,
...
+	}
 
-	process_req_list(dx, &info->si_fifo_list, false);
+	/*
+	 * Insert skipped retried RPCs back to heap.
+	 */
+	d_list_for_each_entry_safe(req, tmp, &tmp_list, sr_link) {
+		d_binheap_insert(&info->si_heap, &req->sr_node);
+		d_list_del_init(&req->sr_link);
+	}
```
NiuYawei, line 1216: This remove & re-insert heap operation looks weird and inefficient. I'm wondering if we'd use this opportunity to track the requests per pool?

src/engine/sched.c
```diff
@@ -1086,16 +1145,75 @@ policy_fifo_enqueue(struct dss_xstream *dx, struct sched_request *req,
...
+	 * If @sra_enqueue_ts is not zero, this is a resent RPC, it will
+	 * be inserted to sorted heap. if it failed to insert because of
+	 * out of memory, use fifo list instead.
+	 */
+	if (attr->sra_enqueue_ts) {
+		rc = d_binheap_insert(&info->si_heap, &req->sr_node);
+		if (rc)
+			goto out;
+		return;
+	}
```
NiuYawei, line 1162: This retry request tracking will be per policy instead of common code shared by different polices? I don't against this, just ask to confirm.

src/object/srv_mod.c
```diff
@@ -232,8 +243,33 @@ obj_get_req_attr(crt_rpc_t *rpc, struct sched_req_attr *attr)
...
+	/* proto v9 doesn't support hint return cart timeout for retry */
+	if (proto_ver == 9)
+		return -DER_TIMEDOUT;
+
+	if (obj_rpc_is_update(rpc) || obj_rpc_is_fetch(rpc)) {
+		struct obj_rw_v10_out	*orwo_v10 = crt_reply_get(rpc);
+
+		orwo_v10->orw_comm_out.req_out_enqueue_ts = attr->sra_enqueue_ts;
+		orwo_v10->orw_ret = -DER_OVERLOAD_RETRY;
+		return crt_reply_send(rpc);
```
NiuYawei, line 261: It looks to me this function should only set request.

src/engine/srv.c
```diff
@@ -290,7 +298,18 @@ dss_rpc_hdlr(crt_context_t *ctx, void *hdlr_arg,
...
+		return rc;
+
+	if (module != NULL && module->sm_mod_ops != NULL &&
+	    module->sm_mod_ops->dms_set_req != NULL)
+		return module->sm_mod_ops->dms_set_req(rpc, &attr);
+
+	/*
+	 * No RPC protocol to return hint, just return timeout.
+	 */
+	return -DER_TIMEDOUT;
```
NiuYawei, line 312: Cart will send replay in this case?

wangshilong: Yes

src/engine/sched.c
wangshilong: Yes.

src/engine/sched.c
```diff
@@ -1086,16 +1145,75 @@ policy_fifo_enqueue(struct dss_xstream *dx, struct sched_request *req,
...
 {
 	struct sched_info	*info = &dx->dx_sched_info;
+	struct sched_req_attr	*attr = &req->sr_attr;
+	int			 rc;
 
+	D_ASSERT(attr->sra_type < SCHED_REQ_TYPE_MAX);
+	/*
+	 * If @sra_enqueue_ts is not zero, this is a resent RPC, it will
+	 * be inserted to sorted heap. if it failed to insert because of
+	 * out of memory, use fifo list instead.
```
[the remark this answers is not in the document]
wangshilong: Fixed.

src/engine/sched.c
```diff
@@ -1061,8 +1113,13 @@ process_pool_cb(d_list_t *rlink, void *arg)
 	if (spi->spi_req_cnt == 0)
 		return 0;
 
-	for (i = SCHED_REQ_UPDATE; i < SCHED_REQ_MAX; i++)
+	for (i = SCHED_REQ_UPDATE; i < SCHED_REQ_MAX; i++) {
 		kick[i] = pool2req_cnt(spi, i);
+		if (info->si_total_req_cnt > MAX_KICKED_REQ_CNT) {
+			tmp = kick[i] * MAX_KICKED_REQ_CNT / info->si_total_req_cnt;
+			kick[i] = tmp;
```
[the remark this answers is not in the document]
NiuYawei: It's weird that total limit of MAX_KICKED_REQ_CNT is against UPDATE & FETCH, but here all request types are reduced based on the "max_io_req_cnt / total_req_cnt", shouldn't we reduce the update & fetch only based on the "max_io_req_cnt / total_io_req_cnt"?

src/engine/sched.c
```diff
@@ -1150,67 +1289,121 @@ should_enqueue_req(struct dss_xstream *dx, struct sched_req_attr *attr)
...
 		D_ASSERT(policy_ops[sched_policy].enqueue_io != NULL);
-		policy_ops[sched_policy].enqueue_io(dx, req, NULL);
+		rc = policy_ops[sched_policy].enqueue_io(dx, req, NULL);
 	} else {
 		d_list_add_tail(&req->sr_link, &sri->sri_req_list);
 	}
 	req->sr_enqueue_ts = info->si_cur_ts;
+	/* Hint is RPC enqueue id */
+	if (attr->sra_enqueue_id == 0)
+		attr->sra_enqueue_id = info->si_cur_ts;
```
NiuYawei, line 1317: Why this is required, I think the 'sra_enqueue_id' has been assigned, right?

src/engine/sched.c
```diff
@@ -1150,67 +1289,121 @@ should_enqueue_req(struct dss_xstream *dx, struct sched_req_attr *attr)
...
 {
 	struct sched_request	*req;
+	struct sched_info	*info = &dx->dx_sched_info;
 
 	if (!should_enqueue_req(dx, attr))
 		return req_kickoff_internal(dx, attr, func, arg);
 
+	info->si_cur_id++;
+	if (attr->sra_enqueue_id == 0)
+		attr->sra_enqueue_id = info->si_cur_id;
```
NiuYawei, line 1376: minor "attr->sra_enqueue_id = ++info->si_cur_id", it's not necessary to bump 'si_cur_id' for resend RPC.

src/engine/sched.c
```diff
@@ -1150,67 +1289,121 @@ should_enqueue_req(struct dss_xstream *dx, struct sched_req_attr *attr)
...
+		return false;
+
+	/*
+	 * Calculate time based on ults on argobots and non-system
+	 * requests queued. It is not easy to estimate how many system
+	 * ults will be executed, reserve 50% of timeout for it.
+	 */
+	for (i = SCHED_REQ_UPDATE; i < SCHED_REQ_MAX; i++) {
+		estimated_time += info->si_kicked_req_cnt[i] * req_latencys[i];
+		req_num += info->si_kicked_req_cnt[i];
```
NiuYawei, line 1347: These "kicked_req_cnt"s are from last schedule cycle or from current cycle, probably we'd just use a pre-calculated maximum cycle time here.

src/engine/sched.c
```diff
@@ -1150,67 +1289,121 @@ should_enqueue_req(struct dss_xstream *dx, struct sched_req_attr *attr)
...
+			req_num += info->si_req_cnt[i];
+		}
+	}
+	/* convert to msecs */
+	estimated_time /= 1000;
+	if ((estimated_time << 1) > attr->sra_timeout)
+		return true;
+
+	if ((req_num << 1) > MAX_SCHED_REQ_NUM)
+		return true;
```
NiuYawei, line 1359: Instead of silently reject the RPCs, It's better to print error messages (or log in telemetry) when the request is rejected?

src/object/srv_mod.c
```diff
@@ -232,8 +309,85 @@ obj_get_req_attr(crt_rpc_t *rpc, struct sched_req_attr *attr)
...
+		orwo_v10->orw_ret = -DER_OVERLOAD_RETRY;
+
+		return -DER_OVERLOAD_RETRY;
+	} else if (obj_rpc_is_migrate(rpc)) {
+		struct obj_migrate_out *om = crt_reply_get(rpc);
+
+		om->om_comm_out.req_out_enqueue_id = attr->sra_enqueue_id;
+		om->om_status = -DER_OVERLOAD_RETRY;
+
+		return -DER_TIMEDOUT;
```
NiuYawei, line 332: Why this kind of RPC is special?

daltonbohning (commented):
ftest LGTM

src/engine/sched.c
```diff
@@ -864,17 +876,20 @@ process_req(struct dss_xstream *dx, struct sched_request *req)
...
 		goto kickoff;
 
-	for (i = SCHED_REQ_UPDATE; i < SCHED_REQ_MAX; i++)
-		ult_timeout += info->si_kicked_req_cnt[i] * req_latencys[i];
-	ult_timeout /= 1000;
-
 	/* Request expired */
-	if ((info->si_cur_ts - req->sr_enqueue_ts + ult_timeout) >
-	    (req->sr_attr.sra_timeout * 2 / 3))
+	if (req->sr_attr.sra_timeout &&
```
NiuYawei, line 880: Is it possible that 'sra_timeout > MAX_CYCLE_TIME'? It's better to change this check from 'sra_timeout != 0" to "sra_timeout > MAX_CYCLE_TIME".

src/engine/sched.c
```diff
@@ -527,8 +527,59 @@ sched_metrics_init(struct dss_xstream *dx)
 			     "ULT", "sched/cycle_size/xs_%u", dx->dx_xs_id);
 	if (rc)
 		D_WARN("Failed to create cycle_size telemetry: "DF_RC"\n", DP_RC(rc));
+
+	rc = d_tm_add_metric(&stats->ss_total_reject, D_TM_COUNTER, "Total rejected requests",
+			     "req", "sched/total_reject/xs_%u", dx->dx_xs_id);
+	if (rc)
+		D_WARN("Failed to create total_reject telemetry: "DF_RC"\n", DP_RC(rc));
```
NiuYawei, line 534: Test script need be adjusted for this new per-engine telemetry.

src/engine/sched.c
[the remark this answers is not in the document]
NiuYawei: kick[i] * MAX_KICKED_REQ_CNT / rpc_cnt?

src/engine/sched.c
NiuYawei: This line should be removed?

src/engine/sched.c
```diff
@@ -1150,67 +1299,119 @@ should_enqueue_req(struct dss_xstream *dx, struct sched_req_attr *attr)
...
+		req_num += info->si_kicked_req_cnt[i];
+		if (!is_system_req(i)) {
+			estimated_time += info->si_req_cnt[i] * req_latencys[i];
+			req_num += info->si_req_cnt[i];
+		}
+	}
+	/* convert to msecs */
+	estimated_time /= 1000;
+	estimated_time += MAX_CYCLE_TIME;
+	if ((estimated_time << 1) > attr->sra_timeout)
```
NiuYawei, line 1362: Isn't it too aggressive to reject an RPC when the (estimated time) > (timeout / 2)?

src/engine/sched.c
wangshilong: Though, MAX_CYCLE_TIME should be less than 1 seconds, and sra_timeout normally larger than 1 second, but check is better.

NiuYawei (approved):

src/include/daos/rpc.h
```diff
@@ -93,6 +93,22 @@ enum daos_rpc_type {
...
+	uint32_t	req_in_gid;
+	uint32_t	req_in_projid;
+	uint64_t	req_in_enqueue_id;
+	uint64_t	req_in_paddings[4];
+	d_string_t	req_in_addr;
+	d_string_t	req_in_jobid;
+};
+
+struct daos_req_comm_out {
+	uint64_t	req_out_enqueue_id;
```
liuxuezhao, line 107: is it a per server XS unique ID to identify retried OP? better with a few comments about what it is

I have not fully understood some details, a possible case for example a fetch request/rpc sent to rank 1's tag 1, and server cannot serve it and reply -DER_OVERLOAD_RETRY with an enqueue_id. But when client retry, that server dead and will do degraded fetch so the retried fetchh request send to rank 2's tag 2 with that enqueue_id got from rank 1's tag 1, will that cause problem? 
similar problem for UPDATE if the retired UPDATE goes to different rank/tag (due to pool map change).
One possible problem maybe that the enque_id got from rank 1 tag 1 is very big, when retry to rank 2 tag 2 it may cause the retried request cannot be served in time?

src/include/daos/rpc.h
```diff
@@ -93,6 +93,22 @@ enum daos_rpc_type {
 	DAOS_REQ_TGT,
 };
 
+struct daos_req_comm_in {
+	uint32_t	req_in_uid;
+	uint32_t	req_in_gid;
+	uint32_t	req_in_projid;
+	uint64_t	req_in_enqueue_id;
+	uint64_t	req_in_paddings[4];
+	d_string_t	req_in_addr;
```
liuxuezhao, line 102: some comments would be great, what is the addr, enqueue_id...

src/engine/sched.c
```diff
@@ -1081,25 +1175,80 @@ process_pool_cb(d_list_t *rlink, void *arg)
...
+	int			 rc;
 
-	process_req_list(dx, &info->si_fifo_list, false);
+	D_INIT_LIST_HEAD(&tmp_list);
+	/*
+	 * All retried RPCs are inserted into a sorted heap, they are sorted
+	 * by RPC enqueue sequence ID in the server side(firstly enqueue time). So
+	 * retried RPCs won't starve forever.
+	 */
+	d_list_for_each_entry_safe(req, tmp, &info->si_fifo_list, sr_link) {
```
liuxuezhao, line 1213: the request in si_fifo_list is new request, and the request in si_heap is retried old request?
here is to ensure if with retried request with smaller sra_enqueue_id than the new request, it will process the retried request? why must ensure this order?
If the retried request has not arrive to server, that order cannot be ensured anyway right?

src/engine/sched.c
```diff
@@ -1081,25 +1175,80 @@ process_pool_cb(d_list_t *rlink, void *arg)
...
+			} else {
+				break;
+			}
+		}
+		process_req(dx, req);
+	}
+
+	/* Process retried RPCs if any */
+	while (!d_binheap_is_empty(&info->si_heap)) {
+		node = d_binheap_root(&info->si_heap);
```
liuxuezhao, line 1232: just confirm why the retried request must be processed by original enqueue_id order? thx

src/engine/sched.c
```diff
@@ -802,24 +876,23 @@ process_req(struct dss_xstream *dx, struct sched_request *req)
...
-		pr = &pressure_gauge[spi->spi_space_pressure];
-		delay_msecs = pr->pr_delay;
-	} else {
-		delay_msecs = max_delay_msecs[req_type];
-	}
-
 	/* Request expired */
-	D_ASSERT(info->si_cur_ts >= req->sr_enqueue_ts);
-	if ((info->si_cur_ts - req->sr_enqueue_ts) > delay_msecs)
+	if (req->sr_attr.sra_timeout > MAX_CYCLE_TIME &&
```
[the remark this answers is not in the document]
wangshilong: Yes.

src/engine/sched.c
```diff
@@ -778,14 +844,22 @@ check_space_pressure(struct dss_xstream *dx, struct sched_pool_info *spi)
...
 }
 
+/*
+ * Current default ult stack size is 16kib, limit amount of memory
+ * that each target can use to process RPC to 100MiB.
+ */
+#define DEFAULT_STACKSIZE	16384
+#define MAX_KICKED_REQ_CNT	((100 << 20) / 16384)
+/* max cycle time in msecs */
+#define MAX_CYCLE_TIME		((MAX_KICKED_REQ_CNT * 20) / 1000)
```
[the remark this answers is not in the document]
wangshilong: Yup, 20 us is max estimate time for different REQ type, just want to be conservative a bit for it.

src/include/daos/rpc.h
```diff
@@ -214,4 +230,15 @@ daos_rpc_from_client(crt_rpc_t *rpc)
 int
 daos_rpc_proto_query(crt_opcode_t base_opc, uint32_t *ver_array, int count, int *ret_ver);
 
+static inline uint32_t
+daos_rpc_rand_delay(uint32_t max_delay)
+{
+	if (max_delay == 0)
+		return 0;
+	if (max_delay > 5)
+		max_delay -= 5;
```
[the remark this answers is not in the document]
wangshilong: just to make random time is between [1, max - 5 ] range.

src/include/daos/rpc.h
wangshilong: In the client codes, enqueue id is per shard, not per object. probably fine? or i mis-write something

src/include/daos/rpc.h
wangshilong: Sure.

src/engine/sched.c
wangshilong: smaller ID means it arrived server earlier in the first time, it is still a kind of "FIFO" to some retried RPCs could be not handlined for a long time which might cause tail-latency very high for some application. 

For second question, yup, we don't know when retried request might arrive to server, but once arrive, smaller ID always
handle firstly to avoid above issue.

src/engine/sched.c
wangshilong: see above replies.

src/object/cli_obj.c
```diff
@@ -2712,6 +2719,9 @@ shard_auxi_set_param(struct shard_auxi_args *shard_arg, uint32_t map_ver,
 		     uint32_t shard, uint32_t tgt_id, struct dtx_epoch *epoch,
 		     uint16_t ec_tgt_idx)
 {
+	/* Reset @enqueue_id if target changed */
+	if (shard_arg->target != tgt_id)
+		shard_arg->enqueue_id = 0;
```
liuxuezhao, line 2724: the query key operation is not affected by this change, you may also need to change queue_shard_query_key_task().
Does it work if use HLC as the enqueue_id at server-side? by that need not such change at client-side.

wangshilong: @NiuYawei and i discussed a bit that way, we think it better to reset enqueued ID as it is a new request. (send a TLC to new target means new RPC could jump a queue).

Let me fix shard_query_key_task() case thanks!

liuxuezhao (approved):

NiuYawei (approved):

PR #13200, opened 2023-10-18
DAOS-14521 client: intercept __pread64_chk() and __read_chk()

wiliamhuang:
Skip-func-test-el9: false
Features: pil4dfs

intercept fstatat() with trampoline. add nlt test for file. also replace "a" with "w" in open since O_APPEND is not supported by DFS.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfuse/pil4dfs/int_dfs.c
+++ b/src/client/dfuse/pil4dfs/int_dfs.c
@@ -410,9 +410,9 @@ static int (*next_xstat)(int ver, const char *path, struct stat *stat_buf);
 
 static int (*libc_lxstat)(int ver, const char *path, struct stat *stat_buf);
 
-static int (*next_fxstatat)(int ver, int dirfd, const char *path, struct stat *stat_buf, int flags);
+static int (*libc_fxstatat)(int ver, int dirfd, const char *path, struct stat *stat_buf, int flags);
 
-static int (*next_fstatat)(int dirfd, const char *path, struct stat *stat_buf, int flags);
+static int (*libc_fstatat)(int dirfd, const char *path, struct stat *stat_buf, int flags);
 
 static int (*next_statx)(int dirfd, const char *path, int flags, unsigned int mask,
 			 struct statx *statx_buf);
@@ -2069,6 +2069,26 @@ ssize_t
 __pread64(int fd, void *buf, size_t size, off_t offset) __attribute__((alias("pread")));
 
 ssize_t
+__pread64_chk(int fd, void *buf, size_t size, off_t offset, size_t buflen)
+{
+	if (size > buflen) {
+		D_FATAL("buffer overflow detected in __pread64_chk().\n");
+		abort();
+	}
+	return pread(fd, buf, size, offset);
+}
+
+ssize_t
+__read_chk (int fd, void *buf, size_t size, size_t buflen)
+{
+	if (size > buflen) {
+		D_FATAL("buffer overflow detected in __read_chk().\n");
+		abort();
+	}
+	return read(fd, buf, size);
+}
+
+ssize_t
 write_comm(ssize_t (*next_write)(int fd, const void *buf, size_t size), int fd, const void *buf,
 	   size_t size)
 {
@@ -2319,7 +2339,7 @@ new_fxstatat(int ver, int dirfd, const char *path, struct stat *stat_buf, int fl
 	char *full_path = NULL;
 
 	if (!hook_enabled)
-		return next_fxstatat(ver, dirfd, path, stat_buf, flags);
+		return libc_fxstatat(ver, dirfd, path, stat_buf, flags);
 
 	if (path[0] == '/') {
 		/* Absolute path, dirfd is ignored */
@@ -2339,7 +2359,7 @@ new_fxstatat(int ver, int dirfd, const char *path, struct stat *stat_buf, int fl
 		else
 			rc = new_xstat(1, full_path, stat_buf);
 	} else {
-		rc = next_fxstatat(ver, dirfd, path, stat_buf, flags);
+		rc = libc_fxstatat(ver, dirfd, path, stat_buf, flags);
 	}
 
 	error = errno;
@@ -2355,17 +2375,13 @@ out_err:
 }
 
 int
-fstatat(int dirfd, const char *__restrict path, struct stat *__restrict stat_buf, int flags)
+new_fstatat(int dirfd, const char *__restrict path, struct stat *__restrict stat_buf, int flags)
 {
 	int  idx_dfs, error = 0, rc;
 	char *full_path = NULL;
 
-	if (next_fstatat == NULL) {
-		next_fstatat = dlsym(RTLD_NEXT, "fstatat");
-		D_ASSERT(fstatat != NULL);
-	}
 	if (!hook_enabled)
-		return next_fstatat(dirfd, path, stat_buf, flags);
+		return libc_fstatat(dirfd, path, stat_buf, flags);
 
 	if (path[0] == '/') {
 		/* Absolute path, dirfd is ignored */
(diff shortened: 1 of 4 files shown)


src/tests/ftest/util/verify_perms.py
```diff
@@ -261,7 +261,10 @@ def _real_w(entry_type, path):
     if entry_type == 'file':
         try:
             # Write a newline to reduce likelihood of corrupting an executable file
-            with open(path, 'a', encoding='utf-8') as file:
+            with open(path, "w", encoding='utf-8') as file:
+                # O_APPEND is not supported by DFS, so "a" is not used in open. Need seek() here
+                # to move file pointer to the end of the file.
+                file.seek(0, 2)
```
ashleypittman, line 267: dfs doesn't support O_APPEND but dfuse does.  This was a regression in your last change and I already have a fix in https://github.com/daos-stack/daos/pull/13204

Please revert this change.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -2068,6 +2068,26 @@ pread64(int fd, void *buf, size_t size, off_t offset) __attribute__((alias("prea
 ssize_t
 __pread64(int fd, void *buf, size_t size, off_t offset) __attribute__((alias("pread")));
 
+ssize_t
+__pread64_chk(int fd, void *buf, size_t size, off_t offset, size_t buflen)
+{
+	if (size > buflen) {
+		D_FATAL("buffer overflow detected in __pread64_chk().\n");
```
ashleypittman, line 2075: I thought we had code in ftest that would check for command output from the libc variant of these functions but I can't see it now, only https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/cmocka_utils.py#L93 Perhaps @daltonbohning could point us to it?

If I'm correct we should make it so that either the output here matches the libc version closely enough to match the check, or we should add the `buffer overflow detected` output to the check.

You also don't need to incorporate the function name here, the logging prefix will include that anyway.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -2068,6 +2068,26 @@ pread64(int fd, void *buf, size_t size, off_t offset) __attribute__((alias("prea
...
+		abort();
+	}
+	return pread(fd, buf, size, offset);
+}
+
+ssize_t
+__read_chk (int fd, void *buf, size_t size, size_t buflen)
```
ashleypittman, line 2082: Whitespace issue.  Other than this line the patch is clang-format clean.
```suggestion
__read_chk(int fd, void *buf, size_t size, size_t buflen)
```

ashleypittman (commented):
How does the __pread64_chk version work, I can see that you intercept a new function but if you didn't do this wouldn't the libc version of this check call pread which would then be intercepted?  Perhaps the libc version is using relative jump, not the symbol table?

I assume this check is doing the same as the libc version with a simple size compare?  Do you have the source to the libc version to check?

I don't know if it helps but I did some work on `_FORTIFY_SOURCE=3` recently which expands the checks here, perhaps that's all taken care of at a higher level?

src/tests/ftest/util/verify_perms.py
wiliamhuang: Yes. dfuse does support O_APPEND. This test will be run with libpil4dfs too which bypasses dfuse. Without this change, "test_dfuse_mu_perms_pil4dfs" fails.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. Thank you! I will revise the message to exactly match error message in glibc.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! Will fix it in next commit.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: You could also presumably call `__chk_fail`()?

wiliamhuang:
> How does the __pread64_chk version work, I can see that you intercept a new function but if you didn't do this wouldn't the libc version of this check call pread which would then be intercepted? Perhaps the libc version is using relative jump, not the symbol table?

000000000013d850 <__pread64_chk>:
  13d850:       f3 0f 1e fa             endbr64
  13d854:       4c 39 c2                cmp    rdx,r8
  13d857:       77 05                   ja     13d85e <__pread64_chk+0xe>
  13d859:       e9 92 10 fe ff          jmp    11e8f0 <__libc_pread>

__pread64_chk() checks buffer length and read size, then jump to pread() as you thought.

> I assume this check is doing the same as the libc version with a simple size compare? Do you have the source to the libc version to check?

Yes. I did check the source code of glibc.
ssize_t
__pread_chk (int fd, void *buf, size_t nbytes, off_t offset, size_t buflen)
{
  if (nbytes > buflen)
    __chk_fail ();

  return __pread (fd, buf, nbytes, offset);
}

> 
> I don't know if it helps but I did some work on `_FORTIFY_SOURCE=3` recently which expands the checks here, perhaps that's all taken care of at a higher level?

Not sure about this. I will take a look. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Yes, we can call __chk_fail(). Will update it soon.

src/tests/ftest/util/verify_perms.py
ashleypittman: Yes, I see you're correct.  IMO we should move this file and one more like it into a subdirectory to make that distinction clear.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: Looking at the ftest code further it does appear that that is the list being used, and `buffer overflow detected` should be added to the list as per https://github.com/lattera/glibc/blob/master/debug/chk_fail.c but I don't know how or why a cmocka function is being referenced from ftest.

wiliamhuang: @ashleypittman The output from __chk_fail() is as following,

*** buffer overflow detected ***: terminated
Aborted (core dumped)

Shall we add "buffer overflow detected" only or the full message?

ashleypittman: just `buffer overflow detected`

wiliamhuang:
> I don't know if it helps but I did some work on `_FORTIFY_SOURCE=3` recently which expands the checks here, perhaps that's all taken care of at a higher level?

Regarding "_FORTIFY_SOURCE=3", it could be used to help improve the security of our own product. The interception library will be used together with users' applications, so we still do need to intercept all relevant APIs.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! Just added it in "src/tests/ftest/util/cmocka_utils.py".

src/tests/ftest/util/verify_perms.py
daltonbohning: > we should move this file and one more like it into a subdirectory to make that distinction clear

I don't understand this. What distinction? And move where?

daltonbohning (approved):

ashleypittman:
> > I don't know if it helps but I did some work on `_FORTIFY_SOURCE=3` recently which expands the checks here, perhaps that's all taken care of at a higher level?
> 
> Regarding "_FORTIFY_SOURCE=3", it could be used to help improve the security of our own product. The interception library will be used together with users' applications, so we still do need to intercept all relevant APIs.

DAOs will already use it if available.  It probably only affects when this code gets called, not how it gets called.

ashleypittman (approved):
I assume this isn't needed for `ioil` because of differences in how the interception is handled.  It would be interesting to try and out-of-bounds read to see how it behaves, we might be able to add this behaviour to ioil.

wiliamhuang:
> I assume this isn't needed for `ioil` because of differences in how the interception is handled. It would be interesting to try and out-of-bounds read to see how it behaves, we might be able to add this behaviour to ioil.

If an application calls "__pread64_chk()", libioil does not intercept it and pread64() will be handled by dfuse.

PR #13212, opened 2023-10-20
DAOS-14245 gurt: Move print_backtrace to gurt.

ashleypittman:
Make this function available to any daos code, and enabled by
default in the engine and dfuse.

Required-githooks: true


src/gurt/signals.c
```diff
@@ -0,0 +1,147 @@
...
+
+void
+d_signal_register()
+{
+	bool enabled = register_handler;
+
+	if (registered)
+		return;
+
+	d_getenv_bool("DAOS_SIGNAL_REGISTER", &enabled);
```
ashleypittman, line 135: @frostedcmos is there a list I need to add this to?

src/gurt/signals.c
```diff
@@ -0,0 +1,147 @@
...
+		/* XXX will old handler get accurate siginfo_t/ucontext_t ?
+		 * we may prefer to call it with the same params we got ?
+		 */
+		raise(signo);
+	}
+
+	memset(&old_handlers[signo], 0, sizeof(struct sigaction));
+}
+
+bool register_handler = false;
```
jolivier23, line 118: Should these be static?

jolivier23 (changes requested):
sorry, changing approval.  I think those variables can and should be static.

src/gurt/signals.c
ashleypittman: Yes.  Will wait for Alexs response and repush.

ashleypittman: I'm slightly annoyed these don't cause a build failure, do we not check for exported symbols from the library any more?  Perhaps modern compiler flags is hiding them though?   Anyway, will update.

src/gurt/signals.c
frostedcmos: > @frostedcmos is there a list I need to add this to?

We have a dump list used in dump_envariables() at init time for debug purposes, but we typically only add cart or mercury/ofi level envariables there.

src/include/gurt/common.h
```diff
@@ -987,6 +987,19 @@ int d_vec_pointers_init(struct d_vec_pointers *pointers, uint32_t cap);
 void d_vec_pointers_fini(struct d_vec_pointers *pointers);
 int d_vec_pointers_append(struct d_vec_pointers *pointers, void *pointer);
 
+/** Change the default setting for if a signal handler should be installed in cart_init()
```
frostedcmos, line 990: minor: crt_init() not cart_init()

src/gurt/signals.c
ashleypittman: Found, and done.

frostedcmos (dismissed):

frostedcmos (approved):

src/gurt/signals.c
jolivier23: right now, most of our libraries export all symbols that are not static.  Only libioil, I think changes this behavior by setting that PUBLIC api flag.  We should probably address that.

jolivier23 (approved):

PR #13213, opened 2023-10-20
DAOS-14537 control: Re-create PMem regions even if state cannot be determined

tanabarr:
In cases where errors with PMem causes region socket affinity problems
proceed to remove regions despite errors reported when evaluating PMem
state. Log failure to NOTICE when state cannot be evaluated.

Change daos_server scm reset behaviour to re-create the AppDirect goal
rather than changing the resource allocation to MemoryMode as before.
Re-setting AppDirect mode resolves the most common situations where
previously un-healthy modules have been replaced. There aren't any
clearly identified benefits of resetting the resource allocation to
MemoryMode.

Features: control
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/server/storage/scm/ipmctl.go
+++ b/src/control/server/storage/scm/ipmctl.go
@@ -431,7 +431,15 @@ func (cr *cmdRunner) prepReset(req storage.ScmPrepareRequest, scanRes *storage.S
 
 	sockState, err := getPMemState(cr.log, regions)
 	if err != nil {
-		return nil, errors.Wrap(err, "getPMemState")
+		err = errors.Wrap(err, "getPMemState")
+		if req.Force {
+			sockState = &storage.ScmSocketState{
+				State: storage.ScmUnknownMode,
+			}
+			cr.log.Error(err.Error())
+		} else {
+			return nil, err
+		}
 	}
 	resp.Socket = sockState
 	if sockState.SocketID == nil {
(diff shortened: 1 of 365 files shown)


knard-intel:
Codespell errors seems to be a CI issue and not related to this PR.
However, unit test seems to be legitimate.
Otherwise, the code is mostly OK for me.
Just one interrogation regarding the log level error when using the force option. From my understanding, the error level message are printed with unrecoverable errors.  Thus, in this case, a warning level could be more suited ?

tanabarr:
> Codespell errors seems to be a CI issue and not related to this PR. However, unit test seems to be legitimate. Otherwise, the code is mostly OK for me. Just one interrogation regarding the log level error when using the force option. From my understanding, the error level message are printed with unrecoverable errors. Thus, in this case, a warning level could be more suited ?

I can see arguments for both, I've labelled them as Errors because they are, it's not "warning there might be something wrong" it's "there is something wrong and the system is not in the expected state". do you want me to make the change and repush?

knard-intel (approved):

kjacque (dismissed):

mjmac (changes requested):
Adding a -1 for the moment to ensure that my questions in the ticket are addressed.

knard-intel (dismissed):
LGTM

tanabarr:
Codespell reported issue unrelated to this PR's changes.

Manually verified changes on wolf-226 by:
- `daos_server scm reset` verifying regions re-created after reboot
- `daos_server scm prepare` then creates namespaces

Unit tests verify that regions are removed even when PMem states cannot be evaluated.

Changes to be verified on cluster to show reset of regions after unhealthy NVDIMM gets replaced.

src/control/server/storage/scm/ipmctl.go
```diff
@@ -460,14 +463,15 @@ func (cr *cmdRunner) prepReset(req storage.ScmPrepareRequest, scanRes *storage.S
 
 	for _, dev := range scanRes.Namespaces {
 		if err := cr.removeNamespace(dev.Name); err != nil {
-			return nil, err
+			cr.log.Noticef("removeNamespace: %s", err.Error())
+			// Continue to reset regions.
```
kjacque, line 467: I'm curious about what happens practically if it's not possible to remove the namespaces for some reason. If there are undeleted namespaces on the DIMM, will `ipmctl create -goal` still work? If not, it would be better to return the error here.

Do we know why namespaces might not be deleted? All I can think of is the device being busy.

kjacque (commented):
Just one question, overall looks good to me.

src/control/server/storage/scm/ipmctl.go
tanabarr: yes the force create -goal should purge namespaces on the NVDIMM on reboot
yes device busy could cause refusal to delete but they will still be removed on reboot

tanabarr: updated

mjmac (approved):

kjacque (approved):

tanabarr:
@daos-stack/daos-gatekeeper can this PR landed please?

PR #13243, opened 2023-10-26
DAOS-14559 ci: use isort to organize python imports

YinglinSun:
Required-githooks: true

Signed-off-by: Yinglin Sun <yinsun@tesla.com>

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate watchers.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
  * [ ] Quick-build or Quick-functional is not used.
* [ ] Fix the commit message upon landing. Check the standard [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments). Edit it to create a single commit. If necessary, ask submitter for a new summary.

PR #13247, opened 2023-10-27
DAOS-14559 ci: fix codespell and pin isort version

daltonbohning:
- Fix Codesepll introduced by #13243
- pin isort==5.9 so it's consistent

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/tests/ftest/launch.py
+++ b/src/tests/ftest/launch.py
@@ -31,7 +31,7 @@ from process_core_files import CoreFileException, CoreFileProcessing
 from slurm_setup import SlurmSetup, SlurmSetupException
 
 # Update the path to support utils files that import other utils files
-# This is not good coding practice. Should use package paths and remove alls these E402.
+# This is not good coding practice. Should use package paths and remove all these E402.
 sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), "util"))
 from data_utils import (dict_extract_values, list_flatten,  # noqa: E402
                         list_unique)
--- a/utils/cq/requirements.txt
+++ b/utils/cq/requirements.txt
@@ -11,4 +11,5 @@ isort
 paramiko
 pyenchant
 pylint
+toml
 yamllint
--- a/utils/cq/requirements.txt
+++ b/utils/cq/requirements.txt
@@ -7,9 +7,8 @@ clustershell
 ## flake8 6 removed --diff option which breaks flake precommit hook.
 ## https://github.com/pycqa/flake8/issues/1389 https://github.com/PyCQA/flake8/pull/1720
 flake8<6.0.0
-isort
+isort==5.9
 paramiko
 pyenchant
 pylint
-toml
 yamllint
--- a/utils/githooks/pre-commit.d/20-isort.sh
+++ b/utils/githooks/pre-commit.d/20-isort.sh
@@ -4,6 +4,13 @@
 
 set -ue
 
+# Disable isort check because it doesn't work
+# - it needs to run with -rc
+# - it conflicts with pylint checks
+# - ~100 files are flagged with import errors
+# - it's really slow
+exit 0
+
 if ! command -v isort > /dev/null 2>&1
 then
     echo "No isort checking, install isort command to improve pre-commit checks"
--- a/utils/githooks/pre-commit.d/20-isort.sh
+++ b/utils/githooks/pre-commit.d/20-isort.sh
@@ -4,17 +4,11 @@
 
 set -ue
 
-# Disable isort check because it doesn't work
-# - it needs to run with -rc
-# - it conflicts with pylint checks
-# - ~100 files are flagged with import errors
-# - it's really slow
-exit 0
 
 if ! command -v isort > /dev/null 2>&1
 then
     echo "No isort checking, install isort command to improve pre-commit checks"
-    echo "python3 -m pip install isort"
+    echo "python3 -m pip install -r ./utils/cq/requirements.txt"
     exit 0
 fi
 
--- a/utils/githooks/pre-commit.d/20-flake.sh
+++ b/utils/githooks/pre-commit.d/20-flake.sh
@@ -18,6 +18,7 @@ set -ue
 echo "Flake8:"
 if ! command -v flake8 > /dev/null 2>&1; then
     echo "  No flake checking, install flake8 command to improve pre-commit checks"
+    echo "python3 -m pip install -r ./utils/cq/requirements.txt"
     exit 0
 fi
 
--- a/utils/githooks/pre-commit.d/30-yamllint.sh
+++ b/utils/githooks/pre-commit.d/30-yamllint.sh
@@ -10,7 +10,7 @@ set -ue
 if ! command -v yamllint > /dev/null 2>&1
 then
     echo "No yaml checking, install yamllint command to improve pre-commit checks"
-    echo "python3 -m pip install yamllint"
+    echo "python3 -m pip install -r ./utils/cq/requirements.txt"
     exit 0
 fi
 


src/tests/ftest/launch.py
```diff
@@ -31,7 +31,7 @@
 from slurm_setup import SlurmSetup, SlurmSetupException
 
 # Update the path to support utils files that import other utils files
-# This is not good coding practice. Should use package paths and remove alls these E402.
+# This is not good coding practice. Should use package paths and remove all these E402.
```
YinglinSun, line 34: Thanks for fixing this. 👍

utils/githooks/pre-commit.d/20-isort.sh
```diff
@@ -4,6 +4,13 @@
 
 set -ue
 
+# Disable isort check because it doesn't work
+# - it needs to run with -rc
+# - it conflicts with pylint checks
+# - ~100 files are flagged with import errors
+# - it's really slow
```
YinglinSun, line 11: How slow is it for you? it is pretty fast for me:

(ai-1775)yinsun@s07ln0001:~/src/daos$ time isort --jobs 8 --check-only .
Skipped 5 files

real    0m0.604s
user    0m0.956s
sys     0m0.282s

utils/githooks/pre-commit.d/20-isort.sh
```diff
@@ -4,6 +4,13 @@
 
 set -ue
 
+# Disable isort check because it doesn't work
+# - it needs to run with -rc
+# - it conflicts with pylint checks
```
YinglinSun, line 9: Interesting ... curious how my PR passed CI pylint check?

utils/cq/requirements.txt
```diff
@@ -11,4 +11,5 @@ isort
 paramiko
 pyenchant
 pylint
+toml
```
YinglinSun, line 14: ah good point. This is the problem. We don't run the exact same env.

utils/githooks/pre-commit.d/20-isort.sh
```diff
@@ -4,6 +4,13 @@
...
+# - ~100 files are flagged with import errors
```
YinglinSun, line 10: what did you run to see these errors?

utils/githooks/pre-commit.d/20-isort.sh
```diff
@@ -4,6 +4,13 @@
 
 set -ue
 
+# Disable isort check because it doesn't work
+# - it needs to run with -rc
```
YinglinSun, line 8: I used `--check-only` which is same as `-c`. but what is `-r`? cannot find that option.

daltonbohning: `-rc` is for recursive
```
$ isort --jobs 8 --check-only .
WARNING: Unable to parse file . due to [Errno 21] Is a directory: '/home/dbohning/daos'
```

utils/githooks/pre-commit.d/20-isort.sh
daltonbohning: It seems your PR ran with a different config/env. When I run `isort` nd update all files, it removes some `pylint: disable` lines, which causes the pylint check to fail

utils/githooks/pre-commit.d/20-isort.sh
daltonbohning: `real    0m3.815s`
But it seems it's because so many files are flagged with warnings for me

utils/cq/requirements.txt
daltonbohning: Right. I think the discrepancy here is a different env

utils/githooks/pre-commit.d/20-isort.sh
YinglinSun: I see. I think your isort is a bit old. `-rc` has been deprecated.

```
/apcv/shared/conda-envs/ai-1775/lib/python3.10/site-packages/isort/main.py:1227: UserWarning: W0501: The following deprecated CLI flags were used and ignored: -rc!
```

utils/githooks/pre-commit.d/20-isort.sh
YinglinSun: Should be due to isort version. I'm running 5.9:

```
(ai-1775)yinsun@s07ln0001:~/src/daos$ isort --version

                 _                 _
                (_) ___  ___  _ __| |_
                | |/ _/ / _ \/ '__  _/
                | |\__ \/\_\/| |  | |_
                |_|\___/\___/\_/   \_/

      isort your imports, so you don't have to.

                    VERSION 5.9.0
```

utils/cq/requirements.txt
YinglinSun: Yes, that's the problem. Let's disable isort for now until we figure out better solution. Everybody is running a different env. Going forward, I would suggest moving these code quality stuff to a GHA which runs before all other tests. That way we have a unified env for code quality.

YinglinSun: isort already provides an official GHA: https://pycqa.github.io/isort/docs/configuration/github_action.html

YinglinSun (dismissed):

utils/githooks/pre-commit.d/20-isort.sh
daltonbohning: Ah! That version doesn't report errors for me. Maybe we should just pin `isort==5.9` in `cq/requirements.txt` so anyone using the githooks gets the right version.
We've ran into similar issues with pylint and flake where we see "new" issues with different versions

utils/cq/requirements.txt
daltonbohning: I think definitely a GHA for this would be good. Then, a githook is just a convenience to catch before pushing.

utils/githooks/pre-commit.d/20-isort.sh
daltonbohning: You're right. My isort was 4.x :(

utils/cq/requirements.txt
```diff
@@ -7,7 +7,7 @@ clustershell
 ## flake8 6 removed --diff option which breaks flake precommit hook.
 ## https://github.com/pycqa/flake8/issues/1389 https://github.com/PyCQA/flake8/pull/1720
 flake8<6.0.0
-isort
+isort==5.9
```
daltonbohning, line 10: @YinglinSun Do you see any issues with pinning this? It ensures everyone uses the same check. Then, when we have a GHA we can pin to the same to avoid discrepancies

utils/githooks/pre-commit.d/20-isort.sh
YinglinSun: For the fresh setup, they will pull the latest version anyway. The issue is for existing env we don't have a way to force the contributors to upgrade.

daltonbohning: > For the fresh setup, they will pull the latest version anyway

For whatever reason, my setup initially pulled in a (cached?) 4.x version. I had to specifically request 5.x+

utils/cq/requirements.txt
YinglinSun: No concern. Should be good idea.

utils/githooks/pre-commit.d/20-isort.sh
```diff
@@ -4,17 +4,11 @@
...
-# - it conflicts with pylint checks
-# - ~100 files are flagged with import errors
-# - it's really slow
-exit 0
 
 if ! command -v isort > /dev/null 2>&1
 then
     echo "No isort checking, install isort command to improve pre-commit checks"
-    echo "python3 -m pip install isort"
+    echo "python3 -m pip install -r ./utils/cq/requirements.txt"
```
YinglinSun, line 11: 👍

utils/githooks/pre-commit.d/20-isort.sh
YinglinSun: Yeah might be from local cache I guess.

YinglinSun (approved):

utils/cq/requirements.txt
ashleypittman: Why 5.9 rather than 5.12 and what's the specific reason here?  I have an idea that I'm working on for making sure that versions are up-to-date and I'd rather address everything at the same time for simplicity.

daltonbohning: 5.9 only because that's what @YinglinSun initially used. If we go higher, we just need to make sure it's available on supported distros without special install

daltonbohning: Or, we need to make sure a higher version doesn't introduce new checks, and maybe pin a range or `>=`

ashleypittman: We shouldn't base this on the lower common denominator for all possible distros that developers could be using, if a developer wants a specific version they should pip install it (and they should probably be doing that anyway).  When pinning a version we should use the most recent version or have a reason not to and a plan to resolve it, plus we should have a process for updating the version over time.

daltonbohning:
Will defer to #13252

jolivier23 (approved):

wangshilong (approved):

PR #13252, opened 2023-10-30
DAOS-14559 ci: Update isort config and add GitHub action for it.

ashleypittman:

### Gatekeeper:

### DIFF
--- a/.github/workflows/flake.yml
+++ b/.github/workflows/flake.yml
@@ -47,10 +47,14 @@ jobs:
   isort:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - uses: actions/setup-python@v2
         with:
           python-version: 3.11
       - uses: isort/isort-action@master
         with:
           requirementsFiles: "requirements.txt utils/cq/requirements.txt"
+      - name: Run on SConstruct file.
+        run: isort --check-only SConstruct
+      - name: Run on build files.
+        run: find . -name SConscript | xargs isort
--- a/.github/workflows/flake.yml
+++ b/.github/workflows/flake.yml
@@ -2,6 +2,11 @@ name: Flake
 
 # Always run on Pull Requests as then this check can be marked as required.
 on:
+  push:
+    branches:
+      - master
+      - 'feature/*'
+      - 'release/*'
   pull_request:
 
 jobs:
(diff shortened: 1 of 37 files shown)


daltonbohning:
@ashleypittman FYI https://github.com/daos-stack/daos/pull/13247

.github/workflows/flake.yml
```diff
@@ -37,3 +42,23 @@ jobs:
           ignore: 'F821,W503,F841'
           max-line-length: '100'
           args: '--filename */SConscript, SConstruct'
+
+  # Run isort on the tree.
+  # This checks .py files only so misses SConstruct and SConscript files are not checked, rather
+  # for these files check them afterwards.  The output-filter will not be installed for this part
+  # so regressions will be detected but not annotated.
+  isort:
+    name: Isort check
```
daltonbohning, line 51: Does an isort check really belong in `flake.yml`?

.github/workflows/flake.yml
```diff
@@ -37,3 +42,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v3
+        with:
+          python-version: 3.11
+      - uses: isort/isort-action@master
```
daltonbohning, line 58: Is this checked in?

ashleypittman: this is a public action, it doesn't need to be checked in.

ashleypittman:
> @ashleypittman FWIW, I have plans to add an ftest-tag lint check at some point. Maybe we're due for a generic "lint" workflow?

What would `ftest-tag` do?

Initially we created lots of GitHub actions jobs and they're all listed at https://github.com/daos-stack/daos/actions even the ones which have only run once so creating lots of jobs with unique names is a bad idea IMO.  It can also take some time to schedule runners so I am thinking it's best if we actually just have fewer actions but each with multiple steps, it would likely schedule faster and therefore run faster although there is a down-side - you can only have 10 annotations per job, and like we see with codespell and the d_log_check if the first one fails then the second doesn't get run, although in that case they're seperate steps but the same job.

daltonbohning:
> > @ashleypittman FWIW, I have plans to add an ftest-tag lint check at some point. Maybe we're due for a generic "lint" workflow?
> 
> What would `ftest-tag` do?
> 
> Initially we created lots of GitHub actions jobs and they're all listed at https://github.com/daos-stack/daos/actions even the ones which have only run once so creating lots of jobs with unique names is a bad idea IMO. It can also take some time to schedule runners so I am thinking it's best if we actually just have fewer actions but each with multiple steps, it would likely schedule faster and therefore run faster although there is a down-side - you can only have 10 annotations per job, and like we see with codespell and the d_log_check if the first one fails then the second doesn't get run, although in that case they're seperate steps but the same job.

It would verify that tests have appropriate tags. e.g.
- each test needs a unique tag
- needs pr, daily, weekly, etc.
- ...

We can discuss this more when I'm ready to add one

.github/workflows/flake.yml
YinglinSun: maybe renaming this flake workflow to something more generic like `cq.yml`? that way we can include all cq related checks in one file.

PR #13262, opened 2023-10-31
DAOS-14571 dfuse: Increase pre-read buffer size.

ashleypittman:
Create a separate pre read buffer so that it is not tied to the
kernel buffer size, use 4Mb as size threshold for pre-read.
Pre-allocate buffers at startup, not first read.
Change to on-by-default for fresh directories.
Do not disable is file is opened but not read from - this could be
 the kernel cache doing it's job.


src/client/dfuse/dfuse.h
```diff
@@ -450,6 +454,7 @@ struct dfuse_pool {
 	ACTION(LISTXATTR)                                                                          \
 	ACTION(RENAME)                                                                             \
 	ACTION(OPEN)                                                                               \
+	ACTION(PRE_READ)                                                                           \
```
jolivier23, line 457: Is this new stat used anywhere?

src/client/dfuse/ops/open.c
```diff
@@ -100,7 +100,7 @@ dfuse_cb_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
 	/* Enable this for files up to the max read size. */
 	if (prefetch && oh->doh_parent_dir &&
 	    atomic_load_relaxed(&oh->doh_parent_dir->ie_linear_read) && ie->ie_stat.st_size > 0 &&
-	    ie->ie_stat.st_size <= DFUSE_MAX_READ) {
+	    ie->ie_stat.st_size <= DFUSE_MAX_PRE_READ) {
```
jolivier23, line 103: Maybe I'm confused about what st_size is here.  Is it the file size?  Shouldn't readahead only happen if the size is >= DFUSE_MAX_PRE_READ ?  Otherwise, wouldn't it cause dfs to do a short read?

jolivier23: ah, nevermind, this is just for small files

ashleypittman: Documentation for pre-read is at https://github.com/daos-stack/daos/blob/master/src/client/dfuse/dfuse.h#L97-L129

I have thought about extending this for larger files, as long as the file is read sequentially then the whole file could be treaded in a similar manner with reads being served from cache and and each read triggering the next one.  There's not a ticket for that yet though.

src/client/dfuse/dfuse.h
ashleypittman: No, other than display none of these are yet.

I've recently made these available via json and am working on adding checking the stats into the testing, https://github.com/daos-stack/daos/pull/13927 adds the ability for ftest to read these stats from a test.

src/client/dfuse/dfuse.h
```diff
@@ -82,12 +82,20 @@ struct dfuse_eq {
...
-/* Maximum size dfuse expects for read requests, this is not a limit but rather what is expected */
+/* Maximum size dfuse expects for read requests, this is not a limit but rather what is expected
+ * This is the maximum size expected from the kernel, increasing this without changing kernel
+ * behavior will have no effect.
+ */
 #define DFUSE_MAX_READ (1024 * 1024)
 
+/* Maximum file-size for pre-read requests.  This can be any value at the cost of
+ * memory consumption */
+#define DFUSE_MAX_PRE_READ (1024 * 1024 * 4)
```
mchaarawi, line 97: it would be great to make that configurable

src/client/dfuse/ops/open.c
mchaarawi: This is separate from this ticket, but yea having a read-ahead on large files so future reads get served from cache is probably worth exploring.

mchaarawi (commented):
the patch looks good to me, but this is one of the PRs that needs to show that it's working as intended with a manual performance test that for example creates a 4m file with FIO, then do like random 4k IO reads and show the latency is basically close to 0 as it should be read from cache.

src/client/dfuse/ops/open.c
ashleypittman: I agree, this is something we've discussed but really is a different feature from this.  I'll file a ticket.

src/client/dfuse/dfuse.h
ashleypittman: I don't really want to make this a per-container setting although it could be, that leave command line of probably environment variable as the best options.  I think it's going to be about tuning a sensible default rather than recommending users change the settings so I think an environment variable would probably be best here.

ashleypittman:
> the patch looks good to me, but this is one of the PRs that needs to show that it's working as intended with a manual performance test that for example creates a 4m file with FIO, then do like random 4k IO reads and show the latency is basically close to 0 as it should be read from cache.

Feature verified by using reading the dfuse statistics in a test.

ashleypittman:
The one test failure is DAOS-15616 which is broken currently.

src/tests/ftest/util/dfuse_utils.py
```diff
@@ -380,6 +381,20 @@ def stop(self):
...
+        cmd = f"daos filesystem query --json {self.mount_dir.value}"
+        result = run_remote(self.log, self.hosts, cmd)
+        if not result.passed:
+            raise CommandFailure(f'"{cmd}" failed on {result.failed_hosts}')
+
+        data = json.loads("\n".join(result.output[0].stdout))
+        print(data)
+        assert data["status"] == 0
+        assert data["error"] is None
+        return data["response"]
```
daltonbohning, line 396: This would only work on a single host, since it returns `output[0]`. Would different hosts potentially return different values? (then, I have further suggestions)

src/tests/ftest/dfuse/read.py
```diff
@@ -0,0 +1,96 @@
...
+        # commands via ssh but there's no harm in adding this.
+        time.sleep(1)
+
+        # Sample the stats, later on we'll check this.
+        data = self.dfuse.get_stats()
+
+        # Check that the inode has been evicted, and there's been no reads so far.
+        assert data["inodes"] == 1, data
+        assert data["statistics"].get("read", 0) == 0, data
+        assert data["statistics"].get("pre_read", 0) == 0, data
```
daltonbohning, line 81: We should use the avocado asserts so this gets propagated properly for failures. The format is
```
self.assertEqual(val1, val2, "message that will propagate to the failure and ultimately the bug ticket")
```

src/tests/ftest/dfuse/read.py
```diff
@@ -0,0 +1,96 @@
...
+        :avocado: tags=dfuse
+        :avocado: tags=DFuseReadTest,test_dfuse_pre_read
+        """
+
+        pool = self.get_pool(connect=False)
+        container = self.get_container(pool)
+
+        self.load_dfuse(self.hostlist_clients, None)
+
+        self.dfuse.disable_wb_cache.value = True
```
daltonbohning, line 32: If this is always on, it's better to set in the config as
```
dfuse:
  disable_wb_cache: true
```

src/tests/ftest/dfuse/read.py
```diff
@@ -0,0 +1,96 @@
...
+
+        self.dfuse.env["D_LOG_MASK"] = "INFO,DFUSE=DEBUG"
+        self.dfuse.env["DD_MASK"] = "ALL"
+        self.dfuse.env["DD_SUBSYS"] = "ALL"
```
daltonbohning, line 36: Similarly, we can set these in the config as
```
dfuse:
  client:
    env_vars:
    - D_LOG_MASK=INFO,DFUSE=DEBUG
    - DD_MASK=ALL
    - DD_SUBSYS=ALL
```

src/tests/ftest/dfuse/read.py
ashleypittman: Done.

src/tests/ftest/dfuse/read.py
ashleypittman: We probably don't need these anymore now the test is complete/working.  I'll remove but yes, I would have put these in the config if I knew the correct syntax.

src/tests/ftest/util/dfuse_utils.py
ashleypittman: Good point.

This is the dfuse I/O counters so yes, each host would have different values.  Do you mind if for now we just check that len(self.hosts) == 1?

daltonbohning: Yeah, that's the simplest solution for now and we can revisit if we need this for multiple hosts

src/tests/ftest/dfuse/read.yaml
```diff
@@ -0,0 +1,22 @@
+hosts:
+  test_servers: 1
```
daltonbohning, line 2: Better to explicitly set clients. One day we need to only use localhost as the runner
```suggestion
hosts:
  test_servers: 1
  test_clients: 1
```

src/tests/ftest/util/dfuse_utils.py
daltonbohning: We should remove `print(data)` because (1) we should use `self.log.info` or `self.log.debug` instead, and (2) `run_remote` already prints all the output

daltonbohning: We shouldn't use direct `assert` here either. And in this case, I think we can raise a `CommandFailure` if the structure isn't correct

mchaarawi (approved):

daltonbohning (commented):
ftest LGTM

jolivier23 (approved):

ashleypittman:
The only failure is DAOS-15616 which is known broken at the minute and I have other 2.6 blockers which depend on this PR, can it be force-landed please.

PR #13263, opened 2023-10-31
DAOS-12961 cart: return error if crt_provider_get_ctx_idx() failed

wangshilong:
Fix to return error for crt_provider_get_ctx_idx(), so we don't need
call crt_provider_put_ctx_idx() if it failed to avoid assertion in crt_finalize().

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/cart/crt_context.c
+++ b/src/cart/crt_context.c
@@ -223,10 +223,10 @@ crt_context_provider_create(crt_context_t *crt_ctx, crt_provider_t provider, boo
 	}
 
 	D_RWLOCK_WRLOCK(&crt_gdata.cg_rwlock);
-	max_ctx_num = crt_provider_get_max_ctx_num(primary, provider);
 	ctx_idx = crt_provider_get_ctx_idx(primary, provider);
 
-	if (ctx_idx < 0 || ctx_idx >= max_ctx_num) {
+	if (ctx_idx < 0) {
+		max_ctx_num = crt_provider_get_max_ctx_num(primary, provider);
 		D_WARN("Provider: %d; Context limit (%d) reached\n",
 		       provider, max_ctx_num);
 		D_RWLOCK_UNLOCK(&crt_gdata.cg_rwlock);
--- a/src/cart/crt_hg.c
+++ b/src/cart/crt_hg.c
@@ -619,7 +619,7 @@ crt_provider_get_ctx_idx(bool primary, int provider)
 	struct crt_prov_gdata	*prov_data = crt_get_prov_gdata(primary, provider);
 	int			i;
 
-	for (i = 0; i < CRT_SRV_CONTEXT_NUM; i++) {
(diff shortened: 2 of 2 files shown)


wiliamhuang (approved):

src/cart/crt_context.c
```diff
@@ -229,6 +229,8 @@ crt_context_provider_create(crt_context_t *crt_ctx, crt_provider_t provider, boo
 	if (ctx_idx < 0 || ctx_idx >= max_ctx_num) {
 		D_WARN("Provider: %d; Context limit (%d) reached\n",
 		       provider, max_ctx_num);
+		if (ctx_idx >=0)
+			crt_provider_put_ctx_idx(primary, provider, ctx_idx);
```
frostedcmos, line 233: This does not look correct to me. if we hit this case, means ctx_idx was >= max_ctx_num, in other words it was invalid.

We dont need to put invalid context back, and doing so today in fact might cause a crash as we access array at the invalid index.

603 void
 604 crt_provider_put_ctx_idx(bool primary, int provider, int idx)
 605 {
 606         struct crt_prov_gdata *prov_data = crt_get_prov_gdata(primary, provider);
 607
 608         if (prov_data->cpg_used_idx[idx] == false) {

frostedcmos (changes requested):

wiliamhuang:
@frostedcmos https://github.com/daos-stack/daos/blob/master/src/cart/crt_init.c#L217-L223
It seems that max_num_ctx (or prov_data->cpg_ctx_max_num) could be smaller than CRT_SRV_CONTEXT_NUM. If this is true, the return value of crt_provider_get_ctx_idx() could be larger than max_ctx_num.

frostedcmos:
> @frostedcmos https://github.com/daos-stack/daos/blob/master/src/cart/crt_init.c#L217-L223 It seems that max_num_ctx (or prov_data->cpg_ctx_max_num) could be smaller than CRT_SRV_CONTEXT_NUM. If this is true, the return value of crt_provider_get_ctx_idx() could be larger than max_ctx_num.

Good point there, however this then tells me we should fix crt_provider_get_ctx_idx() instead and return -1 for any invalid case.  The overall idea is that if 'get' returns an error, then put should not be required.

So for get(), i would say it should only loop up to prov_data->cpg_ctx_max_num instead of going through CRT_SRV_CONTEXT_NUM.

wiliamhuang:
> So for get(), i would say it should only loop up to prov_data->cpg_ctx_max_num instead of going through CRT_SRV_CONTEXT_NUM.

I agree. Thank you!

wangshilong:
Thanks @frostedcmos and @wiliamhuang , sorry i missed to force-update the PR,  but it might be ok since change is trival.

frostedcmos (approved):

wiliamhuang (approved):

PR #13264, opened 2023-10-31
DAOS-14251 control: Allow premounted empty tmpfs (#12968)

mjmac:
In the special case where the tmpfs has already been
mounted but is empty, don't skip NVMe format and
configuration. Enables the use case of running
daos_server in a container with an external tmpfs.

### DIFF
--- a/src/control/cmd/daos_server/start.go
+++ b/src/control/cmd/daos_server/start.go
@@ -64,7 +64,9 @@ func (cmd *startCmd) setCLIOverrides() error {
 	if cmd.Modules != nil {
 		cmd.config.WithModules(*cmd.Modules)
 	}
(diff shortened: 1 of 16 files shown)


kjacque (approved):

mjmac:
@kjacque: Can you see what the hw-medium error is? If it looks transient, would you please restart it?

kjacque:
> @kjacque: Can you see what the hw-medium error is? If it looks transient, would you please restart it?

I restarted it. Not clear to me from the job log what actually failed. The supposed failure was had to do with cart_logtest.py, but I didn't see anything relating to that in the logs.

kjacque:
The latest failure is https://daosio.atlassian.net/browse/DAOS-14352. Since it's a known issue awaiting backport to the branch, I think this should be force landed.

PR #13296, opened 2023-11-06
DAOS-14105 object: misc patch for collectively punch/query object

Nasf-Fan:
Collectively punch object:

Currently, when punch an object with multiple redundancy groups,
to guarantee the atomicity, we handle the whole punch via single
internal distributed transaction. The DTX leader will forward the
CPD RPC to every object shard within the same transaction. For a
large-scaled object, such as a SX object, punching it will generate
N RPCs (N is equal to the count of all the vos targets in the system).
That will be very slow and hold a lot of system resource for relative
long time. If the system is under heavy load, related RPC(s) may get
timeout, then trigger DTX abort, and then client will resend RPC to
the DTX leader for retry, that will make the situation to be worse
and worse.

To resolve such bad situation, we will collectively punch the object.

The basic idea is that: when punch an object with multiple redundancy
groups, the client will scan the object layout and generate bitmap and
targets information for each DAOS engine that have object shards on it.
And then client will send OBJ_COLL_PUNCH RPC to the DTX leader. These
bitmap and targets information may be too large to be transferred via
RPC body, then they will be sent to the DTX leader via RDAM. On DTX
leader side, it will not directly forward the request to all related
engines, instead, the DTX leader will split the engines into multiple
engine groups. For engine group, it will randomly choose a relay engine
that will help current engine to dispatch the OBJ_COLL_PUNCH RPC to the
others in the same engine group. From current engine's perspective, for
each engine group, it only forwards one OBJ_COLL_PUNCH RPC to the relay
engine in such engine group with the bitmap and targets information that
are only related with the engines in such engine group. These method can
be recursively used by the relay engines until all related engines have
received the collective punch request.

We control the count of engine groups on the DTX leader to guarantee that
the size of related bitmap and targets information for each engine group
will not exceed RPC bulk threshold. Then there will be no RDAM among the
engines for collectively punching the object.

For each related DAOS engine, the local punch for related object shards
will be driven via collective task on each own local vos targets. That
will save a lot of RPCs and resources.

We allow the user to set object collective punch threshold via client
side environment "DAOS_OBJ_COLL_PUNCH_THD". The default (and also the
min) value is 31.

On the other hand, for large-scaled object, transferring related DTX
participants information (that will be huge) will be heavy burden in
spite of via RPC body or RDMA. So OBJ_COLL_PUNCH RPC will not transfer
completed dtx_memberships to DAOS engines, instead, related engines in
spite leader or not, will append each local shards bitmap and targets
information to the client given MBS header. That is enough for commit
and abort collectively. As for DTX resync, the (new) DTX leader needs
to re-calculate the completed MBS. Since it is relative rare, even if
the overhead for such re-calculation would be quite high, it will not
affect the whole system too much.

Collectively query object:

Currently, get file size (query key) for large-scaled object is very slow.
Because DAOS does not has logic (metadata) center to store the file size.
The client needs to send query RPCs to all related redundancy groups, then
aggregate related query results. For EC object with parity rotation, it is
worse, the client has to send query RPCs to all shards in every redundancy
group. It will cause a lot of query RPCs. For large-scaled object (such as
the "GX" object class), current method is too heavy loaded for both client
and servers.

To resolve such bad situation, we introduce new mechanism: collective query.
The basic idea is that: before sending query RPCs to related engines, based
on the shards to be queried, the client will generate the bitmap for related
VOS targets on each involved engine. For each engine with non-empty bitmap,
the client only sends one OBJ_COLL_QUERY RPC to it, then the engine generates
collective tasks (based on the bitmap) to query related object shards on each
own local VOS targets. That can save many query RPCs if multiple VOS targets
reside on relative concentrated engines.

On the other hand, it is inefficient for single client to send out hundreds
or even thousands of query RPCs concurrently, that will cause a lot of DRAM
resource being occupied for relative long time. To speedup, once the RPCs
count exceeds some threshold, the client will ask some engine(s) to help to
forward the collective query RPC to other related engine(s), and reply the
aggregated query results to the client. From client perspective, forwarding
causes one additional RPC round-trip, but it is better than single client
handling hundreds or thousands of query RPCs by itself.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/container/srv_target.c
+++ b/src/container/srv_target.c
@@ -1653,6 +1653,8 @@ ds_cont_tgt_open(uuid_t pool_uuid, uuid_t cont_hdl_uuid,
 	struct dss_coll_ops	coll_ops = { 0 };
 	struct dss_coll_args	coll_args = { 0 };
 	struct ds_pool		*pool;
+	int			*exclude_tgts = NULL;
+	uint32_t		exclude_tgt_nr = 0;
 	int			rc;
 
 	/* Only for debugging purpose to compare srv_cont_hdl with cont_hdl_uuid */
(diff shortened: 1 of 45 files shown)


Nasf-Fan:
> Test stage Functional Hardware Medium completed with status FAILURE. https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-13296/15/execution/node/1344/log

Dfuse test failure caused by DAOS-14815, not related with the patch.

Nasf-Fan:
> Test stage Functional Hardware Medium completed with status FAILURE. https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-13296/22/execution/node/1406/log

osa_online_drain_no_csum failed because of rebuild timeout, not related with the patch.

PR #13330, opened 2023-11-13
DAOS-14636 dfuse: Enable symlink caching.

ashleypittman:
Required-githooks: true


daltonbohning (approved):

mchaarawi (approved):
could use adding some unit tests?

ashleypittman:
> could use adding some unit tests?

I think it would be difficult as you're setting kernel behaviour/settings.  I've verified this from the stats counters, on one test READLINK went from 45% of all requests down to 6.

PR #13331, opened 2023-11-13
DAOS-623 cq: Upgrade yamllint package

ashleypittman:
Required-githooks: true

### DIFF
--- a/.github/workflows/version-checks.yml
+++ b/.github/workflows/version-checks.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    paths:
+      - 'utils/cq/requirements.txt'
 
 jobs:
   upgrade-check:
--- a/docs/QSG/build_from_scratch.md
+++ b/docs/QSG/build_from_scratch.md
@@ -83,7 +83,7 @@ PATH.
  ```bash
  $ python3 -m venv venv
  $ source venv/bin/activate
- $ python3 -m pip install -f requirements.txt
+ $ python3 -m pip install -r requirements.txt
  ```
 
 ## Build DAOS
--- a/utils/cq/requirements.txt
+++ b/utils/cq/requirements.txt
@@ -11,5 +11,5 @@ pyenchant
 flake8<6.0.0
 isort==5.12
 pylint==3.0.2
-yamllint==1.32.0
+yamllint==1.33.0
 codespell==2.2.6


brianjmurrell (approved):
Nice to see this working.  Did anyone get an alert about the `version-checks.yml` failures on master or did it take somebody being proactive and looking to see that it was failing?

ashleypittman:
> Nice to see this working. Did anyone get an alert about the `version-checks.yml` failures on master or did it take somebody being proactive and looking to see that it was failing?

People whose PRs land are getting emails about it.

You can also see it via the red dot next to commits at the main daos github page https://github.com/daos-stack/daos and through the actions page at https://github.com/daos-stack/daos/actions/workflows/version-checks.yml

I'm sure there are yet more improvements to come however.

daltonbohning (approved):
We should consider backported to 2.4 after https://github.com/daos-stack/daos/pull/13322. Otherwise, the checks could be out of sync there too.

Something else I'm realizing is that githooks don't necessarily use the newer versions. Maybe we need a githook to check that the currently installed versions are the "recommended"?

ashleypittman:
> We should consider backported to 2.4 after #13322. Otherwise, the checks could be out of sync there too.

Personally I think it's OK if the checks there are a little behind, code should be going through master anyway.  The version_checks job is only on master for now.

> Something else I'm realizing is that githooks don't necessarily use the newer versions. Maybe we need a githook to check that the currently installed versions are the "recommended"?

I don't think that matters.  Users should update periodically but unless they hit a failure there's probably no need to check.

daltonbohning:
> > Something else I'm realizing is that githooks don't necessarily use the newer versions. Maybe we need a githook to check that the currently installed versions are the "recommended"?
> 
> I don't think that matters. Users should update periodically but unless they hit a failure there's probably no need to check.

But if they hit a failure they might not necessarily know it's due to version mismatch locally and in the GHA. Anyway, not a huge deal

ashleypittman:
> > > Something else I'm realizing is that githooks don't necessarily use the newer versions. Maybe we need a githook to check that the currently installed versions are the "recommended"?
> > 
> > 
> > I don't think that matters. Users should update periodically but unless they hit a failure there's probably no need to check.
> 
> But if they hit a failure they might not necessarily know it's due to version mismatch locally and in the GHA. Anyway, not a huge deal

If we assume that newer versions enable more checks then if they've got an old version then they'd miss some checks locally that were caught in GHA, that should be rare and checking you're up-to-date if that happens isn't onerous.

The only problem would be is if a program removed a check, then we updated the codebase to violate that check, it's possible but I'm not going to loose any sleep over it.

PR #13332, opened 2023-11-13
DAOS-14637 dfuse: Avoid use-after-free in open.

ashleypittman:
Use a stack variable rather than oh to determine if feature is on.

Required-githooks: true


mchaarawi (approved):

daltonbohning (approved):

ashleypittman:
The affected code is modified in https://github.com/daos-stack/daos/pull/13073 so the fix is also in that PR which is also approved so I'd prefer to land that one and close this.  I'll keep them both open pending CI results however.

PR #13333, opened 2023-11-13
DAOS-14639 dfuse: Do not allow security. xattr to be set.

ashleypittman:
fuse will call this often to read non-existent xattrs for every write request
so short-circuit these to avoid server round-trips.

Required-githooks: true
Test-tag: dfuse

### DIFF
--- a/src/tests/ftest/dfuse/daos_build.py
+++ b/src/tests/ftest/dfuse/daos_build.py
@@ -216,7 +216,9 @@ class DaosBuild(DfuseTestBase):
                 'daos filesystem query {}'.format(mount_dir),
                 'daos filesystem evict {}'.format(build_dir),
                 'daos filesystem query {}'.format(mount_dir),
-                'scons -C {} --jobs {}'.format(build_dir, intercept_jobs)]
+                'scons -C {} --jobs {}'.format(build_dir, intercept_jobs),
+                'scons -C {} --jobs {} install'.format(build_dir, intercept_jobs),
+                'daos filesystem query {}'.format(mount_dir)]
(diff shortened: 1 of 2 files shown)


src/tests/ftest/dfuse/daos_build.py
```diff
@@ -216,7 +216,9 @@ def run_build_test(self, cache_mode, intercept=False, run_on_vms=False):
                 'daos filesystem query {}'.format(mount_dir),
                 'daos filesystem evict {}'.format(build_dir),
                 'daos filesystem query {}'.format(mount_dir),
-                'scons -C {} --jobs {}'.format(build_dir, intercept_jobs)]
+                'scons -C {} --jobs {}'.format(build_dir, intercept_jobs),
+                'scons -C {} --jobs {} install'.format(build_dir, intercept_jobs),
+                'daos filesystem query {}'.format(mount_dir)]
```
daltonbohning, line 221: So it was previously only building but not installing? And does installing now somehow cover these tests changes?

ashleypittman: Yes, it was previously building but not installing.

The real change in this file is calling `daos filesystem query` at the end of the job to see the statistics, having that data tells us what dfuse is doing which is how I found this issue.  Not doing an install was an oversight but should just be a bunch of renames so I don't expect problems with it.

daltonbohning (dismissed):

mchaarawi (approved):

wiliamhuang (approved):

daltonbohning (approved):

PR #13349, opened 2023-11-16
DAOS-13280 client: intercept libaio functions

wiliamhuang:
Intercept libaio related functions, io_setup, io_destroy, io_submit, io_cancel, and io_getevents.
Performance data on shared performance cluster can be found at https://daosio.atlassian.net/browse/DAOS-13280

Features: dfuse pil4dfs

Required-githooks: true

### Gatekeeper:


wiliamhuang:
I will bump up the release number and add change history about adding the dependency on libaio later (after all CI tests finish).

utils/rpms/daos.spec
```diff
@@ -65,6 +65,7 @@ BuildRequires: protobuf-c-devel
 BuildRequires: lz4-devel
 BuildRequires: capstone-devel
 %endif
+BuildRequires: libaio-devel
```
daltonbohning, line 68: This change requires bumping the release and a changelog entry.

debian/control
```diff
@@ -32,6 +32,7 @@ Build-Depends: debhelper (>= 10),
                libraft-dev (= 0.10.1-1409.gc354cd7),
                python3-tabulate,
                liblz4-dev,
+               libaio-dev,
```
daltonbohning, line 35: Also need a debian/changelog entry

daltonbohning (changes requested):

utils/rpms/daos.spec
wiliamhuang: Thank you! Yes. I will add it later once current CI tests pass and get more feedback

debian/control
wiliamhuang: Thank you!  I will add it later once current CI tests pass and get more feedback.

src/tests/ftest/dfuse/bash.py
```diff
@@ -138,9 +138,12 @@ def run_bashcmd(self, il_lib=None, compatible_mode=False):
                             f"cksum {fuse_root_dir}/src.c",
                             f"bzip2 -z {fuse_root_dir}/lib.a",
                             f"chmod u-r {fuse_root_dir}/lib.a.bz2",
-                            "fio --readwrite=randwrite --name=test --size=\"2M\" --directory "\
-                            f"{fuse_root_dir}/ --bs=1M --numjobs=\"4\" --ioengine=psync "\
-                            "--group_reporting --exitall_on_error --continue_on_error=none",
+                            'fio --readwrite=randwrite --name=test --size="2M" --directory '\
```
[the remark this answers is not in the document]
mchaarawi: furthermore, another ticket is needed to compare the FIO performance of this against DFS with varying io_depth.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5495,6 +5527,332 @@ register_handler(int sig, struct sigaction *old_handler)
...
+		D_ASSERT(next_io_setup != NULL);
+	}
+
+	rc = next_io_setup(maxevents, ctxp);
+	if (rc < 0)
+		return rc;
+
+	allocate_aio_ctx(*ctxp, maxevents);
+	/* we ignore the result value here since we are not sure whether DFS files will
+	 * be involved or not.
```
mchaarawi, line 5615: I don't understand this comment. please clarify

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5495,6 +5527,332 @@ register_handler(int sig, struct sigaction *old_handler)
 	}
 }
 
+static int
+allocate_aio_ctx(io_context_t ctx, int depth)
+{
+	int i;
+
+	pthread_mutex_lock(&lock_aio);
+	if (aio_ctx_count >= MAX_AIO_CTX)
```
mchaarawi, line 5536: do we are limiting to 8? why?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5495,6 +5527,332 @@ register_handler(int sig, struct sigaction *old_handler)
...
+		next_io_submit = dlsym(RTLD_NEXT, "io_submit");
+		D_ASSERT(next_io_submit != NULL);
+	}
+	if (!hook_enabled)
+		return next_io_submit(ctx, nr, ios);
+	io_depth = query_aio_ctx_depth(ctx, &ctx_idx);
+	if (io_depth == 0)
+		return next_io_submit(ctx, nr, ios);
+	if (nr > io_depth) {
+		DS_ERROR(EINVAL, "number of ops submitted is larger than queue size");
```
mchaarawi, line 5701: is this correct? EINVAL?
don't we submit the ones we can and return that?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5495,6 +5527,332 @@ register_handler(int sig, struct sigaction *old_handler)
...
+	}
+	if (n_op_dfs == 0)
+		return next_io_submit(ctx, nr, ios);
+	if (n_op_dfs != nr) {
+		DS_ERROR(EINVAL, "io_submit() does not support mixed non-dfs and dfs files yet");
+		return (-EINVAL);
+	}
+
+	/* assume all io requests are over DFS for now. */
+	rc = create_eq_ev_for_aio(ctx_idx, io_depth);
```
mchaarawi, line 5718: for every io_submit you create a separate EQ? that sounds expensive..
but from the lack of comments in this PR, it's hard to understand.

mchaarawi (changes requested):
Im sorry, but this is really hard to follow and understand. there are no comments in this PR to explain the implementation of how the DAOS EQ and events map to the aio requests.  please add more comments to explain.

src/tests/ftest/dfuse/bash.py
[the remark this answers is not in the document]
wiliamhuang: Another ticket https://daosio.atlassian.net/browse/DAOS-14657 for fio on hw with more threads / procs and access size.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: When applications call io_setup(), we do not know whether the files on DFS will be accessed or not. File descriptor will be passed in io_submit() later. Even allocate_aio_ctx() fails, we ignore it here and return the return code of next_io_setup(maxevents, ctxp).

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: I am not familiar to the use cases of libaio. Not sure how large the limit should be. Sure. We can bump up the default limit and revise the code to adapt larger limit. We could set an env variable to allow user to modify the limit.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: You are right. I will fix it. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: my question was why you set this to 8 and if i understood your answer, it was just random?
I think this is something that needs more investigation before it can be set it seems.

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: i still don't understand what this has to do with ignoring a failure on allocation..

wiliamhuang: ok. A warning message could be printed although the failure on allocation should not change the return code of next_io_setup().

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Yes. I picked it randomly in the preliminary implementation.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: At the very beginning, I hoped to share EQs across aio contexts. To make io_getevents() easier to implement, separate EQs are created right now. Consequently, I set a low limit for the number of aio context. I will think more how we implement io_getevents to use shared EQs.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Currently separate EQs are used for each aio context, so I randomly picked a small number.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -111,26 +110,48 @@ static daos_handle_t          eq_list[MAX_EQ];
...
+struct aio_ev {
+	daos_event_t    ev;
+	struct iocb    *piocb;
+	struct aio_ctx *ctx;
 };
 
 struct aio_ctx {
+	/* the real io_context_t used in libaio */
 	io_context_t          ctx;
+	/* the depth of contex set by io_setup */
```
daltonbohning, line 138: codespell
```suggestion
	/* the depth of context set by io_setup */
```

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -111,26 +110,48 @@ static daos_handle_t          eq_list[MAX_EQ];
...
 	int                   depth;
-	bool                  on_dfs;
+	int                   first_eq;
 	bool                  inited;
-	/* eqs are usable only for current context */
-	struct aio_ctx_eq_ev *eq_ev_list;
+	bool                  on_dfs;
+	_Atomic uint64_t      num_op_submitted;
+	_Atomic uint64_t      num_op_done;
+	/* DFS is involved or not for current contex */
```
daltonbohning, line 145: codespell
```suggestion
	/* DFS is involved or not for current context */
```
```suggestion
	/* DFS is involved or not for current contex */
```

daltonbohning (commented):
ftest LGTM, but I'm not familiar enough to approve the rest

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! I will update it in next commit.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: I will update it in next commit. Thank you!

debian/control
wiliamhuang: debian/changelog entry was added.

utils/rpms/daos.spec
wiliamhuang: Release number was bumped up.

brianjmurrell (commented):
Packaging changes LGTM.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Typo is fixed now. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Typo is fixed now. Thank you!

src/client/dfuse/il/int_posix.c
```diff
@@ -802,7 +802,9 @@ child_hdlr(void)
 {
 	int rc;
 
-	daos_eq_lib_reset_after_fork();
+	rc = daos_eq_lib_reset_after_fork();
+	if (rc)
+		DFUSE_LOG_WARNING("daos_eq_lib_init() failed in child process: " DF_RC, DP_RC(rc));
```
ashleypittman, line 807: Nit, we're trying to move away from these older dfuse specific logging macros.
```suggestion
	if (rc)
		DL_WARN(rc, "daos_eq_lib_init() failed in child process");
```

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -110,6 +111,48 @@ static uint16_t               eq_count_max;
 static uint16_t               eq_count;
 static uint16_t               eq_idx;
 
+/* the number of EQs for aio contexts. (aio_eq_count + eq_count) <= eq_count_max */
+static uint16_t               aio_eq_count;
+
+struct aio_eq {
```
ashleypittman, line 117: A `d` or similar prefix for this struct would make it more obvious that this is something local when it's declared later on.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -110,6 +111,48 @@ static uint16_t               eq_count_max;
...
+	io_context_t     ctx;
+	/* the depth of context set by io_setup */
+	int              depth;
+	int              first_eq;
+	bool             inited;
+	bool             on_dfs;
+	_Atomic uint64_t num_op_submitted;
+	_Atomic uint64_t num_op_done;
+	/* DFS is involved or not for current context */
+	pthread_mutex_t  lock;
```
ashleypittman, line 146: Does this comment match the entry or not?  The comment implies that it's a boolean but the entry is a lock?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -110,6 +111,48 @@ static uint16_t               eq_count_max;
...
+	/* The list of finished ev. ev is removed from the list by io_getevents(). */
+	struct aio_ev  **ev_done_list;
```
ashleypittman, line 148: A list or an array?  I'm still working through the PR so I don't know what this is yet but the comment doesn't match the code.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+create_ev_eq_for_aio(aio_ctx_t *aio_ctx);
+int
+io_setup(int maxevents, io_context_t *ctxp)
+{
+	int        rc;
+	aio_ctx_t *aio_ctx_obj;
+
+	if (next_io_setup == NULL) {
+		next_io_setup = dlsym(RTLD_NEXT, "io_setup");
+		D_ASSERT(next_io_setup != NULL);
```
ashleypittman, line 6436: Is it possible to use pil4dfs with an appliction that does not link libaio, and in which case would this fail?  Perhaps if the user tried that then libaio would be linked anyway as pil4dfs depends on it?

The use of assert here concerns me but it maybe a non-issue.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+	}
+
+	rc = next_io_setup(maxevents, ctxp);
+	if (rc < 0)
+		return rc;
+
+	D_ALLOC(aio_ctx_obj, sizeof(aio_ctx_t));
```
ashleypittman, line 6443: This is a bug I think, it should read as below so that the allocation is the size of the struct rather than the pointer.  To avoid this call of bug we have the `D_ALLOC_PTR` macro which does the work for you, please use it in this case.
```suggestion
	D_ALLOC(aio_ctx_obj, sizeof(*aio_ctx_t));
```
```suggestion
	D_ALLOC_PTR(aio_ctx_obj);
```

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+	if (aio_ctx_obj == NULL) {
+		io_destroy(*ctxp);
+		return -ENOMEM;
+	}
+	aio_ctx_obj->ctx   = *ctxp;
+	aio_ctx_obj->depth = maxevents;
+	atomic_store_relaxed(&aio_ctx_obj->num_op_submitted, 0);
+	atomic_store_relaxed(&aio_ctx_obj->num_op_done, 0);
```
ashleypittman, line 6451: nit: There's an atomic_init() function which would be clearer here.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+		sgl.sg_nr     = 1;
+		sgl.sg_iovs   = &iov;
+		ctx_ev->piocb = ios[i];
+		/* EQs are shared by contexts. Need to save ctx when polling EQs. */
+		ctx_ev->ctx = aio_ctx_obj;
+		if (op == IO_CMD_PREAD) {
+			rc = dfs_read(file_list[fd]->dfs_mt->dfs, file_list[fd]->file, &sgl,
+				      ios[i]->u.c.offset, &read_size, &ctx_ev->ev);
+			if (rc)
+				D_GOTO(err, rc);
```
ashleypittman, line 6627: Are there read requests lost on this error path?  If the I/O has been submitted then the reads could continue after this function but the user could presumably assume the buffer is not in use and free/re-assign it?

src/tests/ftest/dfuse/bash.py
```diff
@@ -126,6 +126,9 @@ def run_bashcmd(self, il_lib=None, compatible_mode=False):
             'fio --readwrite=randwrite --name=test --size="2M" --directory '
             f'{fuse_root_dir}/ --bs=1M --numjobs="4" --ioengine=psync '
             "--group_reporting --exitall_on_error --continue_on_error=none",
+            'fio --readwrite=randwrite --name=test --size="2M" --directory '
+            f'{fuse_root_dir}/ --bs=1M --numjobs="1" --ioengine=libaio --iodepth=4'
+            '--group_reporting --exitall_on_error --continue_on_error=none',
```
ashleypittman, line 131: It would be nicer although not essential to have a dedicated test for aio.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -564,6 +608,13 @@ static int (*next_posix_fallocate64)(int fd, off64_t offset, off64_t len);
 static int (*next_tcgetattr)(int fd, void *termios_p);
 /* end NOT supported by DAOS */
 
+static int (*next_io_setup)(int maxevents, io_context_t *ctxp);
+static int (*next_io_destroy)(io_context_t ctx);
+static int (*next_io_submit)(io_context_t ctx, long nr, struct iocb *ios[]);
+static int (*next_io_cancel)(io_context_t ctx, struct iocb *iocb, struct io_event *evt);
+static int (*next_io_getevents)(io_context_t ctx, long min_nr, long nr, struct io_event *events,
+				struct timespec *timeout);
```
mchaarawi, line 616: it's about time to start creating new files for new features / sets of functions to intercept rather than keep piling on int_dfs.c

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+	return next_io_destroy(ctx_save);
+}
+
+static int
+create_ev_eq_for_aio(aio_ctx_t *aio_ctx)
+{
+	int rc, i, j, idx;
+	int num_aio_eq_free, num_aio_eq_create;
+
+	D_MUTEX_LOCK(&lock_aio);
```
mchaarawi, line 6498: you are not locking the aio context here, but this is a global lock to manage EQs for aio context, so please rename that to aio_eqs_lock_g

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -110,6 +111,48 @@ static uint16_t               eq_count_max;
...
+	daos_handle_t    eq;
+	pthread_mutex_t  lock;
+	_Atomic uint64_t n_op_queued;
+	_Atomic uint64_t n_op_done;
+};
+
+/* list of EQs dedicated for aio contexts. */
+static struct aio_eq aio_eq_list[MAX_EQ];
+/* the accumulated iodepth for aio context. */
+static long int      depth_sum_accu;
```
mchaarawi, line 127: I do not understand what you mean here by accumulated IO depth.
and if this is on an aio context, why is it global? i think the comment is wrong

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -110,6 +111,48 @@ static uint16_t               eq_count_max;
 static uint16_t               eq_count;
 static uint16_t               eq_idx;
 
+/* the number of EQs for aio contexts. (aio_eq_count + eq_count) <= eq_count_max */
+static uint16_t               aio_eq_count;
```
mchaarawi, line 115: the name also is confusing here, so would suggest to rename to:
aio_eqs_count_g

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: perhaps you should consider looking into using d_list_t ?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+		aio_ctx->inited          = true;
+	}
+	aio_ctx->on_dfs = true;
+	D_MUTEX_UNLOCK(&aio_ctx->lock);
+
+	return 0;
+
+free_eq:
+	for (j = 0; j < i; j++) {
+		daos_eq_destroy(aio_eq_list[idx + j].eq, 0);
```
mchaarawi, line 6544: check error condition for coverity

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+		D_ASSERT(next_io_submit != NULL);
+	}
+	if (!hook_enabled)
+		return next_io_submit(ctx_real, nr, ios);
+	io_depth = aio_ctx_obj->depth;
+	if (io_depth == 0)
+		return next_io_submit(ctx_real, nr, ios);
+
+	n_op_dfs = 0;
+	for (i = 0; i < nr; i++) {
```
mchaarawi, line 6576: you execute this for loop three times in this function.. can you combine them?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+		op         = ios[i]->aio_lio_opcode;
+		fd         = ios[i]->aio_fildes - FD_FILE_BASE;
+		idx_aio_eq = (aio_ctx_obj->first_eq + i) % aio_eq_count_local;
+
+		D_ALLOC_PTR(ctx_ev);
+		if (ctx_ev == NULL) {
+			if (i == 0)
+				return -ENOMEM;
+			else
+				return i;
```
mchaarawi, line 6608: is it valid to return success in case of an ENOMEM / allocation failure?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+		}
+
+		rc = daos_event_init(&ctx_ev->ev, aio_eq_list[idx_aio_eq].eq, NULL);
+		if (rc) {
+			D_FREE(ctx_ev);
+			DL_ERROR(rc, "daos_event_init() failed");
+			D_GOTO(err, rc = daos_der2errno(rc));
```
mchaarawi, line 6615: so above you returned success / num io submitted on a failure, but here you disregard i and return error always on failure.. so there is inconsistency in behavior.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+		/* dequeue one ev record */
+		ev = ctx->ev_done_list[ctx->ev_done_h];
+		ctx->ev_done_h++;
+		if (ctx->ev_done_h >= ctx->depth)
+			ctx->ev_done_h = 0;
+		ctx->ev_queue_len--;
+
+		events[*num_ev].obj = ev->piocb;
+		events[*num_ev].res = ev->piocb->u.c.nbytes;
+		daos_event_fini(&ev->ev);
```
mchaarawi, line 6701: rc check

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6355,6 +6408,416 @@ register_handler(int sig, struct sigaction *old_handler)
...
+			dt.tv_sec  = times_1.tv_sec - times_0.tv_sec;
+			dt.tv_nsec = times_1.tv_nsec - times_0.tv_nsec;
+			if (dt.tv_nsec < 0) {
+				dt.tv_sec--;
+				dt.tv_nsec += 1000000000L;
+			}
+			if ((dt.tv_sec > timeout->tv_sec) ||
+			    ((dt.tv_sec == timeout->tv_sec) && ((dt.tv_nsec > timeout->tv_nsec)))) {
+				return op_done;
+			}
```
mchaarawi, line 6815: why implement your own timeout mechanism if daos_eq_poll supports it already?

mchaarawi (changes requested):

src/client/dfuse/il/int_posix.c
wiliamhuang: Fixed as suggested. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Now they are replaced by atomic_init(). Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: This code is inside io_setup(). If user's application does not use libaio, this function should not be triggered.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! I changed it to "D_ALLOC_PTR(aio_ctx_obj);" for better readability although "D_ALLOC(aio_ctx_obj, sizeof(aio_ctx_t));" is correct. aio_ctx_t is struct not struct pointer.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: @ashleypittman I just added "daos_event_fini(&ctx_ev->ev);" at here to destroy the event. Is this you were suggesting? Thank you!

src/tests/ftest/dfuse/bash.py
wiliamhuang: Thank you! We could add a dedicated test for fio using libaio either inside PR https://github.com/daos-stack/daos/pull/13797 or a new PR based on this one.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. I am working on putting aio related stuff into separate files. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! "lock_aio" now is renamed as "lock_aio_eq_g".

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: depth_sum_accu is increased by iodepth when a new aio context is created. Number of "iodepth" event queues (EQs) are created if the total number of EQs is not beyond the set limit.
It is used to determine "first_eq".
aio_ctx->first_eq = depth_sum_accu % aio_eq_count_g;
This is introduced to distribute IO requests from aio context evenly over available EQs and minimize the chances EQs are used by multiple aio contexts.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: I just realized that "aio_eq_count" is renamed as "aio_eq_count_g" now. Do I need to change them to "aio_eq**s**_count_g"? Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Added the check of returned value for daos_eq_destroy(). Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: The first two loops are fused into one loop now. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: You are right. Now they are handled in the same way. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Added the check of returned value for daos_event_fini(). Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Because we need to poll multiple event queues and return once the number of finished requests reaches desired threshold.

wiliamhuang:
All tests in CI finished without issue.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. I will update in next commit. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: OK i had to read the code again to understand what you mean.
if this is the running total of io depths from all io contexts, then please use a clearer name:
io_depth_total_g ? or something to that end..

also update the comment please on why this is needed (choosing the first eq for every io context request submission)

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -263,7 +182,7 @@ static bool            segv_handler_inited;
 struct sigaction       old_segv;
 
 /* the flag to indicate whether initlization is finished or not */
-static bool            hook_enabled;
+bool                   hook_enabled;
```
ashleypittman, line 185: If this is now public then it (and any other exported symbols) needs a `d_` prefix or similar.

wiliamhuang: I will fix them in next commit. Thank you!

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,503 @@
...
+
+static int
+ev_enqueue(struct d_aio_ctx *ctx, struct d_aio_ev *ev)
+{
+	D_MUTEX_LOCK(&ctx->lock);
+	if (ctx->ev_queue_len >= ctx->depth) {
+		D_MUTEX_UNLOCK(&ctx->lock);
+		/* Unexpected here */
+		assert(0);
+	}
```
ashleypittman, line 352: You don't need to unlock on assert and this code as written will not log anything useful on failure, would this work better?
```suggestion
	D_ASSERT(ctx->depth < ctx->ev_queue_len);
```

wiliamhuang: You are right. I will fix it as suggested. Thank you!

wiliamhuang:
The failed test was reported in ticket, https://daosio.atlassian.net/browse/DAOS-15124

src/client/dfuse/pil4dfs/aio.c
wiliamhuang: Fixed as suggested. Thank you!

wiliamhuang: This is fixed as you suggested. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Variables are renamed as suggested. Thank you!

wiliamhuang:
@mchaarawi @ashleypittman Do you have more feedback? Thank you!

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,502 @@
...
+
+struct d_aio_eq {
+	daos_handle_t    eq;
+	pthread_mutex_t  lock;
+	_Atomic uint64_t n_op_queued;
+	_Atomic uint64_t n_op_done;
+};
+
+/* list of EQs dedicated for aio contexts. */
+static struct d_aio_eq aio_eq_list[MAX_EQ];
```
mchaarawi, line 40: I am concerned with the large numbers of EQs/cart network contexts that are just dedicated just for aio.
It would be better to limit those to 16.

wiliamhuang: ok. I will decrease the limit to 16 in next commit. Thank you!

johannlombardi (commented):
Mostly curious why we decided to use a fix number of eq vs one eq per aio context. Is that because we did not want to intercept the functions to create/destroy aio context?

wiliamhuang:
> Mostly curious why we decided to use a fix number of eq vs one eq per aio context. Is that because we did not want to intercept the functions to create/destroy aio context?

fio only uses one aio context. Normally one can get better io performance by increasing iodepth of aio context. If one aio context only uses one EQ, we can not provide better performance for iodepth larger than one unless using multiple EQ per aio context.

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,506 @@
...
+	pthread_mutex_t  lock;
+	_Atomic uint64_t n_op_queued;
+	_Atomic uint64_t n_op_done;
+};
+
+/* The max number of event queues dedicated to AIO */
+#define MAX_NUM_EQ_AIO  (16)
+/* list of EQs dedicated for aio contexts. */
+static struct d_aio_eq aio_eq_list[MAX_NUM_EQ_AIO];
+/* The accumulated sum of iodepth for all created aio contexts. It is introduced to distribute IO
```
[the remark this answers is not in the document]
wiliamhuang: fio only uses one aio context. Normally one can get better io performance by increasing iodepth of an aio context. If one aio context only uses one EQ, we can not provide better performance for iodepth larger than one unless using multiple EQ per aio context.

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,506 @@
...
+			DS_ERROR(EINVAL, "io_submit only supports PREAD and PWRITE for now");
+			return (-EINVAL);
+		}
+	}
+	if (n_op_dfs == 0)
+		return next_io_submit(ctx_real, nr, ios);
+
+	if (n_op_dfs != nr) {
+		DS_ERROR(EINVAL, "io_submit() does not support mixed non-dfs and dfs files yet");
+		return (-EINVAL);
```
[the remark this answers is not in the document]
wiliamhuang: You are right. I will add it if compatibility mode is enabled. Thank you!

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,506 @@
...
+	for (i = 0; i < nr; i++) {
+		if (ios[i]->aio_fildes >= FD_FILE_BASE)
+			n_op_dfs++;
+
+		op = ios[i]->aio_lio_opcode;
+		/* only support IO_CMD_PREAD and IO_CMD_PWRITE */
+		if (op != IO_CMD_PREAD && op != IO_CMD_PWRITE) {
+			DS_ERROR(EINVAL, "io_submit only supports PREAD and PWRITE for now");
+			return (-EINVAL);
+		}
```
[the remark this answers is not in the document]
wiliamhuang: Although "IO_CMD_NOOP" is defined in /usr/include/libaio.h, I got "Invalid argument" when trying "IO_CMD_NOOP". I am not sure kernel does not support "IO_CMD_NOOP" or I did not use it correctly. I could not find any examples how to use it.

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,506 @@
...
+	int               ev_queue_len;
+};
+
+typedef struct d_aio_ctx d_aio_ctx_t;
+
+static int (*next_io_setup)(int maxevents, io_context_t *ctxp);
+static int (*next_io_destroy)(io_context_t ctx);
+static int (*next_io_submit)(io_context_t ctx, long nr, struct iocb *ios[]);
+static int (*next_io_cancel)(io_context_t ctx, struct iocb *iocb, struct io_event *evt);
+static int (*next_io_getevents)(io_context_t ctx, long min_nr, long nr, struct io_event *events,
```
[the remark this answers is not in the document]
wiliamhuang: Both io_queue_run() and io_queue_wait() call io_getevents().
io_queue_release() simply calls io_destroy().
int io_queue_release(io_context_t ctx)
{
	return io_destroy(ctx);
}

io_set_callback() and io_fsync() are defined as inline functions inside /usr/include/libaio.h. io_fsync() is based on io_submit().

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -77,6 +78,7 @@
 #define FAKE_ST_INO(path)   (d_hash_string_u32(path, strnlen(path, DFS_MAX_PATH)))
 
 #define MAX_EQ              64
+#define MAX_AIO_CTX         8
```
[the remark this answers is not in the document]
wiliamhuang: Now we dynamically allocate aio context instead of statically allocated fixed number of aio contexts.

src/client/dfuse/pil4dfs/aio.c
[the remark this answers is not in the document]
johannlombardi: Fair

wiliamhuang:
PR was updated to assign only one event queue for each aio context as Johann suggested. The performance using a single EQ and multiple EQs are same in my benchmark on wolf cluster A.

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,506 @@
...
+	ev_dequeue_batch(ctx, min_nr, nr, events, num_ev);
+	if (*num_ev >= min_nr)
+		return;
+
+	/* loop over all EQs */
+	for (i = 0; i < d_aio_eq_count_g; i++) {
+		if (atomic_load_relaxed(&aio_eq_list[i].n_op_queued) == 0)
+			continue;
+
+		rc = daos_eq_poll(aio_eq_list[i].eq, 0, DAOS_EQ_NOWAIT, AIO_EQ_DEPTH, eps);
```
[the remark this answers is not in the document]
wiliamhuang: EQ could be shared by multiple aio contexts, so we probably still need to use "DAOS_EQ_NOWAIT" when calling daos_eq_poll().

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,506 @@
...
+		return (-EINVAL);
+	}
+	if (timeout)
+		clock_gettime(CLOCK_REALTIME, &times_0);
+
+	if (min_nr > AIO_EQ_DEPTH)
+		min_nr = AIO_EQ_DEPTH;
+
+	while (1) {
+		aio_poll_eqs(aio_ctx_obj, min_nr, nr, events, &op_done);
```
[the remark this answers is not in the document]
wiliamhuang: Now each aio context uses only one EQ. In case of the number of aio contexts is larger than the number of EQs for aio (limit is 16 currently), multiple aio contexts might share EQ.

wiliamhuang:
Just merged with master to fix conflicts in int_dfs.c due to renaming variables in the PR. tests in CI passed previously, so tests are skipped in the rebuild.

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -304,15 +315,20 @@ io_submit(io_context_t ctx, long nr, struct iocb *ios[])
...
+		if (req_args == NULL)
+			D_GOTO(err_loop, rc = ENOMEM);
+		d_iov_set(&req_args->iov, (void *)ios[i]->u.c.buf, ios[i]->u.c.nbytes);
+		req_args->sgl.sg_nr     = 1;
+		req_args->sgl.sg_iovs   = &req_args->iov;
+
 		if (op == IO_CMD_PREAD) {
-			rc = dfs_read(d_file_list[fd]->dfs_mt->dfs, d_file_list[fd]->file, &sgl,
-				      ios[i]->u.c.offset, &read_size, &ctx_ev->ev);
+			daos_event_register_comp_cb(&ctx_ev->ev, aio_req_cb, req_args);
```
mchaarawi, line 329: this needs rc check

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,552 @@
...
+				      &req_args->sgl, ios[i]->u.c.offset, &read_size, &ctx_ev->ev);
+			if (rc) {
+				rc2 = daos_event_fini(&ctx_ev->ev);
+				if (rc2)
+					DL_ERROR(rc2, "daos_event_fini() failed");
+				D_GOTO(err_loop, rc);
+			}
+		}
+		if (op == IO_CMD_PWRITE) {
+			daos_event_register_comp_cb(&ctx_ev->ev, aio_req_cb, req_args);
```
mchaarawi, line 340: same

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,552 @@
...
+	if (!aio_ctx_obj->inited) {
+		rc = create_ev_eq_for_aio(aio_ctx_obj);
+		if (rc)
+			D_GOTO(err, rc);
+	}
+
+	for (i = 0; i < nr; i++) {
+		op         = ios[i]->aio_lio_opcode;
+		fd         = fd_directed[i] - FD_FILE_BASE;
+		idx_aio_eq = aio_ctx_obj->idx_eq;
```
mchaarawi, line 307: is this now the same for every io in this loop?
should it be moved out of the loop?

src/client/dfuse/pil4dfs/aio.c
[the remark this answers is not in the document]
mchaarawi: I think it would be better to just keep it 1 separate EQ for each context and not put a limitation on the number of EQs in pil4dfs for aio. Internally we already have a cap on the number of network contexts that are shared at some point between EQs, so this won't cause a resource limit.
let's address though in a follow on PR.

mchaarawi (approved):

src/client/dfuse/pil4dfs/aio.c
wiliamhuang: ok. I will add rc check later. Thank you!

src/client/dfuse/pil4dfs/aio.c
wiliamhuang: Yes. "idx_aio_eq = aio_ctx_obj->idx_eq;" is same for all requests in the same context. I will move this line out of the loop. Thank you!

src/client/dfuse/pil4dfs/aio.c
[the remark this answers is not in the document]
wiliamhuang: Good to know. I will make such change in a follow up PR. Thank you!

src/client/dfuse/pil4dfs/aio.c
```diff
@@ -0,0 +1,552 @@
...
+	}
+	if (!d_hook_enabled)
+		return next_io_cancel(ctx_real, iocb, evt);
+
+	if (iocb->aio_fildes < FD_FILE_BASE)
+		return next_io_cancel(ctx_real, iocb, evt);
+
+	/* daos_event_abort() may be used to implement this feature later. */
+	DS_ERROR(ENOSYS, "io_cancel() for DFS is not implemented");
+	return (-ENOSYS);
```
johannlombardi, line 390: From the man page:

RETURN VALUES
       0 is returned on success, otherwise returns errno.

This should be return ENOSYS, no?

johannlombardi (approved):
Several cleanups still required. But looks much better to me. Thanks for doing those changes.

src/client/dfuse/pil4dfs/aio.c
wiliamhuang: Please see the example code in kernel for io_submit(). Negative error is used.
https://github.com/torvalds/linux/blob/master/fs/aio.c#L2096

johannlombardi: For io_submit, yes ... but the man page says to use positive error for io_cancel. anyways ....

wiliamhuang: https://github.com/torvalds/linux/blob/master/fs/aio.c#L2193
io_cancel also uses negative errno in kernel.

wiliamhuang:
@mchaarawi I probably need to update the time in debian/changelog and daos.spec? They are out of date now.

wiliamhuang:
@mchaarawi @johannlombardi Just updated the PR to update the time in debian/changelog and daos.spec. Also added rc check for daos_event_register_comp_cb().
Could you please approve? Thank you!

src/client/dfuse/pil4dfs/aio.c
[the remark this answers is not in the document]
wiliamhuang: @mchaarawi Ticket was created for fixing this issue. https://daosio.atlassian.net/browse/DAOS-15803

mchaarawi (approved):

wiliamhuang:
@johannlombardi Could you please review? Only refreshed the PR to update the time in debian/changelog and daos.spec. Also added rc check for daos_event_register_comp_cb().

PR #13353, opened 2023-11-17
DAOS-14625 cart: Print daos opcodes in rpc macros

frostedcmos:
Instead of previously printing (DAOS) for an opcode, it will now print full daos opcode during an error print.

Required-githooks: true

### Gatekeeper:


frostedcmos:
Example of a timeout print now:
`rpc  ERR  src/cart/crt_context.c:1149 crt_context_timeout_check(0x55740f8ccc10) [opc=0x3080000 (DAOS_OPC_SET_PARAMS) rpcid=0x47a6b72900002329 rank:tag=0:0] ctx_id 0, (status: 0x38) timed out (60 seconds), target (0:0)
`

src/pool/srv_pool.c
```diff
@@ -2741,7 +2741,7 @@ ds_pool_create_handler(crt_rpc_t *rpc)
 
 	rc = pool_prop_default_copy(prop_dup, prop);
 	if (rc) {
-		DL_ERROR(rc, "daos_prop_default_copy failed.\n");
+		DL_ERROR(rc, "daos_prop_default_copy failed.");
```
ashleypittman, line 2744: FYI - this is fixed in https://github.com/daos-stack/daos/pull/13351 but the two changes will conflict.

src/cart/crt_rpc.c
```diff
@@ -226,10 +231,16 @@ char
 		CRT_ST_RPCS_LIST
 		CRT_CTL_RPCS_LIST
 	}
-	return "DAOS";
+
+	switch (opc >> CRT_PROTO_BASEOPC_SHIFT) {
```
ashleypittman, line 235: Do you only want to come into this code for some values of the upper bits (basically the protocol id?)

ashleypittman (commented):
Clever!

src/cart/crt_rpc.c
frostedcmos: Hmm, not sure i understand.
Upper 8 bits are for the opcode, while bottom 16 are or the protocol, so we are extracting base opcode here and comparing it to base daos opcodes.

262 #define CRT_PROTO_BASEOPC_MASK  (0xFF000000UL)
263 #define CRT_PROTO_BASEOPC_SHIFT 24
264 #define CRT_PROTO_VER_MASK      (0x00FF0000UL)
265 #define CRT_PROTO_COUNT_MASK    (0x0000FFFFUL)

liw (commented):
Sorry, @frostedcmos, but I'm afraid these are not RPC operation codes; they are used to index `dc_funcs`. For instance, a user may create a pool query task using `DAOS_OPC_POOL_QUERY`, which is 10, find `dc_pool_query` in `dc_funcs[10]`, and create an RPC with operation code `POOL_QUERY`, which is 3.

jgmoore-or (dismissed):

frostedcmos:
> Sorry, @frostedcmos, but I'm afraid these are not RPC operation codes; they are used to index `dc_funcs`. For instance, a user may create a pool query task using `DAOS_OPC_POOL_QUERY`, which is 10, find `dc_pool_query` in `dc_funcs[10]`, and create an RPC with operation code `POOL_QUERY`, which is 3.

hmm i see, let me look further into this

frostedcmos:
> > Sorry, @frostedcmos, but I'm afraid these are not RPC operation codes; they are used to index `dc_funcs`. For instance, a user may create a pool query task using `DAOS_OPC_POOL_QUERY`, which is 10, find `dc_pool_query` in `dc_funcs[10]`, and create an RPC with operation code `POOL_QUERY`, which is 3.
> 
> hmm i see, let me look further into this

Thanks for this mention, I've revised the version now. Not every module is included, as some would need bigger header file changes.

src/cart/crt_internal.h
```diff
@@ -27,28 +27,37 @@
...
-			(rpc)->crp_req_hdr.cch_rpcid,			\
-			(rpc)->crp_pub.cr_ep.ep_rank,			\
-			(rpc)->crp_pub.cr_ep.ep_tag,			\
-			## __VA_ARGS__);				\
+#define RPC_TRACE(mask, rpc, fmt, ...)						\
+	do {									\
+		char *_module;							\
+		char *_opc;							\
+										\
+		crt_opc_decode((rpc)->crp_pub.cr_opc, &_module, &_opc);		\
```
ashleypittman, line 35: This code will be called a lot and crt_opc_decode is potentially expensive, you should use `D_LOG_ENABLED` to check if the logging will be used before calling this function.

src/include/daos/rpc.h
```diff
@@ -40,23 +40,42 @@
...
-	DAOS_MAX_MODULE		= 64  /** Size of uint64_t see dmg profile */
+#define X(a,b) a = b,
+	DAOS_MODULE_LIST
+#undef X
+
 };
 
+static inline char* daos_opc_to_module_str(uint32_t daos_opc)
+{
+#define X(a,b) case a: return #a;
```
ashleypittman, line 69: You can use variadic macros here, they won't help make this PR smaller but it's better practice and means any future changes to the definitions will not require updates to existing users.
```suggestion
#define X(a, ...) case a: return #a;
```

I'm only commenting here but this applies across this PR obviously.

ashleypittman (commented):
What's do you want the eventual output to look like here?  You've got `(%s:%s)` but this disregards the protocol version which will become more important over time.  `%s[%d].%s` might be preferable.

I wonder if there's a way of just computing these strings once rather than every time, perhaps by extending the proto register code to take a string or working these out at startup and saving them in the protocol datastructures.

src/cart/SConscript
```diff
@@ -64,6 +64,8 @@ def scons():
     env.AppendUnique(LIBPATH=[Dir('.')])
 
     env.require('mercury', headers_only=True)
+    env.require('isal', headers_only=True)
+    env.require('argobots', headers_only=True)
```
jolivier23, line 68: why is this needed?   I recently pushed a PR to make sure argobots is only needed for client build

jolivier23 (changes requested):

src/cart/SConscript
frostedcmos: one of these headers was pulling dependencies -- i ll need to go back check which one was it
#include "../mgmt/rpc.h"
#include "../pool/rpc.h"
#include "../container/rpc.h"
#include "../pipeline/pipeline_rpc.h"
#include "../object/obj_rpc.h"

frostedcmos:
> What's do you want the eventual output to look like here? You've got `(%s:%s)` but this disregards the protocol version which will become more important over time. `%s[%d].%s` might be preferable.
> 
> I wonder if there's a way of just computing these strings once rather than every time, perhaps by extending the proto register code to take a string or working these out at startup and saving them in the protocol datastructures.

Yeah I went with versions first locally, but that did not seem that useful, especially given that full opcode itself is printed too. Currently it prints as you said:

` wolf-22 DAOS[4062176/0/153] rpc  DBUG src/cart/crt_rpc.c:1561 crt_reply_send(0x7f26d60b4170) [opc=0x3080003 (DAOS_CONT_MODULE:CONT_TGT_SNAPSHOT_NOTIFY) rpcid=0x4ec12ec60000000a rank:tag=0:0]
`
As for precomputing ahead of time, I am not sure if its worth it. We dont do any string manipulations here at runtime, rather few bit shifts to break uint32 into module/version/opc followed by static jump tables to return corresponding strings

src/cart/crt_internal.h
frostedcmos: done

src/cart/SConscript
frostedcmos: argobots was being pulled by rdb header file, however it starts wanting to pull in flake and other things too, and for that reason for now the rdb module is not included in rpc prints. 
isal was required by object/obj_rpc.h which then includes object/obj_ec.h 

i've removed argobots for now, but have to keep isal for object module

jgmoore-or (approved):

src/cart/crt_rpc.c
```diff
@@ -9,8 +9,15 @@
...
-
+#include <daos_task.h>
 #include "crt_internal.h"
+#include <daos/rpc.h>
+
+#include "../mgmt/rpc.h"
+#include "../pool/rpc.h"
+#include "../container/rpc.h"
+#include "../pipeline/pipeline_rpc.h"
+#include "../object/obj_rpc.h"
```
soumagne, line 20: those includes are not particularly clean but I understand why you did that...

src/include/daos/rpc.h
```diff
@@ -40,23 +40,46 @@
...
-	DAOS_RDB_MODULE		= 7, /** rdb */
-	DAOS_RDBT_MODULE	= 8, /** rdb test */
-	DAOS_SEC_MODULE		= 9, /** security framework */
-	DAOS_DTX_MODULE		= 10, /** DTX */
-	DAOS_PIPELINE_MODULE	= 11,
-	DAOS_NR_MODULE		= 12, /** number of defined modules */
-	DAOS_MAX_MODULE		= 64  /** Size of uint64_t see dmg profile */
+#define X(a, b) a = b,
+	DAOS_MODULE_LIST
+#undef X
```
soumagne, line 62: why do we need to add explicit values for enums ? all enums start from 0

soumagne (approved):
just a few comments but looks good to me

src/include/daos/rpc.h
frostedcmos: > why do we need to add explicit values for enums ? all enums start from 0

in this case i am guessing cause some values are disjoint, like MAX being 64. Also with macros now, you need to either have all values being explicit or none in a macro.

You would need something like this, which I don't think is possible with conventional macros:
#define X(a, b) a = b (if b is set, else) a

src/cart/crt_rpc.c
frostedcmos: Yeah, not a big fan of it myself. 
Eventually it would be good to reorganize some of these headers too, separating rpc-specific stuff from everything else and bringing all rpc definitions to the same form. Some modules today provide opc-to-str functions, some dont, some have rpc list, others have client and server lists and so on. Some headers like rdb's rpc.h couldnt be included at all as they start pulling too much extra stuff

src/include/daos/rpc.h
soumagne: Sure I would certainly not try to complicate the macro. But in this case I honestly think that's a mistake and you should not bother having it in the macro either, DAOS_MAX_MODULE should not be 64, it should be the last enum value + 1. 
 
> some values

There is no other disjoint value :)

src/cart/crt_rpc.c
soumagne: if you can I think it would be good to add a comment about it, as a TODO item + explanation of limitations (in case someone else wants to pick that up).

src/include/daos/rpc.h
ashleypittman: If the last one is special then yes, there are better ways to handle it.

frostedcmos: As discussed in the last cart meeting, ideally this list should also include cart (as we occupy the same module space), and in carts case we reserve 0xff and 0xfe making the list disjoint. I want to unify those lists in the subsequent patch, which also aims at removing headers and storing resolved rpc strings instead in the opc (earlier suggestion by @ashleypittman and @liw )

src/cart/crt_rpc.c
frostedcmos: As discussed in cart meeting these includes will be gone hopefully in the subsequent patch, while we already have 2 TODOs in the code itself indicating modules that need to be added.

PR #13368, opened 2023-11-21
DAOS-14561 vos: Adjust GC metrics

mjmac:
* Switch gc gauges to counters for monitoring
    as rates instead of absolute values
  * Add aggregation failure counter

Also adds a small change to the Go builder for
handling versions with a -suffix.

Required-githooks: true

Change-Id: I0af682bff3dcfe28a78c69d9ed1746afffe9eafa

### DIFF
--- a/src/vos/vos_common.c
+++ b/src/vos/vos_common.c
@@ -759,6 +759,12 @@ vos_metrics_alloc(const char *path, int tgt_id)
 	if (rc)
 		D_WARN("Failed to create 'merged_size' telemetry : "DF_RC"\n", DP_RC(rc));
 
+	/* VOS aggregation failed */
+	rc = d_tm_add_metric(&vam->vam_fail_count, D_TM_COUNTER, "aggregation failures", NULL,
+			     "%s/%s/fail_count/tgt_%u", path, VOS_AGG_DIR, tgt_id);
+	if (rc)
+		D_WARN("Failed to create 'fail_count' telemetry : "DF_RC"\n", DP_RC(rc));
+
 	/* Metrics related to VOS checkpointing */
 	vos_chkpt_metrics_init(&vp_metrics->vp_chkpt_metrics, path, tgt_id);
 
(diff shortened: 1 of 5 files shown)


src/vos/vos_common.c
```diff
@@ -759,6 +759,12 @@ vos_metrics_alloc(const char *path, int tgt_id)
 	if (rc)
 		D_WARN("Failed to create 'merged_size' telemetry : "DF_RC"\n", DP_RC(rc));
 
+	/* VOS aggregation failed */
+	rc = d_tm_add_metric(&vam->vam_fail_count, D_TM_COUNTER, "aggregation failures", NULL,
+			     "%s/%s/fail_count/tgt_%u", path, VOS_AGG_DIR, tgt_id);
+	if (rc)
+		D_WARN("Failed to create 'fail_count' telemetry : "DF_RC"\n", DP_RC(rc));
```
ashleypittman, line 766: FYI - We've updated best-practice here to use the following syntax for improved clarity.   ```suggestion
		DL_WARN(rc, "Failed to create 'fail_count' telemetry");
```

mjmac: Will there be a mass-update of all of the `D_WARN` lines?

ashleypittman: There could be in future but the current focus is in changing developer practices.

mjmac:
I am not sure why clang-format is complaining about the format of the new struct member. It uses tabs, same as the other lines. The suggested change seems to want spaces, which would imply that the existing lines are wrong.

ashleypittman:
> I am not sure why clang-format is complaining about the format of the new struct member. It uses tabs, same as the other lines. The suggested change seems to want spaces, which would imply that the existing lines are wrong.

They may well be, lots of our c code only loosely follows the standard. To get it to pass the way would be to reformat the whole struct but that obviously makes the whole patch bigger so for one entry probably isn't worth it.

jolivier23 (approved):

NiuYawei (approved):

PR #13372, opened 2023-11-22
DAOS-14261 engine: Add dss_chore for I/O forwarding

liw:
As requested by the Jira ticket, add a new I/O forwarding mechanism,
dss_chore, to avoid creating a ULT for every forwarding task.

  - Forwarding of object I/O and DTX RPCs is converted to chores.

  - Cancelation is not implemented, because the I/O forwarding tasks
    themselves do not support cancelation yet.

  - In certain engine configurations, some xstreams do not need to
    initialize dx_chore_queue. This is left to future work.

### Gatekeeper:


src/engine/srv.c
```diff
@@ -518,6 +518,15 @@ dss_srv_handler(void *arg)
 		}
 	}
 
+	/* Some xstreams do not need this, actually... */
```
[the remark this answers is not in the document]
liw: Originally intended for future work... Done.

src/engine/ult.c
```diff
@@ -610,3 +610,227 @@ dss_main_exec(void (*func)(void *), void *arg)
...
+
+		d_list_for_each_entry_safe(chore, chore_tmp, &list, cho_link) {
+			bool is_reentrance = (chore->cho_status == DSS_CHORE_READY);
+
+			D_DEBUG(DB_TRACE, "%p: status=%d\n", chore, chore->cho_status);
+			chore->cho_status = chore->cho_func(chore, is_reentrance);
+			D_ASSERT(chore->cho_status != DSS_CHORE_NEW);
+			if (chore->cho_status == DSS_CHORE_DONE)
+				d_list_del_init(&chore->cho_link);
+			ABT_thread_yield();
```
Nasf-Fan, line 760: If the cho_func() return DSS_CHORE_READY, then the chore is still in the temporary list, who and when will insert it back to the queue->chq_list?

src/engine/ult.c
```diff
@@ -610,3 +610,227 @@ dss_main_exec(void (*func)(void *), void *arg)
...
+dss_chore_delegate(struct dss_chore *chore, dss_chore_func_t func)
+{
+	struct dss_module_info *info = dss_get_module_info();
+	int                     xs_id;
+	struct dss_xstream     *dx;
+	struct dss_chore_queue *queue;
+
+	chore->cho_status = DSS_CHORE_NEW;
+	chore->cho_func   = func;
+
```
Nasf-Fan, line 661: So it is the caller's duty to guarantee that it will call dss_chore_delegate() again inside func(), right? Do we have some mechanism to avoid that?

src/engine/ult.c
```diff
@@ -610,3 +610,227 @@ dss_main_exec(void (*func)(void *), void *arg)
...
+	/* Find the chore queue. */
+	xs_id = sched_ult2xs(DSS_XS_IOFW, info->dmi_tgt_id);
+	D_ASSERT(xs_id != -DER_INVAL);
+	dx = dss_get_xstream(xs_id);
+	D_ASSERT(dx != NULL);
+	queue = dx->dx_chore_queue;
+	D_ASSERT(queue != NULL);
+
+	ABT_mutex_lock(queue->chq_mutex);
+	if (queue->chq_stop) {
```
Nasf-Fan, line 682: In theory, if "queue->chq_stop" is set, then means such XS is to exit. Under such case, the "dx->dx_chore_queue" may has been already been destroyed. So here, we may access invalid DRAM. It seems some trouble although I do not have better idea now.

src/dtx/dtx_rpc.c
```diff
@@ -705,11 +735,18 @@ dtx_rpc_prep(struct ds_cont_child *cont,d_list_t *dti_list,  struct dtx_entry **
...
+	if (dss_has_enough_helper()) {
+		rc = ABT_eventual_create(0, &dca->dca_chore_eventual);
+		if (rc != ABT_SUCCESS) {
+			D_ERROR("failed to create eventual: %d\n", rc);
+			rc = dss_abterr2der(rc);
+			goto out;
+		}
+		rc = dss_chore_delegate(&dca->dca_chore, dtx_rpc_helper);
+	} else {
+		dss_chore_diy(&dca->dca_chore, dtx_rpc_helper);
```
Nasf-Fan, line 747: It may be better to hide dss_chore_diy(), instead, just call dss_chore_delegate() and check whether there is enough helper XS inside dss_chore_delegate (your patch has already does similar check, may need enhancement), then determine:
1. Add chore to the queue if possible. Or
2. Start ULT on neighbor to help me. Or
3. Directly do it by myself.
With such approach, then DTX logic needs not to understand helper XS knowledge. And also save one chore API.

Nasf-Fan (changes requested):

src/dtx/dtx_rpc.c
liw: The problem is that in case 1 and 2, the API call returns asynchronously (i.e., the chore may or may not have started and terminated), whereas in case 3, the API call returns only if the chore has started and terminated. The user code needs to differentiate between these two scenarios. Therefore I feel two API functions for these two scenarios are clearer and thus simpler.

src/dtx/dtx_rpc.c
```diff
@@ -721,8 +758,12 @@ dtx_rpc_post(struct dtx_common_args *dca, int ret, bool keep_head)
 	struct dtx_req_rec	*drr;
 	int			 rc;
 
-	if (dca->dca_helper != ABT_THREAD_NULL)
-		ABT_thread_free(&dca->dca_helper);
+	if (dca->dca_chore_eventual != ABT_EVENTUAL_NULL) {
+		rc = ABT_eventual_wait(dca->dca_chore_eventual, NULL);
```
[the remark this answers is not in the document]
liw: Could you let me know through what code path `dss_chore_queue_ult` can reach here? Or, have I misunderstood the question?

src/engine/ult.c
liw: Do you mean "it will NOT call"? If so, no, we don't check for that. If such coding bugs have occurred, we can see if it's possible to catch those bugs; for now I don't see a great need...

src/engine/ult.c
liw: This is a good point... I'll investigate to see how to address it.

src/engine/ult.c
```diff
@@ -610,3 +610,227 @@ dss_main_exec(void (*func)(void *), void *arg)
...
+		ABT_mutex_unlock(queue->chq_mutex);
+
+		if (stop)
+			break;
+
+		d_list_for_each_entry_safe(chore, chore_tmp, &list, cho_link) {
+			bool is_reentrance = (chore->cho_status == DSS_CHORE_READY);
+
+			D_DEBUG(DB_TRACE, "%p: status=%d\n", chore, chore->cho_status);
+			chore->cho_status = chore->cho_func(chore, is_reentrance);
```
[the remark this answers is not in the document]
liw: Yes, so do not do that. :) I think this is similar to: "If a ULT pthread-blocks for some reason, then the xstream cannot process its other ULTs." (I did, however, propose to Niu in the past some potential idea for detecting such bugs at run time. Anyway, that's definitely out of the scope of this ticket.)

src/engine/ult.c
liw: The `splice` on line 736. Or, have I misunderstood the question?

src/include/daos_srv/daos_engine.h
```diff
@@ -821,4 +821,40 @@ enum dss_drpc_call_flag {
 int dss_drpc_call(int32_t module, int32_t method, void *req, size_t req_size,
 		  unsigned int flags, Drpc__Response **resp);
 
+/** Status of a chore */
+enum dss_chore_status {
+	DSS_CHORE_NEW,		/**< ready to be scheduled for the first time (private) */
+	DSS_CHORE_READY,	/**< ready to be scheduled */
```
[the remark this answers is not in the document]
liw: I'm fine to rename this to `YIELD`.

liw (commented):
@Nasf-Fan, thank you. Please see my replies.

src/engine/ult.c
liw: Moved `dss_chore_queue_fini` to the point after all `dx` streams have completed.

liw:
Looks like CI isn't going to recover any time soon; I'll request Fan Yong to review again before getting the CI result then.

src/engine/srv.c
```diff
@@ -773,6 +787,8 @@ dss_start_one_xstream(hwloc_cpuset_t cpus, int tag, int xs_id)
 	} else {
 		dx->dx_main_xs	= (xs_id >= dss_sys_xs_nr) && (xs_offset == 0);
 	}
+	/* See the DSS_XS_IOFW case in sched_ult2xs. */
+	dx->dx_iofw = xs_id >= dss_sys_xs_nr && (!dx->dx_main_xs || dss_tgt_offload_xs_nr == 0);
```
NiuYawei, line 791: minor: Seems you don't regard vos xstream as iofw xstream, so it could be simplified as "dx->dx_iofw = xs_id >= dss_sys_xs_nr && !dx->dx_main_xs".

NiuYawei (dismissed):

src/engine/srv.c
liw: @NiuYawei, I thought about the same when updating the patch last time. Ended up keeping it this way because if we make such a change to `dx_iofw`, then a `dx_iofw` xstream != a DSS_XS_IOFW xstream, potentially making our terminology even more complex (i.e., we currently talk about "I/O forward xstream", "offload xstream", "helper xstream", "DSS_XS_OFFLOAD xstream", etc.---already too confusing).

src/dtx/dtx_rpc.c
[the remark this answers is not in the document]
Nasf-Fan: Originally, I thought the chore_ult may call dtx_rpc_post(), but current logic seems not, so it's OK.

src/engine/ult.c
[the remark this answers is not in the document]
Nasf-Fan: It may be helpful if we have the ability to check whether the target XS is blocked or not before generating new chore request on it or on some other neighbor. But it seems out of the scope.

src/engine/ult.c
Nasf-Fan: The d_list_splice_init() is for getting chores from queue->chq_list to the temporary list. That means once some chore is moved from queue->chq_list to the temporary list, then it will never be added back to queue->chq_list, right?

src/dtx/dtx_rpc.c
Nasf-Fan: The caller still calls dtx_rpc_post() without any changes to distinguish the cases.

src/engine/ult.c
liw: No, `list` is not cleared before the `splice`. Have I overlooked anything?

src/dtx/dtx_rpc.c
liw: I prefer not to merge `diy` with `delegate`, if that's acceptable to you.

liw (commented):
@Nasf-Fan, thanks for the replies.

src/engine/ult.c
Nasf-Fan: OK, that matches my understand. Means that the new added chores will always have higher priority to be scheduled. I am not sure whether it is reasonable or not, but large task (with multiple steps) will be delayed more under such schedule policy. It is suggested to verify the behavior under heavy loaded system on large scaled cluster.

src/dtx/dtx_rpc.c
Nasf-Fan: Current API can work. Mine is just for suggestion.

Nasf-Fan:
dtx_common.c and dtx_rpc.c have been much changed recently, need some merge work for this patch.

src/dtx/dtx_rpc.c
```diff
@@ -670,11 +700,18 @@ dtx_rpc_prep(struct ds_cont_child *cont,d_list_t *dti_list,  struct dtx_entry **
 	}
 
 	/* Use helper ULT to handle DTX RPC if there are enough helper XS. */
-	if (dss_has_enough_helper())
-		rc = dss_ult_create(dtx_rpc_helper, dca, DSS_XS_IOFW, dca->dca_tgtid, 0,
-				    &dca->dca_helper);
-	else
-		rc = dtx_rpc_internal(dca);
+	if (dss_has_enough_helper()) {
+		rc = ABT_eventual_create(0, &dca->dca_chore_eventual);
```
Nasf-Fan, line 704: Why non-coll case need the eventual, but coll case does not?

Nasf-Fan: Is it possible to not create the dca_chore_eventual, instead, depends on the dra->dra_future to check whether related operation has been done or not, then non-coll case and coll case can share the same mechanism or logic.

liw: > Why non-coll case need the eventual, but coll case does not?

@Nasf-Fan, the `dca_chore_eventual` is for `dtx_rpc_post` to know when it's safe to access `dra_future`.

> Is it possible to not create the dca_chore_eventual, instead, depends on the dra->dra_future to check whether related operation has been done or not, then non-coll case and coll case can share the same mechanism or logic.

AFAICS, it's possible, though not easy. The main trouble is that the number of slots in the future is computed _by_ the chore, so we can't create the future until the chore has run to the point where that number is computed. Also, even if we really want to make such a change in this PR, it will take more changes to share the mechanism with `dtx_coll_rpc_helper`, since the two code paths use different data structures.

So, what do you think?

Nasf-Fan: > > Why non-coll case need the eventual, but coll case does not?
> 
> @Nasf-Fan, the `dca_chore_eventual` is for `dtx_rpc_post` to know when it's safe to access `dra_future`.
> 
> > Is it possible to not create the dca_chore_eventual, instead, depends on the dra->dra_future to check whether related operation has been done or not, then non-coll case and coll case can share the same mechanism or logic.
> 
> AFAICS, it's possible, though not easy. The main trouble is that the number of slots in the future is computed _by_ the chore, so we can't create the future until the chore has run to the point where that number is computed. Also, even if we really want to make such a change in this PR, it will take more changes to share the mechanism with `dtx_coll_rpc_helper`, since the two code paths use different data structures.
> 
> So, what do you think?

Then ignore my former comment and go ahead.

Nasf-Fan (approved):

NiuYawei (approved):

PR #13387, opened 2023-11-27
DAOS-14624 gurt: Fix d_rank_list_del and friends (#13317)

ashleypittman:
The function

  int d_rank_list_del(d_rank_list_t *rank_list, d_rank_t rank)

calls

  d_rank_list_t *d_rank_list_realloc(d_rank_list_t *ptr, uint32_t size)

which may free (or reallocate) the object ptr points and return NULL (or
the address of a new object). Due to d_rank_list_del’s signature, this
can’t be conveyed to the caller, who may continue using the freed
object. A similar pitfall exists in d_rank_list_append.

This patch renames d_rank_list_realloc to d_rank_list_resize and changes
its behavior so that it never frees the d_rank_list_t object.


liw (approved):

ashleypittman:
This is causing a daily test failure and preventing proper testing of https://github.com/daos-stack/daos/pull/13437

PR #13401, opened 2023-11-28
DAOS-14690 control: Fix TestSecurity_DomainInfo_String

mjmac:
The test assumed pid 1 will always be systemd, but on some
distributions it's still init. Just get the actual pid 1
process name to validate the stringer function. Simpler
solution than the previous fix that used hard-coded
process names.

Also adds a t.Helper() call in the dmg tests to better
identify the source of a failed command test.

Change-Id: If18680d7e2d9c08e80df3dfb128d29198e3d6acb
Required-githooks: true

Change-Id: Ia251f9e1621bb7a5a347ee59eb49788fa9d5d566


kjacque:
LGTM, but I'll wait for this to go through the backport/approval process to approve.

kjacque (approved):

tanabarr (approved):

PR #13409, opened 2023-11-29
DAOS-14686 object: revert PR-12217 and add tests to verify

wangdi1:
Revert PR-12217 to use all of target for GX object to avoid imbalance of the target space usages.

Then we might put multiple shards in the same target during rebuild, though the current DAOS should support it.

Add tests to verify the placement algorithm will not put multiple shards from the same group in the same domain.

Required-githooks: true

### Gatekeeper:


liuxuezhao:
the original patch was to handle the RF concern right? (if you put multiple shards in the same target during rebuild, after rebuild lost the tgt it will break RF assumption), DAOS-14686 possibly is not an issue, user did not select right obj class.

gnailzenh:
this patch is also in 2.4 right? If we change this, future upgrade is going to have data placement issue?

wangdi1:
> the original patch was to handle the RF concern right? (if you put multiple shards in the same target during rebuild, after rebuild lost the tgt it will break RF assumption), DAOS-14686 possibly is not an issue, user did not select right obj class.

As long as there are enough domain in the current system, the placement algorithm will only put shards from different group in the same target.   I added the test to verify this.

So even losing the target which has multiple shards should not lose the data.

wangdi1:
> this patch is also in 2.4 right? If we change this, future upgrade is going to have data placement issue?

GX will be convert to the exact group number anyway when the object is created (exists) in the container.  So we just change the GX conversion method, so it should not cause any compatibility issue here.

liuxuezhao (approved):

mchaarawi:
@liuxuezhao could you check if this PR is still valid / needed? tia

PR #13431, opened 2023-12-04
DAOS-14714 vos: return fetch handle only when vos_fetch_begin succeed - b24

Nasf-Fan:
The old logic in vos_fetch_begin() may assign io_context handle to the
output parameter too early, but some subsequent logic may destroy such
handle after that because of some failure. That can cause the returned
handle to be invalid as to related memory to be double freed.

Priority: 2

### Gatekeeper:

### DIFF
--- a/src/vos/vos_io.c
+++ b/src/vos/vos_io.c
@@ -1492,28 +1492,24 @@ vos_fetch_begin(daos_handle_t coh, daos_unit_oid_t oid, daos_epoch_t epoch,
 		       &rc, false)) {
 		if (rc == 0) {
 			if (ioc->ic_read_ts_only)
-				goto set_ioc;
+				goto out;
+
 			if (ioc->ic_obj != NULL &&
 			    has_uncertainty(ioc, &ioc->ic_obj->obj_ilog_info))
 				goto fetch_dkey;
 			for (i = 0; i < iod_nr; i++)
 				iod_empty_sgl(ioc, i);
-			goto set_ioc;
 		}
 		goto out;
 	}
 fetch_dkey:
 	if (dkey == NULL || dkey->iov_len == 0) {
-		if (ioc->ic_read_ts_only)
-			goto set_ioc;
-		D_GOTO(out, rc = -DER_INVAL);
+		if (!ioc->ic_read_ts_only)
+			rc = -DER_INVAL;
+	} else {
+		rc = dkey_fetch(ioc, dkey);
 	}
 
-	rc = dkey_fetch(ioc, dkey);
-	if (rc != 0)
-		goto out;
-set_ioc:
-	*ioh = vos_ioc2ioh(ioc);
 out:
 	vos_dth_set(NULL, ioc->ic_cont->vc_pool->vp_sysdb);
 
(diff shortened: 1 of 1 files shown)


Nasf-Fan:
cont24 failed for DAOS-14353, not related with the patch.

liuxuezhao (dismissed):

src/vos/vos_io.c
```diff
@@ -1532,9 +1532,11 @@ vos_fetch_begin(daos_handle_t coh, daos_unit_oid_t oid, daos_epoch_t epoch,
 	if (rc != 0) {
 		daos_recx_ep_list_free(ioc->ic_recx_lists, ioc->ic_iod_nr);
 		ioc->ic_recx_lists = NULL;
-		return vos_fetch_end(vos_ioc2ioh(ioc), NULL, rc);
+		rc = vos_fetch_end(vos_ioc2ioh(ioc), NULL, rc);
+		*ioh = DAOS_HDL_INVAL;
```
janekmi, line 1536: It is still risky to set and later invalidate the output handle if you experience an error later along the way. It is too easy to overlook another case in the future.

Wouldn't it be simpler to set it only when you are 100% done and nothing can fail?
- replace all `goto set_ioc` would be replaced with `goto out`,
- the `set_ioc` label would be removed
- `*ioh = vos_ioc2ioh(ioc);` would be moved after the last `if`.

So you can exit wherever you would like and the `ioh` would be set only if everything else succeeds. I believe it would be more future-proof.

Please consider my suggestion. Other than that your solution is correct.

janekmi (changes requested):

src/vos/vos_io.c
Nasf-Fan: Your suggestion is reasonable. I will refresh the patch. Hope we still can catch 2.4 timeline.

Nasf-Fan:
> https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-13431/2/execution/node/1322/log

https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-13431/2/artifact/Functional%20Hardware%20Medium%20Verbs%20Provider/launch/functional_hardware_medium_verbs_provider/job.log/*view*/

The failure is caused by DAOS-14052/DAOS-14450, not related with the patch.

src/vos/vos_io.c
```diff
@@ -1532,9 +1528,13 @@ vos_fetch_begin(daos_handle_t coh, daos_unit_oid_t oid, daos_epoch_t epoch,
...
```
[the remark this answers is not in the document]
Nasf-Fan: Generally, as you pointed out, the caller needs to check the return value before accessing the output "ioh". That is the regular logic. But sometime, be as some cleanup logic, the caller may check the "ioh" even if under "rc != 0" case, because such "rc != 0" may be caused by other reason, such as "if (rc != 0) goto out;" and check "ioh" behind "out" label.
 
Here, in my patch, I set DAOS_HDL_INVAL under failure case just for such purpose. So if need refresh the patch, we should set "ioh" as DAOS_HDL_INVAL under all failure cases in vos_fetch_begin(). I think we can do that in another PR since it is not related with this ticket. Then we will not block 2.4 landing. How do you think?

janekmi (approved):

PR #13455, opened 2023-12-07
DAOS-14727 pool: Clean orphaned NVMe blobs on create

mjmac:
When creating a pool on a target, add a step to clean
up any orphaned NVMe blobs. Improves support for
reintegration of a rank that is missing the VOS
metadata for a pool.

Required-githooks: true
Change-Id: I38a314f1e5f140f1d149246802e47385e1c00311


NiuYawei (dismissed):

src/mgmt/srv_target.c
```diff
@@ -1075,6 +1087,14 @@ ds_mgmt_hdlr_tgt_create(crt_rpc_t *tc_req)
 	if (tca.tca_newborn != NULL) {
 		struct vos_pool_arg vpa = {0};
 
+		/* clear any orphaned blobs that are associated with this uuid */
+		rc = clear_vos_pool(tc_in->tc_pool_uuid);
+		if (rc != 0) {
+			D_ERROR(DF_UUID ": failed to clear orphan blobs: " DF_RC "\n",
```
kccain, line 1093: here you may want to change the D_ERROR() to a new macro DL_ERROR(rc, "failed to clear orphan blobs") (no need for DF_RC, \n, and DP_RC(). We are incrementally phasing out D_ERROR() calls.

src/mgmt/srv_target.c
```diff
@@ -1075,6 +1087,14 @@ ds_mgmt_hdlr_tgt_create(crt_rpc_t *tc_req)
 	if (tca.tca_newborn != NULL) {
 		struct vos_pool_arg vpa = {0};
 
+		/* clear any orphaned blobs that are associated with this uuid */
```
kccain, line 1090: No change requested necessarily - just adding some analysis here. The biggest thing I'd like to be sure of is that a retry MGMT_TGT_CREATE corpc will not clean up blobs created in the earlier handling. It seems that the first handling will have created the files in the NEWBORNS directory in "SCM" so in the retry handler tca.tca_newborn == NULL. So I think this will successfully avoid clear_vos_pool()

In a separate chat we had also considered another possible avenue of implementation to update the MGMT_TGT_CREATE cart RPC definition to have a boolean to ask for the clearing to occur. And it could be engaged just by pool reintegration / as opposed to pool create execution flows. But I think settled on the current implementation approach as preferred.

mjmac:
> May benefit from running these changes through Jenkins CI / perhaps with Features: pool.

Done. I merged master and changed to use the new DL_ERROR macro. That commit has the `Features: pool` pragma, seems like it's running fine.

NiuYawei (approved):

kccain (approved):

kccain:
> Test stage Functional Hardware Medium Verbs Provider completed with status UNSTABLE. https://build.hpdd.intel.com/job/daos-stack/job/daos//view/change-requests/job/PR-13455/3/testReport/

The single test failure seems to be an instance of https://daosio.atlassian.net/browse/DAOS-14353

From: https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-13455/3/testReport/junit/FTEST_daos_test/DaosCoreTest-DAOS_Container/Test_Hardware___Functional_Hardware_Medium_Verbs_Provider___CONT27__container_REDUN_LVL_and_RF_test/

`"DER_BUSY" != "DER_SUCCESS"
src/tests/suite/daos_container.c:2932: error: Failure!`

PR #13465, opened 2023-12-08
DAOS-14707 control: Allow multiple agents simultaneously

kjacque:
This allows for the case where multiple agents are used to allow a client connections to multiple systems. The agent log prints a NOTICE-level message to indicate the presence of other daos_agent processes on startup.

Features: control

Required-githooks: true

### Gatekeeper:


Michael-Hennecke (approved):

kjacque:
> This seems reasonable to me... Only question is what happens when two agents are started with the same config. I assume that one will fail because it can't create the socket, but is the error message clear in that case?

When I attempted this, both agents cheerfully started without issue. The agent automatically deletes and re-create the socket file, which does not actually disconnect the old socket. So the first agent thinks it is still listening, although it wouldn't get any new connections on that socket.

I wonder if we should remove the auto-delete of the socket file. It looks like we're cleaning it up during a normal daos_agent shutdown. If someone killed it, though, obviously there would be a dead socket file left over, and that would need to be manually deleted.

knard-intel:
> > This seems reasonable to me... Only question is what happens when two agents are started with the same config. I assume that one will fail because it can't create the socket, but is the error message clear in that case?
> 
> When I attempted this, both agents cheerfully started without issue. The agent automatically deletes and re-create the socket file, which does not actually disconnect the old socket. So the first agent thinks it is still listening, although it wouldn't get any new connections on that socket.
> 
> I wonder if we should remove the auto-delete of the socket file. It looks like we're cleaning it up during a normal daos_agent shutdown. If someone killed it, though, obviously there would be a dead socket file left over, and that would need to be manually deleted.

From what I understand, the agent is removing the socket during the shutdown.  Thus, in your use case, if the first agent is gracefully shutdown, does he will remove the newly created socket by the second agent.

tanabarr (approved):

knard-intel (approved):
LGTM

mjmac (approved):
Looks good. Nice solution for the socket file conflict!

knard-intel (approved):

tanabarr (approved):

Michael-Hennecke:
> > From what I understand, the agent is removing the socket during the shutdown. Thus, in your use case, if the first agent is gracefully shutdown, does he will remove the newly created socket by the second agent.
> 
> Yeah, it would. I think the most efficient way to deal with this, while still allowing two agents, is to not delete the socket on startup. I'll make that change, with a clarifying error message for the admin.

I believe that the "cleanup-on-start" was a conscious decision after discussing more complicated scenarios of cleaning up at the end of a job. it is much more cumbersome to do a cleanup after hangs or aborts (at the end of a job). if we just clean up before we start, that's much easier.  but this certainly makes it more tricky to start two agents on the same node...

is there a way to check if somebody is still at the other end of that socket connection before cleaning it up?

Michael-Hennecke (changes requested):
see comment above

kjacque:
> is there a way to check if somebody is still at the other end of that socket connection before cleaning it up?

Yes, this is actually the solution I ended up using. We attempt to connect to the socket. If there's no error, someone is listening. If connection is refused, the file exists but no one is listening, and we clean it up.

mjmac (approved):

Michael-Hennecke (approved):
lgtm

PR #13477, opened 2023-12-11
DAOS-14769 client: reset crt related global variables after fork()

wiliamhuang:
Once fork() is called, a child process is created and proceeds from the fork point. The parent process may still need to utilize the existing context. It is the parent process's responsibility to destroy the context when the parent process ends. The child process inherited the memory from its parent process due to fork. We do need to reset the global variables and create a fresh context inside the child process.

Features: pil4dfs

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/cart/crt_init.c
+++ b/src/cart/crt_init.c
@@ -52,6 +52,16 @@ crt_lib_init(void)
 	crt_gdata.cg_iv_inline_limit = 19456; /* 19KB */
 }
 
+void
+crt_reset_afterfork(void)
+{
+	gdata_init_flag = 0;
+	memset(&crt_gdata, 0, sizeof(struct crt_gdata));
+	memset(&crt_plugin_gdata, 0, sizeof(struct crt_plugin_gdata));
+	memset(g_prov_settings_applied, 0, sizeof(bool) * CRT_PROV_COUNT);
+	crt_lib_init();
+}
+
 /* Library deinit */
 static void
 crt_lib_fini(void)
--- a/src/include/cart/api.h
+++ b/src/include/cart/api.h
@@ -2227,10 +2227,6 @@ int crt_swim_init(int crt_ctx_idx);
  */
 void crt_swim_fini(void);
 
-/** Reset global variables and call crt_lib_init() after fork().
- */
-void crt_reset_afterfork(void);
-
 #define crt_proc__Bool			crt_proc_bool
 #define crt_proc_d_rank_t		crt_proc_uint32_t
 #define crt_proc_int			crt_proc_int32_t
--- a/src/include/cart/api.h
+++ b/src/include/cart/api.h
@@ -2227,6 +2227,10 @@ int crt_swim_init(int crt_ctx_idx);
  */
 void crt_swim_fini(void);
 
+/** Reset global variables and call crt_lib_init() after fork().
+ */
+void crt_reset_afterfork(void);
+
 #define crt_proc__Bool			crt_proc_bool
 #define crt_proc_d_rank_t		crt_proc_uint32_t
 #define crt_proc_int			crt_proc_int32_t
--- a/src/cart/crt_internal.h
+++ b/src/cart/crt_internal.h
@@ -97,4 +97,10 @@ crt_hdlr_ctl_get_pid(crt_rpc_t *rpc_req);
 
 void
 crt_iv_init(crt_init_options_t *ops);
+
+/** Reset global variables and call crt_lib_init() after fork().
+ */
+void
+crt_reset_afterfork(void);
+
 #endif /* __CRT_INTERNAL_H__ */
--- a/src/client/api/event.c
+++ b/src/client/api/event.c
@@ -15,6 +15,7 @@
 
 #include "client_internal.h"
 #include <daos/rpc.h>
+#include <cart/api.h>
 
 /** thread-private event */
 static __thread daos_event_t	ev_thpriv;
@@ -110,6 +111,7 @@ crt:
 int
 daos_eq_lib_reset_after_fork(void)
 {
+	crt_reset_afterfork();
 	eq_ref            = 0;
 	ev_thpriv_is_init = false;
 	return daos_eq_lib_init();
--- a/src/client/api/event.c
+++ b/src/client/api/event.c
@@ -15,7 +15,7 @@
 
 #include "client_internal.h"
 #include <daos/rpc.h>
-#include <cart/api.h>
+#include <../cart/crt_internal.h>
 
 /** thread-private event */
 static __thread daos_event_t	ev_thpriv;


src/cart/crt_init.c
```diff
@@ -52,6 +52,16 @@ crt_lib_init(void)
 	crt_gdata.cg_iv_inline_limit = 19456; /* 19KB */
 }
 
+void
+crt_reset_afterfork(void)
```
ashleypittman, line 56: `crt_reset` or `cart_lib_reset` would be better names here, the code doesn't know or care when you call it.

Should there be any locking here?  There appears to be no checking of thread safety at all.

There are at least two mutexes being wiped here that should really be finalised before being wiped but it's probably better to change them to be static variables on their own and use static initialisers for them rather than calling _INIT and _FINI on them.

ashleypittman (changes requested):
Why are you sending PRs for review before they've even finished building, let alone testing?

Unit tests would be great for this to lock in the progress, even if it doesn't yet move us forward on anything we can do with system testing.

src/cart/crt_init.c
wiliamhuang: Thank you! This function will be called right after fork() inside the newly spawned child process. I think there is no need to use lock. There is only a single thread in a new process.
It would be more clear with "_after_fork". I am not aware of any other situations we need to call this function at this moment. If we do need to call it somewhere else, we could rename it later.

src/include/cart/api.h
```diff
@@ -2227,6 +2227,10 @@ int crt_swim_init(int crt_ctx_idx);
  */
 void crt_swim_fini(void);
 
+/** Reset global variables and call crt_lib_init() after fork().
+ */
+void crt_reset_afterfork(void);
+
```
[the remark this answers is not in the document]
wiliamhuang: $ ls src/include/cart
api.h  iv.h  swim.h  types.h

I am not aware of any related internal head file. If really necessary, we might create one.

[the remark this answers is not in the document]
ashleypittman: src/cart/crt_internal.h?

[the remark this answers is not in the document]
wiliamhuang: Good to know. I will move it there. Thank you very much!

src/cart/crt_init.c
wiliamhuang: You are right. We could move this function to src/cart/crt_internal.h as you suggested. Thank you!

soumagne:
I understand what this PR is trying to do but I think this requires further review as memsetting global data like that could have hidden side effects.

wiliamhuang:
> I understand what this PR is trying to do but I think this requires further review as memsetting global data like that could have hidden side effects.

This function is called inside the new child process after fork(). It does not affect the parent process's memory.

src/cart/crt_init.c
frostedcmos: I am not sure this approach is safe:

if we assume that fork() can happen at any time after crt_init() was called, then we need to do a clean finalization of all cart, and not just reset some variables.

For example, assume an original process allocated few contexts and initialized a provider.
After that a fork() happens and this function gets eventually called.
Even though a forked child can in theory now re-init cart, you now have lost all references to cart/mercury contexts and other data. 

Is there a reason why in daos lib you cant simply call crt_finalize() followed by crt_init() again?

frostedcmos (changes requested):

src/cart/crt_init.c
wiliamhuang: Once fork() is called, a child process is created and proceed from the fork point. The parent process may still need to utilize the existing context, so we can NOT call crt_finalize() at this point. It is the parent process's responsibility to destroy the context when the parent process ends. The child process inherited the memory from its parent process due to fork. We do need to reset the global variables and create a fresh context inside the child process. The parent and child process have separate memory.

mchaarawi:
> Why are you sending PRs for review before they've even finished building, let alone testing?

@ashleypittman I think you are well aware of our CI queue and how long that takes. so actually getting review feedback is actually more productive if the PR has not started the HW changes to capture the changes and rerun before consuming those resources. but yea ideally if the CI bw is good, it's better to wait till the patch has passed before asking for reviewers.

mchaarawi:
> > Why are you sending PRs for review before they've even finished building, let alone testing?
> 
> @ashleypittman I think you are well aware of our CI queue and how long that takes. so getting review feedback is actually more productive if the PR has not started the HW changes to capture the changes and rerun before consuming those resources. but yea ideally if the CI bw is good, it's better to wait till the patch has passed before asking for reviewers.

maybe a fair compromise is to wait for unit test stage to complete before asking for reviews :-)

wiliamhuang:
> maybe a fair compromise is to wait for unit test stage to complete before asking for reviews :-)

ok. I will follow this protocol later. Thank you!

src/cart/crt_init.c
frostedcmos: I understand now what you are trying to do, but i think it will require a bit more changes.

crt_gdata contains number of elements that get allocated either at init or during cart usage, and would need to be free-ed first before setting gdata_init_flag = 0 and memsetting, in order to not leak memory.

For example crt_gdata->cg_opc_map will be allocated based on registered RPCs in the parent.

There are also other potential SW things you would need to deallocate and reset.
For example crt_group.c contains global cart group list and associated lock.

  19 /* global CRT group list */
  20 D_LIST_HEAD(crt_grp_list);


It also raises a question what to do with cart contexts if any are already allocated.
As you mentioned we cannot destroy the context, as it is parents responsibility, but at the same time we should destroy all SW-related parts of the cart context to not leak memory.

the main concern here is that cart contexts contain rank-to-hg_handle caches and these would be bad to leak after fork().
Similarly crt_gdata contains rank-to-uri cache that also needs to be nuked to not leak memory.

mchaarawi: why would there be leaked memory on the child if one just reset the init flag?
all memory inherited for the child on fork is copy-on-write, no?

frostedcmos: If i understand it correctly with a copy-on-write, upon fork() the pages will get marked as read-only, parent and child both point to the same pages and a write by either a parent or child will result in page fault, followed by a copy of a particular page.

So with this, a parent writing to crt_gdata right after the fork() will result in a child getting its own copy (parents original), and when reset that copy will be lost.
Also since copy-on-write is on per-page boundaries, writing to something unrelated could result in a copy.

Do you think its safe in this case?

wiliamhuang: @frostedcmos . crt_gdata is set zero inside child process here. Do we need to re-initialize "crt_grp_list" too here? Are there any other global variables needed to reset?

mchaarawi: > If i understand it correctly with a copy-on-write, upon fork() the pages will get marked as read-only, parent and child both point to the same pages and a write by either a parent or child will result in page fault, followed by a copy of a particular page.

yea indeed. i did not consider that the parent can change this global data struct.

wiliamhuang:
Test / Functional on EL 8.8 / 1-./dfuse/mu_perms.py failed. I will debug the issue locally and update this PR soon.

wiliamhuang:
As Ashley suggested, I tried letting daos_init() use fd large than 10. The hang issue is gone. configure script uses specific fd, e.g., 7, 6, 5 which could be used by daos. This may cause issues. I am closing this PR. We will create another PR to temporarily block unused low fd (less than 10) before daos_init().

PR #13483, opened 2023-12-12
DAOS-14532 gurt: Replace environment APIs hook

knard-intel:
### Description

This PR is a subset of the PR #13250 allowing thread safe management of environment variables: it has been split into smaller PRs to facilitate the review process.
This PR mainly add thread safe environment variables management functions.
It also remove and replace old non thread safe custom environment management functions.
Finally, it replace the `setenv()` function with` d_setenv()`.

### Gatekeeper:


src/gurt/misc.c
```diff
@@ -947,149 +951,478 @@ d_rank_range_list_free(d_rank_range_list_t *range_list)
...
-	if (!dis_integer_str(env_val)) {
-		D_ERROR("ENV %s is not integer.\n", env_val);
-		return;
+#if UINT_MAX != ULONG_MAX
+	assert(sizeof(unsigned) < sizeof(unsigned long));
+	if (val > UINT_MAX) {
+		rc = -DER_INVAL;
+		goto out;
 	}
+#endif
```
wangshilong, line 1224: [suggestion] i think for somehow dis_unsigned_str() is duplicated as strtoull()?
we could remove dis_unsigned_str() and just use strtoull()?

knard-intel: The best solution from my point of view should be to remove all the unsigned and to replace them with uint32_t or uint64_t.  However, this is beyond the scope of this PR.
For this PR, I would prefer to keep these three functions as the `sizeof()` of the `unsigned` is platform dependent: I would like to avoid any unexpected overflow.  In a follow-up PR, I could replace the `d_getenv_uint()` with relevant `d_getenv_uintXXt()` and update the code accordingly.

wangshilong: I am fine with this, i thought dis_unsigned_str() is unnecessary and might be removed.

wangshilong (dismissed):

src/gurt/misc.c
knard-intel: - [ ] Factorize unsigned fonctions.

src/gurt/misc.c
```diff
@@ -947,149 +951,478 @@ d_rank_range_list_free(d_rank_range_list_t *range_list)
...
+ *
+ * \param[in]	name		name of the environment variable.
+ * \return			0 on success, a negative value on error.
+ */
+int
+d_clearenv(void)
+{
+	int rc;
+
+	pthread_rwlock_wrlock(&d_env_lock);
```
jolivier23, line 1421: Can you elaborate why this is safer than the hook version?   I think the issue we have seen is due to threads calling getenv in other libraries, which in this case would not be locked with respect to calls in daos libraries.  Whereas the hooked version would ensure the lock is acquired for all calls, no?

knard-intel: The hook version can have some side effect with other third party library such as mvapich and eventually lead to segfault.  You could find in the following comment a detailed description of such situation https://daosio.atlassian.net/browse/DAOS-14532?focusedCommentId=121770

I tried to overcome this issue with updating hook or playing with library constructor.  At the end, I did not find an acceptable solution which could not have some side effect with other third party libraries.

After some discussion with @johannlombardi and @gnailzenh , it appeared that the least bad solution was to have our own thread safe functions for managing environment.  From my side, the only disadvantage of this solution is to be much more intrusive in our code.  Happily, we already have some dedicated non thread safe functions for managing the environment.

jolivier23 (approved):

src/gurt/misc.c
jolivier23: we see an intermittent issue with fio that manifests as follows
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007d9faabc195d in getenv () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x7d9fad4335c0 (LWP 60560))]
Missing separate debuginfos, use: yum debuginfo-install boost-atomic-1.66.0-13.el8.x86_64 boost-chrono-1.66.0-13.el8.x86_64 boost-date-time-1.66.0-13.el8.x86_64 boost-iostreams-1.66.0-13.el8.x86_64 boost-program-options-1.66.0-13.el8.x86_64 boost-random-1.66.0-13.el8.x86_64 boost-regex-1.66.0-13.el8.x86_64 boost-system-1.66.0-13.el8.x86_64 boost-thread-1.66.0-13.el8.x86_64 daxctl-libs-71.1-4.el8.x86_64 glibc-2.28-225.el8.x86_64 libaio-0.3.112-1.el8.x86_64 libblkid-2.32.1-42.el8_8.x86_64 libgcc-8.5.0-18.el8.x86_64 libibverbs-44.0-2.el8.1.x86_64 libicu-60.3-2.el8_1.x86_64 libmount-2.32.1-42.el8_8.x86_64 libnl3-3.7.0-1.el8.x86_64 libpmem-1.6.1-1.el8.x86_64 libpmemblk-1.6.1-1.el8.x86_64 librados2-12.2.7-9.el8.x86_64 librbd1-12.2.7-9.el8.x86_64 librdmacm-44.0-2.el8.1.x86_64 libselinux-2.9-8.el8.x86_64 libstdc++-8.5.0-18.el8.x86_64 libyaml-0.1.7-5.el8.x86_64 lz4-libs-1.8.3-3.el8_4.x86_64 ndctl-libs-71.1-4.el8.x86_64 nspr-4.34.0-3.el8_6.x86_64 nss-util-3.79.0-11.el8_7.x86_64 numactl-libs-2.0.12-13.el8.x86_64 openssl-libs-1.1.1k-9.el8_7.x86_64 pcre2-10.32-3.el8_6.x86_64 protobuf-c-1.3.0-6.el8.x86_64 systemd-libs-239-74.el8_8.x86_64 zlib-1.2.11-21.el8_7.x86_64
(gdb) bt
#0  0x00007d9faabc195d in getenv () from /lib64/libc.so.6
#1  0x00005b1f2183098c in check_status_file () at stat.c:2372
#2  0x00005b1f21839c25 in check_for_running_stats () at stat.c:2411
#3  0x00005b1f2186cf51 in do_usleep (usecs=10000) at backend.c:2460
#4  run_threads (sk_out=sk_out@entry=0x0) at backend.c:2460
#5  0x00005b1f2186d5b2 in fio_backend (sk_out=0x0) at backend.c:2513
#6  0x00005b1f2181c595 in main (argc=16, argv=0x7ffe8dcfb228, envp=<optimized out>) at fio.c:60

It doesn't show dlopen in the callstack.  Do you think it's related to this issue?

src/include/gurt/common.h
```diff
@@ -574,12 +574,37 @@ d_sgl_buf_copy(d_sg_list_t *dst_sgl, d_sg_list_t *src_sgl)
...
+bool
+d_isenv_def(char *name);
+int
+d_getenv_str(char *str_val, size_t str_size, const char *name);
+int
+d_agetenv_str(char **str_val, const char *name);
+void
+d_free_env(char **str_val);
+int
+d_getenv_bool(bool *bool_val, const char *env);
```
ashleypittman, line 586: This is changing the API and ABI isn't it?  Existing applications which use this interface will need code changes to work, but if they build with one version and run with another they could also segfault.

jolivier23: I had a similar thought but then kind of assumed that the only user of these APIs are DAOS libraries that will be shipped with the changes.

src/gurt/misc.c
knard-intel: > [suggestion] i think for somehow dis_unsigned_str() is duplicated as strtoull()? we could remove dis_unsigned_str() and just use strtoull()?

Sorry, i did not well understood your question regarding the function `dis_unsigned_str()`.  At the beginning, I also wanted to remove it.  However, `strtoull()` is less restrictive and allow to have spaces before the first digit and even to have a negative integer.  To not change the behavior of the `d_getenv_uint()` functions, I decided to kept the `dis_unsigned_str()` function.

knard-intel: The code of the getenv of unsigned integer function has been factorized in commit 193c24c415c6e4229be3760821b8e7c9adeb3416

src/gurt/misc.c
knard-intel: There is a high probably that it could be related: however it is hard for me to be 100% sure without a more deep investigation in the core file and the related code source.
As for @daltonbohning issue with mvapich, you could use the full patch  #13250 to check if it also solves your issue with fio.

src/include/gurt/common.h
knard-intel: At this time, the only user of this new APIs are indeed the DAOS libraries and thus should not have any side effects for the client applications linked with DAOS.  They will just use the standard non-thread safe `getenv()` functions.

jolivier23 (approved):

wangshilong (approved):

src/include/gurt/common.h
```diff
@@ -574,12 +574,37 @@ d_sgl_buf_copy(d_sg_list_t *dst_sgl, d_sg_list_t *src_sgl)
...
+d_setenv(const char *name, const char *value, int overwrite);
+int
+d_unsetenv(const char *name);
+int
+d_clearenv(void);
+
+int
+d_write_string_buffer(struct d_string_buffer_t *buf, const char *fmt, ...);
+void
+d_free_string(struct d_string_buffer_t *buf);
```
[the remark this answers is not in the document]
knard-intel: - [ ] Rename `d_free_string()` to `d_free_string()`

src/bio/bio_xstream.c
```diff
@@ -195,9 +195,9 @@ set_faulty_criteria(void)
 	 */
 	glb_criteria.fc_max_csum_errs = UINT32_MAX;
 
-	d_getenv_bool("DAOS_NVME_AUTO_FAULTY_ENABLED", &glb_criteria.fc_enabled);
-	d_getenv_int("DAOS_NVME_AUTO_FAULTY_IO", &glb_criteria.fc_max_io_errs);
-	d_getenv_int("DAOS_NVME_AUTO_FAULTY_CSUM", &glb_criteria.fc_max_csum_errs);
+	d_getenv_bool(&glb_criteria.fc_enabled, "DAOS_NVME_AUTO_FAULTY_ENABLED");
```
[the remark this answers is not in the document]
knard-intel: - [ ] Revert interfaces update

jolivier23 (changes requested):
Removing my +1 since the API change needs to be reverted.

knard-intel:
@ashleypittman , my apologies for the misunderstanding I was thinking that you were talking about the standard `getenv()` function and not the `daos_getenv_xxx()` function already existing.  I was expecting that these last ones were only used internaly with DAOS.  For sure, I will revert the interface modification to preserve the compatibility.

src/bio/bio_xstream.c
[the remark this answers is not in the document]
knard-intel: By any chance, do you know if we have some functional tests using MPI_IO ?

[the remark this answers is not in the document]
daltonbohning: You can use `Features: mpiio`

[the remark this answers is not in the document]
ashleypittman: In the past where this has come up then what functional testing we have uses a weird system where:

1. PRs and landings to downstream components test against master daos.
2. PRs and landings to daos test with the latest built downstream component.

The unfortunate implication here is that we have some testing but it's limited and if we change the API or ABI and it's incompatible then we don't notice until there's a new PR to the downstream component or it gets rebuilt for another reason.  If we had test coverage here then it would fail at build time, as it is it'll fail (obscurely) at runtime, but only if the code is called.

It's an area we need to work on.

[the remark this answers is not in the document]
knard-intel: Fixed with commit ac764d169812b82850bd1bf1d9d9177415c2f857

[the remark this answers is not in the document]
knard-intel: > * [x]  Revert interfaces update

Fixed with commit ac764d169812b82850bd1bf1d9d9177415c2f857

src/include/gurt/common.h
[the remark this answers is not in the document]
knard-intel: > In fact the the `d_free_string()` was not introduce by me. More details could be found at [d6d6537#diff-2f91f29b7a93e5a1c4b845a036cb5f7c7193d00655bdd1b834f456b188f769b8](https://github.com/daos-stack/daos/commit/d6d6537b95a1c221e15d271a0afec8f2348dfc84#diff-2f91f29b7a93e5a1c4b845a036cb5f7c7193d00655bdd1b834f456b188f769b8)
> 
> I just changed the formating to make the C linter happy 😉
> 
> * [x]  Rename `d_free_env() `-> `d_free_env_str()`

Fixed with commit ac764d169812b82850bd1bf1d9d9177415c2f857

src/gurt/misc.c
```diff
@@ -947,149 +951,451 @@ d_rank_range_list_free(d_rank_range_list_t *range_list)
...
+ *
+ * \param[in]		name		name of the environment variable.
+ * \return				true iff the environment variable is defined.
+ */
+bool
+d_isenv_def(char *name)
+{
+	char *env;
+
+	pthread_rwlock_rdlock(&d_env_lock);
```
ashleypittman, line 991: These will be coverity warnings unless you either check the return code or use our macros.

ashleypittman: I did wonder.  Does the logging ever call getenv though?

I made the same change to the locking in the logging as that was causing recursion and found out about the coverity issue that way.

knard-intel: From what I remember, I experiment infinite recursion with the function `d_logfac_is_enabled()` from `src/gurt/dlog.c`.  There is other functions from this file which could also probably lead to the same kind of issues.

knard-intel: > Thanks for the info. Sadly, I will not be able to use the D_RWLOCK_RDLOCK macros as it use the DAOS logging facilities which could lead to infinite recursive calls. Thus I will check the return code is a similar way as this last macros.
> 
> * [x]  Checks return code when using standard function calls such as `pthread_rwlock_rdlock()`

Fixed with commit 2791b9323afa26af021220b35b0cffee1e2a0a63

knard-intel:
The only CI issue with the functional test [erasurecode_online_rebuild_mdtest_py](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-13483/8/testReport/FTEST_erasurecode/EcodOnlineRebuildMdtest/Test_Hardware___Functional_Hardware_Large___2___erasurecode_online_rebuild_mdtest_py_EcodOnlineRebuildMdtest_test_ec_online_rebuild_mdtest_run_container_hosts_servers_8_server_mdtest_client_processes_np_48_mdtest_api_dfs_objectclass_pool_server_config_engines_0_storage_0_1_setup_01ee/) is a known issue [DAOS-14845](https://daosio.atlassian.net/browse/DAOS-14845).

wangshilong (approved):

jolivier23 (approved):

mchaarawi (approved):

PR #12330, opened 2023-06-09
DAOS-11406 rdb: Use VOS local TXs

liw:
Every RDB (update) TX invokes multiple VOS writes, each of which
performs a WAL commit. This patch reduces those WAL commits to one per
RDB TX by utilizing the new VOS local TXs to group all VOS writes of one
RDB TX into one VOS local TX.

  - An rdb_vos_tx_t handle is introduced to hide the vos/dtx details
    from the rest of rdb.

  - Some rdb_lc_discard calls are no longer necessary, for the VOS local
    TX mechanism will discard uncommited writes more efficiently.

  - Some "nvops" ("number of VOS TX operations") functions are added to
    provide dtx_begin with an upper bound of the number of VOS TX
    operations that will be invoked using a VOS TX handle. Hopefully
    this will become unnecessary in the future.

  - The rdb_tx_hdr decoding in rdb_tx_apply should use the actual length
    of the data, rather than the expected length.

  - Two places related to rdb_tx_hdr need to be refactored a bit in
    order to avoid compiler warnings along the line of "hdr may be used
    uninitialized". It's unclear why the compiler warnings suddenly come
    up.

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate watchers.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
[...]

PR #13495, opened 2023-12-13
DAOS-11404 dtx: Introduce local transactions

janekmi:
Features: daosio

Required-githooks: true

### Gatekeeper:


utils/utest.yaml
```diff
@@ -48,6 +48,10 @@
   tests:
     - cmd: ["src/gurt/tests/test_gurt"]
     - cmd: ["src/gurt/tests/test_gurt_telem_producer"]
+- name: DTX
+  base: "PREFIX"
+  tests:
+    - cmd: ["bin/dtx_tests"]
```
jolivier23, line 54: if you're installing them, you will want to update the daos.spec file.   Unless you have a really good reason that these need to be installed, I'd recommend running from build directory as with many other tests.

janekmi: I believe in regards it allows testing local VOS transactions it should be installed whenever `vos_tests` is.

Done.

src/vos/vos_io.c
```diff
@@ -2385,6 +2384,38 @@ update_cancel(struct vos_io_context *ioc)
...
+ * local transaction.
+ */
+static int
+vos_insert_oid(struct dtx_handle *dth, struct vos_container *cont, daos_unit_oid_t *oid)
+{
+	/** The array has to grow to accommodate the next record. */
+	if (dth->dth_local_oid_cnt == dth->dth_local_oid_cap) {
+		struct dtx_local_oid_record *oid_array;
+
+		D_ALLOC_ARRAY(oid_array, dth->dth_local_oid_cap << 1);
```
jolivier23, line 2398: any reason not to use D_REALLOC_ARRAY?   This will do the memcpy for you so may be less error prone

src/vos/vos_dtx.c
```diff
@@ -3249,3 +3249,47 @@ vos_dtx_renew_epoch(struct dtx_handle *dth)
...
+	}
+
+	/**
+	 * RDB is intended as the main user of local transactions. RDB is known
+	 * to engage over 32 OIDs per transaction. Hence, the initial value is
+	 * hoped to accommodate all of them.
+	 */
+	dth->dth_local_oid_cap = (1 << 6);
+	dth->dth_local_oid_cnt = 0;
+	D_ALLOC_ARRAY(dth->dth_local_oid_array, dth->dth_local_oid_cap);
```
jolivier23, line 3271: In other places, to avoid having allocation on the critical path, we would have some embedded ones and only do allocation if we go over the embedded amount

utils/utest.yaml
jolivier23: yeah, we aren't very consistent with what we do and don't install.

jolivier23 (dismissed):

src/vos/vos_common.c
```diff
@@ -352,6 +409,16 @@ vos_tx_end(struct vos_container *cont, struct dtx_handle *dth_in,
...
 
+	if (dth->dth_local) {
+		for (int i = 0; i < dth->dth_local_oid_cnt; ++i) {
+			vos_cont_decref(dth->dth_local_oid_array[i].dor_cont);
+		}
+
+		dth->dth_local_oid_cnt = 0;
+		D_FREE(dth->dth_local_oid_array);
+		dth->dth_local_oid_cap = 0;
+	}
```
Nasf-Fan, line 420: Since the array is allocated via vos_dtx_local_begin(), then release such array via vos_dtx_local_end() seems more clean.

src/vos/vos_io.c
```diff
@@ -2458,6 +2489,10 @@ vos_update_end(daos_handle_t ioh, uint32_t pm_ver, daos_key_t *dkey, int err,
 		goto abort;
 	}
 
+	if (dtx_is_valid_handle(dth) && dth->dth_local) {
+		err = vos_insert_oid(dth, ioc->ic_cont, &ioc->ic_oid);
```
Nasf-Fan, line 2490: Will local TX touch punch logic? if not, please add assert() at the beginning of vos_obj_punch(), then all subsequent punch related logic will not handle local TX cases.

src/vos/vos_dtx.c
janekmi: @liw hinted that RDB uses somewhere between 32 and 64 OIDs per transaction. I assumed having a 64-slot embedded array is a little bit too much so I went with having a 64-slot allocated default array which I believe should allow RDB to go with a single alloc per transaction. No reallocs.

src/dtx/dtx_common.c
```diff
@@ -905,23 +912,38 @@ dtx_handle_init(struct dtx_id *dti, daos_handle_t coh, struct dtx_epoch *epoch,
...
-	if (dth->dth_modification_cnt == 0)
-		return 0;
+	rc = vos_dtx_rsrvd_init(dth);
+	if (rc != 0) {
+		D_ERROR("Failed to allocate space for scm reservations: rc=" DF_RC "\n", DP_RC(rc));
+		return rc;
+	}
 
-	return vos_dtx_rsrvd_init(dth);
+	if (local) {
```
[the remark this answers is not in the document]
janekmi: In this case, `dtx_handle_init()` won't produce initialized DTX handles in all the cases. In the case of the `DTX_LOCAL` it will be half-baked which may be confusing. I think it is better to keep it where it is so the role of the `dtx_handle_init()` function will remain unchanged.

src/vos/vos_ts.c
```diff
@@ -304,7 +304,7 @@ vos_ts_set_allocate(struct vos_ts_set **ts_set, uint64_t flags,
 	vos_kh_clear(standalone);
 
 	*ts_set = NULL;
-	if (!dtx_is_valid_handle(dth)) {
+	if (!dtx_is_valid_handle(dth) || dth->dth_local) {
```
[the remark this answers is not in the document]
janekmi: Done.

src/include/daos_srv/dtx_srv.h
```diff
@@ -44,7 +52,7 @@ struct dtx_handle {
 			struct dtx_memberships	*dth_mbs;
 		};
 	};
-	/** The container handle */
+	/** The container handle or pool handle (local transactions only) */
```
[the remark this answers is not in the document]
janekmi: Done.

src/include/daos_srv/dtx_srv.h
```diff
@@ -135,6 +146,13 @@ struct dtx_handle {
 	/* DTX list to be checked */
 	d_list_t			 dth_share_tbd_list;
 	int				 dth_share_tbd_count;
+
+	/** The count of objects stored in dth_local_oid_array. */
+	uint16_t                          dth_local_oid_cnt;
+	/** The total slots in the dth_local_oid_array. */
+	uint16_t                          dth_local_oid_cap;
+	/** The record of all objects touched by the local transaction. */
+	struct dtx_local_oid_record      *dth_local_oid_array;
```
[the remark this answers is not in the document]
janekmi: Done.

src/vos/vos_io.c
```diff
@@ -2385,6 +2384,38 @@ update_cancel(struct vos_io_context *ioc)
...
+
+		memcpy(&oid_array[0], &dth->dth_local_oid_array[0],
+		       sizeof(*oid) * dth->dth_local_oid_cnt);
+
+		D_FREE(dth->dth_local_oid_array);
+		dth->dth_local_oid_array = oid_array;
+		dth->dth_local_oid_cap <<= 1;
+	}
+
+	struct dtx_local_oid_record *record = &dth->dth_local_oid_array[dth->dth_local_oid_cnt];
```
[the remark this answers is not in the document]
janekmi: Done.

src/vos/vos_io.c
```diff
@@ -2414,7 +2445,7 @@ vos_update_end(daos_handle_t ioh, uint32_t pm_ver, daos_key_t *dkey, int err,
 	tx_started = true;
 
 	/* Commit the CoS DTXs via the IO PMDK transaction. */
-	if (dtx_is_valid_handle(dth) && dth->dth_dti_cos_count > 0 &&
+	if (dtx_is_valid_handle(dth) && !dth->dth_local && dth->dth_dti_cos_count > 0 &&
```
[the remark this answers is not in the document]
janekmi: Done. + an assert just in case.

src/vos/vos_common.c
janekmi: Done.

src/vos/vos_io.c
janekmi: I do not see the relationship between the no need for "touch" (I assume you mean any changes to the punch logic needed specifically for local TX) and the notion that punches are out of the scope for local TX. There is also a simple test for local TX making use `vos_obj_punch()` that works flawlessly so far.

Local TX has to support the punch operation. As far as I am concerned a punch is just a "special value" no need for any special logic to handle it for local TX.

Am I missing something?

Nasf-Fan: If punch is in the scope of local TX, why only record the cont/obj that are touched by the local TX for update? Why not for punch?

janekmi: Good catch. Fixed.

brianjmurrell (commented):
Packaging changes LGTM but I will reserve my +1 for others more familiar with the core changes here.

utils/utest.yaml
jolivier23: no, it's something we need to address at some point to ensure we are consistent.

jolivier23 (dismissed):

src/vos/vos_ilog.c
```diff
@@ -434,8 +434,8 @@ int vos_ilog_update_(struct vos_container *cont, struct ilog_df *ilog,
 		return rc;
 	}
 
-	rc = ilog_update(loh, &max_epr, epr->epr_hi, dtx_is_valid_handle(dth) ?
-			 dth->dth_op_seq : VOS_SUB_OP_MAX, false);
+	rc = ilog_update(loh, &max_epr, epr->epr_hi,
+			 (dtx_is_real_handle(dth) ? dth->dth_op_seq : VOS_SUB_OP_MAX), false);
```
Nasf-Fan, line 438: As my understand, one local TX may contain multiple sub-modifications, these sub-modifications belong to the same transaction, shares the same major epoch, and different minor epoch for each one. Here, the 4th parameter is "minor_eph". If we use "VOS_SUB_OP_MAX" for local TX case without distinguishing sub-modifications, then it will be explained as either "conflict" or "already done". That does no match the semantics.

src/vos/vos_ilog.c
```diff
@@ -531,7 +531,7 @@ vos_ilog_punch_(struct vos_container *cont, struct ilog_df *ilog,
 		return rc;
 	}
 
-	if (dth) {
+	if (dtx_is_real_handle(dth)) {
```
Nasf-Fan, line 534: Similar as above.

src/dtx/dtx_common.c
```diff
@@ -1471,7 +1494,7 @@ dtx_leader_end(struct dtx_leader_handle *dlh, struct ds_cont_hdl *coh, int resul
 /**
  * Prepare the DTX handle in DRAM.
  *
- * \param coh		[IN]	Container handle.
+ * \param coh		[IN]	Container handle or pool handle.
```
Nasf-Fan, line 1497: Rename it as "oh"?

src/vos/vos_common.c
```diff
@@ -243,8 +286,26 @@ vos_tx_end(struct vos_container *cont, struct dtx_handle *dth_in,
...
+		if (dth->dth_rsrvd_cnt > 0 && dth->dth_rsrvd_cnt >= dth->dth_modification_cnt) {
+			/*
+			 * Just do your best to release the SCM reservation. Can't handle another
+			 * error while handling one already anyway.
+			 */
+			(void)vos_publish_scm(vos_pool2umm(pool), *rsrvd_scmp, false /* publish */);
+			D_FREE(*rsrvd_scmp);
+			*rsrvd_scmp = NULL;
+			err         = -DER_NOMEM;
+			goto cancel;
```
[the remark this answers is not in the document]
janekmi: There is a silent assumption that the allocation passed as `rsrvd_scmp` is basically taken ownership of and is passed to the `dth->dth_rsrvds[]`. So later on e.g. `vos_ioc_reserve_fini()` assumes `umem_rsrvd_act_cnt(ioc->ic_rsrvd_scm) == 0`. Since in this case the `dth->dth_rsrvds[]` can't accept another record it is disposed of and the whole transaction is aborted.

It is necessary since when doing transactions locally it is sometimes tricky to correctly calculate the number of submodifications upfront. Hence, it is good to have a safeguard in this case.

Ref: https://daosio.atlassian.net/browse/DAOS-11404?focusedCommentId=123025

src/include/daos_srv/dtx_srv.h
```diff
@@ -110,12 +125,24 @@ struct dtx_handle {
...
+		};
+		struct {
+			/** The count of objects stored in dth_local_oid_array. */
+			uint16_t                     dth_local_oid_cnt;
+			/** The total slots in the dth_local_oid_array. */
+			uint16_t                     dth_local_oid_cap;
+			/** The record of all objects touched by the local transaction. */
+			struct dtx_local_oid_record *dth_local_oid_array;
+		};
+	};
```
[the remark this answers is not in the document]
janekmi: Done.

src/dtx/dtx_common.c
janekmi: Done.

src/vos/vos_ilog.c
liw: @janekmi, I discussed this issue with Fan Yong. We speculate the difficult you face here may be that `rdb` does not increment `dth_op_seq` as `obj` does (via `dtx_sub_init`). If that is the case, would this `rdb` change (https://github.com/daos-stack/daos/pull/12330/commits/961c63a273e567353220c359e8dd24c1be115256) allow you to use `dth_op_seq` in the same manner as the non-local TX case?

janekmi: In the [latest fix](https://github.com/daos-stack/daos/pull/13495/commits/3c72881125bcf601659b0916c5daa4f0295f89b0) to this pull request, I introduced both increasing `dth_op_seq` on each update and punch (a little bit different semantics compared to calling `dtx_sub_init()` but should do the trick) and using `dth_op_seq` instead of `VOS_SUB_OP_MAX` exactly as @Nasf-Fan suggested.

Moreover, I wrote additional unit tests ((here)[https://github.com/daos-stack/daos/pull/13495/commits/7592898e87746d80f985df7672a557f7560f5b9a]) after consulting with @Nasf-Fan to verify whether the introduced changes really solved the issue with the conflicting operations. The test with conflicting update and punch ((here)[https://github.com/daos-stack/daos/pull/13495/commits/7592898e87746d80f985df7672a557f7560f5b9a#diff-915f0cd50d1f423d5f44bcc9b8ad81d0e6589df2f932b1cd26f696fe179fdd2bR437]) is currently failing.

From what I have learned ILOG does NOT consider the minor epoch as long as it does not consider the conflicting operations as belonging to the same transaction. Since the local transaction as of now does not track ILOG modifications belonging to a transaction all ILOG modifications are immediately considered `DTX_LID_COMMITTED`. Hence, each operation is considered to belong to different transactions and returns with `DER_TX_RESTART` on conflict. No minor epochs are even considered.

> We speculate the difficult you face here may be that rdb does not increment dth_op_seq as obj does (via dtx_sub_init).

No. This won't solve the problem.

I consulted @jolivier23 and if I cite him correctly, the only way out of this conflict is by introducing a special case just for local transactions to ignore this conflict.

liw: @Nasf-Fan, sounds like our understanding was wrong.

@janekmi, would "every VOS local TX must have valid TX ID" solve the problem?

janekmi: > would "every VOS local TX must have valid TX ID" solve the problem?

Sadly no. To track modifications to ILOGs done by the transaction would require adding to the local transaction a DTX entry as a regular DTX transaction has. The local ID of the DTX entry is what allows to differentiate between transactions. Note it would increase the complexity of a local transaction since all ILOG records would have to be updated as committed when the transaction concludes.

Nasf-Fan: @janekmi , would you please to apply the following patch for such TX_RESTAR issue?

diff --git a/src/vos/vos_ilog.c b/src/vos/vos_ilog.c
index 065f6ed816..1e06e3452b 100644
--- a/src/vos/vos_ilog.c
+++ b/src/vos/vos_ilog.c
@@ -57,7 +57,7 @@ vos_ilog_is_same_tx(struct umem_instance *umm, uint32_t tx_id,
                 * transactional, treat it as the same transaction and let the
                 * minor epoch handle any conflicts.
                 */
-               if (!dtx_is_valid_handle(dth))
+               if (!dtx_is_real_handle(dth))
                        *same = true;
        } else if (tx_id == dtx) {
                *same = true;

janekmi: @Nasf-Fan:

1. Are we 100% sure RDB is ok with ignoring this kind of conflict? @liw 
2. I believe, this solution is not fully correct. Since it allows comparing minor epochs between transactions whereas the minor epoch scope seems to be correct only in the transaction scope. Am I missing something?

Nasf-Fan: It is the same result as without your patch, right?
If with your patch, the local transaction with the same major epoch, should be regarded as belong to the same RDB transaction, it is expected, right? The left things will be handled via minor epoch subsequently, please check update_inplace() logic.

I cannot say it is 100% correct, but please verify whether it can resolve your current test failure. If it did not work, then we will consider other solution.

janekmi: > 2. I believe, this solution is not fully correct. Since it allows comparing minor epochs between transactions whereas the minor epoch seems to be valid only in the transaction scope. Am I missing something?

I got a clarification from @jolivier23. Minor epochs can be compared between transactions since each of the transactions is expected to use a different epoch. Assuming it is ok with RDB works (@liw?) your solution @Nasf-Fan is ok. 👍

Nasf-Fan: As my understand, for a RDB transaction, it may contain several sub-modifications, each of them will have different minor epochs but with the same major epoch. That is basically the same as distributed transaction does.

liw: > As my understand, for a RDB transaction, it may contain several sub-modifications, each of them will have different minor epochs but with the same major epoch. That is basically the same as distributed transaction does.

Sorry, @Nasf-Fan, I forgot about the metadata container (MC), although the conclusion should be the same. While what you said is true for the log container (LC), all writes (from all `vos` TXs) to the MC use the hard-coded epoch: 1. But all these writes are updates; there's no punches. So sounds like we should be okay.

Nasf-Fan: > > As my understand, for a RDB transaction, it may contain several sub-modifications, each of them will have different minor epochs but with the same major epoch. That is basically the same as distributed transaction does.
> 
> Sorry, @Nasf-Fan, I forgot about the metadata container (MC), although the conclusion should be the same. While what you said is true for the log container (LC), all writes (from all `vos` TXs) to the MC use the hard-coded epoch: 1. But all these writes are updates; there's no punches. So sounds like we should be okay.

"hard coded epoch: 1" means for different RDB transactions? If yes, how to distinguish two RDB transactions?

liw: All writes to the MC use hard-coded epoch 1---all such writes, from all `rdb` TXs, in all DAOS systems.

liw: I don't understand the question "how to distinguish two RDB transactions?". Distinguish for what purposes?

Nasf-Fan: Then there will be single RDB transaction? All RDB updates belong to the unique RDB transaction?

liw: In my current plan, `rdb` will perform each of its `vos` TXs like this:

```
dtx_begin(..., &dth);

/* Modify the LC. */
vos_obj_update(LC object, ..., epoch=X, ..., &dth);
vos_obj_punch(LC object, ..., epoch=X, &dth);
...

/* Modify the MC (to record X). */
vos_obj_update(MC object, ..., epoch=1, ..., &dth);

dtx_end(&dth, ...);
```

The set of sub-modification to the LC is determined by `rdb` users. The MC modification is always one update to the same <object, dkey, akey> and always in epoch 1.

Nasf-Fan: > In my current plan, `rdb` will perform each of its `vos` TXs like this:
> 
> ```
> dtx_begin(..., &dth);
> 
> /* Modify the LC. */
> vos_obj_update(LC object, ..., epoch=X, ..., &dth);
> vos_obj_punch(LC object, ..., epoch=X, &dth);
> ...
> 
> /* Modify the MC (to record X). */
> vos_obj_update(MC object, ..., epoch=1, ..., &dth);
> 
> dtx_end(&dth, ...);
> ```
> 
> The set of sub-modification to the LC is determined by `rdb` users. The MC modification is always one update to the same <object, dkey, akey> and always in epoch 1.

Then why not use the same epoch "X" for MC sub-modification? In your current plan, the sub-modifications in one RDB transaction will have different major epochs. That breaks current VOS TX assumption: the sub-modifications belong to the same transaction will use the same major epoch. In your case, the destination for LC and MC sub-modifications are different, that can work. But consider next RDB transaction, its MC sub-modification will also use the fixed major epoch 1, then ilog logic will regard it as belong to former RDB TX, that is confused, may cause trouble, need to be verified.

liw: Because "always using epoch 1 for MC writes" is what `rdb` has been doing since it started to use `vos`. It is part of `rdb`'s storage layout definition. If possible, I don't want to make any changes to that at this time.

liw: > That breaks current VOS TX assumption: the sub-modifications belong to the same transaction will use the same major epoch.

IIRC, that is a DTX assumption, not a VOS local TX assumption. For instance, a umem/pmem TX has no "epoch" or "version" concept in its API.

Nasf-Fan: > Because "always using epoch 1 for MC writes" is what `rdb` has been doing since it started to use `vos`. It is part of `rdb`'s storage layout definition. If possible, I don't want to make any changes to that at this time.

I cannot understand how ilog logic can distinguish different RDB transactions to use the same epoch and handle them as different transactions. Is it possible the subsequent updates to be skipped silently because of the "same" check? not sure, please verify. But I do not think that is a strong reason to have to use fixed epoch. What is the trouble if MC write uses the same epoch X as LC write does? I may miss some important factor.


> > That breaks current VOS TX assumption: the sub-modifications belong to the same transaction will use the same major epoch.
> 
> IIUC, that is a DTX assumption, not a VOS local TX assumption. For instance, a umem/pmem TX has no "epoch" or "version" concept in its API.

It is current ilog logic determined, not DTX assumption. These part changes do not touch non-local DTX logic. Please check the logic check_equal(), it firstly checks the epoch of the modification. From ilog logic perspective, major epoch is the basic clew to distinguish the transactions unless we (define and) offer more identifier. But I do not think that will be more simpler than flexible epoch for different transactions.
As my understand, VOS is version based, epoch is the version, it should be at least locally distinguishable.

liw: > Is it possible the subsequent updates to be skipped silently because of the "same" check? not sure, please verify.

I have verified that the update at the hard-coded epoch 1 works:

```
01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_internal.h:394 rdb_mc_lookup() mc=7fb9d5615650 oid=1 akey=<0x7fb9e77fd9f5, 3> value=<0x7fb9db624270, 304, 304>
01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_raft.c:1211 rdb_raft_log_offer_single() b276ec5f[0]: before update: tail=2
01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_internal.h:382 rdb_mc_update() mc=7fb9d5615650 oid=1 n=1 akeys[0]=<0x7fb9e77fd9f5, 3> values[0]=<0x7fb9d5875280, 304>
01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_internal.h:394 rdb_mc_lookup() mc=7fb9d5615650 oid=1 akey=<0x7fb9e77fd9f5, 3> value=<0x7fb9db624270, 304, 304>
```

An `rdb_mc_lookup` call was added before and after an `rdb_mc_update` call, to read the value before and after the update. All the two `lookup` calls and the `update` call used hard-coded epoch 1.

> But I do not think that is a strong reason to have to use fixed epoch. What is the trouble if MC write uses the same epoch X as LC write does? I may miss some important factor.

The important thing is that "using hard-coded epoch 1" is part of the `rdb` storage layout definition. Changing it would introduce compatibility concerns. Secondly, while I'm okay with using monotonically increasing epochs for MC writes, I'm against using _the same epoch_ as the LC writes, not because of any problem, but because of the confusing model.

Nasf-Fan: > > Is it possible the subsequent updates to be skipped silently because of the "same" check? not sure, please verify.
> 
> I have verified that the update at the hard-coded epoch 1 works:
> 
> ```
> 01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_internal.h:394 rdb_mc_lookup() mc=7fb9d5615650 oid=1 akey=<0x7fb9e77fd9f5, 3> value=<0x7fb9db624270, 304, 304>
> 01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_raft.c:1211 rdb_raft_log_offer_single() b276ec5f[0]: before update: tail=2
> 01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_internal.h:382 rdb_mc_update() mc=7fb9d5615650 oid=1 n=1 akeys[0]=<0x7fb9e77fd9f5, 3> values[0]=<0x7fb9d5875280, 304>
> 01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_internal.h:394 rdb_mc_lookup() mc=7fb9d5615650 oid=1 akey=<0x7fb9e77fd9f5, 3> value=<0x7fb9db624270, 304, 304>
> 01/29-05:55:13.30 boro-66 DAOS[487731/0/1959] rdb  DBUG src/rdb/rdb_raft.c:1228 rdb_raft_log_offer_single() b276ec5f[0]: after update: tail=3
> ```
> 

According to current master logic, if two RDB transactions use the same fixed (major) epoch 1, then ilog logic will only do the first update, and commit. When the second RDB transaction arrives, it changes nothing for ilog (obj/key), and only updates the value (SV). But I am wondering how the following logic can be passed:

```
static int
svt_rec_update(struct btr_instance *tins, struct btr_record *rec,
                d_iov_t *key_iov, d_iov_t *val_iov, d_iov_t *val_out)
{
        struct vos_svt_key      *skey;
        struct vos_irec_df      *irec;
        struct vos_rec_bundle   *rbund;
        int                      rc;

        rbund = iov2rec_bundle(val_iov);
        skey = (struct vos_svt_key *)key_iov->iov_buf;
        irec = vos_rec2irec(tins, rec);

        /** Disallow same epoch overwrite */
        if (skey->sk_minor_epc <= irec->ir_minor_epc)
                return -DER_NO_PERM;
```

Unless each LC update use different object/key. My understand the RDB transactions uses single value update.

liw: Could the following lines from `btr_update` provide an answer?

```
	rc = btr_rec_update(tcx, rec, key, val, val_out);
	if (rc == -DER_NO_PERM) { /* cannot make inplace change */
		struct btr_trace *trace = &tcx->tc_trace.ti_trace[tcx->tc_depth - 1];

		if (btr_has_tx(tcx)) {
			rc = btr_node_tx_add(tcx, trace->tr_node);
			if (rc != 0)
				goto out;
		}

		D_DEBUG(DB_TRACE, "Replace the original record\n");
		rc = btr_rec_free(tcx, rec, NULL);
		if (rc)
			goto out;
		rc = btr_rec_alloc(tcx, key, val, rec, val_out);
	}
```

Nasf-Fan: Yes, it seems so. Based on current ilog implementation, it will not forbid two RDB transactions to use the same fixed epoch 1. With @janekmi 's patch, if the subsequent RDB transaction has more sub-modifications than former ones, it may generate more ilog entries or related object/key, not sure whether it is RDB expected or not. The other parts are basically fine for me.

liw: With this PR, it *seems* ilog entries were not generated for repeated updates in the MC during my testing. (They were generated for LC writes, but that was expected, I think.)

Nasf-Fan (dismissed):

daltonbohning (commented):
ftest LGTM

liw (dismissed):
Thanks, Jan. Looks good to me. Should still seek reviews from @jolivier23 and @Nasf-Fan, I think.

Nasf-Fan (dismissed):

jolivier23 (dismissed):
Looks like it conflicts with master branch now so you'll need to merge

jolivier23:
I think Features: daosio is pretty minimal testing.  Might be good to at least add something like object and tx

See full list here: 
https://daosio.atlassian.net/wiki/spaces/DC/pages/10984259629/Test+Tags#[inlineExtension]List-of-Feature-Tags

jolivier23 (dismissed):

liw (dismissed):

Nasf-Fan (dismissed):

liw (dismissed):

Nasf-Fan (dismissed):

Nasf-Fan (approved):

liw (approved):

gnailzenh (approved):

PR #6185, opened 2021-07-14
DAOS client side metrics API specification

sherintg:
Early Draft for review  Only

Changes created based on the initial discussion with @mchaarawi

PR #13517, opened 2023-12-17
DAOS-8331 client: add client side metrics

wangdi1:
1. Move TLS to common, so both client and server can have TLS, which metrics can be attached metrics on it.

2. Add object metrics on the client side, enabled by export DAOS_CLIENT_METRICS=1. And client metrics are organized as "root/jobid/pid/xxxxx"

And root/jobid/pid are stored in an independent share memory, which will only be destoryed if all jobs are destroyed.

During each daos thread initialization, it will created another shmem (pid/xxx), which all metrics of the thread will be attached to. And this metric will be destoryed once the thread exit.

3. Add DAOS_METRIC_DUMP_ENV dump metrics from current thread once it exit.

4. Some fixes in telemetrics about conv_ptr during re-open the share memory.

5. Add daos_metrics --jobid XXX options to retrieve all metrics of the job.

Required-githooks: true

### Gatekeeper:


ashleypittman:
This is failing with a segfault in pydaos, as a start you could disable this test by commenting out the two lines at https://github.com/daos-stack/daos/blob/master/utils/node_local_test.py#L6031-L6032 which would allow the rest of the testing to proceed which may give some indication of the cause.

wangdi1:
> This is failing with a segfault in pydaos, as a start you could disable this test by commenting out the two lines at https://github.com/daos-stack/daos/blob/master/utils/node_local_test.py#L6031-L6032 which would allow the rest of the testing to proceed which may give some indication of the cause.

Thanks! @ashleypittman

src/utils/daos_metrics/daos_metrics.c
```diff
@@ -199,27 +267,50 @@ main(int argc, char **argv)
...
+			printf("Unable to find job %s\n", jobid);
+			exit(0);
+		}
+
+		iter_arg.dirname = dirname;
+		iter_arg.filter = filter;
+		iter_arg.iter_cb = iter_cb;
+		iter_arg.delay = delay;
+		iter_arg.num_iter = num_iter;
+		d_tm_iterate(ctx, root, 0, D_TM_DIRECTORY, NULL, format, extra_descriptors,
```
mjmac, line 298: I played around with the PR on my development node where I'm running the server and some client processes. When I run daos_metrics -j <jobid>, I get the engine metrics mixed in and the client metrics are all zero. I wonder if it's because the wrong context is being used in the iterator?

wangdi1: did you provide correct jobid after -j ?  hmm, the iteration will iterate all pid under root/jobid, then iterate each pid metrics here. see d_tm_iterate->iter_per_pid->process_metrics().  It should only include client object metrics for the moment. 

I tried this on my side with multiple jobid, it works here.

src/utils/daos_metrics/daos_metrics.c
```diff
@@ -150,7 +239,10 @@ main(int argc, char **argv)
 			delay = atoi(optarg);
 			break;
 		case 'e':
-			ops |= D_TM_ITER_RESET;
+			iter_cb = iter_reset;
+			break;
+		case 'j':
```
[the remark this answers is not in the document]
wangdi1: Sure

src/client/api/metrics.c
```diff
@@ -0,0 +1,167 @@
...
+dc_tm_init(void)
+{
+	struct d_tm_node_t *job_node;
+	struct d_tm_context *current_ctx;
+	struct daos_thread_local_storage *dtls;
+	int		    metrics_tag;
+	pid_t		    pid;
+	int		    rc;
+
+	d_getenv_bool(DAOS_CLIENT_METRICS_ENV, &client_metric);
```
jolivier23, line 36: would it be useful if this was controlled by the agent or the server instead of an environment variable?

wangdi1: Indeed, though I believe @mjmac  will move the client metrics root to agent  based on this PR.

mjmac: > would it be useful if this was controlled by the agent or the server instead of an environment variable?

Hmm. This is an interesting idea. Certainly we could add some control via the GetAttachInfo upcall so that client applications have telemetry enabled/disabled according to an agent or server config setting. I think we can add that later though, as another option in addition to this environment variable.

src/client/api/metrics.c
```diff
@@ -0,0 +1,167 @@
...
+#include <daos/common.h>
+#include <daos/job.h>
+#include <daos/tls.h>
+#include <gurt/telemetry_common.h>
+#include <gurt/telemetry_consumer.h>
+#include <gurt/telemetry_producer.h>
+
+#define INIT_JOB_NUM		1024
+bool client_metric;
+bool client_metric_retain;
```
ashleypittman, line 20: These (and many others) should be static if possible or renamed with a daos_ prefix if not.  There are much stricter naming requirements for symbols in client libraries than there are for the server.

src/client/api/metrics.c
```diff
@@ -0,0 +1,167 @@
...
+
+	metrics_tag = D_TM_CLIENT_PROCESS | D_TM_OPEN_OR_CREATE;
+	if (client_metric_retain)
+		metrics_tag |= D_TM_RETAIN_SHMEM;
+	else
+		metrics_tag |= D_TM_RETAIN_SHMEM_IF_NON_EMPTY;
+
+	rc = d_tm_init(DC_TM_JOB_ROOT_ID, MAX_IDS_SIZE(INIT_JOB_NUM), metrics_tag);
+	if (rc != 0) {
+		D_ERROR("init job root id %u: %d\n", DC_TM_JOB_ROOT_ID, rc);
```
ashleypittman, line 50: I won't comment on all of these, but this one in particular looks like we should log the failure properly.
```suggestion
		DL_ERROR(rc, "init job root id %u", DC_TM_JOB_ROOT_ID);
```

src/client/api/metrics.c
```diff
@@ -0,0 +1,167 @@
...
+		return rc;
+	}
+
+	pid = getpid();
+	D_INFO("INIT %s/%u metrics\n", dc_jobid, pid);
+	rc = d_tm_add_metric(&job_node, D_TM_DIRECTORY,
+			     "job id directory", "dir",
+			     "%s/%u", dc_jobid, pid);
+	/* Close job root sheme */
```
ashleypittman, line 59: These codespell annotations all need to be fixed.

src/common/tls.c
```diff
@@ -0,0 +1,191 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause-Patent
+ */
+/**
+ * It implements thread-local storage (TLS) for DAOS.
+ */
+#include <pthread.h>
+#include <daos/tls.h>
+
+/* The array remember all of registered module keys on one node. */
+struct daos_module_key *daos_module_keys[DAOS_MODULE_KEYS_NR] = { NULL };
```
ashleypittman, line 13: Should be static again and you don't need to initliase it.
```suggestion
static struct daos_module_key *daos_module_keys[DAOS_MODULE_KEYS_NR];
```

src/common/tls.c
```diff
@@ -0,0 +1,191 @@
...
+	D_ASSERT(key->dmk_index < DAOS_MODULE_KEYS_NR);
+	D_MUTEX_LOCK(&daos_module_keys_lock);
+	daos_module_keys[key->dmk_index] = NULL;
+	D_MUTEX_UNLOCK(&daos_module_keys_lock);
+}
+
+/**
+ * Init thread context
+ *
+ * \param[in]dtls	Init the thread context to allocate the
```
ashleypittman, line 48: I don't think doxygen will like this, you need to adjust the whitespace.  Applies elsewhere as well.
```suggestion
 * \param[in] dtls	Init the thread context to allocate the
```

src/common/tls.c
```diff
@@ -0,0 +1,191 @@
...
+static int
+daos_thread_local_storage_init(struct daos_thread_local_storage *dtls,
+			       int xs_id, int tgt_id)
+{
+	int rc = 0;
+	int i;
+
+	if (dtls->dtls_values == NULL) {
+		D_ALLOC_ARRAY(dtls->dtls_values,
+			      (int)ARRAY_SIZE(daos_module_keys));
```
ashleypittman, line 63: This looks weird, would this change improve readability?
```suggestion
		D_ALLOC_ARRAY(dtls->dtls_values, DAOS_MODULE_KEYS_NR);
```

src/common/tls.c
```diff
@@ -0,0 +1,191 @@
...
+				dmk->dmk_fini(dtls->dtls_tag, dtls->dtls_values[i]);
+			}
+		}
+	}
+
+	D_FREE(dtls->dtls_values);
+}
+
+pthread_key_t dss_tls_key;
+pthread_key_t dc_tls_key;
```
ashleypittman, line 110: Again, these should be static if possible.

src/gurt/telemetry.c
```diff
@@ -837,14 +857,24 @@ d_tm_init(int id, uint64_t mem_size, int flags)
...
+
+	if (tm_shmem.retain_non_empty) {
+		struct d_tm_node_t *root;
+
+		root = d_tm_get_root(tm_shmem.ctx);
+		if (root->dtn_child != NULL)
+			destroy_shmem = false;
+	}
 
+	D_INFO("Delete share memory for id %u destory %s\n", tm_shmem.id,
```
ashleypittman, line 876: There's a DF_BOOL macro to use here rather than yes/no trigraph.

src/include/daos/metric.h
ashleypittman, line 1: We don't require all our code to be well formatted with clang-format yet but for new public header files we really should.  I can push an update to this PR for just there new files if you would like, alternatively it should just be a case of "clang-format -i <filename>".

src/include/daos/tls.h
ashleypittman, line 1: Same comment as previous file.

If these are new public header files do they need adding to a sconstruct file somewhere so that they get included in the RPM?

src/include/daos/tls.h
```diff
@@ -0,0 +1,124 @@
...
+{
+	return (struct daos_thread_local_storage *)
+		pthread_getspecific(dc_tls_key);
+}
+
+/* For now TLS is only enabled if metrics are enabled */
+#define METRIC_DUMP_ENV "DAOS_METRIC_DUMP_ENV"
+#define DAOS_CLIENT_METRICS_ENV	"DAOS_CLIENT_METRICS"
+#define DAOS_CLIENT_METRICS_RETAIN_ENV	"DAOS_CLIENT_METRICS_RETAIN"
+extern bool client_metric;
```
ashleypittman, line 83: You can't have a variable called "client_metric" in a shared library that you're linking with end-user applications.

src/include/daos/tls.h
```diff
@@ -0,0 +1,124 @@
...
+	return dtls->dtls_values[key->dmk_index];
+}
+
+#define dss_module_key_get daos_module_key_get
+#define dss_register_key   daos_register_key
+#define dss_unregister_key daos_unregister_key
+#define dss_module_info    daos_module_info
+#define dss_module_tag    daos_module_tag
+#define dss_module_key    daos_module_key
+#define dss_thread_local_storage daos_thread_local_storage
```
ashleypittman, line 115: Are these appropriate for a public interface or should they be private?

ashleypittman (changes requested):
-1 for the namespace pollution.

If this is creating shared memory on the client then we need some documentation or user education on how to manage this and detect/cleanup and leaked resources - even if that is just "the agent does it".

src/common/tls.c
wangdi1: sure

src/common/tls.c
wangdi1: sure

src/common/tls.c
wangdi1: sure

src/client/api/metrics.c
wangdi1: sure

src/client/api/metrics.c
wangdi1: sure

src/client/api/metrics.c
wangdi1: sure

src/client/api/metrics.c
```diff
@@ -0,0 +1,167 @@
...
+	metrics_tag = D_TM_CLIENT_PROCESS;
+	if (client_metric_retain)
+		metrics_tag |= D_TM_RETAIN_SHMEM;
+	rc = d_tm_init(pid, MAX_IDS_SIZE(INIT_JOB_NUM), metrics_tag);
+	if (rc != 0)
+		D_GOTO(out, rc);
+
+	current_ctx = d_tm_open(pid);
+	if (current_ctx == NULL)
+		D_GOTO(out, rc = -DER_NOMEM);
```
mjmac, line 75: I am not 100% sure, but I'm pretty sure that this is the wrong way to do this. Instead of manually managing the shm segments, I think it would be better to use [d_tm_add_ephemeral_dir()](https://github.com/daos-stack/daos/blob/master/src/gurt/telemetry.c#L2455) to attach the child segments to the parent. This will make walking down from the root work better.

@kjacque added this feature to the telemetry for dynamic pool support a few years ago; she may have more thoughts about how to organize this.

wangdi1: Ah, I did not see there is d_tm_add_ephemeral_dir(). hmm, this looks similar as this function. Yes, indeed, I should use this function. let me refresh the patch. Thanks!

src/include/daos/tls.h
```diff
@@ -0,0 +1,124 @@
...
+
+static inline struct daos_thread_local_storage *
+dc_tls_get()
+{
+	return (struct daos_thread_local_storage *)
+		pthread_getspecific(dc_tls_key);
+}
+
+/* For now TLS is only enabled if metrics are enabled */
+#define METRIC_DUMP_ENV "DAOS_METRIC_DUMP_ENV"
```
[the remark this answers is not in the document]
wangdi1: yeah, sure.

src/include/gurt/telemetry_common.h
```diff
@@ -155,6 +155,9 @@ enum {
 	D_TM_SERVER_PROCESS		= 0x000,
 	D_TM_SERIALIZATION		= 0x001,
 	D_TM_RETAIN_SHMEM		= 0x002,
+	D_TM_RETAIN_SHMEM_IF_NON_EMPTY	= 0x004,
+	D_TM_CLIENT_PROCESS		= 0x008,
```
kjacque, line 159: I'm wondering why the metrics library needs to know it's a client process at all.

src/include/gurt/telemetry_consumer.h
```diff
@@ -49,12 +49,21 @@ int d_tm_list(struct d_tm_context *ctx, struct d_tm_nodeList_t **head,
...
 void d_tm_iterate(struct d_tm_context *ctx, struct d_tm_node_t *node,
 		  int level, int filter, char *path, int format,
-		  int opt_fields, uint32_t ops, FILE *stream);
+		  int opt_fields, d_tm_iter_cb_t iter_cb, void *cb_arg);
 void d_tm_print_node(struct d_tm_context *ctx, struct d_tm_node_t *node,
 		     int level, char *name, int format, int opt_fields,
 		     FILE *stream);
+
+void d_tm_reset_node(struct d_tm_context *ctx, struct d_tm_node_t *node, int level,
+		     char *path, int format, int opt_fields, FILE *stream);
```
kjacque, line 65: This shouldn't be available to the telemetry consumer, as this alters the data. If it needs to be accessible publicly, it belongs in telemetry_producer.h. It may not need to be public at all if you use ephemeral directories to handle the job IDs and pids.

src/utils/daos_metrics/daos_metrics.c
```diff
@@ -53,29 +53,120 @@ print_usage(const char *prog_name)
...
+{
+	d_tm_print_node(ctx, node, level, path, format, opt_fields, (FILE *)arg);
+}
+
+static void
+iter_reset(struct d_tm_context *ctx, struct d_tm_node_t *node, int level,
+	   char *path, int format, int opt_fields, void *arg)
+{
+	d_tm_reset_node(ctx, node, level, path, format, opt_fields, (FILE *)arg);
+}
```
kjacque, line 122: I'm not sure about letting people do this with this tool. Things could get very ugly if another process is actively writing.

kjacque (changes requested):

src/common/tls.c
wangdi1: sure

src/client/api/metrics.c
```diff
@@ -0,0 +1,167 @@
...
+	if (!client_metric)
+		return 0;
+
+	d_getenv_bool(DAOS_CLIENT_METRICS_RETAIN_ENV, &client_metric_retain);
+
+	metrics_tag = D_TM_CLIENT_PROCESS | D_TM_OPEN_OR_CREATE;
+	if (client_metric_retain)
+		metrics_tag |= D_TM_RETAIN_SHMEM;
+	else
+		metrics_tag |= D_TM_RETAIN_SHMEM_IF_NON_EMPTY;
```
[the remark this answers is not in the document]
wangdi1: client metrics will only retain with DAOS_CLIENT_METRICS_RETAIN, otherwise it will be delete once all threads exists, which is indicated by D_TM_RETAIN_SHMEM_IF_NOT_EMPTY, though I assume there will be extra options when moving this shmem to agent.

src/engine/srv.c
```diff
@@ -383,7 +383,7 @@ static void
 dss_srv_handler(void *arg)
 {
 	struct dss_xstream		*dx = (struct dss_xstream *)arg;
-	struct dss_thread_local_storage	*dtc;
+	struct daos_thread_local_storage	*dtc;
 	struct dss_module_info		*dmi;
 	int				 rc;
```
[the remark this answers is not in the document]
wangdi1: sure

src/gurt/telemetry.c
```diff
@@ -529,7 +531,7 @@ init_node(struct d_tm_shmem_hdr *shmem, struct d_tm_node_t *node,
 		D_ERROR("cannot allocate node name [%s]\n", name);
 		return -DER_NO_SHMEM;
 	}
-	strncpy(node->dtn_name, name, buff_len);
+	strncpy(conv_ptr(shmem, node->dtn_name), name, buff_len);
```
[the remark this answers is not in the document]
wangdi1: the client side metrics will be organized as  root/jobid/pid/tid/xxx,  and each jobid/pid will create its own share memory by d_tm_add_ephemeral_dir(). so only root/jobid/pid will be written from multiple writer, so I only try to modify the related part to support multiple writer.

src/gurt/telemetry.c
```diff
@@ -2106,6 +2129,115 @@ is_initialized(void)
 	       tm_shmem.ctx->shmem_root != NULL;
 }
 
+static struct d_tm_node_t*
+delete_child(struct d_tm_context *ctx, struct d_tm_node_t *parent, const char *name)
```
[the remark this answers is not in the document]
wangdi1: I will use d_tm_add_ephemeral_dir(), so this function can be removed.

src/include/gurt/telemetry_common.h
```diff
@@ -155,6 +155,9 @@ enum {
 	D_TM_SERVER_PROCESS		= 0x000,
 	D_TM_SERIALIZATION		= 0x001,
 	D_TM_RETAIN_SHMEM		= 0x002,
+	D_TM_RETAIN_SHMEM_IF_NON_EMPTY	= 0x004,
```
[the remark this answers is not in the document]
wangdi1: oh, this tag is only for deleting the job metrics by default.  so if D_TM_RETAIN_SHMEM is set, then job metrics will be kept even after the job exit.   Otherwise if D_TM_RETAIN_SHMEM_IF_NON_EMPTY is set, then job metrics will be delete if no threads/processes left.  Though once these share memory is attached to agent, there maybe more options here.

src/client/api/metrics.c
```diff
@@ -0,0 +1,167 @@
...
+	d_tm_fini();
+
+	if (client_metric_retain)
+		return;
+
+	rc = d_tm_init(DC_TM_JOB_ROOT_ID, MAX_IDS_SIZE(INIT_JOB_NUM),
+		       D_TM_CLIENT_PROCESS | D_TM_RETAIN_SHMEM_IF_NON_EMPTY |
+		       D_TM_OPEN_OR_CREATE);
+	if (rc != 0)
+		return;
```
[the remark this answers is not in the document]
wangdi1: oh, there will not be needed if I use d_tm_add_ephemeral_dir

src/include/gurt/telemetry_common.h
wangdi1: oh, yes, this can be removed. thanks.

src/include/gurt/telemetry_consumer.h
wangdi1: sure

src/utils/daos_metrics/daos_metrics.c
wangdi1: hmm, the user needs to be able to reset the value.

src/include/gurt/telemetry_consumer.h
wangdi1: Hmm, I thought this maybe a consumer, since the user will use it to reset the value from daos_metrics utils? no?

src/include/daos/tls.h
wangdi1: This patch move TLS to both client and server side, so some structures are renamed from dss to daos, and these definition is for avoiding changed too much name on each name. Hmm, I am not sure where to put these, if not with the new TLS definition.

src/include/daos/tls.h
wangdi1: sure

src/gurt/telemetry.c
wangdi1: sure

src/include/daos/tls.h
wangdi1: tls.h is not needed to export to the user.  still need adding to scontract file?

ashleypittman: In that case it probably shouldn't be in this directory.

src/object/cli_mod.c
```diff
@@ -12,21 +12,110 @@
...
+	D_ALLOC_PTR(tls);
+	if (tls == NULL)
+		return NULL;
+
+	/** register different per-opcode sensors */
+	for (opc = 0; opc < OBJ_PROTO_CLI_COUNT; opc++) {
+		/** Start with number of active requests, of type gauge */
+		rc = d_tm_add_metric(&tls->cot_op_active[opc], D_TM_STATS_GAUGE,
+				     "number of active object RPCs", "ops",
+				     "%s/%u/%lu/ops/%s/active", dc_jobid, pid, tid,
```
mjmac, line 43: Please put these ops metrics under `io/`, for consistency with the server-side metric paths. This makes it easier to reuse code for the Prometheus exporter.
```suggestion
				     "%s/%u/%lu/io/ops/%s/active", dc_jobid, pid, tid,
```

src/object/cli_mod.c
```diff
@@ -12,21 +12,110 @@
...
+		if (opc == DAOS_OBJ_RPC_UPDATE ||
+		    opc == DAOS_OBJ_RPC_TGT_UPDATE ||
+		    opc == DAOS_OBJ_RPC_FETCH)
+			/** See below, latency reported per size for those */
+			continue;
+
+		/** And finally the per-opcode latency, of type gauge */
+		rc = d_tm_add_metric(&tls->cot_op_lat[opc], D_TM_STATS_GAUGE,
+				     "object RPC processing time", "us",
+				     "%s/%u/%lu/ops/%s/latency", dc_jobid, pid, tid,
```
mjmac, line 59: Also here:
```suggestion
				     "%s/%u/%lu/io/ops/%s/latency", dc_jobid, pid, tid,
```

mjmac (commented):
This works great! I have a draft PR (#13545) that builds on this work to enable Prometheus metrics exported from daos_agent. Just need a small change in the metrics path for consistency with the server.

Should we add metrics for fetch_bytes and update_bytes, similar to the server?

wangdi1:
> This works great! I have a draft PR (#13545) that builds on this work to enable Prometheus metrics exported from daos_agent. Just need a small change in the metrics path for consistency with the server.
> 
> Should we add metrics for fetch_bytes and update_bytes, similar to the server?

Thanks!  Sure, let me add fetch_bytes and update_bytes.

src/object/cli_mod.c
wangdi1: sure

src/object/cli_mod.c
wangdi1: sure

src/client/api/metrics.c
```diff
@@ -0,0 +1,146 @@
...
+	if (rc != 0) {
+		DL_ERROR(rc, "init job root id.");
+		return rc;
+	}
+
+	pid = getpid();
+	D_INFO("INIT %s/%u metrics\n", dc_jobid, pid);
+
+	/** create new shmem space for per-pool metrics */
+	rc = d_tm_add_ephemeral_dir(NULL, MAX_IDS_SIZE(INIT_JOB_NUM), "%s/%u",
```
mjmac, line 58: Was just chatting with @kjacque about this... I think we need the per-job ephemeral dir, but we also need per-pid ephemeral dirs under it. This would allow the agent to prune old per-pid telemetry without disturbing other processes that are using the same jobid. I am working on that part of it now.

I think we also need to adjust the size of these segments, something along these lines:
  - client root: D_TM_METRIC_SIZE * NUM_SUPPORTED_JOBIDS (256 ?)
  - per-jobid: D_TM_METRIC_SIZE * NUM_SUPPORTED_JOB_PROCS (1024 ?)
  - per-proc: D_TM_METRIC_SIZE * NUM_CLIENT_METRICS (not sure how to calc this)
  
  We may want to make those configurable someday, but I don't think we need to worry about it now.

mjmac: I think what we have now is fine to start. I can play around with any changes to make pruning work better in my follow-on patch.

wangdi1: Trying to add per pool metrics on client side.

wangdi1: @mjmac  Just added per pool metrics for fetch/update bytes, please let me know if you need further improvements.

mjmac: @wangdi1 I played around with this last night... I think there is a bug in the per-pool metrics initialization if a label is used. I had to move the call to dc_pool_metrics_start() until after the label is resolved into a UUID, otherwise the pool UUID is always the nil UUID (0000). Not sure if this is the best place to put it, but it works in my testing:

```
@@ -651,10 +653,6 @@ init_pool(const char *label, uuid_t uuid, uint64_t capas, const char *grp,
        if (rc != 0)
                D_GOTO(err_pool, rc);

-       rc = dc_pool_metrics_start(pool);
-       if (rc != 0)
-               D_GOTO(err_pool, rc);
-
        /** Agent configuration data from pool->dp_sys->sy_info */
        /** sy_info.provider */
        /** sy_info.interface */
@@ -699,6 +697,11 @@ dc_pool_connect_internal(tse_task_t *task, daos_pool_info_t *info, const char *l
	rc = dc_pool_choose_svc_rank(label, pool->dp_pool, &pool->dp_client, &pool->dp_client_lock,
				     pool->dp_sys, &ep);
	if (rc != 0) {
		DL_CDEBUG(rc == -DER_NONEXIST, DB_PL, DLOG_ERR, rc,
			  DF_UUID ":%s: cannot find pool service", DP_UUID(pool->dp_pool),
			  label ? label : "");
                goto out;
        }

+       /** Start the pool metrics now that we are guaranteed to have a UUID */
+       rc = dc_pool_metrics_start(pool);
+       if (rc != 0)
+               D_GOTO(out, rc);
+
```

I also was hitting a segfault on dc_tm_fini() when metrics aren't retained and the shm is cleaned up. I made the following change and it fixed the segfault:
```
--- a/src/gurt/telemetry.c
+++ b/src/gurt/telemetry.c
@@ -2405,14 +2405,18 @@ failure:
 }

 static void
-invalidate_link_node(struct d_tm_node_t *node)
+invalidate_link_node(struct d_tm_shmem_hdr *parent, struct d_tm_node_t *node)
 {
+       struct d_tm_metric_t *link_metric;
+
        if (node == NULL || node->dtn_type != D_TM_LINK)
                return;

        node->dtn_name = NULL;
-       if (node->dtn_metric != NULL)
-               node->dtn_metric->dtm_data.value = 0;
+       if (node->dtn_metric != NULL) {
+               link_metric = conv_ptr(parent, node->dtn_metric);
+               link_metric->dtm_data.value = 0;
+       }
```

I don't fully understand all of the conv_ptr() changes... It seems very tricky to remember which pointers can be accessed directly and which need to be converted for the shm offset, I guess.

wangdi1: Thanks! @mjmac  Sorry, I did not try label case. Yes, you fix seems good to me. 

Hmm, it seems some share memory is not being freed if the metrics are not retained. let me update the patch. Thanks

src/object/obj_utils.c
```diff
@@ -86,6 +90,157 @@ daos_iods_free(daos_iod_t *iods, int nr, bool need_free)
...
+	D_ALLOC_PTR(metrics);
+	if (metrics == NULL)
+		return NULL;
+
+	/** register different per-opcode counters */
+	for (opc = 0; opc < OBJ_PROTO_CLI_COUNT; opc++) {
+		/** Then the total number of requests, of type counter */
+		rc = d_tm_add_metric(&metrics->opm_total[opc], D_TM_COUNTER,
+				     "total number of processed object RPCs",
+				     "ops", "%s/ops/%s/%s%u", path,
```
mjmac, line 178: Unfortunately, this leads to some strange client metric names, e.g. `pool/$uuid/ops/ec_agg/12345`. I changed the logic a little to only append the tgt_id path for server metrics in this commit: https://github.com/daos-stack/daos/pull/13545/commits/abe1830117bc3b28c88bce10fb7b3fc77b03d03e

That commit also fixes a bug where the server metrics are created without the tgt_id.

src/pool/cli.c
```diff
@@ -558,6 +564,81 @@ pool_connect_cp(tse_task_t *task, void *data)
...
+		D_MUTEX_UNLOCK(&pool->dp_client_lock);
+		return 0;
+	}
+	D_MUTEX_UNLOCK(&pool->dp_client_lock);
+
+	snprintf(pool->dp_path, sizeof(pool->dp_path), "%s/%u/pool/"DF_UUIDF,
+		 dc_jobid, pid, DP_UUID(pool->dp_pool));
+
+	/** create new shmem space for per-pool metrics */
+	rc = d_tm_add_ephemeral_dir(NULL, size, pool->dp_path);
```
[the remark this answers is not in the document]
wangdi1: ah, the metrics are shared by all of dc_pool from the same process, hmm, dp_metrics_init probably should not in dc_pool. let me update the patch. Thanks! @mjmac

src/object/obj_utils.c
wangdi1: Thanks.

src/pool/cli.c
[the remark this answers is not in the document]
wangdi1: @mjmac  I moved per pool metrics outside of dc_pool in the updated patch, please try. Thanks!  I will play it more next week.

mjmac (dismissed):
Latest patches work well for me. Unless other reviewers have concerns, I think this is in good enough shape to land, and we can iterate further as we use the new functionality.

mchaarawi:
> Latest patches work well for me. Unless other reviewers have concerns, I think this is in good enough shape to land, and we can iterate further as we use the new functionality.

I was out for a while and would like a chance to review this. but just looking over quickly, there needs 1) more CI testing for this to actually verify metrics; and 2) performance evaluation to see what kind (if any) of performance hit this incurs when enabled.

src/include/daos/tls.h
mchaarawi: the files under include/daos/ are not required to be public, so this should be fine.

src/object/cli_shard.c
```diff
@@ -1125,7 +1227,9 @@ dc_obj_shard_rw(struct dc_obj_shard *shard, enum obj_rpc_opc opc,
 	rw_args.co = shard->do_co;
 	rw_args.shard_args = args;
 	/* remember the sgl to copyout the data inline for fetch */
-	rw_args.rwaa_sgls = (opc == DAOS_OBJ_RPC_FETCH) ? sgls : NULL;
+	rw_args.rwaa_sgls = sgls;
+	rw_args.send_time = daos_get_ntime();
```
mchaarawi, line 1231: calling this on the regular path (no metrics enabled) sounds like a bad idea IMO and should be avoided.

wangdi1:
> > Latest patches work well for me. Unless other reviewers have concerns, I think this is in good enough shape to land, and we can iterate further as we use the new functionality.
> 
> I was out for a while and would like a chance to review this. but just looking over quickly, there needs 1) more CI testing for this to actually verify metrics; and 2) performance evaluation to see what kind (if any) of performance hit this incurs when enabled.

Oh, this is only enabled when the environment is enabled, so it will not impact the current test.  I will add the test in the following patch.

wangdi1:
The failure is due to DAOS-14598

mchaarawi (changes requested):
the old client side metrics patch:
https://github.com/daos-stack/daos/pull/6185
seems more robust on what metrics are actually captured than this one, but the infrastructure is not the same as this one.

mchaarawi:
> > > Latest patches work well for me. Unless other reviewers have concerns, I think this is in good enough shape to land, and we can iterate further as we use the new functionality.
> > 
> > 
> > I was out for a while and would like a chance to review this. but just looking over quickly, there needs 1) more CI testing for this to actually verify metrics; and 2) performance evaluation to see what kind (if any) of performance hit this incurs when enabled.
> 
> Oh, this is only enabled when the environment is enabled, so it will not impact the current test. I will add the test in the following patch.

that is not my point. tests should be part of a PR not after.

mjmac:
@wangdi1 It would probably be easiest to implement tests using ftest, similar to https://github.com/daos-stack/daos/blob/master/src/tests/ftest/control/dmg_telemetry_basic.py ... I can take a stab at this. With my follow-on patch for the agent work, it's easy to query the metrics using dmg.

wangdi1:
> the old client side metrics patch: #6185 seems more robust on what metrics are actually captured than this one, but the infrastructure is not the same as this one.

Hmm, more metrics will be added later as required. This PR is only for implementing the framework with current metrics and TLS on the client side.

#6185 does not use these framework make it hard for metrics tools to retrieve these telemetry.

src/object/cli_shard.c
wangdi1: ah, sure

wangdi1:
> @wangdi1 It would probably be easiest to implement tests using ftest, similar to https://github.com/daos-stack/daos/blob/master/src/tests/ftest/control/dmg_telemetry_basic.py ... I can take a stab at this. With my follow-on patch for the agent work, it's easy to query the metrics using dmg.
> 
> Any objection to merging my PR into this one? Then we can land it all as one chunk of work.

@mjmac  Ah, sure. feel free to merge your PR here. let me update the PR by @mchaarawi  comments first. thanks

src/include/daos/metrics.h
```diff
@@ -0,0 +1,82 @@
...
+
+#include <daos/common.h>
+#include <daos/tls.h>
+#include <daos_types.h>
+#include <gurt/telemetry_common.h>
+
+/* For now TLS is only enabled if metrics are enabled */
+#define METRIC_DUMP_ENV "DAOS_METRIC_DUMP_ENV"
+#define DAOS_CLIENT_METRICS_ENV	"DAOS_CLIENT_METRICS"
+#define DAOS_CLIENT_METRICS_RETAIN_ENV	"DAOS_CLIENT_METRICS_RETAIN"
```
mchaarawi, line 23: some inconsistency in the names here.
first one should include CLIENT, remove ENV, and missing S in MERICS ->   DAOS_CLIENT_METRICS_DUMP
second one does not include any action, so should be something like -> DAOS_CLIENT_METRICS_ENABLE

src/include/gurt/telemetry_common.h
```diff
@@ -176,6 +179,7 @@ enum {
 	D_TM_ITER_RESET			= 0x002,
 };
 
+#define DC_TM_JOB_ROOT_ID	256
```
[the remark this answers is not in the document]
wangdi1: makes sense, sure

wangdi1:
ping

mjmac (dismissed):

gnailzenh (dismissed):

src/object/cli_shard.c
gnailzenh: are you going to change this or fix in following patch?

wangdi1: This has been fixed by the following commit in PR , not sure why it is not shown here. hmm
[
](https://github.com/daos-stack/daos/pull/13517/commits/8182b70a22a0fae570c438ae3cb29cd7c3940d56)

daltonbohning:
Did we want to run performance tests with these changes first?

src/client/api/metrics.c
mjmac: For the record, I went ahead and implemented this in my follow-on PR: https://github.com/daos-stack/daos/pull/13545

wangdi1:
> @wangdi1: Are you planning to merge latest master in? Would be nice to get the initial cut of this work landed so that we can continue to iterate.

Sure, I will. Though @daltonbohning  needs to evaluation the performance impacts before landing.

daltonbohning:
> > @wangdi1: Are you planning to merge latest master in? Would be nice to get the initial cut of this work landed so that we can continue to iterate.
> 
> Sure, I will. Though @daltonbohning needs to evaluation the performance impacts before landing.

I plan to run some performance tests for this over the weekend. Hopefully by Monday we'll have an idea

src/gurt/telemetry.c
```diff
@@ -2522,15 +2586,18 @@ d_tm_add_ephemeral_dir(struct d_tm_node_t **node, size_t size_bytes,
...
-	link_node->dtn_metric->dtm_data.value = key;
 
 	/* track attached regions within the parent shmem */
 	parent_shmem = get_shmem_for_key(ctx, link_node->dtn_shmem_key);
 	if (parent_shmem == NULL) {
 		D_ERROR("failed to get parent shmem pointer\n");
 		D_GOTO(fail_link, rc = -DER_NO_SHMEM);
 	}
+
+	D_ASSERT(link_node->dtn_type == D_TM_LINK);
```
daltonbohning, line 2597: I'm tripping this assert on Frontera. See this comment for details
https://daosio.atlassian.net/browse/DAOS-15083?focusedCommentId=124040

daltonbohning (dismissed):

src/gurt/telemetry.c
daltonbohning: I can no longer reproduce this, so maybe something strange in my environment. I'm running performance testing now

mjmac: @daltonbohning Any updates on the test results?

daltonbohning: Funny. I _**just**_ pulled them together :)
I ran IOR Easy with SX on Frontera, based on our standard performance suite. There doesn't seem to be any significant difference between `D_CLIENT_METRICS_ENABLE=0`, `D_CLIENT_METRICS_ENABLE=1`, and master.

Unless we think more performance testing is needed, I think it's okay to move forward.

daltonbohning: After discussing this internally, we need to check the performance overhead with many RPCs. The IOR Easy tests I ran use 1M transfer, so there aren't many RPCs. I'll verify with MDTest Hard, which has many smaller IOs

daltonbohning: The queue has been busy, so most of my performance jobs are still pending

wangdi1:
Update to resolve conflict.

kjacque (dismissed):
Just a few comments, nothing I'd block on. Nice work. A bit painful to re-review something this big after a force push though.

mjmac:
@wangdi1: Look like there's another conflict to resolve. :(

@mchaarawi, @ashleypittman: Have your -1 concerns been addressed? Would be good to get this landed, as there is a bunch of follow-on work waiting on it. I understand the desire to have ftest stuff added as part of the landing, but IMO it's too early for that.

src/object/cli_shard.c
```diff
@@ -1125,7 +1227,9 @@ dc_obj_shard_rw(struct dc_obj_shard *shard, enum obj_rpc_opc opc,
 	rw_args.co = shard->do_co;
 	rw_args.shard_args = args;
 	/* remember the sgl to copyout the data inline for fetch */
-	rw_args.rwaa_sgls = (opc == DAOS_OBJ_RPC_FETCH) ? sgls : NULL;
+	rw_args.rwaa_sgls = sgls;
+	rw_args.send_time = daos_client_metric ? daos_get_ntime() : 0;
+	obj_shard_update_metrics_begin(req);
```
mchaarawi, line 1232: this is better, but still i'd rather have:
if (daos_client_metrics) {
  rw_args.send_time = daos_get_ntime();
  obj_shard_update_metrics_begin(req);
} else {
  rw_args.send_time = 0
}
and same for the obj_shard_update_metrics_end() calls.
i know we currently return immediately from those calls, but this is probably cleaner / safer for future changes that might go into those funcitons.

mchaarawi: it's fine unless you need to repush which looks like you might need to anyway because of a merge conflict

mchaarawi:
> @wangdi1: Look like there's another conflict to resolve. :(
> 
> @mchaarawi, @ashleypittman: Have your -1 concerns been addressed? Would be good to get this landed, as there is a bunch of follow-on work waiting on it. I understand the desire to have ftest stuff added as part of the landing, but IMO it's too early for that.

we discussed this at the perf working group and i thought there is another round of testing needed than what was done.

as for the ftest stuff, TBH this is exactly what we proposed feature branches for. I know feature branches are a pain and I have suffered from it, but that was the process.. I won't block on this though if other gatekeepers are OK.

daltonbohning (changes requested):
I'm seeing seg faults again with `D_CLIENT_METRICS_ENABLE=1` for most of my Frontera runs. Will discuss with @wangdi1

src/include/daos/pool.h
```diff
@@ -96,6 +98,7 @@ struct dc_pool {
 	pthread_rwlock_t	dp_map_lock;
 	struct pool_map	       *dp_map;
 	tse_task_t	       *dp_map_task;
+	void			**dp_metrics;
```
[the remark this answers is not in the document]
wangdi1: I thought each module (object/container/pool etc) will register its own per pool metrics  to the pool, so it can not be declared as a specific type.

src/client/api/metrics.c
```diff
@@ -0,0 +1,157 @@
...
+	rc = d_tm_init(DC_TM_JOB_ROOT_ID, MAX_IDS_SIZE(INIT_JOB_NUM), metrics_tag);
+	if (rc != 0) {
+		DL_ERROR(rc, "init job root id.");
+		return rc;
+	}
+
+	D_INFO("INIT %s metrics\n", dc_jobid);
+	rc = d_tm_add_ephemeral_dir(NULL, MAX_IDS_SIZE(INIT_JOB_NUM), "%s", dc_jobid);
+	if (rc != 0 && rc != -DER_EXIST) {
+		DL_ERROR(rc, "add metric %s failed.\n", dc_jobid);
```
ashleypittman, line 58: You need to resolve these.  DL_ERROR will add the formatting for you so remove the ".\n" at the end.
```suggestion
		DL_ERROR(rc, "add metric %s failed", dc_jobid);
```

src/cart/README.env
```diff
@@ -174,3 +174,15 @@ This file lists the environment variables used in CaRT.
  . CRT_TEST_CONT
    When set to 1, orterun does not automatically shut down other servers when
    one server is shutdown. Used in cart internal testing.
+
+ . D_CLIENT_METRICS_ENABLE
```
ashleypittman, line 178: Should these be added to the list in src/cart/crt_init.c as well?

ashleypittman (commented):
I think my other issues are resolved now.

src/gurt/telemetry.c
```diff
@@ -569,13 +572,16 @@ alloc_node(struct d_tm_shmem_hdr *shmem, struct d_tm_node_t **newnode,
 		rc = -DER_NO_SHMEM;
 		goto out;
 	}
-	rc = init_node(shmem, node, name);
+
+	tmp = conv_ptr(shmem, node);
```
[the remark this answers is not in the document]
wangdi1: the return value is assigned to the sharemem struct directly somewhere, so it can blindly convert the ptr internally in shmalloc.  Though I agree the convertion is troublesome indeed, will try refining it in the following patch.

daltonbohning:
My seg faults are
```
==== backtrace (tid: 144701) ====
0 0x0000000000036400 killpg()  ???:0
1 0x0000000000033660 add_child()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517/20240126/daos/src/gurt/telemetry.c:661
2 0x0000000000033660 add_metric()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517/20240126/daos/src/gurt/telemetry.c:2196
3 0x0000000000035665 d_tm_add_ephemeral_dir()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517/20240126/daos/src/gurt/telemetry.c:2584
4 0x000000000004665f dc_tm_init()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517/20240126/daos/src/client/api/metrics.c:56
5 0x000000000003abae daos_init()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517/20240126/daos/src/client/api/init.c:246
6 0x0000000000409872 DFS_Init()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517/20240126/ior/src/aiori-DFS.c:492
7 0x0000000000418df8 mdtest_run()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517/20240126/ior/src/mdtest.c:2386
8 0x0000000000404c02 main()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517/20240126/ior/src/mdtest-main.c:6
9 0x0000000000022555 __libc_start_main()  ???:0
10 0x0000000000404c37 _start()  ???:0
```
Which is line
https://github.com/daos-stack/daos/blob/fa5eea605cc4b0500b1d93a29dbbc978b6bf1b2a/src/gurt/telemetry.c#L661

and
```
(gdb) bt
#0  get_free_region_entry (entry=<synthetic pointer>, shmem=0x2b1b11fd3000) at src/gurt/telemetry.c:2443
#1  d_tm_add_ephemeral_dir (node=0x0, node@entry=<error reading variable: Cannot access memory at address 0x7ffec755ca48>, size_bytes=<optimized out>, fmt=<optimized out>) at src/gurt/telemetry.c:2601
```
Which is line
https://github.com/daos-stack/daos/blob/fa5eea605cc4b0500b1d93a29dbbc978b6bf1b2a/src/gurt/telemetry.c#L2443

src/gurt/telemetry.c
```diff
@@ -831,19 +854,46 @@ d_tm_init(int id, uint64_t mem_size, int flags)
...
+/* Check if all children are invalid */
+static bool
+is_node_empty(struct d_tm_node_t *node)
+{
+	struct d_tm_context   *ctx = tm_shmem.ctx;
+	struct d_tm_shmem_hdr *shmem;
+	struct d_tm_node_t    *child;
+
+	shmem = get_shmem_for_key(ctx, node->dtn_shmem_key);
+	child = conv_ptr(shmem, node->dtn_child);
```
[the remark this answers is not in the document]
kjacque: Just looking at the existing telemetry code, I noticed a few helpers already exist. `d_tm_get_child` and `d_tm_get_sibling` in particular. It may be a good idea to use these to clean up now, perhaps it will help to narrow down where the segfaults Dalton discovered are happening.

src/client/api/metrics.c
wangdi1: > You need to resolve these. DL_ERROR will add the formatting for you so remove the ".\n" at the end.

sure, thanks

src/cart/README.env
wangdi1: ah, sure.

wangdi1:
@daltonbohning  Could you please try this update PR? thanks.

daltonbohning:
> @daltonbohning Could you please try this update PR? thanks.

Another segfault:
```
[c171-133:53563:0:53563] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x8)
[c179-202:242334:0:242334] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x8)
==== backtrace (tid:  53563) ====
 0 0x0000000000036400 killpg()  ???:0
 1 0x00000000000336e0 add_child()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517_2/20240206/daos/src/gurt/telemetry.c:661
 2 0x00000000000336e0 add_metric()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517_2/20240206/daos/src/gurt/telemetry.c:2196
 3 0x00000000000356f5 d_tm_add_ephemeral_dir()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517_2/20240206/daos/src/gurt/telemetry.c:2588
 4 0x000000000004679f dc_tm_init()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517_2/20240206/daos/src/client/api/metrics.c:56
 5 0x000000000003ac5e daos_init()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517_2/20240206/daos/src/client/api/init.c:246
 6 0x0000000000409872 DFS_Init()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517_2/20240206/ior/src/aiori-DFS.c:492
 7 0x0000000000418df8 mdtest_run()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517_2/20240206/ior/src/mdtest.c:2386
 8 0x0000000000404c02 main()  /work2/08126/dbohninx/frontera/BUILDS/daos-15083_pr13517_2/20240206/ior/src/mdtest-main.c:6
 9 0x0000000000022555 __libc_start_main()  ???:0
10 0x0000000000404c37 _start()  ???:0
```

kjacque:
The places where Dalton's failures are happening imply to me that the shmem is disappearing out from under the caller while they are still in the process of adding the ephemeral dir. Seems like a race to me.

@wangdi1 maybe we need to do some more access synchronization around creating/deleting ephemeral directories. It's more overhead, but in a world of multiple writers, seems necessary now.

wangdi1:
> The places where Dalton's failures are happening imply to me that the shmem is disappearing out from under the caller while they are still in the process of adding the ephemeral dir. Seems like a race to me.
> 
> @wangdi1 maybe we need to do some more access synchronization around creating/deleting ephemeral directories. It's more overhead, but in a world of multiple writers, seems necessary now.

indeed. hmm

daltonbohning:
> maybe we need to do some more access synchronization around creating/deleting ephemeral directories. It's more overhead, but in a world of multiple writers, seems necessary now.

FWIW, I am running with 32 client processes per node. Which might be why I'm the first to see this

wangdi1:
> > maybe we need to do some more access synchronization around creating/deleting ephemeral directories. It's more overhead, but in a world of multiple writers, seems necessary now.
> 
> FWIW, I am running with 32 client processes per node. Which might be why I'm the first to see this

I added ephemeral directory synchronization @kjacque . please take a look. @daltonbohning  please try the patch. Thanks!

daltonbohning:
> > > maybe we need to do some more access synchronization around creating/deleting ephemeral directories. It's more overhead, but in a world of multiple writers, seems necessary now.
> > 
> > 
> > FWIW, I am running with 32 client processes per node. Which might be why I'm the first to see this
> 
> I added ephemeral directory synchronization @kjacque . please take a look. @daltonbohning please try the patch. Thanks!

@wangdi1 @kjacque Does it make sense for me to test this with multiple client nodes, or would 1 client and many processes be similar? I.e. is the directory creation per agent/client node?

wangdi1:
> > > > maybe we need to do some more access synchronization around creating/deleting ephemeral directories. It's more overhead, but in a world of multiple writers, seems necessary now.
> > > 
> > > 
> > > FWIW, I am running with 32 client processes per node. Which might be why I'm the first to see this
> > 
> > 
> > I added ephemeral directory synchronization @kjacque . please take a look. @daltonbohning please try the patch. Thanks!
> 
> @wangdi1 @kjacque Does it make sense for me to test this with multiple client nodes, or would 1 client and many processes be similar? I.e. is the directory creation per agent/client node?

yes, you can test it with single client and multiple threads. and the metrics directory  does create per client here.

wangdi1:
> @wangdi1 Two issues with the latest.
> 
> 1. Most mdtest runs just hang forever with no errors in logs.
> 2. One reported `ERROR (aiori-DFS.c:493): 244: -1038: Failed to initialize daos`
>    And from the logs
> 
> ```
> DAOS[12827/12827/0] telem ERR  src/gurt/telemetry.c:629 add_child() can't get parent node shmem region, key=0x0
> DAOS[12827/12827/0] telem ERR  src/gurt/telemetry.c:676 add_child() Failed to add metric [c142-072.frontera.tacc.utexas.edu-12827]: DER_NO_SHMEM(-1038): 'Not enough shared memory free'
> 0DAOS[12827/12827/0] telem ERR  src/gurt/telemetry.c:2605 d_tm_add_ephemeral_dir() can't set up the link node, DER_NO_SHMEM(-1038): 'Not enough shared memory free'
> DAOS[12827/12827/0] telem ERR  src/gurt/telemetry.c:2648 d_tm_add_ephemeral_dir() Failed to add ephemeral dir [c142-072.frontera.tacc.utexas.edu-12827]: DER_NO_SHMEM(-1038): 'Not enough shared memory free'
> DAOS[12827/12827/0] daos ERR  src/client/api/metrics.c:67 dc_tm_init() add metric c142-072.frontera.tacc.utexas.edu-12827/12827 failed.
> : DER_NO_SHMEM(-1038): 'Not enough shared memory free'
> ```
> 
> If this code is complicated / problematic then we probably need CI tests to verify.

Could you please delete those sharememory before running the job. from ipcrm -a on all client nodes? Ah there is a known shmem size calculation issue if you run too much ranks per node.  I think @mjmac  point this out before, but we decided to resolve this later in the following patch. 

Could you please run this with fewer ranks per clients? like <= 16 ranks per client.

I will see what I can do here. Thanks.

daltonbohning:
> Could you please delete those sharememory before running the job. from ipcrm -a on all client nodes? Ah there is a known shmem size calculation issue if you run too much ranks per node. I think @mjmac point this out before, but we decided to resolve this later in the following patch.

Does it matter to run `ipcrm -a` if there are known issues with shmem calculation?

> Could you please run this with fewer ranks per clients? like <= 16 ranks per client.

Even if it works, the performance numbers won't be meaningful because we typically run with 32 ranks per client.


IMO, we should resolve the synchronization issues before running on Frontera. It's time-consuming to test small changes like this on Frontera, and we're running v2.4.2 testing now. Let me try to get a reproducer on boro so we can debug this more quickly.

daltonbohning:
> IMO, we should resolve the synchronization issues before running on Frontera. It's time-consuming to test small changes like this on Frontera, and we're running v2.4.2 testing now. Let me try to get a reproducer on boro so we can debug this more quickly.

And if I can get a reproducer on boro, I can help with a CI test

daltonbohning:
Different issue on boro. I get past daos_init and ior or mdtest runs okay. But sometimes during cleanup (maybe daos_fini?) I see
```
telem ERR  src/gurt/telemetry.c:253 destroy_shmem() Unable to remove shared memory segment (shmid=917536). shmctl failed, Invalid argument.
```

wangdi1:
> Different issue on boro. I get past daos_init and ior or mdtest runs okay. But sometimes during cleanup (maybe daos_fini?) I see
> 
> ```
> telem ERR  src/gurt/telemetry.c:253 destroy_shmem() Unable to remove shared memory segment (shmid=917536). shmctl failed, Invalid argument.
> ```

only with this  D_CLIENT_METRICS_ENABLE=1 setting?  how many clients and ranks per client in your run? @daltonbohning  thanks.

daltonbohning:
> > Different issue on boro. I get past daos_init and ior or mdtest runs okay. But sometimes during cleanup (maybe daos_fini?) I see
> > ```
> > telem ERR  src/gurt/telemetry.c:253 destroy_shmem() Unable to remove shared memory segment (shmid=917536). shmctl failed, Invalid argument.
> > ```
> 
> only with this D_CLIENT_METRICS_ENABLE=1 setting? how many clients and ranks per client in your run? @daltonbohning thanks.

Only with `D_CLIENT_METRICS_ENABLE=1` and with as few as 1 client node, 2 client procs. FWIW, I'm running the server on the same node.

wangdi1:
> > Could you please delete those sharememory before running the job. from ipcrm -a on all client nodes? Ah there is a known shmem size calculation issue if you run too much ranks per node. I think @mjmac point this out before, but we decided to resolve this later in the following patch.
> 
> Does it matter to run `ipcrm -a` if there are known issues with shmem calculation?
> 
yes, it will help to resolve this shmem calculation issue. And also the updated PR actually change the format of shmem a bit. so if you do not delete the old shmem, which might cause hang.  I will play with larger scale in boro see how it goes. Thanks.

daltonbohning:
@wangdi1 I so far can't reproduce the hangs or segfaults on boro. Even with 144 ranks one one client. I'll queue some more jobs on Frontera, but I'm not sure when they'll finish. We're running 2.4.2 testing and Frontera will be unavailable Friday through Tuesday

src/gurt/telemetry.c
```diff
@@ -31,12 +31,17 @@ struct shmem_region_list {
 struct d_tm_shmem_hdr {
 	uint64_t		 sh_base_addr;	/** address of this struct */
 	key_t			 sh_key;	/** key to access region */
-	bool			 sh_deleted;	/** marked for deletion */
+	uint32_t		 sh_deleted:1,	/** marked for deletion */
+				 sh_require_ephemeral_lock:1; /** ephemeral_lock required */
```
kjacque, line 35: I am a little hesitant to have the flag be quite this specific. Maybe a name like "sh_multiple_writers" would better document our reasoning for needing a flag to begin with, and would allow us to do more things dependent on that, if we find more synchronization is needed later.

src/gurt/telemetry.c
```diff
@@ -809,6 +816,11 @@ d_tm_init(int id, uint64_t mem_size, int flags)
 		D_INFO("Retaining shared memory for id %d if not empty\n", id);
 	}
 
+	if (flags & D_TM_EPHEMERAL_DIR_LOCK) {
```
kjacque, line 819: Same concerns about this name as above.

src/gurt/telemetry.c
```diff
@@ -2556,10 +2572,13 @@ d_tm_add_ephemeral_dir(struct d_tm_node_t **node, size_t size_bytes,
 	if (rc != 0)
 		D_GOTO(fail, rc);
 
+	if (tm_shmem.ephemeral_dir_lock)
```
kjacque, line 2575: Are we only concerned with the local setting, and not the flag in the shmem header? If we don't want each shmem region to potentially have its own settings, we should only have the flag in one place.

kjacque (changes requested):

mjmac:
@wangdi1, @kjacque: Can we please arrive at a consensus on how to get this landed? I have two follow-on PRs that are waiting for this one (#13545, #5382). The most recent comments by Kris look like only small adjustments. There will undoubtedly be some more adjustments needed as we start to really use this feature, but as long as we're not introducing any regressions for the non-telemetry use cases, then I don't think it needs to be perfect before landing.

src/gurt/telemetry.c
```diff
@@ -2455,7 +2467,11 @@ get_free_region_entry(struct d_tm_shmem_hdr *shmem,
 	}
 
 	next = conv_ptr(shmem, head->next);
-	cur = head->next;
+	if (d_list_empty(&shmem->sh_subregions))
+		cur = (d_list_t *)(shmem->sh_base_addr +
+		       (uint64_t)(&((struct d_tm_shmem_hdr *)(0))->sh_subregions));
```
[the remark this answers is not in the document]
wangdi1: The list is initialized by D_INIT_LIST_HEAD(&header->sh_subregions), so it is not a shmem address if it is empty, if I understand correctly.

src/gurt/telemetry.c
wangdi1: yes, makes sense. Thanks

kjacque (approved):

mchaarawi:
> @wangdi1, @kjacque: Thanks for the quick turnaround on the feedback. FYI, I found an issue in my testing that I've fixed in my follow-on: https://github.com/daos-stack/daos/pull/13545/files#diff-49b244164a1877abd5a7363633cb22baec5bce3b270d978f1beea89a00761bbcR2730
> 
> With that fix in place, the agent was able to manage the per-job/pid subdirs with no problems, and ran for a few hours with telemetry managed for hundreds of client processes.
> 
> @mchaarawi, @ashleypittman, @daltonbohning: Can we please land these two PRs so that we can keep making progress? I think Dalton's testing has shown that there is no appreciable performance impact with metrics enabled or disabled, correct?

I don't see any feedback from Dalton on his latest run whether that segfaulted again or not.. but we probably need to run the test again since it was 2 weeks ago and more changes were landed to this PR.

daltonbohning:
> I don't see any feedback from Dalton on his latest run whether that segfaulted again or not.. but we probably need to run the test again since it was 2 weeks ago and more changes were landed to this PR.

The last I ran with `D_CLIENT_METRICS_ENABLE=1` I was still getting segfaults or hangs on init.
With `D_CLIENT_METRICS_ENABLE=0`, I don't yet have a definite answer on the performance impact.
I should finish 2.4.2 RC testing soon and can get back to this.

daltonbohning:
> > I don't see any feedback from Dalton on his latest run whether that segfaulted again or not.. but we probably need to run the test again since it was 2 weeks ago and more changes were landed to this PR.
> 
> The last I ran with `D_CLIENT_METRICS_ENABLE=1` I was still getting segfaults or hangs on init. With `D_CLIENT_METRICS_ENABLE=0`, I don't yet have a definite answer on the performance impact. I should finish 2.4.2 RC testing soon and can get back to this.

Looking back at my mdtest hard results, it seems unlikely that `D_CLIENT_METRICS_ENABLE=0` introduces significant overhead, though I need to run more tests.
So @mchaarawi if only the non-default (`D_CLIENT_METRICS_ENABLE=1`) is an issue, should we move forward with this PR and fix that after? Due to some issues on Frontera I'm not sure when I can re-test this.

mchaarawi:
> For what it's worth, my follow-on PR based on this one passes all of the linter checks: #13545
> 
> Can we please get another +1 on it and get it landed? @johannlombardi, @mchaarawi, @ashleypittman? Thanks in advance. :)

im going to object on this until 1) we have proved that the segfault issue is resolved/tested, and 2) perf numbers show there is no affect on the non-metric path.

mjmac:
> im going to object on this until 1) we have proved that the segfault issue is resolved/tested, and 2) perf numbers show there is no affect on the non-metric path.

OK. @daltonbohning: Please advise when you'll have time to re-run the performance tests.

daltonbohning:
> > im going to object on this until 1) we have proved that the segfault issue is resolved/tested, and 2) perf numbers show there is no affect on the non-metric path.
> 
> OK. @daltonbohning: Please advise when you'll have time to re-run the performance tests.

Sorry, I'm behind. I should get back to this either this week or early next week.

daltonbohning:
With the latest version, I still see mdtest hanging with `D_CLIENT_METRICS_ENABLE=0`. And it seems to only happen (or at least be more likely) with multiple client nodes.

PR #13519, opened 2023-12-18
DAOS-14539 client: intercept execve() and it variants

wiliamhuang:
Features: pil4dfs

envp different from environ could be passed into execve() and execvpe() to override the default environ. Under rare circumstances (e.g., scons, git), IL related envs are not propagated from parent process to newly created child processes. D_IL_ENFORCE_EXEC_ENV=1 will force libpil4dfs related env variables appended when exec() is called. This is turned off as default.

Required-githooks: true

### Gatekeeper:


wiliamhuang:
Just converted this PR back to draft. I will spend a little time to double check all resources are cleaned up before exec().

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
...
+				if (strstr(envp[num_entry], "libpil4dfs.so"))
+					pil4dfs_in_preload = true;
+			} else if (strncmp(envp[num_entry], "D_IL_REPORT", 11) == 0) {
+				report_included = true;
+			} else if (strncmp(envp[num_entry], "DAOS_MOUNT_POINT", 16) == 0) {
+				mp_included = true;
+			} else if (strncmp(envp[num_entry], "DAOS_POOL", 9) == 0) {
+				pool_included = true;
+			} else if (strncmp(envp[num_entry], "DAOS_CONTAINER", 14) == 0) {
+				cont_included = true;
```
mchaarawi, line 3367: what about other env variables that DAOS users might set? 
D_LOG_MASK, D_LOG_FILE, etc ?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
 	return &mydir->ents[mydir->num_ents];
 }
 
+extern char **__environ;
+
+/* This number may be updated later to be consistent!*/
+#define N_ENV_CHECK (7)
```
ashleypittman, line 3315: Consistent with what?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
...
- *		return pid;
- *	}
- *}
+ * char    env_list[N_ENV_CHECK][32] = {"LD_PRELOAD", "D_IL_REPORT", "DAOS_MOUNT_POINT",
+ *				     "DAOS_POOL", "DAOS_CONTAINER", "D_IL_MAX_EQ",
+ *				     "D_IL_ENFORCE_EXEC_ENV"};
  */
 
+static char **
+pre_envp(char *const envp[])
```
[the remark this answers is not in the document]
ashleypittman: Perhaps you could also explain why this is needed as well?  Environment variables are usually inherited by default by both child processes and their children, why is there a hook to re-add them needed?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
...
+	char  *str_maxeq            = NULL;
+	char  *str_enforcement      = NULL;
+
+	if (envp == NULL) {
+		num_entry = 0;
+	} else if (envp[0] == NULL) {
+		num_entry = 0;
+	} else {
+		while (envp[num_entry]) {
+			if (strncmp(envp[num_entry], "LD_PRELOAD", 10) == 0) {
```
ashleypittman, line 3355: This is a very error prone way of handling strings although it may be that the numbers you have in this PR are correct, please consider something like the following or better still wrapping this up in a macro so the string can only be passed in once.
```suggestion
			if (strncmp(envp[num_entry], "LD_PRELOAD", sizeof"LD_PRELOAD")) == 0) {
```

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
...
+			} else if (strncmp(envp[num_entry], "D_IL_MAX_EQ", 11) == 0) {
+				maxeq_included = true;
+			} else if (strncmp(envp[num_entry], "D_IL_ENFORCE_EXEC_ENV", 21) == 0) {
+				enforcement_included = true;
+			}
+			num_entry++;
+		}
+	}
+
+	/* The fd for daos logging may be not valid here, so use plain printf */
```
ashleypittman, line 3377: This comment applies to many places in this function, assuming it's correct then put it once in the function description at the start.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
...
+	new_envp = malloc(sizeof(char *) * (num_entry + N_ENV_CHECK + 1));
+	if (new_envp == NULL) {
+		printf("Error: failed to allocate memory for new_envp. Use existing envp\n");
+		goto err_out0;
+	}
+
+	pil4df_path = query_pil4dfs_path();
+	len2        = strnlen(pil4df_path, PATH_MAX);
+	/* copy existing entries */
```
ashleypittman, line 3386: You need a better comment for what this loop is doing.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
...
+		new_envp[i] = str_report;
+		i++;
+	}
+	if (!mp_included) {
+		fs_root = getenv("DAOS_MOUNT_POINT");
+		if (fs_root) {
+			rc = asprintf(&str_mp, "DAOS_MOUNT_POINT=%s", fs_root);
+			if (rc < 0) {
+				printf("Error: failed to allocate memory for DAOS_MOUNT_POINT "
+				       "env!\n");
```
ashleypittman, line 3434: Applies elsewhere but please don't use ! in error messages!

It looks like this would all fit on line line as well.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
...
+err_out2:
+	free(new_preload_str);
+err_out1:
+	free(new_envp);
+err_out0:
+	return (char **)envp;
+}
+
+/* check whether fd 0, 1, and 2 are located on DFS mount or not. If yes, reopen files and
+ * set offset.
```
ashleypittman, line 3513: Is offset meaningful here?  At least for 1 and 2 it could be, but if it is then this seems like something that would benefit from a test.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3284,71 +3309,392 @@ new_readdir(DIR *dirp)
...
+	for (i = 0; i < MAX_FD_DUP2ED; i++) {
+		/* only check fd 0, 1, and 2 */
+		if (fd_dup2_list[i].fd_src >= 0 && fd_dup2_list[i].fd_src <= 2) {
+			fd        = fd_dup2_list[i].fd_src;
+			idx       = fd_dup2_list[i].fd_dest - FD_FILE_BASE;
+			offset    = file_list[idx]->offset;
+			open_flag = file_list[idx]->open_flag;
+
+			/* get a real fd from kernel */
+			fd_tmp = libc_open(file_list[idx]->path, open_flag);
```
ashleypittman, line 3533: Do you have any consideration for if the path has changed, this could legitimately return a different file or ENOENT I assume?

src/tests/ftest/daos_test/dfuse.py
```diff
@@ -91,8 +91,15 @@ def run_test(self, il_lib=None):
...
             daos_test_env['D_LOG_MASK'] = 'INFO,IL=DEBUG'
-
-        command = [self.daos_test, '--test-dir', mount_dir, '--io', '--stream']
+            if il_lib == 'libpil4dfs.so':
+                daos_test_env['D_IL_REPORT'] = '0'
+                daos_test_env['DAOS_MOUNT_POINT'] = f'{mount_dir}'
+                daos_test_env['DAOS_POOL'] = f'{self.pool.identifier}'
+                daos_test_env['DAOS_CONTAINER'] = f'{self.container.uuid}'
+                daos_test_env['D_IL_MAX_EQ'] = '2'
+                daos_test_env['D_IL_ENFORCE_EXEC_ENV'] = '1'
```
ashleypittman, line 100: This is not a new test, what has changed to mean these are required now?

src/tests/ftest/dfuse/bash.py
```diff
@@ -119,6 +119,9 @@ def run_bashcmd(self, il_lib=None, compatible_mode=False):
                             f'realpath {link_name}',
                             f'head {fuse_root_dir}/src.c',
                             f'tail {fuse_root_dir}/src.c',
+                            'printf "exec 5>config.log\n{\ncat <<EOF\n------\nEOF\n} >&5\n"'
+                            f'> {fuse_root_dir}/1.sh; cd {fuse_root_dir}; '
+                            f'/usr/bin/sh {fuse_root_dir}/1.sh',
```
ashleypittman, line 124: I assume this is covered in https://github.com/daos-stack/daos/pull/13681 in which case please drop the changes to this file.

ashleypittman (commented):
Neither the ticket or PR description elaborates, why do we need to intercept evecve?  Is this so that we can run binaries from daos using pil4dfs without a dfuse backend?

src/tests/ftest/dfuse/bash.py
wiliamhuang: ok. I will drop the changes here. Thank you!

src/tests/ftest/daos_test/dfuse.py
wiliamhuang: The new tests in dfuse_test will check libpil4dfs related env to make sure they are enforced to set in child process.

src/client/dfuse/pil4dfs/hook.c
```diff
@@ -308,6 +309,24 @@ determine_lib_path(void)
 	}	
 	D_FREE(lib_dir_str);
 
+	pos = strstr(read_buff_map, "libpil4dfs.so");
```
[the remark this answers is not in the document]
wiliamhuang: The comment was revised.

src/client/dfuse/pil4dfs/int_dfs.c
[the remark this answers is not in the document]
wiliamhuang: Comments are added. Yes. env variables are normally inherited from the parent process. envp could be passed into execve() and execvpe() to override the default envp. env was cleared in many child processes created by python. Intercepting execve() etc. can force the needed env set if absent.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! I rearranged the code here and added comments.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! Removed "!". What did you mean "all fit on line line"? Is it "all fit on one line"?

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: One test extracted from "configure" is added in dfuse/bash.py. The code here was removed and put in dup2() as a special case for bash/sh.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: The code is moved to dup2(). If libc_open() fails, then dup2() fails with libc_open()'s errno.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: The comment was revised. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Updated as you suggested. Thank you!
I will replace them with a macro as you suggested when I refresh my PR next time.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Moved this line after function description. Thank you!

wiliamhuang:
I will set D_IL_ENFORCE_EXEC_ENV off as default when I refresh this PR.

src/tests/ftest/daos_test/dfuse.py
```diff
@@ -91,8 +91,15 @@ def run_test(self, il_lib=None):
             daos_test_env['DD_MASK'] = 'all'
             daos_test_env['DD_SUBSYS'] = 'all'
             daos_test_env['D_LOG_MASK'] = 'INFO,IL=DEBUG'
-
-        command = [self.daos_test, '--test-dir', mount_dir, '--io', '--stream']
+            if il_lib == 'libpil4dfs.so':
+                daos_test_env['D_IL_REPORT'] = '0'
+                daos_test_env['DAOS_MOUNT_POINT'] = f'{mount_dir}'
+                daos_test_env['DAOS_POOL'] = f'{self.pool.identifier}'
+                daos_test_env['DAOS_CONTAINER'] = f'{self.container.uuid}'
```
daltonbohning, line 98: No need to use f-strings for these. Also, any reason why the container label wouldn't work?
```suggestion
                daos_test_env['DAOS_MOUNT_POINT'] = mount_dir
                daos_test_env['DAOS_POOL'] = self.pool.identifier
                daos_test_env['DAOS_CONTAINER'] = self.container.identifier
```

wiliamhuang: Container label should work. I will fix them. Thank you very much!

wiliamhuang:
One ftest failed. It will need a bit of work to fix the issue.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: You could remove the newline so the code could fit on one line of the source file rather than being wrapped onto 2.  We allow 100 characters per line.

src/tests/ftest/daos_test/dfuse.py
wiliamhuang: Fixed as you suggested. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! I double-checked this line. There are still 101 characters in this line after "\n" is removed.

wiliamhuang:
@ashleypittman @mchaarawi Could you please review this PR? Thank you very much!

src/tests/suite/dfuse_test.c
```diff
@@ -48,6 +49,9 @@ print_usage()
 	print_message("dfuse_test -m|--metadata\n");
 	print_message("dfuse_test -d|--directory\n");
 	print_message("dfuse_test -l|--lowfd\n");
+	print_message("dfuse_test -e|--exec\n");
+	/* verifyenv is only run by exec test. Should not be executed directly */
+	/* print_message("dfuse_test -v|--verifyenv\n");                       */
```
ashleypittman, line 54: Is this is internal then only support the long option here.  -v could well use run by someone wanting to check the version.
```suggestion
	/* print_message("dfuse_test    --verifyenv\n");                       */
```

src/tests/suite/dfuse_test.c
```diff
@@ -654,9 +787,11 @@ main(int argc, char **argv)
 					       {"metadata", no_argument, NULL, 'm'},
 					       {"directory", no_argument, NULL, 'd'},
 					       {"lowfd", no_argument, NULL, 'l'},
+					       {"exec", no_argument, NULL, 'e'},
+					       {"verifyenv", no_argument, NULL, 'v'},
 					       {NULL, 0, NULL, 0}};
 
-	while ((opt = getopt_long(argc, argv, "aM:imsdl", long_options, &index)) != -1) {
+	while ((opt = getopt_long(argc, argv, "aM:imsdlev", long_options, &index)) != -1) {
```
ashleypittman, line 794: You can support the long option only by leaving the short option out of this string.
```suggestion
	while ((opt = getopt_long(argc, argv, "aM:imsdle", long_options, &index)) != -1) {
```

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,347 @@ new_readdir(DIR *dirp)
 	return &mydir->ents[mydir->num_ents];
 }
 
+extern char **__environ;
+
+/* This is the number of environmental variables that would be forced to set in child process. */
+#define N_ENV_APPEND (7)
```
ashleypittman, line 3341: Should this be forwarding n D_LOG_FILE as well?

src/tests/suite/dfuse_test.c
wiliamhuang: Thank you! I will remove "-v".

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5680,6 +5983,13 @@ init_myhook(void)
 			report = false;
 		d_freeenv_str(&env_log);
 	}
+	enforce_exec_env = false;
+	d_agetenv_str(&env_exec, "D_IL_ENFORCE_EXEC_ENV");
```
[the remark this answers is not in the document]
wiliamhuang: Thank you! I will use d_getenv_bool() instead.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! I will add "D_LOG_FILE". Shall we add "DD_MASK", "DD_SUBSYS", and "D_LOG_MASK" too?

src/tests/suite/dfuse_test.c
wiliamhuang: Thank you! I will fix it as you suggested.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: Yes, I think so.

It would be a good idea to re-write this code to use and array (possilby you had it like that before?) so that adding new ones is easy as I'm sure we're going to find new entries to add to the list at some point.  I think LD_PRELOAD is the only one that needs a special case here, the rest all copy from the current process I assume?

wiliamhuang: Thank you! I will look into rewriting the code here as you suggested.

wiliamhuang: I just updated the code to loop over a given env list to append. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,282 @@ new_readdir(DIR *dirp)
...
+			goto err_out1;
+		}
+		new_envp[i] = new_preload_str;
+		i++;
+	}
+
+	for (j = 0; j < LEN_ENV_LIST; j++) {
+		/* env is not found. Need to be appended if present in current process. */
+		if (!env_found[j]) {
+			rc = d_agetenv_str(&env_value, env_list[j]);
```
ashleypittman, line 3460: Does this work?  At this point if the environment has been cleared then this might fail?  In which case you'd need to sample the value in init() and keep it around to use it here.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,282 @@ new_readdir(DIR *dirp)
...
+			if (rc != -DER_NONEXIST && rc != -DER_SUCCESS) {
+				printf("Error: failed to allocate memory in querying env %s.\n",
+				       env_list[j]);
+				goto err_out2;
+			}
+			rc = asprintf(&env_buf_list[j], "%s=%s", env_list[j], env_value);
```
ashleypittman, line 3466: This looks incorrect for the -DER_NONEXIST case.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,282 @@ new_readdir(DIR *dirp)
...
- *		init_dfs();
- *		return pid;
- *	}
- *}
+extern char **__environ;
+
+/* This is the number of environmental variables that would be forced to set in child process.
+ * "LD_PRELOAD" is a special case and it is not included in the list.
+ */
+#define LEN_ENV_LIST (10)
```
ashleypittman, line 3343: You could now use ARRAY_SIZE(env_list) for this so to avoid having to update it.  You'd need to make env_list a static global or this a local variable though.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,282 @@ new_readdir(DIR *dirp)
...
+	char  *pil4df_path;
+	char  *new_preload_str;
+	int    num_entry                  = 0;
+	int    num_entry_found            = 0;
+	int    idx_preload                = -1;
+	bool   preload_included           = false;
+	bool   pil4dfs_in_preload         = false;
+
+	/* "LD_PRELOAD" is a special case which needs .*/
+	char   env_list[LEN_ENV_LIST][32] = {"D_IL_REPORT", "DAOS_MOUNT_POINT", "DAOS_POOL",
```
ashleypittman, line 3369: Is the 32 or LEN_ENV_LIST required here?  As you're populating this list you can probably let the compiler do the work here.  See https://github.com/daos-stack/daos/blob/master/src/cart/crt_init.c#L21 for now to do this dynamically without using constants.
```suggestion
	char   *env_list[] = {"D_IL_REPORT", "DAOS_MOUNT_POINT", "DAOS_POOL",
```

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: I think this should work. unsetenv() is not called in the case of scons, so env variables should be there.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! Fixed as you suggested.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Fixed as you suggested. Thank you very much!

wiliamhuang:
The failures in Functional Hardware Medium Verbs Provider are not related.
@ashleypittman @mchaarawi Could you please review?

daltonbohning (commented):
ftest code changes LGTM

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,284 @@ new_readdir(DIR *dirp)
...
- *	pid = real_fork();
- *	if (pid) {
- *		// parent process: do nothing
- *		return pid;
- *	} else {
- *		init_dfs();
- *		return pid;
- *	}
- *}
+extern char **__environ;
```
ashleypittman, line 3338: If I've read the man page correctly then this is also available as environ and available through a standard header file.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,284 @@ new_readdir(DIR *dirp)
...
+static char  *env_list[] = {"D_IL_REPORT", "DAOS_MOUNT_POINT", "DAOS_POOL", "DAOS_CONTAINER",
+			    "D_IL_MAX_EQ", "D_LOG_FILE", "D_IL_ENFORCE_EXEC_ENV",  "DD_MASK",
+			    "DD_SUBSYS", "D_LOG_MASK"};
+
+/* Environmental variables could be cleared in some applications. To make sure all libpil4dfs
+ * related env properly set, we intercept execve and its variants to check envp[] and append our
+ * env when needed.
+ * pre_envp() scans provided envp to look for libpil4dfs related envs (env_list[]) and returns a
+ * new envp with missing libpil4dfs related envs appended.
+ * The fd for daos logging may be not valid here, so printf() is used in this function.
```
ashleypittman, line 3352: This is a callback between for and exec, I'm not sure printf() would always work here but for errors then fprintf() to stderr would be better.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,284 @@ new_readdir(DIR *dirp)
...
+				preload_included = true;
+				idx_preload      = num_entry;
+				num_entry_found++;
+				if (strstr(envp[num_entry], "libpil4dfs.so"))
+					pil4dfs_in_preload = true;
+			}
+			/* The list of env is not too long. We use a simple loop to lookup for
+			 * simplicity. This function is not performance critical.
+			 */
+			for (i = 0; i < ARRAY_SIZE(env_list); i++) {
```
ashleypittman, line 3393: There's a case here where this code will be invoked but not required to.  If LD_PRELOAD is being used but one of the env vars you're testing for is not set then this will be detected and the code will start to modify the env, but if the env isn't set then it won't be able to be copied so in effect this will make a copy but then not modify it.

To catch this case then I think this while code should loop and call d_agetenv_str() first for each entry and then only search the list for ones which are set in the parent.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: It seems likely that this will work for programs that call evecve without the env set but not for functions that call unsetenv and then execv, this is something we should test for and if necessary note.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,284 @@ new_readdir(DIR *dirp)
...
+	reset_daos_env_before_exec();
+
+	if (!enforce_exec_env)
+		return next_execv(filename, argv);
+
+	if (next_execve == NULL) {
+		next_execve = dlsym(RTLD_NEXT, "execve");
+		D_ASSERT(next_execve != NULL);
+	}
+	return next_execve(filename, argv, pre_envp(__environ));
```
ashleypittman, line 3574: I still think this is unlikely to work, if a env var has been removed then d_agetenv_str() on it would fail.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5673,13 +5913,19 @@ init_myhook(void)
...
 
-	d_agetenv_str(&env_log, "D_IL_REPORT");
+	rc = d_agetenv_str(&env_log, "D_IL_REPORT");
 	if (env_log) {
 		report = true;
 		if (strncmp(env_log, "0", 2) == 0 || strncasecmp(env_log, "false", 6) == 0)
 			report = false;
 		d_freeenv_str(&env_log);
+	} else if (rc == -DER_NONEXIST) {
+		d_setenv("D_IL_REPORT", "0", 1);
```
ashleypittman, line 5923: I don't think you should be setting these.  I assume the intention is to save them for if required later in exec but that has two problems:

1. They might have been stripped from the environment
2. When running pil4dfs then the child and grandchild will have different environment

Assuming the intention here is to save them for later then use a static global variable for this.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,284 @@ new_readdir(DIR *dirp)
...
+extern char **__environ;
+
+/* This is the number of environmental variables that would be forced to set in child process.
+ * "LD_PRELOAD" is a special case and it is not included in the list.
  */
+static char  *env_list[] = {"D_IL_REPORT", "DAOS_MOUNT_POINT", "DAOS_POOL", "DAOS_CONTAINER",
+			    "D_IL_MAX_EQ", "D_LOG_FILE", "D_IL_ENFORCE_EXEC_ENV",  "DD_MASK",
+			    "DD_SUBSYS", "D_LOG_MASK"};
+
+/* Environmental variables could be cleared in some applications. To make sure all libpil4dfs
```
ashleypittman, line 3347: I'm trying to understand the error handling here, it seems this function is unlikely to fail, however if it does then it would do it in a way that it does nothing which were it to cause problems would be very difficult to debug so this should probably return an error and then propagate to the user.  I would probably just remove the prints about memory allocation failure as well.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: You are right.
"/usr/include/unistd.h" does have "extern char **__environ;" and "char **environ". I will fix this. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. I will use fprintf(stderr, ). Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: I believe applications call unsetenv() to clear only common system env or the env they set. I only observed "MAKEFLAGS" cleared by unsetenv() two times during daos building with scons.
We can intercept unsetenv(env) and ignore the operation if env is in our list and D_IL_ENFORCE_EXEC_ENV == 1.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. I will add the interception of unsetenv() then. Thank you!

ashleypittman: It would be simpler and more reliable to save the env vars that might be required at the time that the parent process itself is initialized.

wiliamhuang: Saving the value at initialization might cause inconsistency in application like bash. If "export" is used to update env in bash script, the saved value would be out of dated. Using env does not have such issue.

ashleypittman: That's true, but if that's the case then it would be set and the saved value wouldn't be needed.

getenv() probably just checks the environ list so having code which checks environ and adds anything that's missing from environ isn't going to achieve much.

ashleypittman:
FYI - this has a conflict now with some other changes I just made to the dfuse_test.

wiliamhuang:
> FYI - this has a conflict now with some other changes I just made to the dfuse_test.

Thank you for letting me know! I will fix them when I refresh my PR.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: It would be more reasonable to let the lib/application set default value if an env is not set instead of using a saved value at the initialization phase.

Assuming env_var's default value is 0.
bash runs a bash script with env_var=1.
"unset env_var" is invoked inside bash script to clear env on purpose.
"env_var=1" will still be used in following new processes.

Relying on getenv would be more consistent to me.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: same as the above comment. 

It would lead to better consistency to let the lib/application set default value if an env is not set instead of using a saved value at the initialization phase.

Assuming env_var's default value is 0.
bash runs a bash script with env_var=1.
"unset env_var" is invoked inside bash script to clear env on purpose.
"env_var=1" will still be used in following new processes.

For the case of creating a child process with stripped env, parent process does not unset env in environ. Instead, the parent process just passes a new env list in exec().

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! Revised as you suggested.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you very much! Revised as you suggested.

wiliamhuang:
@ashleypittman The failures in CI are not relevant. Could you please review and provide your feedback?
We only append env if the provided env list is different from environ. I think relying on getenv would achieve better consistency especially for bash.
When a parent process creates a child process with a stripped env, it uses an env list different from environ. If a process updates env, we should honor such changes.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,350 @@ new_readdir(DIR *dirp)
...
+	/* 0 means no error. */
+	*error = 0;
+	/* simply return for environ. append pil4dfs env only for stripped env list. */
+	if (envp == environ)
+		return (char **)envp;
+
+	/* check D_IL_ENFORCE_EXEC_ENV env in case it was updated. */
+	rc = d_getenv_bool("D_IL_ENFORCE_EXEC_ENV", &enforce_exec_env);
+	if (rc == -DER_NONEXIST || enforce_exec_env == false)
+		return (char **)envp;
```
ashleypittman, line 3387: If this environment has been cleaned then this won't be set so the code would never be invoked, at least for the exec case.

You also don't want to read the same environment variable more than once so can you read this at the time daos_init() is called or elsewhere where the library is setup and save the result for use here.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,350 @@ new_readdir(DIR *dirp)
...
+
+	/* the number of env in env_list[] that are set in environ */
+	num_env_append = 0;
+
+	/* check whether "LD_PRELOAD" exists in environ */
+	rc = d_agetenv_str(&env_value, "LD_PRELOAD");
+	if (rc == -DER_NONEXIST) {
+		return (char **)envp;
```
ashleypittman, line 3395: So if `LD_PRELOAD` isn't set then this code does not set it?  Isn't adding this if it's missing the point of this PR?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,350 @@ new_readdir(DIR *dirp)
...
+	} else if (rc == -DER_NOMEM) {
+		*error = ENOMEM;
+		goto err_out0;
+	}
+	if (strstr(env_value, "libpil4dfs.so")) {
+		pil4dfs_set_preload = true;
+		num_env_append++;
+	}
+	free(env_value);
```
ashleypittman, line 3404: I'm not completely up to date here but should this use `d_freeenv_str`?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -2011,6 +2027,14 @@ open_common(int (*real_open)(const char *pathname, int oflags, ...), const char
 	file_list[idx_fd]->open_flag   = oflags;
 	/* NEED to set at the end of file if O_APPEND!!!!!!!! */
 	file_list[idx_fd]->offset = 0;
+	if (strncmp(full_path, "/", 2) == 0)
```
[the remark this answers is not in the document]
wiliamhuang: Yes. I though about using "STR_AND_SIZE" previously and chose not changing it. This string is short and current form is easier to read.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Flag enforce_exec_env is set in the constructor function by getenv. We do not monitor whether this env is updated or not. In the case of running a bash script, env "D_IL_ENFORCE_EXEC_ENV" could be updated inside the script. Checking it again will make sure we do get an updated value.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: If "LD_PRELOAD" is unset inside a bash script, we should honor it.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: You are right. I will replace "free" with "d_freeenv_str". Thank you!

wiliamhuang: Fixed as you suggested. Thank you!

ashleypittman:
@wiliamhuang I had missed the changes in https://github.com/daos-stack/daos/pull/13519/commits/ebf7a5b944877bdcd141cdfc50cd954e408a4670 so that this PR no longer modifies `execv` but only `execve` variants.

The assumption now appears to be that the process still has everything set in the environment but that if it's using `execve` then this should add back in things from the environment.

I'll re-review with this in mind.

wiliamhuang:
> @wiliamhuang I had missed the changes in [ebf7a5b](https://github.com/daos-stack/daos/commit/ebf7a5b944877bdcd141cdfc50cd954e408a4670) so that this PR no longer modifies `execv` but only `execve` variants.
> 
> The assumption now appears to be that the process still has everything set in the environment but that if it's using `execve` then this should add back in things from the environment.
> 
> I'll re-review with this in mind.

Right. We only append env for execve, execvpe, and fexecve where an env list different from environ can be passed in. If envs in environ updated (modified/removed/add), we do nothing since we assume that are done on purpose.

wiliamhuang:
The failures in hardware medium verb providers are not relevant. I can rerun the hardware tests.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,349 @@ new_readdir(DIR *dirp)
...
+			fprintf(stderr, "Error: env for LD_PRELOAD is too long.\n");
+			rc = E2BIG;
+			goto err_out1;
+		}
+		rc = asprintf(&new_preload_str, "%s:%s", envp[idx_preload], pil4df_path);
+		if (rc < 0) {
+			rc = ENOMEM;
+			goto err_out1;
+		}
+		(*new_envp)[idx_preload] = new_preload_str;
```
ashleypittman, line 3495: Do you need to free the old string?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,349 @@ new_readdir(DIR *dirp)
...
+			num_entry++;
+		}
+	}
+
+	/* All required env are found and pil4dfs is in LD_PRELOAD. No need to create a new envp. */
+	if (num_entry_found == num_env_append && pil4dfs_in_preload == true)
+		return 0;
+
+	/* the new envp holds the existing envs & the envs forced to append plus NULL at the end */
+	*new_envp = malloc(sizeof(char *) * (num_entry + num_env_append + 1));
```
ashleypittman, line 3463: calloc would be better here as it would be zeroed memory.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -3310,70 +3335,349 @@ new_readdir(DIR *dirp)
...
+			}
+			(*new_envp)[i] = env_buf_list[j];
+			i++;
+			/* free the buffer allocated by d_agetenv_str() */
+			d_freeenv_str(&env_value);
+		}
+	}
+
+	/* append NULL pointer to the end of new_envp */
+	(*new_envp)[i] = NULL;
```
ashleypittman, line 3536: using calloc above would avoid the need for this.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: The above discussion was in the context of a PR where execv was being modified but now the PR only modified execve.

I still don't see any case for setting this here.  If a user has set it and it's propagated then fine, but why would a process turn it on for it's child processes?

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: I prefer not freeing the old string. The string could be static.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Revised as you suggested. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Revised as you suggested. Thank you!

wiliamhuang:
@ashleypittman I removed unneeded setenv() for "D_IL_REPORT" and "D_IL_ENFORCE_EXEC_ENV". The default value will be set in case env is not found. Thank you!

ashleypittman:
> @ashleypittman I removed unneeded setenv() for "D_IL_REPORT" and "D_IL_ENFORCE_EXEC_ENV". The default value will be set in case env is not found. Thank you!

I think the same applies to `D_IL_MAX_EQ` as well.

wiliamhuang:
> > @ashleypittman I removed unneeded setenv() for "D_IL_REPORT" and "D_IL_ENFORCE_EXEC_ENV". The default value will be set in case env is not found. Thank you!
> 
> I think the same applies to `D_IL_MAX_EQ` as well.

Thank you! Just fixed it.

ashleypittman (approved):

mchaarawi (approved):

wiliamhuang:
The failures of "Functional Hardware Medium Verbs Provider " are not relevant to libpil4dfs.

mchaarawi:
> The failures of "Functional Hardware Medium Verbs Provider " are not relevant to libpil4dfs.

what is the ticket number associated with the HW medium failure?

PR #13532, opened 2023-12-21
DAOS-14826 client: reserve low fds

wiliamhuang:
Features: pil4dfs

Reserve low fd to avoid daos network contexts use them.
Fix a corner case in fstatat() to mimic libc and add unit test.
Rely on libc dup2() to get desired fd to avoid error "failed to get the desired fd in dup2()".
Use atomic APIs to access daos_inited.
Move pthread_atfork() inside locked region to avoid multiple execution.
Call glibc chdir() to update process's cwd in kernel state.
Introduce fd duplication reference count to handle complex situation

Add a unit test to make sure daos does not use low fd.

### Gatekeeper:


src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1007,6 +1010,19 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
 		/* trying to avoid lock as much as possible */
 		if (!daos_inited) {
```
ashleypittman, line 1011: Looking at this existing code it doesn't appear thread-safe in the case of two threads entering this code-path concurrently, not a regression but it would be good to add locking to this at the same time.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1007,6 +1010,19 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
...
 			/* daos_init() is expensive to call. We call it only when necessary. */
+			int low_fd_list[DAOS_MIN_FD], low_fd_count = 0, fd_kernel, idx;
+
+			fd_kernel = open("/proc/self/maps", O_RDONLY);
```
ashleypittman, line 1015: What's wrong with opening "/"? 

Presumably this code can only get here after the hooks have been installed so perhaps it would be better to call the redirected open directly?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1024,6 +1040,8 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
 
 			daos_inited = true;
 			atomic_fetch_add_relaxed(&daos_init_cnt, 1);
+			for (idx = 0; idx < low_fd_count; idx++)
+				close(low_fd_list[idx]);
```
ashleypittman, line 1044: There are lots of error paths where these files would not be closed, you need to add extra error handling to this function.  Overall this function is large and would probably benefit from being split up.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -2540,6 +2558,10 @@ new_fstatat(int dirfd, const char *__restrict path, struct stat *__restrict stat
 			return new_xstat(1, path, stat_buf);
 	}
 
+	if (dirfd >= FD_FILE_BASE && dirfd < FD_DIR_BASE && path[0] == 0)
+		/* same as fstat for a file. May need further work to handle flags */
+		return fstat(dirfd, stat_buf);
+
```
ashleypittman, line 2564: Is this related to the commit message?  Same applies for the rest of the code changes in this file.

ashleypittman (commented):
Didn't you say you had a working test for the low fd issue?  It would be good to see that included alongside the code.

wiliamhuang:
> Didn't you say you had a working test for the low fd issue? It would be good to see that included alongside the code.

Yes. The test bash script needs some code in exec interception to pass. I will put the test case over there. fd 0, 1, and 2 are preserved after exec(). We need to handle such issue in https://github.com/daos-stack/daos/pull/13519. We also probably need to close all fds higher than 2.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Yes. We were aware of this possible issue. daos_init() uses lock already. query_path() is called in many intercepted functions. We hoped to avoid lock if possible.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: This is a corner case of fstatat(). dirfd is allowed to be a fd of a regular file.
Yes. There are probably a few places where we need to handle flags. I added a comment here to make sure I will not forget it in future.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! Added a wrapper function as you suggested.

ashleypittman:
> > Didn't you say you had a working test for the low fd issue? It would be good to see that included alongside the code.
> 
> Yes. The test bash script needs some code in exec interception to pass. I will put the test case over there. fd 0, 1, and 2 are preserved after exec(). We need to handle such issue in #13519. We also probably need to close all fds higher than 2.

That link and some of the code in this PR is for the execv issue, but most of the code in this PR and the commit message relate to the low FD issue.  Do you have a test for the low FD issue and can you include it in this PR.

wiliamhuang:
> That link and some of the code in this PR is for the execv issue, but most of the code in this PR and the commit message relate to the low FD issue. Do you have a test for the low FD issue and can you include it in this PR.

Just added a unit test to make sure daos does not use low fd. Thank you!

wiliamhuang:
@ashleypittman @mlawsonca Could you please review? Thank you very much!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: In case daos_init() is called more than one time, daos_init_cnt has the count. daos_fini() will be called daos_init_cnt times in destruction phase.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5565,6 +5676,20 @@ init_myhook(void)
 		eq_count_max = MAX_EQ;
 	}
 
+	/* D_IL_DAOS_MIN_FD=3 or smaller value will disable blocking low fds */
+	rc = d_getenv_uint64_t("D_IL_DAOS_MIN_FD", &daos_min_fd);
```
mchaarawi, line 5680: is this env documented somewhere?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1007,23 +1075,41 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
...
 				goto out_normal;
 			}
+
 			if (eq_count_max) {
-				rc = daos_eq_create(&td_eqh);
-				if (rc)
-					DL_WARN(rc, "daos_eq_create() failed");
-				main_eqh = td_eqh;
-				rc       = pthread_atfork(NULL, NULL, &child_hdlr);
+				D_MUTEX_LOCK(&lock_eqh);
```
mchaarawi, line 1097: i don't understand why we need nested locks now

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1007,23 +1075,41 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
 		/* trying to avoid lock as much as possible */
 		if (!daos_inited) {
 			/* daos_init() is expensive to call. We call it only when necessary. */
+
+			D_MUTEX_LOCK(&lock_global);
```
mchaarawi, line 1079: why is this needed for the entire block? isn't this just needed internally for consume_low_fd()?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -2540,6 +2626,10 @@ new_fstatat(int dirfd, const char *__restrict path, struct stat *__restrict stat
 			return new_xstat(1, path, stat_buf);
 	}
 
+	if (dirfd >= FD_FILE_BASE && dirfd < FD_DIR_BASE && path[0] == 0)
```
mchaarawi, line 2629: im not sure i understand what you are doing here, nor i understand this commit message:
"Also fix a corner case in fstatat() to mimic libc. Rely on libc dup2() to get desired fd to avoid error
"failed to get the desired fd in dup2()"."

what is the corner case being fixed? can you be more clear. if this corner case is fixed, can you add a test for the corner case? thanks.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: It is not in our document. Normally users do not need to set it. Default value 10 will be used.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: dirfd is allowed to be a fd of a regular file. I did not expect such situation. ok. I will add a unit test for this corner case.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. I will rearrange the code to avoid nested locks. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: It looks like you've re-worked this code since yesterday but it's still not thread safe although as you daos_fini() does look like it'll be called the correct number of times.

pthread_atfork() wants to be called precisely once, and if you're using the daos_inited boolean in this manner then you need to access it atomically.

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: ok well if users don't need to set it then i do not see why it needs to be an env variable..
if it needs to, then we should document it in the IL readme, and also consider changing it's name D_ and DAOS are repetitive.
maybe call it DAOS_RESERVE_MIN_FD? other suggestion welcome

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -78,6 +78,18 @@
...
 
+/* the default min fd that will be used by DAOS */
+#define DAOS_MIN_FD         10
+/* an upper limit is set in case that env was not set properly */
+#define MAX_DAOS_MIN_FD     128
+
+/* the number of low fd blocked */
+static uint16_t               low_fd_count;
+/* upper limit of low fd that will be blocked */
+static uint16_t               max_low_fd_blocked;
```
mchaarawi, line 89: is it blocked or reserved?

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: You are right. I will move the lock inside consume_low_fd(). Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Maybe "reserved" is a little better.

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: well my point was just that if you already added a lock on top (not sure if that one is needed), do you need this one too?

mchaarawi: actually nevemind that comment was for the next lock not the lock_global

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. We can just use a hard-coded value for now. We add an env later if necessary. Thank you!

mchaarawi: sorry, I am not against making this an env variable. im just saying that an env variable needs to be documented somewhere if we need to ask users to play with it. otherwise only the dev (you) knows about it and what it does.

wiliamhuang: Thank you! Understood. We can use hardcoded value to make our code simpler and decrease the number of env users need to deal with. Originally, I thought bash script can use arbitrary fd. An env may be helpful if a larger value is needed. However, it could be very hard for a regular user to figure out the situation. I will add a warning in dup2() if a fd pointing to "socket:[xxxx]" or "anon_inode:[eventpoll]" is closed, so it would help for debugging related issues.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: I am trying to add the corner case in dfuse unit test. Looks libioil has the same issue.

wiliamhuang: Sorry. False alarm. Used wrong flag in fstatat().

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! Moved pthread_atfork() into the locked region. Could you please check again?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -241,6 +249,7 @@ struct sigaction       old_segv;
 /* the flag to indicate whether initlization is finished or not */
 static bool            hook_enabled;
 static bool            hook_enabled_bak;
+static pthread_mutex_t lock_global;
```
mchaarawi, line 252: not a big deal, but if a repush is needed, better to rename that to lock_reserve_fd

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -4818,6 +4903,35 @@ futimens(int fd, const struct timespec times[2])
...
+		return;
+	snprintf(fd_path, sizeof(fd_path) - 1, "/proc/self/fd/%d", fd);
+	len = readlink(fd_path, path, DFS_MAX_PATH - 1);
+	if (len < 0) {
+		DS_ERROR(errno, "readlink() failed");
+		goto out;
+	}
+	path[len] = 0;
+	if (strncmp(path, "socket:", 7) == 0 || strncmp(path, "anon_inode:[eventpoll]", 22) == 0) {
+		D_WARN("dup2/fcntl is closing fd %d (%s) which may be used by DAOS!", fd, path);
```
mchaarawi, line 4928: i guess i don't understand what a warning is supposed to do. how can we end up in this situation?

mchaarawi (approved):

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: This message is mainly for debugging. Not sure whether this is always related to real issues, especially for an application uses network. Shall I use D_DEBUG() instead?
If a bash script uses larger fd (e.g., 15) directly, this piece of code will be triggered.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Thank you! I will fix it if I need a repush.

src/client/dfuse/pil4dfs/int_dfs.c
ashleypittman: For what it's worth the ioil code is at https://github.com/daos-stack/daos/blob/master/src/client/dfuse/il/int_posix.c#L741-L797

It just uses a hard-coded value of 10 and does no checking.  It also uses dup for all but the first fd and does it all in one function rather than using globals.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -913,6 +922,55 @@ child_hdlr(void)
...
+static int
+consume_low_fd(void)
+{
+	int rc = 0;
+
+	if (atomic_load_relaxed(&daos_inited) == true)
+		return 0;
+
+	low_fd_count              = 0;
+	low_fd_list[low_fd_count] = libc_open("/", O_RDONLY);
```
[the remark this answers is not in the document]
wiliamhuang: Yes. "O_PATH|O_DIRECTORY" works too. I can use it to match the code in other places. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -4989,15 +5106,21 @@ dup2(int oldfd, int newfd)
...
+
 	if (oldfd == newfd) {
 		if (oldfd < FD_FILE_BASE)
 			return next_dup2(oldfd, newfd);
 		else
 			return newfd;
 	}
-	if ((oldfd < FD_FILE_BASE) && (newfd < FD_FILE_BASE))
+	oldfd_directed = query_fd_forward_dest(oldfd);
+	if ((oldfd_directed < FD_FILE_BASE) && (oldfd < FD_FILE_BASE) && (newfd < FD_FILE_BASE))
```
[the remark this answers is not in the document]
wiliamhuang: This change is not related with low fd. I found some warning messages when running ftest/dfuse/bash.py and found linpil4dfs failed in a corner case where fd points to a regular file and path is empty.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -4818,6 +4903,35 @@ futimens(int fd, const struct timespec times[2])
 	return 0;
 }
 
+/* check the fd to be closed. If the fd pointing to something like
+ * 'socket:[xxxx]' or 'anon_inode:[eventpoll]', print out a warning
+ * since DAOS network contexts may be using such fds.
+ */
+static void
+check_fd_to_close(int fd)
+{
```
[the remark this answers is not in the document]
wiliamhuang: ok. I can remove this check for now. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -913,6 +922,55 @@ child_hdlr(void)
 	context_reset = true;
 }
 
+/* only free the reserved low fds when application exits or encounters error */
+static void
+free_reserved_low_fd(void)
```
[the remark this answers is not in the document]
wiliamhuang: fixed as you suggested. Thank you!

wiliamhuang:
@mchaarawi @ashleypittman Just updated this PR to fix the failure of unit test in CI. Could you please review? We need to land this one very soon, so we can get the PR for exec() interception ready quickly. Thank you very much!

wiliamhuang:
All tests in CI passed. @ashleypittman @mchaarawi Could you please review? Thank you!

mchaarawi (approved):

wiliamhuang:
The failed case is not related. Sounds like ticket, https://daosio.atlassian.net/browse/DAOS-14884

@ashleypittman @mchaarawi Master landing is open now. Could you please review? We need to land this one soon. Thank you very much!

mchaarawi (approved):

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -913,6 +922,59 @@ child_hdlr(void)
...
+		low_fd_list[low_fd_count] = libc_open("/", O_RDONLY);
+	}
+
+	D_MUTEX_UNLOCK(&lock_reserve_fd);
+	return rc;
+
+err:
+	rc = errno;
+	free_reserved_low_fd();
+	low_fd_count = 0;
```
ashleypittman, line 972: Not required now.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -4818,6 +4915,39 @@ futimens(int fd, const struct timespec times[2])
 	return 0;
 }
 
+/* check the fd to be closed. If the fd pointing to something like
```
ashleypittman, line 4918: Please delete this before landing.

src/client/dfuse/pil4dfs/int_dfs.c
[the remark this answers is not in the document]
ashleypittman: Should that test be in this PR or does this PR depend on the exec PR then?

src/tests/ftest/daos_test/dfuse.py
```diff
@@ -93,6 +93,8 @@ def run_test(self, il_lib=None):
             daos_test_env['D_LOG_MASK'] = 'INFO,IL=DEBUG'
 
         command = [self.daos_test, '--test-dir', mount_dir, '--io', '--stream']
+        if use_dfuse:
+            command.append('--lowfd')
```
ashleypittman, line 97: Is there a reason for only enabling this when dfuse is enabled?  The purpose of this test is to ensure dfuse or interception behave the same as natively, does this test fail when run on a native filesystem?

ashleypittman (changes requested):
This PR seems to correctly protect file descriptors used by daos from being accessed incorrectly, however it does not allow bash scripts to write to fds backed by dfs at all so whilst this may prevent a segfault in the daos library the operations that we're trying to support still do not work.

I've written a test-case this morning that loosely resembles what the configure script is doing, I'll push that as a ftest in a PR to explain what I mean, but in essence bash is managing to open a fd and reporting success for that, but then getting errors when writing to it.

src/client/dfuse/pil4dfs/int_dfs.c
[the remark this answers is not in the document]
wiliamhuang: Right. The above is added in the exec PR. It requires the new code there.

src/tests/ftest/daos_test/dfuse.py
wiliamhuang: This test fails under a native filesystem. consume_low_fd() and daos_init() are called on the first time accessing file/dir on DFS. They are not triggered when only native filesystem is used.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Could you please provide more information? Thank you!

ashleypittman: the free_reserved_low_fd() function sets it.

mchaarawi (approved):

wiliamhuang:
@ashleypittman I just made the changes you suggested.
Functional and hardware tests are skipped since we only removed some commented code and one line duplicated code.

ashleypittman:
> @ashleypittman I just made the changes you suggested. Functional and hardware tests are skipped since we only removed some commented code and one line duplicated code.

See my previous review and the code I put into the ticket, I don't think this PR solves the problem it's aiming to.

wiliamhuang:
> > @ashleypittman I just made the changes you suggested. Functional and hardware tests are skipped since we only removed some commented code and one line duplicated code.
> 
> See my previous review and the code I put into the ticket, I don't think this PR solves the problem it's aiming to.

Can we land this PR and work on your new test later? I am working on daos compile with pil4dfs compatible mode. "configuration" can pass in my latest code.

mchaarawi:
> > @ashleypittman I just made the changes you suggested. Functional and hardware tests are skipped since we only removed some commented code and one line duplicated code.
> 
> See my previous review and the code I put into the ticket, I don't think this PR solves the problem it's aiming to.

there are multiple issues for the config, and afaiu this resolves one of them. we asked Lei to break PRs into smaller ones, so this resolves one of them, no?
I think you need to be a little more clear with your reviews as to why this PR needs a change and what is the change requested.

ashleypittman:
As I put in my review comment earlier I don't believe this fixes the issue stated in the ticket.  I've taken the time to write a reproducer which I had said I'd push in a PR and I still intend to do that but I have been pulled away by other priorities today, in lieu of pushing a PR with the test in I've attached the test to the ticket this PR is linked to.

It is possible that for the use-case in the ticket this PR changes the behaviour from a crash or hang into either pil4dfs returning an error (which would potentially be OK) or misdirected writes (which would not be OK), if you want me to review the code on that basis then I can do that but the description in this PR says this resolves the issue and my testing which I have shared with you demonstrates that it does not and therefore to me this PR requires more work.

mchaarawi:
> As I put in my review comment earlier I don't believe this fixes the issue stated in the ticket. I've taken the time to write a reproducer which I had said I'd push in a PR and I still intend to do that but I have been pulled away by other priorities today, in lieu of pushing a PR with the test in I've attached the test to the ticket this PR is linked to.
> 
> It is possible that for the use-case in the ticket this PR changes the behaviour from a crash or hang into either pil4dfs returning an error (which would potentially be OK) or misdirected writes (which would not be OK), if you want me to review the code on that basis then I can do that but the description in this PR says this resolves the issue and my testing which I have shared with you demonstrates that it does not and therefore to me this PR requires more work.

I do appreciate writing that reproducer. but my understanding is that this reproducer requires also the second PR https://github.com/daos-stack/daos/pull/13519 which is still in progress. or am i wrong in my assumption?
I am trying to avoid having giant PRs and have smaller ones to make reviews easier.

wiliamhuang:
@ashleypittman With the latest commit, https://github.com/daos-stack/daos/pull/13532/commits/b1bf1e99a8d4fd633092f73bb6944607420c874c, your bash test seems showing same behavior with and without libpil4dfs. Please try it and let me know. Thank you!

ashleypittman:
> @ashleypittman With the latest commit, [b1bf1e9](https://github.com/daos-stack/daos/commit/b1bf1e99a8d4fd633092f73bb6944607420c874c), your bash test seems showing same behavior with and without libpil4dfs. Please try it and let me know. Thank you!

I can confirm that my test now runs and passes, It looks like the remaining issues were in the dup() code from your change?

There is one error I'm getting from pil4dfs during this operation which you should look at but once that's resolved then I think this PR might be complete.

> src/client/dfuse/pil4dfs/int_dfs.c:1525 free_fd() dfs_release() failed: 22 (Invalid argument)

There are some IO requests to dfuse during the test, I assume this is down to the getcwd handling although I also see two opens and two reads - we can deal with that elsewhere though.

wiliamhuang:
> I can confirm that my test now runs and passes, It looks like the remaining issues were in the dup() code from your change?

Yes. One situation in dup2() was not implemented previously. Also glibc chdir() is needed to update the current working directory before exec(), so we always call glibc chdir inside chdir() and fchdir() now.


> There is one error I'm getting from pil4dfs during this operation which you should look at but once that's resolved then I think this PR might be complete.
> 
> > src/client/dfuse/pil4dfs/int_dfs.c:1525 free_fd() dfs_release() failed: 22 (Invalid argument)

Thank you! I will look into this issue.

> There are some IO requests to dfuse during the test, I assume this is down to the getcwd handling although I also see two opens and two reads - we can deal with that elsewhere though.

chdir()/fchdir() will be passed through to dfuse. exec() also needs dfuse.

wiliamhuang:
> src/client/dfuse/pil4dfs/int_dfs.c:1525 free_fd() dfs_release() failed: 22 (Invalid argument)

I just pushed a commit to introduce fd duplication reference count to handle complex situations. This issue is resolved in my local test. Could you please give it a try? Thank you!

ashleypittman (approved):

wiliamhuang:
@mchaarawi Could you please review the latest changes? Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1913,14 +1919,15 @@ open_common(int (*real_open)(const char *pathname, int oflags, ...), const char
...
 		mode_query = S_IFREG;
 	} else if (!parent && (strncmp(item_name, "/", 2) == 0)) {
-		rc = dfs_lookup(dfs_mt->dfs, "/", oflags, &dfs_obj, &mode_query, NULL);
+		rc =
+		    dfs_lookup(dfs_mt->dfs, "/", oflags & (~O_APPEND), &dfs_obj, &mode_query, NULL);
 	} else {
-		rc = dfs_lookup_rel(dfs_mt->dfs, parent, item_name, oflags, &dfs_obj, &mode_query,
-				    NULL);
+		rc = dfs_lookup_rel(dfs_mt->dfs, parent, item_name, oflags & (~O_APPEND), &dfs_obj,
+				    &mode_query, NULL);
```
mchaarawi, line 1930: some comments need to be added here as to how O_APPEND is being handled.

mchaarawi (approved):
my comment can be addressed in the next PR.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. I will add comments in my another PR. Thank you!

wiliamhuang:
The failure in hardware tests are not related to libpil4dfs.

PR #13533, opened 2023-12-21
Add a check for Jenkins results.

ashleypittman:
Skip-build: true

Required-githooks: true


ashleypittman:
test

ashleypittman:
Test 2

ci/jenkins_status.py
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env python
```
daltonbohning, line 1: This is probably okay, but some distros point `python -> python2` so personally I tend to use `python3` here

ci/jenkins_status.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def __init__(self, data, blocks):
+        name = data["name"]
+        if "-" in name:
+            try:
+                (num, full) = name.split("-", 1)
+                test_num = int(num)
+                if full[-5] == "-":
+                    full = full[:-5]
+                name = f"{full} ({test_num})"
```
daltonbohning, line 25: What does the original `data["name"]` look like?

ci/jenkins_status.py
```diff
@@ -0,0 +1,212 @@
...
+            except ValueError:
+                pass
+        self.name = name
+        self.cname = data["className"]
+        self.skipped = False
+        self.passed = False
+        self.failed = False
+        assert data["status"] in ("PASSED", "FIXED", "SKIPPED", "FAILED", "REGRESSION")
```
daltonbohning, line 33: Maybe better to gracefully handle?

ci/jenkins_status.py
ashleypittman: I hope we're beyond that now!

ci/jenkins_status.py
ashleypittman: this is the weird 8 digits of sha hash at the end of the PR name which is making matching fail.

ci/jenkins_status.py
ashleypittman: For an internal script I'm OK with this, if there is another status then we want to know about it so we handle it properly.  As we're scraping Jenkins then we can stress-test this code against previous builds.

ci/jenkins_status.py
```diff
@@ -0,0 +1,212 @@
...
+    for build in data["builds"]:
+        jid = build["number"]
+        if jid > lcb:
+            print(f"Job {jid} is of {job_name} is still running, skipping")
+            continue
+        failed = show_job(job_name, jid)
+        if not isinstance(failed, list):
+            continue
+        for test in failed:
+            all_failed.add(test)
```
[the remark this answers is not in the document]
ashleypittman: all_failed is a set, not list.

ci/jenkins_status.py
```diff
@@ -0,0 +1,212 @@
...
+        jdata = je_load(job_name, jid=jid, what="testReport")
+    except urllib.error.HTTPError:
+        print(f"Job {jid} of {job_name} has no test results")
+        return None
+
+    print(f"Checking job {jid} of {job_name}")
+
+    failed = []
+
+    assert not jdata["testActions"]
```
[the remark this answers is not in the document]
ashleypittman: I could just remove this, it was me trying to understand the output.

ci/jenkins_status.py
daltonbohning: Hopefully, but some older (still supported) distros still use python2 by default last I checked. Also, some distros don't symlink `python` at all. E.g.
```
$ cat /etc/os-release
NAME="SLES"
VERSION="15-SP4"
VERSION_ID="15.4"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP4"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp4"
DOCUMENTATION_URL="https://documentation.suse.com/"

$ which python
which: no python in ...

$ which python3
/usr/bin/python3
```

ci/jenkins_status.py
daltonbohning: The problem with `assert` is that the entire script will fail without much debug info.

ci/jenkins_status.py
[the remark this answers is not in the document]
daltonbohning: Ah, then JFYI
```
>>> set1 = set([1,2,3])
>>> list2 = [4,5,6]
>>> set1.update(list2)
>>> set1
{1, 2, 3, 4, 5, 6}
```

jolivier23 (approved):

PR #13545, opened 2023-12-28
DAOS-8331 client: Export client metrics via agent

mjmac:
Adds new agent config parameters and code to
optionally export client metrics in Prometheus
format.

Example daos_agent.yml updates:
telemetry_port: 9192 # export on port 9192
telemetry_retain: 5m # retain metrics for 5 minutes
		   # after client exit

Change-Id: I77864682cc19fa4c33f326d879e20704ef57a7ea
Required-githooks: true

### DIFF
--- a/src/control/cmd/daos_agent/config.go
+++ b/src/control/cmd/daos_agent/config.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2023 Intel Corporation.
+// (C) Copyright 2020-2024 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -54,6 +54,13 @@ type Config struct {
 	DisableAutoEvict    bool                      `yaml:"disable_auto_evict,omitempty"`
 	ExcludeFabricIfaces common.StringSet          `yaml:"exclude_fabric_ifaces,omitempty"`
 	FabricInterfaces    []*NUMAFabricConfig       `yaml:"fabric_ifaces,omitempty"`
+	TelemetryPort       int                       `yaml:"telemetry_port,omitempty"`
+	TelemetryRetain     time.Duration             `yaml:"telemetry_retain,omitempty"`
+}
+
+// TelemetryEnabled returns true if client telemetry collection/export is enabled.
+func (c *Config) TelemetryEnabled() bool {
+	return c.TelemetryPort > 0
 }
 
 // NUMAFabricConfig defines a list of fabric interfaces that belong to a NUMA
@@ -86,6 +93,10 @@ func LoadConfig(cfgPath string) (*Config, error) {
 
 	if !daos.SystemNameIsValid(cfg.SystemName) {
 		return nil, fmt.Errorf("invalid system name: %s", cfg.SystemName)
+	}
+
+	if cfg.TelemetryRetain > 0 && cfg.TelemetryPort == 0 {
+		return nil, errors.New("telemetry_retain requires telemetry_port")
 	}
 
 	return cfg, nil
--- a/src/control/cmd/daos_agent/infocache.go
+++ b/src/control/cmd/daos_agent/infocache.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2023 Intel Corporation.
+// (C) Copyright 2020-2024 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -8,6 +8,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"strings"
 	"sync"
@@ -22,6 +23,7 @@ import (
 	"github.com/daos-stack/daos/src/control/lib/control"
 	"github.com/daos-stack/daos/src/control/lib/hardware"
 	"github.com/daos-stack/daos/src/control/lib/hardware/hwprov"
+	"github.com/daos-stack/daos/src/control/lib/telemetry"
 	"github.com/daos-stack/daos/src/control/logging"
 )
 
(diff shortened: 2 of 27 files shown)


daltonbohning (commented):
Should there be new entries added to utils/config/daos_agent.yml?

Also, could I ask for this small change? It would allow functional tests - specifically, the performance tests - to set the config
https://github.com/daos-stack/daos/commit/578d9075c0e432c3121fade934ff8ced3eff24e7
And since it's unused, no special testing is needed

mjmac:
> Should there be new entries added to utils/config/daos_agent.yml?
> 
Yes, good catch. I forgot about those.

> Also, could I ask for this small change? It would allow functional tests - specifically, the performance tests - to set the config [578d907](https://github.com/daos-stack/daos/commit/578d9075c0e432c3121fade934ff8ced3eff24e7) And since it's unused, no special testing is needed
>
I'll merge that in, thanks. Actually, I may try to add a ftest for this work, so that change makes it even easier.

mjmac:
> 
> I'll merge that in, thanks. Actually, I may try to add a ftest for this work, so that change makes it even easier.

Just refreshed this patch. I did add the `agent_utils_params.py` changes. I have not gotten to adding the ftest yet. As these metrics are still somewhat of a WIP, IMO it's premature to add tests that are expecting fixed sets of metrics while we're iterating. I agree with @wangdi1 that we should add the ftest later.

mjmac:
Requesting early reviews while waiting for the base patch to land, TIA.

tanabarr (approved):
Not overly familiar with the telemetry code but these changes LGTM.

src/control/cmd/daos_agent/config.go
```diff
@@ -88,6 +95,10 @@ func LoadConfig(cfgPath string) (*Config, error) {
 		return nil, fmt.Errorf("invalid system name: %s", cfg.SystemName)
 	}
 
+	if cfg.TelemetryRetain > 0 && cfg.TelemetryPort == 0 {
```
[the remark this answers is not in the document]
mjmac: Hmm. I see where you're coming from on this suggestion, but IMO there's a subtle difference. It's true that the current implementation of `TelemetryEnabled()` just checks to see if `TelemetryPort` is greater than zero. If we change that logic to depend on other checks in order to determine if it's enabled, though, then we have to go back and change this line which is specifically about validating the configuration values.

src/control/lib/telemetry/promexp/engine.go
```diff
@@ -0,0 +1,271 @@
...
+		}
+	}
+}
+
+func (c *EngineCollector) getSources() []*EngineSource {
+	c.sourceMutex.RLock()
+	defer c.sourceMutex.RUnlock()
+
+	sourceCopy := make([]*EngineSource, len(c.sources))
+	_ = copy(sourceCopy, c.sources)
```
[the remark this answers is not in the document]
mjmac: TBH, not sure why I added this placeholder assignment in the original method. I've removed it though, because it's distracting.

daltonbohning (commented):
ftest code changes LGTM. Thanks for adding!

src/control/lib/telemetry/promexp/source.go
```diff
@@ -0,0 +1,200 @@
...
+
+func newSourceMetricSchema(addFn func(logging.Logger, telemetry.Metric) *sourceMetric) *sourceMetricSchema {
+	return &sourceMetricSchema{
+		sourceMetrics: make(map[string]*sourceMetric),
+		seen:          make(map[string]struct{}),
+		addFn:         addFn,
+	}
+}
+
+func (s *sourceMetricSchema) Prune() {
```
kjacque, line 49: nit - needs documentation comment

src/control/lib/telemetry/promexp/source.go
```diff
@@ -0,0 +1,200 @@
...
+
+type (
+	sourceMetricSchema struct {
+		mu            sync.Mutex
+		sourceMetrics map[string]*sourceMetric
+		seen          map[string]struct{}
+		addFn         func(logging.Logger, telemetry.Metric) *sourceMetric
+	}
+
+	MetricSource struct {
```
kjacque, line 32: nit - needs documentation comment

src/control/lib/telemetry/promexp/source.go
```diff
@@ -0,0 +1,200 @@
...
+func (s *MetricSource) Enable() {
+	s.enabled.SetTrue()
+}
+
+// Disable disables the source.
+func (s *MetricSource) Disable() {
+	s.enabled.SetFalse()
+}
+
+func (s *MetricSource) Collect(log logging.Logger, ch chan<- *sourceMetric) {
```
kjacque, line 153: nit - needs documentation comment

src/control/lib/telemetry/promexp/source.go
```diff
@@ -0,0 +1,200 @@
...
+		s.tmSchema.Prune()
+	}()
+
+	for metric := range metrics {
+		ch <- s.smSchema.add(log, metric)
+	}
+	s.smSchema.Prune()
+}
+
+func (s *MetricSource) PruneSegments(log logging.Logger, maxSegAge time.Duration) {
```
kjacque, line 184: needs documentation comment

src/control/cmd/daos_agent/infocache.go
```diff
@@ -292,6 +299,41 @@ func (c *InfoCache) EnableStaticFabricCache(ctx context.Context, nf *NUMAFabric)
...
+	if err != nil {
+		return nil, err
+	}
+	c.adjustAttachInfo(resp)
+	return resp, nil
+}
+
+// adjustAttachInfo performs any necessary adjustments to the attach info
+// before returning it.
+func (c *InfoCache) adjustAttachInfo(resp *control.GetAttachInfoResp) {
```
kjacque, line 320: We make other modifications to the GetAttachInfoResp struct in mgmt_rpc.go before returning to the client. Infocache has been caching only what the server sends back. I'm not exactly opposed to doing it this way, but it feels a little odd making our client-side modifications in two different layers.

kjacque (changes requested):

mjmac:
Closed in favor of the approach in #14030.

PR #13549, opened 2024-01-02
DAOS-14804 vos: fix memory track counter

wangshilong:
1. remove duplicated vtl_dtx_cmt_ent_cnt.
2. mallocinfo() hurts performance significantly, only got about 100MiB per seconds
output with memory track enabled, reduce frequency of recording.
3. fix per xstream memory track counter.

Required-githooks: true


src/vos/vos_container.c
```diff
@@ -200,6 +200,8 @@ cont_free_internal(struct vos_container *cont)
 	cont->vc_pool->vp_dtx_committed_count -= cont->vc_dtx_committed_count;
 	d_tm_dec_gauge(vos_tls_get(cont->vc_pool->vp_sysdb)->vtl_committed,
 		       cont->vc_dtx_committed_count);
+	d_tm_dec_gauge(vos_tls_get(cont->vc_pool->vp_sysdb)->vtl_dtx_cmt_ent_cnt,
+		       cont->vc_dtx_committed_count);
```
Nasf-Fan, line 204: I am wondering why we maintain both "vtl_committed" and "vtl_dtx_cmt_ent_cnt"? They seems have the same functionality for the count of committed DTX entries. @wangdi1

jolivier23 (dismissed):

src/vos/vos_container.c
wangdi1: ah, if they are the same, then maybe just keep one here. thanks

wangdi1 (dismissed):

liuxuezhao (dismissed):

src/vos/vos_container.c
wangshilong: Removed.

Nasf-Fan (dismissed):

liuxuezhao (dismissed):

jolivier23 (dismissed):
You'll need to adjust the description

wangdi1 (dismissed):

Nasf-Fan (dismissed):

jolivier23 (dismissed):

liuxuezhao (dismissed):

NiuYawei (approved):

src/gurt/telemetry.c
```diff
@@ -1814,6 +1846,31 @@ d_tm_record_meminfo(struct d_tm_node_t *metric)
...
+		D_ERROR("Failed to update memusage on item %s not a "
+			"memusage.  Operation mismatch: " DF_RC "\n",
+			metric->dtn_name, DP_RC(-DER_OP_NOT_PERMITTED));
+		return;
+	}
+	d_tm_node_lock(metric);
+	metric->dtn_metric->dtm_data.memusage += bytes;
+	d_tm_node_unlock(metric);
+
+}
```
kjacque, line 1872: Unclear to me why this new type of metric is needed. Isn't this still basically a gauge?

kjacque (changes requested):

src/gurt/telemetry.c
wangshilong: This new counter metric might be negative, and current infra for gauge based on uint64_t. I think my use case is a bit special, so i add a new type here.

mjmac: Why would memory usage be a negative value?

wangshilong: Memory might be allocated and freed by different XS, using an atomic for all XS might be too much overhead. since this is almost a debug operation for developers, sum all all XS memory usage here is accurate and helpful.

NiuYawei: It's bit weird to define a signed integer as "D_TM_MEMUSAGE", also the "int bytes" can only support up to 2GB bytes.
Why don't we just change the current gauge to signed 64bits integer? Or introduce a new "signed gauge"?

NiuYawei (changes requested):
Overlooked the metric type changes in prior review.

src/gurt/telemetry.c
wangshilong: memusage is defined as "long long int",  i thought arg "int bytes" might be fine for normal allocation, but i might declare it to "long long int" as well, I am not aware of any other case that might use signed gauge.....

just change the current gauge to signed 64 bits integer will break current API, eg: d_tm_get_gauge()  which is a public api 
@kjacque might confirm. another way is introduce new API like you said "signed gauge", i wondered if it really deserved to do so?

kjacque: I don't think it is a good idea to change the type for gauge, as we have a number of consumers of metrics. Changing that would have definite ripple effects. If we must have a new type, I think signed gauge is not a bad idea.

However, I still do not totally buy into the idea that we need to account for negative values in this case. Are we actually seeing underflows of these metrics in the field, @wangshilong? If so, could we update the d_tm_dec_gauge function to check if input > current value, and if so, set the result to 0? Would that be adequate for accuracy if this is a debug metric we are OK with approximating?

PR #13569, opened 2024-01-04
DAOS-14850 control: Allow logging.Logger in Context

mjmac:
Updates control plane tools to set a context in a logger
for ease of debug/trace logging.

Change-Id: I94ea454ca83afa3ba035bdcc8ec419b0ad7897e7
Required-githooks: true


tanabarr (approved):

brianjmurrell:
@mjmac Just an FYI/reminder that you can use the `Run-GHA: true` commit pragma to run your functional testing on GitHub Actions and see the results of the testing in this PRs _Checks_ tab.

mjmac:
> @mjmac Just an FYI/reminder that you can use the `Run-GHA: true` commit pragma to run your functional testing on GitHub Actions and see the results of the testing in this PRs _Checks_ tab.

This is an either/or choice though, right? In other words, if I run the tests in GHA, then the Jenkins tests won't run, which means that I have to run it again in order to land. In cases like there, where the change is focused and test analysis is unlikely to be needed, there's no particular benefit to the extra step, IMO.

mjmac:
> @mjmac Just an FYI/reminder that you can use the `Run-GHA: true` commit pragma to run your functional testing on GitHub Actions and see the results of the testing in this PRs _Checks_ tab.

fyi, I tried this on a draft PR that I'm not trying to land with any urgency... Seems like it's totally busted? https://github.com/daos-stack/daos/actions/runs/7466679565/job/20318574941

brianjmurrell:
> This is an either/or choice though, right?

Correct.

> In other words, if I run the tests in GHA, then the Jenkins tests won't run, which means that I have to run it again in order to land.

You should not have to re-run to land.  The GHA testing should be accepted as equal to Jenkins testing.  We have some work to do to get the _Required) checks to align (either by making Jenkins use the GHA check names or having GHA manually set the existing Jenkins labels -- which way is yet to be determined) between the two, so a force-landing would be required but I think should be acceptable.

> In cases like this, where the change is focused and test analysis is unlikely to be needed, there's no particular benefit to the extra step, IMO.

Agreed, if there were an extra step.  But I don't think there should be.  If you set the force-landing label and fill out the checklist explaining explaining why the force-landing is necessary the gatekeeper should be able to see the alternative GHA testing checks and be comfortable landing.

Indeed, this is all new process so there are things to be worked out, but we won't know what unless we start using it.

But ultimately it's up to you as the submitter.  I just thought/think it might be helpful if you can actually see your test results and not (presumably -- I'm not sure how you get failed test results from Jenkins) need a relay from inside of Intel to tell you why a test failed.

brianjmurrell:
> fyi, I tried this on a draft PR that I'm not trying to land with any urgency... Seems like it's totally busted? [daos-stack/daos/actions/runs/7466679565/job/20318574941](https://github.com/daos-stack/daos/actions/runs/7466679565/job/20318574941)

This is known.  This is just an "alternate" (to the one that works -- but only if your run completes within 24h of your submitting it) JUnit report that could not be tested before landing due to a catch-22 in some kinds of GHA workflow structures.

The actual workflow for that run can be found at https://github.com/daos-stack/daos/actions/runs/7466584253 which you can get to by clicking the _Checks_ tab on the PR:

image

and then choosing the _RPM Build and Test_ workflow:

image

You can see there that your Leap 15 RPM build failed.  If you click the failed step:

image

and then open the _Build RPM failure log_ step:

image

But the log is missing so something more rudimentary failed.  Look at the _Build RPM_ step and you can see there is a repo missing in artifactory.  That's because the GH workflow is still at 15.4.  This workflow should have been updated to 15.5 when the Jenkins workflow was updated to 15.5.  This stuff is still pretty new so it got forgotten about in that effort unfortunately.  Apologies for that.

If you change https://github.com/daos-stack/daos/blob/fb8c169a6782aaad2a77931e982a9168f3197442/.github/workflows/rpm-build-and-test.yml#L11 to `15.5` (and may as well change https://github.com/daos-stack/daos/blob/fb8c169a6782aaad2a77931e982a9168f3197442/.github/workflows/rpm-build-and-test.yml#L8 to `8.8` although for your purposes that is probably not particularly important but would be consistent with Jenkins) that should fix that Leap 15 build failure.

knard-intel (approved):
LGTM

brianjmurrell:
> That's because the GH workflow is still at 15.4. This workflow should have been updated to 15.5 when the Jenkins workflow was updated to 15.5. This stuff is still pretty new so it got forgotten about in that effort unfortunately. 

Fixed in daos-stack/daos#13608 -- just waiting to land.

kjacque (approved):
A nice addition. Will the actual usage of the logger from the context be added with some future change?

PR #13570, opened 2024-01-05
DAOS-14845 vos: avoid TX_RESTART for modification with fixed epoch

Nasf-Fan:
The ilog_tree_modify() logic should not return -DER_TX_RESTART when
hit existing ilog entry for the modification with fixed epoch, such
as rebuild, ec_agg, ec_rep and so on. Because related modification
must either has been committed or is committable under such cases.
Please note that rebuild may split modifications belong to the same
transaction into multiple vos operations. So the modifications with
the same fixed epoch should be regarded as the same transaction and
need to be further handled with minor epoch.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/vos/ilog.c
+++ b/src/vos/ilog.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2019-2023 Intel Corporation.
+ * (C) Copyright 2019-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -81,6 +81,8 @@ struct ilog_context {
 	bool				 ic_in_txn;
 	/** version needs incrementing */
 	bool				 ic_ver_inc;
+	/** For operation with fixed epoch (rebuild, ec_agg, ec_rep, etc.) */
+	bool				 ic_fixed_epoch;
 };
 
 D_CASSERT(sizeof(struct ilog_id) == sizeof(struct ilog_tree));
@@ -401,7 +403,7 @@ ilog_create(struct umem_instance *umm, struct ilog_df *root)
 
 int
 ilog_open(struct umem_instance *umm, struct ilog_df *root,
-	  const struct ilog_desc_cbs *cbs, daos_handle_t *loh)
+	  const struct ilog_desc_cbs *cbs, bool fixed_epoch, daos_handle_t *loh)
 {
 	struct ilog_context	*lctx;
 	int			 rc;
@@ -412,6 +414,7 @@ ilog_open(struct umem_instance *umm, struct ilog_df *root,
 	if (rc != 0)
 		return rc;
 
+	lctx->ic_fixed_epoch = fixed_epoch;
 	*loh = ilog_lctx2hdl(lctx);
 
 	return 0;
@@ -603,7 +606,15 @@ check_equal(struct ilog_context *lctx, struct ilog_id *id_out, const struct ilog
 			    id_out->id_update_minor_eph >
 			    id_out->id_punch_minor_eph)
 				return -DER_ALREADY;
+		} else if (lctx->ic_fixed_epoch) {
+			/*
+			 * For operation with fixed epoch, when update existing ilog entry,
+			 * regard them as the same and use minor epoch for further handling.
+			 */
+			*is_equal = true;
+			return 0;
 		}
+
 		D_DEBUG(DB_IO, "Access of incarnation log from multiple DTX"
 			" at same time is not allowed: rc=DER_TX_RESTART\n");
 		return -DER_TX_RESTART;
--- a/src/vos/ilog.h
+++ b/src/vos/ilog.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2019-2023 Intel Corporation.
+ * (C) Copyright 2019-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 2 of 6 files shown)


Nasf-Fan:
[DAOS_FileSystem_DFS_Parallel_DTX_cmocka_results.xml.[failed-to-read]](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-13570/2/testReport/junit/DAOS_FileSystem_DFS_Parallel_DTX_cmocka_results/xml/_failed_to_read_/)

This is a known XML formatted issue, another failure instance of DAOS-14594, not related with the patch.

src/vos/ilog.c
```diff
@@ -575,6 +578,10 @@ check_equal(struct ilog_context *lctx, struct ilog_id *id_out, const struct ilog
 		return 0;
 
 	if (update) {
+		/* For rebuild case, related modification must be either committed or committble. */
+		if (lctx->ic_replay)
+			return -DER_ALREADY;
+
```
wangdi1, line 584: A little concern about this change. I thought ecoch for rebuild update should be detected by the following check? Or I miss sth?
` if (dtx_is_committed(id_out->id_tx_id, ilog_ctx2cont(lctx), id_out->id_epoch)) {
                        /** Need to differentiate between updates that are
                         * overwrites and others that are conflicts.  Return
                         * a different error code in this case if the result
                         * would be the same (e.g. not mixing update with
                         * punch
                         */
                        if (id_in->id_punch_minor_eph &&
                            id_out->id_punch_minor_eph >
                            id_out->id_update_minor_eph)
                                return -DER_ALREADY;

                        if (id_in->id_update_minor_eph &&
                            id_out->id_update_minor_eph >
                            id_out->id_punch_minor_eph)
                                return -DER_ALREADY;
                }


Btw: do you think this might be related with this fix in PR-13493 ?
diff --git a/src/vos/ilog.c b/src/vos/ilog.c
index e7dd2bff53..08528e40c4 100644
--- a/src/vos/ilog.c
+++ b/src/vos/ilog.c
@@ -587,7 +587,7 @@ check_equal(struct ilog_context *lctx, struct ilog_id *id_out, const struct ilog
                        D_DEBUG(DB_IO, "No entry found, done\n");
                        return 0;
                }
-               if (dtx_is_committed(id_in->id_tx_id, ilog_ctx2cont(lctx), id_in->id_epoch)) {
+               if (dtx_is_committed(id_out->id_tx_id, ilog_ctx2cont(lctx), id_out->id_epoch)) {
                        /** Need to differentiate between updates that are
                         * overwrites and others that are conflicts.  Return
                         * a different error code in this case if the result

Nasf-Fan: The original implementation "if (dtx_is_committed(id_in->id_tx_id, ilog_ctx2cont(lctx), id_in->id_epoch)) " before PR-13493 is NOT correct.

As my comment pointed out, the input parameter "id_in->id_tx_id" for ilog_update() is zero, that will be regarded as "committed", so such check "dtx_is_committed()" is always true, that is wrong. The expected behavior is to check whether the "id_out" is a committed/committable ilog entry or not, if yes, we can directly use it; otherwise, we found two in-processing transactions using the same epoch and the current one need to be restarted.

It means that the original improper dtx_is_committed() checking hide TX_RESTART issue. Maybe @jolivier23 can give more check for related logic. Thanks!

Nasf-Fan: > A little concern about this change. I thought ecoch for rebuild update should be detected by the following check? Or I miss sth?

For most of cases, rebuild can be detected by dtx_is_committed() and return "return -DER_ALREADY". But for committable (instead of committed) case, current in-rebuilding target cannot know whether its local ilog entry is committable or not. So if we do not offer "ic_replay" flag, the logic will regard it as DTX conflict and require TX restart. That is the logic we fixed in this patch.

jolivier23 (dismissed):

src/vos/ilog.c
wangdi1: Sorry, I do not fully understand the check_equally logic here. so id_out stands for the cached/uncommited ops? then how id_in and id_out are equal for this case?  there is clear boundary between inflight IO and rebuild. hmm. 

My concern is whether we should unconditional return EALDEY for rebuild case, just wonder if we can tell it from epoch.  I assume in_id epoch can be set to some special rebuild epoch etc. Though I am not so familiar with the logic here. might miss sth.

wangdi1 (approved):

src/vos/ilog.c
Nasf-Fan: id_out is the existing ilog entry, it is the one to be committed, aborted or shared. The id_in is the one for current operation. For rebuild case, the regular IO request has already been sent to current in-rebuilding target, and prepared, and then the rebuild request with the same epoch is sent to such target for rebuilding, then find the existing ilog entry with the same epoch. Under such case, the existing ilog entry must be committable, means related rebuild IO has already been done.

Nasf-Fan:
Sorry, I have to force re-pushed the patch because the new version much changed the logic to fix more issues than before, not only for rebuild, but also for ec_agg, ec_rep, and so on.

liuxuezhao (approved):

Nasf-Fan:
> Test stage Functional Hardware Medium completed with status UNSTABLE. https://build.hpdd.intel.com/job/daos-stack/job/daos//view/change-requests/job/PR-13570/8/testReport/

create_no_space_loop failed for DAOS-14884, not related with the patch.

PR #13598, opened 2024-01-12
DAOS-12209 pool: force destroy ignore evict errors

kccain:
With this change, the daos_server pool destroy handler, when invoked with force destroy (true), is changed to ignore any errors that may occur when it first invokes PoolEvict before transitioning the pool to Destroying state and performing the rest of the destroy procedure, This allows the remaining destroy procedure to execute, and to clear the pool's resources including its entry in the MS list.

This is needed because in some edge cases for pool create failure a pool may be left in a Creating state and on subsequent server restarts, even a pool destroy --force is unable to remove the pool entry from the MS list (and is also unable to clear any engine target and rdb allocations that may have been performed in the previous, incomplete, pool create operation).

The scenario that can cause the stranded pool in Creating state:
- a pool create gets delayed/stuck after allocating targets and rdb, but before being able to run the final step, the POOL_CREATE RPC send (and its handling).
- The admin kills the dmg pool create command in progress (CTRL-C).
- The admin also restarts the daos servers in the system.
- After server restart, the pool is in the MS directory in a Creating state, and a dmg pool destroy --force will not succeed (rather, it repeatedly gets -DER_AGAIN replies).

Features: pool

Signed-off-by: Kenneth Cain <kenneth.c.cain@intel.com>          
Co-authored-by: Tom Nabarro <tom.nabarro@intel.com>

### Gatekeeper:


kccain:
Jenkins [build 4](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-13598/4/) seems to have run well, with the exception of some known failures. While waiting for master branch stability, I think it makes sense to start reviewing this patch.

liw (approved):

mjmac (approved):
The change as-is seems fine to me, although I wonder if there's any good reason to bother with the evict before destroy when it's being forced. In other words, maybe it would make the logic a little less convoluted if we just skipped directly to the destroy when the force flag is set?

From the client perspective, I think the end result would be roughly the same. If the client happened to be doing I/O when the evict happened, then I'm pretty sure it would get a -DER_NO_HDL error. I don't remember if it retries at that point or if it just fails completely. If it did retry, then I'm guessing it would get a -DER_NO_SERVICE or something similar. Either way, it should be pretty clear to the user that something is pretty broken.

kccain:
> The change as-is seems fine to me, although I wonder if there's any good reason to bother with the evict before destroy when it's being forced. In other words, maybe it would make the logic a little less convoluted if we just skipped directly to the destroy when the force flag is set?
> 
> From the client perspective, I think the end result would be roughly the same. If the client happened to be doing I/O when the evict happened, then I'm pretty sure it would get a -DER_NO_HDL error. I don't remember if it retries at that point or if it just fails completely. If it did retry, then I'm guessing it would get a -DER_NO_SERVICE or something similar. Either way, it should be pretty clear to the user that something is pretty broken.

Once master branch gets a few fixes for some unrelated failures, I could try the following patch and see if it holds up in regression testing. How do these diffs look?

`diff --git a/src/control/server/mgmt_pool.go b/src/control/server/mgmt_pool.go
index 4758ac582..996f4aa9e 100644
--- a/src/control/server/mgmt_pool.go
+++ b/src/control/server/mgmt_pool.go
@@ -697,23 +697,31 @@ func (svc *mgmtSvc) PoolDestroy(parent context.Context, req *mgmtpb.PoolDestroyR
                        }
                }
 
-               // Perform separate PoolEvict _before_ possible transition to destroying state.
-               evStatus, err := svc.poolEvictConnections(ctx, req)
-               if !req.Force && (err != nil) {
-                       return nil, err
+               updatePoolService := req.Force
+               evStatus := daos.Success
+
+               if !req.Force {
+                       // Perform separate PoolEvict _before_ possible transition to destroying state.
+                       evStatus, err := svc.poolEvictConnections(ctx, req)
+                       if err != nil {
+                               return nil, err
+                       }
+                       if evStatus != daos.Busy && evStatus != daos.NoService {
+                               updatePoolService = true
+                       }
                }
 
                // If the request is being forced, or the evict request did not fail
                // due to the pool being busy or service not up, then transition to the
                // destroying state and persist the update(s).
-               if req.Force || (evStatus != daos.Busy && evStatus != daos.NoService) {
+               if updatePoolService {
                        ps.State = system.PoolServiceStateDestroying
                        if err := svc.sysdb.UpdatePoolService(ctx, ps); err != nil {
                                return nil, errors.Wrapf(err, "failed to update pool %s", poolUUID)
                        }
                }
 
-               if !req.Force && evStatus != daos.Success {
+               if evStatus != daos.Success {
                        svc.log.Errorf("PoolEvict during pool destroy failed: %s", evStatus)
                        resp.Status = int32(evStatus)
                        return resp, nil
`

liw:
Currently, we don't take good care of local pool handle objects when destroying a pool on an engine. So it's better to keep the `PoolEvict` even for force until we fix the `ds_pool_stop` to close all associated pool handles.

mjmac:
> Currently, we don't take good care of local pool handle objects when destroying a pool on an engine.

Hi @liw. Would you please elaborate on this statement? What are the possible ramifications of destroying a pool without an evict first? I thought that the handles were stored in per-pool metadata. Is that not reclaimed when the pool is destroyed?

liw:
> > Currently, we don't take good care of local pool handle objects when destroying a pool on an engine.
> 
> Hi @liw. Would you please elaborate on this statement? What are the possible ramifications of destroying a pool without an evict first? I thought that the handles were stored in per-pool metadata. Is that not reclaimed when the pool is destroyed?

Hi @mjmac. When we handle a `POOL_TGT_DESTROY` on an engine, we don't (but should) release in-memory pool handle objects. These objects in turn may prevent us from releasing the space allocated to the pool on this engine.

mjmac:
> Hi @mjmac. When we handle a `POOL_TGT_DESTROY` on an engine, we don't (but should) release in-memory pool handle objects. These objects in turn may prevent us from releasing the space allocated to the pool on this engine.

Would it make sense to just always invoke the local evict handling code from the target destroy handler? That seems like a straightforward and simple solution (he says, naively).

liw:
> Would it make sense to just always invoke the local evict handling code from the target destroy handler? That seems like a straightforward and simple solution (he says, naively).

We don't know which pool handles belong to this pool. Plus, code structure wise, I think the closing of pool handles should be done in `ds_pool_stop`.

liw:
I've created https://daosio.atlassian.net/browse/DAOS-15040 for that issue.

src/control/server/mgmt_pool.go
```diff
@@ -699,7 +699,7 @@ func (svc *mgmtSvc) PoolDestroy(parent context.Context, req *mgmtpb.PoolDestroyR
 
 		// Perform separate PoolEvict _before_ possible transition to destroying state.
 		evStatus, err := svc.poolEvictConnections(ctx, req)
-		if err != nil {
+		if !req.Force && (err != nil) {
```
tanabarr, line 702: parentheses around equality statement is not needed in go

src/control/server/mgmt_pool.go
```diff
@@ -699,7 +699,7 @@ func (svc *mgmtSvc) PoolDestroy(parent context.Context, req *mgmtpb.PoolDestroyR
...
 			return nil, err
```
tanabarr, line 703: In force mode, do we want to log the errors whilst ignoring them?

src/control/server/mgmt_pool.go
```diff
@@ -713,7 +713,7 @@ func (svc *mgmtSvc) PoolDestroy(parent context.Context, req *mgmtpb.PoolDestroyR
 			}
 		}
 
-		if evStatus != daos.Success {
+		if !req.Force && evStatus != daos.Success {
 			svc.log.Errorf("PoolEvict during pool destroy failed: %s", evStatus)
```
tanabarr, line 717: In force mode we will probably want to log this.

src/control/server/mgmt_pool_test.go
```diff
@@ -826,25 +826,6 @@ func TestServer_MgmtSvc_PoolDestroy(t *testing.T) {
 			},
 			expSvc: svcWithState(testPoolService, system.PoolServiceStateDestroying),
 		},
-		"force=true, evict dRPC fails due to engine error": {
```
tanabarr, line 829: We should have 2 test cases that verify the destroy flow completes with either of the failures occurring.

tanabarr (changes requested):

tanabarr (commented):
I won't block on the suggestions.

tanabarr (dismissed):
LGTM, suggestions in case you need to re-push or want to improve unit test coverage

src/control/server/mgmt_pool.go
kccain: I'm looking at changing the Debugf to Errorf within poolEvictConnections() for this

src/control/server/mgmt_pool_test.go
kccain: I tried changing this test case but could not figure out how to arrange it properly. Do you have ideas on how to structure the test cases?

kjacque (changes requested):
Basically looks okay to me, I would like the test case case fixed though.

src/control/server/mgmt_pool_test.go
tanabarr: I'm having a go at this, @kccain do you want me to push a Commit to the PR or paste the diff in the comments?

kccain: > I'm having a go at this, @kccain do you want me to push a Commit to the PR or paste the diff in the comments?

Thanks Tom. Pushing a commit would be fine, making sure to use Features: pool commit pragma for CI testing.

tanabarr: Pushed with pragma to test pool features and enable MD-on-SSD coverage. The test refactor should enable coverage of different code-paths by enabling customisation of serial dRPC responses.

tanabarr (commented):
I have updated the tests.

src/control/server/mgmt_pool_test.go
```diff
@@ -719,37 +719,42 @@ func TestServer_MgmtSvc_PoolCreateDownRanks(t *testing.T) {
...
+			State:     system.PoolServiceStateReady,
+			Storage: &system.PoolServiceStorage{
+				CreationRankStr: ranklist.MustCreateRankSet("0-7").String(),
+			},
+		}
 	}
-	svcWithState := func(in *system.PoolService, state system.PoolServiceState) (out *system.PoolService) {
-		out = new(system.PoolService)
+	ready := system.PoolServiceStateReady
+	destroying := system.PoolServiceStateDestroying
```
kccain, line 734: I think we need test cases in which pool state starts out as Creating, which this PR aims to resolve (problem was that a pool destroy --force still was unable to destroy such a pool). Can it be represented in this framework?

1. PoolDestroy request with Recursive: true and Force:false. Expect evict dRPC failure with -DER_AGAIN, and PoolDestroy response with the same error. evict fails since there is no pool service alive (previous create attempt didn't create pool service entirely, and got interrupted with CTRL-C of dmg, and termination of servers). destroy dRPC doesn't happen because force=false.
2. case 1 with Recursive: false and Force:false - same result I think
3. PoolDestroy request with Recursive: true and Force:true. Expect evict dRPC failure, which implementation shall ignore, and pool destroy dRPC success.
4. case 3 with Recursive: true - same result I think

kccain (commented):
just commenting / asking some questions
perhaps may want some pool destroy test cases in which the starting state = Creating.

src/control/server/mgmt_pool_test.go
tanabarr: sounds good, could you add those cases or do you want me to?

src/control/server/mgmt_pool_test.go
```diff
@@ -759,164 +764,248 @@ func TestServer_MgmtSvc_PoolDestroy(t *testing.T) {
...
 			},
 			expResp: &mgmtpb.PoolDestroyResp{
 				Status: int32(daos.MiscError),
 			},
-			expSvc: svcWithState(testPoolService, system.PoolServiceStateDestroying),
+			expSvcState: &destroying,
 		},
-		"already destroying, destroy dRPC fails due to engine error": {
-			req: &mgmtpb.PoolDestroyReq{Id: mockUUID, Recursive: true},
+		"force=true, recursive=true, evict dRPC fails -DER_BUSY, pool in destroying state": {
```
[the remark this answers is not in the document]
tanabarr: correct, which results in the pool service remaining in the destroying state

tanabarr:
GATEKEEPER: please use the PR title and description as the commit message when landing (omitting any test pragmas) TIA

src/control/server/mgmt_pool_test.go
kccain: I can add them.

liw (dismissed):

src/control/server/mgmt_pool_test.go
tanabarr: adding...

tanabarr: There is a discrepancy between error handling in ListCont and PoolEvict in the case of creating pool service state. The former returns the error in response whereas the latter returns the error outright.

kjacque (dismissed):

tanabarr:
@kccain added requested creating test cases, please check whether behaviour is as you expect

liw (approved):

kccain:
> @kccain added requested creating test cases, please check whether behaviour is as you expect

Yes it looks good, thank you.

tanabarr (approved):

tanabarr:
probably want one more go reviewer to approve

knard-intel (approved):
LGTM

tanabarr:
All tests passed in https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-13598/7/pipeline/ apart from the known go unit test failure which has been resolved on master with PR-13672. Requesting force landing.

gnailzenh (approved):

PR #13605, opened 2024-01-14
DAOS-14550 control: Prevent SPDK calls if disable_hugepages true in cfg

tanabarr:
Avoid calling into SPDK if disable_hugepages is set in the server
config file. This change prevents access during StorageScan,
StorageFormat, StorageNvmeRebind and StorageNvmeAddDevice server
ControlService handlers and when cleaning hugepages during start-up.

Also:
- Improve fault information messages
- Improve skip related messaging during format
- Cover missing config in storage command handler unit tests
- Fix quirks in storage command display output
- Improve unit test coverage of storage pretty printers

Features: control
Required-githooks: true

### Gatekeeper:


knard-intel (approved):
LGTM

mjmac (dismissed):

tanabarr:
If we want coverage for disable_hugepages in CI we probably want to run a SCM-only test after clearing hugepages on a host, that would have caught the `permission denied` SPDK errors. `sudo su -c "echo 0 > /proc/sys/vm/nr_hugepages"`

src/control/server/ctl_storage_rpc.go
```diff
@@ -29,8 +29,11 @@ import (
 )
 
 const (
-	msgFormatErr      = "instance %d: failure formatting storage, check RPC response for details"
-	msgNvmeFormatSkip = "NVMe format skipped on instance %d as SCM format did not complete"
+	msgFormatErr             = "instance %d: failure formatting storage, check RPC response for details"
+	msgNvmeFormatSkip        = "NVMe format skipped on instance %d"
+	msgNvmeFormatSkipHPD     = msgNvmeFormatSkip + ", use of hugepages disabled in config"
+	msgNvmeFormatSkipFail    = msgNvmeFormatSkip + ", SCM format failed"
+	msgNvmeFormatSkipNotDone = msgNvmeFormatSkip + ", SCM was not formatted"
```
knard-intel, line 36:
```suggestion
	msgNvmeFormatSkipNotDone = msgNvmeFormatSkip + ", SCM was already formatted"
```

knard-intel (approved):
Mostly OK, just one small remarks regarding a confusing information message.

src/control/server/ctl_storage_rpc.go
tanabarr: that Skip message is accurate, it represents the case when no formatting was performed, it could be because SCM was already formatted or because of something else

tanabarr:
GATEKEEPER: Please use the PR title and description above as the commit message when landing. TIA

src/control/server/ctl_storage_rpc.go
knard-intel: OK, I was understanding that it was only possible if the SCM was previously formatted.
Thanks for the clarification.

mjmac (commented):
A few things:
  * First, a general grumble... If you're going to sneak in unrelated formatting changes that don't affect functionality, please put them in their own commit so that reviewers don't have to wade through pages of whitespace updates before getting to the "meat" of the patch.
  * Second, the general approach here seems to be to turn a number of function calls into no-ops if hugepages are disabled in the server config. Wouldn't it be better to just not call those functions? It would certainly be easier to understand.
  * Finally, the fact that the code needs to be updated in so many places indicates some problems with delegation of responsibility. Is it possible to restructure the code so that the behavior is controlled in one place? My concern is that having responsibility for this sprinkled around is harder to understand and maintain, and increases the odds that future changes will miss a spot.

tanabarr:
> A few things:
> 
> * First, a general grumble... If you're going to sneak in unrelated formatting changes that don't affect functionality, please put them in their own commit so that reviewers don't have to wade through pages of whitespace updates before getting to the "meat" of the patch.
Sure
> * Second, the general approach here seems to be to turn a number of function calls into no-ops if hugepages are disabled in the server config. Wouldn't it be better to just not call those functions? It would certainly be easier to understand.
I will try to do that
> * Finally, the fact that the code needs to be updated in so many places indicates some problems with delegation of responsibility. Is it possible to restructure the code so that the behavior is controlled in one place? My concern is that having responsibility for t his sprinkled around is harder to understand and maintain, and increases the odds that future changes will miss a spot.
That sounds sensible and I have thought the same, as this is a reactive fix are you okay if I do that in a separate PR? I have to balance against other priorities for 2.6.

src/control/server/ctl_storage_rpc.go
```diff
@@ -966,8 +990,15 @@ func (cs *ControlService) StorageFormat(ctx context.Context, req *ctlpb.StorageF
 		skipped:     instanceSkips,
 		mdFormatted: mdFormatted,
 	}
+	if cs.srvCfg != nil && cs.srvCfg.DisableHugepages {
+		cs.log.Notice("skipping bdev format as use of hugepages disabled in config")
```
kjacque, line 994: nit - The reasoning for the skip is included in the result returned from formatNvme, right? Just wondering if we need this message at the Notice level if the user will be notified anyway.

src/control/server/faults.go
```diff
@@ -56,6 +56,11 @@ var (
 		"cannot destroy a pool with existing containers",
 		"retry the operation with the recursive flag set to remove containers along with the pool",
 	)
+	FaultHugepagesDisabledBadAction = serverFault(
+		code.ServerHugepagesDisabledBadAction,
+		"action cannot be performed as the use of hugepages has been disabled in the server config",
+		"set false (or remove) disable_hugepages parameter in config and reformat storage before retrying the desired action",
+	)
```
kjacque, line 63: Just a suggestion for tightening up the naming/messaging:
```suggestion
	FaultHugepagesDisabled = serverFault(
		code.ServerHugepagesDisabled,
		"the use of hugepages has been disabled in the server config",
		"set false (or remove) disable_hugepages parameter in config and reformat storage, then retry the operation",
	)
```

src/control/server/instance_storage_rpc_test.go
```diff
@@ -185,22 +185,7 @@ func TestIOEngineInstance_bdevScanEngine(t *testing.T) {
...
-								Rank:     uint32(ranklist.NilRank),
-								RoleBits: storage.BdevRoleAll,
-							},
-						}
-						return c
-					}(),
-				},
-				State: new(ctlpb.ResponseState),
-			},
+			expErr: errors.New("nil superblock"),
```
kjacque, line 188: I wonder about changing this behavior... Does this have to be fatal? It is just a scan, right? Does the engine have to be joined for that?

kjacque (commented):
If the change in behavior has already been discussed and thought through elsewhere, I would be willing to approve as-is. The other two items are minor and could be addressed in another PR.

src/control/server/instance_storage_rpc_test.go
kjacque: Discussed in a chat -- the "old" behavior here is a relatively new change, and logically we should not get into a position where the engine has a bad rank but allows a storage scan. This makes sense to me.

kjacque (approved):
Works for me. My suggestions, and greater improvements suggested by Mike, would be fine to address in a future PR.

mjmac:
> That sounds sensible and I have thought the same, as this is a reactive fix are you okay if I do that in a separate PR? I have to balance against other priorities for 2.6.

That's fine with me. Please create a ticket so that the work isn't forgotten, though.

tanabarr:
@mjmac does the most recent commit address your second point sufficiently? I'm assuming you don't want me to address the first point in this PR as it has already been reviewed and I will address your third point in another PR (and create a ticket for it). If it looks okay then I will update the unit tests appropriately.

tanabarr:
> > That sounds sensible and I have thought the same, as this is a reactive fix are you okay if I do that in a separate PR? I have to balance against other priorities for 2.6.
> 
> That's fine with me. Please create a ticket so that the work isn't forgotten, though.

https://daosio.atlassian.net/browse/DAOS-15086

mjmac (commented):
This looks better to me. Conceptually, it's much easier to understand smaller functions that either succeed at doing the thing or fail to do the thing rather than maybe doing the thing.

Holding off on approval for now, as there have been a number of changes in related areas and it would be prudent to re-test after landing to ensure that logic changes don't result in test failures.

src/control/server/faults.go
tanabarr: done

tanabarr:
GATEKEEPER: please use the PR title and description as commit message when landing, TIA

knard-intel (approved):

mjmac (approved):
LGTM

PR #13606, opened 2024-01-15
DAOS-14824 dfuse: Allow parallel readdir calls.

ashleypittman:
Remove the inode lock preventing parallel readdir calls.
Update readdir caching to be thread safe.
Required-githooks: true


ashleypittman:
This ran and passed all the testing but that's mostly pointless as due to a libfuse bug then the kernel feature isn't enabled in our testing.

I'm separately working on a test for this which shows significant performance improvement locally but again that won't show any change in performance or exercise this new code until we can get a new libfuse in CI so it should not block this (although I can see that the libfuse update itself might).

mchaarawi (approved):

wiliamhuang (approved):

PR #13613, opened 2024-01-16
DAOS-12710 build: Improve daos/tse.h handling.

ashleypittman:
Fix doxygen comments in daos_task.h and daos/tse.h

Change includes so all public daos headers are independent and can
be included without including any headers prior.
Add a unit test about the above for all installed headers.
Remove the scons check for gurt and cart headers in favour
of the new unit test.

### DIFF
--- a/src/common/tse.c
+++ b/src/common/tse.c
@@ -17,8 +17,10 @@
 
 #include <stdint.h>
 #include <pthread.h>
+#include <daos_task.h>
+
 #include <daos/common.h>
-#include <daos/tse.h>
+
 #include "tse_internal.h"
 
 D_CASSERT(sizeof(struct tse_task) == TSE_TASK_SIZE);
--- a/src/common/tse.c
+++ b/src/common/tse.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2023 Intel Corporation.
+ * (C) Copyright 2016-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -17,7 +17,6 @@
 
 #include <stdint.h>
 #include <pthread.h>
-#include <daos_task.h>
 
 #include <daos/common.h>
 
--- a/src/include/daos/common.h
+++ b/src/include/daos/common.h
@@ -29,7 +29,6 @@
 #include <daos_obj.h>
 #include <daos_prop.h>
 #include <daos_security.h>
-#include <daos/tse.h>
 
 #include <daos/debug.h>
 #include <gurt/hash.h>
--- a/src/include/daos/common.h
+++ b/src/include/daos/common.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2015-2023 Intel Corporation.
+ * (C) Copyright 2015-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -29,7 +29,7 @@
 #include <daos_obj.h>
 #include <daos_prop.h>
 #include <daos_security.h>
-#include <daos_task.h>
+#include <daos/tse.h>
 
 #include <daos/debug.h>
 #include <gurt/hash.h>
@@ -39,7 +39,6 @@
 #include <daos/profile.h>
 #include <daos/dtx.h>
 #include <daos/cmd_parser.h>
-#include <daos/tse.h>
 
 #define DAOS_ON_VALGRIND D_ON_VALGRIND
 
--- a/src/include/daos/common.h
+++ b/src/include/daos/common.h
@@ -25,17 +25,21 @@
 #include <byteswap.h>
 
 #include <daos_errno.h>
-#include <daos/debug.h>
-#include <gurt/hash.h>
-#include <gurt/common.h>
-#include <cart/api.h>
 #include <daos_types.h>
 #include <daos_obj.h>
 #include <daos_prop.h>
 #include <daos_security.h>
+#include <daos_task.h>
+
+#include <daos/debug.h>
+#include <gurt/hash.h>
+#include <gurt/common.h>
+#include <cart/api.h>
+
 #include <daos/profile.h>
 #include <daos/dtx.h>
 #include <daos/cmd_parser.h>
+#include <daos/tse.h>
 
 #define DAOS_ON_VALGRIND D_ON_VALGRIND
 
--- a/src/include/daos/common.h
+++ b/src/include/daos/common.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2015-2024 Intel Corporation.
+ * (C) Copyright 2015-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -25,16 +25,14 @@
 #include <byteswap.h>
 
 #include <daos_errno.h>
-#include <daos_types.h>
-#include <daos_obj.h>
-#include <daos_prop.h>
-#include <daos_security.h>
-
 #include <daos/debug.h>
 #include <gurt/hash.h>
 #include <gurt/common.h>
 #include <cart/api.h>
-
+#include <daos_types.h>
+#include <daos_obj.h>
+#include <daos_prop.h>
+#include <daos_security.h>
 #include <daos/profile.h>
 #include <daos/dtx.h>
 #include <daos/cmd_parser.h>
--- a/src/include/daos/container.h
+++ b/src/include/daos/container.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2023 Intel Corporation.
+ * (C) Copyright 2016-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
--- a/src/include/daos/container.h
+++ b/src/include/daos/container.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2024 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
--- a/src/include/daos/container.h
+++ b/src/include/daos/container.h
@@ -12,7 +12,6 @@
 
 #include <daos/common.h>
 #include <daos/pool_map.h>
-#include <daos/tse.h>
 #include <daos_types.h>
 #include <daos_cont.h>
 #include <daos/cont_props.h>
--- a/src/include/daos/container.h
+++ b/src/include/daos/container.h
@@ -12,6 +12,7 @@
 
 #include <daos/common.h>
 #include <daos/pool_map.h>
+#include <daos/tse.h>
 #include <daos_types.h>
 #include <daos_cont.h>
 #include <daos/cont_props.h>
--- a/src/include/daos/dpar.h
+++ b/src/include/daos/dpar.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2021-2022 Intel Corporation.
+ * (C) Copyright 2021-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
--- a/src/include/daos/dpar.h
+++ b/src/include/daos/dpar.h
@@ -9,10 +9,6 @@
 #include <stdbool.h>
 #include <inttypes.h>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 #define DPAR_MAJOR 2
 #define DPAR_MINOR 0
 
@@ -92,11 +88,6 @@ par_comm_split(uint32_t comm, int color, int key, uint32_t *new_comm);
 /** Free a communicator */
 int
 par_comm_free(uint32_t comm);
-
-
-#ifdef __cplusplus
-}
-#endif
 
 #endif /** __DAOS_PAR_LIB_H__ */
 
--- a/src/include/daos/mgmt.h
+++ b/src/include/daos/mgmt.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2023 Intel Corporation.
+ * (C) Copyright 2016-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
--- a/src/include/daos/mgmt.h
+++ b/src/include/daos/mgmt.h
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2024 Intel Corporation.
+ * (C) Copyright 2016-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -10,10 +10,9 @@
 #ifndef __DC_MGMT_H__
 #define __DC_MGMT_H__
 
-#include <daos_types.h>
-#include <daos/tse.h>
-
 #include <daos/common.h>
+#include <daos/tse.h>
+#include <daos_types.h>
 #include <daos/pool.h>
 #include "svc.pb-c.h"
 
--- a/src/include/daos/mgmt.h
+++ b/src/include/daos/mgmt.h
@@ -11,6 +11,7 @@
 #define __DC_MGMT_H__
 
 #include <daos_types.h>
+#include <daos/tse.h>
 
 #include <daos/common.h>
 #include <daos/pool.h>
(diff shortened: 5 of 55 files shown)


ashleypittman:
@mchaarawi this is technically an API change but did we intend to export daos/tse.h?  If there are functions there we want exported then we should move them to daos_task.h IMO.

mchaarawi:
> @mchaarawi this is technically an API change but did we intend to export daos/tse.h? If there are functions there we want exported then we should move them to daos_task.h IMO.

not really. daos_task is something different than tse. tse by itself is unrelated to DAOS and meant as a separate package. some libraries use tse without daos. but daos_task is different and depends on DAOS.

ashleypittman:
> > @mchaarawi this is technically an API change but did we intend to export daos/tse.h? If there are functions there we want exported then we should move them to daos_task.h IMO.
> 
> not really. daos_task is something different than tse. tse by itself is unrelated to DAOS and meant as a separate package. some libraries use tse without daos. but daos_task is different and depends on DAOS.

Does anything outside the daos tree use tse directly?

With this PR you can't include daos/tse.h without previously including daos_task.h which goes against what you're saying but I suspect that was the case anyway - I can check.

mchaarawi (changes requested):
tse is used by HDF5 and possible other packages, so this cannot land.

mchaarawi:
> > > @mchaarawi this is technically an API change but did we intend to export daos/tse.h? If there are functions there we want exported then we should move them to daos_task.h IMO.
> > 
> > 
> > not really. daos_task is something different than tse. tse by itself is unrelated to DAOS and meant as a separate package. some libraries use tse without daos. but daos_task is different and depends on DAOS.
> 
> Does anything outside the daos tree use tse directly?
> 
> With this PR you can't include daos/tse.h without previously including daos_task.h which goes against what you're saying but I suspect that was the case anyway - I can check.

daos_task.h depends on tse.h not the other way around. so this PR is not correct.

ashleypittman:
> tse is used by HDF5 and possible other packages, so this cannot land.

Ok, updated on that basis.  daos/tse.h was and still is includable without pulling in any daos_*.h headers.

src/common/tse.c
```diff
@@ -17,8 +17,10 @@
 
 #include <stdint.h>
 #include <pthread.h>
+#include <daos_task.h>
```
mchaarawi, line 20: where does this file require daos_task.h ?

src/include/daos/common.h
```diff
@@ -25,17 +25,21 @@
...
 #include <daos_errno.h>
-#include <daos/debug.h>
-#include <gurt/hash.h>
-#include <gurt/common.h>
-#include <cart/api.h>
 #include <daos_types.h>
 #include <daos_obj.h>
 #include <daos_prop.h>
 #include <daos_security.h>
+#include <daos_task.h>
```
mchaarawi, line 32: why add daos_task.h here?

src/include/daos/common.h
```diff
@@ -25,17 +25,21 @@
...
+
+#include <daos/debug.h>
+#include <gurt/hash.h>
+#include <gurt/common.h>
+#include <cart/api.h>
+
 #include <daos/profile.h>
 #include <daos/dtx.h>
 #include <daos/cmd_parser.h>
+#include <daos/tse.h>
```
mchaarawi, line 42: same, why is that needed here?

src/include/daos/dpar.h
```diff
@@ -9,10 +9,6 @@
 #include <stdbool.h>
 #include <inttypes.h>
 
-#ifdef __cplusplus
```
mchaarawi, line 12: why remove those?

src/include/daos/tse.h
```diff
@@ -15,18 +15,24 @@
...
-/* internal tse private data size (struct tse_task_private) */
+/** internal tse private data size (struct tse_task_private) */
 #define TSE_PRIV_SIZE		(TSE_TASK_ARG_LEN + 136)
-/* tse_task is used to track single asynchronous operation (8 bytes used for public members). */
+/** tse_task is used to track single asynchronous operation (8 bytes used for public members). */
 #define TSE_TASK_SIZE		(TSE_PRIV_SIZE + 8)
 
+/** A single task */
 typedef struct tse_task {
+	/** the result */
```
mchaarawi, line 31: that sounds like a useless comment.

src/include/daos/tse.h
```diff
@@ -15,18 +15,24 @@
...
+/** tse_task arguments max length (pthread_mutex_t is of different size between x86 and aarch64). */
 #define TSE_TASK_ARG_LEN	(840 + sizeof(pthread_mutex_t))
-/* internal tse private data size (struct tse_task_private) */
+/** internal tse private data size (struct tse_task_private) */
 #define TSE_PRIV_SIZE		(TSE_TASK_ARG_LEN + 136)
-/* tse_task is used to track single asynchronous operation (8 bytes used for public members). */
+/** tse_task is used to track single asynchronous operation (8 bytes used for public members). */
 #define TSE_TASK_SIZE		(TSE_PRIV_SIZE + 8)
 
+/** A single task */
```
mchaarawi, line 29: that sounds like a useless comment.

src/include/daos_api.h
```diff
@@ -16,6 +16,8 @@
 extern "C" {
 #endif
 
+#include <daos_types.h>
+
```
mchaarawi, line 20: so even though those headers are installed, i believe we discussed at one point that really the user should just include daos.h for the daos API, then add daos_fs.h, or others for the higher level APIs vs being able to add each header. so your changes to make each public header byitself pull the other needed ones is not necessary.

mchaarawi (changes requested):

src/include/daos/dpar.h
ashleypittman: This is an internal header file so doesn't need c++ protection.

tse.h however does not have these guards in but I'm not sure if it's worth adding them.

src/include/daos/tse.h
ashleypittman: Our documentation requires all structures and entries to be documented, some text is needed here, feel free to suggest something better.

src/include/daos/mgmt.h
```diff
@@ -10,9 +10,9 @@
 #ifndef __DC_MGMT_H__
 #define __DC_MGMT_H__
 
-#include <daos/common.h>
-#include <daos/tse.h>
```
daltonbohning, line 14: This seems improper, or at least not how I'm familiar with imports. Generally, if a file needs a definition, it should import the header that contains that definition.
For example, `mgmt.h` uses `tse_task_t`, which is defined in `tse.h`. So `mgmt.h` should include `tse.h` directly instead of implicitly relying on other imports to import those definitions, which could change.

src/include/daos/common.h
```diff
@@ -25,14 +25,17 @@
...
 #include <daos_errno.h>
-#include <daos/debug.h>
-#include <gurt/hash.h>
-#include <gurt/common.h>
-#include <cart/api.h>
 #include <daos_types.h>
 #include <daos_obj.h>
 #include <daos_prop.h>
 #include <daos_security.h>
+#include <daos/tse.h>
```
daltonbohning, line 32: `daos/common.h` includes `daos/tse.h`, but `daos/common.h` doesn't actually need the definitions in `daos/tse.h`. So this seems improper to me

src/include/daos/mgmt.h
ashleypittman: You're correct.  I started this PR assuming that tse.h was a private header and trying to remove it's special handing but as @mchaarawi pointed it it's part of the API and it's special handing is warranted.  An early change I made incorrectly meant that daos_task.h neeed to included before daos/tse.h and these changes were to support that.

What you're saying about includes only being included to satisfy the header they're included in is good practice and this PR is an attempt to improve on that for the public headers but we do tend to have a few header files that just catch everything and I didn't mean to exacerbate that.

I've backed those changes out now and this PR is smaller as a result, thank you.

ashleypittman: The other issue here is that we should also be using <> for public headers and "" for internal ones, plus including system headers first, then daos public, then daos private but I don't think any of our code does that.

We've also got a few headers which are included both with or without a directory in places, daos_test.h is included as "<daos_test>" 25 times, <daos_test.h> 6 times and "<suite/daos_test.h>" once.

src/include/daos/container.h
```diff
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2023 Intel Corporation.
+ * (C) Copyright 2016-2024 Intel Corporation.
```
daltonbohning, line 2: erroneous copyright update after changes were backed out. Here and others

src/include/daos/pool.h
```diff
@@ -12,13 +12,15 @@
...
+#include <daos_pool.h>
+#include <daos_task.h>
+
 #include <daos/common.h>
 #include <gurt/hash.h>
 #include <daos/pool_map.h>
 #include <daos/rsvc.h>
-#include <daos/tse.h>
-#include <daos_types.h>
-#include <daos_pool.h>
```
daltonbohning, line 21: This file should still `include <daos/tse.h>` right?

src/include/daos/task.h
```diff
@@ -12,7 +12,6 @@
 
 #include <daos_types.h>
 #include <daos_api.h>
-#include <daos/tse.h>
```
daltonbohning, line 15: And this one?

src/object/obj_internal.h
```diff
@@ -14,7 +14,6 @@
 #include <stdint.h>
 #include <daos/common.h>
 #include <daos/event.h>
-#include <daos/tse.h>
```
daltonbohning, line 17: ditto

daltonbohning (commented):
Please do correct me if I'm wrong about how includes should be organized

src/include/daos/pool.h
ashleypittman: On what basis?  That in uses some of the defines in daos/tse.h I assume?  We have to assume that every file that includes this already includes daos/tse.h via some route.

It's easy enough to run my new unit test on the private headers as well, I'll update this PR with a few more code changes to allow all of them to be included but I won't enable this by default as it means the test would need to access abt.h and also we have daos/qat.h which we don't seem to include from anywhere today.

I've also found that using the macros from gurt/debug.h depends on gurt/common.h but you can't include gurt/common.h before gurt/debug.h so you have to include gurt/debug.h then gurt/common.h or just gurt/common.h

src/include/daos/container.h
ashleypittman: I'll do another pass and fix these when I'm done with other changes.

src/include/daos/pool.h
daltonbohning: > On what basis? That in uses some of the defines in daos/tse.h I assume?

Right.

>  We have to assume that every file that includes this already includes daos/tse.h via some route.

That works accidentally, but not in the general sense. `daos/pool.h` now has a silent dependency on `daos/tse.h`.


> the macros from gurt/debug.h depends on gurt/common.h

That issue is exactly the kind of issue I'm trying to avoid. If "the macros from gurt/debug.h depends on gurt/common.h" then `gurt/debug.h` should include `gurt/common.h`

ashleypittman: > > On what basis? That in uses some of the defines in daos/tse.h I assume?
> 
> Right.
> 
> > We have to assume that every file that includes this already includes daos/tse.h via some route.
> 
> That works accidentally, but not in the general sense. `daos/pool.h` now has a silent dependency on `daos/tse.h`.

Having testing more now I don't think it does, `daos/pool.h` obviously pulls in `daos/tse.h` from one of the header files that it includes.  We could no doubt do some pruning of includes from c files but getting the headers right is the first step towards that.

> > the macros from gurt/debug.h depends on gurt/common.h
> 
> That issue is exactly the kind of issue I'm trying to avoid. If "the macros from gurt/debug.h depends on gurt/common.h" then `gurt/debug.h` should include `gurt/common.h`

It can't, `gurt/common.h` includes `gurt/debug.h`.  I'll deal with that another day.

daltonbohning: This is what I mean. Currently, `pool.h` has an _**implicit**_ dependency on `tse.h`. If, for whatever reason, `something.h` gets refactored and doesn't need `tse.h` anymore, then `pool.h` will be missing an include. And since `pool.h` already includes `tse.h`, I don't see the value in removing it.
include

daltonbohning: > We could no doubt do some pruning of includes from c files

I don't think the goal should be to reduce the includes to the minimal possible that allows us to build each header independently. We should include what we need in the headers that need it. Otherwise, we end up breaking unrelated files if we reorganize, like in my example above.

src/include/daos/common.h
```diff
@@ -1,5 +1,5 @@
 /**
```
mchaarawi, line 1: not quite sure why the change is needed in this file. could you explain?

src/include/daos/mgmt.h
```diff
@@ -10,9 +10,10 @@
...
 #include <daos_types.h>
+#include <daos/tse.h>
+
+#include <daos/common.h>
```
mchaarawi, line 16: the change here just seems to remove and re-add the header with a line in between? just revert this so this file is not changed in the PR then.

src/include/daos/sys_debug.h
```diff
@@ -10,9 +10,7 @@
 #ifndef __DC_DEBUG_H__
 #define __DC_DEBUG_H__
 
-#include <daos/common.h>
-#include <daos/tse.h>
-#include <daos_types.h>
+#include <daos_task.h>
```
mchaarawi, line 13: so why here add daos_task.h but other places like https://github.com/daos-stack/daos/pull/13613/files#diff-930d69e66aac2e78ef8537ea29d976fc69620cbdeee43d117662b9e6f7f1f036R14 you added aos/tse.h

seems inconsistent.

src/include/daos/tse.h
mchaarawi: "struct for a task object that is used for tracking an operation"

src/include/daos/tse.h
mchaarawi: "result of the task operation - valid after task completion"

src/include/daos/tse.h
```diff
@@ -36,37 +42,41 @@ typedef struct tse_task {
  * Track all of the tasks under a scheduler.
  */
 typedef struct {
+	/** the result */
```
mchaarawi, line 45: same as above, but this is for the schedule

src/include/daos/tse.h
```diff
@@ -36,37 +42,41 @@ typedef struct tse_task {
...
-	/* daos schedule internal */
+	/** daos schedule internal */
 	struct {
 		uint64_t	ds_space[48];
 	}			ds_private;
 } tse_sched_t;
 
+/** Callaback function for when a scheduler is done */
 typedef int (*tse_sched_comp_cb_t)(void *args, int rc);
+/** Callaback function for when a task is done */
```
mchaarawi, line 59: this comment is wrong.

src/include/daos_array.h
```diff
@@ -19,6 +19,10 @@
 extern "C" {
 #endif
 
+#include <daos_types.h>
+#include <daos_obj.h>
+#include <daos_obj_class.h>
+
```
mchaarawi, line 25: im not going to comment again on every one of those public headers and will just comment on his one.
so the idea is for user to include daos.h, and that pulls in all what's needed. so "fixing" all the headers to be able to be included by itself is not required.

src/include/daos_fs.h
```diff
@@ -21,8 +21,15 @@ extern "C" {
...
 #include <dirent.h>
+#include <inttypes.h>
 #include <sys/stat.h>
 
+#include <daos_types.h>
+#include <daos_obj.h>
+#include <daos_obj_class.h>
+#include <daos_array.h>
+#include <daos_cont.h>
+
```
mchaarawi, line 32: not required. expectation is that user pulls in daos.h then daos_fs.h

mchaarawi (changes requested):

mchaarawi (approved):

utils/utest.yaml
```diff
@@ -1,6 +1,9 @@
-#  (C) Copyright 2023 Intel Corporation.
+#  (C) Copyright 2023-2024 Intel Corporation.
 #
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
+- name: include checks
+  tests:
+    - cmd: ["utils/include_test.py"]
```
daltonbohning, line 6: This is another case where my might want a directory of pytest tests somewhere

daltonbohning (approved):

PR #13617, opened 2024-01-16
DAOS-12710 build: Add unit testing that header can be included.

ashleypittman:
Add a unit test to verify all installed headers can be included
standalone.

Remove scons check for cart and gurt headers for the same.

Required-githooks: true

### DIFF
--- a/utils/include_test.py
+++ b/utils/include_test.py
@@ -1,23 +1,22 @@
 #!/usr/bin/env python3
 """Check that all daos headers can be included stand-alone"""
 
+import json
 import os
 import subprocess  # nosec
 import tempfile
 
-INCLUDE_DIR = "install/include"
 
-
-def check_dir(sub_dir):
+def check_dir(include_dir, sub_dir):
     """Check all files in one directory"""
 
-    h_dir = INCLUDE_DIR
+    h_dir = include_dir
 
     if sub_dir:
-        h_dir = os.path.join(INCLUDE_DIR, sub_dir)
+        h_dir = os.path.join(include_dir, sub_dir)
     for entry in sorted(os.listdir(h_dir)):
         if not os.path.isfile(os.path.join(h_dir, entry)):
-            check_dir(entry)
+            check_dir(include_dir, entry)
             continue
         with tempfile.NamedTemporaryFile(suffix=".c", mode="w+t") as tf:
             header = entry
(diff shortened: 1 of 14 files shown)


utils/include_test.py
```diff
@@ -0,0 +1,46 @@
...
+            print(f"Header file {header} is OK.")
+
+
+def main():
+    """Check the whole tree"""
+
+    with open(".build_vars.json", "r") as ofh:
+        bc = json.load(ofh)
+
+    include_dir = f"{bc['PREFIX']}/include"
```
daltonbohning, line 39:
```suggestion
    include_dir = os.path.join(bc['PREFIX'], "include")
```

utils/include_test.py
```diff
@@ -0,0 +1,46 @@
...
+
+    check_dir(include_dir, None)
+    os.unlink("a.out")
```
daltonbohning, line 42: Nit. I think it's better to move this unlink to the helper function which generates it

utils/include_test.py
```diff
@@ -0,0 +1,46 @@
...
+            print(f"Checking {header}")
+            subprocess.run(["gcc", "-I", include_dir, tf.name], check=True)
+            print(f"Header file {header} is OK.")
+
+
+def main():
+    """Check the whole tree"""
+
+    with open(".build_vars.json", "r") as ofh:
+        bc = json.load(ofh)
```
daltonbohning, line 37: What does `bc` stand for? Could we use something more self-documenting?

utils/include_test.py
```diff
@@ -0,0 +1,46 @@
...
+def check_dir(include_dir, sub_dir):
+    """Check all files in one directory"""
+
+    h_dir = include_dir
+
+    if sub_dir:
+        h_dir = os.path.join(include_dir, sub_dir)
+    for entry in sorted(os.listdir(h_dir)):
+        if not os.path.isfile(os.path.join(h_dir, entry)):
+            check_dir(include_dir, entry)
```
daltonbohning, line 19: FYI
```suggestion
        if os.path.isdir(os.path.join(h_dir, entry)):
            check_dir(include_dir, entry)
```

daltonbohning, line 19: This only works recursively once, right? But that's okay because our includes aren't deep

utils/include_test.py
ashleypittman: Each compile test will create/overwrite it, removing it on each iteration would be extra work although possibly clearer as you say.

utils/include_test.py
ashleypittman: build_config?  I'll change it to bv to match the filename.

utils/include_test.py
ashleypittman: Correct.

ashleypittman:
Pr has been merged into https://github.com/daos-stack/daos/pull/13613 but I'll push feedback there.

PR #13644, opened 2024-01-22
DAOS-15055 tools: Add build host/time to JSON version

mjmac:
Include build timestamp in all builds. Only
include build host in non-release builds.

Change-Id: I2d5525c21da2537a9583e9dbd57265e34639913f
Required-githooks: true

### DIFF
--- a/src/control/SConscript
+++ b/src/control/SConscript
@@ -2,8 +2,8 @@
 # pylint: disable=too-many-locals
 import os
 import socket
-from datetime import datetime, timezone
 from binascii import b2a_hex
+from datetime import datetime, timezone
 from os import urandom
 from os.path import join
 
--- a/src/control/SConscript
+++ b/src/control/SConscript
@@ -1,6 +1,8 @@
 """Build DAOS Control Plane"""
 # pylint: disable=too-many-locals
 import os
+import socket
+from datetime import datetime, timezone
 from binascii import b2a_hex
 from os import urandom
 from os.path import join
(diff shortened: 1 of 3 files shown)


daltonbohning (dismissed):
I'm curious what the use case is for needing BuildHost?

mjmac:
> I'm curious what the use case is for needing BuildHost?

For non-release builds, it's useful for determining the provenance of binaries. If I don't have to guess, I'd rather not. :)

tanabarr (approved):

kjacque (approved):

daltonbohning (approved):

src/control/SConscript
```diff
@@ -43,13 +45,19 @@ def gen_build_id():
...
 
+    build_host = ''
+    if not is_release_build(benv):
+        build_host = socket.getfqdn()
+    build_time = datetime.now(timezone.utc).astimezone().isoformat()
     Import('daos_version', 'conf_dir')
     path = 'github.com/daos-stack/daos/src/control/build'
     return ' '.join([f'-X {path}.DaosVersion={daos_version}',
                      f'-X {path}.ConfigDir={conf_dir}',
+                     f'-X {path}.BuildTime={build_time}',
```
ashleypittman, line 59: This has the unfortunate side effect of causing an expensive re-build of the go components for every source build regardless of if anything has changed.

If there is a go version of the c __date__ or __time__ macros then you should use that, if not there's a way of telling scons that a subset of options shouldn't be compared against the cache when considering rebuilds, if you need me to I can look into this.

PR #13654, opened 2024-01-23
DAOS-15069 dfs: use memcpy instead of snprintf for readdir

mchaarawi:
- memcpy is more efficient to avoid scanning the entire kds string.
- remove an assert and replace with error when looking up "."

Features: dfs

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfs/dfs.c
+++ b/src/client/dfs/dfs.c
@@ -3790,7 +3790,7 @@ readdir_int(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
 			D_GOTO(out, rc = daos_der2errno(rc));
 
 		for (ptr = enum_buf, i = 0; i < number; i++) {
-			strncpy(dirs[key_nr].d_name, ptr, kds[i].kd_key_len);
+			memcpy(dirs[key_nr].d_name, ptr, kds[i].kd_key_len);
 			dirs[key_nr].d_name[kds[i].kd_key_len] = '\0';
 			ptr += kds[i].kd_key_len;
 
@@ -6539,8 +6539,6 @@ dfs_listxattr(dfs_t *dfs, dfs_obj_t *obj, char *list, daos_size_t *size)
 			continue;
 
 		for (ptr = enum_buf, i = 0; i < number; i++) {
-			int len;
-
 			if (strncmp("x:", ptr, 2) != 0) {
 				ptr += kds[i].kd_key_len;
 				continue;
@@ -6553,10 +6551,8 @@ dfs_listxattr(dfs_t *dfs, dfs_obj_t *obj, char *list, daos_size_t *size)
 			if (list_size < kds[i].kd_key_len - 2)
 				continue;
 
-			len = snprintf(ptr_list, kds[i].kd_key_len - 1, "%s",
-				       ptr + 2);
-			D_ASSERT(len >= kds[i].kd_key_len - 2);
-
+			memcpy(ptr_list, ptr + 2, kds[i].kd_key_len - 2);
+			ptr_list[kds[i].kd_key_len - 2] = '\0';
 			list_size -= kds[i].kd_key_len - 1;
 			ptr_list += kds[i].kd_key_len - 1;
 			ptr += kds[i].kd_key_len;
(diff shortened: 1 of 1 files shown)


yuvale2 (approved):
Thank you!

src/client/dfs/dfs.c
```diff
@@ -568,7 +568,7 @@ fetch_entry(dfs_layout_ver_t ver, daos_handle_t oh, daos_handle_t th, const char
 
 	/** TODO - not supported yet */
 	if (strcmp(name, ".") == 0)
-		D_ASSERT(0);
+		return ENOTSUP;
```
ashleypittman, line 571: This is progress but wasn't there a different ticket with a reprodyucer for this?  Without seeing that ticket could dfs_lookup_rel() skip over entiries named "." when doing the path walk, then the rerroducer should pass rather than exit gracefully as presumably it does with this fix.

src/client/dfs/dfs.c
```diff
@@ -3790,11 +3790,8 @@ readdir_int(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
 			D_GOTO(out, rc = daos_der2errno(rc));
 
 		for (ptr = enum_buf, i = 0; i < number; i++) {
-			int len;
-
-			len = snprintf(dirs[key_nr].d_name,
-				       kds[i].kd_key_len + 1, "%s", ptr);
-			D_ASSERT(len >= kds[i].kd_key_len);
+			strncpy(dirs[key_nr].d_name, ptr, kds[i].kd_key_len);
+			dirs[key_nr].d_name[kds[i].kd_key_len] = '\0';
```
ashleypittman, line 3794: I thought I'd been through daos and remove most of the places where snprintf() was used with "%s" but I must have missed this one.  I think here though as dfs isn't ever doing anything with the entry names then it could just treat them as a memory region and use memcpy().  strncpy() will do two extra steps, check the length of the input string and pad the output buffer with \0, neither of which are of benefit in this case.

The same fix should also be applied to the dfs_listxattr code I assume?

src/client/dfs/dfs.c
mchaarawi: we can leave that case for the ticket. this is fine for now

src/client/dfs/dfs.c
mchaarawi: agree, memcpy is better in this case.

src/client/dfs/dfs.c
```diff
@@ -3790,7 +3790,7 @@ readdir_int(dfs_t *dfs, dfs_obj_t *obj, daos_anchor_t *anchor, uint32_t *nr,
 			D_GOTO(out, rc = daos_der2errno(rc));
 
 		for (ptr = enum_buf, i = 0; i < number; i++) {
-			strncpy(dirs[key_nr].d_name, ptr, kds[i].kd_key_len);
+			memcpy(dirs[key_nr].d_name, ptr, kds[i].kd_key_len);
```
yuvale2, line 3793: Maybe check that `kds[i].kd_key_len < sizeof(dirs[key_nr].d_name)`? (I realize that this wasn't checked before, maybe it's guaranteed elsewhere)

yuvale2 (approved):

src/client/dfs/dfs.c
mchaarawi: If that condition would be false, then we have landed some data corruption in the container and this would be the least of the problems :-)

yuvale2: But if we don't detect it here, we'll have two data corruptions. :)

mchaarawi: my point was we probably won't even get here if there is data corruption in the container.
but sure, i can add that check if needed to repush. otherwise can add it in another PR im working on.

yuvale2: Adding in another PR sounds good to me.

yuvale2: Alternatively, if the check it's 100% redundant, a comment that explains this would help (it's not obvious unless you know the server side, I guess).

jolivier23 (approved):

src/client/dfs/dfs.c
ashleypittman: I think there is an issue here but adding in another PR seems the way to go.

PR #13655, opened 2024-01-23
DAOS-15068 engine: check IV refresh result

wangdi1:
Check IV refresh result during cart IV fetch, to
avoid further failure if the fetched value is corrupted.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/cart/crt_iv.c
+++ b/src/cart/crt_iv.c
@@ -1086,8 +1086,8 @@ handle_ivfetch_response(const struct crt_cb_info *cb_info)
 	}
 
 	/* In case of a failure, call on_refresh with NULL iv_value */
-	iv_ops->ivo_on_refresh(ivns, &input->ifi_key, 0, /* TODO: iv_ver */
-			       iv_value, false, rc, iv_info->ifc_user_priv);
+	rc = iv_ops->ivo_on_refresh(ivns, &input->ifi_key, 0, /* TODO: iv_ver */
+				    iv_value, false, rc, iv_info->ifc_user_priv);
 
 	if (iv_info->ifc_bulk_hdl)
 		crt_bulk_free(iv_info->ifc_bulk_hdl);
--- a/src/engine/server_iv.c
+++ b/src/engine/server_iv.c
@@ -509,6 +509,9 @@ ivc_on_refresh(crt_iv_namespace_t ivns, crt_iv_key_t *iv_key,
 	       crt_iv_ver_t iv_ver, d_sg_list_t *iv_value,
 	       bool invalidate, int refresh_rc, void *priv)
 {
+	if (refresh_rc != 0)
+		return refresh_rc;
+
 	return iv_on_update_internal(ivns, iv_key, iv_ver, iv_value, invalidate,
 				     true, refresh_rc, priv);
 }


src/engine/server_iv.c
```diff
@@ -509,6 +509,9 @@ ivc_on_refresh(crt_iv_namespace_t ivns, crt_iv_key_t *iv_key,
 	       crt_iv_ver_t iv_ver, d_sg_list_t *iv_value,
 	       bool invalidate, int refresh_rc, void *priv)
 {
+	if (refresh_rc != 0)
+		return refresh_rc;
+
```
frostedcmos, line 514: Hmm, I don't think it should return refresh_rc here but instead SUCCESS. callback implementation's return indicates the status of its execution, i dont think you should just return iv framework the same rc it gave you for this value.

src/cart/crt_iv.c
```diff
@@ -1086,8 +1086,8 @@ handle_ivfetch_response(const struct crt_cb_info *cb_info)
 	}
 
 	/* In case of a failure, call on_refresh with NULL iv_value */
-	iv_ops->ivo_on_refresh(ivns, &input->ifi_key, 0, /* TODO: iv_ver */
-			       iv_value, false, rc, iv_info->ifc_user_priv);
+	rc = iv_ops->ivo_on_refresh(ivns, &input->ifi_key, 0, /* TODO: iv_ver */
```
frostedcmos, line 1089: i dont think this is correct. rc is only used in this function later to determine the order of finalizing pending events on lines 1113-1120, and it should be using the rc from the fetch, not the success/failure of refreshing the value locally.
it functionally will do the right thing today, because of how you changed iv_on_refresh() implementation to return the same rc, but I also don't think that is a proper way, as it should return success instead.

frostedcmos (changes requested):

src/engine/server_iv.c
wangdi1: hmm, refresh_rc is the result of remote fetch, if it failed, then it suppose to return skip the refresh. I thought  we should return the failure if refresh does not succeed here. since refresh_rc is one of the input anyway?

src/cart/crt_iv.c
wangdi1: yes, rc will be used later indeed.  But if on_refresh failed, which also validate the fetch value actually,  then the return value should be updated here.   So there are two cases

if fetch failed, i.e. the original rc is not 0, then ivo_on_refresh will return the original rc anyway(see above change),  

if fetch succeeds, but ivo_on_refresh failed due to local validation or refresh, then rc should be updated and used later, then if the local engine will have to do IV remote fetch again, if it want to get the update value.

src/engine/server_iv.c
frostedcmos: no, not really. ivo_on_refresh() is just a notification to the IV member that the value is being updated. If IV value is unavailable/invalid then its still a 'valid' update, even if you chose to ignore it in your case. 
As far as IV goes ivo_on_refresh() errors should not affect any logic. So I think the right way would be to return DER_SUCCESS if refresh_rc != 0, and keep iv logic as it is today

frostedcmos: The concern is that you are changing IV behavior where previously ivo_on_refresh() errors were not meant to affect the fetch path (they totally should fail the update path) but now they do. I don't have issues if you want to change ivo_on_refresh() semantics to mean that any error should stop the fetch process and return appropriate on_refresh() error, but then this is incomplete and needs to address all other instances where ivo_on_refresh() is called.

In your particular case, if the fetch itself was successful, but for some reason local ivo_on_refresh() failed on the current node in the middle, then it will cause the wrong order of pending events finalization (as it would assume overall fetch failed).

wangdi1: >but then this is incomplete and needs to address all other instances where ivo_on_refresh() is called.

Not sure I follow here, but some of ivo_on_fresh() return values are being checked in crt_iv.c

> In your particular case, if the fetch itself was successful, but for some reason local ivo_on_refresh() failed on the current node in the middle, then it will cause the wrong order of pending events finalization (as it would assume overall fetch failed).

As I understand,  if IV fetch failed in the middle of the path (after get the value from the root), those nodes already got the value(before failure) will keep those value anyway (by setting entry valid),   those nodes (after failure)  will be notified by failure anyway.  I do not understand why it has to be succeed along the path, how it is different with network failure. or I miss sth?

PR #13657, opened 2024-01-24
DAOS-15073 pool: avoid deadlock when put last ref on stopped pool

Nasf-Fan:
Sometimes, when a pool is stopped, some user may be still using the pool with reference held on such pool. For the user that is holding the last reference on the stopped pool, when it releases reference, it will trigger pool_free_ref that will stop (with block mode) the container that may wait current user (ULT) to exit.

To avoid potential deadlock, we will use another independent ULT to release the reference (on the stopped pool) asynchronously and make current ULT to go ahead.

### Gatekeeper:

### DIFF
--- a/src/include/daos/lru.h
+++ b/src/include/daos/lru.h
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2016-2022 Intel Corporation.
+ * (C) Copyright 2016-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -156,6 +156,15 @@ static inline void
 daos_lru_ref_add(struct daos_llink *llink)
 {
 	llink->ll_ref++;
+}
+
+/**
+ * Return true if the caller is the last user of the LRU element.
+ */
+static inline bool
+daos_lru_is_last_user(struct daos_llink *llink)
+{
+	return llink->ll_ref <= 2;
 }
 
 #endif
--- a/src/pool/srv_target.c
+++ b/src/pool/srv_target.c
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2016-2023 Intel Corporation.
+ * (C) Copyright 2016-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -637,6 +637,16 @@ pool_obj(struct daos_llink *llink)
 	return container_of(llink, struct ds_pool, sp_entry);
 }
 
+static inline void
+pool_put_sync(void *args)
+{
+	struct ds_pool	*pool = args;
+
+	D_ASSERT(pool != NULL);
+	D_ASSERT(dss_get_module_info()->dmi_xs_id == 0);
+	daos_lru_ref_release(pool_cache, &pool->sp_entry);
+}
+
 struct ds_pool_create_arg {
 	uint32_t	pca_map_version;
 };
@@ -857,7 +867,7 @@ ds_pool_lookup(const uuid_t uuid, struct ds_pool **pool)
 	*pool = pool_obj(llink);
 	if ((*pool)->sp_stopping) {
 		D_DEBUG(DB_MD, DF_UUID": is in stopping\n", DP_UUID(uuid));
-		ds_pool_put(*pool);
+		pool_put_sync(*pool);
 		*pool = NULL;
 		return -DER_SHUTDOWN;
 	}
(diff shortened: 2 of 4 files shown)


src/pool/srv_target.c
```diff
@@ -876,9 +886,25 @@ ds_pool_get(struct ds_pool *pool)
...
-	daos_lru_ref_release(pool_cache, &pool->sp_entry);
+	int	rc;
+
+	/*
+	 * Someone has stopped the pool. Current user may be the one that is holding the last
+	 * reference on the pool, then drop such reference will trigger pool_free_ref() as to
+	 * stop related container that may wait current user (ULT) to exit. To avoid deadlock,
+	 * let's use independent ULT to drop the reference asynchronously and make current ULT
+	 * to go ahead.
+	 */
```
NiuYawei, line 897: This comment is ambiguous, could you write down the deadlock example of aggregation ULT calls ds_cont_csummer_init() which synchronously waiting IV fetch?

src/pool/srv_target.c
```diff
@@ -876,9 +886,25 @@ ds_pool_get(struct ds_pool *pool)
...
+	if (unlikely(pool->sp_stopping)) {
```
NiuYawei, line 898: It's better to introduce a daos_lru_ref_peek() to check the reference and only creating ULT when (stopping && ref == 2).

Nasf-Fan: I Has ever considered that, but it means that the caller needs to understand how the LRU reference works, for example, it is "ref == 2" instead of "ref == 1". That is too hack since it expose the LRU internal details to the  callers. Personally, I do not like such usage, because once LRU logic changed in the further, such hack may be bug. On the other hand, stopping the pool with others users referencing the pool is relative case, then even if we start another ULT, it will not cause much overhead.

src/pool/srv_target.c
Nasf-Fan: Done.

src/pool/srv_target.c
NiuYawei: Well, if that's your concern, you could export a function for checking if it's holding the last reference.

We actually does compare vos object refcount with 2 to see if it's the last refcount. (see vos_obj_hold()), I agree it looks not that clean.

Nasf-Fan: It is not only unclean, but also seems unnecessary, because once "stopping" flag is set, nobody can find such pool (in cache) any longer. So only the former users can continuously access such pool until released the reference. So three will be at most "N - 1" (N is the system level pool users by race count, very limited) additional ULTs to be created that will neither affect the correctness nor slowdown the system too much. That is different from vos_obj_hold() case.

NiuYawei: After "stopping" is set, the pool can still be found until the last reference of the ds_pool being dropped and being removed from cache. See ds_pool_lookup(), the ds_pool is still in cache, so every ds_pool_lookup() will have to call ds_pool_put() to generate an ULT.

Please look into ds_pool_stop() too, it could yield and wait on many things before putting the last reference. 

What's your major concern of adding a simple function to check the last reference?

Nasf-Fan: > After "stopping" is set, the pool can still be found until the last reference of the ds_pool being dropped and being removed from cache. See ds_pool_lookup(), the ds_pool is still in cache, so every ds_pool_lookup() will have to call ds_pool_put() to generate an ULT.
 
I have changed ds_pool_lookup() logic in the patch, it will not call ds_pool_put(), instead, it will call pool_put_sync() when the found pool has "stopping" flag, that will not generate additional ULT.

> Please look into ds_pool_stop() too, it could yield and wait on many things before putting the last reference.

What's the matter for that?

> What's your major concern of adding a simple function to check the last reference?

As I mentioned above, it is not worth to introduce a hack and potential buggy "solution" but without benefit.

src/include/daos/lru.h
```diff
@@ -158,4 +158,13 @@ daos_lru_ref_add(struct daos_llink *llink)
 	llink->ll_ref++;
 }
 
+/**
+ * Return true if the caller is the last user of the LRU element.
+ */
+static inline bool
+daos_lru_is_last_user(struct daos_llink *llink)
+{
+	return llink->ll_ref <= 2;
```
NiuYawei, line 167: This relies on that hash table holds one refcount, and looks current daos lru implementation assumes the hash table is always created without D_HASH_FT_EPHEMERAL, but I didn't see where how it's guranteed.

If the patch needs be refreshed, please add a D_ASSERT(!(feats & D_HASH_FT_EPHEMERAL)) in daos_lru_cache_create().

NiuYawei (approved):

liuxuezhao (approved):
this may need to backport to 2.4?

Nasf-Fan:
> this may need to backport to 2.4?

Yes, 2.4 has the same trouble in theory.

PR #13660, opened 2024-01-24
DAOS-15075 test: Skip the device role check.

rpadma2:
Test-tag: test_csum_error_logging always_passes test_tiers

Summary:
- The test seems to check device role string which is either set to NULL or NA.
- Looks like there is no real benefit in checking the status of the string.
- Removing the device role check.
- Added the if condition check only using the role_bits information.

### Gatekeeper:


src/tests/ftest/checksum/csum_error_logging.py
```diff
@@ -72,7 +72,7 @@ def test_csum_error_logging(self):
                 if not device['tgt_ids']:
                     self.log_step('Skipping device without targets on {}'.format(device['uuid']))
                     continue
-                if device['roles'] and not int(device['role_bits']) & 1:
+                if not int(device['role_bits']) & 1:
```
daltonbohning, line 75: If we don't care about `roles` should we remove it from here too?
https://github.com/daos-stack/daos/blob/8c91a8a138099a3d2164e4582891e5dfd93438f5/src/tests/ftest/checksum/csum_error_logging.py#L65-L71

rpadma2: I noticed this place... Since it just checks whether there is an entry for the device [keys], I thought it is ok to keep it for now. If you feel we should remove it, I am fine to remove it.

daltonbohning: I think if we don't need `roles` for this test, we should remove it. But not a hard blocker.

daltonbohning (approved):

src/tests/ftest/checksum/csum_error_logging.py
rpadma2: Done.

daltonbohning (dismissed):

shimizukko (dismissed):

src/tests/ftest/checksum/csum_error_logging.py
```diff
@@ -62,17 +62,17 @@ def test_csum_error_logging(self):
...
                 self.log.info(
-                    'Host %s device: uuid=%s, targets=%s, role=%s, role_bits=%s',
-                    host, device['uuid'], device['tgt_ids'], device['roles'], device['role_bits'])
+                    'Host %s device: uuid=%s, targets=%s, role_bits=%s',
+                    host, device['uuid'], device['tgt_ids'], device['role_bits'])
                 if not device['tgt_ids']:
                     self.log_step('Skipping device without targets on {}'.format(device['uuid']))
                     continue
-                if device['roles'] and not int(device['role_bits']) & 1:
+                if not int(device['role_bits']) & 1:
```
tanabarr, line 75: I don't think this will work for MD-on-PMem mode where 'role_bits' will be 0. Either use approach in https://github.com/daos-stack/daos/pull/13667 or 
```suggestion
                if int(device['role_bits']) > 0 and not (int(device['role_bits']) & 1):
```

tanabarr (changes requested):

src/tests/ftest/checksum/csum_error_logging.py
tanabarr: It looks like the CI run didn't execute the desired test: https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-13660/2/pipeline/

tanabarr (changes requested):

src/tests/ftest/checksum/csum_error_logging.py
```diff
@@ -62,17 +62,17 @@ def test_csum_error_logging(self):
         host_devices = get_dmg_smd_info(dmg.storage_query_list_devices, 'devices')
         for host, devices in host_devices.items():
             for device in devices:
-                for entry in ('uuid', 'tgt_ids', 'role_bits', 'roles'):
+                for entry in ('uuid', 'tgt_ids', 'role_bits'):
```
phender, line 65: This change is not needed. We still want to verify the `roles` entry.
```suggestion
                for entry in ('uuid', 'tgt_ids', 'role_bits', 'roles'):
```

src/tests/ftest/checksum/csum_error_logging.py
```diff
@@ -62,17 +62,17 @@ def test_csum_error_logging(self):
...
                     if entry not in device:
                         self.fail(
                             'Missing {} info from dmg storage query list devices'.format(entry))
                 self.log.info(
-                    'Host %s device: uuid=%s, targets=%s, role=%s, role_bits=%s',
-                    host, device['uuid'], device['tgt_ids'], device['roles'], device['role_bits'])
+                    'Host %s device: uuid=%s, targets=%s, role_bits=%s',
+                    host, device['uuid'], device['tgt_ids'], device['role_bits'])
```
phender, line 71: Similarly this change should be reverted:
```suggestion
                    'Host %s device: uuid=%s, targets=%s, role=%s, role_bits=%s',
                    host, device['uuid'], device['tgt_ids'], device['roles'], device['role_bits'])
```

src/tests/ftest/checksum/csum_error_logging.py
phender, line 75: The problem in the ticket is `roles` used to be assigned to `''` when unset and now it's `NA`, so we only need to support treating `NA` the same as when `device['roles']` is not set:
```suggestion
                if (device['roles'] != 'NA') and not int(device['role_bits']) & 1:
```

phender (changes requested):

src/tests/ftest/checksum/csum_error_logging.py
rpadma2: I can agree with this... I can revert the change.

src/tests/ftest/checksum/csum_error_logging.py
rpadma2: This change I don't feel why we need to do? Yes, it was NULL before and now it is NA... What is the point in checking it? At some point if the value changes, the script will once again will be broken. Also, if we want to re-test with previous version of DAOS, the test will fail. I am not clear about the reasoning behind this check. I can add the check but I don't see the use.

tanabarr: I think the correct solution is to keep your original changes and just adjust the check to `if int(device['role_bits']) > 0 and not (int(device['role_bits']) & 1):` then you are not relying on any changes in syntax of `roles` and instead just relying on the behaviour of `role_bits` which is less likely to change.

tanabarr: In MD-on-PMem mode, role_bits will always be 0. Roles are only assigned in MD-on-SSD mode so given that roles == "NA" whenever role_bits = 0 you can just test if `bits > 0 && !(bits & 1)`.

rpadma2: Thanks @tanabarr ... I think this seems to be the correct suggestion. I will proceed with this change.

src/tests/ftest/checksum/csum_error_logging.py
rpadma2: I am going with Tom's change to make the test more flexible and generic.

src/tests/ftest/checksum/csum_error_logging.py
rpadma2: I am going with Tom's change to make the test more flexible and generic.

daltonbohning (approved):

src/tests/ftest/checksum/csum_error_logging.py
daltonbohning: IMO It doesn't look like this test is intended to verify the output of the command, so I don't think we should add extraneous checks. If we do want to verify, then we should go back to check if it's `NA`.

tanabarr (approved):
Happy to land as is assuming we have coverage on "roles" elsewhere as per Phil's concern. Otherwise we could use https://github.com/daos-stack/daos/pull/13667/files#diff-408e10113afc30cd28bc6b24fafefd57ca80e4c63de476ff69f1aad8d8daac70R75 (that PR unfortunately needs pragma adjusted as it didn't run any tests).

rpadma2:
> Happy to land as is assuming we have coverage on "roles" elsewhere as per Phil's concern. Otherwise we could use https://github.com/daos-stack/daos/pull/13667/files#diff-408e10113afc30cd28bc6b24fafefd57ca80e4c63de476ff69f1aad8d8daac70R75 (that PR unfortunately needs pragma adjusted as it didn't run any tests).

Looks like we have coverage for testing the roles in ftest:
./src/tests/ftest/control/dmg_storage_query.py:        self.log_step('Verify storage device targets and roles')
./src/tests/ftest/control/dmg_storage_query.py:                    for key in ('tgt_ids', 'roles'):
./src/tests/ftest/server/storage_tiers.py:                    storage_roles = self.params.get("roles", namespace, None)
./src/tests/ftest/server/storage_tiers.py:                    if storage_roles:
./src/tests/ftest/server/storage_tiers.py:                        data["roles"] = storage_roles


I updated the Test-tag from just calling test_csum_error_logging  to include test_tiers and all_passes... Hope this will invoke the test_csum_error_logging.

shimizukko (approved):

phender (approved):

PR #13672, opened 2024-01-25
DAOS-14985 test: Address unit test failure due to merge

tanabarr:
Bad merge involving PR-13614 resulted in broken bdevScanEngine go unit test.

- Fix merge conflict affecting unit test by adding roles to bdev in scan.
- Improve reliability of unit test by sorting response content.
- Move filtering of bdevs in bad state so that retry logic is not inadvertently triggered.

Features: control
Required-githooks: true

### Before requesting gatekeeper:

* [x] Two review approvals and any prior change requests have been resolved.
* [x] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [x] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [x] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [x] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate watchers.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
  * [ ] Quick-build or Quick-functional is not used.
* [ ] Fix the commit message upon landing. Check the standard [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments). Edit it to create a single commit. If necessary, ask submitter for a new summary.

PR #13614, opened 2024-01-16
DAOS-14985 bio: Handle NVMe unplugged in list devs

tanabarr:
dmg storage query list-devices fails if run after a NVMe device is
unplugged during a physical hot-remove. The failure is due to
UNPLUGGED NVMe device state resulting in a SMD object not being
populated with a NVMe controller reference. Unplugged devices should
be reported in list-devices and ignored in storage scan so handle
UNPLUGGED device state without returning an error.

Features: control nvme
Required-githooks: true

### Before requesting gatekeeper:

* [x] Two review approvals and any prior change requests have been resolved.
* [x] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [x] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [x] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [x] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate watchers.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
  * [ ] Quick-build or Quick-functional is not used.
* [ ] Fix the commit message upon landing. Check the standard [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments). Edit it to create a single commit. If necessary, ask submitter for a new summary.

PR #13675, opened 2024-01-26
Revert "DAOS-14985 bio: Handle NVMe unplugged in list devs (#13614)"

mjmac:
This reverts commit e86a3b39dff156e9b438f945920c839428c0b2f2.

### DIFF
--- a/src/bio/bio_device.c
+++ b/src/bio/bio_device.c
@@ -460,10 +460,7 @@ alloc_ctrlr_info(uuid_t dev_id, char *dev_name, struct bio_dev_info *b_info)
 	D_ASSERT(b_info->bdi_ctrlr == NULL);
 
 	if (dev_name == NULL) {
-		D_DEBUG(DB_MGMT,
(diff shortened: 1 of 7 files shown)


tanabarr (changes requested):
Please review and land fix https://github.com/daos-stack/daos/pull/13672 rather than reverting. I've done various manual verification checks on the fix and it is solid. Also to be clear the problem wasn't specifically due to this change it was the combination of this and another PR landing close together.

mjmac:
I pushed this after discussion with other gatekeepers and particularly in response to (#13674), as I was concerned that skipping a broken unit test may have been hiding other problems in the patch. I don't have a problem with landing #13672 instead, if there is a high level of confidence that it will fix the problem and not cause more problems. If there is any doubt, though, it seems advisable to just land this one to take some pressure off and really test the new code.

tanabarr:
https://github.com/daos-stack/daos/pull/13672 landed to master

PR #13678, opened 2024-01-29
DAOS-14348 client: remove dcache entry when a dir is deleted

wiliamhuang:
Features: pil4dfs

Dcache can expire to achieve better consistency. Add a unit test.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/docs/user/filesystem.md
+++ b/docs/user/filesystem.md
@@ -974,6 +974,13 @@ libpil4dfs intercepting summary for ops on DFS:
 [op_sum ]  5003
 ```
 
+### Change dir cache timeout with env
+
+Directory caching is employed for better performance. The default timeout is 10 seconds. User can change it if necessary. The unit is second and an integer should be provided.
+```
+$ export D_IL_DCACHE_TIMEOUT=5
+```
+
 ### Limitations of using libpil4dfs
 Stability issues: This is a preview version. Some features are not implemented yet. Many APIs are involved in libpil4dfs. There may be bugs, uncovered/not intercepted functions, etc. 
 
--- a/src/client/dfuse/pil4dfs/int_dfs.c
+++ b/src/client/dfuse/pil4dfs/int_dfs.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2022-2023 Intel Corporation.
+ * (C) Copyright 2022-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -77,6 +77,10 @@
 #define FAKE_ST_INO(path)   (d_hash_string_u32(path, strnlen(path, DFS_MAX_PATH)))
 
 #define MAX_EQ              64
+
+/* default dir cache time-out in seconds*/
+#define DCACHE_TIMEOUT      10
+static uint16_t               dcache_timeout;
 
 /* In case of fork(), only the parent process could destroy daos env. */
 static bool                   context_reset;
@@ -262,10 +266,11 @@ get_eqh(daos_handle_t *eqh);
  * entity.
  */
 struct dir_hdl {
-	d_list_t   entry;
-	dfs_obj_t *oh;
-	size_t     name_size;
-	char       name[];
+	d_list_t        entry;
+	dfs_obj_t      *oh;
+	struct timespec t_expire;
+	size_t          name_size;
+	char            name[];
 };
 
 static inline struct dir_hdl *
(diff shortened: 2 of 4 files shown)


docs/user/filesystem.md
```diff
@@ -974,6 +974,13 @@ libpil4dfs intercepting summary for ops on DFS:
 [op_sum ]  5003
 ```
 
+### Change dir cache timeout with env
+
+Directory caching is employed for better performance. The default timeout is 10 seconds. User can change it if necessary. The unit is second and an integer should be provided.
+```
+$ export D_IL_DCACHE_TIMEOUT=5
```
ashleypittman, line 981: This is the equivalent setting as dfuse-dentry-dir-time so it would be good to use a matching name.  Perhaps PIL4DFS_DENTRY_DIR_TIME given as this is a env var rather than a container attribute.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5565,6 +5602,12 @@ init_myhook(void)
 		eq_count_max = MAX_EQ;
 	}
 
+	rc = d_getenv_uint64_t("D_IL_DCACHE_TIMEOUT", &dcache_timeout_loc);
+	if (rc != -DER_NONEXIST)
+		dcache_timeout = (uint16_t)dcache_timeout_loc;
+	else
+		dcache_timeout = DCACHE_TIMEOUT;
```
ashleypittman, line 5609: I'd probably make dcache_timeout be a uint32_t rather than uint16_t but don't call a function for one size and then change it, just use a 32 bit number throughout and set it to a sensible default when declared.
```suggestion
	rc = d_getenv_uint32_t("D_IL_DCACHE_TIMEOUT", &dcache_timeout);
	if (rc != -DER_NONEXIST)
	        ... handle fatal error.
```

ashleypittman (commented):
There are lots of gaps here, the hash table will still grow forever except for the case of directories that are deleted locally, what about pruning of expired entries to speed lookup, what about entries that have been deleted by another process, what about entries which are longer supersets of ones that have been deleted.

The whole approach to using a hash table in this way is wrong, I don't think you should be making small adjustments to the implementation when this whole area will need to be rewritten anyway.

mchaarawi:
> There are lots of gaps here, the hash table will still grow forever except for the case of directories that are deleted locally, what about pruning of expired entries to speed lookup, what about entries that have been deleted by another process, what about entries which are longer supersets of ones that have been deleted.
> 
> The whole approach to using a hash table in this way is wrong, I don't think you should be making small adjustments to the implementation when this whole area will need to be rewritten anyway.

I have not reviewed this yet to comment on the PR itself. it needs a rebase btw..
But i think the gaps are clear here in this hash implementation. the ultimate goal is to move the caching in the DFS layer where we can have a more proper implementation, but for now we need to address some of those gaps as we discussed several times in the WG. and it's better for now to address those in separate, smaller, and tested PRs, and document the ones that we don't get to for 2.6.

ashleypittman:
> > There are lots of gaps here, the hash table will still grow forever except for the case of directories that are deleted locally, what about pruning of expired entries to speed lookup, what about entries that have been deleted by another process, what about entries which are longer supersets of ones that have been deleted.
> > The whole approach to using a hash table in this way is wrong, I don't think you should be making small adjustments to the implementation when this whole area will need to be rewritten anyway.
> 
> I have not reviewed this yet to comment on the PR itself. it needs a rebase btw.. But i think the gaps are clear here in this hash implementation. the ultimate goal is to move the caching in the DFS layer where we can have a more proper implementation, but for now we need to address some of those gaps as we discussed several times in the WG. and it's better for now to address those in separate, smaller, and tested PRs, and document the ones that we don't get to for 2.6.

I'm not 100% sure I understand your comment but let me rephrase what I meant above to see if that clarifies things.

The current hash table in pil4dfs contains paths, that is long strings with multiple intermediate directory names in it, so it might have "a", "a/b" and "a/b/c" all in the same hash table.  Adding logic to expire "a/b" from the hash table based on age does not move us forward, there is no point spending time improving anything around the hash table until it's been updated to store only dentries (parent + name pairs) and have pil4dfs doing the path walk properly for every lookup.  This code cannot exit it's technology preview state until that's resolved as an absolute minimum and this PR does nothing to move us closer to that point.

ashleypittman:
> > The current hash table in pil4dfs contains paths, that is long strings with multiple intermediate directory names in it, so it might have "a", "a/b" and "a/b/c" all in the same hash table. Adding logic to expire "a/b" from the hash table based on age does not move us forward, there is no point spending time improving anything around the hash table until it's been updated to store only dentries (parent + name pairs) and have pil4dfs doing the path walk properly for every lookup. This code cannot exit it's technology preview state until that's resolved as an absolute minimum and this PR does nothing to move us closer to that point.
> 
> Thanks for the clarification. Yea I do agree with this part of your comment. I did not review the PR to see the age part of it and was just going by the title that deleted entries need to be removed from the cache. @wiliamhuang you do need to move away from this cache implementation that is used in IOR (use-case specific) and have finer grain lookups of each entry from the cache. As for the cache growing continuously, consistency with other processes, or background expiry of entries, this can be addressed in different PRs.

By "the cache growing continuously" what I mean is that with this PR in the example above "a/b" would only be removed from cache after in had expired during a lookup of "a/b" (which would presumably then put it back in the cache), there's no provision for removing expired entries that are not looked up although these would still be in the hash table which has an ongoing cost.

But - we need to correct what's in the hash table first before putting effort into optimising how the hash table is accessed or managed etc.

daltonbohning (commented):
ftest code changes LGTM. Will defer on the C changes

mchaarawi:
I believe this can be closed / deleted @wiliamhuang ?

PR #13681, opened 2024-01-29
DAOS-14826 test: Add a dfuse/pil4dfs test for bash fd usage.

ashleypittman:
Add a test for pil4dfs that tries to replicate what configure
scripts do with file descriptors.

Required-githooks: true
Test-tag: FdTest
Skip-init-tests: true

### DIFF
--- a/src/tests/ftest/dfuse/bash_fd.py
+++ b/src/tests/ftest/dfuse/bash_fd.py
@@ -1,0 +1,165 @@
+"""
+  (C) Copyright 24 Intel Corporation.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+"""
+
+import os
+import stat
+
+from dfuse_test_base import DfuseTestBase
+from run_utils import run_remote
+
+OUTER = """#!/bin/bash
+
+set -uex
+
+[ -d out_dir ] && rm -rf out_dir
+[ -d e_out_dir ] && rm -rf e_out_dir
+mkdir out_dir
+
+cd out_dir
+
+.././bash_fd_inner.sh
+
+cd -
+
+grep . out_dir/*
+
+mkdir e_out_dir
+echo first file > e_out_dir/out_file
+echo first file >> e_out_dir/out_file
+
+echo second file > e_out_dir/other_file
+echo five >> e_out_dir/other_file
+echo six >> e_out_dir/other_file
+
+diff --new-file --recursive out_dir e_out_dir
+"""
+
+INNER = """#!/bin/bash
+
+set -ue
+
+echo Hello, about to perform some bash I/O testing similar to "configure" scripts.
+
+# Open file for read/write access.
+exec 3<>out_file
+echo first file >&3
+
+ls -l /proc/$$/fd
+
+# Open file in /proc and hold it open.
+exec 4</proc/self/maps
+
+# Open file for write access.
+exec 5>other_file
+echo second file >&5
+
+# Duplicate fd.
+exec 6>&5
+echo five >&6
+
+ls -l /proc/$$/fd
+
+# Close fds as output file descriptors.
+exec 4>&-
+exec 5>&-
+echo six >&6
+exec 6>&-
+
+echo first file >&3
+exec 3>&-
+
+ls -l /proc/$$/fd
+
+exit 0
+"""
+
+
+class FdTest(DfuseTestBase):
+    """Base FdTest test class.
+
+    "avocado: recursive
+    """
+
+    def run_bashfd(self, il_lib=None):
+        """Run a shell script which opens and writes to files.
+
+        This attempts to replicate the way that configure scripts manipulate fds in bash.
+        """
+
+        if il_lib is not None:
+            lib_path = os.path.join(self.prefix, "lib64", il_lib)
+            env_str = f"export LD_PRELOAD={lib_path}; "
+        else:
+            env_str = ""
+
+        self.add_pool(connect=False)
+        self.add_container(self.pool)
+        mount_dir = f"/tmp/{self.pool.identifier}_daos_dfuse"
+        self.start_dfuse(self.hostlist_clients, self.pool, self.container, mount_dir=mount_dir)
+
+        fuse_root_dir = self.dfuse.mount_dir.value
+
+        with open(os.path.join(fuse_root_dir, "bash_fd_inner.sh"), "w") as fd:
+            fd.write(INNER)
+
+        os.chmod(os.path.join(fuse_root_dir, "bash_fd_inner.sh"), stat.S_IXUSR | stat.S_IRUSR)
+
+        with open(os.path.join(fuse_root_dir, "bash_fd_outer.sh"), "w") as fd:
+            fd.write(OUTER)
+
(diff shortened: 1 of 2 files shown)


ashleypittman:
Results are at https://build.hpdd.intel.com/job/daos-stack/job/daos//view/change-requests/job/PR-13681/3/testReport/FTEST_dfuse/

Test passed for dfuse and ioil, failed for pil4dfs as expected.  The test is weekly so could probably land before https://github.com/daos-stack/daos/pull/13532

wiliamhuang (approved):

wiliamhuang:
The new test works fine in my local test.

mchaarawi:
please rebase with latest master and the test-tag

src/tests/ftest/dfuse/bash_fd.py
```diff
@@ -0,0 +1,165 @@
...
+    """Base FdTest test class.
+
+    :avocado: recursive
+    """
+
+    def run_bashfd(self, il_lib=None):
+        """Run a shell script which opens and writes to files.
+
+        This attempts to replicate the way that configure scripts manipulate fds in bash.
+        """
```
daltonbohning, line 90: We try to document all arguments
```suggestion
    def run_bashfd(self, il_lib=None):
        """Run a shell script which opens and writes to files.

        This attempts to replicate the way that configure scripts manipulate fds in bash.

        Args:
            il_lib (str, optional): interception library to run with. Defaults to None
        """
```

src/tests/ftest/dfuse/bash_fd.py
```diff
@@ -0,0 +1,165 @@
...
+        if il_lib is not None:
+            lib_path = os.path.join(self.prefix, "lib64", il_lib)
+            env_str = f"export LD_PRELOAD={lib_path}; "
+        else:
+            env_str = ""
+
+        self.add_pool(connect=False)
+        self.add_container(self.pool)
+        mount_dir = f"/tmp/{self.pool.identifier}_daos_dfuse"
+        self.start_dfuse(self.hostlist_clients, self.pool, self.container, mount_dir=mount_dir)
```
daltonbohning, line 101: We're trying to move away from these "magic" pool and container functions and instead use explicit variables. Also, you could let the framework generic a mount_dir for you.
```suggestion
        pool = self.get_pool(connect=False)
        container = self.get_container(pool)
        self.start_dfuse(self.hostlist_clients, pool, container)
```

src/tests/ftest/dfuse/bash_fd.py
```diff
@@ -0,0 +1,165 @@
...
+        with open(os.path.join(fuse_root_dir, "bash_fd_outer.sh"), "w") as fd:
+            fd.write(OUTER)
+
+        os.chmod(os.path.join(fuse_root_dir, "bash_fd_outer.sh"), stat.S_IXUSR | stat.S_IRUSR)
+
+        cmd = f"cd {fuse_root_dir}; ./bash_fd_outer.sh"
+
+        result = run_remote(self.log, self.hostlist_clients, env_str + cmd)
+        if not result.passed:
+            self.fail(f'"{cmd}" failed on {result.failed_hosts}')
```
daltonbohning, line 119: Just pointing out that this creates _**local**_ files and then issues a _**remote**_ command to execute them. This only works because the test runner is also the only client. That's okay for now, but eventually (not sure when) we'd like to not rely on the test runner being a client.

src/tests/ftest/dfuse/bash_fd.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # stop dfuse
+        self.stop_dfuse()
+        # destroy container
+        self.container.destroy()
+        # destroy pool
+        self.pool.destroy()
```
daltonbohning, line 126: This generally isn't necessary. It will all be handled in teardown

src/tests/ftest/dfuse/bash_fd.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        self.run_bashfd()
+
+    def test_bashfd_ioil(self):
+        """
+
+        Test Description:
+            Test a typical I/O pattern for bash based configure scripts.
+
+        :avocado: tags=all,pr,full_regression
```
daltonbohning, line 147: pr and weekly? Is that intentional?

src/tests/ftest/dfuse/bash_fd.py
```diff
@@ -0,0 +1,165 @@
...
+echo first file >&3
+exec 3>&-
+
+ls -l /proc/$$/fd
+
+exit 0
+"""
+
+
+class FdTest(DfuseTestBase):
```
daltonbohning, line 80: Can we please put `Dfuse` or similar in the name? Out of context, `FdTest` doesn't give much description. And these are harder to change after the file is merged and we have results under the old name

src/tests/ftest/dfuse/bash_fd.yaml
```diff
@@ -0,0 +1,21 @@
...
+    0:
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos
+  system_ram_reserved: 1
+pool:
+  scm_size: 1000000000
```
daltonbohning, line 17: This could be more generic, and could use a human-readable size
```suggestion
  size: 1G
```

src/tests/ftest/dfuse/bash_fd.yaml
```diff
@@ -0,0 +1,21 @@
...
+  control_method: dmg
```
daltonbohning, line 18: This is an outdated option we don't use anymore
```suggestion
```

src/tests/ftest/dfuse/bash_fd.py
ashleypittman: Is there a "write remote file" option in ftest yet?  There's no other way to write fuse tests now, and there never has been.

src/tests/ftest/dfuse/bash_fd.py
ashleypittman: Forgot to update this one, no that's not intentional.

src/tests/ftest/dfuse/bash_fd.py
daltonbohning: No, there isn't. But it's something we'll likely need in the future. For now, I think this approach is okay.

src/tests/ftest/dfuse/bash_fd.py
```diff
@@ -0,0 +1,160 @@
...
+        """
+
+        Test Description:
+            Test a typical I/O pattern for bash based configure scripts.
+
+        :avocado: tags=all,full_regression
+        :avocado: tags=vm
+        :avocado: tags=pil4dfs,dfs
+        :avocado: tags=DFuseFdTest,test_bashfd_pil4dfs
+        """
```
daltonbohning, line 159: Do you want the pil4dfs variant also tagged with `dfuse`?

daltonbohning (approved):

src/tests/ftest/dfuse/bash_fd.py
ashleypittman: No.

wiliamhuang (approved):

mchaarawi (approved):

daltonbohning (approved):

PR #13692, opened 2024-01-30
DAOS-14981 test: Have daosbuild test buuld 2.4 version of daos.

ashleypittman:
Test-tag: test_dfuse_daos_build_wt_il
Sigted-off-by: Ashley Pittman <ashley.m.pittman@intel.com>

Required-githooks: true

### DIFF
--- a/src/tests/ftest/dfuse/daos_build.py
+++ b/src/tests/ftest/dfuse/daos_build.py
@@ -1,5 +1,5 @@
 """
-  (C) Copyright 2020-2023 Intel Corporation.
+  (C) Copyright 2020-2024 Intel Corporation.
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 """
@@ -208,6 +208,7 @@ class DaosBuild(DfuseTestBase):
 
         cmds = ['python3 -m venv {}/venv'.format(mount_dir),
                 'git clone https://github.com/daos-stack/daos.git {}'.format(build_dir),
+                'git -C {} checkout release/2.4'.format(build_dir),
                 'git -C {} submodule init'.format(build_dir),
                 'git -C {} submodule update'.format(build_dir),
                 'python3 -m pip install pip --upgrade',


knard-intel (approved):
LGTM

src/tests/ftest/dfuse/daos_build.py
```diff
@@ -208,6 +208,7 @@ def run_build_test(self, cache_mode, intercept=False, run_on_vms=False):
 
         cmds = ['python3 -m venv {}/venv'.format(mount_dir),
                 'git clone https://github.com/daos-stack/daos.git {}'.format(build_dir),
+                'git -C {} checkout release/2.4'.format(build_dir),
```
daltonbohning, line 211: We'll have the same problem when master branches to the next 2.x release.
If we wanted a more generic solution, we could rely on
```
$ daos version -j
{
  "response": {
    "name": "DAOS CLI",
    "version": "2.5.100",
    "revision": "0aae80f5909b8ec63f9b46dd31bb56cf88d23fa2",
    "dirty": true
  },
  "error": null,
  "status": 0
}
```
And either
1. Checkout the exact revision hash. Which means we would build the exact same commit that the PR/branch is running on.
2. Map the version to master, 2.4, etc.

If we do that, it should go into master as well, so no need to block this PR. And I could make that change myself. Thoughts?

daltonbohning (approved):

src/tests/ftest/dfuse/daos_build.py
ashleypittman: This has only come up as a problem now despite having branched in June, I agree that if/when we find we need to do this on release/2.6 then we should generalise it but for now I don't see it as being a major benefit.

ashleypittman:
@daos-stack/daos-gatekeeper this is not a cherry-pick but a change that's only needed on the branch.  The affected test passed.

src/tests/ftest/dfuse/daos_build.py
daltonbohning: But prior to this change, the release/2.4 version of this test was building master, correct? So even if it wasn't failing, it wasn't quite doing the right thing

daltonbohning:
> @daos-stack/daos-gatekeeper this is not a cherry-pick but a change that's only needed on the branch. The affected test passed.

I think we still need approved to merge first

src/tests/ftest/dfuse/daos_build.py
ashleypittman: There's no reason for it to match.  We could just as well put in a SHA hash here.

ashleypittman:
> > @daos-stack/daos-gatekeeper this is not a cherry-pick but a change that's only needed on the branch. The affected test passed.
> 
> I think we still need approved to merge first

Didn't occur to me for a 2.4 only ticket.  I'll ask.

jolivier23 (approved):

PR #13702, opened 2024-01-30
DAOS-14822, 14820 dfuse: Better manage container connections.

ashleypittman:
Reference count timeout values.
Keep track of detached containers for re-use.

Required-githooks: true


daltonbohning:
Is this ready for review?

ashleypittman:
> Is this ready for review?

I thought it was but it had a test failure overnight. It may need further revision on the grace period but otherwise it's ready for review yes

utils/node_local_test.py
```diff
@@ -2721,6 +2717,92 @@ def test_xattr(self):
...
+        Access container path check ino
+        """
+
+        # Magic value for how long to sleep.  This needs to be long enough for entry timeout,
+        # whatever grace period is configured and some additional to let the eviction process
+        # happen.  This makes for a very long test, plus in addition there is no way of telling
+        # if the eviction has actually happened.
+        # A second test pool would be useful here so we could use filesystem query to check the
+        # inode count.
+        sleep_time = 307 + 5 + 10
```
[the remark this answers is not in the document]
ashleypittman: Yes, that's true.  The `5` in the above calculation is wrong now as well.  This tests in this object run in parallel and this one is configured to start first so it doesn't add 15 minutes to run-time but it does add about six as it's the first test to start and the last to finish.  The other thought I had was to use the fault injection code to set the thresholds differently for the dfuse instance running this test, I'll look into that.

src/client/dfuse/dfuse_core.c
```diff
@@ -671,204 +756,178 @@ dfuse_set_default_cont_cache_values(struct dfuse_cont *dfc)
...
+		D_SPIN_LOCK(&dfuse_info->di_lock);
+		d_list_for_each_entry(dfcc, &dfp->dfp_historic, dfcc_entry) {
+			if (dfcc->dfcc_ino == 0)
+				continue;
+			if (uuid_compare(dfcc->dfcc_uuid, dfc->dfc_uuid) != 0)
+				continue;
+			dfc->dfs_ino = dfcc->dfcc_ino;
+			break;
+		}
+		if (dfc->dfs_ino == 0) {
```
mchaarawi, line 892: not quite sure what this block is doing. it was not found in the historic list, what is the next list (di_pool_historic) for?

src/client/dfuse/dfuse_core.c
```diff
@@ -671,204 +756,178 @@ dfuse_set_default_cont_cache_values(struct dfuse_cont *dfc)
...
+	DFUSE_TRA_DEBUG(dfp, "New cont " DF_UUIDF " in pool " DF_UUIDF, DP_UUID(dfc->dfc_uuid),
+			DP_UUID(dfp->dfp_uuid));
+
+	if (dfc->dfs_ino == 0) {
+		dfc->dfs_ino      = atomic_fetch_add_relaxed(&dfuse_info->di_ino_next, 1);
+		dfc->dfc_save_ino = true;
+		DFUSE_TRA_INFO(dfc, "Assigned new inode number %ld", dfc->dfs_ino);
+
+	} else {
+		DFUSE_TRA_INFO(dfc, "Reusing inode number %ld", dfc->dfs_ino);
```
mchaarawi, line 929: do we need to update the historic list by removing the entry from there?

ashleypittman: No.  In theory we could remove it from the historic list and it would all work, but then when this new container was closed then it would get added to the historic list so I don't see any point adding/removing it.  It would complicate the _ch_free() logic on close to make the decision there as well.

src/client/dfuse/dfuse_core.c
ashleypittman: It searches the list of historic container for the current pool and then if not found there then it needs to look at the historic pools.

If you connected to p1, opened and closed c1 then c1 would be on the open pool historic list.

If you connected to p1, opened and closed c1, disconnected from p1, connected to p1 and opened c1 then c1 would not be in the historic list for the open pool descriptor but would be on the historic list.  On p1 being closed if it's already on the historic pool list then it splices the list from the p1 descriptor it's closing to the p1 descriptor on the historic list.

This is actually the exact case the new (but disabled) test case covers.

ashleypittman:
This has previously been passing all tests but it failed the test_dfuse_daos_build_wt_il test in the last push with possibly an invalid handle.  I've merged with master and am running just this test.

src/client/dfuse/dfuse_core.c
mchaarawi: got it

mchaarawi (approved):
looks ok to me. although i do not see anything changed since my last review when you had mentioned test were failing with invalid handle, which sounds somehow related to this PR as it manages handles.
please make sure that we don't introduce some intermittent issue by running that failing test a few times to make sure.

ashleypittman:
> looks ok to me. although i do not see anything changed since my last review when you had mentioned test were failing with invalid handle, which sounds somehow related to this PR as it manages handles. please make sure that we don't introduce some intermittent issue by running that failing test a few times to make sure.

That was affecting all PRs, and was caused by a cart change.  Master is working again as of https://github.com/daos-stack/daos/pull/13817

jolivier23 (approved):

PR #13744, opened 2024-02-05
DAOS-15121 test: fix run_remote log without timeout

daltonbohning:
Test-tag: pr DaosCoreTestDfuse
Skip-unit-tests: true

Fix bug in the log.debug call when there is not a timeout.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/tests/ftest/util/run_utils.py
+++ b/src/tests/ftest/util/run_utils.py
@@ -405,7 +405,7 @@ def run_remote(log, hosts, command, verbose=True, timeout=120, task_debug=False,
     task.set_info("ssh_options", "-oForwardAgent=yes")
     if verbose:
         if timeout is None:
-            log.debug("Running on %s without a timeout: %s", hosts, timeout, command)
+            log.debug("Running on %s without a timeout: %s", hosts, command)
         else:
             log.debug("Running on %s with a %s second timeout: %s", hosts, timeout, command)
     task.run(command=command, nodes=hosts, timeout=timeout)


ashleypittman (approved):

src/tests/ftest/util/run_utils.py
```diff
@@ -405,7 +405,7 @@ def run_remote(log, hosts, command, verbose=True, timeout=120, task_debug=False,
     task.set_info("ssh_options", "-oForwardAgent=yes")
     if verbose:
         if timeout is None:
-            log.debug("Running on %s without a timeout: %s", hosts, timeout, command)
+            log.debug("Running on %s without a timeout: %s", hosts, command)
```
phender, line 408: Thoughts on adding a `test_harness_unit_run_remote_no_timeout` test case to `harness/unit.py` to verify this arguably simple change?

ashleypittman: I'd sooner ask why are we allowing commands without timeouts.

daltonbohning: The default timeout is 120, so a test would have to explicitly pass `timeout=None`. And the entire test itself is still bound by a timeout. Most cases probably don't need `timeout=None`, but I think it's okay to support it.

daltonbohning: That said, maybe tests should just adapt to always having a timeout

daltonbohning: > test_harness_unit_run_remote_no_timeout

I think that's overkill because we should trip over the error this PR fixes if `verbose=True` and `timeout is None`. And those don't tend to be dynamic for an individual test. I suspect something subtle changed somewhere for this test to trip over this now.

ashleypittman: > That said, maybe tests should just adapt to always having a timeout

If the run_remote() timeout times out then that's a test failure, if the overall test times out then that's a test error.  It's a minor point but there's no case for passing in timeout=None as a timeout value here.

daltonbohning: > there's no case for passing in timeout=None as a timeout value here

It appears to just be used for legacy cases where commands just didn't specify a timeout in the past. E.g. in `fio_utils` the timeout is defaulted to `None`. There isn't a sensible default here because adjusting the fio command options could very easily cause the command to run for a long time. But I think it's better for tests to just specify a timeout.

In short, I agree with you. But it would take a little digging to ensure we catch every case

ashleypittman: Let's not waste time adding additional test harnesses for configurations that should be dis-allowed then.

daltonbohning:
Closed in favor of #13743

PR #13750, opened 2024-02-05
DAOS-15138 dfs: improve output of daos fs get-attr

mchaarawi:
Instead of just returning the oclass of new objects, return the oclass of the directory, the creation oclass of new directories, and the creation oclass of new files.

Features: dfs
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/cmd/daos/filesystem.go
+++ b/src/control/cmd/daos/filesystem.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2021-2023 Intel Corporation.
+// (C) Copyright 2021-2024 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -32,11 +32,11 @@ type fsCmd struct {
 	FixRoot        fsFixRootCmd        `command:"fix-root" description:"Relink root object in the container"`
 	FixEntry       fsFixEntryCmd       `command:"fix-entry" description:"Fix Entries in case the type or chunk size were corrupted"`
 	Copy           fsCopyCmd           `command:"copy" description:"copy to and from a POSIX filesystem"`
-	SetAttr        fsSetAttrCmd        `command:"set-attr" description:"set fs attributes"`
-	GetAttr        fsGetAttrCmd        `command:"get-attr" description:"get fs attributes"`
-	ResetAttr      fsResetAttrCmd      `command:"reset-attr" description:"reset fs attributes"`
-	ResetChunkSize fsResetChunkSizeCmd `command:"reset-chunk-size" description:"reset fs chunk size"`
-	ResetObjClass  fsResetOclassCmd    `command:"reset-oclass" description:"reset fs obj class"`
+	SetAttr        fsSetAttrCmd        `command:"set-attr" description:"set default object class and/or chunk size"`
+	GetAttr        fsGetAttrCmd        `command:"get-attr" description:"get default object class and chunk size"`
+	ResetAttr      fsResetAttrCmd      `command:"reset-attr" description:"reset default creation attributes on a directory"`
+	ResetChunkSize fsResetChunkSizeCmd `command:"reset-chunk-size" description:"reset default creation chunk size on a directory"`
+	ResetObjClass  fsResetOclassCmd    `command:"reset-oclass" description:"reset default creation object class on a directory"`
 	DfuseQuery     fsDfuseQueryCmd     `command:"query" description:"Query dfuse for memory usage"`
 	DfuseEvict     fsDfuseEvictCmd     `command:"evict" description:"Evict object from dfuse"`
 }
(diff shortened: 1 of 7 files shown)


src/include/daos_fs.h
```diff
@@ -98,6 +98,10 @@ typedef struct {
 	daos_oclass_id_t	doi_oclass_id;
 	/** chunk size */
 	daos_size_t		doi_chunk_size;
+	/** In case of dir, return default object class for dirs created in that dir */
+	daos_oclass_id_t        doi_dir_oclass_id;
+	/** In case of dir, return default object class for files created in that dir */
+	daos_oclass_id_t        doi_file_oclass_id;
```
daltonbohning, line 104: Do we care to make this backward compatible? E.g. if an application is built against 2.4.x, but then daos is updated and the app not rebuilt

daltonbohning (approved):
Sanity tested locally and output looks sane

mchaarawi:
the failure is https://daosio.atlassian.net/browse/DAOS-14232
auto_oc_selection.py actually passed, but it timed out in the container destroy phase as it ran out of time. so unrelated to this PR

src/include/daos_fs.h
mchaarawi: this is used in the daos tools and I'm not aware of other middleware or apps using that.
that being said, just adding these fields at the end of the struct won't cause issues. the main issue is the change in behavior of what doi_oclass_id is for dirs. but as i said this is just used today by DAOS itself.

daltonbohning: > just adding these fields at the end of the struct won't cause issues.

I think it would if an application was compiled against an older version and, e.g., the size of the struct was 1K but now with the additional fields the size is 2K. So when the application initializes an instance of the struct, it uses too small of a size. Then when dfs tries to access the struct members it's a memory error.

IIRC, we ran into a similar issue before. Anyway, I'll defer to your judgement

mchaarawi: yea sorry i was thinking compiling issues. you are right though.
but im not worried about it as this was mainly written for the tools.

jolivier23: Probably safe for now.   As we use more of these management APIs though, we will probalby need to be more careful and bump major/minor API versions

jolivier23 (dismissed):

src/include/daos_fs.h
mchaarawi: Yea i thought about bumping the API major version, but it would be an overkill in this case because of how this is used today. a minor version bump probably makes more sense, but i will evaluate that with 2.6 release to see if it was already done or not to avoid bumping it multiple times.

src/control/cmd/daos/filesystem.go
```diff
@@ -262,27 +262,69 @@ func (cmd *fsGetAttrCmd) Execute(_ []string) error {
...
+	}
+
 	if cmd.JSONOutputEnabled() {
-		jsonAttrs := &struct {
-			ObjClass  string `json:"oclass"`
-			ChunkSize uint64 `json:"chunk_size"`
-		}{
-			ObjClass:  C.GoString(&oclassName[0]),
-			ChunkSize: uint64(attrs.doi_chunk_size),
+		if C.mode_is_dir(cmode) {
```
tanabarr, line 281: nit: I think it's tidier to return early so that you can reduce nesting:
```
	if C.mode_is_dir(cmode) {
	        if !cmd.JSONOutputEnabled() {
                	cmd.Infof("Object Class = %s", C.GoString(&oclassName[0]))
                        cmd.Infof("Directory Creation Object Class = %s", C.GoString(&diroclassName[0]))
          		cmd.Infof("File Creation Object Class = %s", C.GoString(&fileoclassName[0]))
          		cmd.Infof("File Creation Chunk Size = %d", attrs.doi_chunk_size)
           	        return nil
          	}

		return cmd.OutputJSON(struct {
			ObjAttr struct {
				ObjClass string `json:"oclass"`
			} `json:"Object"`
			DirAttr struct {
				DirObjClass  string `json:"dir_oclass"`
				FileObjClass string `json:"file_oclass"`
				ChunkSize    uint64 `json:"chunk_size"`
			} `json:"Directory"`
		}{
			ObjAttr: struct {
				ObjClass string `json:"oclass"`
			}{
				ObjClass: C.GoString(&oclassName[0]),
			},
			DirAttr: struct {
				DirObjClass  string `json:"dir_oclass"`
				FileObjClass string `json:"file_oclass"`
				ChunkSize    uint64 `json:"chunk_size"`
			}{
				FileObjClass: C.GoString(&diroclassName[0]),
				DirObjClass:  C.GoString(&fileoclassName[0]),
				ChunkSize:    uint64(attrs.doi_chunk_size),
			},
		}, nil)
	}

        if !cmd.JSONOutputEnabled() {
               	cmd.Infof("Object Class = %s", C.GoString(&oclassName[0]))
		cmd.Infof("Object Chunk Size = %d", attrs.doi_chunk_size)
	        return nil
	}
	
	return cmd.OutputJSON(&struct {
		ObjClass  string `json:"oclass"`
		ChunkSize uint64 `json:"chunk_size"`
	}{
		ObjClass:  C.GoString(&oclassName[0]),
		ChunkSize: uint64(attrs.doi_chunk_size),
	}, nil)
}
```

tanabarr (approved):
non-blocking code style suggestion

src/include/daos_fs.h
ashleypittman: APIs/ABIs can be tricky but extending structs is one of the few things that you can get away with.  Doing a scan for API (header) changes and ensuring appropriate major/minor bumps wouldn't be a bad thing to have on the TODO list at the beginning of a release cycle.

src/control/cmd/daos/filesystem.go
```diff
@@ -262,27 +262,69 @@ func (cmd *fsGetAttrCmd) Execute(_ []string) error {
...
+			jsonAttrs := struct {
+				ObjAttr struct {
+					ObjClass string `json:"oclass"`
+				} `json:"Object"`
```
ashleypittman, line 285: I think in general we've standardised on lower case keys in json haven't we?
```suggestion
				} `json:"object"`
```

src/control/cmd/daos/filesystem.go
```diff
@@ -262,27 +262,69 @@ func (cmd *fsGetAttrCmd) Execute(_ []string) error {
...
+				DirAttr struct {
+					DirObjClass  string `json:"dir_oclass"`
+					FileObjClass string `json:"file_oclass"`
+					ChunkSize    uint64 `json:"chunk_size"`
+				} `json:"Directory"`
```
ashleypittman, line 290:
```suggestion
				} `json:"directory"`
```

utils/node_local_test.py
```diff
@@ -1854,6 +1854,28 @@ def check_dfs_tool_output(output, oclass, csize):
     return True
 
 
+def check_dir_attr(output, oclass, dir_oclass, file_oclass, csize):
```
ashleypittman, line 1857: The other option would be to change the calling code to pass the run_daos_cmd() the json=True option, in which case it would return a dict of the json contents.

ashleypittman (changes requested):
- 1 just to get some feedback on Mixed-case json keys.

src/control/cmd/daos/filesystem.go
mchaarawi: is that documented somewhere?

utils/node_local_test.py
mchaarawi: sure.. can keep that to an improvement patch for later

src/control/cmd/daos/filesystem.go
ashleypittman: Not that I'm aware of but I have seen PRs to move things to lower-case and I have seen PRs to update tests after the case was updated and not all test-cases were changed at the time.  A quick git grep of the source shows 78 matches for `json:"` all of which are lower case throughout.

mchaarawi: but it's not a "standard" if it's not documented or at least gofmt does not throw an error on it.

ashleypittman: I wouldn't expect gofmt to throw an error on it, to go it's just a string.

https://daosio.atlassian.net/browse/DAOS-10579 is an old ticket but covers making changes so that snake_case rather than CamelCase is used throughout.  I'm not aware of this being writing down in a document as a policy anywhere but it seems clear to me that there was a decision made on this at some point and it seems reasonable to stick to it.

mchaarawi: i will argue again with your "standardize" comment. sorry but how is it clear that a decision is made on this just because there was a ticket created 2 years ago..
ill push the change, but if there is some standard as you claim for this that we want applied, it has to be clearly documented in the coding rules. otherwise it's just a preference and **not** a standard.

ashleypittman (approved):

tanabarr (approved):

src/control/cmd/daos/filesystem.go
ashleypittman: Perhaps standardised was too strong a term to use here.  Catching issues like this is part of the review process and documentation/process improvements are always welcome.

mchaarawi:
force landing after unit test stage passed since only change in go to change to lower case json attrs.

PR #13803, opened 2024-02-14
DAOS-15135 object: handle non-continuous ranks array for coll operation

Nasf-Fan:
When scan the object targets for collective operation, the ranks array for related pool may be not continuous. Under such case the scan logic needs to extend the ranks array dynamically to avoid accessing invalid DRAM or falling into dead loop.

Features: ec rebuild

### Gatekeeper:

### DIFF
--- a/src/object/cli_coll.c
+++ b/src/object/cli_coll.c
@@ -132,12 +132,15 @@ obj_coll_prep_one(struct coll_oper_args *coa, struct dc_object *obj,
 	if (rc != 0 || (shard->do_rebuilding && !coa->coa_for_modify))
 		goto out;
 
-	/*
-	 * More ranks joined after obj_coll_oper_args_init().
-	 * The object layout may be changed, need to re-scan.
-	 */
-	if (unlikely(shard->do_target_rank >= coa->coa_dct_cap))
-		D_GOTO(out, rc = -DER_AGAIN);
+	/* The ranks for the pool may be not continuous, let's extend the coa_dcts array. */
+	if (unlikely(shard->do_target_rank >= coa->coa_dct_cap)) {
+		D_REALLOC_ARRAY(dct, coa->coa_dcts, coa->coa_dct_cap, shard->do_target_rank + 4);
+		if (dct == NULL)
+			D_GOTO(out, rc = -DER_NOMEM);
+
+		coa->coa_dcts = dct;
+		coa->coa_dct_cap = shard->do_target_rank + 4;
+	}
 
 	dct = &coa->coa_dcts[shard->do_target_rank];
 	dct->dct_rank = shard->do_target_rank;
@@ -402,21 +405,14 @@ dc_obj_coll_punch(tse_task_t *task, struct dc_object *obj, struct dtx_epoch *epo
 	int				 rc;
 	int				 i;
 
-re_scan:
 	rc = obj_coll_oper_args_init(coa, obj, true);
 	if (rc != 0)
 		goto out;
 
 	for (i = 0; i < obj->cob_shards_nr; i++) {
 		rc = obj_coll_prep_one(coa, obj, map_ver, i);
-		if (rc != 0) {
-			if (unlikely(rc == -DER_AGAIN)) {
-				obj_coll_oper_args_fini(coa);
-				goto re_scan;
-			}
-
+		if (rc != 0)
 			goto out;
-		}
 	}
 
 	rc = obj_coll_oper_args_collapse(coa, &tgt_size);
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -7087,7 +7087,6 @@ dc_obj_query_key(tse_task_t *api_task)
 
 	/* Some optimization for get dkey collectively since 2.6 which version is 10. */
 	if (api_args->flags & DAOS_GET_DKEY && grp_nr > 1 && dc_obj_proto_version >= 10) {
-re_scan:
 		rc = obj_coll_oper_args_init(&obj_auxi->cq_args.cqa_coa, obj, false);
 		if (rc != 0)
 			goto out_task;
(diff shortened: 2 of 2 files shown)


src/object/cli_coll.c
```diff
@@ -132,12 +132,15 @@ obj_coll_prep_one(struct coll_oper_args *coa, struct dc_object *obj,
...
+	if (unlikely(shard->do_target_rank >= coa->coa_dct_cap)) {
+		D_REALLOC_ARRAY(dct, coa->coa_dcts, coa->coa_dct_cap, shard->do_target_rank + 4);
+		if (dct == NULL)
+			D_GOTO(out, rc = -DER_NOMEM);
+
+		coa->coa_dcts = dct;
+		coa->coa_dct_cap = shard->do_target_rank + 4;
+	}
 
 	dct = &coa->coa_dcts[shard->do_target_rank];
```
wangdi1, line 145: I thought this is just to get an empty slot from coa_dcts?  No?  using rank as index seems improper here, especially if the pool is created by some high number ranks.

Nasf-Fan:
Rebuild_29 timeout because of DAOS-15165, not related with the patch.

src/object/cli_coll.c
Nasf-Fan: The coa_dcts is continuous array. It is fine to get empty slot as long as it is not hole. Using rank as the index is the most convenient way. There will be obj_coll_oper_args_collapse after object layout parsing, that will filter out the empty slots.

wangdi1: If there are thousands rank, and creating pool with a few higher number of ranks, then it will allocate a lot empty slot? I thought you can easily get the empty & available slot?

wangdi1: Btw: you need run the test by Features: rebuild ec to make sure pool extend tests are being run.

Nasf-Fan: > If there are thousands rank, and creating pool with a few higher number of ranks, then it will allocate a lot empty slot? I thought you can easily get the empty & available slot?

The ranks array is pre-allocated before scanning the object layout, at that time, we do not know which ranks may be used by the object shards. Under such case, using the rank as array index to pre-allocate a large array (with empty slots at the beginning) may be the most simple way.
It is not impossible to dynamically allocate the array during scanning the object layout, but we still need to use rank as the array index to efficiently locate the rank_slot.

Nasf-Fan: > Btw: you need run the test by Features: rebuild ec to make sure pool extend tests are being run.

OK, I will re-run the test with ec_rebuild enabled.

wangdi1: > > If there are thousands rank, and creating pool with a few higher number of ranks, then it will allocate a lot empty slot? I thought you can easily get the empty & available slot?
> 
> The ranks array is pre-allocated before scanning the object layout, at that time, we do not know which ranks may be used by the object shards. Under such case, using the rank as array index to pre-allocate a large array (with empty slots at the beginning) may be the most simple way. It is not impossible to dynamically allocate the array during scanning the object layout, but we still need to use rank as the array index to efficiently locate the rank_slot.

I thought you can pre-allocate the array by object layout size, then get one empty slot from the array each time, why this is expensive.  Or I miss sth?

What if the pool is created by 1,2,3,1002,   then the array size is 1002?

Nasf-Fan: Object layout is per target based, not per rank based. Its count may be quite larger than rank count. So pre-allocating against object layout may be not save space. On the other hand, if the ranks array (or target array) is not continuous, how can you efficiently locate related slot with index? Maintain a sorted array may be optional, but it still needs some data movement, so cannot say it is more efficient than the continuous array.

> What if the pool is created by 1,2,3,1002, then the array size is 1002?

Yes, currently, the pre-allocated array size will be 1002, but finally after the object layout scanning, it will be shrieked to 4.

wangdi1: >Object layout is per target based, not per rank based. Its count may be quite larger than rank count. So pre-allocating against object layout may be not save space. 

I thought you can either choose the total engine size or total target size to pre-allocate the array.  or even allocate by with a normally enough array, and extend later.

> On the other hand, if the ranks array (or target array) is not continuous, how can you efficiently locate related slot with index? 

Hmm, when does it need lookup, I thought it is only needed for choosing an empty slot?  Then the rank will be assigned?  Does it have to do a lot lookup?

> With this patch, the pre-allocated array size will be 4, then will be extended to 1002,

Allocate 1000 "useless" entries, just to favor the lookup in a small entry?

Nasf-Fan:
Reran the test with enable offline_extend, according to the test result, the original timeout issue has been resolved.

Ping @wangdi1 @liuxuezhao

src/object/cli_coll.c
wangdi1: @Nasf-Fan  I really think we need to discuss the issue before landing. I really concern about allocate a lot unless array. here.

liuxuezhao: just confirm here the array is 32B (struct daos_coll_target) * #total_number_of_engine_ranks is for local memory usage and will be shrinked to actual size when send to RPC?

Nasf-Fan: > > Object layout is per target based, not per rank based. Its count may be quite larger than rank count. So pre-allocating against object layout may be not save space.
> 
> I thought you can either choose the total engine size or total target size to pre-allocate the array. or even allocate by with a normally enough array, and extend later.
> 
> > On the other hand, if the ranks array (or target array) is not continuous, how can you efficiently locate related slot with index?
> 
> Hmm, when does it need lookup, I thought it is only needed for choosing an empty slot? Then the rank will be assigned? Does it have to do a lot lookup?

When you scan a object shard with some rank, how do you know whether some former scanned shard has already reside on such rank? we need to generate the bitmap per rank, so we need to know such per-rank information. If the pre-allocated array is not continuous, then we have search the given rank. 


> 
> > With this patch, the pre-allocated array size will be 4, then will be extended to 1002,
> 
> Allocate 1000 "useless" entries, just to favor the lookup in a small entry?

It is for speedup the search rank with index.

Nasf-Fan: > just confirm here the array is 32B (struct daos_coll_target) * #total_number_of_engine_ranks is for local memory usage and will be shrinked to actual size when send to RPC?

Yes, the rank array is pre-allocated before scanning the object shards, and then shrieked to the real size before sending RPC to server.

Nasf-Fan: The key topic is whether we should use the rank# as the pre-allocated array to generate the bitmap for collective operation. Let's discuss and optimize that via DAOS-15256.

This patch is to fix the correctness bug, not focus on the optimization.

wangdi1 (approved):

liuxuezhao (approved):

Nasf-Fan:
ec_multiple_rank_failure failed for DAOS-15127, the others ec_online_rebuild related failures exist on current master, not related with the patch.

Ping @daos-stack/daos-gatekeeper

gnailzenh (approved):

PR #13807, opened 2024-02-14
DAOS-15063 dfs: chown of root when daos set-owner is called on container

mchaarawi:
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/api/container.c
+++ b/src/client/api/container.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2015-2023 Intel Corporation.
+ * (C) Copyright 2015-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -495,6 +495,7 @@ daos_cont_set_owner(daos_handle_t coh, d_string_t user, d_string_t group,
 	}
 
 	rc = daos_cont_set_prop(coh, prop, ev);
+
 	daos_prop_free(prop);
 	return rc;
 }
--- a/src/client/api/container.c
+++ b/src/client/api/container.c
@@ -448,10 +448,10 @@ int
 daos_cont_set_owner(daos_handle_t coh, d_string_t user, d_string_t group,
 		    daos_event_t *ev)
 {
-	daos_prop_t	*prop;
-	uint32_t	nr = 0;
-	uint32_t	i = 0;
-	int		rc;
+	daos_prop_t *prop;
+	uint32_t     nr = 0;
+	uint32_t     i  = 0;
+	int          rc;
 
 	if (user != NULL) {
 		if (!daos_acl_principal_is_valid(user)) {
@@ -495,7 +495,6 @@ daos_cont_set_owner(daos_handle_t coh, d_string_t user, d_string_t group,
 	}
 
 	rc = daos_cont_set_prop(coh, prop, ev);
-
 	daos_prop_free(prop);
 	return rc;
 }
--- a/src/client/api/container.c
+++ b/src/client/api/container.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2015-2024 Intel Corporation.
+ * (C) Copyright 2015-2023 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -448,10 +448,10 @@ int
 daos_cont_set_owner(daos_handle_t coh, d_string_t user, d_string_t group,
 		    daos_event_t *ev)
 {
-	daos_prop_t *prop;
-	uint32_t     nr = 0;
-	uint32_t     i  = 0;
-	int          rc;
+	daos_prop_t	*prop;
+	uint32_t	nr = 0;
+	uint32_t	i = 0;
+	int		rc;
 
 	if (user != NULL) {
 		if (!daos_acl_principal_is_valid(user)) {
--- a/src/client/dfs/dfs.c
+++ b/src/client/dfs/dfs.c
@@ -7939,3 +7939,128 @@ out:
 	D_FREE(buf_keys);
 	return rc;
 }
+
+int
+dfs_root_chown(daos_handle_t coh, d_string_t user, d_string_t group)
+{
+	uid_t                      uid;
+	gid_t                      gid;
+	daos_key_t                 dkey;
+	d_sg_list_t                sgl;
+	d_iov_t                    sg_iovs[4];
+	daos_iod_t                 iod;
+	daos_recx_t                recxs[4];
+	daos_handle_t              oh;
+	int                        i;
+	struct timespec            now;
+	daos_prop_t               *prop;
+	uint32_t                   props[]   = {DAOS_PROP_CO_LAYOUT_TYPE, DAOS_PROP_CO_ROOTS};
+	const int                  num_props = ARRAY_SIZE(props);
+	struct daos_prop_entry    *entry;
+	struct daos_prop_co_roots *roots;
+	int                        rc;
+
+	prop = daos_prop_alloc(num_props);
+	if (prop == NULL)
+		return ENOMEM;
+
+	for (i = 0; i < num_props; i++)
+		prop->dpp_entries[i].dpe_type = props[i];
+	rc = daos_cont_query(coh, NULL, prop, NULL);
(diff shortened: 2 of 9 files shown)


src/client/api/container.c
```diff
@@ -448,10 +448,10 @@ int
 daos_cont_set_owner(daos_handle_t coh, d_string_t user, d_string_t group,
 		    daos_event_t *ev)
 {
-	daos_prop_t	*prop;
-	uint32_t	nr = 0;
-	uint32_t	i = 0;
-	int		rc;
+	daos_prop_t *prop;
```
ashleypittman, line 451: whitespace only changes in this file?

src/client/dfs/dfs.c
```diff
@@ -7939,3 +7939,128 @@ dfs_readdir_with_filter(dfs_t *dfs, dfs_obj_t *obj, dfs_pipeline_t *dpipe, daos_
 	D_FREE(buf_keys);
 	return rc;
 }
+
+int
+dfs_root_chown(daos_handle_t coh, d_string_t user, d_string_t group)
```
ashleypittman, line 7944: Couldn't you just dfs_mount/dfs_chmod instead here?  That would be a lot less code to verify and there's no performance consideration here.

src/control/cmd/daos/util.go
```diff
@@ -69,6 +69,15 @@ func daosError(rc C.int) error {
 	return daos.Status(rc)
 }
 
+func dfsError(rc C.int) error {
```
ashleypittman, line 72: We have plenty of go code that works with error numbers, is there something dedicated needed here and why is it tied to dfs?

src/control/cmd/daos/util.h
```diff
@@ -33,6 +33,7 @@
 #include <dfuse_ioctl.h>
 
 #include "daos_hdlr.h"
+#include "dfs_internal.h"
```
ashleypittman, line 36: Is this needed?  Perhaps the include could be added directory to the acl.go file directory?  I'm not that good at includes from go code however.

src/client/api/container.c
mchaarawi: yea i originally had some change in here and clang-format did it's thing. then i moved my changes but those remain.

src/client/dfs/dfs.c
mchaarawi: no because chmod does not handle root directory and for now i don't want it to.
also i don't want to log an error if container is not posix. so this path handles it fine.

src/control/cmd/daos/util.go
mchaarawi: yes this was only used in filesystem.go and now i use it in acl.go. so i moved it here. did not add something new.

src/control/cmd/daos/acl.go
```diff
@@ -405,6 +405,10 @@ func (cmd *containerSetOwnerCmd) Execute(args []string) error {
 			cmd.ContainerID())
 	}
 
+	if err := dfsError(C.dfs_root_chown(ap.cont, user, group)); err != nil {
+		return errors.Wrapf(err, "%s failed", fsOpString((ap.fs_op)))
+	}
```
kjacque, line 410: Should this be part of daos_cont_set_owner, so it is always called when the owner is changed, not just from the daos tool? Might also make the #include issue easier to resolve.

src/control/cmd/daos/util.go
tanabarr: fyi you don't need to move it to this file to use it, you can still use the unexported function in any of the "daos" package's files

src/control/cmd/daos/util.h
tanabarr: +1 to not accessing a private internal header from here, is there anyway to move what is needed externally to a public header

mchaarawi: i do not want to expose that as a public function for now.

src/control/cmd/daos/util.go
mchaarawi: got it, will try then. i thought i had an error when i tried it that way.

src/control/cmd/daos/acl.go
mchaarawi: i thought about this, but im not seeing a way we can call DFS from a container module.
DFS relies on container module but not the other way around.

mchaarawi: what we can add is a new function in dfs that calls daos_cont_set_owner and we call it dfs_cont_set_owner. let me explore that

src/control/cmd/daos/util.h
mchaarawi: > You should move this include to the top of acl.go, where util.h is already included. You can add it as another #include line there. However it seems odd to me to include a private header from another directory. If the daos tool is using the dfs function, it seems like it should be public.

just FYI, adding that at top of acl.go, the build error is:
scons: *** [build/debug/gcc/src/control/daos] Implicit dependency `src/control/cmd/daos/dfs_internal.h' not found, needed by target `build/debug/gcc/src/control/daos'.

ashleypittman:
Looking at the existing usage of dfs_internal.h more closely it seems to have a combination of three uses

- Function prototypes and structs for sharing code between dfs.c and dfs_internal.c
- Function prototypes for functions that dfuse uses
- Function prototypes for functions that go command use.

I think there would be a good case for splitting this file into two, keeping the existing dfs_internal.h purely for dfs internals and only including it from dfs.c, dfs_internal.c and test code.  Then having a new file containing a "private" dfs api for code which is external to dfs but internal to daos where we don't want to commit to having the interfaces in the public API.  I assume this new file would then go in src/include/daos where it could be included from other areas of the source without having to modify CPPPATH.

mchaarawi:
> Looking at the existing usage of dfs_internal.h more closely it seems to have a combination of three uses
> 
> * Function prototypes and structs for sharing code between dfs.c and dfs_internal.c
> * Function prototypes for functions that dfuse uses
> * Function prototypes for functions that go command use.
> 
> I think there would be a good case for splitting this file into two, keeping the existing dfs_internal.h purely for dfs internals and only including it from dfs.c, dfs_internal.c and test code. Then having a new file containing a "private" dfs api for code which is external to dfs but internal to daos where we don't want to commit to having the interfaces in the public API. I assume this new file would then go in src/include/daos where it could be included from other areas of the source without having to modify CPPPATH.

yea that was on my plate to split dfs_internal as dfs.c is becoming too large and we need something just internal to dfs.
for this particular case, i will just make a new public function dfs_cont_set_owner. will refresh pr soon

mchaarawi:
apologies for the force push.. i mistakenly did a master rebase and not a merge.
the changes themselves to the PR were not forced and in a separate commit.

src/client/dfs/dfs.c
```diff
@@ -7939,3 +7939,136 @@ dfs_readdir_with_filter(dfs_t *dfs, dfs_obj_t *obj, dfs_pipeline_t *dpipe, daos_
...
+	}
+
+	entry = daos_prop_entry_get(prop, DAOS_PROP_CO_LAYOUT_TYPE);
+	if (entry == NULL || entry->dpe_val != DAOS_PROP_CO_LAYOUT_POSIX) {
+		rc = EINVAL;
+		D_ERROR("container is not of type POSIX: %d (%s)\n", rc, strerror(rc));
+		D_GOTO(out_prop, rc);
+	}
+
+	rc = daos_cont_set_owner(coh, user, group, NULL);
```
daltonbohning, line 7982: daos_cont_set_owner requires user or group. Should dfs_cont_set_owner check and report an error above? Or just rely on this call failing?

daltonbohning (approved):

src/client/dfs/dfs.c
mchaarawi: it would be repetitive to add the checking here. daos_cont_set_owner() should properly log an error in this case.

kjacque (approved):

src/control/cmd/daos/acl.go
kjacque: I like that solution. Looks good!

ashleypittman (approved):

mchaarawi:
failing a rebuild test that did not report cmocka results. unrelated to PR.

PR #13810, opened 2024-02-15
DAOS-14978 dfuse: Add a read-only mount option.

ashleypittman:
Add a command line option for read-only mounting of dfuse.
When set instruct the kernel to mount read-only
Construct the dfuse_ops struct dynamically and do not set
callback that modify data when in read-onlyu mode.
Make several functions static and structs const.
Add ro and multi_user to the options that can be set via -o

Required-githooks: true


src/client/dfuse/dfuse.h
```diff
@@ -24,6 +24,10 @@
 
 #include "dfuse_common.h"
 
+/* TODO: This should update the container open flags once the update has landed
+ * https://github.com/daos-stack/daos/pull/13702
+ */
+
```
daltonbohning, line 30: So we we're wanting to land #13702 first, then rebase this PR?

ashleypittman: That's my hope.  I could make the change in this PR but it'll conflict.

ashleypittman: Ok, updated this PR, it'll just have to conflict.

src/client/dfuse/dfuse_core.c
```diff
@@ -691,10 +691,16 @@ dfuse_cont_open_by_label(struct dfuse_info *dfuse_info, struct dfuse_pool *dfp,
...
+	if (dfuse_info->di_read_only) {
 		dfs_flags = O_RDONLY;
 		rc = daos_cont_open(dfp->dfp_poh, label, DAOS_COO_RO, &dfc->dfs_coh, &c_info, NULL);
+	} else {
+		rc = daos_cont_open(dfp->dfp_poh, label, DAOS_COO_RW, &dfc->dfs_coh, &c_info, NULL);
+		if (rc == -DER_NO_PERM) {
+			dfs_flags = O_RDONLY;
+			rc        = daos_cont_open(dfp->dfp_poh, label, DAOS_COO_RO, &dfc->dfs_coh,
+						   &c_info, NULL);
+		}
```
mchaarawi, line 703: i feel like it's better to fail in this case rather than retry with RO if user did not request RO.

ashleypittman: This is existing code but for dfuse then in many cases the user doesn't explicitly request either mode, they just see a path and try to access it.  Failing back to RO was something that came up early on in the multi-user testing as something that is likely to happen, I'll see if I can find the ticket.

Even this PR doesn't allow setting of read-only mode on a per container basis, it's more saying "never write" rather than read-only, in general dfuse just doesn't know if access will be RO or RW so it needs to be flexible.

mchaarawi (approved):

wiliamhuang (approved):

daltonbohning (approved):

PR #13818, opened 2024-02-15
DAOS-12710 build: Update include order of header files and add missing dependencies.

ashleypittman:
Sanitize the include order and handling for header files.
Include local (same directory) headers first, then non-local but
internal, then system, then library and then daos public headers.

Add python code for checking the order.

This commit only changes headers which need manual changes due
to missing includes.  Headers which can be programatically updated
will be changed later.


src/container/srv_container.c
```diff
@@ -14,17 +14,18 @@
 
 #define D_LOGFAC DD_FAC(container)
 
-#include <daos_srv/container.h>
```
liw, line 17: [Request changes] This is the module API header of `ds_cont`, and was included first so that if this header forgot to include some required headers, the compilation of `ds_cont` would fail.

src/pool/srv_pool.c
```diff
@@ -14,29 +14,31 @@
 
 #define D_LOGFAC DD_FAC(pool)
 
-#include <daos_srv/pool.h>
```
liw, line 17: Ditto.

liw (changes requested):
@ashleypittman, would it make sense to first educate the team of your new rules, and hopeful to get some consensus, before enforcing it, I wonder? What do you think?

src/container/srv_container.c
ashleypittman: We spoke about this in some detail, I had misunderstood some of the description inhttps://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes which I was trying to follow.  I've taken the PR back to draft and will work through the issues again.

PR #13823, opened 2024-02-16
DAOS-15237 vos: implement temporary cache

jolivier23:
We should not pollute the VOS object cache when
rebuilding on a new target.  Add an ephemeral flag to the VOS object cache.

We will keep one ephemeral entry at a time. If
the same oid is read, the entry will be added to
the cache.

Required-githooks: true

Change-Id: Ic76e4ca56f5a4d1d8b83df536e6dc644d7bc396b

### Gatekeeper:


src/vos/vos_obj_cache.c
```diff
@@ -320,12 +435,16 @@ vos_obj_hold(struct daos_lru_cache *occ, struct vos_container *cont,
 	if (cont->vc_pool->vp_dying)
```
NiuYawei, line 435: To be honest, it's bit hard for me to follow this vos_obj_hold() function already :) , but seems we have to complicate it further (in this PR and in md-on-ssd phase 2), so I'm wondering if we'd split this gigantic vos_obj_hold() for the sake of code maintainance?

It looks to me there are three major functionalities implemented in current vos_obj_hold():
1. Object cache lookup/create;
2. OI lookup/create;
3. ilog check;

Is it possible to make it into three individual functions respectively?

I think the pure object cache hold function (without those 'intent' and ilog complexities involved) would be very helpful for future extension. (In md-on-ssd phase 2, we will need a function to hold object cache & pin the necessary memory buckets, and it could be called in vos_update_begin() no matter if the OI exists or not.)

Does this proposal sound good to you? Anything I overlooked?

jolivier23: I don't disagree with it being difficult to follow

Are you suggesting callers would have 3 functions to call?  Not necessarily opposed to it but I don't think this particular patch is the place to do it.

I might need you to elaborate on the phase 2 plan a bit.   Why does memory need to be pinned in vos_update_begin?  Maybe we can chat about it in the MD on SSD call.

NiuYawei: In phase 2, we'll have to know which bucket the object is using (or will use when the object doesn't exist) in vos_update_begin(), and the bucket has to be pinned in memory until vos_update_end(), few allocations (reserve) could happen between vos_update_begin() and vos_update_end():
1. small value located on meta blob;
2. single value vos_irec_df & csum;
 
To ensure concurrent updates use same bucket for reserve, I think the best place to track bucket info would be object cache, and since bucket pin/load operation could yield, I'm afraid that we might need to make vos_obj_hold() into at least two parts in the future:
1. Pure object cache part, which includes bucket pin and being called outside of tx.
2. Other part which might need be called within tx. 

Yeah, let's discuss it in the meeting.

src/vos/vos_obj_cache.c
```diff
@@ -38,6 +38,26 @@ struct obj_lru_key {
...
+	/** Temporary scratch space for holding an object before caching it. */
+	struct vos_object      oc_scratch;
+	/** For objects that are not cached with an expected short life such as
+	 *  for rebuild or aggregation where we are working on one at at a time
+	 *  and post many operations for the same object without a yield. Avoid
+	 *  polluting the object cache for user I/O.
+	 */
+	struct vos_object     *oc_ephemeral;
+	/** container for ephemeral object (for same oid case) */
+	struct vos_container  *oc_eph_cont;
```
NiuYawei, line 53: Why do we need this? Can't get get it through oc_ephemeral->obj_cont?

src/vos/vos_obj_cache.c
```diff
@@ -238,69 +287,134 @@ vos_obj_discard_hold(struct daos_lru_cache *occ, struct vos_container *cont, dao
...
-	clean_object(&obj_local);
-	memset(&obj_local, 0, sizeof(obj_local));
+	if (ephemeral) {
+		if (old_eph) {
+			clean_object(old_eph);
+			if (occ->oc_eph_in_use)
+				old_eph->obj_dropped = 1;
+			else
+				D_FREE(old_eph);
+		}
```
NiuYawei, line 398: This looks duplicated with the latter part (start from switch(action)) in move_object(), it would be more read-able if you move that part out of move_object().

NiuYawei (approved):

PR #13828, opened 2024-02-17
DAOS-15201 cart: Multi-interface support for clients.

frostedcmos:
- Interface can now be specified as either a coma-separated list or as an empty string, e.g. "eth0,eth1" to both D_INTERFACE environment variable and cio_interface init field.
- If no interface is provided, mercury will pick one automatically.

- IP address is no longer required on the interface to initialize CaRT on it. Instead mercury is now passed the interface name directly.

- some of old context functions removed
- New client API: crt_num_ifaces_get() - returns number of interfaces CaRT was initialized with.
- New client API: crt_context_create_on_iface_idx() - Allow creation of context on the specified interface.
- 'int provider' replaced with proper 'crt_provider_t' enum

Required-githooks: true

### Gatekeeper:


src/cart/crt_init.c
```diff
@@ -1161,14 +1118,37 @@ crt_na_config_init(bool primary, crt_provider_t provider,
...
+	/* CXI provider requires domain to be set */
+	if (provider == CRT_PROV_OFI_CXI && !domain) {
+		D_ERROR("Domain must be set for CXI provider\n");
+		D_GOTO(out, rc = -DER_INVAL);
+	}
+
+	if (interface) {
+		if (provider == CRT_PROV_OFI_CXI) {
+			D_WARN("Interface '%s' ignored for CXI. Using domain '%s' instead\n",
+			       interface, domain);
```
soumagne, line 1140: I don't necessarily think we should emit a warning for that.

frostedcmos: I agree it probably shouldn't be a warning, but I still want it to be mentioned in logs in some way. changed to D_INFO for now.

src/cart/README.env
```diff
@@ -112,25 +112,33 @@ This file lists the environment variables used in CaRT.
...
+   allowing allocation of CaRT contexts on them using crt_context_create_on_iface_idx() API.
+   If no interface is specified, the network layer will select one automatically
+   during context creation (crt_context_create() API)
 
  . D_DOMAIN (Deprecated: OFI_DOMAIN)
-   Set it as the network domain name to be used for OFI communications. Domain
-   name can be determined by running 'fi_info' app. If not set, OFI_DOMAIN is
-   attempted to be automatically determined.
+   Set it as the network domain name to be used for network communications. Domain
+   name can be determined by running 'fi_info' app. If not set, D_DOMAIN is
```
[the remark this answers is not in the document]
frostedcmos: Changed

soumagne (dismissed):

jgmoore-or (dismissed):

src/include/cart/api.h
```diff
@@ -104,6 +104,36 @@ crt_init(crt_group_id_t grpid, uint32_t flags)
...
+ *
+ * Note: This is a client-side only API.
+ *
+ * \param[in]  iface_index     index of the interface.
+ * \param[out] crt_ctx         created CRT transport context
+ *
+ * \return                     DER_SUCCESS on success, negative value if error
+ */
+int
+crt_context_create_on_iface_idx(uint32_t iface_index, crt_context_t *crt_ctx);
```
mchaarawi, line 135: how can i get information on an interface given an index?
the way i would use this is that i would init cart with all possible interfaces, then i want to create a context with an interface where this context will be running on (the numa binding).

src/cart/README.env
```diff
@@ -112,25 +112,33 @@ This file lists the environment variables used in CaRT.
...
 
  . D_INTERFACE (Deprecated: OFI_INTERFACE)
-   Set it as the network device name to be used for OFI communication, for
+   Set it as the network device name to be used for network communication, for
    example "eth0", "ib0" or "ens33" etc.
+   Supports a comma separated list of interfaces, for example "eth0,ib0,ens33".
+   If multiple interfaces are specified then CaRT is initialized on all of them,
+   allowing allocation of CaRT contexts on them using crt_context_create_on_iface_idx() API.
+   If no interface is specified, the network layer will select one automatically
+   during context creation (crt_context_create() API)
```
mchaarawi, line 129: since we do not use D_INTERFACE in DAOS client, where can this be provided in cart init?

mchaarawi (commented):
should add some tests when landing new APIs/features.

src/include/cart/api.h
frostedcmos: The idea here is that as an app you wouldn't want to hardcode any interface names, so instead when you initialize CaRT with a list of interfaces you can then refer to them in a generic manner by the index in the order in which they were listed. 

The way I imagine daos layer would use new apis is something like:
`num_ifaces = crt_num_ifaces_get();
for (i = 0; i < num_ifaces; i++) {
    crt_context_create_on_iface_idx(i, &ctx);
    // Create event queue with this context
}
`

If it helps I can also add lookup functions to return an interface name by the index and vice versa as we already have this information stored internally. Beyond the name of the interface (as passed at init) we don't have any other information about it today in CaRT.

frostedcmos:
> should add some tests when landing new APIs/features.

will do with the next push

src/cart/README.env
frostedcmos: most D_* params such as D_INTERFACE, D_PROVIDER, D_PORT etc.. we have a corresponding crt_init_options_t setting that daos is using today when initing cart with crt_init_opt().

In this case for the interface it would be crt_init_options_t::cio_interface field, just like it is today.

src/include/cart/api.h
mchaarawi: no i was not talking about hardcoding of course.
in the end, we need to make some correlation between the iface index and the CPU of the core to select the proper one. i guess we can also keep that mapping cached on daos_init when talking to the agent first.
anyway might be a good idea to discuss this in more detail with Kris and sketch a use case out to just verify.

frostedcmos: Since daos_agent is already going to provide a client with the interface to be used (I assume it is based on the core on which app was running?), perhaps we can have an iface_name2indx API to do the conversion, and then subsequently use crt_context_create_on_iface_idx() to allocate the required one.  Do you think this will work?

mchaarawi: the gist of the problem today is that the agent provided one thing during init, but with multi threaded apps, the thread might be created on a different CPU/numa node, and so it needs a different one than what we selected on init.
So my thought of what we need is that on daos_init() we tell cart to use all interfaces (i guess the agent can provide those?), but it would be nice to also provide the ability to specify a default one to maintain original behavior like today with context_create without an iface idx.

Then on every context creation (with this new API), we need to specify the index of the iface/nic of the calling thread cpu affinity. this last part sounds like we need to ping the agent again for that info (agent would provide a string of the iface name I assume ?). but then we need to do the translation to that index.
so I see two potential issues:
1) pinging the agent on every context create - might be expensive.
2) the translation from the iface from the agent to the crt idx.
so for 2 if the iface_name2idx() is consistent / deterministic, then it should be good for that one.

frostedcmos: Perhaps then we can provide APIs such as crt_interface_select(char *) / crt_interface_select_by_idx(int) which would modify the default interface on which crt_context_create() will allocate the context?

name2idx is already just strcmps but with this then can be done only once if app needs to create multiple contexts on the same interface.

for 1. -- is the mapping of core to nic a fixed thing? I was assuming daos_agent does some sort of balancing as well to 'spread the load' across the nics.

If the mapping is fixed then perhaps we can query it from the agent once and do the translation ourselves (whether in daos or cart layer).
@kjacque Any thoughts on the agent interaction here?

mchaarawi: > for 1. -- is the mapping of core to nic a fixed thing? I was assuming daos_agent does some sort of balancing as well to 'spread the load' across the nics.

not always. on aurora for example there are 4 nics per CPU on the client side and the agent round robins in that case.

src/cart/crt_context.c
```diff
@@ -370,19 +350,90 @@ crt_context_provider_create(crt_context_t *crt_ctx, crt_provider_t provider, boo
...
+
+	num_ifaces = crt_provider_num_ifaces_get(true, crt_gdata.cg_primary_prov);
+
+	for (i = 0; i < num_ifaces; i++) {
+		name = crt_provider_iface_str_get(true, crt_gdata.cg_primary_prov, i);
+
+		if (strcmp(name, iface_name) == 0) {
+			*idx = i;
+			rc = DER_SUCCESS;
+			break;
```
soumagne, line 411: you could directly return without the need for `rc`.

src/tests/ftest/cart/no_pmix_multi_ctx.c
```diff
@@ -154,13 +148,78 @@ int main(int argc, char **argv)
...
+		env = getenv("OFI_INTERFACE"); /* Support deprecated envar for now */
+	D_ASSERTF(env != NULL, "either D_INTERFACE or OFI_INTERFACE must be set");
+
+	/* Save current interface value */
+	D_ASPRINTF(cur_iface_str, "%s", env);
+	D_ASSERTF(cur_iface_str != NULL, "Failed to allocate string");
+
+	/* Append loopback to the current interface list */
+	D_ASPRINTF(new_iface_str, "%s,lo", cur_iface_str);
+	D_ASSERTF(new_iface_str != NULL, "Failed to allocate string");
```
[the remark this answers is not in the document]
frostedcmos: currently its sm and tcp:
https://github.com/daos-stack/daos/blob/master/src/tests/ftest/cart/no_pmix/multictx_one_node.yaml

src/cart/crt_context.c
frostedcmos: sure, will simplify if i do another push

jgmoore-or (dismissed):

src/tests/ftest/cart/no_pmix_multi_ctx.c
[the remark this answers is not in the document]
soumagne: does it work for sm ? I guess it's just picking some random ID in that case since it ignores the interface that you pass :)

src/cart/crt_context.c
soumagne: was that pushed ? I did not see it

src/tests/ftest/cart/no_pmix_multi_ctx.c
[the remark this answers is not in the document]
frostedcmos: Yeah, after the latest change (to initialize provider fully for sm) the test works with sm, and yeah the interface just gets ignored with it.

src/cart/crt_context.c
frostedcmos: no, i didnt push that one:) i will if there are more changes to make, otherwise would like to avoid another ci run

soumagne (dismissed):

daltonbohning (commented):
High-level ftest code changes LGTM

src/include/cart/api.h
```diff
@@ -104,6 +104,68 @@ crt_init(crt_group_id_t grpid, uint32_t flags)
...
+ * can then be used to create a context on a specific interface using
+ * crt_context_create_on_iface_idx() API.
+ *
+ * \param[in]  iface_name       Interface name to look up.
+ * \param[out] idx              Returned index
+ *
+ * \return                      DER_SUCCESS on success, negative value if error
+ */
+int
+crt_iface_name2idx(char *iface_name, int *idx);
```
mchaarawi, line 148: should be const char *iface_name

src/include/cart/api.h
```diff
@@ -104,6 +104,68 @@ crt_init(crt_group_id_t grpid, uint32_t flags)
...
+ *
+ * Note: This is a client-side only API.
+ *
+ * \param[in]  iface_name      name of the interface.
+ * \param[out] crt_ctx         created CRT transport context
+ *
+ * \return                     DER_SUCCESS on success, negative value if error
+ */
+int
+crt_context_create_on_iface(char *iface_name, crt_context_t *crt_ctx);
```
mchaarawi, line 166: ditto

mchaarawi (approved):

src/include/cart/api.h
frostedcmos: done

src/include/cart/api.h
frostedcmos: done

src/cart/crt_context.c
frostedcmos: done

mchaarawi (dismissed):

soumagne (dismissed):

soumagne:
I think at some point we should also make sure we address clang format in PRs as much as possible (i.e., only for the subset of changes without having two styles in the same file like in the headers) otherwise we'll never fix it :)

mchaarawi:
> I think at some point we should also gradually make sure we address clang format in PRs as much as possible (i.e., only for the subset of changes without having two styles in the same file like in the headers) otherwise we'll never fix it :)

On el9, all i had to do was install git-clang-format and the githook will automatically format the PR.

soumagne:
ok, well I don't know if we necessarily want to do that because like I said, fixing all formatting for the current set of changes will make things half ugly in the headers but applying some of it would be a first good step ? just a thought.

jgmoore-or (dismissed):

src/cart/crt_init.c
```diff
@@ -747,23 +747,35 @@ crt_init_opt(crt_group_id_t grpid, uint32_t flags, crt_init_options_t *opt)
 			D_GOTO(unlock, rc = -DER_NONEXIST);
 		}
 
-		rc = __split_arg(interface, &iface0, &iface1);
+		/*
+		 * TODO: revisit this.
+		 * interfaces for multi-provider mode for now are delimited with ':'
+		 * to allow multi-interface selection (delimited by ',') in a single provider mode.
```
kjacque, line 753: As I was reading the code, I was wondering about this. This will probably have some ripple effects in the multiprovider branch.

src/cart/README.env
kjacque: I wonder, should we be using D_INTERFACE/etc. in the client? I see we are still using the old OFI_INTERFACE/etc. there. Seems like either the client should be setting both or Cart detecting both.

frostedcmos: @kjacque when updating the code we should be switching to D_* variant of envariables and away from old OFI_* ones. Currently cart still supports both, with D_* variants (e.g. D_INTERFACE) taking a precedence if set. OFI_* versions have been marked as deprecated for a while now.

src/cart/crt_init.c
frostedcmos: Yeah, we need to chat on what to do here. 

One option is to keep it coma separated for both cases, and determine whether it's a multi-interface or a multi-provider case based on whether its a client and server side init:

- multi-interface setup is only available on clients
- multi-provider setup is only available on servers

This will keep it contained within cart mostly, but can make it confusing to the end user.

src/cart/README.env
kjacque: I created a ticket for this. It's a bit of a tangent from this changeset, but IMO we should begin this transition in libdaos. https://daosio.atlassian.net/browse/DAOS-15401

src/cart/crt_init.c
kjacque: Yeah, I feel like that could be a bit confusing. One thing that occurred to me, maybe as we productize multiprovider, we can hide the change from the user. Right now the config file reflects Cart's format for these variables, but it doesn't have to.

kjacque (approved):

soumagne (approved):

PR #13833, opened 2024-02-19
DAOS-15240 cq: Add a GitHub actions unit test stage.

ashleypittman:

### Gatekeeper:


src/control/run_go_tests.sh
```diff
@@ -117,50 +117,6 @@ function setup_environment()
 	fi
 }
 
-function emit_junit_failure()
```
mjmac, line 120: What's the motivation for removing this function? If it's not currently needed, then we could adjust the logic so that it's not called for the current CI environment. I'm hesitant to just toss it altogether though, as this precise format was found to work with Jenkins' JUnit parser via trial and error.

src/control/run_go_tests.sh
```diff
@@ -117,50 +117,6 @@ function setup_environment()
...
-<?xml version="1.0" encoding="UTF-8" ?>
-<testsuites>
-  <testsuite tests="1" failures="1" errors="0" skipped="0" >
-    ${teststr}
-  </testsuite>
-</testsuites>
-EOF
-}
-
-function check_formatting()
```
mjmac, line 143: I think this function is still useful for developers who run the script locally (anyone who touches Go code is encouraged to run it prior to pushing a PR). ... After reading the rest of this PR, I see that you extracted it to a standalone script. I guess maybe we could also add that script to the pre-commit git hooks?

ashleypittman: The logic with this file was to move the check to where it ran outside of the Intel CI infrastructure so that the results were available to all.  This just seemed an easy thing to include and it doesn't natively support JUnit - which works with Jenkins but not with GitHub actions unfortunately.

I'm torn between wanting to run things quicker with the results available to all in GitHub actions or with the better reporting but taking longer and visible to fewer people.

It's easy enough to back out this change, I just saw it as something easy to move but hadn't considered your point about having a single entry point for developers - that's nice that you've got that process that works.

mjmac: I had completely forgotten about the gofmt githook that I added in December of 2022. I have no problem with removing this from the test runner script and into its own GHA-compatible script. At some point, it would be nice to make the GHA gofmt check work like the other linters, so that it shows the required updates inline as comments, but that could be done later.

ashleypittman: To make annotations work then you need the source line, gofmt gives you a diff but it's not easy to go from there to the line.

You've just reminded me of something though, in Jenkins checkpatch we used to have a process where we had scripts that could "fix" the source tree by patching them as gofmt has an option to do, then our code would run "git diff" to see where the changes were and if they were in a line modified by the PR then they'd annotate them as comments to the PR.  This was used by codespell and a gofmt hook, but was also really flaky and used to break a lot so I removed it when I moved codespell to GitHub actions.

utils/utest.yaml
```diff
@@ -2,8 +2,10 @@
 #
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 - name: include checks
+  memcheck: False
+  gha: True
   tests:
-    - cmd: ["utils/include_test.py"]
+    - cmd: ["utils/unit/include_test.py"]
```
jolivier23, line 8: I presume the idea is to eventually run more of them?

jolivier23: Maybe this could be more generic...such as 
tags: gha

and using --tags gha

jolivier23 (approved):
Ideally, we would move all unit testing to gha

utils/utest.yaml
ashleypittman: @daltonbohning could comment on this.  There already is a `--suite` option but it does string matching rather than using tags so it's not much use IMO.

ashleypittman:
> Ideally, we would move all unit testing to gha

I agree, that's the whole point of making a start to see how far we can go with that.

Jenkins has better reporting of junit results and can handle valgrind xml files which seems to be entirely missing from GitHub actions so in some ways it would be a step backwards.

utils/utest.yaml
daltonbohning: I recently added this one, which might belong here ( I need to think about it some more)
https://github.com/daos-stack/daos/blob/40f9ffaa9dbc97f68be6450cdf9ec58c371c6c1e/utils/utest.yaml#L185-L189
And I suspect there are more we could group together.

And at least for python tests, we might consider running `pytest` against a directory of unit tests.

daltonbohning:
> > Ideally, we would move all unit testing to gha
> 
> I agree, that's the whole point of making a start to see how far we can go with that.
> 
> Jenkins has better reporting of junit results and can handle valgrind xml files which seems to be entirely missing from GitHub actions so in some ways it would be a step backwards.

Do you know if valgrind can report in a format other than XML? XML has been losing popularity and generally has less support for parsing and displaying than, for example, json.

.github/workflows/bash_unit_testing.yml
```diff
@@ -1,7 +1,6 @@
 name: bash Unit Testing
 
 on:
-  push:
```
daltonbohning, line 4: Does removing this mean the action only runs once when the pull request is created?

daltonbohning, line 4: Does removing this mean the action only runs once when the pull request is created?

.github/workflows/bash_unit_testing.yml
```diff
@@ -15,6 +14,7 @@ defaults:
 jobs:
   Test-gha-functions:
     name: Tests in ci/gha_functions.sh
+    if: github.repository == 'daos-stack/daos'
```
daltonbohning, line 17: Is this related to pushing PRs from forks? Why would the repo not be `daos-stack/daos`?

.github/workflows/bash_unit_testing.yml
ashleypittman: Yes.  As it stands this runs whenever there's a commit, so for example when you push a branch without creating a PR or more commonly when a PR is merged then it runs against master, neither of which is helpful or done by any other of our checks.

.github/workflows/bash_unit_testing.yml
ashleypittman: Kind of, it's related to pushing PRs to forks.  Doing anything in a fork I get lots of email about unable to find any self-hosted runners, with the bash code running on push then even syncing a fork gets me an email with errors in it.

.github/workflows/bash_unit_testing.yml
daltonbohning: What about when a new commit is pushed to this same branch/PR?

.github/workflows/bash_unit_testing.yml
daltonbohning: That sounds like we still fundamentally don't support PRs from forks then?

.github/workflows/bash_unit_testing.yml
ashleypittman: The line below catches that and is all any of our other actions use./

.github/workflows/bash_unit_testing.yml
ashleypittman: Not at all.  If there's a PR from a form to daos-stack then this would still trigger.  This change is for if there's a PR to a fork where the action would currently fail due to lack of self-hosted runners.

utils/utest.yaml
daltonbohning: Sorry, I didn't mean you had to group those here, but that works! :)

ashleypittman: I did it mostly to see what the output looks like.  actions only appears to have one level of grouping.

daltonbohning (dismissed):

src/control/run_go_tests.sh
ashleypittman: I deleted this part of the change.  The intention wasn't ever to change the workflow, I just wanted to be able to get the gofmt results viewable outside of Intel CI systems.

mjmac (approved):

daltonbohning (approved):

PR #13834, opened 2024-02-20
DAOS-15154 pool: ds_pool_child_lookup() could return NULL

wangshilong:
When a faulty device is replaced with a new one using the “dmg storage replace nvme” command, 
the reintegration of affected pool targets is automatically triggered. The followings steps outline the
device replacement process on the engine side:

1) Replace the old device with the new device in the SMD. 
2) Setup all SPDK related stuff for the new device. 
3) Start ds_pool_child

It is important to note that manual reintegration may be initiated before step 3, which trigger assertion.
For above case, return DER_AGAIN error, this patch also tries to fix another two places which forgot to
check return of ds_pool_child_lookup().

Required-githooks: true

### DIFF
--- a/src/object/srv_obj_migrate.c
+++ b/src/object/srv_obj_migrate.c
@@ -405,6 +405,8 @@ struct migrate_pool_tls_create_arg {
 	unsigned int generation;
 	uint32_t opc;
 	uint32_t new_layout_ver;
+	/* for sys sstream */
+	bool	 track_status;
 };
 
 int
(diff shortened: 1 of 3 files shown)


tanabarr (dismissed):

src/object/srv_obj_migrate.c
```diff
@@ -452,7 +458,7 @@ migrate_pool_tls_create_one(void *data)
 	pool_tls->mpt_executed_ult = 0;
 	pool_tls->mpt_root_hdl = DAOS_HDL_INVAL;
 	pool_tls->mpt_max_eph = arg->max_eph;
-	pool_tls->mpt_pool = ds_pool_child_lookup(arg->pool_uuid);
+	pool_tls->mpt_pool = child;
```
NiuYawei, line 461: ds_pool_child could leak when any of above operation fails. I think simpler change could be:

`
pool_tls->mpt_pool = ds_pool_child_lookup(arg->pool_uuid);
 if (pool_tls->mpt_pool == NULL)
    D_GOTO(out, rc = -DER_NO_HDL);
`

NiuYawei (changes requested):

NiuYawei (dismissed):

NiuYawei (approved):

jolivier23 (approved):

PR #13851, opened 2024-02-22
DAOS-15254 pydaos: Add a open_mode option for containers.

ashleypittman:
Allow access to read-only containers by adding a open_mode option
to the container object.  Default is 'RW' so the existing behaviour is unchanged.

Pools do not require write access so open these read-only always.


ziegenbalg (approved):
After some thought, I do like the open_ro flag better than automatically dropping to RO if users don't have permission.

johannlombardi (approved):

src/client/pydaos/pydaos_core.py
```diff
@@ -69,7 +69,7 @@ class DCont():
         Create new DArray object.
     """
 
-    def __init__(self, pool=None, cont=None, path=None):
+    def __init__(self, pool=None, cont=None, path=None, open_ro=False):
```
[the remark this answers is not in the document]
daltonbohning: Instead of `open_ro=True` or `open_ro=False`, should we just take a `flags` argument? Or, for python convenience, we could even accept a string of `RW` or `RO`.
`open_ro` seems oddly specific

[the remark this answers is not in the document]
ashleypittman: It is oddly specific but I was trying to maintain 100% compatibility with existing code where no mode is provided.  Yes, in hindsight the interface should have required a `RW` or `RO` option from the start.

Perhaps this would be better?  Then we could support `RW` or `RO`.
```suggestion
    def __init__(self, pool=None, cont=None, path=None, open_mode='RW`):
```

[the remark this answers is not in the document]
daltonbohning: I think that's good. And defaulting to `RW` is backwards compatible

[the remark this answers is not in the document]
ashleypittman: The current PR is backwards compatible, that's why I wrote it how I did.  I can update tomorrow.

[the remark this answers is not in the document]
daltonbohning: The current is backward compatible, but if we ever need to open with a different flag in the future, we'll need to add another argument. Whereas a generic `flags` argument makes it easier for future changes. E.g. if someone wanted to use `DAOS_COO_EX`.

Also, defaulting `flags=RW` makes it clear that the default is `RW`. But `open_ro=False` only tells you that it's NOT `RO`.

ashleypittman:
@johannlombardi @daltonbohning can you re-review this please with a view to landing it.

src/client/pydaos/pydaos_shim.c
```diff
@@ -153,7 +132,7 @@ __shim_handle__err_to_str(PyObject *self, PyObject *args)
  */
 
 static PyObject *
-cont_open(int ret, char *pool, char *cont, int flags)
+cont_open(int ret, char *pool, char *cont, int ro)
```
daltonbohning, line 135: Can we not just pass the flag here? Changing `flags` to `ro` seems like a reduction in support

src/client/pydaos/pydaos_core.py
```diff
@@ -69,20 +75,25 @@ class DCont():
...
+        open_ro = False
+        if open_mode == 'RO':
+            open_ro = True
+        elif open_mode != 'RW':
+            raise InvalidModeError('open_mode is neither "RO" or "RW"')
         if path is not None:
             self.pool = None
             self.cont = None
-            (ret, hdl) = pydaos_shim.cont_open_by_path(DAOS_MAGIC, path, 0)
+            (ret, hdl) = pydaos_shim.cont_open_by_path(DAOS_MAGIC, path, open_ro)
```
daltonbohning, line 92: I think it's better if we can pass RW or RO by flags

src/client/pydaos/pydaos_core.py
```diff
@@ -69,20 +75,25 @@ class DCont():
...
         self._dc = DaosClient()
         self._hdl = None
         if path is None and (pool is None or cont is None):
             raise PyDError("invalid pool or container UUID",
                            -pydaos_shim.DER_INVAL)
+        open_ro = False
+        if open_mode == 'RO':
+            open_ro = True
+        elif open_mode != 'RW':
+            raise InvalidModeError('open_mode is neither "RO" or "RW"')
```
daltonbohning, line 88: I don't know what's best here, but we could use the built-in `ValueError` which is most similar to EINVAL

src/client/pydaos/pydaos_shim.c
ashleypittman: We could, but this is the c code which is not part of the public API.  Previously flags existed but was never used and as the public interface to this is via python we should just pass in the exact options we want and do the logic checking on the python side.

src/client/pydaos/pydaos_core.py
ashleypittman: If you see on L23 this inherits from `ValueError` so could be caught as ValueError but also allows more specific error messages and catching.

src/client/pydaos/pydaos_core.py
daltonbohning: Couldn't it just be similar to this?
https://github.com/daos-stack/daos/blob/80f92739b294d97004fe5db141139030a7dc803c/src/client/pydaos/pydaos_shim.c#L1484

Maybe even just
```
PyModule_AddIntConstant(module, "DAOS_COO_RO", DAOS_COO_RO);
PyModule_AddIntConstant(module, "DAOS_COO_RW", DAOS_COO_RW);
PyModule_AddIntConstant(module, "DAOS_COO_EX", DAOS_COO_EX);
...
```

ashleypittman: Are you proposing a change to the (external) python API here or the (internal) C API?

daltonbohning: It would be a change in `pydaos_shim.c` to export definitions based on the core C code.
E.g. it's used like this
https://github.com/daos-stack/daos/blob/80f92739b294d97004fe5db141139030a7dc803c/src/client/pydaos/pydaos_core.py#L86


So in `pydaos_shim.c` we'd have
```
PyModule_AddIntConstant(module, "DAOS_COO_RO", DAOS_COO_RO);
```

and in the python we'd do
```
pydaos_shim.DAOS_COO_RO
```

ashleypittman: Going forward that would be better I agree but whilst there's only one value I don't see the benefit in making that change yet.  This is all internal so there's no compatibility issue that requires us to commit to something now.

daltonbohning (approved):

johannlombardi (approved):

PR #13852, opened 2024-02-22
DAOS-15863 container: fix container destroy error

wangshilong:
After stopping the container service, certain user cases
may still retain references to the container's child objects,
such as ongoing IO operations. To prevent errors during container
destruction, ensure that the container is only destroyed after
all dependent operations have completed.

Fix daos_lru_cache_create() to return an error if D_HASH_FT_EPHEMERAL
is passed in.

Required-githooks: true


src/container/srv_target.c
```diff
@@ -1192,6 +1199,11 @@ cont_child_destroy_one(void *vin)
 
 		cont_child_stop(cont);
 
+		/*
+		 * This might be racy, as dtx_resync() might yeild after
+		 * ds_cont_child_lookup(), but before @sc_dtx_resyncing set,
```
jolivier23, line 1204: Need to fix the spelling errors, they will affect later PRs

jolivier23 (dismissed):

src/container/srv_target.c
```diff
@@ -1209,10 +1221,13 @@ cont_child_destroy_one(void *vin)
...
-		/* Make sure rebuild has stopped */
+		/**
+		 * After @sc_stopping is 1, all cont lookup will fail, so we wait
+		 * existed ds_cont users exit, it might come from rebuild, dtx.
+		 */
 		ABT_mutex_lock(cont->sc_mutex);
-		if (cont->sc_rebuilding)
-			ABT_cond_wait(cont->sc_rebuild_cond, cont->sc_mutex);
+		if (!daos_lru_is_last_user(&cont->sc_list))
+			ABT_cond_wait(cont->sc_fini_cond, cont->sc_mutex);
```
NiuYawei, line 1230: Do we abort dtx resync & rebuild on container destroy or just wait for them done here?

I think if the dtx resync & rebuild just hold container for a short period, we could just retry the loop in function for few more times, if they hold container for a long period, we'd inform them to abort.

Looks not necessary to introduce this sc_fini_cond and hack the cont_child_put()?

src/container/srv_target.c
```diff
@@ -1192,6 +1199,11 @@ cont_child_destroy_one(void *vin)
 
 		cont_child_stop(cont);
 
+		/*
+		 * This might be racy, as dtx_resync() might yield after
+		 * ds_cont_child_lookup(), but before @sc_dtx_resyncing set,
+		 * we use @sc_fini_cond to guarantee all users exit properly.
+		 */
```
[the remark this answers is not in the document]
wangshilong: @Nasf-Fan I think DTX will check sc_stopping, it call ds_cont_child_lookup() internally and inside ds_cont_child_lookup() it will check sc_stopping and return error if sc_stopping has been set.

src/container/srv_target.c
```diff
@@ -752,7 +752,14 @@ ds_cont_child_cache_destroy(struct daos_lru_cache *cache)
 static void
 cont_child_put(struct daos_lru_cache *cache, struct ds_cont_child *cont)
 {
+	bool		 wake_cond = false;
+
+	if (cont->sc_stopping == true && daos_lru_ref_count(&cont->sc_list) == 3)
```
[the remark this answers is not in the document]
wangshilong: Because in cont_child_destroy_one() it hold one extra refrences, and conf ref is inited as 1. so if ref count is 2 that means there is no other references for this containers.

Logic is like firstly we set sc_stopping for containers, and new lookup will fail.
but we wait existed references finish, that is why we could not wake up as long as sc_stopping is set. makes sense?

src/container/srv_target.c
wangshilong: I think we abort resync, but for rebuild is a bit different because it is not based on container but based on pool. now the logic if we found container is destroying, we will skip that container scan/pulling before. but there that might take a bit longer if container has been held, we need wait existed operation finished..

src/container/srv_target.c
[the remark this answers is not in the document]
NiuYawei: I think we'd try to avoid introducing such ad-hoc reference checking code further.

I think the '3' means we assume the daos_lru_cache for ds_cont_child is always created without D_HASH_FT_EPHEMERAL (no reference increasing when adding to cache) flag, I think we'd better don't make such assumption.

It looks to me current daos_lru code is flawed: in daos_lru_ref_release() it assumes 'in cache' holds one reference, but daos_lru_cache_create() doesn't regard D_HASH_FT_EPHEMERAL as an invalid feature for daos_lru.

I suggest implementing the common 'waiting for the last reference on daos_lru_ref_release()' for daos_lru, for example: introduce two more callbacks in daos_llink_ops, one for 'wakeup' and the other for 'wait', they could be called in daos_lru_ref_release() accordingly.

BTW, if this ticket isn't urgent, should we use this opportunity try to incorporate some the changes for what we discussed on I/O meeting? Thanks.

src/container/srv_target.c
NiuYawei: I think the rebuild code needs be changed to abort current iterating jobs (for scan or pull)?

src/common/lru.c
```diff
@@ -269,3 +287,17 @@ daos_lru_ref_release(struct daos_lru_cache *lcache, struct daos_llink *llink)
...
+{
+	lru_ref_release_internal(lcache, llink, false);
+}
+
+void
+daos_lru_ref_wait_evict(struct daos_lru_cache *lcache, struct daos_llink *llink)
+{
+	D_ASSERT(lcache->dlc_ops->lop_wait);
+
+	lru_ref_release_internal(lcache, llink, false);
```
NiuYawei, line 302: s/false/true

wangshilong: Sorry, indeed.

NiuYawei (approved):

src/container/srv_target.c
```diff
@@ -1232,8 +1260,7 @@ cont_child_destroy_one(void *vin)
 			D_GOTO(out_pool, rc = -DER_BUSY);
 		} /* else: resync should have completed, try again */
 
-		/* If it is the last user, ds_cont_child will be removed from hash & freed. */
-		cont_child_put(tls->dt_cont_cache, cont);
+		daos_lru_ref_wait_evict(tls->dt_cont_cache, &cont->sc_list);
```
Nasf-Fan, line 1263: How can we prevent new coming IO during the wait? If cannot, more and more new IO request may block the destroy process for ever?

Nasf-Fan: After discussed with shilong, it seems work well.

Nasf-Fan (approved):

PR #13877, opened 2024-02-26
DAOS-623 build: Use space in meson options

jolivier23:
While = may work, in the documentation, spaces are used so let's try this and see if it fixes build issues we've seen.

### Gatekeeper:


site_scons/components/__init__.py
```diff
@@ -301,9 +301,9 @@ def define_components(reqs):
     reqs.define('fuse', libs=['fuse3'], defines=['FUSE_USE_VERSION=35'],
                 retriever=GitRepoRetriever('https://github.com/libfuse/libfuse.git'),
                 commands=[['meson', 'setup', '../fuse'],
-                          ['meson', 'configure', '--prefix=$FUSE_PREFIX', '-Ddisable-mtab=True',
-                           '-Dudevrulesdir=$FUSE_PREFIX/udev', '-Dutils=False',
-                           '--default-library', 'both'],
+                          ['meson', 'configure', '--prefix', '$FUSE_PREFIX', '-D',
```
daltonbohning, line 304: Where does FUSE_PREFIX get set?

jolivier23: in site_scons/prereq_tools/base.py _modify_prefix

`            self._save_component_prefix(f'{comp_def.name.upper()}_PREFIX', '/usr')`

jolivier23: actually, in this case, it would be in get_prefixes

        self._save_component_prefix(comp_prefix, target_prefix)

jamesanunez (approved):

ashleypittman (changes requested):
This is fine for diagnostics but we shouldn't be landing it unless we can prove it solves the problem.

jolivier23:
> This is fine for diagnostics but we shouldn't be landing it unless we can prove it solves the problem.

sure, I suppose that remains to be seen.  @jamesanunez are you able to reproduce with this patch?

daltonbohning:
> > This is fine for diagnostics but we shouldn't be landing it unless we can prove it solves the problem.
> 
> sure, I suppose that remains to be seen. @jamesanunez are you able to reproduce with this patch?

This does seem to resolve a similar issue on Frontera. Without this PR, meson tries to configure fuse in `/usr`. But with this, it correctly uses my local prefix

ashleypittman:
> > > This is fine for diagnostics but we shouldn't be landing it unless we can prove it solves the problem.
> > 
> > 
> > sure, I suppose that remains to be seen. @jamesanunez are you able to reproduce with this patch?
> 
> This does seem to resolve a similar issue on Frontera. Without this PR, meson tries to configure fuse in `/usr`. But with this, it correctly uses my local prefix

If you can verify this change fixes the problem then yes, let's land it, absolutely.

daltonbohning:
> > This does seem to resolve a similar issue on Frontera. Without this PR, meson tries to configure fuse in `/usr`. But with this, it correctly uses my local prefix
> 
> If you can verify this change fixes the problem then yes, let's land it, absolutely.

Actually. I suspect some kind of race condition here. It sometimes works and sometimes does not.
I just had another build fail with
```
Installation failed due to insufficient permissions.
Attempting to use polkit to gain elevated privileges...
Error executing command as another user: Not authorized
```

And at least the fuse configure looks right
```
meson configure --prefix /work2/08126/dbohninx/frontera/BUILDS/dev/20240307/daos/install/prereq/release/fuse -D disable-mtab=True -D udevrulesdir=/work2/08126/dbohninx/frontera/BUILDS/dev/20240307/daos/install/prereq/release/fuse/udev -D utils=False --default-library both
```

daltonbohning:
> Maybe `ninja install` needs to specify a prefix too?

Nevermind. According to https://stackoverflow.com/questions/62475190/meson-and-ninja-build-system-specify-where-binaries-are-stored
the prefix should be set with meson, not ninja.

ashleypittman:
> > > This does seem to resolve a similar issue on Frontera. Without this PR, meson tries to configure fuse in `/usr`. But with this, it correctly uses my local prefix
> > 
> > 
> > If you can verify this change fixes the problem then yes, let's land it, absolutely.
> 
> Actually. I suspect some kind of race condition here. It sometimes works and sometimes does not. I just had another build fail with
> 
> ```
> RUN: ninja install
> [0/1] Installing files.
> Installing lib/libfuse3.so.3.16.2 to /usr/local/lib64
> Installation failed due to insufficient permissions.
> Attempting to use polkit to gain elevated privileges...
> Error executing command as another user: Not authorized
> ```
> 
> And at least the fuse configure looks right
> 
> ```
> meson configure --prefix /work2/08126/dbohninx/frontera/BUILDS/dev/20240307/daos/install/prereq/release/fuse -D disable-mtab=True -D udevrulesdir=/work2/08126/dbohninx/frontera/BUILDS/dev/20240307/daos/install/prereq/release/fuse/udev -D utils=False --default-library both
> ```

The evidence here all points to a `meson` bug to me.  Presumably an old one as it's only on el7 and not when using virtual environments which will be using newer software.  If that's the case (and it's still happening on supported distros) then we should add a check for the meson version in our scons code.

daltonbohning:
> The evidence here all points to a `meson` bug to me. Presumably an old one as it's only on el7 and not when using virtual environments which will be using newer software. If that's the case (and it's still happening on supported distros) then we should add a check for the meson version in our scons code.

I'm trying to use newer meson and ninja now to see if it fixes the issue. Will report back when I know

daltonbohning:
> > The evidence here all points to a `meson` bug to me. Presumably an old one as it's only on el7 and not when using virtual environments which will be using newer software. If that's the case (and it's still happening on supported distros) then we should add a check for the meson version in our scons code.
> 
> I'm trying to use newer meson and ninja now to see if it fixes the issue. Will report back when I know

I've gotten two successful builds now with updated meson and ninja. And @jamesanunez mentioned in a DM to me that might have fixed for him too. I'm going to try again without this patch

daltonbohning:
Another datapoint, without this patch and updated meson + ninja I still got a failure.
This time, meson explicitly says it's not a bug. Probably because it's such a common issue (and probably actually a bug)
```
RUN: ninja install
[0/1] Installing files.
Installing lib/libfuse3.so.3.16.2 to /usr/local/lib64
Installation failed due to insufficient permissions.
...
ERROR: Unhandled python OSError. This is probably not a Meson bug, but an issue with your build environment.
```

daltonbohning:
This is 99% most definitely a bug with meson. Trying manually
```
$ meson configure --prefix /work2/08126/dbohninx/frontera/BUILDS/dev/20240307/daos/install/prereq/release/fuse

$ meson install
ninja: Entering directory `/work2/08126/dbohninx/frontera/BUILDS/dev/20240307/daos/build/external/release/fuse.build'
ninja: no work to do.
Installing lib/libfuse3.so.3.16.2 to /usr/local/lib64
Installation failed due to insufficient permissions.
Attempt to use /usr/bin/sudo to gain elevated privileges? [y/n]
```

And the configuration file shows the prefix as I have set it.


There seems to be a workaround to use destdir, which overrides the prefix.
Older versions of meson:
```
DESTDIR=<path> ninja install
```
Newer versions of meson since 0.47.0. Also, use `meson` instead of `ninja` directly. (`ninja` still used in the backend)
```
meson install --destdir <path>
```

daltonbohning:
The fix that works for me is #13961

daltonbohning:
FWIW I've been able to reproduce the prefix issue on boro with EL 8.8. So I suspect master is currently doing the wrong thing for other people's builds

daltonbohning:
Others have reported this on boro/wolf, so https://daosio.atlassian.net/browse/DAOS-15417

daltonbohning:
Recommend closing this in favor of #13961

PR #13881, opened 2024-02-27
DAOS-15256 object: coll ops against object with sparse ranks

Nasf-Fan:
We define the degree of rank# sparse for an object as: (the count of ranks on which the object shards reside) / (the count of the ranks in the pool map)

For example: the pool is built on the 10-ranks set: [1, 3, 5, 7, 9, 10, 11, 12, 13, 14]

Case1: If the object shards reside on the ranks set [1, 3, 9, 5], then its sparse degree is equal to 4 / 10 = 0.4

Case2: If the object shards reside on the ranks set [12, 10, 13], then its sparse degree is equal to 3 / 10 = 0.3

Based on such sparse degree, when handle collective operation on client, we will prepare related target bitmap as following:

If the sparse degree is not less than 0.35, then pre-allocate continuous array (that indexed with the rank#) to record related targets bitmap for the collective operation. It is the most efficient solution for locating the rank slot with the rank#. It may waste some DRAM space, but that is temporary before sending out the collective RPC. The percentage of waste DRAM will not be more than 65%.

Otherwise, if the sparse degree is less than 0.5, then use b+tree as the basic data structure to record related targets bitmap. It will introduce some overhead for data comparing and movement when locate the rank slot with the rank#, but it avoids occupying too much DRAM space, especially when the sparse degree is very small.

In practice, before completely scanning the object layout, we cannot exactly know on which ranks the object shards reside. So we need to calculate the sparse degree based on some estimation. For example, when generate the object layout, record the maximum and the minimum rank# on which related object shards reside, then use the width of the ranks range to estimate the ranks count for the given object. Because these ranks may be not continuous, then the count may be over-estimated. Consider that, we can adjust the boundary of sparse degree for above ranks array and b+tree solutions, such as 0.35 => 0.45, or similarly.

Features: offline_extend

### Gatekeeper:


Nasf-Fan:
telemetry_list failed for DAOS-15337, to be retested.

src/object/cli_mod.c
```diff
@@ -71,15 +69,24 @@ dc_obj_init(void)
...
+	obj_coll_thd = OBJ_COLL_THD_MIN;
+	d_getenv_uint("DAOS_OBJ_COLL_THD", &obj_coll_thd);
+	if (obj_coll_thd == 0) {
+		D_INFO("Disable collective operation.\n");
+	} else {
+		if (obj_coll_thd < OBJ_COLL_THD_MIN) {
+			D_WARN("Invalid collective operation threshold %u, either larger than %u, "
+			       "or zero for disabling collective operation. Use default value %u\n",
+			       obj_coll_thd, OBJ_COLL_THD_MIN - 1, OBJ_COLL_THD_MIN);
+			obj_coll_thd = OBJ_COLL_THD_MIN;
```
gnailzenh, line 86: if I'm understanding correctly, if the threshold is too high, it means no collective correct?

gnailzenh (approved):

src/object/cli_coll.c
```diff
@@ -132,17 +365,30 @@ obj_coll_prep_one(struct coll_oper_args *coa, struct dc_object *obj,
 	if (rc != 0 || (shard->do_rebuilding && !coa->coa_for_modify))
 		goto out;
 
-	/* The ranks for the pool may be not continuous, let's extend the coa_dcts array. */
-	if (unlikely(shard->do_target_rank >= coa->coa_dct_cap)) {
-		D_REALLOC_ARRAY(dct, coa->coa_dcts, coa->coa_dct_cap, shard->do_target_rank + 4);
-		if (dct == NULL)
-			D_GOTO(out, rc = -DER_NOMEM);
+	D_RWLOCK_RDLOCK(&obj->cob_lock);
```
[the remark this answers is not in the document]
Nasf-Fan: To prevent the new added cob_xxx being changed by others by race.

src/object/cli_coll.c
```diff
@@ -57,20 +221,74 @@ obj_coll_oper_args_init(struct coll_oper_args *coa, struct dc_object *obj, bool
...
+}
+
+static int
+obj_coll_collapse_tree(struct coll_oper_args *coa, uint32_t *size)
+{
+	struct coll_sparse_targets	*tree = coa->coa_tree;
+	int				 rc = 0;
+
+	if (unlikely(coa->coa_dct_cap == 0))
+		D_GOTO(out, rc = 1);
```
liuxuezhao, line 266: this func's caller obj_coll_oper_args_collapse() does not differenciate rc 1 or rc 0, why here return 1 rather 0?

src/object/cli_coll.c
```diff
@@ -101,14 +319,31 @@ obj_coll_oper_args_collapse(struct coll_oper_args *coa, uint32_t *size)
...
+	if (coa->coa_sparse)
+		rc = obj_coll_collapse_tree(coa, size);
+	else
+		rc = obj_coll_collapse_array(coa, size);
+
+	if (rc >= 0) {
+		obj->cob_rank_nr = coa->coa_dct_nr;
+		/* If all shards are NONEXIST, then need not to send RPC(s). */
+		if (unlikely(coa->coa_dct_nr == 0))
+			rc = 1;
```
liuxuezhao, line 344: same, is it really useful for rc 1 for caller?

src/object/cli_coll.c
```diff
@@ -132,17 +371,30 @@ obj_coll_prep_one(struct coll_oper_args *coa, struct dc_object *obj,
...
+		rc = dbtree_upsert(coa->coa_tree->cst_tree_hdl, BTR_PROBE_EQ, DAOS_INTENT_UPDATE,
+				   &kiov, &riov, &viov);
+		if (rc != 0)
+			goto out;
 
-		coa->coa_dcts = dct;
-		coa->coa_dct_cap = shard->do_target_rank + 4;
+		dct = viov.iov_buf;
+	} else {
+		dct = &coa->coa_dcts[shard->do_target_rank - obj->cob_min_rank];
```
liuxuezhao, line 394: looks possible after unlock, the obj refresh layout and the "obj->cob_min_rank" changed, in that case is it still safe to use the dct pointer to assign at below code?

liuxuezhao (approved):

src/object/cli_coll.c
```diff
@@ -20,30 +20,194 @@
...
 
+static int
+coll_rec_alloc(struct btr_instance *tins, d_iov_t *key_iov, d_iov_t *val_iov,
+	       struct btr_record *rec, d_iov_t *val_out)
+{
+	struct daos_coll_target	*dct;
+	struct coll_oper_args	*coa = val_iov->iov_buf;
+	int			 rc = 0;
+
+	D_ALLOC_PTR(dct);
```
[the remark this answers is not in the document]
Nasf-Fan: We will return the new "dct" via "d_iov_set(val_out, dct, sizeof(*dct));". If use umem_zalloc(), then need umem_off2ptr() inside d_iov_set(). That is almost the same overhead.

src/object/cli_coll.c
```diff
@@ -20,30 +20,194 @@
...
+		return false;
+
+	if (*shard_nr > obj_coll_thd)
+		 return true;
+
+	if (*shard_nr <= 4)
+		return false;
+
+	D_RWLOCK_RDLOCK(&obj->cob_lock);
+	if (*shard_nr >= (obj->cob_max_rank - obj->cob_min_rank + 1) * 2)
```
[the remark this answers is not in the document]
Nasf-Fan: We may not know the real number of ranks in the object until after parsing the whole object layout via obj_coll_oper_args_collapse(). Here cob_{min,max}_rank are always available. But of course, we can make some optimization if we know cob_rank_nr. I will enhance that in subsequent patch.

src/object/cli_coll.c
Nasf-Fan: Returning 1 will make the caller to "goto out" with positive number, that will not fail the operation.

src/object/cli_coll.c
Nasf-Fan: The same as above.

src/object/cli_obj.c
```diff
@@ -358,7 +362,15 @@ obj_layout_create(struct dc_object *obj, unsigned int mode, bool refresh)
...
 		obj_shard->do_target_rank = layout->ol_shards[i].po_rank;
 		obj_shard->do_target_idx = layout->ol_shards[i].po_index;
+		if (obj->cob_min_rank > obj_shard->do_target_rank)
+			obj->cob_min_rank = obj_shard->do_target_rank;
+		if (obj->cob_max_rank < obj_shard->do_target_rank)
+			obj->cob_max_rank = obj_shard->do_target_rank;
 	}
+
+	if (unlikely(obj->cob_min_rank == -1))
+		obj->cob_min_rank = 0;
```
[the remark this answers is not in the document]
Nasf-Fan: Yes, it will avoid unnecessary overhead. The "cob_rank_nr" will be only be set after parsing the whole object layout via obj_coll_oper_args_collapse() since collective operations are the users.

PR #13885, opened 2024-02-27
DAOS-623 ci: Modify githooks to allow skip

jolivier23:
Allow one to set DAOS_SKIP=exp1,exp2 to skip
git hooks on a case by case basis as needed.

Required-githooks: true
Skipped-githooks: flake

### Gatekeeper:

### DIFF
--- a/utils/githooks/commit-msg.d/10-watermark.py
+++ b/utils/githooks/commit-msg.d/10-watermark.py
@@ -30,6 +30,15 @@ def find_hooks():
     return hooks
 
 
+def emit_watermark(msg):
+    """Print the Required-githooks value"""
+    msg.write("Required-githooks: true\n")
+    skipped = os.environ.get("DAOS_SKIP", None)
+    if skipped:
+        msg.write(f"Skipped-githooks: {skipped}\n")
+    msg.write("\n")
+
+
 def run_check():
     """Run the checks for the required commit hooks"""
     empty_commit = True
@@ -60,10 +69,10 @@ def run_check():
         for line in msg:
             if not hook_emitted:
                 if line.startswith("Signed-off-by"):
-                    commit_msg.write("Required-githooks: true\n\n")
+                    emit_watermark(commit_msg)
                     hook_emitted = True
                 elif line.startswith("#"):
-                    commit_msg.write("Required-githooks: true\n\n")
+                    emit_watermark(commit_msg)
                     hook_emitted = True
             commit_msg.write(line)
 
--- a/utils/githooks/pre-commit
+++ b/utils/githooks/pre-commit
@@ -7,12 +7,22 @@ export TARGET
 hook=${0##*/}
 rm -f ".${hook}"
 
+IFS=', ' read -r -a skip_list <<< "${DAOS_SKIP:-none}"
+echo "skip_list=${skip_list[*]}"
+
 run-parts() {
     local dir="$1"
     shift
 
     for i in $(LC_ALL=C; echo "${dir%/}"/*[^~,]); do
-        # don't run vim .swp files
+        skip_item=0
+        for skip in "${skip_list[@]}"; do
+            if [[ "${i}" =~ ${skip} ]]; then
+                skip_item=1
+                break
+            fi
+        done
+        [ $skip_item = 1 ] && continue
         [ "${i%.sw?}" != "${i}" ] && continue
         $i "$@"
     done


jolivier23:
Tested with subsequent commit passing multiple options:

`commit 746485e57f6ead549355aaa988b4d344bab06951 (HEAD -> jvolivie/test)
Author: Jeff Olivier <jeffolivier@google.com>
Date:   Tue Feb 27 14:09:13 2024 -0700

    do a check
    
    Required-githooks: true
    Skipped-githooks: clang,flake
    
    Signed-off-by: Jeff Olivier <jeffolivier@google.com>
`

utils/githooks/commit-msg.d/10-watermark.py
```diff
@@ -30,6 +30,15 @@ def find_hooks():
     return hooks
 
 
+def emit_watermark(msg):
+    """Print the Required-githooks value"""
+    msg.write("Required-githooks: true\n")
+    skipped = os.environ.get("DAOS_SKIP", None)
+    if skipped:
+        msg.write(f"Skipped-githooks: {skipped}\n")
+    msg.write("\n")
```
daltonbohning, line 39: FWIW, I think we should phase out the Required-githooks watermark. They're really just a convenience, since we have GHA to do the same checks. The only one missing (I think) is the copyright githook doesn't have a GHA.

There are holes in the watermark, like:
1. Users can just manually paste the watermark. As a gatekeeper, I see this A LOT.
2. It doesn't get added in every scenario. E.g. some merges.
3. Even this new "Skipped-githooks" wouldn't be added if you use `--amend`

Thoughts @brianjmurrell?

utils/githooks/pre-commit
```diff
@@ -7,12 +7,22 @@ export TARGET
...
     shift
 
     for i in $(LC_ALL=C; echo "${dir%/}"/*[^~,]); do
-        # don't run vim .swp files
+        skip_item=0
+        for skip in "${skip_list[@]}"; do
+            if [[ "${i}" =~ ${skip} ]]; then
+                skip_item=1
+                break
+            fi
```
daltonbohning, line 23: Should this print what is being skipping?

utils/githooks/commit-msg.d/10-watermark.py
```diff
@@ -30,6 +30,15 @@ def find_hooks():
     return hooks
 
 
+def emit_watermark(msg):
+    """Print the watermark to the commit message"""
+    msg.write("Required-githooks: true\n")
+    skipped = os.environ.get("DAOS_SKIP", None)
```
daltonbohning, line 36: I don't know a better name, but `DAOS_SKIP` seems vague with regard to the githooks.

utils/githooks/commit-msg.d/10-watermark.py
jolivier23: Honestly, this would be fine with me but we should fix the copyright issue and check that in gha

utils/githooks/commit-msg.d/10-watermark.py
jolivier23: fair enough

utils/githooks/pre-commit
jolivier23: not a bad idea

utils/githooks/commit-msg.d/10-watermark.py
daltonbohning: Agree. Beyond the scope of this PR anyway, but I guess where I was going with that is that if @brianjmurrell agrees on phasing out the watermark, you wouldn't need to bother touching it

daltonbohning (approved):

utils/githooks/commit-msg.d/10-watermark.py
brianjmurrell: > FWIW, I think we should phase out the Required-githooks watermark. They're really just a convenience, since we have GHA to do the same checks. The only one missing (I think) is the copyright githook doesn't have a GHA.

Should we fix that before removing the watermark?  `Jenkinsfile` would need updating if/when removing that.

> There are holes in the watermark, like:
> 
> 1. Users can just manually paste the watermark. As a gatekeeper, I see this A LOT.

I would liken this to faking test results to get your PR landed and treated as such (with whatever kind of consequences that might have).  Not using them does not fail your PR so there is on reason to fake them.

> 2. It doesn't get added in every scenario. E.g. some merges.

That would be a bug that needs fixing, assuming we were keeping them.

> 3. Even this new "Skipped-githooks" wouldn't be added if you use `--amend`

This is a general problem of having a facility to carry pragams forward in commits.  I have some hackery around that (but it's rough and fails all kinds of corner cases) as does @ashleypittman I believe.  I wonder how fail-proof his is.

daltonbohning: Yes, I do think we need a copyright GHA before removing the watermark.
As for faking Required-githooks, I think think the problem is that the watermark so often doesn't get added, or the githooks themselves fail for various reasons. So I suspect people have just added the watermark to their standard copy-paste commit pragmas so they don't see the stage warning. It's hard to enforce not faking it because (one example):
You could modify a lot of files and use `--amend --no-verify` and the hook would still be there from the first commit. This one is nearly impossible to enforce because if you use `--no-verify` the hooks don't run. So no amount of intelligent hooks can work around this.

I think the core problem is that we're effectively relying on a client-side check to tell the server (GitHub/us) that something is correct. That's why I think removing the watermark entirely and having a corresponding GHA for each hook is the proper solution.


> This is a general problem of having a facility to carry pragams forward in commits. I have some hackery around that (but it's rough and fails all kinds of corner cases)

According to [githooks.com](https://githooks.com/), there are 17 hooks, most of which we don't use. To account for most corner cases, I think we would need to break the githooks into utility files and write duplicate hooks for the various hook points.

utils/githooks/pre-commit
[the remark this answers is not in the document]
jolivier23: I took a stab at that...seems to work

brianjmurrell:
Not sure if the refactoring/file renaming resulted in a bunch of my prior suggestions being auto-resolved or what, but I've unresolved them as they were resolved without applying or any comment otherwise -- which is why I wondered if that was some kind of auto-resolving done by GH.

jolivier23:
> Not sure if the refactoring/file renaming resulted in a bunch of my prior suggestions being auto-resolved or what, but I've unresolved them as they were resolved without applying or any comment otherwise -- which is why I wondered if that was some kind of auto-resolving done by GH.

No, it actually is the case that I had two copies of hook_base and I edited the wrong one.  I didn't check the uploaded changes.  Sorry about that.

utils/githooks/pre-commit
```diff
@@ -7,13 +7,24 @@ export TARGET
 hook=${0##*/}
 rm -f ".${hook}"
 
+IFS=', ' read -r -a skip_list <<< "${DAOS_GITHOOK_SKIP:-none}"
```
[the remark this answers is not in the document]
jolivier23: I got "unbound string" when I tried this but :- does what you want, I think.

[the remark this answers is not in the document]
brianjmurrell: Indeed `${…:-}` is the solution for an unbound string that can be empty.

brianjmurrell (approved):

utils/githooks/pre-commit
```diff
@@ -7,13 +7,24 @@ export TARGET
...
 run-parts() {
     local dir="$1"
     shift
 
     for i in $(LC_ALL=C; echo "${dir%/}"/*[^~,]); do
         # don't run vim .swp files
         [ "${i%.sw?}" != "${i}" ] && continue
+        skip_item=0
+        for skip in "${skip_list[@]}"; do
+            if [[ "${i}" =~ ${skip} ]]; then
```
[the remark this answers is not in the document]
jolivier23: I think we can refine it later if needed.

PR #13927, opened 2024-03-04
DAOS-15338 dfuse: Configure fuse to allow writes to populate cache.

ashleypittman:
Change dfuse data caching settings to allow writes to populate cache.
Track the data cache validity separately from attribute cache validity to
allow finer grained control.

On close of a file then speculatively call getattr to get updated size/mtime
information and use this as a reference for future dcache invalidation.

For getattr calls after close then use the pre-loaded stat data leading to
increased performance by moving the getattr off the critical path (only for
files which were written to, files which aren't do not invalidate the atttr cache).

Use the FUSE_CAP_EXPLICIT_INVAL_DATA kernel feature to instruct the
kernel not to automatically invalidate data in the size/mtime has changed, this
is now performed by dfuse.

Overall this is a performance improvement for stat-after-close but the major
change is that a write/read workflow will now correctly use the kernel page
cache.  Previously writes were modifying mtime values and that was incorrectly
triggering the kernel to discard it's cache unnecessarily.

Add testing of write/read and read/read I/O patterns verified by the dfuse
statistics that data caching is working as expected.


jolivier23 (dismissed):

src/tests/ftest/dfuse/read.py
```diff
@@ -96,3 +96,98 @@ def test_dfuse_pre_read(self):
...
+        pool = self.get_pool(connect=False)
+        container = self.get_container(pool)
+
+        self.load_dfuse(self.hostlist_clients, None)
+
+        self.dfuse.disable_wb_cache.value = True
+
+        self.dfuse.env["D_LOG_MASK"] = "INFO,DFUSE=DEBUG"
+        self.dfuse.env["DD_MASK"] = "ALL"
+        self.dfuse.env["DD_SUBSYS"] = "ALL"
```
daltonbohning, line 131: Do you want this set for both test cases?
If so, we can put this in the config
```
dfuse:
  disable_wb_cache: true
  env_vars:
    - D_LOG_MASK=INFO,DFUSE=DEBUG
    - DD_MASK=ALL
    - DD_SUBSYS=ALL
```


And if it needs to be different, I'd recommend different namespaces like
```
dfuse:
  disable_wb_cache: true
dfuse_debug:
  disable_wb_cache: true
  env_vars:
    - D_LOG_MASK=INFO,DFUSE=DEBUG
    - DD_MASK=ALL
    - DD_SUBSYS=ALL
```

And then initialize like
```
self.load_dfuse(self.hostlist_clients, '/run/dfuse/*')
self.load_dfuse(self.hostlist_clients, '/run/dfuse_debug/*')
```

src/tests/ftest/dfuse/read.py
```diff
@@ -96,3 +96,98 @@ def test_dfuse_pre_read(self):
...
+        result = run_remote(self.log, self.hostlist_clients, cmd)
+        if not result.passed:
+            self.fail(f'"{cmd}" failed on {result.failed_hosts}')
+
+        data = self.dfuse.get_stats()
+
+        read_calls = data["statistics"].get("read", 0)
+        write_calls = data["statistics"].get("write")
+
+        print(f"Test caused {write_calls} write and {read_calls} reads calls")
```
daltonbohning, line 161:
```suggestion
        self.log.info("Test caused %d write and %d reads calls", write_calls, read_calls)
```

src/client/dfuse/ops/open.c
```diff
@@ -203,16 +304,22 @@ dfuse_cb_release(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
...
 	if (oh->doh_evict_on_close) {
 		ie = oh->doh_ie;
 		atomic_fetch_add_relaxed(&ie->ie_ref, 1);
 	}
 
+	if (update_data_cache) {
+		dfuse_dcache_set_time(oh->doh_ie);
+		if (oc == 1)
+			/* Fire off a ostatx call to fetch the new mtime */
+			size_resample(dfuse_info, oh->doh_ie);
```
mchaarawi, line 320: a stat on a file close is going to be pretty bad for workloads that have millions of files ..

ashleypittman: This is only for files which were written to.  If there's any access to the file afterwards then the kernel would do this on next access anyway so the only additional network load here is for files which are written to but then not accessed again.

mchaarawi: most of workloads we have seen today write/read to the file with 1 handle. i.e. there is no close and re-access.
What workload are you thinking about here that does this write, close, open, read that would benefit from this?

src/client/dfuse/dfuse_core.c
```diff
@@ -1048,6 +1064,51 @@ dfuse_mcache_get_valid(struct dfuse_inode_entry *ie, double max_age, double *tim
...
+bool
+dfuse_dc_cache_get_valid(struct dfuse_inode_entry *ie, double max_age, double *timeout)
+{
+	bool            use = false;
+	struct timespec now;
+	struct timespec left;
+	double          time_left;
+
+	D_ASSERT(max_age != -1);
+	D_ASSERT(max_age >= 0);
```
wiliamhuang, line 1086: Minor issue, if "max >= 0", then the first assertion always is true. Do we need the first assertion?

src/client/dfuse/dfuse_core.c
```diff
@@ -1048,6 +1064,51 @@ dfuse_mcache_get_valid(struct dfuse_inode_entry *ie, double max_age, double *tim
 	return use;
 }
 
+/* Set a timer to mark cache entry as valid */
```
wiliamhuang, line 1067: minor: "timer"-->"time"?

src/client/dfuse/ops/open.c
wiliamhuang: Most applications do not read the files in a short time after writing files. Some applications (e.g., computation chemistry) may do write & close then open & read later because intermediate data are too large to fit in memory.

src/client/dfuse/dfuse_core.c
ashleypittman: True, we don't.  I've used this check elsewhere in the code though, for some cache types (data) then -1 is a valid cache age so the reason for having two checks here is that if it does fail we know why, was the cache incorrectly initialised/loaded in which case it'll be -1 or was a specific age incorrectly calculated in which case it's be a different negative value.

src/client/dfuse/dfuse.h
```diff
@@ -998,6 +1000,37 @@ struct dfuse_inode_entry {
 	/** File has been unlinked from daos */
 	bool                      ie_unlinked;
 
+	/* Data cache metadata, list known size/mtime for file, if these have been updated then
```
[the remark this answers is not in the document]
ashleypittman: This is the information used to determine if the data cache is to be considered valid - it's data about the data cache.

src/client/dfuse/ops/open.c
ashleypittman: You're right this will help with some I/O patterns and not others, I don't think we can speculate accurately on which is more common but the bug is clear - with current master the cache is dropped on open in the open/write/close -> open/read/close case which is a simple bug due to having the kernel handle cache validity decisions based on incorrect assumptions on how mtime updates are handled.

Some applications and workflows won't be triggering this bug and will therefore be unaffected by it, a lot more will be triggering it and benefit from the fix, interactive workloads will probably benefit the most but plenty of HPC ones will.  IOR certainly would if using the `-C`/reorder flag and the right rank allocation (although this option is mainly there to force clients onto other nodes to avoid the use of in-node caching).

mchaarawi: my main concern is the effect of an extra stat on those unaffected workloads. i do not agree that they are "unaffected".
anyway since you mentioned IOR is a use case that can benefit, let's gather some results then before we decide on this fix.

ashleypittman: I did have some numbers on this when I was developing it and it was showing a very high proportion of these speculative stat calls were returned to the kernel - but it'll be entirely application dependant as you say.

Anyway - the extra stat is asynchronous but launched on close so it'll be impossible to measure from a client perspective as the extra cost, if it is indeed extra, is all server side so won't show up in any simple benchmark.

src/client/dfuse/dfuse_core.c
wiliamhuang: I agree. Thank you for your clarification!

src/client/dfuse/dfuse.h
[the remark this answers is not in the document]
mchaarawi: yea maybe reword to something like "metadata for the data-cache"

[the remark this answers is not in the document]
ashleypittman: I can re-push with that, any other changes whilst I'm at it?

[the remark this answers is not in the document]
mchaarawi: this is just a minor comment that does not require a repush. there is a bigger concern on the stat after close that is more fundamental and needs to be discuss more before any repush..

src/client/dfuse/ops/open.c
```diff
@@ -132,6 +135,101 @@ dfuse_cb_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
...
+
+	rc = dfs_ostatx(ie->ie_dfs->dfs_ns, ie->ie_obj, &ie->ie_dc.stat, &ev->de_ev);
+	if (rc != 0)
+		goto err;
+
+	atomic_fetch_add(&ie->ie_dc.active, 1);
+
+	sem_post(&eqt->de_sem);
+
+	return;
```
knard-intel, line 226: In case of success, the resources such as `ev` are not freed ?

knard38 (commented):
Mostly LGTM for what I understand.
Some points to clarify as I am not used to this part of the code.

mchaarawi (changes requested):
as discussed in WG, the stat on close is pretty significant hit performance wise and to the DAOS cluster itself in case of large deployments and widely striped files

mchaarawi:
> @mchaarawi having the extra asynchronous getattr on close will be a big win for many of the first impression applications (such as when people try to clone and compiler in a dfuse mount). In such cases, fuse will do subsequent getattr anyway, this just caches it preemptively and, more importantly, keeps the write cache for the subsequent reads. For performance sensitive cases, I suspect the extra asynchronous stat is somewhat negligible compared to the expected normal overhead of dfuse. It already does a million getattr's.

im pretty sure that this does server some use cases. im worried though about use cases that are just now doing an extra stat on close.. say an app that just writes a file and never accesses it again till much later. the stat in this case of widely stripped files if there are a bunch of them, would be bad. so yes this probably improves 1 use case but can be problemetic in others

mchaarawi:
> Actually, maybe we can augment dfs_write to return the max mtime for the chunks it writes? And we can approximate the size based on what we wrote and what we already know?

the io path should never be modified IMO to do extra metadata things like this.

jolivier23:
> > @mchaarawi having the extra asynchronous getattr on close will be a big win for many of the first impression applications (such as when people try to clone and compiler in a dfuse mount). In such cases, fuse will do subsequent getattr anyway, this just caches it preemptively and, more importantly, keeps the write cache for the subsequent reads. For performance sensitive cases, I suspect the extra asynchronous stat is somewhat negligible compared to the expected normal overhead of dfuse. It already does a million getattr's.
> 
> 
> 
> im pretty sure that this does serve some use cases. im worried though about use cases that are just now doing an extra stat on close.. say an app that just writes a file and never accesses it again till much later. the stat in this case of widely stripped files if there are a bunch of them, would be bad. so yes this probably improves 1 use case but can be problemetic in others

Along the lines of my next comment, what if we locally tracked size and mtime on the client object and allowed the application to query the values.  This would be a good approximation in the cases I'm talking about and just get invalidated for others.

mchaarawi:
> > > @mchaarawi having the extra asynchronous getattr on close will be a big win for many of the first impression applications (such as when people try to clone and compiler in a dfuse mount). In such cases, fuse will do subsequent getattr anyway, this just caches it preemptively and, more importantly, keeps the write cache for the subsequent reads. For performance sensitive cases, I suspect the extra asynchronous stat is somewhat negligible compared to the expected normal overhead of dfuse. It already does a million getattr's.
> > 
> > 
> > im pretty sure that this does serve some use cases. im worried though about use cases that are just now doing an extra stat on close.. say an app that just writes a file and never accesses it again till much later. the stat in this case of widely stripped files if there are a bunch of them, would be bad. so yes this probably improves 1 use case but can be problemetic in others
> 
> Along the lines of my next comment, what if we locally tracked size and mtime on the client object and allowed the application to query the values. This would be a good approximation in the cases I'm talking about and just get invalidated for others.
> 
> In other words the local object would keep track of locally highest offset and mtime

the mtime we use today there is based on server time stamps (the max epoch thing). so if dfuse wants to use it's own timestamps to do this, there might be issues of client and servers are not properly synchronized.
but in a single case dfuse instance, it might not be a big deal?
we cannot do his at the dfs level because it we currently do not do any caching there. but we might do that in the future.

PR #13940, opened 2024-03-06
DAOS-14348: pil4dfs dentry caching

knard-intel:
### Description

TODO

### Gatekeeper:


utils/ansible/ftest/file/configure-nofile.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# set -x
+set -e -o pipefail
```
brianjmurrell, line 4: `-u` is another good flag to add so that you don't have variables that are unexpectedly unassigned going unchecked for assignment.  It does mean your code needs to use `${varname:-}` where you intentionally want to allowa  variable to be legitimately unassigned.

utils/ansible/ftest/file/enable-coredumps.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+# set -x
+set -e -o pipefail
```
brianjmurrell, line 4: Ditto.

knard-intel: - [ ] Add -u flags bash script header

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -914,14 +1175,16 @@ retrieve_handles_from_fuse(int idx)
 		goto err;
 	}
 
-	rc = d_hash_table_create(D_HASH_FT_EPHEMERAL | D_HASH_FT_MUTEX | D_HASH_FT_LRU, 6, NULL,
-				 &hdl_hash_ops, &dfs_list[idx].dfs_dir_hash);
-	if (rc) {
+	rc = dcache_init(&dfs_list[idx]);
```
mchaarawi, line 1178: I think we should add an option (env variable) to disable the dcache entirely (not the default, but just an option).

mchaarawi (commented):
nice patch! just a few comments from an initial review.
Can we move the dcache code itself into a separate file please and add an internal header in this case to avoid keep growing int_dfs.c endlessly?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -110,17 +120,35 @@ static uint16_t               eq_count_max;
 static uint16_t               eq_count;
 static uint16_t               eq_idx;
 
+/* Hash table entry for directory handles.  The struct and name will be allocated as a single
+ * entity.
+ */
+struct dcache_rec {
```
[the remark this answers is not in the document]
knard-intel: - [ ] Add field's documentation

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -100,6 +103,13 @@ static uint16_t               low_fd_count;
 /* the list of low fd reserved */
 static int                    low_fd_list[DAOS_MIN_FD];
 
+/* default dir cache time-out in seconds*/
+#define DCACHE_TIMEOUT 10
```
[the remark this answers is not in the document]
knard-intel: - [ ] Change default timeout

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -285,126 +313,358 @@ get_eqh(daos_handle_t *eqh);
...
+		      "%016" PRIx64 "-%016" PRIx64 ":", obj_id.hi, obj_id.lo);
+	D_ASSERT(rc == DCACHE_KEY_PREF_SIZE - 1);
 
-	rlink = d_hash_rec_find(mt->dfs_dir_hash, name, len);
-	if (rlink != NULL) {
-		hdl  = hdl_obj(rlink);
-		*obj = hdl->oh;
-		return 0;
-	}
+	rc = d_hash_table_create(D_HASH_FT_MUTEX | D_HASH_FT_LRU, 6, NULL, &dcache_hash_ops,
```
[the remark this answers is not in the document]
knard-intel: - [ ] Size of hash table configurable
- [ ] Change default value of the hash table size

src/client/dfuse/pil4dfs/int_dfs.c
knard-intel: - [ ] Add configuration variable allowing to disable cache

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -285,126 +313,358 @@ get_eqh(daos_handle_t *eqh);
...
-	/* TODO: Remove this after testing. */
-	D_ASSERT(strlen(name) == len);
+	rc = dfs_lookup(mt->dfs, "/", O_RDONLY, &obj, &mode, NULL);
+	if (rc != 0)
+		D_GOTO(out, rc = daos_der2errno(rc));
+	rc = dfs_obj2id(obj, &obj_id);
+	D_ASSERT(rc == 0);
+	rc = snprintf(&mt->dcache.dd_key_root_prefix[0], DCACHE_KEY_PREF_SIZE,
+		      "%016" PRIx64 "-%016" PRIx64 ":", obj_id.hi, obj_id.lo);
+	D_ASSERT(rc == DCACHE_KEY_PREF_SIZE - 1);
```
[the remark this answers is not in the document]
knard-intel: The value are displayed in hex, and thus 16 +1 +16 + 1 + 1 seems correct for me.

knard-intel:
> nice patch! just a few comments from an initial review. Can we move the dcache code itself into a separate file please and add an internal header in this case to avoid keep growing int_dfs.c endlessly?

Put dcache code into a seperate file

src/client/dfuse/pil4dfs/int_dfs.c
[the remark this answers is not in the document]
mchaarawi: i mean the size of the buffer here: https://github.com/daos-stack/daos/pull/13940/files#diff-5bf38ce94f48630a1e3e81661870ed9bce01de6715bd10ab9195c4677263d518R111

that is not hex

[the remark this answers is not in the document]
mchaarawi: sorry my bad.. i misunderstood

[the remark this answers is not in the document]
knard-intel: OK.

src/client/dfuse/pil4dfs/int_dfs.c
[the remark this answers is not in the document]
ashleypittman: You should either use the value in the `dfuse-dentry-dir-time` contaioner attribute or create a new `pil4dfs-dentry-dir-time`

https://github.com/daos-stack/daos/blob/master/docs/user/filesystem.md#caching

This code is going to need to handle multiple containers shortly so the code will need to handle different paths using different values here but that can be added later.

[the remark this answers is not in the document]
ashleypittman: This is exactly the same as the dfuse caching.  It's a trade-off between performance and correctness, and the optimum value will depend on what the application is doing.

There are obviously subtleties and requirement different here so having a different setting is wise but perhaps checking a pil4dfs setting and then backing off to a dfuse setting would be acceptable.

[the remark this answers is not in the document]
mchaarawi: i should have been more clear. I was not talking about functionality. i meant dfuse in this case is not used so it should not use the same value.
Since this is a user space thing, it would be better IMO to allow users to set this with an env variable (with 0 meaning dcache is off/not used) and not having to set an attr on a container. 

my concern with using the attr setting is that users almost never use or set them in practice because they are not that intuitive; or one can be using a default container that is pre created and they might not have permissions to set attrs on the container. We can try and set the default to what dfuse sees if that is provided in the ioctl of the container connect. but adding another attr on the container that users should set before is not a great option IMO.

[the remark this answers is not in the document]
ashleypittman: Users are going to do what users are going to do.  What would be good would be to have the option to set on a per-container basis for power users who will use the flexibility, there's nothing wrong with also having a environment variable as well which as you say is convenient and will cover most users I expect.

Probably the most important thing (other than to have a way to set it in the first place) is sensible default, and I absolutely agree that 60 seconds feels about right.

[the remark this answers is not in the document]
knard-intel: I have removed the timeout of the dcache entry from this PR (not yet pushed).
However, the default timeout value (from 10s to 60s) has been already fixed in the follow-up PR.
https://github.com/daos-stack/daos/blob/84ff78657bda9589886d4c0cdefc5682b357737e/src/client/dfuse/pil4dfs/int_dfs.c#L105

knard-intel:
> > nice patch! just a few comments from an initial review. Can we move the dcache code itself into a separate file please and add an internal header in this case to avoid keep growing int_dfs.c endlessly?
> 
> * [x]  Put dcache code into a seperate file

Fixed wit commit 28ebb6b8293026515227201207a1b43529649505

src/client/dfuse/pil4dfs/int_dfs.c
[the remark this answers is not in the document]
knard-intel: Fixed with commit 61a3f3248

src/client/dfuse/pil4dfs/int_dfs.c
knard-intel: Fixed with commit b530789d44f8fded639bcf685b2e3621bf9b24b0

src/client/dfuse/pil4dfs/int_dfs.c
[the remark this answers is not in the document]
knard-intel: Fixed with commit b530789d44f8fded639bcf685b2e3621bf9b24b0

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1952,7 +1856,7 @@ open_common(int (*real_open)(const char *pathname, int oflags, ...), const char
 		if (!parent && (strncmp(item_name, "/", 2) == 0))
 			rc = dfs_access(dfs_mt->dfs, NULL, NULL, X_OK | W_OK);
 		else
-			rc = dfs_access(dfs_mt->dfs, parent, item_name, X_OK | W_OK);
+			rc = dfs_access(dfs_mt->dfs, drec2obj(parent), item_name, X_OK | W_OK);
```
mchaarawi, line 1859: we are calling drec2obj a lot of times here in this function. it's probably better to just call it once and use a temp variable for the parent dfs_obj_t handle.

knard-intel: - [ ] Add temporary var parent_dfs

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -2055,6 +1960,14 @@ open_common(int (*real_open)(const char *pathname, int oflags, ...), const char
 	}
 	strncpy(file_list[idx_fd]->item_name, item_name, DFS_MAX_NAME);
 
+	file_list[idx_fd]->parent = NULL;
+	if (parent != NULL) {
+		rc = dfs_dup(dfs_mt->dfs, drec2obj(parent), O_RDWR, &file_list[idx_fd]->parent);
```
knard-intel, line 1965: - [ ] Remove dfs_dup

knard-intel: Fixed with commit e283ccd3a

src/client/dfuse/pil4dfs/int_dfs.c
knard-intel: Fixed with commit e283ccd3a

src/tests/ftest/daos_test/dfuse.py
```diff
@@ -100,6 +100,17 @@ def run_test(self, il_lib=None):
                 daos_test_env['D_IL_MAX_EQ'] = '2'
                 daos_test_env['D_IL_ENFORCE_EXEC_ENV'] = '1'
 
+        command = [
```
ashleypittman, line 103: Looks like you have a merge conflict here.  This new alignment is nice and will reduce conflicts in future but you do need to remove L114-115.

knard-intel: > Arghhh, merging failure from my side.
> 
> * [x]  Remove deprecated command

Fixed with commit dd12a3828

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,360 @@
...
+
+        # Run pil4dfs_dcache remotely
+        result = run_remote(self.log, self._host, self.with_exports, timeout=None)
+        if raise_exception and not result.passed:
+            stdout = result.all_stdout[self.host].replace(r'\n', os.linesep)
+            stderr = result.all_stderr[self.host].replace(r'\n', os.linesep)
+            raise CommandFailure(
+                f"Error running pil4dfs_dcache_utest on: {result.failed_hosts}\n"
+                f"-- STDERR START --\n{stderr}\n-- STDERR END --\n"
+                f"-- STDOUT START --\n{stdout}\n-- STDOUT END --\n")
```
daltonbohning, line 85: When the command fails, `run_remote` still logs the output. It's better to not put this output in an exception message because it's very verbose when looking at test failures

src/tests/ftest/dfuse/pil4dfs_dcache.yaml
```diff
@@ -0,0 +1,38 @@
...
+      log_file: daos_server1.log
+      storage:
+        0:
+          class: dcpm
+          scm_list: ["/dev/pmem1"]
+          scm_mount: /mnt/daos1
+
+pool:
+  size: 90%
+  control_method: dmg
```
daltonbohning, line 34: `control_method: dmg` isn't used anymore
```suggestion
```

src/tests/ftest/dfuse/pil4dfs_dcache.yaml
```diff
@@ -0,0 +1,38 @@
...
+    1:
+      pinned_numa_node: 1
+      fabric_iface: ib1
+      fabric_iface_port: 31417
+      log_file: daos_server1.log
+      storage:
+        0:
+          class: dcpm
+          scm_list: ["/dev/pmem1"]
+          scm_mount: /mnt/daos1
```
daltonbohning, line 30: Could we use `storage: auto` here?

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,360 @@
...
+import re
+
+from command_utils import ExecutableCommand
+from command_utils_base import BasicParameter
+from dfuse_test_base import DfuseTestBase
+from exception_utils import CommandFailure
+from run_utils import run_remote
+
+
+class Pil4dfsDcacheCmd(ExecutableCommand):
```
daltonbohning, line 17: Will you please move this class definition to `dfuse_utils.py`? We try to define all command classes in a utility for re-use later.

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,360 @@
...
+    """Defines an object representing a pil4dfs_dcache unit test command."""
+
+    def __init__(self, host, path, mnt, timeout=None):
+        """Create a Pil4dfsDcacheUtestCmd object.
```
daltonbohning, line 21:
```suggestion
        """Create a Pil4dfsDcacheCmd object.
```

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,360 @@
...
+        lib_dir = os.path.join(path, 'lib64', 'libpil4dfs.so')
+        self.env['LD_PRELOAD'] = lib_dir
+        self.env['D_DFUSE_MNT'] = mnt
+        self.env['D_LOG_MASK'] = 'DEBUG'
+        self.env['DD_SUBSYS'] = 'il'
+        self.env['DD_MASK'] = 'trace'
+        self.env['D_IL_REPORT'] = '1'
+
+        if timeout is not None:
+            self.env["D_IL_DCACHE_REC_TIMEOUT"] = timeout
```
daltonbohning, line 44: It's better to not hardcode these envs in the class itself. Especially after moved to `dfuse_utils.py`. The test should do
```
cmd = Pil4dfsDcacheCmd(...)
cmd.env.update({...})
```

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,360 @@
...
+
+        self.test_id = BasicParameter(None)
+
+    @property
+    def host(self):
+        """Get the host on which to remotely run the command via run().
```
daltonbohning, line 50: Should this only every be one host? If so, maybe `__init__` should raise a `ValueError` if given more than one host

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,360 @@
...
+                if Pil4dfsDcache.__start_test_re__.match(line):
+                    state = 'check_dcache'
+
+            elif state == 'check_dcache':
+                for dcache_name, dcache_re in Pil4dfsDcache.__dcache_re__.items():
+                    if dcache_re.match(line):
+                        dcache_count[dcache_name] += 1
+                if Pil4dfsDcache.__end_test_re__.match(line):
+                    for key in Pil4dfsDcache.__dcache_re__:
+                        self.assertEqual(test_case[key], dcache_count[key])
```
daltonbohning, line 297: Please add a descriptive message (third argument) for asserts so failures are clear

daltonbohning (commented):
If these ftest changes are the only remaining blockers, we can address separately.

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: - [ ] Remove verbose error log message

src/tests/ftest/dfuse/pil4dfs_dcache.yaml
knard-intel: - [ ] Remove useless control_method

src/tests/ftest/dfuse/pil4dfs_dcache.yaml
knard-intel: - [ ] use auto storage

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: - [ ] Move class Pil4dfsDcacheCmd to dfuse_utils.py

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: - [ ] Fix comment

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: - [ ] Makes environment configurable

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: - [ ] Add assertion error message

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: Check that host contain only one host

knard-intel: - [ ] Check client nodeset size

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: Fixed with commit eace57b3b479abd420dec07bf85cc46b4a1c9c31

src/tests/ftest/dfuse/pil4dfs_dcache.yaml
knard-intel: Fixed with commit eace57b3b479abd420dec07bf85cc46b4a1c9c31

src/tests/ftest/dfuse/pil4dfs_dcache.yaml
knard-intel: Fixed with commit eace57b3b479abd420dec07bf85cc46b4a1c9c31

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: Fixed with commit eace57b3b479abd420dec07bf85cc46b4a1c9c31

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: Fixed with commit eace57b3b479abd420dec07bf85cc46b4a1c9c31

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: Fixed with commit eace57b3b479abd420dec07bf85cc46b4a1c9c31

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: Fixed with commit eace57b3b479abd420dec07bf85cc46b4a1c9c31

knard-intel:
The failure happening are known issues which are not related with this PR:
* [daos_test/suite.py:DaosCoreTest.test_daos_degraded_mode](https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-13940/36/testReport/FTEST_daos_test/DaosCoreTest/Test_Hardware___Functional_Hardware_Medium_Verbs_Provider___01___daos_test_suite_py_DaosCoreTest_test_daos_degraded_mode_run_agent_config_transport_config_daos_tests_args_daos_test_num_clients_pools_created_scalable_endpoint_stopped_ranks_test_name_dmg_hosts_pool_server_config_engines_0_storage_0_1_timeouts_b912/): ticket [DAOS-15598](https://daosio.atlassian.net/browse/DAOS-15598)
* [daos_test/suite.py:DaosCoreTest.test_daos_rebuild_simple](https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-13940/36/testReport/FTEST_daos_test.DaosCoreTest/DAOS_Rebuild_Simple/Test_Hardware___Functional_Hardware_Medium_Verbs_Provider___18___daos_test_suite_py_DaosCoreTest_test_daos_rebuild_simple_run_agent_config_transport_config_daos_tests_args_daos_test_num_clients_pools_created_scalable_endpoint_stopped_ranks_test_name_dmg_hosts_pool_server_config_engines_0_storage_0_1_timeouts_b912/): ticket [DAOS-15124](https://daosio.atlassian.net/browse/DAOS-15124)

knard-intel:
> Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE. https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-13940/36/execution/node/1541/log

The failure happening are known issues which are not related with this PR:
* [daos_test/suite.py:DaosCoreTest.test_daos_degraded_mode](https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-13940/36/testReport/FTEST_daos_test/DaosCoreTest/Test_Hardware___Functional_Hardware_Medium_Verbs_Provider___01___daos_test_suite_py_DaosCoreTest_test_daos_degraded_mode_run_agent_config_transport_config_daos_tests_args_daos_test_num_clients_pools_created_scalable_endpoint_stopped_ranks_test_name_dmg_hosts_pool_server_config_engines_0_storage_0_1_timeouts_b912/): ticket [DAOS-15598](https://daosio.atlassian.net/browse/DAOS-15598)
* [daos_test/suite.py:DaosCoreTest.test_daos_rebuild_simple](https://build.hpdd.intel.com/job/daos-stack/job/daos/view/change-requests/job/PR-13940/36/testReport/FTEST_daos_test.DaosCoreTest/DAOS_Rebuild_Simple/Test_Hardware___Functional_Hardware_Medium_Verbs_Provider___18___daos_test_suite_py_DaosCoreTest_test_daos_rebuild_simple_run_agent_config_transport_config_daos_tests_args_daos_test_num_clients_pools_created_scalable_endpoint_stopped_ranks_test_name_dmg_hosts_pool_server_config_engines_0_storage_0_1_timeouts_b912/): ticket [DAOS-15124](https://daosio.atlassian.net/browse/DAOS-15124)

mchaarawi (approved):
the pil4dfs core changes look good to me.

src/tests/ftest/dfuse/utest/pil4dfs_dcache.c
```diff
@@ -0,0 +1,355 @@
+/*
+ * (C) Copyright 2024 Intel Corporation.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause-Patent
+ */
+/**
+ * This tests a threaded client sending RPCs on a single context
```
wiliamhuang, line 7: Need to revise this comment.

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+
+/** Entry of a DFS directory cache */
+struct dcache_rec {
+	/** Entry in the hash table of the DFS cache */
+	d_list_t         dr_entry;
+	/** Cached DFS directory */
+	dfs_obj_t       *dr_obj;
+	/** Reference counter used to manage memory deallocation */
+	_Atomic uint32_t dr_ref;
+	/** True iff this entry was deleted */
```
wiliamhuang, line 65: Typo "iff"

src/tests/ftest/dfuse/utest/pil4dfs_dcache.c
knard-intel: - [ ] Update invalid comment

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: Not a typo, it is an acronym of [if and only if](https://www.dictionary.com/browse/iff) ;)

wiliamhuang: Good to know. Thank you! :)

mchaarawi (dismissed):

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+	if (rc != 0)
+		D_GOTO(error, rc = d_errno2der(rc));
+	if (!S_ISDIR(mode))
+		D_GOTO(error, rc = -DER_NOTDIR);
+	rec_tmp->dr_obj = obj;
+
+	rc = dfs_obj2id(obj, &obj_id);
+	D_ASSERT(rc == 0);
+	rc = snprintf(&rec_tmp->dr_key_child_prefix[0], DCACHE_KEY_PREF_SIZE,
+		      "%016" PRIx64 "-%016" PRIx64 ":", obj_id.hi, obj_id.lo);
```
wiliamhuang, line 282: Text version of obj_id is used here. Shall we use the original 16 bytes obj_id instead to make the key shorter?

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+	int           rc;
+
+	D_ALLOC(rec_tmp, sizeof(*rec_tmp) + key_len + 1);
+	if (rec_tmp == NULL)
+		D_GOTO(error, rc = -DER_NOMEM);
+
+	atomic_init(&rec_tmp->dr_ref, 1);
+	atomic_init(&rec_tmp->dr_deleted, false);
+
+	rc = dfs_lookup(dcache->dd_dfs, path, O_RDWR, &obj, &mode, NULL);
```
wiliamhuang, line 272: dfs_lookup() always starts from the root. We should consider dfs_lookup_rel() here to avoid unnecessary work.

ashleypittman (changes requested):
I haven't fully understood this PR yet but I would like to know why it's calling `dfs_lookup` rather than `dfs_lookup_rel`.

src/client/dfuse/pil4dfs/dfs_dcache.c
mchaarawi: indeed this looks like it's doing unnecessary work as in the loop where this is being used, we can keep track of each entry that is looked up and then passed to the next entry as the parent.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1220,20 +1114,23 @@ query_path(const char *szInput, int *is_target_path, dfs_obj_t **parent, char *i
...
 			if (pos == (*dfs_mt)->len_fs_root) {
 				(*parent_dir)[0] = '/';
-				path_len         = 1;
+				parent_dir_len   = 1;
 			} else {
+				char *parent_path;
+
+				full_path_parse[pos] = '\0';
+				parent_path          = full_path_parse + (*dfs_mt)->len_fs_root;
+				/* path_dir_len is length of the string, without termination */
```
wiliamhuang, line 1123: "path_dir_len" in comment should be updated as "parent_dir_len".

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: From my side, I would prefer to keep the textual version: makes things easier for maintenance and debugging.  I am also not very found of mixing binary values (i.e. obj_id) and textual values (file name).

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: - [ ] use dfs_lookup_rel() function

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+dcache_rec_free(struct d_hash_table *htable, d_list_t *rlink)
+{
+	int           rc;
+	dcache_rec_t *rec;
+
+	rec = dlist2drec(rlink);
+	D_DEBUG(DB_TRACE, "delete record '%s' (ref=%u)", rec->dr_key, atomic_load(&rec->dr_ref));
+	rc = dfs_release(rec->dr_obj);
+	if (rc == ENOMEM)
+		rc = dfs_release(rec->dr_obj);
```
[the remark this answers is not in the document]
knard-intel: I took this code from the original dir-entry cache code, which was looking strange for me.
Thus, I am not sure from where it was taken, maybe @wiliamhuang could have more info on this point.

- [ ] Remove retry dfs_release()

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+}
+
+static void
+dcache_rec_free(struct d_hash_table *htable, d_list_t *rlink)
+{
+	int           rc;
+	dcache_rec_t *rec;
+
+	rec = dlist2drec(rlink);
+	D_DEBUG(DB_TRACE, "delete record '%s' (ref=%u)", rec->dr_key, atomic_load(&rec->dr_ref));
```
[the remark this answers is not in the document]
knard-intel: - [ ] Use dedicated path logging macro

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+	dcache_tmp->drec_del_at_fn = drec_del_at_act;
+	dcache_tmp->drec_del_fn    = drec_del_act;
+
+	*dcache = dcache_tmp;
+	D_GOTO(out, rc = -DER_SUCCESS);
+
+error:
+	D_FREE(dcache_tmp);
+out:
+	dfs_release(obj);
```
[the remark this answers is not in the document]
knard-intel: It was done on purpose as `dfs_release(NULL)` is well supported. The goal is to have one return point from the function: I do not like not mix goto and several return statements.  However this is a personal coding practice, and thus I have no problem to change the GOTO from line 178 to a return statement.

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: To be honest, I did not know the existence of the `dfs_lookup_rel()` function: good things to know :)

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+
+	rlink = d_hash_rec_find(&dcache->dd_dir_hash, key, key_len);
+	if (rlink == NULL)
+		return NULL;
+	return dlist2drec(rlink);
+}
+
+static inline int
+dcache_add(dcache_rec_t **rec, dfs_dcache_t *dcache, const char *path, const char *key,
+	   size_t key_len)
```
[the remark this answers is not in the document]
knard-intel: From what I know, in C or at least in C++, lvalue are normally put at the beginning of the list of function parameters.  However, I have no problem to change it, if it is DAOS best practice coding.

- [ ] Put lvalue rec parameter at the end of the list of parameters

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+dcache_destroy_dact(dfs_dcache_t *dcache);
+static int
+dcache_find_insert_dact(dcache_rec_t **rec, dfs_dcache_t *dcache, char *path, size_t path_len);
+static void
+drec_del_at_dact(dfs_dcache_t *dcache, dcache_rec_t *rec);
+
+static int
+dcache_create_dact(dfs_dcache_t **dcache, dfs_t *dfs, uint32_t bits, uint32_t rec_timeout)
+{
+	dfs_dcache_t *dcache_tmp = NULL;
```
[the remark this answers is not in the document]
knard-intel: - [ ] Remove useless initialization

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+static void
+drec_del_at_dact(dfs_dcache_t *dcache, dcache_rec_t *rec)
+{
+	int rc;
+
+	D_DEBUG(DB_TRACE, "delete record %p", rec);
+
+	rc = dfs_release(rec->dr_obj);
+	if (rc == ENOMEM)
+		rc = dfs_release(rec->dr_obj);
```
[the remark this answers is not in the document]
knard-intel: Remove useless double deallocation

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+{
+	D_ASSERT(dcache != NULL);
+	D_ASSERT(dfs != NULL);
+
+	if (rec_timeout == 0) {
+		return dcache_create_dact(dcache, dfs, bits, rec_timeout);
+	}
+
+	return dcache_create_act(dcache, dfs, bits, rec_timeout);
+}
```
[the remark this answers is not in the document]
knard-intel: xxx_dact() function are used when dir-cache is deactivated.
xxx_act() function are used when dir-cache is activated

src/client/dfuse/pil4dfs/dfs_dcache.c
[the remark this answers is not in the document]
wiliamhuang: Indeed, I previously found a few places in other code retrying dfs_release() to avoid fault injection testing issues more than a half year ago. I copied it into libpil4dfs.
I just went through all code with dfs_release(). It looks like all of them are fixed now except in int_dfs.c.

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,616 @@
...
+
+int
+dcache_create(dfs_dcache_t **dcache, dfs_t *dfs, uint32_t bits, uint32_t rec_timeout)
+{
+	D_ASSERT(dcache != NULL);
+	D_ASSERT(dfs != NULL);
+
+	if (rec_timeout == 0) {
+		return dcache_create_dact(dcache, dfs, bits, rec_timeout);
```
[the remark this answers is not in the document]
knard-intel: Indeed, the bits and rec_timeout are useless.

- [ ] Remove useless bits and rec_timeout parameters

knard-intel:
> I haven't fully understood this PR yet but I would like to know why it's calling `dfs_lookup` rather than `dfs_lookup_rel`.

As a DFS young padawan, I did not know the existence of the `dfs_lookup_rel` function => I will change this for sure.

src/client/dfuse/pil4dfs/dfs_dcache.c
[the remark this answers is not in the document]
ashleypittman: Normally the unwinding at the end of the function is done in the opposite order to allocations etc in the main function so that goto can be used to jump to the appropriate place and only unwind the things it needs to.

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,329 @@
...
+from dfuse_utils import Pil4dfsDcacheCmd
+
+
+class Pil4dfsDcache(DfuseTestBase):
+    """Test class Description: Runs set of unit test on pil4dfs directory cache.
+
+    :avocado: recursive
+    """
+
+    __tests_suite__ = {
```
ashleypittman, line 21: Unless I'm mistaken in python a `_` prefix means something is "private", a `__` prefix means something is actually private and `__` prefix with `__` suffix is reserved for python internals.  Unless this is overriding some avocado default then I'd expect this to have a single `_` prefix.

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    __dcache_re__ = {
+        "dcache_add": re.compile(r'^.+ il +DBUG .+ dcache_add\(\) .+$'),
+        "dcache_del": re.compile(r'^.+ il +DBUG .+ dcache_rec_free\(\) .+$'),
+        "dcache_hit": re.compile(r'^.+ il +DBUG .+ dcache hit:.+$'),
+        "dcache_miss": re.compile(r'^.+ il +DBUG .+ dcache miss:.+$'),
+        "no_dcache_new": re.compile(r'^.+ il +DBUG .+ dcache_find_insert_dact\(\).+$'),
+        "no_dcache_del": re.compile(r'^.+ il +DBUG .+ drec_del_at_dact\(\) .+$')
+    }
```
ashleypittman, line 175: Same comment as above.

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,329 @@
...
+
+    def setUp(self):
+        """Set up each test case."""
+        # obtain separate logs
+        self.update_log_file_names()
+
+        # Start the servers and agents
+        super().setUp()
+
+    def mount_dfuse(self):
```
ashleypittman, line 188:
```suggestion
    def _mount_dfuse(self):
```

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,329 @@
...
+        self.log.info("Creating DAOS pool")
+        self.add_pool()
+
+        self.log.info("Creating DAOS container")
+        self.add_container(self.pool)
+
+        self.log.info("Mounting DFuse mount point")
+        self.start_dfuse(self.hostlist_clients, self.pool, self.container)
+
+    def update_cmd_env(self, env, mnt, timeout=None):
```
ashleypittman, line 199:
```suggestion
    def _update_cmd_env(self, env, mnt, timeout=None):
```

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,329 @@
...
+        env['D_DFUSE_MNT'] = mnt
+        env['D_LOG_MASK'] = 'DEBUG'
+        env['DD_SUBSYS'] = 'il'
+        env['DD_MASK'] = 'trace'
+        env['D_IL_REPORT'] = '1'
+
+        if timeout is not None:
+            env["D_IL_DCACHE_REC_TIMEOUT"] = timeout
+
+    def check_result(self, test_case, lines):
```
ashleypittman, line 219:
```suggestion
    def _check_result(self, test_case, lines):
```

src/tests/ftest/dfuse/pil4dfs_dcache.py
```diff
@@ -0,0 +1,329 @@
...
+        """Check consistency of the test output.
+
+        Args:
+            test_case (dict): dictionary of the expected output result
+            lines (list): list of the test command stdout lines
+        """
+        dcache_count = {key: 0 for key in Pil4dfsDcache.__dcache_re__}
```
ashleypittman, line 226: Perhaps I'm mis-reading this but can't this (and lots of other instances) be accessed via self rather than the class name?
```suggestion
        dcache_count = {key: 0 for key in self.__dcache_re__}
```

gnailzenh:
sorry there is a conflict, please update the patch

knard-intel:
> sorry there is a conflict, please update the patch

I will do it asap.  Moreover, there is some part of the code which must be fixed: use of dfs_lookup_rel() instead of dfs_lookup()

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: I was thinking that python naming convention was used to respectively identify protected and private member with `_` and `__`.  However, it seems that I was wrong https://realpython.com/python-double-underscore/#public-interfaces-and-naming-conventions-in-python

Thus, I will use a `_` as you suggested.

- [ ] Use _ instead of __

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: - [ ]  Use _ instead of __

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: - [ ]  refactor mount_dfuse() to _mount_dfuse()

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: - [ ]  refactor update_cmd_env() to _update_cmd_env()

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: refactor check_result() to _check_result()

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: Nope, it can not be accessed through self, as they are class attributes: they are common to all class instances.

src/client/dfuse/pil4dfs/dfs_dcache.c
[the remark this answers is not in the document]
knard-intel: > I took this code from the original dir-entry cache code, which was looking strange for me. Thus, I am not sure from where it was taken, maybe @wiliamhuang could have more info on this point.
> 
> * [x]  Remove retry dfs_release()

Fixed with commit e6dd70d

src/client/dfuse/pil4dfs/dfs_dcache.c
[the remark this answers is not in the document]
knard-intel: > * [x]  Use dedicated path logging macro

Fixed with commit 4c54fb2

src/client/dfuse/pil4dfs/dfs_dcache.c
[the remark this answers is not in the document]
knard-intel: > * [x]  Remove useless initialization

Fixed with commit 53b6af17e

src/client/dfuse/pil4dfs/dfs_dcache.c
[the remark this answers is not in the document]
knard-intel: > * [x]  Remove useless double deallocation

Fixed with commit e6dd70d

src/client/dfuse/pil4dfs/dfs_dcache.c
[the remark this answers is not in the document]
knard-intel: > Indeed, the bits and rec_timeout are useless.
> 
> * [x]  Remove useless bits and rec_timeout parameters

Fixed with commit 9acaccad0

src/client/dfuse/pil4dfs/int_dfs.c
knard-intel: > * [ ]  Fix comment error

Fixed with commit 9eb019038

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: > I was thinking that python naming convention was used to respectively identify protected and private member with `_` and `__`. However, it seems that I was wrong https://realpython.com/python-double-underscore/#public-interfaces-and-naming-conventions-in-python
> 
> Thus, I will use a `_` as you suggested.
> 
> * [x]  Use _ instead of __

Fixed with commit 38b3dbc77

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: Fixed with commit 454a7371d

knard-intel: > * [x]   Use _ instead of __

Fixed with commit 454a7371d

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: > * [x]   refactor update_cmd_env() to _update_cmd_env()

Fixed with commit 454a7371d

src/tests/ftest/dfuse/pil4dfs_dcache.py
knard-intel: > * [x]  refactor check_result() to _check_result()

Fixed with commit 454a7371d

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: > * [x]  use dfs_lookup_rel() function

Fixed with commit 3a6baaa10

knard-intel:
> > I haven't fully understood this PR yet but I would like to know why it's calling `dfs_lookup` rather than `dfs_lookup_rel`.
> 
> As a DFS young padawan, I did not know the existence of the `dfs_lookup_rel` function => I will change this for sure.

Fixed with commit 3a6baaa10

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,663 @@
...
+#include "dfs_dcache.h"
+
+/** Size of the hash key prefix */
+#define DCACHE_KEY_PREF_SIZE 35
+#define DCACHE_KEY_MAX       (DCACHE_KEY_PREF_SIZE - 1 + PATH_MAX)
+
+#ifdef DAOS_BUILD_RELEASE
+
+#define DF_DK          "dk[%zi]"
+#define DP_DK(_dk)     strnlen((_dk), DCACHE_KEY_MAX)
```
ashleypittman, line 32: Is there a difference between `DK` and the `DE` macros that already exist?  `DE` was supposed to mean `directory entry` so is what I'd expect the cache to use.

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,663 @@
...
+
+#define DF_PATH        "path[%zi]"
+#define DP_PATH(_path) strnlen((_path), PATH_MAX)
```
ashleypittman, line 35: There isn't a `DF_PATH` macro, these could go in the public header file at some point although it's no harm to have them here for now.

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,663 @@
...
+
+#else
+
+#define DF_DK          "dk'%s'"
+#define DP_DK(_dk)     (_dk)
```
ashleypittman, line 40: It's worth noting that the `DF_DE` macro will mask out ` from the logs so they can be parsed by tools correctly.

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,663 @@
...
+	/** Delete a given dir-cache record */
+	drec_del_at_fn_t    drec_del_at_fn;
+	/** Delete the dir-cache record of a given location */
+	drec_del_fn_t       drec_del_fn;
+};
+
+/** Entry of a DFS directory cache */
+struct dcache_rec {
+	/** Entry in the hash table of the DFS cache */
+	d_list_t         dr_entry;
```
ashleypittman, line 79: Is there a requirement for this entry to be first in the struct?  If so it should be commented as such.

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,663 @@
...
+		D_GOTO(error, rc = -DER_NOMEM);
+
+	rec->dr_obj = obj;
+	atomic_init(&rec->dr_ref, 0);
+	atomic_init(&rec->dr_deleted, false);
+	memcpy(&rec->dr_key_child_prefix[0], &dcache->dd_key_root_prefix[0], DCACHE_KEY_PREF_SIZE);
+	memcpy(&rec->dr_key[0], &dcache->dd_key_root_prefix[0], DCACHE_KEY_PREF_SIZE);
+	rec->dr_key_len = DCACHE_KEY_PREF_SIZE - 1;
+	rc = d_hash_rec_insert(&dcache->dd_dir_hash, rec->dr_key, rec->dr_key_len, &rec->dr_entry,
+			       true);
```
ashleypittman, line 202: Does `exclusive` need to be true here?  If the hash table is empty then presumably it could be false although for the first entry it won't make much difference.

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,663 @@
...
+	D_ASSERT(key_len > 0);
+
+	rlink = d_hash_rec_find(&dcache->dd_dir_hash, key, key_len);
+	if (rlink == NULL)
+		return NULL;
+	return dlist2drec(rlink);
+}
+
+static inline int
+dcache_add(dfs_dcache_t *dcache, dcache_rec_t *parent, const char *name, const char *key,
```
ashleypittman, line 304: How does the key differ from the name here?  Shouldn't we by using the `directory entry` as the key, otherwise how can we search for it?

src/client/dfuse/pil4dfs/dfs_dcache.c
```diff
@@ -0,0 +1,663 @@
...
+	mode_t        mode;
+	int           rc;
+
+	(void)path_len; /* unused */
+
+	D_ALLOC_PTR(rec_tmp);
+	if (rec_tmp == NULL)
+		D_GOTO(error, rc = -DER_NOMEM);
+
+	rc = dfs_lookup(dcache->dd_dfs, path, O_RDWR, &obj, &mode, NULL);
```
ashleypittman, line 558: I assume this is the non-caching case here and it's delegating the path walk to dfs?  This is OK for now but in future then we'll want to do this in pil4dfs in future.

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: I was not sure if it was valuable enough to put in a common header.
I will let you decide, if it is worth of it or not.

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: Just a problem of size in release mode with strnlen

src/client/dfuse/pil4dfs/dfs_dcache.c
wiliamhuang: We used "true" for "exclusive" originally in pil4dfs code.
Yes. we could use "false" here since the record should be unique.

src/client/dfuse/pil4dfs/dfs_dcache.c
ashleypittman: Ok, now I understand.  Thank you.

src/client/dfuse/pil4dfs/dfs_dcache.c
ashleypittman: We will need this in other places so yes, at some point we will but obviously outside the scope of this work.

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: At this point, the hash table shall be empty.  If this record is already present, then we have a consistancy issue as this function should only be called at the creation of the dcache

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: There is no requirements from what I know.
No call to containerof or such struct pointer trick until now.

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: The key is constructed in the following way:
<low bit dirobj_id>-<high bit dirobj_id>:<dir name>

ashleypittman: We spoke about this off-line, the key is already formatted for the previous (failed) lookup so the key is passed in here in addition to the name as it's already been generated, but at this point yes, the key is the oid+name as you say.  I was confused by the string seemingly being passed in twice as two different parameters.

ashleypittman (approved):
I think all my other comments were with questions or cosmetic so can be resolved later.

src/client/dfuse/pil4dfs/dfs_dcache.c
ashleypittman: https://github.com/daos-stack/daos/blob/master/docs/user/filesystem.md#unified-namespace-uns

dfs supports one container but all our other middleware layered on top if it will detect UNS links on directories and open a new container as necessary.  To do that the path walk needs to happen at the higher level, check the xattr on loopup and open new pools/containers as required.  Out of scope for here but this will likely be the next development work on pil4dfs.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -5525,6 +5229,7 @@ utimens_timespec(const char *path, const struct timespec times[2], int flags)
 		times_us[0].tv_usec = times[0].tv_nsec / 100;
 		times_us[1].tv_sec  = times[1].tv_sec;
 		times_us[1].tv_usec = times[1].tv_nsec / 100;
+		drec_decref(dfs_mt->dcache, parent);
```
wiliamhuang, line 5232: This line should be removed since "is_target_path" is false here.

knard-intel: Agree, that this statement is useless but should be harmless.
Indeed `parent` will be `NULL` and thus `drec_decref()` will do nothing (as for the `free()` function with NULL pointer).
Thus, if it is OK for you, I will fix this in the follow-up PR.
Is it OK for you.

mchaarawi: yes let's fix that in the GC follow on PR please

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: OK thanks for the explanation.

knard-intel: I will add a comment about this in the follow-up PR

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Sounds good to me.

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: I will move these macros in the common header in the follow-up PR.

wiliamhuang (approved):

ashleypittman:
@knard-intel can you update the PR description with an appropriate commit message please?

knard-intel:
Hello @daos-stack/daos-gatekeeper ,

Could this PR be landed as the only CI failure of the last build is a known issue:
- Build failure [daos_test/suite.py:DaosCoreTest.test_daos_container CONT35: container duplicate op detection timing failure](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-13940/49/testReport/FTEST_daos_test/DaosCoreTest-DAOS_Container/Test_Hardware___Functional_Hardware_Medium_Verbs_Provider___CONT35__container_duplicate_op_detection_timing/) Jira ticket [DAOS-15662](https://daosio.atlassian.net/issues/DAOS-15662)

Please could you use the following commit message:
Title: `DAOS-14348 client: fix pil4dfs's dentry cache`
Body:
```
Fix path walk of the pil4dfs's dentry cache.
Fix pil4dfs rename() function.
Add enable/disable feature of the pil4dfs's dentry cache.
Add new functional test of the pil4dfs's dentry cache,

Signed-off-by: Cedric Koch-Hofer <cedric.koch-hofer@intel.com>
Co-authored-by: Lei Huang <lei.huang@intel.com>
```

knard-intel:
> @knard-intel can you update the PR description with an appropriate commit message please?

Done https://github.com/daos-stack/daos/pull/13940#issuecomment-2074199645

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: Fixed in the GC follow-up PR with commit 23ec8eac9a17a2ddc7872bfcbcdaee1df043ac3e

src/client/dfuse/pil4dfs/int_dfs.c
knard-intel: Fixed in the GC follow-up PR with commit ece16c1146b4eeba9f6ec2769b558c2abcdda989

mchaarawi (approved):

src/client/dfuse/pil4dfs/dfs_dcache.c
knard-intel: Sounds logical, I will do it.

knard-intel: Fixed with commit e0bc79fd6e6b926dd2efa405bdd66bb9e3762766

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -4190,18 +4135,29 @@ rmdir(const char *path)
...
 	if (rc)
 		D_GOTO(out_err, rc);
 
+	if (parent != NULL) {
+		rc = drec_del(dfs_mt->dcache, full_path, parent);
+		if (rc != -DER_SUCCESS && rc != -DER_NONEXIST)
+			DL_ERROR(rc, "DAOS directory cache cleanup failed");
+	}
+
+	drec_decref(dfs_mt->dcache, parent);
```
wiliamhuang, line 4148: Should this line, "drec_decref(dfs_mt->dcache, parent);", be removed? "drec_del(dfs_mt->dcache, full_path, parent);" was executed already.

knard-intel: The `drec_del()` will remove the dentry of the path given in parameter of the function, if this last one exist.
The `drec_decref()` will decrement the dentry counter of the parent of the path given as parameter of the function.

wiliamhuang:
@knard-intel One question: Why "drec_decref(dfs_mt->dcache, parent);" is called for opening a directory with open/opendir?

wiliamhuang:
> @knard-intel One question: Why "drec_decref(dfs_mt->dcache, parent);" is called for opening a directory with open/opendir?

Never mind. I understand now. Sorry for the noise.

knard-intel:
> > @knard-intel One question: Why "drec_decref(dfs_mt->dcache, parent);" is called for opening a directory with open/opendir?
> 
> Never mind. I understand now. Sorry for the noise.

No problem, thanks for your time with this PR.

PR #13962, opened 2024-03-08
DAOS-15388 tools: daos fs copy: error when src is dst

daltonbohning:
Features: daos_fs_copy

Check if the source and destination are the same.
Check if the source is a subset of the destination.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/utils/daos_hdlr.c
+++ b/src/utils/daos_hdlr.c
@@ -925,9 +925,8 @@ fs_copy(struct cmd_args_s *ap, struct dm_args *ca, struct file_dfs *src_file_dfs
 				src_is_subpath = (strncmp(src_path, dst_path, src_path_len) == 0);
 			} else {
 				src_is_subpath =
-					(strncmp(src_path, dst_path, src_path_len) == 0 &&
-					 dst_path_len > src_path_len &&
-					 dst_path[src_path_len] == '/');
+				    (strncmp(src_path, dst_path, src_path_len) == 0 &&
+				     dst_path_len > src_path_len && dst_path[src_path_len] == '/');
 			}
 
 			if (src_is_subpath) {
--- a/src/utils/daos_hdlr.c
+++ b/src/utils/daos_hdlr.c
@@ -924,10 +924,10 @@ fs_copy(struct cmd_args_s *ap, struct dm_args *ca, struct file_dfs *src_file_dfs
 			if (src_path[src_path_len - 1] == '/') {
 				src_is_subpath = (strncmp(src_path, dst_path, src_path_len) == 0);
 			} else {
-				src_is_subpath = (
-					strncmp(src_path, dst_path, src_path_len) == 0 &&
-					dst_path_len > src_path_len &&
-					dst_path[src_path_len] == '/');
+				src_is_subpath =
+					(strncmp(src_path, dst_path, src_path_len) == 0 &&
+					 dst_path_len > src_path_len &&
+					 dst_path[src_path_len] == '/');
 			}
 
 			if (src_is_subpath) {
--- a/src/utils/daos_hdlr.c
+++ b/src/utils/daos_hdlr.c
@@ -901,7 +901,7 @@ fs_copy(struct cmd_args_s *ap, struct dm_args *ca, struct file_dfs *src_file_dfs
 	}
 
 	/* Make sure the source and destination are not the same.
-	 * Make sure the source is not a subset of the destination. */
+	 * Make sure the source is not a subpath of the destination. */
 	if ((src_file_dfs->type == POSIX && dst_file_dfs->type == POSIX) ||
 	    (src_file_dfs->type == DAOS && dst_file_dfs->type == DAOS &&
 	     strcmp(ca->src_pool, ca->dst_pool) == 0 && strcmp(ca->src_cont, ca->dst_cont) == 0)) {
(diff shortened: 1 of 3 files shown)


daltonbohning:
Failures in https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-13962/10/tests
are
https://daosio.atlassian.net/browse/DAOS-15431

src/utils/daos_hdlr.c
```diff
@@ -899,6 +900,41 @@ fs_copy(struct cmd_args_s *ap, struct file_dfs *src_file_dfs, struct file_dfs *d
...
+			bool src_is_subpath;
+
+			/* Always compare src_path with a trailing slash */
+			src_path_len = strlen(src_path);
+			if (src_path[src_path_len - 1] == '/') {
+				src_is_subpath = (strncmp(src_path, dst_path, src_path_len) == 0);
+			} else {
+				char *tmp_src = NULL;
+
+				D_ASPRINTF(tmp_src, "%s/", src_path);
```
mchaarawi, line 924: should check tmp_src != NULL, but do you really need an extra allocation and a tmp string here?
why not just compare:
(strncmp(src_path, dst_path, src_path_len) == 0 && dst_path[src_path_len] == "/") ?


im not sure though if this is correct though, because consider
what if src == dst == /d1/d2
I see that the same as
src == /d1/d2 and dst == /d1/d2/
but the result is different in this case, no?

src/utils/daos_hdlr.c
```diff
@@ -899,6 +900,41 @@ fs_copy(struct cmd_args_s *ap, struct file_dfs *src_file_dfs, struct file_dfs *d
...
+		}
+		if (S_ISDIR(src_stat.st_mode)) {
+			int  src_path_len;
+			bool src_is_subpath;
+
+			/* Always compare src_path with a trailing slash */
+			src_path_len = strlen(src_path);
+			if (src_path[src_path_len - 1] == '/') {
+				src_is_subpath = (strncmp(src_path, dst_path, src_path_len) == 0);
+			} else {
```
wiliamhuang, line 921: What if src_path ends with '/' but dst_path does not end with '/'?
Example,
src_path = "/dir0/dir1/"
dst_path = "/dir0/dir1"
src_is_subpath is set false.

[leihuan1@boro-46 daos]$ cp -R /dev/shm/dir/ /dev/shm/dir
cp: cannot copy a directory, '/dev/shm/dir/', into itself, '/dev/shm/dir/dir'

src/utils/daos_hdlr.c
ashleypittman: Alternatively you could stat the src and dest and if the `st_dev` and `st_ino` fields are the same then they two are the same.  This has the advantage that it'll work if you access the same tree via different paths, full vs relative or in the presence of symlinks etc.

src/utils/daos_hdlr.c
daltonbohning: That behaves as expected
```
$ daos fs copy --src dir0/dir1/ --dst dir0/dir1
ERROR: cannot copy a directory into itself: Invalid parameters (-1003)
```
If the source doesn't end with `/`, we append it before comparing.
And because we always generate `dst_path` like this:
https://github.com/daos-stack/daos/blob/d4cc14fd97a7cc522669f7899622920202015ae1/src/utils/daos_hdlr.c#L894-L900
it will either contain a slash at the some location as the trailing slash on the source, or it will not be a subpath of the source

wiliamhuang: I see. Thank you!

src/utils/daos_hdlr.c
daltonbohning: > why not just compare:
(strncmp(src_path, dst_path, src_path_len) == 0 && dst_path[src_path_len] == "/") ?

You're right. I tested locally and that works the same (with minor tweaks).

Per my other comment to @wiliamhuang: https://github.com/daos-stack/daos/pull/13962#discussion_r1567945589

At this point, `dst_path` is either `/`, has a `/` at the same location as `src_path`, or `src_path` is not a subpath of `dst_path`.


For my previous implementation and your suggestion, all four of these error as expected:
```
daos fs copy --src src  --dst src
daos fs copy --src src  --dst src/
daos fs copy --src src/ --dst src/
daos fs copy --src src/ --dst src
```

I pushed your suggestion and added more comments here

daltonbohning: > Alternatively you could stat the src and dest and if the st_dev and st_ino fields are the same

That would be more complete, but I think we'd have to do some recursive parse + stat to detect subpaths. E.g.
```
src_path = /some/deep/dir
dst_path = /some/deep/dir/that/keeps/going
```
We'd have to stat like this to figure it out (unless I'm missing something)
```
src - /some/deep/dir
dst - /some/deep/dir/that/keeps/going
dst - /some/deep/dir/that/keeps
dst - /some/deep/dir/that
dst - /some/deep/dir
```

mchaarawi (approved):

wiliamhuang (approved):

daltonbohning:
Failure in https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-13962/14/tests
is
https://daosio.atlassian.net/browse/DAOS-15547

PR #13971, opened 2024-03-12
DAOS-15415 control: Enforce no more than 3 bdev tiers in MD-on-SSD

tanabarr:
The maximum number of bdev tiers allowed with MD-on-SSD is 3. Update
some comments that indicate otherwise and prevent assigning roles in
config generate mode when there are more than 3.

Features: control
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/server/storage/config.go
+++ b/src/control/server/storage/config.go
@@ -374,8 +374,8 @@ func (tcs TierConfigs) Validate() error {
 //     scenario allow the use of a single NVMe tier co-locating all three roles, three
 //     separate NVMe tiers with each tier dedicated to exactly one role, or two
 //     separate NVMe tiers where two of the three roles are co-located on one of the
-//     two NVMe tiers. In the latter case, all combinations to co-locate two of the
-//     roles shall be allowed, although not all those combinations may be technically
+//     two NVMe tiers. In the latter case, the only combination of two co-located roles
+//     that is not allowed is Wal+Data. Not all combinations may be technically
 //     desirable in production environments.
 func (tcs TierConfigs) validateBdevRoles() error {
 	scmConfs := tcs.ScmConfigs()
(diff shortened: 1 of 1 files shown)


tanabarr:
Note to reviewers: No test updates have been provided with this change because the validation of the config would always enforce that the effects of AssignBdevTierRoles() were valid during the process of generating a config automatically. In no case would the algorithm generate a config with more than 3 tiers so the change merely updates the checks inside AssignBdevTierRoles() in case it is called with more than 3 due to a future code update. In that case the config validation would fail anyway.

src/control/server/storage/config.go
```diff
@@ -461,10 +461,11 @@ func (tcs TierConfigs) validateBdevRoles() error {
 // Role assignments will be decided based on the following rule set:
 //   - For 1 bdev tier, all roles will be assigned to that tier.
 //   - For 2 bdev tiers, WAL role will be assigned to the first bdev tier and Meta and Data to
```
NiuYawei, line 463: This is inaccurate. For 2 tiers, the combination could be "wal + meta" & "data" or "wal" & "meta + data", we just don't allow "wal + data" & "meta".

tanabarr: Right, this comment is for `AssignBdevTierRoles()` function which is used to automatically assign roles during `dmg|daos_server config generate`. The comment "For 2 bdev tiers, WAL role will be assigned to the first bdev tier and Meta and Data to the second bdev tier." is describing the behaviour of the auto config population where when certain number of SSDs  have been detected for a given engine and the algorithm is trying to create a MD-on-SSD config then the distribution of roles across two tiers will be "wal" & "meta+data" and as you have described this is a valid combination. This isn't a validation function. Does that makes sense?

knard-intel: For my understanding, to not have more than 3 bdev tiers with md-on-ssd is a limitation of the engine ?

NiuYawei: Yes. I see.

NiuYawei (approved):

src/control/server/storage/config.go
tanabarr: @knard-intel This I have discussed with @NiuYawei and it seems unnecessary to be able to support that, it will be simpler for an administrator to understand and there doesn't seem to be any compelling use cases to support more than one data tier Rather than just adding any SSDs to a single data tier. From the perspective of the engine currently, each SSD is simply presented with a role that is assigned to the tier it came from, the tier ID is currently not used (although the information is present in the nvme.conf). If/when we move to differentiating between different SSD types TLC vs QLC and 3D NAND vs NAND then a use case for this may emerge. Also the situation where we have multiple data tiers in the form of e.g. "meta+data" and "data" it's not currently possible to distribute the roles engine side without changing the implementation. HTH

knard-intel: Make sense, thanks for the explanation :)

knard-intel (approved):
LGTM

kjacque (approved):

PR #14014, opened 2024-03-18
DAOS-15322 common: fix untrusted loop bound

juszhan1:
Coverity reports `old_key.ivo_len` as a `TAINTED_SCALAR`. Adding a size check to `ktr_hkey_gen`.

Run-GHA: true

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/common/btree.c
+++ b/src/common/btree.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2016-2023 Intel Corporation.
+ * (C) Copyright 2016-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
--- a/src/common/btree.c
+++ b/src/common/btree.c
@@ -987,7 +987,6 @@ btr_embedded_create_hash(struct btr_context *tcx, bool force)
 			return rc;
 		}
 		D_ASSERT(rec != NULL);
-		D_ASSERT(old_key.iov_len > 0);
 		btr_hkey_gen(tcx, &old_key, &rec->rec_hkey[0]);
 		btr_embedded_hash_set(tcx);
 	}
--- a/src/common/btree.c
+++ b/src/common/btree.c
@@ -987,6 +987,7 @@ btr_embedded_create_hash(struct btr_context *tcx, bool force)
 			return rc;
 		}
 		D_ASSERT(rec != NULL);
+		D_ASSERT(old_key.iov_len > 0);
 		btr_hkey_gen(tcx, &old_key, &rec->rec_hkey[0]);
 		btr_embedded_hash_set(tcx);
 	}
--- a/src/vos/vos_tree.c
+++ b/src/vos/vos_tree.c
@@ -154,7 +154,10 @@ ktr_hkey_gen(struct btr_instance *tins, d_iov_t *key_iov, void *hkey)
 {
 	struct ktr_hkey		*kkey = (struct ktr_hkey *)hkey;
 	struct umem_pool        *umm_pool = tins->ti_umm.umm_pool;
+	struct vos_pool		*pool = (struct vos_pool *)tins->ti_priv;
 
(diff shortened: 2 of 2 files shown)


src/common/btree.c
```diff
@@ -987,6 +987,7 @@ btr_embedded_create_hash(struct btr_context *tcx, bool force)
 			return rc;
 		}
 		D_ASSERT(rec != NULL);
+		D_ASSERT(old_key.iov_len > 0);
```
jolivier23, line 990: I'm not entirely sure this is a valid assertion in that I'm not entirely sure a zero length key is unsupported.  I think what it's complaining about here (and I'm not sure how static analysis is figuring this out) is that the value is used as a loop boundary and there is no checking to see if it's within a reasonable range.   This check doesn't check that. 

The only way static analysis could see this is if it is checking all of the functions used for to_hkey_gen such as ktr_hkey_gen.

The problem is there is no maximum key length (though there are practical limits).  What we can do is make sure the size is correct in all of the gen functions, which, outside of ktr_hkey_gen , should have fixed sized keys for the most part.

But I would limit checking to places it matters rather than trying to do it generically in the top level function.

jolivier23: For ktr_hkey_gen (which may be the only one that has a non-fixed sized key), you could maybe check the maximum based on the size of the vos pool (e.g. <= pool->vp_pool_df->pd_scm_sz)

juszhan1: Thanks for the detailed comment. Added `>= 0` and `< pool->vp_pool_df->pd_scm_sz`.

Coverity was right. ktr_hkey_gen eventually uses the unchecked length to calculate the hash (d_hash_murmur64). 

ktr_hkey_gen is indeed the only function that doesn't have the size check. The functions below do check size or does not look like they need the size.
dtx_cos_hkey_gen
svt_hkey_gen
dtx_hkey_gen
nv_hkey_gen
uv_hkey_gen - doesn't check for size and doesn't necessarily need size.
cont_df_hkey_gen
oi_hkey_gen
ik_hkey_gen - doesn't check for size and doesn't necessarily need size.

What is the difference between `d_iov_t.iov_buf_len` and `d_iov_t.iov_len`?   
https://github.com/daos-stack/daos/blob/master/src/include/gurt/types.h#L70
I would expect to see a memcpy use to use iov_buf_len, but I see it often uses iov_len.

mjmac:
@wangshilong: Jeff is on vacation this week. Would you please review, and suggest someone else on that side for another review? Thanks in advance.

src/vos/vos_tree.c
```diff
@@ -154,7 +154,9 @@ ktr_hkey_gen(struct btr_instance *tins, d_iov_t *key_iov, void *hkey)
 {
 	struct ktr_hkey		*kkey = (struct ktr_hkey *)hkey;
 	struct umem_pool        *umm_pool = tins->ti_umm.umm_pool;
+	struct vos_pool         *pool     = (struct vos_pool *)tins->ti_priv;
```
wangshilong, line 157: I am not sure as far as i checked ti_priv might be inited as cont as well, could we make sure this function could be only used for pool?

NiuYawei: looks the ti_priv is vos_pool for this vos key tree.

NiuYawei (approved):

src/vos/vos_tree.c
```diff
@@ -154,7 +154,9 @@ ktr_hkey_gen(struct btr_instance *tins, d_iov_t *key_iov, void *hkey)
...
 
+	D_ASSERT(key_iov->iov_len < pool->vp_pool_df->pd_scm_sz);
```
jolivier23, line 159: if we want to be really "safe", we could ensure it doesn't go off the end of the pool but let's see if this works for now to quiet coverity

jolivier23 (approved):

PR #14017, opened 2024-03-18
DAOS-15417 build: Fix intermittent issue with libfuse build.

ashleypittman:
Meson "setup" sets up a package for buidlding, meson configure
sets a configuration option, but does not do the setup.
Previously our code would do setup, then configure which would
set configuration options but not apply them.

Ninja has file age checking built-in so if the config file was
older than the build file then it would re-run setup to apply
the correct config, and this was happening most times so
the build would work, but occasionally the file timestamps would
be the same so the check would not fail and the build would be run
without the configuration options applied.

### DIFF
--- a/site_scons/components/__init__.py
+++ b/site_scons/components/__init__.py
@@ -300,14 +300,12 @@ def define_components(reqs):
 
     reqs.define('fuse', libs=['fuse3'], defines=['FUSE_USE_VERSION=35'],
                 retriever=GitRepoRetriever('https://github.com/libfuse/libfuse.git'),
-                commands=[['meson', 'setup', '../fuse'],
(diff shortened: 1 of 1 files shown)


daltonbohning (approved):

jamesanunez (approved):

jolivier23 (approved):

ashleypittman:
@daos-stack/daos-gatekeeper this is a build-only change so ftest results are not relevant, it could be force landed.

mchaarawi:
> @daos-stack/daos-gatekeeper this is a build-only change so ftest results are not relevant, it could be force landed.

in that case, please next time use a pragma to skip functional testing so we don't have to navigate to jenkins and stop the run manually.

PR #14020, opened 2024-03-19
DAOS-15345 dfuse: Ignore return of cont_close.

ashleypittman:
Fixes coverity ID 1979831

Also fix 1981011 about correct handling of fd 0

### DIFF
--- a/src/client/dfuse/dfuse_core.c
+++ b/src/client/dfuse/dfuse_core.c
@@ -964,7 +964,7 @@ dfuse_cont_open(struct dfuse_info *dfuse_info, struct dfuse_pool *dfp, const cha
 err_umount:
 	dfs_umount(dfc->dfs_ns);
 err_close:
-	daos_cont_close(dfc->dfs_coh, NULL);
+	(void)daos_cont_close(dfc->dfs_coh, NULL);
 err_free:
 	D_FREE(dfc);
 err:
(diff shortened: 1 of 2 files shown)


src/client/dfuse/dfuse_core.c
```diff
@@ -964,7 +964,7 @@ dfuse_cont_open(struct dfuse_info *dfuse_info, struct dfuse_pool *dfp, const cha
 err_umount:
 	dfs_umount(dfc->dfs_ns);
 err_close:
-	daos_cont_close(dfc->dfs_coh, NULL);
+	(void)daos_cont_close(dfc->dfs_coh, NULL);
```
mchaarawi, line 967: the same issue can happen on line 965? so might as well do the same here for coverity

mchaarawi (approved):

mchaarawi:
@ashleypittman please stop testing of the PR. i will force land it as soon as you have another +1

mjmac (approved):
Seems fine to me... As a general rule, though, I wonder if we should at least log something if these return nonzero. Maybe a macro like `DL_WARNFAIL()` that emits a warning, e.g.

```
DL_WARNFAIL(dfs_umount(dfc->dfs_ns), "DFS umount failed");
```

It's not necessarily actionable by the user, but might add some context if something weird is going on.

PR #14021, opened 2024-03-19
DAOS-15418 cq: Add documentation for isort githook.

ashleypittman:

### DIFF
--- a/utils/githooks/README.md
+++ b/utils/githooks/README.md
@@ -45,6 +45,11 @@ the version output, it will fail.  Try running
 3. flake8
 4. yamllint
 5. gofmt
+6. isort
+
+The default python version on `rocky 8.8` is `python 3.6` and the latest `isort` available for that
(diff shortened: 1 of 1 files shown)


utils/githooks/README.md
```diff
@@ -45,6 +45,11 @@ the version output, it will fail.  Try running
 3. flake8
 4. yamllint
 5. gofmt
+6. isort
+
+The default python version on `rocky 8.8` is `python 3.6` and the latest `isort` available for that
```
kjacque, line 50: Might be good to specify the isort version affected, and generalize the text for people using python 3.6 on other OS versions.

daltonbohning: It's not an issue with the python version directly, right? It's an issue with the isort version? So should we just say which version of isort has bad performance? To Kris's point, there are many OS flavors and versions, many version of python, and many versions of isort.

PR #14049, opened 2024-03-22
DAOS-15497 build: Only set LD_LIBRARY_PATH in scons build where required.

ashleypittman:
This avoids a long-standing but previously unknown issue where
the build directory was in LD_LIBRARY_PATH when running gcc etc.

Update site_scons to not se LD_LIBRARY_PATH for all commands launched
during the build but rather only set it for the step where the dmg man
pages are generated.

This impacts the daos_build test which bas previously always needed
to run the daos_build with --jobs=1, with this change then the build
can be run in parallel which reduces the run-time of this test by a third.

### DIFF
--- a/site_scons/site_tools/daos_builder.py
+++ b/site_scons/site_tools/daos_builder.py
@@ -64,20 +64,15 @@ def _add_build_rpath(env, pathin="."):
     path = Dir(pathin).path
     env.AppendUnique(LINKFLAGS=[f'-Wl,-rpath-link={path}'])
     env.AppendENVPath('CGO_LDFLAGS', f'-Wl,-rpath-link={path}', sep=' ')
+
+
+def _enable_ld_path(env, part_list):
+    """Add a build directory to LD_LIBRARY_PATY"""
     # We actually run installed binaries from the build area to generate
     # man pages.  In such cases, we need LD_LIBRARY_PATH set to pick up
     # the dependencies
-    env.AppendENVPath("D_LD_PATH", path)
-
-
-def _enable_ld_path(env):
-    """Add a build directory to rpath"""
-    # We actually run installed binaries from the build area to generate
-    # man pages.  In such cases, we need LD_LIBRARY_PATH set to pick up
-    # the dependencies
-    # env.AppendENVPath("LD_LIBRARY_PATH", path)
-    print(env["ENV"]["D_LD_PATH"])
-    env["ENV"]["LD_LIBRARY_PATH"] = env["ENV"]["D_LD_PATH"]
+    for part in part_list:
+        env.AppendENVPath("LD_LIBRARY_PATH", os.path.join(env["BUILD_DIR"], "src", part))
 
 
 def _known_deps(env, **kwargs):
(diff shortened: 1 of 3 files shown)


ashleypittman:
Test result showing the improved daos_build test result is at https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14049/4/testReport/FTEST_dfuse/DaosBuild

Due to the nature of that test where it builds master the PR had to checkout itself in order to run the test, that change is no longer in the PR so the test would fail if run with the PR now hence this is not running with test-tags or features, however once this PR is landed then all tests will pick up the new code.

soumagne:
interesting I would have thought that as long as rpath is properly set there's no need for setting LD_LIBRARY_PATH at all.

soumagne (dismissed):

ashleypittman:
> interesting I would have thought that as long as rpath is properly set there's no need for setting LD_LIBRARY_PATH at all.

For the install path then rpath works as you say.  Some of the build runs the binaries from the staging build directory without installing it and that's what's being addressed here.

site_scons/site_tools/daos_builder.py
```diff
@@ -64,10 +64,15 @@ def _add_build_rpath(env, pathin="."):
     path = Dir(pathin).path
     env.AppendUnique(LINKFLAGS=[f'-Wl,-rpath-link={path}'])
     env.AppendENVPath('CGO_LDFLAGS', f'-Wl,-rpath-link={path}', sep=' ')
+
+
+def _enable_ld_path(env, part_list):
+    """Add a build directory to LD_LIBRARY_PATY"""
```
ashleypittman, line 70:
```suggestion
    """Add a build directory to LD_LIBRARY_PATH"""
```

daltonbohning (commented):
ftest LGTM, but I don't understand the build enough to approve

daltonbohning (commented):
Latest commit need `Features: test_dfuse_daos_build_wt_il`

ashleypittman:
> Latest commit need `Features: test_dfuse_daos_build_wt_il`

See my previous comment, and link to test results. Due to the nature of how this test works that would fail.

soumagne:
> > interesting I would have thought that as long as rpath is properly set there's no need for setting LD_LIBRARY_PATH at all.
> 
> For the install path then rpath works as you say. Some of the build runs the binaries from the staging build directory without installing it and that's what's being addressed here.

ok we're probably missing something then. There should be a build rpath and an install rpath ideally (I have no idea what it takes to do that properly with scons though).

wiliamhuang (approved):

ashleypittman:
> > > interesting I would have thought that as long as rpath is properly set there's no need for setting LD_LIBRARY_PATH at all.
> > 
> > 
> > For the install path then rpath works as you say. Some of the build runs the binaries from the staging build directory without installing it and that's what's being addressed here.
> 
> ok we're probably missing something then. There should be a build rpath and an install rpath ideally (I have no idea what it takes to do that properly with scons though).

Should there?  Why should there be a build rpath as well as a install one?  Would that then get removed when the files are "copied" to the install location?

jolivier23 (approved):

jolivier23:
I assume by build_rpath, you mean what we do with -Wl,-rpath-link.   We shouldn't add build paths to the rpath embedded in the binary unless that binary is a test that isn't installed.

PR #14055, opened 2024-03-22
DAOS-15493 mgmt: Deprecate OFI_INTERFACE/OFI_DOMAIN in libdaos

kjacque:
Collect both current and deprecated network env variables in libdaos.

Required-githooks: true


frostedcmos:
Thank you, I've merged your changes into my PR

PR #14052, opened 2024-03-22
DAOS-14780 rebuild: fix a bug with mpt_inflight_size

liuxuezhao:
one dkey migrate possible exceed the mpt_inflight_max_size, in this case original code possibly cause the dkey migrate ULT dead loop and then rebuild cannot complete.
Example log - "migrate_one_ult() mrone 0x7f3c91fe1ec0 wait 0/33554432", that case will cause the ULT wait again after wakeup until shutdown.

Required-githooks: true

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate owners.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
  * [ ] Quick-build or Quick-functional is not used.
* [ ] Fix the commit message upon landing. Check the standard [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments). Edit it to create a single commit. If necessary, ask submitter for a new summary.

PR #14059, opened 2024-03-25
DAOS-15226 rebuild: expose more settings to control memory

johannlombardi:
Add new environment variables to control how much memory is allocated on the target and system xstreams. Add the ability to tune independently the number of ULTs on both type of xstream.
Also add some documentation.

### Gatekeeper:


kccain (dismissed):
Suggestions for changes only if the patch needs to be updated otherwise.

Also I assume documentation does not need to specify that the resulting memory usages (e.g., 1.0GiB + 2x256MiB) are per pool? I guess since rebuild/reintegration/rebalance would be done one pool at a time? (note: I don't know that for sure, have to go re-review that code while I'm studying rebuild implementation in general).

jolivier23 (approved):

jolivier23 (dismissed):
compiler failure

docs/admin/troubleshooting.md
```diff
@@ -522,14 +522,50 @@ To resolve the issue:
...
+### Running out of Memory during Rebuild
+
+Several environment variables can be set in the env_vars section of the yaml file of the DAOS server
+to control how much memory will be used by the migration service to transfer data while handling
+a rebuild, reintegrate, drain or rebalancing operation.
+
+The D\_MIGRATE\_TGT\_ULT\_CNT and D\_MIGRATE\_SYS\_ULT\_CNT environment variables control how many
+user level threads (i.e. ULT) will be created at most in the engine while processing one of the
+operation listed above. D\_MIGRATE\_TGT\_ULT\_CNT is for the total number of ULTs across the target
+xstreams while D\_MIGRATE\_SYS\_ULT\_CNT is for the system xstream in charge of object scanning.
```
liuxuezhao, line 534: now the "tls->mpt_obj_ult_cnts[tgt_idx]" is to control the number of "migrate_obj_ult()" created on each target VOS XS, "tls->mpt_dkey_ult_cnts[tgt_idx]" is to control the number of "migrate_one_ult()" (dkey/akey migrate ULT) created on each target VOS XS.
we probably need not separately expose sys_ult_cnt and tgt_ult_cnt, exposing one "D_MIGRATE_ULT_CN" as the total number of migrate ULT looks enough to control the number of ULTs used for migration.
The details of how to control the number of tls->mpt_obj_ult_cnts[tgt_idx] and tls->mpt_dkey_ult_cnts[tgt_idx] on each vos target can be implementation internals so can be tuned  later when we want.

johannlombardi:
> Suggestions for changes only if the patch needs to be updated otherwise.
> 
> Also I assume documentation does not need to specify that the resulting memory usages (e.g., 1.0GiB + 2x256MiB) are per pool? I guess since rebuild/reintegration/rebalance would be done one pool at a time? (note: I don't know that for sure, have to go re-review that code while I'm studying rebuild implementation in general).

hm, good point ... actually, i thought that all pool config changes were done in parallel ... let me dig more into the code.

docs/admin/troubleshooting.md
```diff
@@ -522,14 +522,50 @@ To resolve the issue:
...
+Each ULT will consume a stack size of 128KiB. To sum up, the following formula defines how much
+memory will be allocaged at most for all the ULTs in an engine to manage data migration:
+
+```
+(D_MIGRATE_TGT_ULT_CNT + D_MIGRATE_SYS_ULT_CNT) x 128KiB
+```
+
+The default value for both environment variables is 4096. This gives us a 1GB of memory.
+
+Moreover, the D\_MIGRATE\_TGT\_MEM\_SIZE and D\_MIGRATE\_SYS\_MEM\_SIZE environment variables
```
liuxuezhao, line 544: Now the static macro "MIGRATE_MAX_SIZE" is to control the total inflight rebuild data size across all VOS tgt XS, currently the default value set as 25% of system DMA buffer size.
One point is now rebuild possibly generate a few large dkey migrate ULT (migrate_one_ult()) with data size exceed the limit in that case we have to make it pass to avoid the rebuild task cannot complete (https://github.com/daos-stack/daos/pull/14052 fixed a related bug).

This one looks need not to separately define two ENVs for  SYS and TGT either, the number of tls->mpt_inflight_size/tls->mpt_inflight_max_size is not really useful on SYS XS.

docs/admin/troubleshooting.md
```diff
@@ -522,14 +522,56 @@ To resolve the issue:
...
+
+```
+D_MIGRATE_ULT_CNT x 128KiB
+```
+
+The default value for D\_MIGRATE\_ULT\_CNT is 4096. This gives us 0.5GiB of memory used at most
+for the ULT stacks.
+
+Moreover, the D\_MIGRATE\_MEM\_SIZE environment variable defines the maximum total memory in bytes
+allocated by all those migrate ULTs to buffer the data.
```
liuxuezhao, line 544: not very sure if this description is accurate, if user think that rebuild is limited to use these amount of memory to rebuild a pool, then that possibly is not accurate. For example rebuild scan possibly can use more memory for enumerate.

src/object/srv_obj_migrate.c
```diff
@@ -590,9 +593,13 @@ migrate_pool_tls_lookup_create(struct ds_pool *pool, unsigned int version, unsig
...
 		return rc;
 	}
-
-	d_getenv_uint(ENV_MIGRATE_ULT_CNT, &max_migrate_ult);
 	D_ASSERT(generation != (unsigned int)(-1));
+
+	arg.max_ult_cnt = MIGRATE_DEFAULT_MAX_ULT;
+	d_getenv_uint(ENV_MIGRATE_ULT_CNT, &arg.max_ult_cnt);
+	arg.max_mem_size = MIGRATE_DEFAULT_MAX_SIZE;
+	d_getenv_uint64_t(ENV_MIGRATE_MEM_SIZE, &arg.max_mem_size);
```
liuxuezhao, line 601: maybe need to with a safe min value to restrict what the ENV can set, because if set it as a very small value possibly cause the rebuild cannot work well. I can change it in one of my following PR so need not refresh specially for this point.

liuxuezhao (dismissed):

src/object/srv_obj_migrate.c
```diff
@@ -35,13 +35,11 @@
  * the moment, will adjust it later if needed.
  */
 #define MIGRATE_DEFAULT_MAX_SIZE	(1 << 28)
```
kccain, line 37: is the intent of this patch to change the default maximum size in bytes to 256MiB as mentioned in the documentation files? It is still 512MiB here.

docs/admin/troubleshooting.md
```diff
@@ -522,39 +522,45 @@ To resolve the issue:
...
 operation:
 
 ```
-(D_MIGRATE_TGT_ULT_CNT + D_MIGRATE_SYS_ULT_CNT) x 128KiB + D_MIGRATE_TGT_MEM_SIZE + D_MIGRATE_SYS_MEM_SIZE
+D_MIGRATE_ULT_CNT x 128KiB + D_MIGRATE_MEM_SIZE
 ```
 
-With a default of 256MiB for both D\_MIGRATE\_TGT\_MEM\_SIZE and D\_MIGRATE\_SYS\_MEM\_SIZE, this
-gives us an overall upper bound of 1.5GiB of memory allocated during rebuild.
+With a default of 256MiB for both D\_MIGRATE\_MEM\_SIZE, this gives us an overall upper bound
```
kccain, line 553: suggest removing the word "both", since there is just 1 environment variable now.

kccain (changes requested):

src/object/srv_obj_migrate.c
johannlombardi: Nope, i actually did not intend to change any default in this PR (yet).
2^28 bytes = 268435456 bytes = 256MiB
Am i missing something?

src/object/srv_obj_migrate.c
johannlombardi: need to refresh this PR so let me add this. thanks.

src/object/srv_obj_migrate.c
kccain: oops yeah, my mistake on 2^28. Oof.

docs/admin/troubleshooting.md
liuxuezhao: If the enumerate means migrate_one_epoch_object() -> dsc_obj_list_obj(), looks the memory used there is not controlled by tls->mpt_inflight_size/mpt_inflight_max_size.

johannlombardi: Right, but it is allocating everything on the stack (which is pretty bad in itself and should eventually be fixed), but not allocating extra memory.

src/object/srv_obj_migrate.c
```diff
@@ -590,9 +599,17 @@ migrate_pool_tls_lookup_create(struct ds_pool *pool, unsigned int version, unsig
...
 	D_ASSERT(generation != (unsigned int)(-1));
+
+	arg.max_ult_cnt = MIGRATE_DEFAULT_MAX_ULT;
+	d_getenv_uint(ENV_MIGRATE_ULT_CNT, &arg.max_ult_cnt);
+	if (arg.max_ult_cnt < MIGRATE_MIN_MAX_SIZE)
+		arg.max_ult_cnt = MIGRATE_MIN_MAX_SIZE;
+	arg.max_mem_size = MIGRATE_DEFAULT_MAX_SIZE;
+	if (arg.max_mem_size < MIGRATE_MIN_MAX_ULT)
+		arg.max_mem_size = MIGRATE_MIN_MAX_ULT;
+	d_getenv_uint64_t(ENV_MIGRATE_MEM_SIZE, &arg.max_mem_size);
```
liuxuezhao, line 611: looks like it is mis-ordered? should first get the env, then compare with min?

BTW, the "MIGRATE_MIN_MAX_ULT 32" looks too small, in that case if with 16 tgts, then each VOS can only create 2 ULT for migration? if with 32 tgts then can only with 1ULT per tgt that even cannot work. Maybe the min ult count could be 512?
the "MIGRATE_MIN_MAX_SIZE" 1MB looks also too small, but it is not a hard limit although.

liuxuezhao: maybe change not committed before push?
the "d_getenv_uint64_t(ENV_MIGRATE_MEM_SIZE, &arg.max_mem_size);" still mis-ordered, should be moved to L611?

liuxuezhao (approved):

kccain (approved):

jolivier23 (approved):

PR #14094, opened 2024-03-30
DAOS-15575 dfs: replace DAOS_TX_NONE with dfs th

johannlombardi:
This patch lays the ground work for adding support for snapshot to dfs in subsequent patches. It does not change anything from a functionality standpoint and just replaces all occurences of DAOS_TX_NONE in libdfs with dfs->th which can only be set to DAOS_TX_NONE for now.

A dfs::th_epoch field is added to the global dfs handle to record the epoch associated with the handle (can only be set to DAOS_EPOCH_MAX for now, which means DAOS_TX_NONE).

### Gatekeeper:


wiliamhuang (dismissed):

wiliamhuang:
@johannlombardi  one minor question: there are 12 places of "DAOS_TX_NONE" in total in file common.c and cont.c. Do we need to define a macro (example, #define DEFAULT_DFS_TH DAOS_TX_NONE) and change them too?

src/client/dfs/lookup.c
```diff
@@ -141,16 +142,16 @@ lookup_rel_path(dfs_t *dfs, dfs_obj_t *root, const char *path, int flags, dfs_ob
...
 			}
 
-			rc = daos_array_open_with_attr(
-			    dfs->coh, entry.oid, DAOS_TX_NONE, daos_mode, 1,
-			    entry.chunk_size ? entry.chunk_size : dfs->attr.da_chunk_size, &obj->oh,
-			    NULL);
+			rc = daos_array_open_with_attr(dfs->coh, entry.oid, dfs->th, daos_mode, 1,
+						       entry.chunk_size ? entry.chunk_size
+									: dfs->attr.da_chunk_size,
+						       &obj->oh, NULL);
```
mchaarawi, line 148: that tabbing looks bad to me..

mchaarawi (commented):
seems fine to me, although if this is only for allowing mounting a snapshot, we should not change really the write / punch calls to the th and should just use TX_NONE as before to be more explicit / clear.

johannlombardi:
@mchaarawi I think that using dfs->th on the write path has two benefits:
- I/Os will still fail on a snapshot if we somehow does not properly return RDONLY earlier instead of writing (and potentially corrupting) the latest epoch
- if we support writeable snapshot in the future, we could just hide all the complexity behind the transaction handle without needed to replace again TX_NONE everywhere.

johannlombardi:
@wiliamhuang  you mean that there are other places where i should have changed TX_NONE to dfs->th that i missed?

mchaarawi:
> @mchaarawi I think that using dfs->th on the write path has two benefits:
> 
> * I/Os will still fail on a snapshot if we somehow does not properly return RDONLY earlier instead of writing (and potentially corrupting) the latest epoch
> * if we support writeable snapshot in the future, we could just hide all the complexity behind the transaction handle without needed to replace again TX_NONE everywhere.

im not sure how we can support a writeable snapshot in our TX model TBH.
we can support in the future (although pretty tricky for users to understand) writing to a mounted snapshot at EPOCH_MAX, in which case we would not use the th (but we would need to revert this change again).

mchaarawi:
> > @mchaarawi I think that using dfs->th on the write path has two benefits:
> > 
> > * I/Os will still fail on a snapshot if we somehow does not properly return RDONLY earlier instead of writing (and potentially corrupting) the latest epoch
> > * if we support writeable snapshot in the future, we could just hide all the complexity behind the transaction handle without needed to replace again TX_NONE everywhere.
> 
> im not sure how we can support a writeable snapshot in our TX model TBH. we can support in the future (although pretty tricky for users to understand) writing to a mounted snapshot at EPOCH_MAX, in which case we would not use the th (but we would need to revert this change again).

anyway, in any case, im not against this change in particular. as of this PR now, nothing has changed :-)

src/client/dfs/lookup.c
johannlombardi: hm, it looks fine in the patch ...

mchaarawi (dismissed):

mchaarawi:
as a gatekeeper, i would point out that this needs to run with `Features: dfs` pragma :-)

wiliamhuang (approved):

johannlombardi:
Should be good to go. Please review again.

mchaarawi (approved):

wiliamhuang (approved):

PR #14118, opened 2024-04-04
DAOS-15522 test: daos_test fi always for potential dmg retry

kccain:
Adjust daos_test -p, pool_op_retry() fault injection before dmg operations to specify fail always (until explicitly cleared), rather than fail once with implicit clear.

The change is needed because dmg / control plane can "fan out" gRPC requests to multiple MS daos_server instances, and may (intermittently) initiate retries when one returns ErrNotLeader and the other returns an error(e.g., -DER_MISC in this case). Before this change, dmg retry would result in rc=0 surfaced to daos_test (contrary to the expected -DER_MISC). With this change, the dmg retry (that is actually a distinct operation from the engine standpoint) will also have the fault injected in the engine handler, surfacing the intended -DER_MISC result to the dmg caller (daos_test).

Test-tag: test_daos_pool
Test-repeat: 10
Skip-unit-tests: true

### Gatekeeper:

### DIFF
--- a/src/tests/suite/daos_pool.c
+++ b/src/tests/suite/daos_pool.c
@@ -760,10 +760,11 @@ pool_op_retry(void **state)
 
 	/* TODO: implement dup op detection in ds_pool_update_handler()? */
 #if 0
-	test_set_engine_fail_loc(arg, info.pi_leader, DAOS_MD_OP_PASS_NOREPLY | DAOS_FAIL_ONCE);
+	test_set_engine_fail_loc(arg, leader_rank, DAOS_MD_OP_PASS_NOREPLY | DAOS_FAIL_ALWAYS);
 	print_message("draining rank %d target idx 0 ... ", info.pi_leader);
 	rc = dmg_pool_drain(arg->dmg_config, arg->pool.pool_uuid, arg->group, info.pi_leader, 0);
 	assert_rc_equal(rc, 0);
+	test_set_engine_fail_loc(arg, leader_rank, 0);
 	print_message("success\n");
 #endif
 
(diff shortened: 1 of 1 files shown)


kccain:
10 test loops of daos_test -p (all pool tests) passed, including the pool_op_retry() that is modified here. Ready for review.

src/tests/suite/daos_pool.c
```diff
@@ -797,20 +798,24 @@ pool_op_retry(void **state)
 	assert_rc_equal(rc, -DER_MISC);
 	print_message("success\n");
 
-	/* pool update ACL entry failure committed, "lost" reply - duplicate RPC retry */
-	test_set_engine_fail_loc(arg, leader_rank, DAOS_MD_OP_FAIL_NOREPLY | DAOS_FAIL_ONCE);
+	/* pool update ACL entry failure committed, "lost" reply - duplicate RPC retry.
+	 * NB: with dmg, fail until cleared (in case control plane "retries")
+	 */
+	test_set_engine_fail_loc(arg, leader_rank, DAOS_MD_OP_FAIL_NOREPLY | DAOS_FAIL_ALWAYS);
```
liw, line 804: [Question] Is it correct that this fault injection causes a related operation to complete with result `-DER_MISC` while the reply `rc` is `-DER_TIMEDOUT`? If that's correct, the test will hang at such an operation. Obviously my understanding is missing something. Could you point out what I'm missing, please? :)

kccain: RPC handling 1: complete with -DER_MISC (and save it into svc_ops KVS) and return -DER_TIMEDOUT
RPC handling 2 (the retry): look up the previous -DER_MISC result and reply with that error rc.

If the control plane intermittently may initiate the whole thing all over again as a brand-new operation (because of its "fan out" design to multiple MS daos_server instnaces), then the sequence above should repeat.

kccain: actually let me think about it more. I tested it and it ran successfully, but how could it be if we set fault injection for always. I'll be off for a short while and will check next week. Maybe a workaround for control plane logic is not possible with this approach.

kccain: After reviewing the code again, and the artifacts from the latest CI testing:

1. client test_set_engine_fail_loc(fail_noreply|always)
2. gRPC 1, cart RPC handling 1 returns -DER_TIMEDOUT and saves -DER_MISC as the result of the operation
3. gRPC 1, cart RPC handling 2 sees fault injection is (still) configured, *and* also sees that this is a duplicate operation. Because it is dup, the logic inside the handler (in this case ds_pool_acl_update_handler()) replies with the retrieved status, -DER_MISC. i.e. the duplicate detection takes precedence over the fault injection configuration (which remains enabled even after the reply).

In most cases in which the control plane does not retry the gRPC from dmg utility, we're done and we have passed the test (and the test would clear the fault injection with test_set_engine_fail_loc(value=0). But let's say we hit the rare case that dmg retries. In this case steps 2 and 3 are repeated with "gRPC 2" and daos_test will ultimately see -DER_MISC returned, which is what the test expects.

I tried to find a "smoking gun" MS retry scenario among the 10 repeat loops from the Jenkins build 1 CI testing. At least for the acl update and delete handlers (which are easier to grep for than the other operation for which this applies, pool evict), I did not see the case we want to solve occur. So at this point I am going on the above code analysis that this will work when the MS on rare occasions will retry the gRPC.

liw: Thanks for taking a closer look at this.

liw (approved):

wangshilong (approved):

PR #14122, opened 2024-04-05
DAOS-15598 cart: Prevent engines from setting wrong number of contexts

frostedcmos:
- Prevent engines from using "CRT_CTX_NUM" setting to restrict number of contexts. This option is only proper for scalable endpoint setting which is no longer supported by any provider.

- This is intended to fix a rebuild test that currently sets SEP=true, CRT_CTX_NUM=8 while trying to start engines with more contexts than 8.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/cart/crt_init.c
+++ b/src/cart/crt_init.c
@@ -197,18 +197,20 @@ prov_data_init(struct crt_prov_gdata *prov_data, crt_provider_t provider,
 	if (rc != 0)
 		return rc;
 
-	/* Set max number of contexts. Defaults to the number of cores */
-	ctx_num = 0;
-
-	if (crt_is_service())
+	if (crt_is_service()) {
 		ctx_num = CRT_SRV_CONTEXT_NUM;
-	else
+		max_num_ctx = CRT_SRV_CONTEXT_NUM;
+	} else {
+		/* Only limit the number of contexts for clients */
 		d_getenv_uint("CRT_CTX_NUM", &ctx_num);
 
-	if (opt)
-		max_num_ctx = ctx_num ? ctx_num : max(crt_gdata.cg_num_cores, opt->cio_ctx_max_num);
-	else
-		max_num_ctx = ctx_num ? ctx_num : crt_gdata.cg_num_cores;
+		/* Default setting to the number of cores */
+		if (opt)
+			max_num_ctx = ctx_num ? ctx_num :
+				      max(crt_gdata.cg_num_cores, opt->cio_ctx_max_num);
+		else
+			max_num_ctx = ctx_num ? ctx_num : crt_gdata.cg_num_cores;
+	}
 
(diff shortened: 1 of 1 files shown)


src/cart/crt_init.c
```diff
@@ -215,26 +219,6 @@ prov_data_init(struct crt_prov_gdata *prov_data, crt_provider_t provider,
 
 	D_DEBUG(DB_ALL, "Max number of contexts set to %d\n", max_num_ctx);
 
-	/* Assume for now this option is only available for a primary provider */
```
frostedcmos, line 218: Note: This section is going away in next patch anyway, as we no longer support SEP mode anywhere. Decided to nuke it here as well, instead of trying to rework the code that will get nuked

soumagne (approved):

src/cart/crt_init.c
```diff
@@ -198,13 +197,20 @@ prov_data_init(struct crt_prov_gdata *prov_data, crt_provider_t provider,
...
 		return rc;
 
-	/* Set max number of contexts. Defaults to the number of cores */
-	ctx_num = 0;
-	d_getenv_uint("CRT_CTX_NUM", &ctx_num);
-	if (opt)
-		max_num_ctx = ctx_num ? ctx_num : max(crt_gdata.cg_num_cores, opt->cio_ctx_max_num);
-	else
-		max_num_ctx = ctx_num ? ctx_num : crt_gdata.cg_num_cores;
+	if (crt_is_service()) {
```
kccain, line 200: in a current test (test_daos_degraded_mode) it appears to have set CRT_CTX_SHARE_ADDR = 1 and CRT_CTX_NUM = 8 which (after a recent master commit) causes engine startup errors, but previously worked.

With this change, it looks like the test environment's setting of CRT_CTX_SHARE_ADDR will be ignored, and the CRT_CTX_NUM value will also be ignored, and instead be set to CRT_SRV_CONTEXT_NUM(=64)? 

Ok, actually in separate chat with @frostedcmos previously when the test worked the share addr (SEP) and CTX_NUM settings were ignored in engine anyway. So this change gets the engine setup the way we had it before when the testing worked?

kccain (approved):

src/cart/crt_init.c
frostedcmos: Correct. Previous setting of CRT_CTX_NUM=8 would be ignored on the engines due to SEP being unsupported by providers, so this patch should bring engines back to the same behavior as before my patch that broke things.

phender:
The failures in https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14122/3/ appear to known issues:
- https://daosio.atlassian.net/browse/DAOS-15515
- https://daosio.atlassian.net/browse/DAOS-15124

PR #14138, opened 2024-04-09
DAOS-15603: Support of FIO with ansible ftest script

knard-intel:
### Description

Miscellaneous improvements allowing to run fio functional tests, and to use the shared cluster.

Required-githooks: true

### Gatekeeper:


utils/ansible/ftest/file/enable-coredumps.sh
```diff
@@ -0,0 +1,44 @@
...
+set -u -e -o pipefail
+
+if [[ $(id -u) -ne 0 ]] ; then
+	echo "[ERROR] Could only be used by root"
+fi
+
+sed -i -E -e "/DefaultLimitCORE/s/.*/DefaultLimitCORE=infinity:infinity/" /etc/systemd/system.conf
+
+cat > /etc/security/limits.d/daos-coredumps.conf << EOF
+# /etc/security/limits.d/daos-coredumps.conf
```
shimizukko, line 13: Is this just a commented out line that you can remove? It's the same path as the line above.

utils/ansible/ftest/inventory-sample.yml
```diff
@@ -3,11 +3,13 @@
 all:
   vars:
     daos_runtime_dir: /home/foo/daos
+    # daos_ofi_provider: ofi+tcp;ofi_rxm
```
shimizukko, line 6: If you comment it out for now and plan to use it later, you might want to add some explanation.

shimizukko (approved):

utils/ansible/ftest/file/enable-coredumps.sh
knard-intel: > Is this just a commented out line that you can remove? It's the same path as the line above.

Not sure to understand. Line 13 is the header of the output file created at line 12 with the cat command.  From what know, it is a common practice to have the first line of such configuration file defining the location of the file.  Does it make sense ?

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -13,13 +14,18 @@ SERVERS_LIST="{{ groups["daos_servers"] | join(',') }}"
 {% if "daos_clients" in groups and groups["daos_clients"] | length > 0 %}
 CLIENTS_LIST="{{ groups["daos_clients"] | join(',') }}"
 {% endif %}
+BUILD_TYPE="{{ daos_build_mode | default(daos_build_mode_default) }}"
 
 SCONS_EXE=scons-3
```
ashleypittman, line 19: Everything should be `scons` now, `scons-3` was about transitioning from python2 which should finally be complete.

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -242,14 +256,28 @@ run $CLUSH_EXE $CLUSH_OPTS -l root -w $SERVERS_LIST rm -f /etc/ld.so.cache
 run $CLUSH_EXE $CLUSH_OPTS -l root -w $SERVERS_LIST ldconfig
 {% endif %}
 
-if [[ $MPICH_PATH ]] ; then
+if [[ ! -f "$DAOS_INSTALL_DIR/bin/activate" ]] ; then
+	info "Installing avocado"
+	run $PYTHON_EXE -m venv "$DAOS_INSTALL_DIR"
+	source "$DAOS_INSTALL_DIR/bin/activate"
+	run $PIP_EXE install --upgrade pip
+	run $PIP_EXE install -r "$DAOS_SOURCE_DIR/requirements.txt"
```
ashleypittman, line 264: After my change this could install fewer packages.
```suggestion
	run $PIP_EXE install -r "$DAOS_SOURCE_DIR/requirements-build.txt"
```

utils/ansible/ftest/inventory-sample.yml
knard-intel: This comment present and example of an optional parameters.
if it is confusing, I can remove it or add something like this 
```suggestion
    # Optional parameter
    # daos_ofi_provider: ofi+tcp;ofi_rxm
```

utils/ansible/ftest/file/enable-coredumps.sh
shimizukko: ah okay. Thanks for the explanation.

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: I started with requirements-ftest.txt, and I thought that what can do more, can do less.  This is why I switched to the full install.  I am also hopping that other tasks than running ftest could be perform with installing all the python packages.
However, I have no strong opinion on this point and thus I will do what you think is better.

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: - [ ] Replace scons-3 with scons

utils/ansible/ftest/inventory-sample.yml
knard-intel: - [ ] Add default parameter explanation

knard-intel: > * [x]  Add default parameter explanation

Fixed with commit 07b4777df

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: > * [x]  Replace scons-3 with scons

Fixed with commit 07b4777df

utils/ansible/ftest/templates/daos-launch.sh.j2
```diff
@@ -248,12 +248,16 @@ common_opts="-ar --provider={{ daos_ofi_provider | default(daos_ofi_provider_def
...
-env OFI_INTERFACE=$OFI_INTERFACE $PYTHON_EXE launch.py $common_opts "${args[@]}"
+run env OFI_INTERFACE=$OFI_INTERFACE $PYTHON_EXE launch.py $common_opts "${args[@]}"
 {% else %}
-$PYTHON_EXE launch.py $common_opts "${args[@]}"
+run $PYTHON_EXE launch.py $common_opts "${args[@]}"
 {% endif %}
 rc=$?
+set +u
+deactivate
+set -u
```
ashleypittman, line 260: FYI - if you copied this from my code then when using a later python version than 3.6 it's the +u/-u settings are no longer required.  @phender this will apply to your PR as well, if using a newer python then you could remove these extra lines.

knard-intel: I read the comment about 3.6 or higher, I will remove it as it is clearly "belt and braces" as we say in France ;-)

- [ ] Remove useless set +u before activate

ashleypittman: In fact I can elaborate further.  Phil is upgrading the avocado version in https://github.com/daos-stack/daos/pull/14148 and that requires a newer python which them means that the python bug doesn't work with `+u` set will no longer apply.  For this PR you should IMO also use a newer python version (there's no reason not to use the latest version available in the distro which is probably 3.11)  even though it may not be strictly needed today - otherwise you'll have to rework this soon.

The `-u` changes needed applies to all instances of the `deactivate` command in this PR obviously.

knard-intel: > I will remove it as it is clearly "belt and braces" as we say in France ;-)
> 
> * [x]  Remove useless set +u before activate

Fixed with commit ad2c133d7

ashleypittman: Awesome.

utils/ansible/ftest/templates/daos-make.sh.j2
ashleypittman: The disadvantage of using `requirements.txt` is that more software will be installed and therefore it'll take slightly longer and use minimally more bandwidth and disk space.  If you're running the build, NLT (which counts as unit testing) and ftest from the same virtual environment then this is almost correct - you'll probably have a package or two that are only used for interacting with GitHub but the alternative would be to use 3 different requirements files here which would be a mess as well.

I only mentioned it as the behaviour/requirements of this file has changed since you started this PR and I wanted to let you know that you might need to re-evaluate this.  It sounds like no change is needed.

knard-intel: From my side, I would prefer to keep as it is: the overhead (i.e. disk usage and overhead time) compared to the build process of DAOS and dependencies is neglectable.
I confirm that one aim of this ansible stuff is to be able to build DAOS, run ftest and run NLT with the same virtual env.

shimizukko (approved):

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -196,6 +208,18 @@ if [[ ! -d "$DAOS_INSTALL_DIR" ]] ; then
...
+	deactivate
+	info "Customize venv for DAOS install"
+	run rm -fv "$DAOS_INSTALL_DIR/lib64"
+	run mkdir -pv "$DAOS_INSTALL_DIR/lib64"
```
ashleypittman, line 220: What is this bit doing?

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -196,6 +208,18 @@ if [[ ! -d "$DAOS_INSTALL_DIR" ]] ; then
...
+fi
+
 cd "$DAOS_SOURCE_DIR"
 SCONS_OPTS="-j $JOBS_NB --implicit-deps-changed"
```
ashleypittman, line 224: FYI - scons has a `-C <directory>` option you could use here to avoid the use of `cd`

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -196,6 +208,18 @@ if [[ ! -d "$DAOS_INSTALL_DIR" ]] ; then
...
 case $CMD in
```
ashleypittman, line 225: `--config-force` is an option that seems to be used ubiquitously but not actually required.  If you can demonstrate that it's required then please file a ticket but in practice it shouldn't be needed here.  The same probably applies to the `--implicit-deps-changed` option.

There's a `--build-deps=only` option, the install could call this and then fall through to the update case which would perhaps be clearer.

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -205,12 +229,19 @@ case $CMD in
 esac
 
 info "Building and installing DAOS from source tree $PWD to $DAOS_INSTALL_DIR"
-if ! run env MPI_PKG=any "$SCONS_EXE" BUILD_TYPE=debug PREFIX="$DAOS_INSTALL_DIR" install $SCONS_OPTS ; then
+source "$DAOS_INSTALL_DIR/bin/activate"
+if ! run env MPI_PKG=any "$SCONS_EXE" BUILD_TYPE="$BUILD_TYPE" PREFIX="$DAOS_INSTALL_DIR" install $SCONS_OPTS ; then
```
ashleypittman, line 233: `MPI_PKG`, 'BUILD_TYPE` and 'PREFIX' will should all be set on the build line above on L227, otherwise the tree will be built twice I'd expect?

If this is expected to do both the build and install then what is L227 doing?

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: good to know, thanks for the tips.

- [ ] Use scons with -C option

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: I added these two options, as they were at required at a point of time.
I will check, if they are still needed for my use case.
If no, I will remove them.

- [ ] Check if --config-force and --implicit-deps-changed options are still needed.
- [ ] Build in two steps with --build-deps=only option.

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: The L226 is used  for building and install a fresh DAOS and line 227 just to update a previous local install done with the line 226.  However, it is highly probable that it is the not the way to use scons as I am no used to it.

utils/ansible/ftest/templates/daos-make.sh.j2
ashleypittman: It would be good to keep these separate if we can.  They may be some issues with `pydaos.raw` doing that however, but I think you address that with `PYTHONPATH` above?

utils/ansible/ftest/templates/daos-make.sh.j2
ashleypittman: For the general case and indeed this work there are three stages of the build that scons needs to perform

1. Building and installing all the deps.
2. Building daos
3. installing daos.

For 1. you can use `--build-deps=yes` to perform it at the same time as 2. however using `--build-deps=only` will perform 1. but not 2.
2. Options like `BUILD_TYPE`, `PREFIX` etc can be passed to any invocation and they'll be saved in `daos.conf` and used for future invocations.
For 3. if daos needs to be rebuilt because of options then it'll always rebuilt it.  As far as I'm aware there's no "install but fail if rebuild is required" option.

I think this code will call `scons --build-deps=yes` which will then build and install the dependencies, and then build daos.  It's then calling `scons` which will be a no-op as the code is built and then it's calling `scons install PREFIX=` which will re-build and then install the whole tree.

For correctness and to get any errors reported where you'd expect them and for the timings to be as expected then this code should be doing

- `scons --build-deps=only`
- `scons PREFIX=<> BUILD_TYPE=<>`
- `scons install`

This way the three commands that this script launches will match with the three build steps I illustrated above.  Rebuilds from the same source tree can miss the first step.  In theory they could skip the second as well unless the options have changed but they would want to do steps 2 and 3.

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: - [ ] Try to split virtualenv and DAOS local install

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: - [ ] Properly call scons

knard-intel: > * [x]  Properly call scons

Fixed with commit 8a62b1f000ca6c212e790b7aaddf992c98651969

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: > I added these two options, as they were at required at a point of time. I will check, if they are still needed for my use case. If no, I will remove them.
> 
> * [x]  Remove useless scons option --config-force and --implicit-deps-changed.

Fixed with commit f8c49965c035fd3a39fe24c51e09a44e628bf0d4

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: > good to know, thanks for the tips.
> 
> * [x]  Use scons with -C option

Fixed with commit bce8c6cb288ad0605a53f42649b035f6e81d8546

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: > * [x]  Split virtualenv and DAOS local install

Fixed with commit 79615a794810dfe810a571416598c8cf8f9584b4

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -196,21 +204,49 @@ if [[ ! -d "$DAOS_INSTALL_DIR" ]] ; then
...
+fi
+
+info "Building DAOS from source tree $DAOS_SOURCE_DIR"
+if ! run env MPI_PKG=any "$SCONS_EXE" BUILD_TYPE="$DAOS_BUILD_TYPE" PREFIX="$DAOS_INSTALL_DIR" $SCONS_OPTS --build-deps=no ; then
+	fatal "DAOS could not be properly build"
+fi
+
+info "Installing DAOS to $DAOS_INSTALL_DIR"
+source "$DAOS_VENV_DIR/bin/activate"
+if ! run env MPI_PKG=any "$SCONS_EXE" BUILD_TYPE="$DAOS_BUILD_TYPE" PREFIX="$DAOS_INSTALL_DIR" install $SCONS_OPTS ; then
```
ashleypittman, line 239: The opts are used here but they'll be saved in daos.conf by the invocation above.
```suggestion
if ! run "$SCONS_EXE" install $SCONS_OPTS ; then
```

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -196,21 +204,49 @@ if [[ ! -d "$DAOS_INSTALL_DIR" ]] ; then
...
+
+if "$DAOS_BUILD_DEPS" ; then
+	info "Building DAOS dependencies from source tree $DAOS_SOURCE_DIR"
+	if ! run env MPI_PKG=any "$SCONS_EXE" BUILD_TYPE="$DAOS_BUILD_TYPE" PREFIX="$DAOS_INSTALL_DIR" $SCONS_OPTS --build-deps=only ; then
+		fatal "DAOS dependencies could not be properly build"
+	fi
+fi
+
+info "Building DAOS from source tree $DAOS_SOURCE_DIR"
+if ! run env MPI_PKG=any "$SCONS_EXE" BUILD_TYPE="$DAOS_BUILD_TYPE" PREFIX="$DAOS_INSTALL_DIR" $SCONS_OPTS --build-deps=no ; then
```
ashleypittman, line 233: --build-deps=no is the default.
```suggestion
if ! run env MPI_PKG=any "$SCONS_EXE" BUILD_TYPE="$DAOS_BUILD_TYPE" PREFIX="$DAOS_INSTALL_DIR" $SCONS_OPTS ; then
```

utils/ansible/ftest/templates/daos-make.sh.j2
```diff
@@ -196,21 +204,49 @@ if [[ ! -d "$DAOS_INSTALL_DIR" ]] ; then
...
+info "Installing DAOS python virtualenv $DAOS_VENV_DIR"
+source "$DAOS_VENV_DIR/bin/activate"
+run $PIP_EXE install --upgrade pip
+run $PIP_EXE install -r "$DAOS_SOURCE_DIR/requirements.txt"
+
+SCONS_OPTS="--directory="$DAOS_SOURCE_DIR" --jobs=$JOBS_NB"
+
+if "$DAOS_BUILD_DEPS" ; then
+	info "Building DAOS dependencies from source tree $DAOS_SOURCE_DIR"
+	if ! run env MPI_PKG=any "$SCONS_EXE" BUILD_TYPE="$DAOS_BUILD_TYPE" PREFIX="$DAOS_INSTALL_DIR" $SCONS_OPTS --build-deps=only ; then
```
ashleypittman, line 227: These options aren't used here.
```suggestion
	if ! run "$SCONS_EXE" SCONS_OPTS --build-deps=only ; then
```

utils/ansible/ftest/vars/Rocky8.yml
```diff
@@ -6,39 +6,60 @@ daos_openmpi_module_path: /usr/share/modulefiles/mpi/openmpi-x86_64
...
-  pips:
-    - "avocado=={{ daos_avocado_version | default(daos_avocado_version_default) }}"
-    - "avocado-framework=={{ daos_avocado_framework_version | default(daos_avocado_framework_version_default) }}"
-    - "avocado-framework-plugin-glib=={{ daos_avocado_framework_version | default(daos_avocado_framework_version_default) }}"
-    - "avocado-framework-plugin-golang=={{ daos_avocado_framework_version | default(daos_avocado_framework_version_default) }}"
-    - "avocado-framework-plugin-loader-yaml=={{ daos_avocado_framework_version | default(daos_avocado_framework_version_default) }}"
-    - "avocado-framework-plugin-result-html=={{ daos_avocado_framework_version | default(daos_avocado_framework_version_default) }}"
-    - "avocado-framework-plugin-result-upload=={{ daos_avocado_framework_version | default(daos_avocado_framework_version_default) }}"
-    - "avocado-framework-plugin-varianter-yaml-to-mux=={{ daos_avocado_framework_version | default(daos_avocado_framework_version_default) }}"
-    - pyelftools
```
ashleypittman, line 63: This was better as it was, we should be installing avocado and the other tools from pip.

knard-intel: Not sure to understand this comment.
With this new version, avocao is indeed installed with pip in the virtualenv.

ashleypittman: Sorry, my mistake.

utils/ansible/ftest/vars/Rocky8.yml
```diff
@@ -6,39 +6,60 @@ daos_openmpi_module_path: /usr/share/modulefiles/mpi/openmpi-x86_64
...
+    - python3-distro
+    - python3-jira
+    - python3-junit_xml
+    - python3-pip
+    - python3-pyelftools
+    - python3-scons
+    - python3-tabulate
+    - python3-wheel
+    - python3-pyxattr
+    - python3-pyyaml
```
ashleypittman, line 57: Are a lot of these new packages needed now then?  python3.11, python3.11-devel and python-wheel might be enough here.

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: Are you sure PREFIX and BUILD_TYPE are not used.
I was expecting that at least these two variables to be used by the "configure" script of the dependencies.

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: I would prefer to keep it as reminder of the option use.
If it is bothering you, I will add a comment indicating that this option is used with this default value.

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: - [ ] Remove useless opts

utils/ansible/ftest/vars/Rocky8.yml
knard-intel: Agree with you, this is some reminder of old experiments.

- [ ] Remove useless python related packages

utils/ansible/ftest/templates/daos-make.sh.j2
ashleypittman: You're probably right.  There's a DEPS_BUILD_TYPE but now that I think about it maybe the deps use BUILD_TYPE and the build uses BUILD_TYPE and also DEPS_BUILD_TYPE if that was different.

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: > * Remove useless opts

Fixed with commit 97ec61dff

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: > * [x]  Remove useless option and add a comment

Fixed with commit 97ec61dff

utils/ansible/ftest/templates/daos-make.sh.j2
knard-intel: > > You're probably right. There's a DEPS_BUILD_TYPE but now that I think about it maybe the deps use BUILD_TYPE and the build uses BUILD_TYPE and also DEPS_BUILD_TYPE if that was different.
> 
> * [x]  Remove useless env

Fixed with commit 97ec61dff

shimizukko (dismissed):

utils/ansible/ftest/vars/Rocky8.yml
```diff
@@ -6,39 +6,51 @@ daos_openmpi_module_path: /usr/share/modulefiles/mpi/openmpi-x86_64
...
+    - gcc-c++
+    - gcc-gfortran
+    - golang
     - libaec-devel
+    - libattr-devel
+    - libcmocka-devel
+    - libevent-devel
+    - libyaml-devel
     - meson
+    - ninja-build
```
ashleypittman, line 42: clustershell, meson and ninja-build also might not be needed here.

utils/ansible/ftest/vars/Rocky8.yml
```diff
@@ -6,39 +6,51 @@ daos_openmpi_module_path: /usr/share/modulefiles/mpi/openmpi-x86_64
...
+    - openssl-devel
+    - python3.11
+    - python3.11-devel
     - python3-clustershell
```
ashleypittman, line 46: and possibly python-clustershell.  For regular ftest then it's installed via pip but I do remember there being something special about it where it's also required outside the environment as well so maybe this is needed still.

ashleypittman (dismissed):
Build/setup changes look goot to me.

utils/ansible/ftest/vars/Rocky8.yml
knard-intel: - [ ] Remove useless dependencies

utils/ansible/ftest/vars/Rocky8.yml
knard-intel: ClusterShell dependencies are indeed needed outside ftest and the build of DAOS.
Thus I would like to keep them.

utils/ansible/ftest/vars/Rocky8.yml
knard-intel: > * [x]  Remove useless dependencies

Fixed with commit 03ae13204aacf1821a8cd75734207322c7ed3881

ashleypittman (approved):

shimizukko (dismissed):

knard-intel:
@daos-stack/daos-gatekeeper  please could you lend this PR with the following commit message:
Title: `DAOS-15603 test: Support of FIO with ansible ftest script`
Message:
```
Add support to build and run fio functional tests.
Add support of the shared cluster nodes.
Add support of MOFED libraries.
Miscellaneous improvements of the build process:
- improvement of scons usage
- use of virtualenv with pip
- etc...
Miscellaneous improvements allowing to run fio functional tests, and to use the shared cluster.
```

utils/ansible/ftest/file/configure-nofile.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# set -x
+set -u -e -o pipefail
+
+if [[ $(id -u) -ne 0 ]] ; then
+	echo "[ERROR] Could only be used by root"
+fi
```
daltonbohning, line 8: Should this `exit 1`?

knard-intel: Arghhhh, thanks for your attention.

utils/ansible/ftest/file/enable-coredumps.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+# set -x
+set -u -e -o pipefail
+
+if [[ $(id -u) -ne 0 ]] ; then
+	echo "[ERROR] Could only be used by root"
```
knard-intel, line 7: - [ ] Fix invalid missing exit 1

utils/ansible/ftest/file/configure-nofile.sh
knard-intel: > * [x]  Fix invalid missing exit 1

Fixed wit commit [7d876bc](https://github.com/daos-stack/daos/pull/14138/commits/7d876bc371e97e6a19ce79e006142022e2fff1a6)

utils/ansible/ftest/file/enable-coredumps.sh
knard-intel: > * [x]  Fix invalid missing exit 1

Fixed wit commit [7d876bc](https://github.com/daos-stack/daos/pull/14138/commits/7d876bc371e97e6a19ce79e006142022e2fff1a6)

shimizukko (approved):

PR #14170, opened 2024-04-16
DAOS-15655 control: fix support for non default system name

mchaarawi:
Features: control

Required-githooks: true

### Gatekeeper:


mjmac (approved):

tanabarr (approved):

src/include/daos/pool.h
```diff
@@ -78,6 +78,9 @@
  */
 #define DAOS_POOL_GLOBAL_VERSION 3
 
+/** Default system name used in srv_pool.c */
+extern char daos_sysname[];
+
```
liw, line 83: Since `daos_sysname` is defined in `engine`, its declaration should be in daos_engine.h instead, no?

src/engine/init.c
```diff
@@ -46,7 +44,7 @@ static char		modules[MAX_MODULE_OPTIONS + 1];
 static unsigned int	nr_threads;
 
 /** DAOS system name (corresponds to crt group ID) */
-static char	       *daos_sysname = DAOS_DEFAULT_SYS_NAME;
+char                    daos_sysname[DAOS_SYS_NAME_MAX + 1] = DAOS_DEFAULT_SYS_NAME;
```
liw, line 47: [Question] Why is this change to array necessary?

src/engine/init.c
```diff
@@ -1032,7 +1030,7 @@ parse(int argc, char **argv)
 				rc = -DER_INVAL;
 				break;
 			}
-			daos_sysname = optarg;
+			strcpy(daos_sysname, optarg);
```
liw, line 1033: [Nit?] For this to be safe, we have to assume that `optarg` won't be too long. Although that may indeed be true, it's not obvious---we might want to check the length (either by `strlen` or `strncpy`).

src/engine/init.c
mchaarawi: to allow for a sys name that might be larger than DAOS_DEFAULT_SYS_NAME and could be up to DAOS_SYS_NAME_MAX

src/include/daos/pool.h
mchaarawi: im not sure that it matters? we only use it in the the server client pool operations, but i will move it, which would require though including daos_engine.h

src/engine/init.c
mchaarawi: yes, but if you look at the check right before this strcpy, we already use strnlen for the check and error if it's larger. so this strcpy is safe.

src/include/daos/pool.h
mchaarawi: actually pool/srv_cli.c already includes that, so that actually works better. thanks ill update.

src/engine/init.c
mchaarawi: I will update it though to use memcpy to avoid coverity warnings though in case it's not smart enough

mchaarawi:
> I found a few other areas where changes are needed to fully support a non-default system name:
> 
> * System property always displays the default instead of the configured name: https://github.com/daos-stack/daos/blob/master/src/control/lib/daos/system_prop.go#L502

I tried:
```
dmg system get-prop -o=/home/mschaara/config/daos.yml
Name                                             Value   
----                                             -----   
Checksum scrubbing threshold (pool_scrub_thresh) 0       
DAOS system name (daos_system)                   my_srv  
DAOS version (daos_version)                      2.5.101 
Checksum scrubbing mode (pool_scrub_mode)        off 
```
so it might be taken into account already?

> * If the request doesn't supply a system name, the agent uses the built-in default system name to cache system GetAttachInfo responses. This isn't very important -- right now we don't support multi-system access for a single agent. But it should be a fairly simple fix: https://github.com/daos-stack/daos/blob/master/src/control/cmd/daos_agent/infocache.go#L348

we gave up on the idea of 1 agent supporting multiple systems i believe.. but in this case attach info would return the system name anyway so it should not hit that part of the code that you indicated?

kjacque:
> I tried:
> 
> ```
> dmg system get-prop -o=/home/mschaara/config/daos.yml
> Name                                             Value   
> ----                                             -----   
> Checksum scrubbing threshold (pool_scrub_thresh) 0       
> DAOS system name (daos_system)                   my_srv  
> DAOS version (daos_version)                      2.5.101 
> Checksum scrubbing mode (pool_scrub_mode)        off 
> ```
> 
> so it might be taken into account already?

Oops, yes, I found the place where this is updated. Disregard.

> we gave up on the idea of 1 agent supporting multiple systems i believe.. but in this case attach info would return the system name anyway so it should not hit that part of the code that you indicated?

The agent case is looking for a locally-stored value based on the system name requested by the client. If no system name was requested, it tries the built-in default. But yeah, I agree, the single-agent-multi-system case isn't on the table for the foreseeable future, so this isn't important. I'm OK approving as-is.

kjacque (approved):

johannlombardi (approved):

src/engine/init.c
liw: `daos_sysname` was a pointer, it could be assigned the address of a string longer than `DAOS_DEFAULT_SYS_NAME`, is that not so?

liw (approved):

tanabarr (approved):

src/engine/init.c
mchaarawi: yea, i'm not sure then why i made that change :-)
I'll revert it back on my follow on PR.

mchaarawi:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-14170/4/pipeline/267/
was the last test result before accidently restarting.
 the failures there are known: DAOS-15431. DAOS-15662

PR #14172, opened 2024-04-16
DAOS-13151 client: cache and reuse the attach info for the default system

mchaarawi:
improve the daos_init() process to reuse the attach info instead of doing agent drpc upcalls multiple times.

Required-githooks: true

### Gatekeeper:


src/client/api/init.c
```diff
@@ -195,19 +199,29 @@ daos_init(void)
...
+		D_FREE(info);
+		D_GOTO(out_job, rc);
+	}
+	dc_set_global_attach_info(info, resp);
+
 	/**
 	 * get CaRT configuration (see mgmtModule.handleGetAttachInfo for the
 	 * handling of NULL system names)
 	 */
 	rc = dc_mgmt_net_cfg(NULL);
```
liw, line 216: [Question] @mchaarawi, @frostedcmos, does anybody know why we have the `dc_mgmt_net_cfg_check` call below? How does `daos_eq_lib_init` "change net cfg"?

src/client/api/init.c
```diff
@@ -195,19 +199,29 @@ daos_init(void)
...
 
+	D_ALLOC_PTR(info);
+	if (info == NULL)
+		D_GOTO(out_job, rc);
+	rc = dc_get_attach_info(NULL, true, info, &resp);
+	if (rc != 0) {
+		D_FREE(info);
+		D_GOTO(out_job, rc);
+	}
+	dc_set_global_attach_info(info, resp);
```
liw, line 210: [Suggestion] Would it be clearer if we move (the ownership of) the cached "info" and "resp" into `mgmt`? Then `info` and `resp` would no longer need to be global, and the cached "info" and "resp" would be freed when `daos_fini` calls `dc_clear_global_attach_info`.

src/mgmt/cli_mgmt.c
```diff
@@ -943,12 +952,16 @@ sys_attach(const char *name, struct dc_mgmt_sys **sysp)
...
 {
 	if (name == NULL)
 		name = DAOS_DEFAULT_SYS_NAME;
 
-	return sys_attach(name, sysp);
+	/** only use the global attach response if default system (force to re-attach if not) */
+	if (strcmp(name, DAOS_DEFAULT_SYS_NAME) != 0)
+		use_global_resp = false;
+
+	return sys_attach(name, use_global_resp, sysp);
```
liw, line 964: [Question] If a system's name is not "daos_server", this `if` statement would be wrong?

src/client/api/init.c
frostedcmos: From my reading, this has to do with 'FI_OFI_RXM_USE_SRX' (using shared receive queues), as this setting must match between clients and servers. 

daos_eq_lib_init() calls crt_init_opt() which then will do:

 550         if (prov == CRT_PROV_OFI_VERBS_RXM ||
 551             prov == CRT_PROV_OFI_TCP_RXM) {
 552                 /* Use shared receive queues to avoid large mem consumption */
 553                 d_setenv("FI_OFI_RXM_USE_SRX", "1", 0);

this can cause mismatched use_srx setting between clients and servers which dc_mgmt_net_cfg_check() checks.

src/mgmt/cli_mgmt.c
```diff
@@ -56,7 +73,7 @@ dc_mgmt_profile(char *path, int avg, bool start)
 	crt_opcode_t		opc;
 	int			rc;
 
-	rc = dc_mgmt_sys_attach(NULL, &sys);
+	rc = dc_mgmt_sys_attach(NULL, false, &sys);
```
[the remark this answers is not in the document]
mchaarawi: because i didn't want to change the default in this case as it's not on the init path that im trying to improve.

src/mgmt/cli_mgmt.c
mchaarawi: im not sure what you mean by wrong. in any cae, i would like to land the other PR that allows a different sys name than daos_server first, and this PR needs to be adjusted anyway.

src/pool/srv_cli.c
```diff
@@ -67,7 +67,7 @@ dsc_pool_open(uuid_t pool_uuid, uuid_t poh_uuid, unsigned int flags,
 	pool->dp_capas = flags;
 
 	/** attach to the server group and initialize rsvc_client */
-	rc = dc_mgmt_sys_attach(NULL, &pool->dp_sys);
+	rc = dc_mgmt_sys_attach(NULL, false, &pool->dp_sys);
```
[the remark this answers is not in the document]
mchaarawi: because i didn't want to change the default in this case as it's not on the init path that im trying to improve.
so passing false just uses this original behavior.

src/client/api/init.c
mchaarawi: OK, ill look into on the next set of changes and rebase to this PR.

src/include/daos/mgmt.h
```diff
@@ -53,7 +53,8 @@ struct dc_mgmt_sys {
 	struct dc_mgmt_sys_info	sy_info;
 };
 
-int dc_mgmt_sys_attach(const char *name, struct dc_mgmt_sys **sysp);
+int
+	dc_mgmt_sys_attach(const char *name, bool use_global_info, struct dc_mgmt_sys **sysp);
```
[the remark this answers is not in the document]
mchaarawi: im not quite sure i understand you question, so please clarify if my answer does not answer it.
we use the cached info if use_global_info is passed as true (basically in net_cfg, proto query, and pool connect calls).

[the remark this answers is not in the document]
kjacque: Oh sorry. To clarify, I'm just wondering why we would have some code paths use the locally cached version and some not. Why not have them all use the cache? Or have some env variable to make it configurable, if you envision cases where we may not want it cached?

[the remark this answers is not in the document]
mchaarawi: there are some cases where i am not familiar with the code that i chose to keep the default behavior (the server side attach that uses the client stack and the rbt cases). if someone more familiar with the code thinks otherwise, we can update those too and remove this flag. I was just trying to address the ticket where just a simple daos command is triggering 5-6 attach calls to the agent.

src/client/api/init.c
mchaarawi: ok i looked more into this, and dc_get_attach_info() expects the caller to own info and resp. and since we use them in init and fini, i had to make them global here.
I'll try and make this a little more clear.

src/rdb/tests/rdbt.c
```diff
@@ -92,7 +92,7 @@ multi_tests_common_parse(int argc, char *argv[])
 
 	D_ASSERT(g_nreps < g_nranks);
 
-	return dc_mgmt_sys_attach(group_id, &sys);
+	return dc_mgmt_sys_attach(group_id, false, &sys);
```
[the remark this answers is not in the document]
kccain: Yes, I think so.

src/include/daos/mgmt.h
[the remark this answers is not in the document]
kjacque: Ah, understood. From my POV I don't understand why any individual client process would need to query the agent multiple times for the same info. The only element that could change is the suggested interface/domain.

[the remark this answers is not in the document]
mchaarawi: well what im saying that it's not the client only that's using that function. see:
pool/srv_cli.c
rdb/tests/rdbt.c

for the former i think the flag does not matter because we do not call the agent attach anyway since this is a server process. but for the other rdbt test, i do not know. So i think with some review from code owners, this flag might be removed, but for now i added to maintain the original behavior in those cases.

src/mgmt/cli_mgmt.c
mchaarawi: OK I believe this should be fixed now.

src/client/api/init.c
mchaarawi: i reworked this a bit and ultimately did what you suggested with a new dc_mgmt function to populate those global variables that are owned in cli_mgmt.

src/mgmt/cli_mgmt.c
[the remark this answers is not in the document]
mchaarawi: I also have no clue what this mgmt dc function is about. i did not see any usage of it from the code.

src/include/daos/mgmt.h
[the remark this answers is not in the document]
kjacque: @kccain Any thoughts about the rdbt test?

mchaarawi:
@liw @kjacque would appreciate another review. thanks!

src/client/api/init.c
liw: Thanks, @frostedcmos. This is quite eye-opening. :)

src/mgmt/cli_mgmt.c
```diff
@@ -873,21 +872,32 @@ attach(const char *name, struct dc_mgmt_sys **sysp)
...
-	rc = get_attach_info(name, true /* all_ranks */, &sys->sy_info, &resp);
-	if (rc != 0)
-		goto err_sys;
+	if (!resp_g) {
+		need_free_resp = true;
+		rc = get_attach_info(name, true /* all_ranks */, &sys->sy_info, &resp);
+		if (rc != 0)
+			goto err_sys;
+	} else {
+		resp = resp_g;
```
liw, line 881: Do we need to check if `name` is the same with `agent_sys_name`? Not sure if my understanding of `agent_sys_name` is correct.

liw (approved):

src/include/daos/mgmt.h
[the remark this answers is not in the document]
liw: Thanks.

src/client/api/init.c
liw: Thanks. Looks much clearer now.

src/mgmt/cli_mgmt.c
mchaarawi: the problem is that today we only support 1 system per agent. so unless the agent can support multiple systems then this check is not necessary.

src/mgmt/cli_mgmt.c
```diff
@@ -507,13 +524,8 @@ int dc_mgmt_net_cfg(const char *name)
...
 	char                     buf[SYS_INFO_BUF_SIZE];
-	struct dc_mgmt_sys_info  info;
-	Mgmt__GetAttachInfoResp *resp;
-
-	/* Query the agent for the CaRT network configuration parameters */
-	rc = get_attach_info(name, true /* all_ranks */, &info, &resp);
-	if (rc != 0)
-		return rc;
+	struct dc_mgmt_sys_info *info = &info_g;
+	Mgmt__GetAttachInfoResp *resp = resp_g;
```
kjacque, line 528: This is just a suggestion, to take or leave if you like. But I think it would be better to use a function to access these globals rather than having multiple places access them directly. I was thinking, you could reduce some of these additional code changes by having `get_attach_info` check the globals first, and only upcall to the agent if they are absent.

src/mgmt/cli_mgmt.c
kjacque: In theory a host could have multiple agents, though, and a client application may want access to both. IMO if a client has requested a system whose info isn't cached, it's worth it to try fetching it, and see if the name has changed.

kjacque (commented):
I think it would be worth addressing the system name issue, so if anyone is doing anything tricky with multiple agents, we don't silently give them bad information. The other comment is just a suggestion.

kjacque (approved):

kjacque (approved):

liw (dismissed):

mchaarawi:
sorry added gatekeeper after i realized that the PR had a small conflict.

kjacque (approved):

PR #14174, opened 2024-04-16
DAOS-10028 control: Move core types into daos package

mjmac:
The lib/daos package should be the main repository for
types (e.g. PoolInfo) used throughout the codebase. It
should be possible to import these types without pulling
in extra dependencies (e.g. gRPC, protobufs, etc).

Also reverts an inadvertant change to the system
database's serialization that resulted in 2.4/2.6
incompatibilities.

Features: pool
Required-githooks: true
Change-Id: I152742e5d10e689da391fc941f7a9d1afa4e3874


daltonbohning (commented):
ftest LGTM

src/control/system/raft/raft_recovery_test.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/daos-stack/daos/src/control/common/test"
 	. "github.com/daos-stack/daos/src/control/lib/ranklist"
 	"github.com/daos-stack/daos/src/control/logging"
+	"github.com/daos-stack/daos/src/control/system"
 	. "github.com/daos-stack/daos/src/control/system"
```
kjacque, line 30: For clarity, want to get rid of the duplicate import?

src/control/lib/daos/pool.go
```diff
@@ -0,0 +1,246 @@
...
+		PoolLayoutVer    uint32               `json:"pool_layout_ver"`
+		UpgradeLayoutVer uint32               `json:"upgrade_layout_ver"`
+	}
+
+	// StorageTargetUsage represents DAOS target storage usage
+	StorageTargetUsage struct {
+		Total     uint64           `json:"total"`
+		Free      uint64           `json:"free"`
+		MediaType StorageMediaType `json:"media_type"`
+	}
```
kjacque, line 79: Suggestion - how about changing this to `StorageUsage` and embedding it as part of `StorageUsageStats` above? It seems like a subset of the same data, and could be genericized.

kjacque (approved):
Nothing I'd block on. Nice work!

src/control/system/raft/raft_recovery_test.go
mjmac: Ah, good call. I remember using the dot-imports when I moved stuff in order to reduce the reviewer burden, but now seems like a good time to get rid of them.

src/control/lib/daos/pool.go
mjmac: Yeah... I think @kccain added with the `dmg pool query-targets` command. I don't see any reason not to just use the `StorageUsage` struct, though. It will have a few extra fields that aren't used in that case but I think that's fine.

src/control/cmd/dmg/pretty/pool.go
```diff
@@ -270,15 +233,15 @@ func poolListCreateRowVerbose(pool *control.Pool) txtfmt.TableRow {
 
 	row := txtfmt.TableRow{
 		"Label":          label,
-		"UUID":           pool.UUID,
-		"State":          pool.State,
+		"UUID":           pool.UUID.String(),
+		"State":          pool.State.String(),
 		"SvcReps":        svcReps,
-		"Disabled":       fmt.Sprintf("%d/%d", pool.TargetsDisabled, pool.TargetsTotal),
+		"Disabled":       fmt.Sprintf("%d/%d", pool.DisabledTargets, pool.TotalTargets),
```
tanabarr, line 239: why the change from <component><verb> to the other way round? thought we tended to stay with the former?

src/control/lib/daos/pool.go
tanabarr: I had a similar thought last time I visited this. +1

src/control/lib/daos/pool.go
```diff
@@ -0,0 +1,246 @@
...
+	PoolServiceStateDegraded
+	// PoolServiceStateUnknown indicates that the pool service state is unknown
+	PoolServiceStateUnknown
+)
+
+func (pss PoolServiceState) String() string {
+	if pss < PoolServiceStateCreating || pss > PoolServiceStateUnknown {
+		return fmt.Sprintf("Invalid pool service state %d", pss)
+	}
+	return [...]string{
```
tanabarr, line 138: just out of curiosity, why use an array literal rather than just a slice? is there a performance/storage benefit?

src/control/lib/daos/pool.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func (pss *PoolServiceState) UnmarshalJSON(data []byte) error {
+	stateStr := strings.Trim(string(data), "\"")
+
+	state, ok := mgmtpb.PoolServiceState_value[stateStr]
+	if !ok {
+		// Try converting the string to an int32, to handle the
+		// conversion from protobuf message using convert.Types().
+		si, err := strconv.ParseInt(stateStr, 0, 32)
```
tanabarr, line 158: nice, approve of using this method that we've used elsewhere, handles the multiple cases well

tanabarr (approved):
Thanks for taking the time to rework this!

src/control/lib/daos/pool.go
mjmac: OK, I went ahead and moved those types too. While I was doing that, I realized that I'd forgotten to update the daos tool to use the non-dmg pretty printers, so I fixed that.

src/control/lib/daos/pool.go
```diff
@@ -0,0 +1,246 @@
...
+	// PoolServiceStateDegraded indicates that the pool service is in a degraded state
+	PoolServiceStateDegraded
+	// PoolServiceStateUnknown indicates that the pool service state is unknown
+	PoolServiceStateUnknown
+)
+
+func (pss PoolServiceState) String() string {
+	if pss < PoolServiceStateCreating || pss > PoolServiceStateUnknown {
+		return fmt.Sprintf("Invalid pool service state %d", pss)
+	}
```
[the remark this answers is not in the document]
mjmac: As I was looking at these stringers and marshallers, I realized that we were in an in-between state, where some of the types were backed by the protobuf values, and some were just using native values. I started to go down the road of getting rid of all of the protobuf stuff, with the thought that the `lib/daos` package shouldn't depend on the protobufs. I backed away from that position though, because I realized that I was adding maintenance burden to keeping everything in sync.

I still think there's value in revisiting whether or not we should have protobuf dependencies in `lib/daos`, but I don't think this PR is the place to do that.

src/control/cmd/dmg/pretty/pool.go
mjmac: Had to choose one or the other when doing the merge, and the `PoolInfo` names had been around longest.

src/control/lib/daos/pool.go
mjmac: I backed this out in the latest commit in favor of using the protobuf names as in the other stringer methods. As I indicated in my other comment to Kris, I had initially taken the position that we shouldn't have any protobuf dependencies in the `lib/daos` package, but I'm not so sure now. It is certainly less of a maintenance headache to accept the dependency.

src/control/lib/daos/pool.go
```diff
@@ -120,50 +129,53 @@ type PoolServiceState uint
...
-		mgmtpb.PoolServiceState_Unknown.String(),
-	}[pss]
+	return psss
 }
 
 func (pss PoolServiceState) MarshalJSON() ([]byte, error) {
 	return []byte(`"` + pss.String() + `"`), nil
 }
 
+func unmarshalStrVal(inStr string, name2Vals map[string]int32, val2Names map[int32]string) (int32, error) {
```
tanabarr, line 155: Given this is essentially the same approach we are using with `NvmeDevState` and `LedState` do we want to formalise it somehow with a helper so we can update any improvements to be approach in one place?

tanabarr (dismissed):

src/control/lib/daos/pool.go
mjmac: Probably not a bad idea to create a helper in the `common/proto` package. I'd like to do that as a follow-on, though.

src/proto/mgmt/pool.proto
```diff
@@ -229,13 +229,14 @@ message PoolQueryResp {
...
 	uint32 version = 10; // latest pool map version
-	uint32 leader = 11; // current raft leader
+	uint32                     leader             = 11; // current raft leader (2.4)
 	string enabled_ranks = 12; // optional set of ranks enabled
 	string disabled_ranks = 13; // optional set of ranks disabled
 	uint32 total_engines = 14; // total engines in pool
 	uint32 pool_layout_ver = 15; // current pool global version
 	uint32 upgrade_layout_ver = 16; // latest pool global version to upgrade
 	PoolServiceState state = 17; // pool state
+	uint32                     svc_ldr            = 18; // current raft leader (2.6+)
```
kjacque, line 239: Weird indentation on the new fields.

kjacque (dismissed):
LGTM

src/proto/mgmt/pool.proto
mjmac: Ugh, thanks clang-format. I think maybe something got changed because it didn't use to try to mess with the .proto or generated files.

kjacque: Weird, on my system it doesn't seem to do anything to the .proto files.

mjmac:
Unrelated failures:
  * [DAOS-15793](https://daosio.atlassian.net/browse/DAOS-15793): Container destroy times out
  * [DAOS-15774](https://daosio.atlassian.net/browse/DAOS-15774): Failure to destroy container in teardown

mjmac:
@daltonbohning, @kjacque, @tanabarr: Sorry for the churn on this... Should be ready for final re-review now, thanks.

daltonbohning (commented):
It seems we need to update this test as well?
https://github.com/daos-stack/daos/blob/7861ebbaf2fd24ec4d2ab7a1c9fc8f7640196eb5/src/tests/ftest/control/dmg_pool_query_test.py#L69
And we should probably add a `pool` tag, which would have helped you catch this.
These lines:
https://github.com/daos-stack/daos/blob/7861ebbaf2fd24ec4d2ab7a1c9fc8f7640196eb5/src/tests/ftest/control/dmg_pool_query_test.py#L38
https://github.com/daos-stack/daos/blob/7861ebbaf2fd24ec4d2ab7a1c9fc8f7640196eb5/src/tests/ftest/control/dmg_pool_query_test.py#L108
https://github.com/daos-stack/daos/blob/7861ebbaf2fd24ec4d2ab7a1c9fc8f7640196eb5/src/tests/ftest/control/dmg_pool_query_test.py#L155

mjmac:
> It seems we need to update this test as well?
> 
> https://github.com/daos-stack/daos/blob/7861ebbaf2fd24ec4d2ab7a1c9fc8f7640196eb5/src/tests/ftest/control/dmg_pool_query_test.py#L69
> 
Bah. Sigh, will do.

> 
> And we should probably add a `pool` tag, which would have helped you catch this.
> These lines:
> https://github.com/daos-stack/daos/blob/7861ebbaf2fd24ec4d2ab7a1c9fc8f7640196eb5/src/tests/ftest/control/dmg_pool_query_test.py#L38
> 
I'll add it.

daltonbohning (commented):
ftest LGTM

tanabarr (dismissed):

src/control/lib/control/pool.go
```diff
@@ -492,7 +496,15 @@ func (pqr *PoolQueryResp) MarshalJSON() ([]byte, error) {
...
-	return json.Marshal(&aux)
+	auxJSON, err := json.Marshal(&aux)
+	if err != nil {
+		return nil, err
+	}
+
+	// Kinda gross, but needed to merge the embedded struct's MarshalJSON
+	// output with this one's.
+	piJSON[0] = ','
+	return append(auxJSON[:len(auxJSON)-1], piJSON...), nil
```
kjacque, line 507: I'm a little confused. If PoolInfo marshals cleanly, why is it necessary to separate it out and tack it on, rather than including it as part of the aux struct?

mjmac: It's due to one of the quirks of using embedded types. The `MarshalJSON` method on `daos.PoolInfo` is [promoted](https://go.dev/ref/spec#Struct_types), so the logic in `PoolQueryResp`'s `MarshalJSON` is never used.

kjacque (approved):

PR #14176, opened 2024-04-17
DAOS-15276 object: refine sgl valid check

liuxuezhao:
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfs/xattr.c
+++ b/src/client/dfs/xattr.c
@@ -38,6 +38,8 @@ dfs_setxattr(dfs_t *dfs, dfs_obj_t *obj, const char *name, const void *value, da
 		return EINVAL;
 	if (name == NULL)
 		return EINVAL;
+	if (value == NULL && size != 0)
+		return EINVAL;
 	if (strnlen(name, DFS_MAX_XATTR_NAME + 1) > DFS_MAX_XATTR_NAME)
 		return EINVAL;
 	if (size > DFS_MAX_XATTR_LEN)
@@ -142,6 +144,8 @@ dfs_getxattr(dfs_t *dfs, dfs_obj_t *obj, const char *name, void *value, daos_siz
 	if (obj == NULL)
 		return EINVAL;
 	if (name == NULL)
+		return EINVAL;
+	if (value == NULL && size != NULL && *size > 0)
 		return EINVAL;
 	if (strnlen(name, DFS_MAX_XATTR_NAME + 1) > DFS_MAX_XATTR_NAME)
 		return EINVAL;
--- a/src/client/dfs/xattr.c
+++ b/src/client/dfs/xattr.c
@@ -38,8 +38,6 @@ dfs_setxattr(dfs_t *dfs, dfs_obj_t *obj, const char *name, const void *value, da
 		return EINVAL;
 	if (name == NULL)
 		return EINVAL;
-	if (value == NULL && size != 0)
-		return EINVAL;
 	if (strnlen(name, DFS_MAX_XATTR_NAME + 1) > DFS_MAX_XATTR_NAME)
 		return EINVAL;
 	if (size > DFS_MAX_XATTR_LEN)
@@ -64,7 +62,10 @@ dfs_setxattr(dfs_t *dfs, dfs_obj_t *obj, const char *name, const void *value, da
 	iods[0].iod_recxs = NULL;
 	iods[0].iod_type  = DAOS_IOD_SINGLE;
 	iods[0].iod_size  = size;
-	d_iov_set(&sg_iovs[0], (void *)value, size);
+	if (value == NULL)
+		d_iov_set(&sg_iovs[0], NULL, 0);
+	else
+		d_iov_set(&sg_iovs[0], (void *)value, size);
 	sgls[0].sg_nr     = 1;
 	sgls[0].sg_nr_out = 0;
 	sgls[0].sg_iovs   = &sg_iovs[0];
@@ -145,8 +146,6 @@ dfs_getxattr(dfs_t *dfs, dfs_obj_t *obj, const char *name, void *value, daos_siz
 		return EINVAL;
 	if (name == NULL)
 		return EINVAL;
-	if (value == NULL && size != NULL && *size > 0)
-		return EINVAL;
 	if (strnlen(name, DFS_MAX_XATTR_NAME + 1) > DFS_MAX_XATTR_NAME)
 		return EINVAL;
 
@@ -172,7 +171,10 @@ dfs_getxattr(dfs_t *dfs, dfs_obj_t *obj, const char *name, void *value, daos_siz
 		iod.iod_size = *size;
 
 		/** set sgl for fetch */
-		d_iov_set(&sg_iov, value, *size);
+		if (value == NULL)
+			d_iov_set(&sg_iov, NULL, 0);
+		else
+			d_iov_set(&sg_iov, value, *size);
 		sgl.sg_nr     = 1;
 		sgl.sg_nr_out = 0;
 		sgl.sg_iovs   = &sg_iov;
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -2216,6 +2216,18 @@ obj_iod_sgl_valid(daos_obj_id_t oid, unsigned int nr, daos_iod_t *iods,
 				return -DER_INVAL;
 			}
 		}
+		if (sgls != NULL && sgls[i].sg_nr > 0) {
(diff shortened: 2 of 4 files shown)


src/object/cli_obj.c
```diff
@@ -2216,6 +2216,18 @@ obj_iod_sgl_valid(daos_obj_id_t oid, unsigned int nr, daos_iod_t *iods,
...
 			}
 		}
+		if (sgls != NULL && sgls[i].sg_nr > 0) {
+			d_sg_list_t	*sg = &sgls[i];
+			d_iov_t		*iov;
+
+			for (j = 0; j < sg->sg_nr; j++) {
+				iov = sg->sg_iovs + j;
+				if (iov == NULL || (iov->iov_buf_len > 0 && iov->iov_buf == NULL)) {
+					D_ERROR("Bad iov\n");
```
NiuYawei, line 2226: it's better to make the error message more informative, so that one can tell which IOV is bad and why it's bad IOV.

NiuYawei (dismissed):

wangshilong (dismissed):

src/object/cli_obj.c
liuxuezhao: OK, I can refine the log in following PR, thx

liuxuezhao:
@daos-stack/daos-gatekeeper ping, thx

mchaarawi (commented):
please add a proper commit message (PR description) for landing.
please fix the error message "Bad iov". you can do it in same PR or following PR, but it needs to be fix as it's user facing/

src/client/dfs/xattr.c
```diff
@@ -143,6 +145,8 @@ dfs_getxattr(dfs_t *dfs, dfs_obj_t *obj, const char *name, void *value, daos_siz
 		return EINVAL;
 	if (name == NULL)
 		return EINVAL;
+	if (value == NULL && size != NULL && *size > 0)
+		return EINVAL;
```
mchaarawi, line 149: im actually going to request changes for this. if the buffer itself is NULL, there should be no requirement to set the size value. technically this might break existing apps if we return EINVAL now.

src/client/dfs/xattr.c
```diff
@@ -38,6 +38,8 @@ dfs_setxattr(dfs_t *dfs, dfs_obj_t *obj, const char *name, const void *value, da
 		return EINVAL;
 	if (name == NULL)
 		return EINVAL;
+	if (value == NULL && size != 0)
+		return EINVAL;
```
mchaarawi, line 42: same here. if the buffer is NULL, the size is totally ignored and should not be required to be init to 0.

mchaarawi (changes requested):

src/client/dfs/xattr.c
Nasf-Fan: @mchaarawi , if the caller does not know how large the EA (xattr) will be, it needs to call getxattr() with \size parameter as zero for detection firstly, then with the returned size value to allocate buffer and re-call getxattr() with the real buffer size. Under such case, for the first call "value == NULL && size != NULL && *size == 0", is it valid use case or not?

liuxuezhao: @Nasf-Fan the "value == NULL && size != NULL && *size == 0" is allowed parameter, but "value == NULL && size != NULL && *size > 0" is not allowed by this PR.

@mchaarawi why don't allow the case of "value == NULL && size != NULL && *size > 0" is below code in this func -
L175.     d_iov_set(&sg_iov, value, *size);
it will generate a sgl for obj_fetch with one iov "iov.iov_buf = NULL, iov.iov_buf_len = *size (>0)", that IOV cannot be a valid and must be reject by client IO.
I don't think this IOV is a valid input parameter.
If you don't want to check that case inside dfs_getxattr(), it will be rejected by client IO parameter check anyway. Another option is for the case of "value == NULL && size != NULL && *size > 0" do "d_iov_set(&sg_iov, NULL, 0)" at L175.

src/client/dfs/xattr.c
liuxuezhao: same here, for the case of (value == NULL && size != 0), below code at L67 "d_iov_set(&sg_iovs[0], (void *)value, size)" it will generate sgl with IOV of "iov.iov_buf = NULL, iov.iov_buf_len != 0" that cannot be allowed by client IO.

src/client/dfs/xattr.c
Nasf-Fan: > @Nasf-Fan the "value == NULL && size != NULL && *size == 0" is allowed parameter, but "value == NULL && size != NULL && *size > 0" is not allowed by this PR.

If that (value == NULL && size != NULL && *size == 0) is valid use case, then we cannot just check "value == NULL", that is what I want to express.

mchaarawi: I understand what you mean. Im juut worried about this change in behavior with the additional check:
for object daos_obj_fetch, and daos_kv_get, we explicitly say that if the size of the buffer is unknown, one can set to DAOS_REC_ANY and DAOS will return the buffer size.
In DFS, we do not explicitly say that. The user can think that If the buffer is NULL we would get the size (size[IN] is ignored and need not be initialized). if the buffer is not NULL, we would fill up to size and return the size.  

I can see what you mean and how it can mean an issue for client IO. to make sure we don't introduce regressions, please rerun this with this feature tag:
`Features: dfs dfuse datamover`
thanks.

src/object/cli_obj.c
mchaarawi: since a re-run is needed with Features since DFS was changed, please change it with this PR.

src/client/dfs/xattr.c
liuxuezhao: I think that providing a NULL buffer and with an non-zero buffer_size should not be valid parameter for any case.
But I am fine to rerun with "Features: dfs dfuse datamover hdf5" although I don't really think it is necessary TBH.

mchaarawi: > I think that providing a NULL buffer and with an non-zero buffer_size should not be valid parameter for any case.

the problem is users can forget to initialize the size if just passing a NULL buffer especially in an [in,out] parameter, since the size is to be returned. so i disagree with you there. since this is only done on xattr for DFS, it's not going to be a huge change. we need to update the header file to indicate that. I can take care of this later. 

> But I am fine to rerun with "Features: dfs dfuse datamover hdf5" although I don't really think it is necessary TBH.

you touched DFS code, so Features: dfs is definitely required
you also touched cli_obj.c code in a way that can add errors to users that were not returned before.
dfuse uses xattr a lot, so we need to check dfuse layer if it missed setting size to 0.
hdf5-vol uses the object API and can be affected by this for the same reason.
datamover uses dfuse and dfs heavily and can also make the same usage issue
so you might be introducing CI failures in those tests and can affect future landings. it's better to be safe and detect those to make sure that we fix other places / tests and avoid breaking master for everyone else.

liuxuezhao:
@mchaarawi I revert the parameter check in dfs, changed to "if (value == NULL) d_iov_set(&sg_iovs[0], NULL, 0);" that probably is safer. Is it find to you?

mchaarawi:
> @mchaarawi I reverted the parameter check in dfs, changed to "if (value == NULL) d_iov_set(&sg_iovs[0], NULL, 0);" that probably is safer. Is it find to you?

yes thanks for changing that

liuxuezhao:
> > @mchaarawi I reverted the parameter check in dfs, changed to "if (value == NULL) d_iov_set(&sg_iovs[0], NULL, 0);" that probably is safer. Is it find to you?
> 
> yes thanks for changing that

OK, there still with a "-1" left by you, any other concern?

mchaarawi:
> > > @mchaarawi I reverted the parameter check in dfs, changed to "if (value == NULL) d_iov_set(&sg_iovs[0], NULL, 0);" that probably is safer. Is it find to you?
> > 
> > 
> > yes thanks for changing that
> 
> OK, there still with a "-1" left by you, any other concern?

oh sorry, i thought i removed that.

mchaarawi (approved):

src/object/cli_obj.c
```diff
@@ -2216,6 +2216,22 @@ obj_iod_sgl_valid(daos_obj_id_t oid, unsigned int nr, daos_iod_t *iods,
 				return -DER_INVAL;
 			}
 		}
+		if (sgls != NULL && sgls[i].sg_nr > 0) {
```
Nasf-Fan, line 2219: "if (sgls != NULL) {" seems enough.

Nasf-Fan (approved):

PR #14179, opened 2024-04-17
DAOS-623 cq: cancel in progress GHA

daltonbohning:
Group workflows by the workflow and PR number, or the SHA if not a PR.
Then cancel in-progress runs since we only care about the latest.

Skip-test: true
Skip-tests: true
Skip-unit-tests: true

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/.github/workflows/bash_unit_testing.yml
+++ b/.github/workflows/bash_unit_testing.yml
@@ -3,8 +3,9 @@ name: bash Unit Testing
 on:
   pull_request:
 
+# Unique to this workflow and PR, or SHA if not a PR
 concurrency:
-  group: bash-unit-test-${{ github.head_ref  || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
 defaults:
--- a/.github/workflows/ci2.yml
+++ b/.github/workflows/ci2.yml
@@ -3,8 +3,9 @@ name: Build
 on:
   pull_request:
 
+# Unique to this workflow and PR, or SHA if not a PR
 concurrency:
-  group: ci2-${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
 permissions: {}
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -18,6 +18,11 @@ name: clang-format
 on:
   pull_request:
 
+# Unique to this workflow and PR, or SHA if not a PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
 permissions: {}
 
 jobs:
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -9,6 +9,11 @@ on:
       - 'release/*'
   pull_request:
 
+# Unique to this workflow and PR, or SHA if not a PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
 permissions: {}
 
 jobs:
--- a/.github/workflows/rpm-build-and-test.yml
+++ b/.github/workflows/rpm-build-and-test.yml
@@ -3,7 +3,7 @@ name: RPM Build and Test
 env:
   # TODO: we really need to define a list of supported versions (ideally it's no more than 2)
   #       build is done on the lowest version and test on the highest with a "sanity test"
-  #       stage done on all versions in the list ecept the highest
+  #       stage done on all versions in the list except the highest
   EL8_BUILD_VERSION: 8.6
   EL8_VERSION: 8.8
   EL9_BUILD_VERSION: 9
@@ -18,8 +18,9 @@ on:
         required: false
   pull_request:
 
+# Unique to this workflow and PR, or SHA if not a PR
 concurrency:
-  group: rpm-build-and-test-${{ github.head_ref  || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
 defaults:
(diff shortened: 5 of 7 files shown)


daltonbohning:
See this post for background:
https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

And this page GHA references available:
https://docs.github.com/en/actions/learn-github-actions/contexts

.github/workflows/bash_unit_testing.yml
```diff
@@ -3,8 +3,9 @@ name: bash Unit Testing
 on:
   pull_request:
 
+# Unique to this workflow and PR, or SHA if not a PR
 concurrency:
-  group: bash-unit-test-${{ github.head_ref  || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
```
daltonbohning, line 8: The docs for `run_id` say
> A unique number for each workflow run within a repository. This number does not change if you re-run the workflow run.

Which means if a new commit is pushed, the runs for the previous commit are not cancelled because the `group` here is different for each commit.

.github/workflows/ci2.yml
```diff
@@ -3,8 +3,9 @@ name: Build
 on:
   pull_request:
 
+# Unique to this workflow and PR, or SHA if not a PR
 concurrency:
-  group: ci2-${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
```
daltonbohning, line 8: We only run this on pull_request so it's not an issue, but I updated for consistency. If we only use `head_ref` for uniqueness, and if this was ran on landing, then e.g. if we landed two commits to master back-to-back, the second would cancel the workflow for the first. Using `sha` for non pull requests means we maintain uniqueness for landing runs

daltonbohning:
The idea is to cancel in progress runs for the same PR, similar to how Jenkins CI runs are cancelled for a new commit.
We don't have time usage limits, but we are rate limited and IMO we should be responsible with the resources we have.

.github/workflows/clang-format.yml
```diff
@@ -18,6 +18,11 @@ name: clang-format
 on:
   pull_request:
 
+# Unique to this workflow and PR, or SHA if not a PR
```
ashleypittman, line 21: For the linting ones where the run-time is measured in seconds rather than minutes I don't see any point in having this.  It's unlikely to ever trigger, it won't save much in the way of resources when it does and it's extra overhead to maintain.

daltonbohning: Another problem I've ran into is:
- I push some commit A
- GHA is really backed up and takes a while to start some jobs
- I push a new commit B
- GHA linting for commit A runs and reports failures, but it's already been addressed in commit B
- I get confused

daltonbohning: It might be seconds now, but if/when we add longer running checks it will become an issue. I'm trying to proactively trim the GHA before we run into the problem we have with Jenkins runs: waaaaay too much being ran waaay too often and stopping progress

ashleypittman: The longer running checks (or at least ci2) already has this feature enabled.  One thing I don't like with this however is that if you push commit A and then push commit B before GHA has completed which is easy for ci2 (it happens for example if you need to re-push for a linting error) then you get an email about the jobs for the push A being killed which is also confusing.

Jenkins is different as it's the pipeline that's killed, it's all or nothing so you have to kill it.  As I say for longer running jobs in GHA this is a no-brainer but linting jobs as a whole should be under the threshold where there's a benefit.

I think GHA rate-limiting is per person by the way, I've certainly had a couple of days where I've been iterating on GHA triggers and just stopped getting results for hours.

daltonbohning: > if you push commit A and then push commit B before GHA has completed which is easy for ci2 (it happens for example if you need to re-push for a linting error) then you get an email about the jobs for the push A being killed which is also confusing.

Yeah I agree. And unfortunately it's a known issue that hasn't been addressed: https://github.com/orgs/community/discussions/13015

Even if we don't **_need_** it for short jobs, I don't think the maintenance is too much since it's just copy paste across workflows. And I think it's good to not waste the resources. We can't keep adding more GHA without regard for resource usage. The sooner we start addressing that, the easier it is to manage in the future.

ashleypittman: We can't, hence my initial comment about short-running linting jobs only.

This will almost never trigger, and then it does it'll make the user experience worse but save minimal cycles.  The purpose of CI is to use compute resources to assist the developer, saving resources at the cost of developer overhead is not the right tradeoff in this case IMO.

daltonbohning: > it'll make the user experience worse but save minimal cycles

I disagree. Yes, it will almost never trigger. But if it does trigger, it's probably because the workflows are backed up. IMO, the experience is worse if I suddenly get a flood of failed lint checks.

Anyway, I'll remove it for the short jobs since it's not a perfect solution anyway. Mostly because GitHub sends a notification when the workflow is cancelled.

daltonbohning: Oh. This one runs in a few seconds but already had the `cancel-in-progress`. So we're already doing this for some short running jobs

daltonbohning: I guess it also gets confusing to wonder why some jobs get cancelled and some don't. It's just arbitrary at this point

ashleypittman: > Oh. This one runs in a few seconds but already had the `cancel-in-progress`. So we're already doing this for some short running jobs https://github.com/daos-stack/daos/actions/runs/8727893484/job/23946437185

is that the bash one? That runs on our network so won't be subject to scheduling delays and it takes less than 5 seconds.  There's next to zero chance that will ever actually trigger.

We need a threshold and yes, it's arbitrary but we have lots of jobs that run for less than a minute and a couple that run for more than 15 minutes, I'm not aware of any that take between 1 and 15 minutes.

daltonbohning:
Will revisit when needed

PR #14188, opened 2024-04-18
use dependency injection, not globals

mjmac:
Required-githooks: true

Change-Id: I8c179af2ce220e28228e2410e29e873028163958


tanabarr:
nice!

PR #14193, opened 2024-04-19
DAOS-14834 control: Enable parallel server->engine dRPCs

kjacque:
The idea here is to remove the bottleneck in `daos_server` that serializes dRPC calls, to enable `daos_server` to pass along multiple dRPC calls even if the first one hasn't yet returned. In the current master branch, we have a single dRPC client structure that uses RW locks to control access to its internals. dRPC calls that take a long time can potentially impede other commands.

My proposed solution is to create a new `drpc.ClientConnection` for each command that needs to be sent to the `daos_engine`. Each command is handled on its own connection. We were using a connect->send->disconnect pattern on the client connection anyway.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/server/instance_drpc.go
+++ b/src/control/server/instance_drpc.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2023 Intel Corporation.
+// (C) Copyright 2020-2024 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -25,23 +25,29 @@ import (
 )
 
 var (
-	errDRPCNotReady   = errors.New("no dRPC client set (data plane not started?)")
+	errDRPCNotReady   = errors.New("dRPC socket not ready (data plane not started?)")
 	errEngineNotReady = errors.New("engine not ready yet")
 )
 
-func (ei *EngineInstance) setDrpcClient(c drpc.DomainSocketClient) {
+func (ei *EngineInstance) setDrpcSocket(sock string) {
 	ei.Lock()
 	defer ei.Unlock()
-	ei._drpcClient = c
+	ei._drpcSocket = sock
 }
 
-func (ei *EngineInstance) getDrpcClient() (drpc.DomainSocketClient, error) {
+func (ei *EngineInstance) getDrpcSocket() string {
 	ei.RLock()
 	defer ei.RUnlock()
-	if ei._drpcClient == nil {
-		return nil, errDRPCNotReady
+	return ei._drpcSocket
+}
+
+func (ei *EngineInstance) getDrpcClient() drpc.DomainSocketClient {
+	ei.Lock()
+	defer ei.Unlock()
+	if ei.getDrpcClientFn == nil {
+		ei.getDrpcClientFn = drpc.NewClientConnection
 	}
-	return ei._drpcClient, nil
+	return ei.getDrpcClientFn(ei._drpcSocket)
 }
 
 // NotifyDrpcReady receives a ready message from the running Engine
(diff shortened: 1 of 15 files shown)


src/control/server/instance_drpc.go
```diff
@@ -29,28 +29,31 @@ var (
 	errEngineNotReady = errors.New("engine not ready yet")
 )
 
-func (ei *EngineInstance) setDrpcClient(c drpc.DomainSocketClient) {
+func (ei *EngineInstance) setDrpcSocket(sock string) {
 	ei.Lock()
 	defer ei.Unlock()
-	ei._drpcClient = c
+	ei._drpcSocket = sock
```
tanabarr, line 35: nit: don't see a need for the underscore variable prefix

tanabarr (dismissed):
LGTM, any issues in practice?

src/control/server/instance_drpc.go
kjacque: Based on a comment on the struct definition, the reason this was initially done was to mark variables that need to be behind a lock. As this variable subject to the same limitations _drpcClient originally was, I kept the underscore.

src/control/server/instance_drpc.go
```diff
@@ -29,28 +29,31 @@ var (
...
 	ei.RLock()
 	defer ei.RUnlock()
-	if ei._drpcClient == nil {
-		return nil, errDRPCNotReady
+	return ei._drpcSocket
+}
+
+func (ei *EngineInstance) getDrpcClient() drpc.DomainSocketClient {
+	if ei.getDrpcClientFn == nil {
+		ei.getDrpcClientFn = drpc.NewClientConnection
```
mjmac, line 46: Isn't this racy? Could be called from multiple goroutines, I think...

src/control/server/instance_drpc.go
```diff
@@ -29,28 +29,31 @@ var (
...
-	ei._drpcClient = c
+	ei._drpcSocket = sock
 }
 
-func (ei *EngineInstance) getDrpcClient() (drpc.DomainSocketClient, error) {
+func (ei *EngineInstance) getDrpcSocket() string {
 	ei.RLock()
 	defer ei.RUnlock()
-	if ei._drpcClient == nil {
-		return nil, errDRPCNotReady
```
mjmac, line 42: If you're getting rid of this logic, how will we detect/handle requests that are made before the engine has sent its ready upcall? I agree that now is a good time to rework some of this fiddly logic, though.

mjmac (commented):
Interesting to see how this was achieved with relatively minor changes to the logic. I do think that most of the really tricky concerns with concurrent dRPCs have been alleviated by locking at the pool level and serialization of handlers (e.g. PoolCreate). Might be a good opportunity to go even further to streamline and simplify some of this instance management code.

src/control/server/instance_drpc.go
kjacque: Good catch, I'll lock around this.

kjacque: Fixed in the latest push.

src/control/server/instance_drpc.go
kjacque: Fixed in the latest push. I moved this check earlier in the CallDrpc flow.

tanabarr (approved):

kjacque:
Test failures are unrelated.
- DaosCoreTest failure on UCX: https://daosio.atlassian.net/issues/DAOS-15275
- DmgStorageQueryTest failure: https://daosio.atlassian.net/browse/DAOS-15808
- The other one is a DaosCoreTest rebuild timeout, for which we have multiple tickets.

tanabarr:
have we verified that this fixes the customer issue or is that only possible after landing?

liw (approved):
Thanks, @kjacque. AFAICS, it looks good to me.

kjacque:
> have we verified that this fixes the customer issue or is that only possible after landing?

Hasn't been verified (I don't think we have a reliable reproducer for the issue). However at least part of the problem was the server being blocked from further dRPCs to the engine by one dRPC hogging the channel, so that should be resolved by this PR.

kjacque:
@daos-stack/daos-gatekeeper Test failures are documented above. Can this be force-landed please?

mjmac (approved):
LGTM. Nice work, Kris. The unit test looks really solid, too.

PR #14207, opened 2024-04-22
DAOS-15329 cq: Disable debug locking macros for coverity.

ashleypittman:
The coverity tool gets confused about the use of assert in the
debug version of the logging macros and can think a lock is
being unlocked twice which it reports as a API usage error.

Disable the complex macros for coverity to reduce the instances
of false positives in the tool.

Fixes coverity ID 1975167 and others.

### DIFF
--- a/src/include/gurt/common.h
+++ b/src/include/gurt/common.h
@@ -401,7 +401,7 @@ d_realpath(const char *path, char *resolved_path) _dalloc_;
 #define D_SPIN_INIT(x, y)	__D_PTHREAD_INIT(pthread_spin_init, x, y)
 #define D_RWLOCK_INIT(x, y)	__D_PTHREAD_INIT(pthread_rwlock_init, x, y)
 
-#ifdef DAOS_BUILD_RELEASE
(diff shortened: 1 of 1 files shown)


wiliamhuang:
Maybe coverity does not recognize the WRLOCK after unlock.

https://github.com/daos-stack/daos/blob/master/src/client/dfuse/ops/readdir.c#L371

We should set up coverity for local scan to check whether this PR fix the issue or not.

ashleypittman:
> Maybe coverity does not recognize the WRLOCK after unlock.
> 
> https://github.com/daos-stack/daos/blob/master/src/client/dfuse/ops/readdir.c#L371

Yes, coverity thinks that `__D_PTHREAD_TIMED` can return without holding the lock as it appears to have problems with D_ASSERT in that code. https://github.com/daos-stack/daos/blob/master/src/include/gurt/common.h#L368

wiliamhuang (approved):

johannlombardi (approved):

PR #14212, opened 2024-04-22
DAOS-15682 dfuse: Perform reads in larger chunks.

ashleypittman:
When dfuse sees I/O as well-aligned 128k reads then read MB at
a time and cache the result allowing for faster read bandwidth
for well behaved applicaions and large files.

Create a new in-memory descriptor for file contents, pull in a
whole descriptor on first read and perform all other reads from
the same result.

This should give much higher bandwidth for well behaved applications
and should be easy to extend to proper readahead in future.

Test-tag: test_dfuse_caching_check


src/client/dfuse/ops/read.c
```diff
@@ -102,6 +102,334 @@ dfuse_readahead_reply(fuse_req_t req, size_t len, off_t position, struct dfuse_o
...
+
+/* Chunk read and coalescing
+ *
+ * This code attempts to predict application and kernel I/O patterns and preemptively read file
+ * data ahead of when it's requested.
+ *
+ * For some kernels read I/O size is limited to 128k when using the page cache or 1Mb when using
+ * direct I/O.  To get around the performance impact of them detect when well aligned 128k reads
+ * are received and read an entire buffers worth, then for future requests the data should already
+ * be in cache.
```
mchaarawi, line 122: does the limitation of 128k alignment come from fuse?
i would not call only doing 128k reads from an application "well behaved". but im struggling to understand really when is this feature is actually used and what is the benefit.

If app is doing 4k reads, is the kernel reading in 128k blocks and this stores the chunks? or is this code used only if an app is doing 128k aligned IO?

src/client/dfuse/ops/read.c
```diff
@@ -102,6 +102,334 @@ dfuse_readahead_reply(fuse_req_t req, size_t len, off_t position, struct dfuse_o
...
+ *
+ * This code is entered when caching is enabled and reads are correctly size/aligned and not in the
```
mchaarawi, line 124: which test is using this new feature?

src/client/dfuse/ops/read.c
wiliamhuang: Inside chunk_read(),

	if (len != K128)
		return false;
	if ((position % K128) != 0)
		return false;

chunk_read() seems only handles special case of reading length ==128K and offset aligned on 128K.

src/client/dfuse/ops/read.c
```diff
@@ -111,6 +111,333 @@ dfuse_readahead_reply(fuse_req_t req, size_t len, off_t position, struct dfuse_o
...
+	D_MUTEX_LOCK(&rc_lock);
+	if (ie->ie_chunk == NULL) {
+		D_ALLOC_PTR(ie->ie_chunk);
+		if (ie->ie_chunk == NULL)
+			goto err;
+		D_INIT_LIST_HEAD(&ie->ie_chunk->entries);
+	}
+	cc = ie->ie_chunk;
+
+	d_list_for_each_entry(cd, &cc->entries, list)
```
knard-intel, line 371: Storing the buckets in a LRU list seems could have a high search cost when it will contains a lot of entries.
Maybe a structure such as a binary search tree could be more efficient ?

knard-intel (commented):
Mostly looks good to me, from what I understand.
Regarding the limitations notified by @mchaarawi , I have the following newbie questions:
- does user read < 128k, will be done by a 128k read by the kernel ?
- does user read > 128k but <1MB will be serialized with n 128k read by the kernel ?

src/client/dfuse/ops/read.c
```diff
@@ -111,6 +111,333 @@ dfuse_readahead_reply(fuse_req_t req, size_t len, off_t position, struct dfuse_o
...
+	struct read_chunk_data *cd = ev->de_cd;
+	fuse_req_t              req;
+	bool                    done = false;
+
+	cd->rc = ev->de_ev.ev_error;
+
+	if (cd->rc == 0 && (ev->de_len != CHUNK_SIZE)) {
+		cd->rc = EIO;
+		DS_WARN(cd->rc, "Unexpected short read bucket %ld (%#zx) expected %i got %zi",
+			cd->bucket, cd->bucket * CHUNK_SIZE, CHUNK_SIZE, ev->de_len);
```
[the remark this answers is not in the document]
ashleypittman: `DFUSE_REPLY_ERR_RAW` be called on L254 in that case to return an error to the caller.  It needs to be done in the loop as it needs to be called for every slot where the I/O request has been entered.  For slots where the I/O request is made after the callback is complete this handled on L417 so it falls back into the standard read case there rather than returning an error.

src/client/dfuse/ops/read.c
ashleypittman: It could - entry count is key here.  Entries are removed from the list when all slots in a bucket have been read so there should not really be more than one entry per client process outside of some very contrived I/O patterns and it'll only be that high if the app is using single-shard-file and clients are reading different offsets in it.

src/client/dfuse/ops/read.c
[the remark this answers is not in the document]
knard-intel: Make sense (it is what I was suspected).
Thanks for the clarification.

src/client/dfuse/ops/read.c
ashleypittman: Yes.  For reads which go through the page cache then fuse will split the read into 128k chunks and submit them in parallel which is really what this feature is trying to address.

This is about ensuring that dfuse is delivering peak performance for well behaved or larger applications which I'd expect to be issuing I/Os in MB chunks - and in that case this deliveries significant benefit.

If the kernel sees 4k reads then dfuse will see 4k reads and this code won't be activated - the 4k read will be sent to the network.  I've written it in a way that could be extended to that use-case, it would need to use an extent map rather than the current `slots` and I question if an application reading 4k at a time would be as likely to end up reading the entire buffer so that would sit at a different place on the cost/benefit curve.  It would be good to have that as a option but it's probably better for the system not to enable it by default.

src/client/dfuse/ops/read.c
knard-intel: I have to admit, that I am not familiar enough with HPC and AI use cases to know if this unfavorable use case is common.

Just to be sure to understand: if a client read do first aligned read of 128k, and never read the following 7 128K. Then the entry will stay in the cache as long as the client do not close the inode ?

ashleypittman: > Just to be sure to understand: if a client read do first aligned read of 128k, and never read the following 7 128K. Then the entry will stay in the cache as long as the client do not close the inode ?

Close - file handle rather than inode but yes.  The cache is tied to the inode but is dropped when the number of open file handles on that inode drops to zero.

src/client/dfuse/ops/read.c
ashleypittman: Here's the result from a test on this PR:
https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14212/23/artifact/Functional%20Hardware%20Medium/dfuse/caching_check.py/
And the same result against master:
https://build.hpdd.intel.com/job/daos-stack/job/daos/job/weekly-testing/380/artifact/Functional%20Hardware%20Medium/dfuse/caching_check.py/

You can see the read performance in the forth iteration goes from `1317.47` to `3087.97`MiB/s

dfuse can deliver wire-speed on reads but this change will improve the single-client performance and the in-node scaling so higher bandwidth figures will be available with smaller client counts.

src/client/dfuse/ops/read.c
knard-intel: Thanks for the explanation.

knard-intel (approved):
LGTM for what I understand.

src/client/dfuse/ops/read.c
mchaarawi: im not questioning your conclusion that this does provide a better result, but I think you know and you are the one who always says that CI is not valid for looking at performance numbers ;-)

wiliamhuang (approved):

mchaarawi (approved):

src/client/dfuse/ops/read.c
ashleypittman: You're right that this is a major frustration/limitation with CI.  Once you understand the kernel behaviour though and look at what I/O is being requested by dfuse then the potential should be obvious.  I also see performance improvements locally but that's harder to link to here and whilst we can't trust CI numbers entirely a > 100% improvement in bandwidth is not to be disregarded, even if there are large error bars on that.

mchaarawi: again, im not disputing your assessment. 
But i do want to point something obvious that you are missing here. In many applications and middleware on top of POSIX, usually there is always an offset or metadata in the file that breaks that 128k alignment. so "well-behaved" IO can be large and contiguous, but in most cases not 128k aligned.
I did not block your PR here because this is not hurting anything really, but i doubt this provides much benefit beyond some ior runs that are 128k aligned.

ashleypittman: This absolutely won't catch everyone, but it should help for all applications that have put the effort into optimising their I/O patterns.  Lustre for example has been recommending MB size/ MB aligned I/Os for twenty years now.

The more flexible we make this the more applications will potentially benefit but also the greater the cost of incorrectly triggering which will lead to wasted I/O which would be more cost for less benefit.  I view this as a conservative but safe set of requirements today that we can extend, once we have mechanisms in place in order to allow tuning to avoid overburden servers.

mchaarawi: lustre can recommend anything, that does not mean in reality this is what is happening.

mchaarawi: and im definitely disputing your claim that optimizing IO pattern == making it 128k aligned.. this is definitely false IMO

johannlombardi (approved):

PR #14262, opened 2024-04-26
DAOS-15616 test: Update dfuse/find.py to work in a python venv

phender:
Move the self-executed part of dfuse/find.py into its own file.  Also enable running python files setup using the ExecutableCommand with the python executable.

Skip-unit-tests: true
Test-tag: DfuseFind

Required-githooks: true

### Gatekeeper:


src/tests/ftest/dfuse/find.py
```diff
@@ -193,54 +192,58 @@ def _search_needles(file_name, sample_tag, expected_res):
...
         for path in paths:
-            self.log.info("Populating: %s", path)
-            prefix = "t{:05d}_".format(count)
+            dir_tree = DirectoryTreeCommand(self.hostlist_clients)
+            dir_tree.path.value = path
+            dir_tree.height.value = height
+            dir_tree.subdirs.value = subdirs
+            dir_tree.files.value = files_per_node
+            dir_tree.needles.value = needles
+            dir_tree.prefix.value = "t{:05d}_".format(count)
```
daltonbohning, line 224: FYI I think this is the f-string way
```suggestion
            dir_tree.prefix.value = f"t{count:05d}_".
```

src/tests/ftest/directory_tree.py
```diff
@@ -0,0 +1,52 @@
+"""
+  (C) Copyright 2018-2024 Intel Corporation.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+"""
+import sys
+from argparse import ArgumentParser
+
+from util.io_utilities import DirTree
```
daltonbohning, line 9: I'm not sure if these imports will always work in any generic virtual env. But we can cross that bridge when/if needed

src/tests/ftest/directory_tree.py
```diff
@@ -0,0 +1,52 @@
...
+    dir_tree = DirTree(path, height, subdirs_per_node, files_per_node)
+    dir_tree.set_needles_prefix(prefix)
+    dir_tree.set_number_of_needles(needles)
+    tree_path = dir_tree.create()
+    print(f'Directory tree created at: {tree_path}')
+
+
+def main():
+    """Run the program."""
+    parser = ArgumentParser(prog='directory_tree_test.py', description='Create a directory tree')
```
daltonbohning, line 33: `directory_tree_test.py` or `directory_tree.py`?
FWIW: `os.path.basename(__file__)`

src/tests/ftest/directory_tree.py
```diff
@@ -0,0 +1,52 @@
...
+    parser.add_argument('-n', '--needles', type=int, help='number of files in the bottom directory')
+    parser.add_argument('-pr', '--prefix', type=str, help='bottom directory file prefix')
+    args = parser.parse_args()
+
+    try:
+        _populate_dir_tree(
+            args.path, args.height, args.subdirs, args.files, args.needles, args.prefix)
+    except Exception as error:      # pylint: disable=broad-except
+        print(f'Error detected: {str(error)}')
+        sys.exit(1)
```
daltonbohning, line 47: We don't want to let the exception just be raised?

src/tests/ftest/util/command_utils.py
```diff
@@ -167,6 +168,24 @@ def with_exports(self):
...
+
+    @property
+    def with_python_and_env(self):
+        """Get the command string with bind_cores and sudo, and env exports executed by python.
+
+        Returns:
+            str: the command string with bind_cores, sudo, and env exports executed by python
+        """
+        return " ".join([sys.executable, command_as_user(self.with_bind, self.run_user, self.env)])
+
```
daltonbohning, line 188: Do we need to support this? Ideally I think commands just shouldn't set an env if they don't want it. Previous functions to support with and without env are mostly to support existing usage.
I think we could just let `with_python` always include the env

daltonbohning: Another idea is maybe `__init__` could take a `python=True` argument and this can just be rolled into the existing functions? Or... maybe automatically use python if the file extension is `.py`?

src/tests/ftest/util/io_utilities.py
```diff
@@ -10,8 +10,12 @@
 import time
 from logging import getLogger
 
+from command_utils import ExecutableCommand
+from command_utils_base import FormattedParameter
+from exception_utils import CommandFailure
```
daltonbohning, line 15: F401 'exception_utils.CommandFailure' imported but unused

src/tests/ftest/directory_tree.py
phender: I missed this this in the rename, thanks.

src/tests/ftest/util/io_utilities.py
phender: Hmm, that means the githook linting is still not working correctly.

src/tests/ftest/util/command_utils.py
ashleypittman: sys.executable might be `$HOME/venv/bin/python3.6` which is then a symlink to `/usr/bin/python3.6` but on the remote node then the venv probably won't exist - and the env won't be setup for it anyway.  I wonder if we should just us as straight `python` here and let the OS deal with it.

Of course what we're running remotely won't have any access to avocado classes etc.

src/tests/ftest/util/io_utilities.py
daltonbohning: Yeah, the flake hook and some others still need fixing. I'm doing that here #14214

src/tests/ftest/util/command_utils.py
daltonbohning: Luckily we don't need access to avocado, but it's possible something in the future might need some pip-installed package. And I think it's good to make sure we're using the same python. As long as the virtual env is in a shared directory we should be okay

ashleypittman: > As long as the virtual env is in a shared directory we should be okay

Is that an assumption we currently make?  We do already have examples of remote python scripts however so I guess whatever works there is correct here.

daltonbohning: I'm not sure we've though this through enough to know if the assumption is valid. But we could just start making that assumption now :)

daltonbohning: FWIW on systems like Aurora and Endeavour, we put `resnet`, etc. in a shared path

ashleypittman: > '/localhome/jenkins/venv/bin/python3': [Errno 2] No such file or directory\n

Looks like CI doesn't.

src/tests/ftest/util/command_utils_base.py
```diff
@@ -478,6 +479,10 @@ def __init__(self, namespace, command, path=""):
         super().__init__(namespace)
         self._command = command
         self._path = path
+        self._python = None
+        if self.command.endswith('.py'):
+            # Run python scripts with the python command
+            self._python = os.path.basename(sys.executable)
```
ashleypittman, line 485: From the logs this will be `python3`, I suspect `f"python{sys.version_info.major}.{sys.version_info.minor}"` would be safer, with the correct imports of course.

daltonbohning: I'm not sure we can rely on `f"python{sys.version_info.major}.{sys.version_info.minor}"` existing on remote nodes either. E.g. if `python3` is a symlink to a python installed in a non-standard directory, which I have seen before.
IMO the most proper thing is to setup the virtual env on all nodes or put it in a shared directory. Even if `pythonX.Y` exists on the remote node, it possibly doesn't have the necessary pip packages anyway.
But that's more work so I think we should just keep it simple for now and cross that bridge when we need to

ashleypittman: Out of scope for this PR but:

> IMO the most proper thing is to setup the virtual env on all nodes or put it in a shared directory. Even if pythonX.Y exists on the remote node, it possibly doesn't have the necessary pip packages anyway.

When there's a need for these helper/remote python scripts to access that then yes they should.  At that point the front-end node should pip install what it needs, then save the versions to a file and have the remote nodes all install from that file rather than from the requirments-ftest.txt file directly.  On that node CI should probably save that data as an artefact anyway.

A bigger problem for now is that the `util` directory has an identity crisis, it's mostly used by ftest but the user (I think still) needs to set `PYTHONPATH` in order to access it, and that itself messes up pylint testing which we have a hack in place for, then there's are a load of apparent circular dependencies that we have to ignore as well.  Remote python scripts which are run by tests on test nodes should not reside in the same directory in the filesystem tree, that just obfuscates and confuses things further IMO.

daltonbohning: > When there's a need for these helper/remote python scripts to access that then yes they should ...

I agree and think we could leave it at that until we need better handling, since it's only a couple oddball tests that need it anyway. I'm okay with whatever works for now.

daltonbohning: Yeah, we don't really install or setup ftest as a module. It's just a collection of python files that "works"

src/tests/ftest/util/profiler_utils.py
daltonbohning, line 1: What do you think about calling this `util/profiler.py` instead?

daltonbohning (approved):

phender:
In https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14262/12/testReport/ all tests passed, except `test_dfuse_find_perf` in https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14262/12/artifact/Functional%20Hardware%20Medium/dfuse/find.py which is a manual test and appears to require the preexistence of the `challenger_path` mount point to run.  The other test `test_dfuse_find` (a weekly CI test) did pass.

src/tests/ftest/util/profiler_utils.py
phender: I was keeping it generic in case we needed future profilers, but I don't have a major preference.

daltonbohning (approved):

ashleypittman (approved):

PR #14273, opened 2024-04-29
DAOS-14739 pool: Add service-level metrics

mjmac:
Adds a new /svc group under each pool which contains
the following set of metrics:
  * leader (gauge): Current pool service leader rank
  * map_version (counter): Current pool map version
  * open_pool_handles (gauge): Current count of open handles
  * total_ranks (gauge): Number of ranks in pool map
  * degraded_ranks (gauge): Number of ranks with disabled targets
  * total_targets (gauge): Number of targets in pool map
  * disabled_targets (gauge): Number of targets marked disabled
  * draining_targets (gauge): Number of targets in draining state

For non-leader ranks, the service metrics will have zero
values. Telemetry consumers may positively identify the
current leader by checking the value of map_version, which
will always be non-zero for the leader.

Features: pool telemetry
Required-githooks: true
Change-Id: If289250ebf6481b4bcc9783768f5bf07502b52a9


mjmac:
Marking this as ready for review now. The two hw-medium failures are known issues due to the container cleanup bug. I'm assuming that I'll need to make some adjustments based on feedback, so I'll re-run with a merge from master after that.

daltonbohning:
> Marking this as ready for review now. The two hw-medium failures are known issues due to the container cleanup bug. I'm assuming that I'll need to make some adjustments based on feedback, so I'll re-run with a merge from master after that.

Sorry about that. Those two tests are fixed in master now so you shouldn't see the failures again.

src/tests/ftest/telemetry/pool_svc_metrics.py
```diff
@@ -0,0 +1,133 @@
...
+    """Verify pool service metric values.
+
+    :avocado: recursive
+    """
+
+    def collect_svc_telemetry(self, pool_uuid):
+        """Collect the pool service metric values.
+
+        Args:
+            pool_uuid: The UUID of the pool to collect metrics from.
```
daltonbohning, line 25:
```suggestion
            pool_uuid (str): The UUID of the pool to collect metrics from.
```

src/tests/ftest/telemetry/pool_svc_metrics.py
```diff
@@ -0,0 +1,133 @@
...
+        def _map_version(rank_metrics):
+            if rank_metrics is None or MAP_VERSION_METRIC not in rank_metrics:
+                return 0
+            return rank_metrics[MAP_VERSION_METRIC]
+
+        def _pool_rank_metrics(host_metrics, pool_uuid):
+            self.log.debug("collecting rank metrics for pool: %s", pool_uuid)
+            pm = {}
+            for hm in host_metrics.values():
+                for k, v in hm.items():
```
daltonbohning, line 39: We used to have a lint check for this and I'm not sure why it's not flagged anymore, but we should use 3+ characters for variable names. And ideally they should be self-descriptive

src/tests/ftest/telemetry/pool_svc_metrics.py
```diff
@@ -0,0 +1,133 @@
...
+                        continue
+                    for m in v['metrics']:
+                        if 'labels' not in m:
+                            continue
+                        if 'pool' not in m['labels'] or \
+                                m['labels']['pool'].casefold() != str(pool_uuid).casefold():
+                            continue
+                        if 'rank' not in m['labels']:
+                            continue
+                        rank = m['labels']['rank']
```
daltonbohning, line 50: It's somewhat personal preference, but FYI you could do
```suggestion
                        try:
                            if m['labels']['pool'].casefold() != str(pool_uuid).casefold():
                                continue
                            rank = m['labels']['rank']
                        except KeyError:
                            continue
```

src/tests/ftest/telemetry/pool_svc_metrics.py
```diff
@@ -0,0 +1,133 @@
...
+        if failed_ranks:
+            self.fail(f"Rank {restart_rank} didn't stop!")
+
+        self.log_step("Wait for rank exclusion to show up in the telemetry.")
+        metrics = self.collect_svc_telemetry(pool.uuid)
+        while True:
+            if MAP_VERSION_METRIC in metrics and metrics[MAP_VERSION_METRIC] > 1:
+                break
+
+            time.sleep(5)
```
daltonbohning, line 105: Please add a log message that we're sleeping here so the logs don't look like something is hanging (it's been an issue before)

src/tests/ftest/util/telemetry_utils.py
```diff
@@ -486,8 +496,12 @@ def list_metrics(self, hosts=None):
         host_list = hosts or self.hosts
         self.log.info("Listing telemetry metrics from %s", host_list)
         for host in host_list:
-            data = self.dmg.telemetry_metrics_list(host=host)
             info[host] = []
+            try:
+                data = self.dmg.telemetry_metrics_list(host=host)
+            except CommandFailure as e:
+                self.log.error("Failed to list metrics on %s: %s", host, e)
+                continue
```
daltonbohning, line 504:
```suggestion
            except CommandFailure as err:
                self.log.error("Failed to list metrics on %s: %s", host, err)
                continue
```

Also, do we really want to just skip this failure?

src/tests/ftest/util/telemetry_utils.py
```diff
@@ -549,8 +563,12 @@ def get_metrics(self, name, hosts=None):
         host_list = hosts or self.hosts
         self.log.info("Querying telemetry metric %s from %s", name, host_list)
         for host in host_list:
-            data = self.dmg.telemetry_metrics_query(host=host, metrics=name)
             info[host] = {}
+            try:
+                data = self.dmg.telemetry_metrics_query(host=host, metrics=name)
+            except CommandFailure as e:
+                self.log.error("Failed to get metrics for %s: %s", host, e)
```
daltonbohning, line 570:
```suggestion
            except CommandFailure as err
                self.log.error("Failed to get metrics for %s: %s", host, err)
```

src/tests/ftest/telemetry/pool_svc_metrics.py
mjmac: Hmm, good point. I've gotten out of the Python way of thinking. That's cleaner.

src/tests/ftest/telemetry/pool_svc_metrics.py
mjmac: Will do.

src/tests/ftest/util/telemetry_utils.py
mjmac: We're not just skipping, though. We're logging it and proceeding to attempt to collect metrics from other hosts in the hostlist. If the intended default behavior of this function is that it shouldn't succeed unless all hosts return the expected metrics, then I could add a bool to enable this as optional behavior. IMO that's kind of pointless and less flexible, because any callers should be sanity checking the results, and if they care specifically about whether or not every host in the results has data, then they should check for that.

src/tests/ftest/telemetry/pool_svc_metrics.py
mjmac: Will add.

src/tests/ftest/telemetry/pool_svc_metrics.py
daltonbohning: I'm on the fence, so not a blocker. But yes I agree that a shorter scope makes it obvious what they are

src/tests/ftest/util/telemetry_utils.py
daltonbohning: Okay I'm inclined to agree with you. Looks like current usages do verify in some way or would otherwise fail. And obviously if a test was previously getting an exception it should already be failing.

daltonbohning (commented):
ftest LGTM

jolivier23 (approved):

src/pool/srv_pool.c
```diff
@@ -1831,6 +1831,139 @@ pool_svc_check_node_status(struct pool_svc *svc)
...
+		D_GOTO(out, rc);
+	}
+	num_disabled = ranks->rl_nr;
+	d_tm_set_gauge(metrics->degraded_ranks, num_disabled);
+
+	d_rank_list_free(ranks);
+	rc = pool_map_get_ranks(uuid, map, true, &ranks);
+	if (rc != 0) {
+		DL_ERROR(rc, DF_UUID ": failed to get enabled ranks", DP_UUID(uuid));
+		D_GOTO(out_ranks, rc);
```
kccain, line 1871: ranks is freed already on L1867 so maybe here when rc != 0 and I think we can be assured pool_map_get_ranks() did not allocate ranks,  the D_GOTO can jump to the out label instead?

src/pool/srv_pool.c
```diff
@@ -7088,6 +7239,13 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc,
 			DL_INFO(rc, DF_UUID": failed to schedule RF check", DP_UUID(svc->ps_uuid));
 	}
 
+	rc = pool_svc_update_map_metrics(svc->ps_uuid, map,
+					 svc->ps_pool->sp_metrics[DAOS_POOL_MODULE]);
+	if (rc != 0) {
+		DL_ERROR(rc, DF_UUID ": failed to update pool metrics", DP_UUID(svc->ps_uuid));
+		rc = 0; /* not fatal */
```
kccain, line 7246: Should the same be true in the step_up code, that if the call to pool_svc_update_map_metrics() fails there, it would be considered non-fatal?

kccain (changes requested):

src/pool/srv_pool.c
mjmac: I have to admit to a bit of dithering on this due to my unfamiliarity with this area of the codebase. In other areas of the code, there is no defensive programming around access to the per-module metrics. I guess maybe the idea is that it's a "can't happen unless things are really messed up" kind of situation and so there's no point? Seems like I should probably pick one side or the other though, I agree.

src/pool/srv_metrics.c
```diff
@@ -65,6 +65,42 @@ ds_pool_metrics_alloc(const char *path, int tgt_id)
...
+
+	rc = d_tm_add_metric(&metrics->total_ranks, D_TM_GAUGE, "Pool storage ranks (total)", NULL,
+			     "%s/svc/total_ranks", path);
+	if (rc != 0)
+		DL_WARN(rc, "Failed to create pool total_ranks metric");
+
+	rc = d_tm_add_metric(&metrics->degraded_ranks, D_TM_GAUGE, "Pool storage ranks (degraded)",
+			     NULL, "%s/svc/degraded_ranks", path);
+	if (rc != 0)
+		DL_WARN(rc, "Failed to create pool excluded_ranks metric");
```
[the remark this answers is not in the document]
mjmac: Fixed.

src/pool/srv_pool.c
```diff
@@ -1831,6 +1831,139 @@ pool_svc_check_node_status(struct pool_svc *svc)
...
+	rc = pool_map_find_upin_tgts(map, NULL, &num_enabled);
+	if (rc != 0) {
+		DL_ERROR(rc, DF_UUID ": failed to get upin targets", DP_UUID(uuid));
+		D_GOTO(out, rc);
+	}
+	d_tm_set_gauge(metrics->total_targets, num_enabled + num_disabled);
+
+	rc = pool_map_get_ranks(uuid, map, false, &ranks);
+	if (rc != 0) {
+		DL_ERROR(rc, DF_UUID ": failed to get disabled ranks", DP_UUID(uuid));
```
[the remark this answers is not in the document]
mjmac: Fixed.

src/pool/srv_pool.c
mjmac: Ah, right. Missed that when I was moving things around, thanks.

src/pool/srv_pool.c
```diff
@@ -1831,6 +1831,139 @@ pool_svc_check_node_status(struct pool_svc *svc)
...
+		D_GOTO(out, rc);
+	}
+	d_tm_set_gauge(metrics->disabled_targets, num_disabled);
+
+	rc = pool_map_find_upin_tgts(map, NULL, &num_enabled);
+	if (rc != 0) {
+		DL_ERROR(rc, DF_UUID ": failed to get upin targets", DP_UUID(uuid));
+		D_GOTO(out, rc);
+	}
+	d_tm_set_gauge(metrics->total_targets, num_enabled + num_disabled);
```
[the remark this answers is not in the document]
mjmac: Fixed. I changed it to use `pool_map_find_tgts_by_state()` with a state mask of `-1` in order to get the total target count. While I was at it, I also added a gauge for number of targets in the draining state.

src/pool/srv_pool.c
mjmac: At any rate... I just adjusted the other call to `pool_svc_update_map_metrics()` to be non-fatal as well. Also changed the ERRORs to WARNs.

daltonbohning (commented):
ftest LGTM

mjmac:
@jolivier23, @kccain: ping... Sorry for the churn, but should be ready to go again, thanks.

src/pool/srv_pool.c
```diff
@@ -1849,16 +1851,23 @@ pool_svc_update_map_metrics(uuid_t uuid, struct pool_map *map, struct pool_metri
...
+	rc = pool_map_find_tgts_by_state(map, PO_COMP_ST_DRAIN, NULL, &num_draining);
 	if (rc != 0) {
-		DL_ERROR(rc, DF_UUID ": failed to get upin targets", DP_UUID(uuid));
+		DL_ERROR(rc, DF_UUID ": failed to get draining targets", DP_UUID(uuid));
 		D_GOTO(out, rc);
 	}
-	d_tm_set_gauge(metrics->total_targets, num_enabled + num_disabled);
+	d_tm_set_gauge(metrics->draining_targets, num_draining);
+
+	rc = pool_map_find_tgts_by_state(map, -1, NULL, &num_total);
```
kccain, line 1861: I wonder if this could (I'm not 100% sure here without trying it directly) be replaced with a call to pool_map_target_nr(struct pool_map *map) that seems to be able to get the total number of targets in the pool map without having to loop and match on state criteria?

kccain (approved):

jolivier23 (approved):

src/pool/srv_pool.c
mjmac: Will check this and create a followup if it works.

PR #14280, opened 2024-04-29
DAOS-15773 tests: add "--implicit-deps-unchanged" in scons install command in daos build test

wiliamhuang:
To speedup scons install with "--implicit-deps-unchanged" to save CI resource. It decreases this step a little over 10 minutes in local test when libpil4dfs.so is loaded.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/tests/ftest/dfuse/daos_build.py
+++ b/src/tests/ftest/dfuse/daos_build.py
@@ -236,12 +236,7 @@ class DaosBuild(DfuseTestBase):
                 'git -C {} submodule update'.format(build_dir),
                 'python3 -m pip install pip --upgrade',
                 'python3 -m pip install -r {}/requirements-build.txt'.format(build_dir),
-                'scons -C {} --jobs {} --build-deps=only'.format(build_dir, build_jobs),
-                'daos filesystem query {}'.format(mount_dir),
-                'daos filesystem evict {}'.format(build_dir),
-                'daos filesystem query {}'.format(mount_dir),
-                'scons -C {} --jobs {}'.format(build_dir, build_jobs),
-                'scons -C {} --jobs {} install'.format(build_dir, build_jobs),
+                'scons -C {} --jobs {} --build-deps=yes install'.format(build_dir, build_jobs),
                 'daos filesystem query {}'.format(mount_dir)]
         for cmd in cmds:
             command = '{};{}'.format(preload_cmd, cmd)
--- a/src/tests/ftest/dfuse/daos_build.py
+++ b/src/tests/ftest/dfuse/daos_build.py
@@ -236,7 +236,13 @@ class DaosBuild(DfuseTestBase):
                 'git -C {} submodule update'.format(build_dir),
                 'python3 -m pip install pip --upgrade',
                 'python3 -m pip install -r {}/requirements-build.txt'.format(build_dir),
-                'scons -C {} --jobs {} --build-deps=yes install'.format(build_dir, build_jobs),
+                'scons -C {} --jobs {} --build-deps=only'.format(build_dir, build_jobs),
+                'daos filesystem query {}'.format(mount_dir),
+                'daos filesystem evict {}'.format(build_dir),
+                'daos filesystem query {}'.format(mount_dir),
+                'scons -C {} --jobs {}'.format(build_dir, build_jobs),
+                'scons -C {} --jobs {} install --implicit-deps-unchanged'.format(build_dir,
+                                                                                 build_jobs),
                 'daos filesystem query {}'.format(mount_dir)]
         for cmd in cmds:
             command = '{};{}'.format(preload_cmd, cmd)


ashleypittman (changes requested):
Looking at the ticket the aim here is to save CI time however the PR that resolved the issue with building in parallel already changed this test to build in parallel so there is no gain to be had.

Running the deps/daos and install steps separately makes the output clearer, it easier to diagnose steps when they do wrong and us to set more appropriate timeout values for the different stages.

Two calls to `daos filesystem query` are not strictly needed but the `evict` call does provide a lot of additional coverage for fduse (it will cause the whole tree to be forgotten and therefore verify the reference counts on all inodes are correct) so I'd rather keep these.  Ideally `daos filesystem evict` would take a `--wait` option which would negate the need for the second `query` but that's a different issue.

wiliamhuang:
> Looking at the ticket the aim here is to save CI time however the PR that resolved the issue with building in parallel already changed this test to build in parallel so there is no gain to be had.
> 
> Running the deps/daos and install steps separately makes the output clearer, it easier to diagnose steps when they do wrong and us to set more appropriate timeout values for the different stages.
> 
> Two calls to `daos filesystem query` are not strictly needed but the `evict` call does provide a lot of additional coverage for fduse (it will cause the whole tree to be forgotten and therefore verify the reference counts on all inodes are correct) so I'd rather keep these. Ideally `daos filesystem evict` would take a `--wait` option which would negate the need for the second `query` but that's a different issue.

Merging scons commands together can decrease the test time from ~2 hours to 1.5 hours. If you do not think the gain is big enough, I will close this PR then. Thank you!

ashleypittman:
> Merging scons commands together can decrease the test time from ~2 hours to 1.5 hours. If you do not think the gain is big enough, I will close this PR then. Thank you!

Can you justify that with data?  In terms of what the build is doing then it's the same, the only difference will be the new scons process doesn't have the directory tree in cache and I can't believe that takes 30 minutes.  I thought you'd said yesterday that this test took between 2-4 hours due to natural variance anyway?

wiliamhuang:
> > Merging scons commands together can decrease the test time from ~2 hours to 1.5 hours. If you do not think the gain is big enough, I will close this PR then. Thank you!
> 
> Can you justify that with data? In terms of what the build is doing then it's the same, the only difference will be the new scons process doesn't have the directory tree in cache and I can't believe that takes 30 minutes. I thought you'd said yesterday that this test took between 2-4 hours due to natural variance anyway?

https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14280/1/testReport/FTEST_dfuse/DaosBuild/
"1 hr 33 min"

The shortest time I previously observed is ~2 hours.
It seems scons repeats some go applications compiling which takes some time.

ashleypittman:
Ok, this completed now.  Comparing build 1 vs 2 of this PR I see:

B1: combined 90:30
B2: deps: 89:55 build: 55:58 install 2:01

There is no reference weekly test yet to compare against, but from one of my PRs I can see

deps: 62:08 build: 34:11 install 26:05 for a total of 2h 5m.

I think from that we can say there is significant variance anyway as you say, but also that the flag is helping.

scons doesn't run the configure checks when doing --build-deps=yes so adding this flag to the build step would likely not help and with this PR the install step taking 2 minutes is not a significant time so I think this PR is good as-is but further work is needed to understand the variance.

wiliamhuang:
> Ok, this completed now. Comparing build 1 vs 2 of this PR I see:
> 
> B1: combined 90:30 B2: deps: 89:55 build: 55:58 install 2:01
> 
> There is no reference weekly test yet to compare against, but from one of my PRs I can see
> 
> deps: 62:08 build: 34:11 install 26:05 for a total of 2h 5m.
> 
> I think from that we can say there is significant variance anyway as you say, but also that the flag is helping.
> 
> scons doesn't run the configure checks when doing --build-deps=yes so adding this flag to the build step would likely not help and with this PR the install step taking 2 minutes is not a significant time so I think this PR is good as-is but further work is needed to understand the variance.

Thank you!
"install 2:01" looks good since it takes ~12 minutes in my local test without "--implicit-deps-unchanged".
Large variance could be caused by the contention of shared resources on the vm host. Do all nodes running functional tests have similar hardware?

ashleypittman:
> Thank you! "install 2:01" looks good since it takes ~12 minutes in my local test without "--implicit-deps-unchanged".

Yes, this is definitely worth having.

> Large variance could be caused by the contention of shared resources on the vm host. Do all nodes running functional tests have similar hardware?

True, and no, I don't think I they do.  I've not observed such a big variance with the other versions of this test but I have not been actively looking for it.  The only way to tell would be to use Test-repeat and get several instances on the same set test node, or as you're doing test locally.

One other change that might be a factor is that we updated libfuse earlier today, that will enable PARALLEL_DIROPS which won't make any difference if dfuse caching were on but for this test where it isn't then it might have a positive impact on performance.

ashleypittman (approved):

daltonbohning (approved):

PR #14284, opened 2024-04-30
DAOS-15774 test: fix pool_acl and pool_groups

daltonbohning:
Test-tag: SecurityPoolGroupsTest SecurityPoolACLTest
Skip-unit-tests: true

Skip container cleanup in teardown since the pool is destroyed explicitly first.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/tests/ftest/util/pool_security_test_base.py
+++ b/src/tests/ftest/util/pool_security_test_base.py
@@ -326,6 +326,7 @@ class PoolSecurityTestBase(TestWithServers):
             if action.lower() == "write":
                 container = self.get_container(pool, create=False, daos=daos_cmd)
                 result = container.create()
+                container.skip_cleanup()
             elif action.lower() == "read":
                 result = daos_cmd.pool_query(pool.identifier)
             else:


mjmac:
Would it make more sense to make `skip_cleanup()` the default behavior? Seems like there were two commits prior to this one that are adding the `skip_cleanup()` because it's not needed (perhaps due to some other change?). 

To put it another way: If most tests should not be doing the container cleanup, then maybe this should be the new default, because otherwise this is likely to be a recurring problem as tests are added.

I guess another way to approach it would be to make the container destroy in cleanup optional. If the destroy fails due to the container not existing, and it's in the teardown phase, just ignore the error?

daltonbohning:
> Would it make more sense to make `skip_cleanup()` the default behavior? Seems like there were two commits prior to this one that are adding the `skip_cleanup()` because it's not needed (perhaps due to some other change?).
> 
> To put it another way: If most tests should not be doing the container cleanup, then maybe this should be the new default, because otherwise this is likely to be a recurring problem as tests are added.
> 

Most tests should - and do - want to destroy containers in teardown. It's just oddball cases that destroy the container by some other means, or tests that have overly complicated handling, that need `skip_cleanup`


> I guess another way to approach it would be to make the container destroy in cleanup optional. If the destroy fails due to the container not existing, and it's in the teardown phase, just ignore the error?

That's not easy to do automatically because we could end up ignoring real errors.

Part of the trouble here is the codebase has years of tech debt built up and fixing these architectural issues is unfortunately not simple. Most code has been throwing all variables into the `self.*` namespace, which is effectively using global variables for everything.

daltonbohning:
Tests are passing now:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-14284/1/tests

mjean308 (approved):

jamesanunez (approved):

rpadma2:
@daltonbohning  : With this PR, can we address the other two issues mentioned in DAOS-15774 [control/dmg_system_cleanup.py, nvme/enospace.py] ?

PR #14285, opened 2024-04-30
DAOS-15774 test: fix nvme/enospace.py container destroy

daltonbohning:
Test-tag: NvmeEnospace
Skip-unit-tests: true

Track containers created with start_ior_load and delete them explicitly.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/tests/ftest/util/test_utils_container.py
+++ b/src/tests/ftest/util/test_utils_container.py
@@ -43,7 +43,8 @@ def add_container(test, pool, namespace=CONT_NAMESPACE, create=True, daos=None, 
         container.update_params(**params)
     if create:
         container.create()
-    test.register_cleanup(remove_container, test=test, container=container)
+    if container.register_cleanup.value is True:
+        test.register_cleanup(remove_container, test=test, container=container)
     return container
 
 
@@ -370,6 +371,8 @@ class TestContainer(TestDaosApiBase):  # pylint: disable=too-many-public-methods
         self.daos_timeout = BasicParameter(None)
         self.label = BasicParameter(None, "TestContainer")
         self.label_generator = label_generator
+
+        self.register_cleanup = BasicParameter(True)  # call register_cleanup by default
 
         self.container = None
         self.uuid = None
--- a/src/tests/ftest/nvme/enospace.yaml
+++ b/src/tests/ftest/nvme/enospace.yaml
@@ -43,6 +43,7 @@ pool:
 
 container:
   control_method: daos
+  register_cleanup: False  # Skip teardown destroy. Test manually destroys containers.
   type: POSIX
 ior:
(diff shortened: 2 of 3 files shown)


daltonbohning:
Test passed with skipping register_cleanup:
https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14285/4/testReport/

src/tests/ftest/util/test_utils_container.py
```diff
@@ -43,7 +43,8 @@ def add_container(test, pool, namespace=CONT_NAMESPACE, create=True, daos=None,
         container.update_params(**params)
     if create:
         container.create()
-    test.register_cleanup(remove_container, test=test, container=container)
+    if container.register_cleanup.value is True:
```
phender, line 46: I don't think we can properly identify a TestContainer object as not needing to be cleaned up when we instantiate the object.  Some other action needs to be performed - like destroying its parent pool - before we know TestContainer.destroy() should be skipped.  If we:

1. Create a pool
2. Create a container
3. Run ior
4. Destroy the pool
5. Perform additional test steps
6. Enter tearDown()

Having tearDown() call TestContainer.destroy() is valid if the test times out any time after step #2 and before step #4.  In this case we need to notify TestContainer that it should ignore any calls to destroy() either just before step #4 or after it completes successfully.

daltonbohning: If we destroy the pool, all containers in that pool are also destroyed. We don't necessarily need to destroy each container individually for every test case.
And this particular test is doing this to destroy containers:
https://github.com/daos-stack/daos/blob/c7675eb0c55745b3d19b974e44574dcc87211044/src/tests/ftest/nvme/enospace.py#L206

The problem is we don't have a reference to the container after it's created. And the "proper" fix is to refactor the utilities.

There are also generally cases where a test creates A LOT of containers and we don't need to destroy each container individually - it's enough to just destroy the pool. Like this one which creates 100 containers and is also broken right now:
https://github.com/daos-stack/daos/blob/c7675eb0c55745b3d19b974e44574dcc87211044/src/tests/ftest/nvme/pool_capacity.py#L224

daltonbohning: If the pool destroy has already been registered, it's not necessary to register container destroy and then just call skip_cleanup. IMO having to call `container.skip_cleanup` is more of a hack than not registering the cleanup in the first place

phender (approved):

daltonbohning:
Modified test passed, and arbitrarily IorSmall for shared utility verification:
https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14285/8/testReport/

dinghwah (approved):

PR #14289, opened 2024-04-30
DAOS-8331 telemetry: Support client metrics dump without agent

mjmac:
Enable scenarios where client telemetry is collected and dumped
to a CSV without agent config changes or involvement.

Setting D_CLIENT_METRICS_DUMP_DIR in the client process
environment will enable client telemetry dump to the
specified directory even if the agent is not configured
to export telemetry.

Features: telemetry
Required-githooks: true
Change-Id: I243d11a2e00059ef3115d392d63c523048477122

### DIFF
--- a/src/client/api/metrics.c
+++ b/src/client/api/metrics.c
@@ -21,13 +21,20 @@
 #include <gurt/telemetry_producer.h>
 
 #define INIT_JOB_NUM 1024
-bool daos_client_metric        = false;
-bool daos_client_metric_retain = false;
+bool daos_client_metric;
+bool daos_client_metric_retain;
 
 #define MAX_IDS_SIZE(num) (num * D_TM_METRIC_SIZE)
 /* The client side metrics structure looks like
  * root/job_id/pid/....
  */
+
+static int
+shm_key(pid_t pid)
+{
+	/* Set the key based on our pid so that it can be easily found. */
+	return pid - D_TM_SHARED_MEMORY_KEY;
+}
 
 static int
 shm_chown(key_t key, uid_t new_owner)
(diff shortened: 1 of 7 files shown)


mjmac:
Moving this out of draft for review. Note, NLT seems to be having issues with this PR. The actual tests pass, but the stage is marked as failed due to an error in calculating deltas. I wound up pushing with `Skip-NLT: true` out of frustration, because I wanted to get the rest of the test coverage.

Given that NLT actually passed and the failure does not seem to be related to changes made in this PR, I don't think it's worth re-running again to try and get an NLT pass.

tanabarr (approved):

src/client/api/metrics.c
```diff
@@ -60,35 +67,43 @@ shm_chown(key_t key, uid_t new_owner)
...
+	/* If the metrics will not be retained, don't register them with the agent. */
+	if (!daos_client_metric_retain)
+		goto skip_agent;
+
 	/* Request that the agent adds our segment into the tree. */
 	rc = dc_mgmt_tm_register(NULL, dc_jobid, pid, &agent_uid);
 	if (rc != 0) {
-		DL_ERROR(rc, "client telemetry setup failed.");
+		if (rc == -DER_UNINIT && d_isenv_def(DAOS_CLIENT_METRICS_DUMP_DIR)) {
+			D_INFO("telemetry dump dir set -- proceeding without agent management.\n");
```
kccain, line 91: would this be appropriate to classify as a warning / D_WARN? Since the user has requested metrics to be retained but it can't be done?

kccain (approved):

src/client/api/metrics.c
mjmac: I considered this, but decided that it would be too noisy/alarming. The intent behind this PR is to allow a user to dump client telemetry without any involvement from the admin (i.e. no need to reconfigure the agent, etc). 

Rather than requiring the user to set yet another environment variable to indicate that they want telemetry to work even if the agent isn't configured to manage it, I just modified the code to quietly handle this special case (-DER_UNINIT indicates that the agent hasn't initialized the telemetry library). To me, this follows the principle of least surprise (or annoyance, depending on your outlook).

The exact use cases here are still being hashed out, but in case it's not clear, my expectation is that if the user is asking to dump the client telemetry at program exit, then they probably don't care about whether or not the agent is configured to manage the telemetry. Agent management is only needed if the telemetry will be sampled in realtime while the application is still running.

mjmac:
*grumble* ... Just merged master, re-running.

kjacque (approved):
LGTM

PR #14293, opened 2024-05-01
DAOS-15685 control: Make failure clear when mountpt goes missing

tanabarr:
In the case that a normally persistent mountpoint directory goes
missing, e.g. after a catastrophic failure, the control log is not
clear about the reason for the mount failure. The message indicating
that the superblock is missing is particularly confusing and could
result in unnecessary data loss if the admin decides they need to wipe
PMem as a result.

Fix by stopping engine start-up if PMem device has ext4 filesystem
signature and the target mount-point location doesn't exist. The
engine process will not be started in this situation and a clear fault
description and resolution message will be displayed. No "waiting for
metadata format" log entry will be printed.

Features: control

### Gatekeeper:


src/control/server/storage/provider.go
```diff
@@ -185,11 +185,15 @@ func (p *Provider) ControlMetadataIsMounted() (bool, error) {
 		return false, errors.New("nil provider")
 	}
 
-	p.log.Debugf("control metadata config: %+v", p.engineStorage.ControlMetadata)
 	if !p.engineStorage.ControlMetadata.HasPath() {
 		// If there's no control metadata path, we use SCM for control metadata
-		return p.ScmIsMounted()
+		if isMounted, err := p.ScmIsMounted(); os.IsNotExist(errors.Cause(err)) {
+			return false, nil // Mountpoint maybe missing
```
knard-intel, line 191: Some logging message could be useful here

knard-intel (approved):
Overall PR LGTM.
NIT, a unit test could has been added to cover this new behavior.

src/control/server/storage/provider.go
tanabarr: TBH I don't think we need extra logging here, the aim of the change is to prevent a missing mountpoint from being confused as a missing superblock. This case will indicate that the mountpoint is not mounted and further down the chain the path existence will be checked explicitly.

kjacque: Instead of returning nil here, IMO we should return a more informative, helpful error. Masking the error with the assumption that it gets caught later doesn't seem like the right approach to me.

kjacque (changes requested):
I don't feel comfortable with returning nil at this lower level -- IMO it assumes too much about what will happen later in the flow, which it doesn't control. I think it would be better to report a new error (instead of wrapping the error returned).

src/control/server/storage/provider.go
tanabarr: My thinking behind the change is that `ControlMetadataIsMounted()` shouldn't really be worried about whether the mountpoint exists or not, either it is mounted or not. If the mountpoint doesn't exist then it's simply not mounted so that should be the return result. Returning NotExist in the context of the IsMounted check doesn't make sense to me. I don't think this is masking the error, I see it as making the function output more intuitive.

knard-intel (dismissed):
LGTM

mjmac (dismissed):

tanabarr:
> I don't feel comfortable with returning nil at this lower level -- IMO it assumes too much about what will happen later in the flow, which it doesn't control. I think it would be better to report a new error (instead of wrapping the error returned).

After off-line discussion we agreed at looking to terminate the server process on engine exit in this case. I had a look and I'm not sure it's a sensible thing to do. The `server/instance_exec.go` and `server/harness.go` code has evolved to enable the engines to be handled pretty much independently of each other and the fault we will generate from the missing mount directory case will be returned from `awaitStorageReady()->format()->startRunner()`. In theory we could add logic inside of `handleExit()` that caused a cancellation in the parent context for a specific fault but that feels like it's going against the notion of independent engine instances as well as being a special case.

The situation where both engines are running, one crashes and `dmg system stop` is used to restart the dead rank would also cause the running rank to be killed through server exit in the case that the crashed rank was missing a mount directory. To avoid that scenario we would need to do some checking as to the state of the other engine when deciding whether to percolate the engine error to the server or not. 

In summary I think we should limit the solution to causing a fault to be raised and exit engine instance startup when the mount directory missing. @kjacque @mjmac thoughts?

kjacque:
To re-iterate the offline discussion, from my POV, it is confusing to leave the server running and asking for a metadata format when, in fact, the admin needs to either:

1) (to keep existing data) Shut down the daos_server, create the mountpoint, and restart to bring everything back up in a mode that doesn't ask for a format.
OR 
2) (to destroy existing data and start over) Perform a full format, which would clobber the existing pmem data and create the mountpoint.

We probably don't want to bring either engine up if we detect a situation like this, I don't think. Is there a legitimate reason why someone would want to fix it later (letting another rank on the node start running, doing IO, etc.) rather than fixing it immediately?

tanabarr:
I agree that exiting engine start in this case with a descriptive fault would be helpful.

startRunner() is the function that gets triggered each time an engine instance is started (either on service start or dmg system start) and this calls format()->awaitStorageReady() is the blocking function that evaluates the state of the storage, and is where we are concerned about.

I'm working on identifying and refusing that special case during awaitStorageReady(), this does mean that any engine with fs and missing mount target/dir will exit before reaching "format required" prompt.

kjacque (dismissed):
I don't have access to a pmem-enabled system to verify what the output looks like, but I like this approach better. Thanks.

tanabarr:
> I don't have access to a pmem-enabled system to verify what the output looks like, but I like this approach better. Thanks.

Updated PR results in the following output: https://daosio.atlassian.net/browse/DAOS-15685?focusedCommentId=128178

knard-intel (approved):

mjmac (commented):
Looks pretty good. Just a quibble about the complex logic in `checkScmNeedFormat()`.

mjmac (approved):
LGTM, thanks

kjacque (approved):

PR #14318, opened 2024-05-06
DAOS-13151 client: cache and reuse the attach info for the default system

mjmac:
Contains the following backports:
  - DAOS-15655 control: fix support for non default system name (#14170)
  - DAOS-13151 client: cache and reuse the attach info for the default system (#14172)


mjmac:
@techbasset: FYI, I'd like to get this merged before we do the weekly update.

jolivier23 (approved):

PR #14337, opened 2024-05-09
mjmac/DAOS 15810

mjmac:
- **DAOS-15810 control: Rework daos_agent AttachInfo handling**
- **DAOS-15810 control: Rework daos_agent AttachInfo handling**


mjmac:
The hw-medium-verbs failure appears to be a typical rebuild suite timeout. This PR doesn't change anything having to do with rebuild or pools. Waiting for feedback from @johannlombardi on whether or not the approach in this PR seems to help with the problem he is seeing.

PR #14367, opened 2024-05-14
DAOS-14834 control: Enable parallel server->engine dRPCs (#14193)

mjmac:
The idea here is to remove the bottleneck in daos_server that serializes
dRPC calls, to enable daos_server to pass along multiple dRPC calls even
if the first one hasn't yet returned. In the current master branch, we
have a single dRPC client structure that uses RW locks to control access
to its internals. dRPC calls that take a long time can potentially impede
other commands.

My proposed solution is to create a new drpc.ClientConnection for each
command that needs to be sent to the daos_engine. Each command is handled
on its own connection. We were using a connect->send->disconnect pattern
on the client connection anyway.

Required-githooks: true
Change-Id: Ibe5a03d28ecc5099b5827ef22fbbace9e3d8b963
Signed-off-by: Kris Jacque <kris.jacque@intel.com>

### DIFF
--- a/src/control/cmd/daos_agent/security_rpc_test.go
+++ b/src/control/cmd/daos_agent/security_rpc_test.go
@@ -73,7 +73,7 @@ func setupTestUnixConn(t *testing.T) (*net.UnixConn, func()) {
 	return newConn, cleanup
 }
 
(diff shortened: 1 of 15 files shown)


kjacque (approved):
LGTM

PR #14381, opened 2024-05-15
DAOS-15849 control: Add client uid map to agent config

mjmac:
Allow daos_agent to optionally handle unresolvable client
uids via custom mapping. In deployments where the agent
may not have access to the same user namespace as client
applications (e.g. in containerized deployments), the
client_user_map can provide a fallback mechanism for
resolving the client uids to known usernames for the
purpose of applying ACL permissions tests.

Example agent config:

    credential_config:
      client_user_map:
        default:
          user: nobody
          group: nobody
        1000:
          user: joe
          group: blow

Features: control
Required-githooks: true
Change-Id: I72905ccc5ddee27fc2101aa4358a14e352c86253


src/control/lib/control/pool_test.go
```diff
@@ -704,9 +702,6 @@ func TestControl_PoolCreate(t *testing.T) {
 			ctx := test.Context(t)
 			mi := NewMockInvoker(log, mic)
 
-			if tc.req.userExt == nil {
-				tc.req.userExt = mockExt
-			}
```
kjacque, line 709: Huh, so the test wasn't actually using the mocks for anything?

src/control/security/auth/auth_sys.go
```diff
@@ -146,67 +68,188 @@ func sysNameToPrincipalName(name string) string {
...
+		for i, gID := range groupIds {
+			g, err := req.getGroupFn(gID)
+			if err != nil {
+				return nil, err
+			}
+			groupNames[i] = g.Name
+		}
+
+		return groupNames, nil
+	}
```
kjacque, line 130: suggestion - To test the internals of these functions, I'd suggest breaking them out as normal named methods, and keeping the mockable interface as thin as possible.

kjacque (approved):
I wouldn't block on any of my suggestions. This is pretty cool. Nice work.

I find myself wondering if there might be some benefit to making use of interface types for this scenario. Having two different data sources for user/group/etc., it seems like a good fit for the concept.

src/control/cmd/daos_agent/security_rpc_test.go
```diff
@@ -188,3 +197,125 @@ func TestAgentSecurityModule_RequestCreds_BadUid(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			log, buf := logging.NewTestLogger(t.Name())
+			defer test.ShowBufferOnFailure(t, buf)
+
+			// Set up a real unix socket so we can make a real connection
+			conn, cleanup := setupTestUnixConn(t)
+			defer cleanup()
+
+			mod := NewSecurityModule(log, tc.secCfg)
+			mod.signCredential = func() func(req *auth.CredentialRequest) (*auth.Credential, error) {
```
tanabarr, line 298: Why `func() sig {...} ()` Rather than just `sig {...}` here and specify `var idx int` at top-level inside `t.Run()`?

src/control/lib/control/pool_test.go
tanabarr: looks like it was until this change https://github.com/daos-stack/daos/pull/14381/files#diff-b86b7f9f25785e10ce878665ccc27f1a331e16ab733a0ff7729b7e404d0d659fR53

src/control/security/auth/auth_sys.go
```diff
@@ -146,67 +68,188 @@ func sysNameToPrincipalName(name string) string {
...
+	return stripHostName(name), nil
+}
+
+type (
+	// CredentialRequest defines the request parameters for GetSignedCredential.
+	CredentialRequest struct {
+		DomainInfo      *security.DomainInfo
+		SigningKey      crypto.PrivateKey
+		getHostnameFn   func() (string, error)
+		getUserFn       func(string) (*user.User, error)
```
tanabarr, line 91: nit: would be nice to have aliases for these function signatures

tanabarr (commented):
Are we sure this change won't enable any undesired security consequences outside container environments? The prospect of nefarious mappings seems like something we really should think about, but perhaps it has? Is the assumption that the mappings can only be applied by an administrator who has write access to the configuration files and there aren't any other vectors that we need to be concerned about?

Also the Commit message seems to me to imply that the changes are only required for testing purposes: "...client_user_map can provide a fallback mechanism for resolving the client uids to known usernames for the purpose of applying ACL permissions tests.". I assume that this is a typo (or me misunderstanding) and the changes are indeed needed for production deployments?

kjacque:
> Are we sure this change won't enable any undesired security consequences outside container environments? The prospect of nefarious mappings seems like something we really should think about, but perhaps it has? Is the assumption that the mappings can only be applied by an administrator who has write access to the configuration files and there aren't any other vectors that we need to be concerned about?

I thought about this for a bit, too. As long as the admin secures their DAOS config files and certificates from unauthorized users, I think we should be OK. If the admin has secured the system with certificates with appropriate permissions (private key readable by daos_agent only), attackers will not be able to start their own private daos_agent to interface with the system with a custom config either. I see these as the two main attack vectors. @dpquigl can you think of anything else?

tanabarr (dismissed):
Concerns assuaged by @dpquigl 's comments elsewhere:
> So if I understand this correctly you have a multi-container environment where the socket for daos-agent is exported from one container and used in another. So when you check the sender’s credentials on the socket you’re getting he euid and egid from the process in the other container. However since the daos-agent container probably only has the users needed for the applications running in the container you can’t do a lookup on the users and groups. This seems like a reasonable solution. NFSv4 has a similar mechanism where it will translate a local uid to a name and back into a local uid on the remote side because of just this issue. The agent is a trusted entity in the security architecture so I agree that as long as you keep the agent integrity its ok for it to take on the additional responsibility of performing the translation. Is the reason this is in the same config file just for convenience? It might be a good idea to have it in a separate file for manageability.

I think moving the new section to a separate file is probably something we don't have bandwidth for currently so I'm happy to merge this now.

mjmac:
> Are we sure this change won't enable any undesired security consequences outside container environments? The prospect of nefarious mappings seems like something we really should think about, but perhaps it has? Is the assumption that the mappings can only be applied by an administrator who has write access to the configuration files and there aren't any other vectors that we need to be concerned about?
> 
As Kris & David have pointed out, the agent is considered a trusted component, and only the admin should be able to modify its configuration. As long as the agent trust is maintained, then there's no concern that malicious mappings could be created. If the agent trust is broken (due to file permissions or whatever), then pretty much anything could happen.

> Also the Commit message seems to me to imply that the changes are only required for testing purposes: "...client_user_map can provide a fallback mechanism for resolving the client uids to known usernames for the purpose of applying ACL permissions tests.". I assume that this is a typo (or me misunderstanding) and the changes are indeed needed for production deployments?
>

Perhaps there's a better term, but I was using "test" in the sense that an access request is tested by applying the ACL and either accepting or rejecting the request. Maybe "check" would have been clearer.

daltonbohning:
FYI a fix for the DEB build just landed to master

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -57,17 +74,37 @@ func (m *SecurityModule) getCredential(session *drpc.Session) ([]byte, error) {
...
 		return m.credRespWithStatus(daos.BadCert)
 	}
 
-	cred, err := auth.AuthSysRequestFromCreds(m.ext, info, signingKey)
+	req := auth.NewCredentialRequest(info, signingKey)
+	cred, err := m.signCredential(req)
 	if err != nil {
-		m.log.Errorf("%s: failed to get AuthSys struct: %s", info, err)
-		return m.credRespWithStatus(daos.MiscError)
+		if err := func() error {
```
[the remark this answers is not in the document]
mjmac: If the initial signing fails, then we have a number of checks to do before attempting to re-sign with the mapped user and group. If any of those steps fail, then we want to log that the signing failed and return a wrapped error.

I could have broken this out into a new named function (e.g. `resignWithUserMap()` or something like that), but then the reader would have to jump to a different spot in the code to follow that flow, whereas with this approach it's just a natural extension of the signing logic.

src/control/cmd/daos_agent/security_rpc_test.go
mjmac: I've changed the code to use the `credSignerFn` type in the return signature to make it clearer that the closure is generating a function. The index variable is scoped to the generator closure and the generated closure.

kjacque (approved):

PR #14382, opened 2024-05-15
DAOS-15540 cart: Change proto query default timeout

jolivier23:
Set it to 5 seconds so we can try a different target more quickly if a rank is down.

Features: rebuild

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/cart/crt_register.c
+++ b/src/cart/crt_register.c
@@ -629,9 +629,13 @@ crt_proto_query_int(crt_endpoint_t *tgt_ep, crt_opcode_t base_opc, uint32_t *ver
 	proto_query->pq_user_arg = arg;
 	proto_query->pq_coq->coq_base = base_opc;
 
-	rc = crt_req_set_timeout(rpc_req, timeout);
-	/** Should only fail if invalid parameter */
-	D_ASSERT(rc == 0);
+	if (timeout != 0) {
+		if (timeout < crt_gdata.cg_timeout) {
+			rc = crt_req_set_timeout(rpc_req, timeout);
+			/** Should only fail if invalid parameter */
+			D_ASSERT(rc == 0);
+		}
+	}
 
 	rc = crt_req_send(rpc_req, proto_query_cb, proto_query);
 	if (rc != 0)
--- a/src/cart/crt_register.c
+++ b/src/cart/crt_register.c
@@ -579,6 +579,7 @@ crt_proto_query_int(crt_endpoint_t *tgt_ep, crt_opcode_t base_opc, uint32_t *ver
 	struct crt_proto_query_in	*rpc_req_input;
 	struct proto_query_t		*proto_query = NULL;
 	uint32_t			*tmp_array = NULL;
+	uint32_t                         default_timeout;
 	int				 rc = DER_SUCCESS;
 
 	if (ver == NULL) {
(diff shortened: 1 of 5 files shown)


wangshilong (approved):

src/cart/crt_register.c
```diff
@@ -629,7 +629,7 @@ crt_proto_query_int(crt_endpoint_t *tgt_ep, crt_opcode_t base_opc, uint32_t *ver
 	proto_query->pq_user_arg = arg;
 	proto_query->pq_coq->coq_base = base_opc;
 
-	rc = crt_req_set_timeout(rpc_req, 5);
+	rc = crt_req_set_timeout(rpc_req, timeout);
```
frostedcmos, line 632: I would prefer here if timeout passed is 0 to not crt_req_set_timeout() at all. This way callers like crt_rpc.c can then call with 0 and use whatever defaults were already set via CRT_TIMEOUT instead of having to specify 60 seconds (for fi/ctl proto queries below).

mchaarawi: you mean timeout passed from the cart tests, but not the daos proto query from init? i agree in that case that passing 0 should just be a no-op and not do anything in this case.

src/client/api/rpc.c
```diff
@@ -114,8 +115,10 @@ query_cb(struct crt_proto_query_cb_info *cb_info)
 
 	if (daos_rpc_retryable_rc(cb_info->pq_rc)) {
 		rproto->ep.ep_rank = (rproto->ep.ep_rank + 1) % rproto->nr_ranks;
+		if (rproto->timeout < 30)
```
mchaarawi, line 118: it would be better here that instead of 30 to use the default timeout that cart was inited with. not sure if there is an api to get default timeout from cart, but should be easy to add one and use that instead of 30 and if current timeout exceeds the default to just use the default.

src/cart/crt_rpc.c
```diff
@@ -441,8 +441,7 @@ crt_register_proto_fi(crt_endpoint_t *ep)
 	if (rc != 0)
 		return -DER_MISC;
 
-	rc = crt_proto_query(ep, cpf.cpf_base, &cpf.cpf_ver,
-			     1, crt_pfi_cb, &pfi);
+	rc = crt_proto_query(ep, cpf.cpf_base, &cpf.cpf_ver, 1, 60, crt_pfi_cb, &pfi);
```
mchaarawi, line 444: would be better to pass 0 here and in cart we just not set the timeout of the value passed is 0.

mchaarawi (dismissed):

mchaarawi (approved):

daltonbohning (commented):
ftest LGTM

src/cart/crt_register.c
```diff
@@ -629,9 +629,13 @@ crt_proto_query_int(crt_endpoint_t *tgt_ep, crt_opcode_t base_opc, uint32_t *ver
 	proto_query->pq_user_arg = arg;
 	proto_query->pq_coq->coq_base = base_opc;
 
-	rc = crt_req_set_timeout(rpc_req, timeout);
-	/** Should only fail if invalid parameter */
-	D_ASSERT(rc == 0);
+	if (timeout != 0) {
+		if (timeout < crt_gdata.cg_timeout) {
```
frostedcmos: i would remove this line, as crt_gdata.cg_timeout isnt necessarily the one being used either if per-rpc timeout isnt set. the order is -- global timeout (crt_gdata.cg_timeout), per-context timeout, per-rpc timeout.

src/client/api/rpc.c
frostedcmos: nope there isnt a single/simple way, as there are multiple timeouts involved (global, per context, per rpc).

src/cart/crt_register.c
jolivier23: So we should use the crt_req_get_timeout() instead to get the default?

frostedcmos (approved):

mchaarawi (approved):

wangshilong (approved):

PR #14395, opened 2024-05-17
DAOS-9247: Add PCIe link speed and width to NVMe health stats

tanabarr:

### Gatekeeper:


tanabarr:
https://daosio.atlassian.net/browse/DAOS-15959 NLT memcheck errors in CI run 15

src/control/server/instance_storage_rpc.go
```diff
@@ -163,6 +166,63 @@ func (ei *EngineInstance) StorageFormatSCM(ctx context.Context, force bool) (mRe
 	return
 }
 
+func setLnkStats(log logging.TraceLogger, inStr string, health *ctlpb.BioHealthResp) error {
```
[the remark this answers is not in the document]
tanabarr: you want to send the protobuf struct into a method in the hardware library to get populated? that doesn't really sound like a good fit to me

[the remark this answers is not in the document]
tanabarr: The use of reflect seems appropriate here as it can be used to avoid specifying each health stat explicitly in a switch case when scanning through the input text (lspci output).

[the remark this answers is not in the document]
tanabarr: I've attempt to address your comments by moving to the hardware library and commenting on the use of reflect.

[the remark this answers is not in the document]
mjmac: > I've attempt to address your comments by moving to the hardware library and commenting on the use of reflect.

OK, I think I'm not communicating effectively. I am saying that the approach you've got here, even after moving the function into the hardware library, is very narrowly focused on this specific use case of decorating a protobuf message in a gRPC handler before returning it to dmg.

The code works now for this specific use case, sure. What about down the road when we have a different use case where we'd like to key some decisions based off of the link speed or link status? How would we reuse this code you're adding in order to accomplish that? As far as I can tell, it would be pretty difficult. The developer would have to know about this new `RunLspciWithInput()` function and figure out how to use it. It takes a string and returns a string, so they'd have to figure out what those look like and how to generate/interpret them.

And then we've got this `SetPciLinkStats()` function which accepts a string and a protobuf message. In order to use this, a developer has to figure out what the input string looks like or where it comes from. At the moment, I guess you'd have to copy whatever is going on in `addLinkInfoToHealthStats()`? Why should another developer (or you, after a few months of looking at other stuff) have to go through and figure all of this out? Wouldn't it be nicer to just call into the hardware library and get a list of [PCIDevice](https://github.com/daos-stack/daos/blob/master/src/control/lib/hardware/pci.go#L438)s with all of this info populated on them?

The solution as it is now accomplishes the immediate task, but we really should be thinking in terms of libraries and reuse. A change to the logic for getting PCI link speed should ideally not result in changes outside of the hardware library, if we've done it correctly. It's more work up-front, but it yields dividends down the road as the codebase grows and more developers are working with it.

I'm willing to be convinced that what I'm asking for is impossible or unreasonable, but I don't think it is.

src/control/server/instance_storage_rpc_test.go
```diff
@@ -18,13 +18,154 @@ import (
...
+	"github.com/daos-stack/daos/src/control/provider/system"
 	"github.com/daos-stack/daos/src/control/server/config"
 	"github.com/daos-stack/daos/src/control/server/engine"
 	"github.com/daos-stack/daos/src/control/server/storage"
 	"github.com/daos-stack/daos/src/control/server/storage/bdev"
 	"github.com/daos-stack/daos/src/control/server/storage/scm"
 )
 
+func TestIOEngineInstance_populateCtrlrHealth(t *testing.T) {
+	mockLspciOut := `01:00.0 Non-Volatile memory controller: Intel Corporation PCIe Data Center SSD (rev 01) (prog-if 02 [NVM Express])
```
kjacque, line 30: I think I get what Mike is saying here. I don't think the server package should need to know about how to parse lspci output.

Perhaps in the lib/hardware library there could be some definitions like:
```go
type DeviceLinkStats struct {
        Speed uint64 // maybe specify units?
        Capabilities []string
        // etc... any other useful fields
}

type DeviceLinkStatsProvider interface {
        GetDeviceLinkStats(device string) (*DeviceLinkStats, error)
}
```

And then create an lspci provider that implements this one function. It keeps all the lspci output string parsing down at that low level. Then at all the upper levels, we can interpret and format them as we please. Also if we need to use some other data source later it could be easily changed. It would also be easier to mock.

src/control/server/instance_storage_rpc.go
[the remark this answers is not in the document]
tanabarr: Thanks for the detailed explanation I will try again.

src/control/server/instance_storage_rpc_test.go
tanabarr: I didn't want to go into that detail within this PR, i.e. parsing each key/val within all of the PCIe config space link parameters. Providing the string value for each of the Lnk* lspci output lines in a health stat seemed like an efficient way of meeting the requirements of the single customer for something that's not particular high priority. To me this feels like over engineering but I will proceed with the consensus.

src/control/common/env_utils.go
```diff
@@ -154,3 +154,21 @@ func MergeKeyValues(curVars []string, newVars []string) (merged []string) {
 
 	return
 }
+
+// AppendToEnv appends entries to the supplied colon-separated string if non-existent.
+func AppendToEnv(envVal string, toAdd ...string) string {
```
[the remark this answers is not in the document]
tanabarr: agreed, changed

src/control/server/instance_storage_rpc.go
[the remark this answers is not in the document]
tanabarr: By implementing the approach you are prescribing we lose the brevity and simplicity benefits of the current solution (using reflect to directly parse lspci into BioHealthResp). Forcing delegation into the hardware module (obviously there is bias for using a module one has authored) may be cleaner from one perspective but causes unnecessary bloat in this solution IMO and indirection (passing config space string received over dRPC, decoding individual parameters from an already decoded and formatted lspci output, populating PCIDevice structs, copying fields to a BioHealthResp struct) and the promise of future benefits doesn't warrant the level of change requested. In my mind this is an example of premature optimisation. However as this is the consensus I will disagree and commit...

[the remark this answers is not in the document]
tanabarr: Having looked again there is at least one significant drawback to the suggested approach - decoding the config space bytes through lspci then picking apart the text generated to convert the "humanised" metrics to raw types is nonsensical and prone to error if lspci code changes (unlikely) or some output format is not handled correctly by us. The ideal solution would be to provide a go implementation of PCIe config space decoder (or implement on the engine side which I did take a look into) but this to me is not within the scope of effort to address the minor customer ask. I would like to ask that we accept the current changes and look to spend more effort on a fully fledged PCIDevice implementation when more feature development work is required.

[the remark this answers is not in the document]
mjmac: Frankly, I'm not convinced that the approach of shelling out to lspci (which may or may not be installed on a system (you haven't added it as a dependency in the RPM specfile or debian control file) is great to begin with. I played around a little with creating a binding for pciutils under `lib/hardware` and it wasn't too bad to work with. I don't have time to flesh that out myself, though.

I understand that you're just trying to quickly satisfy a customer request, but the proposed approach adds technical debt. Quick solutions here and there pile up into a bad codebase and ultimately slow down velocity because changes are hard to make when they cut across multiple layers and it's difficult to figure out how to use code that wasn't designed for testability and reuse.

src/control/server/instance_storage_rpc_test.go
kjacque: An intermediate approach could be to have the link stats fields be strings, but still create the lib/hardware framework around the implementation. Give the fields the same names as the protobuf fields for easy conversion. That would create a framework that allows for future refinement if we ever need to customize display or otherwise use these fields at a higher level.

My main concern is having the server package parsing stuff from lspci. If it ever becomes a problem it will be harder to understand/replace if assumptions about using lspci are sprinkled around the code and unit tests.

src/control/server/instance_storage_rpc.go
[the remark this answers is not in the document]
tanabarr: do you have a branch for the pciutils bindings?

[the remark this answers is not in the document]
mjmac: > do you have a branch for the pciutils bindings?

Extremely WIP to play with the idea: https://github.com/daos-stack/daos/pull/14521

src/control/lib/hardware/pci.go
```diff
@@ -590,3 +591,39 @@ func (d PCIDevices) Get(addr *PCIAddress) []*PCIDevice {
...
+	}
+
+	for _, line := range strings.Split(inStr, "\n") {
+		toks := strings.Split(strings.TrimSpace(line), ":")
+		if len(toks) != 2 {
+			continue // Not valid key:val entry.
+		}
+		lineKey := strings.TrimSpace(toks[0])
+		if !strings.HasPrefix(lineKey, "Lnk") {
+			continue // Bail early to reduce chance of false-positive lookups.
```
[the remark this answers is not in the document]
tanabarr: "Bail" was intentional :-)

src/control/lib/hardware/pci.go
```diff
@@ -590,3 +591,39 @@ func (d PCIDevices) Get(addr *PCIAddress) []*PCIDevice {
...
+		}
+
+		field := healthVal.FieldByName(lineKey)
+		// Check that matching field exists in health stats resp and that field is settable
+		// (exported) and of type string before attempting to set it.
+		if !field.IsValid() || !field.CanSet() || field.Kind() != reflect.String {
```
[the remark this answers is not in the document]
tanabarr: the new approach will add only negotiated and maximum speeds and widths

tanabarr:
@kjacque @mjmac @knard-intel have revised the PR to use pciutils bindings, converted to draft as I still need to remove some surplus codes now using bindings and fix unit tests but I would appreciate early review feedback on this approach. TIA

src/control/lib/hardware/pci.go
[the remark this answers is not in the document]
knard-intel: My apologies... still have to work on my english :-)

src/proto/ctl/smd.proto
```diff
@@ -112,6 +116,7 @@ message NvmeController {
 	LedState led_state = 10;		// NVMe device LED state
 	string pci_dev_type = 11;		// PCI device type, vmd or pci
 	string vendor_id = 12;			// controller's vendor ID
+	string             pci_cfg      = 13;                  // PCIe configuration space
```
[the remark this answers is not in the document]
knard-intel: Did not know that proto files were linted: sounds awkward.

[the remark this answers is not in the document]
tanabarr: agreed

src/control/server/instance_storage_rpc.go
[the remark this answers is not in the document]
tanabarr: have integrated

src/proto/ctl/smd.proto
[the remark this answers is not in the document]
kjacque: There's a ticket against the proto and generated file linting, by the way. It only started happening recently for me, not sure why.

src/control/lib/hardware/hwloc/bindings.go
```diff
@@ -554,7 +554,8 @@ func (o *object) linkSpeed() (float64, error) {
 	if pciDevAttr == nil {
 		return 0, errors.Errorf("device %q attrs are nil", o.name())
 	}
-	return float64(pciDevAttr.linkspeed), nil
+	// Attribute in Giga units, return raw transactions per sec.
+	return float32(pciDevAttr.linkspeed) * 1e+9, nil
```
mjmac, line 558: nit: The multiplier should be defined as a const.

src/control/lib/hardware/pci.go
```diff
@@ -514,8 +519,8 @@ func (b *PCIBus) IsZero() bool {
 
 func (d *PCIDevice) String() string {
 	var speedStr string
-	if d.LinkSpeed > 0 {
-		speedStr = fmt.Sprintf(" @ %.2f GB/s", d.LinkSpeed)
+	if d.LinkNegSpeed > 0 {
+		speedStr = fmt.Sprintf(" @ %.2f GT/s", d.LinkNegSpeed/1e+9)
```
mjmac, line 523: The divisor should probably use the same named const as the multiplier elsewhere. Or, probably better, define a `pciutils.LinkSpeedString()` function or something similar, as it will encapsulate responsibility for knowing these details in one place.

src/control/lib/hardware/topology.go
```diff
@@ -353,8 +353,8 @@ func (t *Topology) Merge(newTopo *Topology) error {
 							oldDev.Type = newDev.Type
 						}
 
-						if oldDev.LinkSpeed == 0 {
-							oldDev.LinkSpeed = newDev.LinkSpeed
+						if oldDev.LinkNegSpeed == 0 {
+							oldDev.LinkNegSpeed = newDev.LinkNegSpeed
 						}
```
mjmac, line 358: Just a reminder that you'll need to handle the new fields here, too.

src/control/lib/hardware/pciutils/bindings.go
```diff
@@ -0,0 +1,132 @@
...
+	case 5:
+		return float32(tera) * 32
+	case 6:
+		return float32(tera) * 64
+	default:
+		return 0
+	}
+}
+
+func (api *api) PCIDeviceFromConfig(cfgBytes []byte, dev *hardware.PCIDevice) error {
```
mjmac, line 65: Looking at how we're using this now... I'd suggest either fleshing this out to fully-populate the incoming PCIDevice from the config, or, alternatively, renaming this to something like `PCICapsFromConfig`. I'm kind of inclined toward the latter, as I don't think we need a completely different way to populate a device. At least not for the moment.

mjmac (commented):
Looks much better to me, thanks. I know this seemed like a bit of scope creep, but I think it will serve the codebase better in the long run to do it this way.

src/bio/bio_device.c
```diff
@@ -438,19 +450,29 @@ fetch_pci_dev_info(struct nvme_ctrlr_t *w_ctrlr, const char *tr_addr)
 		return -DER_INVAL;
 	}
 
+	w_ctrlr->pci_cfg    = NULL;
+	w_ctrlr->pci_cfg_sz = 0;
+	if (fetch_pci_cfg) {
+		D_ALLOC(w_ctrlr->pci_cfg, NVME_PCI_CFG_SPC_MAX_LEN);
+		if (w_ctrlr->pci_cfg == NULL)
+			return -DER_NOMEM;
+		w_ctrlr->pci_cfg_sz = NVME_PCI_CFG_SPC_MAX_LEN;
```
NiuYawei, line 459: Looks this 'pci_cfg_sz' isn't useful?

src/bio/bio_device.c
```diff
@@ -438,19 +450,29 @@ fetch_pci_dev_info(struct nvme_ctrlr_t *w_ctrlr, const char *tr_addr)
 		return -DER_INVAL;
 	}
 
+	w_ctrlr->pci_cfg    = NULL;
+	w_ctrlr->pci_cfg_sz = 0;
+	if (fetch_pci_cfg) {
+		D_ALLOC(w_ctrlr->pci_cfg, NVME_PCI_CFG_SPC_MAX_LEN);
```
NiuYawei, line 456: Minor: looks to me this D_ALLOC should be done in alloc_ctrlr_info() along with other allocations, and this extra 'fetch_pci_cfg' parameter could be removed.

NiuYawei (commented):
The BIO changes look good to me.

src/bio/bio_device.c
tanabarr: correct isn't used anymore and forgot to remove it, done

src/bio/bio_device.c
tanabarr: My concern was I didn't want to be fetching this for standard Scans only when health is requested. If you think the overhead isn't worth worrying about then I can remove the fetch parameter and always populate it regardless of the scan type?

tanabarr: removed adding fetch parameter

NiuYawei: Sorry, I didn't make it clear. I meant if you move the 'pci_cfg' allocation from this function to alloc_ctrlr_info(), then you don't need this extra 'fetch_pci_cfg' parameter for this function.

But I'm fine with totally removing this 'fetch_pci_cfg' too.

src/control/lib/hardware/hwloc/bindings.go
tanabarr: done

src/control/lib/hardware/pciutils/bindings.go
tanabarr: done

src/control/lib/hardware/topology.go
tanabarr: done

src/bio/bio_device.c
tanabarr: I removed the flag so now the config space is populated on each dRPC scan but only processed in the control-plane if a health flag is specified in gRPC request.

src/control/lib/hardware/pci.go
tanabarr: used `humanize` library to decide the right text

utils/rpms/daos.spec
```diff
@@ -146,6 +147,7 @@ Requires: ndctl
 # needed to set PMem configuration goals in BIOS through control-plane
 %if (0%{?suse_version} >= 1500)
 Requires: ipmctl >= 03.00.00.0423
+Requires: pciutils
```
brianjmurrell, line 150: This is superfluous with line 162.

```suggestion
```

utils/ansible/ftest/vars/Rocky8.yml
```diff
@@ -72,6 +73,8 @@ daos_server_deps:
     - libuuid
     - lz4-libs
     - ndctl
+    - pciutils
+    - pciutils-libs
```
brianjmurrell, line 77: This is implied by the line above it.

```suggestion
```

ci/unit/required_packages.sh
```diff
@@ -26,6 +26,8 @@ pkgs="argobots                         \
       numactl-devel                    \
       openmpi$OPENMPI_VER              \
       patchelf                         \
+      pciutils                         \
```
brianjmurrell, line 29: Not needed.  Implied by below.

```suggestion
```

debian/control
```diff
@@ -34,6 +34,7 @@ Build-Depends: debhelper (>= 10),
                liblz4-dev,
                libaio-dev,
                libcapstone-dev
```
brianjmurrell, line 36: You need a `,` after this item.

utils/rpms/daos.spec
brianjmurrell: @tanabarr Please don't resolve conversations I start.  I use them to validate my outstanding requests and will resolve them when I see that they are fixed correctly.

brianjmurrell (dismissed):

brianjmurrell (commented):
Packaging changes LGTM.

src/control/lib/hardware/pciutils/bindings.go
```diff
@@ -0,0 +1,120 @@
...
+#cgo LDFLAGS: -lpci
+
+#include <stdlib.h>
+#include <pci/pci.h>
+*/
+import "C"
+
+type api struct{}
+
+func (api *api) Cleanup() {}
```
kjacque, line 27: If we don't expect `Cleanup` to be used, no need to define it in the interface, in my opinion.

src/control/lib/hardware/pciutils/bindings.go
```diff
@@ -0,0 +1,120 @@
...
+	access.method = C.PCI_ACCESS_DUMP
+
+	tmpFile, err := os.CreateTemp("", "pciutils")
+	if err != nil {
+		return err
+	}
+	fName := tmpFile.Name()
+	defer os.Remove(fName)
+
+	cfgBytes = append([]byte("01:00.0 device #1\n"), cfgBytes...)
```
kjacque, line 73: This is a little confusing to me. Why do we need to tack this on? Are this address and device number always right?

src/control/lib/hardware/pciutils/bindings.go
```diff
@@ -0,0 +1,120 @@
...
+	case 6:
+		mant = 64
+	default:
+		return 0
+	}
+
+	return mant * 1e+9
+}
+
+func (api *api) PCIeCapsFromConfig(cfgBytes []byte, dev *hardware.PCIDevice) error {
```
kjacque, line 52: Don't forget to add doc comments to any new exported functions

src/control/lib/hardware/pciutils/provider.go
```diff
@@ -0,0 +1,64 @@
...
+type (
+	accessProvider interface {
+		PCIeCapsFromConfig(cfgBytes []byte, dev *hardware.PCIDevice) error
+		Cleanup()
+	}
+)
+
+func Init(parent context.Context) (context.Context, error) {
+	return context.WithValue(parent, AccessKey, &api{}), nil
+}
```
kjacque, line 31: Is there some way to avoid needing the caller to init/fini the `Context`? Especially considering cleanup doesn't do anything at the moment.

src/control/lib/hardware/pciutils/provider.go
```diff
@@ -0,0 +1,64 @@
...
+func Fini(ctx context.Context) {
+	ap, err := accessFromContext(ctx)
+	if err != nil {
+		return
+	}
+
+	ap.Cleanup()
+}
+
+func PCIeCapsFromConfig(ctx context.Context, cfgBytes []byte) (*hardware.PCIDevice, error) {
```
kjacque, line 51: Hmm, I don't mind a utility function necessarily, but I was thinking of something similar to the other lib/hardware providers, a struct with methods that could be mocked, rather than requiring higher level unit tests to know something about the (non-exported) `accessProvider` interface.

src/control/server/server_utils.go
```diff
@@ -828,3 +828,26 @@ func checkFabricInterface(name string, lookup ifLookupFn) error {
...
 	return nil
 }
+
+// Convert bytestring to format accepted by lspci, 16 bytes per line.
+func formatBytestring(in string, sb *strings.Builder) {
+	if sb == nil {
+		return
+	}
+	for i, s := range in {
+		rem := i % 32
```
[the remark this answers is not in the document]
tanabarr: "remainder", constant added

src/control/lib/hardware/hwloc/provider.go
```diff
@@ -298,7 +298,7 @@ func (p *Provider) getPCIDevsPerNUMANode(topo *topology, nodes hardware.NodeMap)
 		numaID := p.getDeviceNUMANodeID(osDev, topo)
 
 		var addr *hardware.PCIAddress
-		var linkSpeed float64
+		var linkSpeed float32
```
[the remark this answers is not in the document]
tanabarr: protobuf float is 32-bits, link related values will not exceed max of 3.4028237 × 10^38

src/control/lib/hardware/pciutils/bindings.go
tanabarr: this was added by the initial implementation. removed

src/control/lib/hardware/pciutils/bindings.go
tanabarr: This preamble is expected by the library when reading config from file (lspci "-F" option). The actual address is irrelevant because we are just extracting from the config content, we don't read the address. Without an address and device values in the preamble, the library will refuse to parse the file content.

tanabarr: I've decided that the preamble should be applied to the config content before calling into the library. The library itself shouldn't be concerned with altering the config bytes input. The preamble is now applied in the server code before calling into the library. I've added some more comments to improve clarity.

src/control/lib/hardware/pciutils/bindings.go
tanabarr: fixed

src/control/lib/hardware/pciutils/provider.go
tanabarr: we could revert @mjmac 's provider style of implementation but I don't feel that it's a particular problem. maybe I can visit that as a follow-up if it is deemed important.

src/control/provider/system/distribution.go
```diff
@@ -187,3 +190,21 @@ func getDistribution(open openFunc) Distribution {
...
+func GetLspciPath() (string, error) {
+	distro := GetDistribution()
+
+	switch distro.ID {
+	case "opensuse-leap", "opensuse", "sles":
+		envPath, ok := os.LookupEnv("PATH")
+		if !ok {
+			return "", errors.New("couldn't find PATH env")
+		}
+		if err := os.Setenv("PATH", common.AppendToPathEnv(envPath, "/sbin")); err != nil {
```
knard-intel, line 204: Could be more safe to add this path at the beginning of the PATH environment variable.

src/control/server/instance_storage_rpc.go
```diff
@@ -392,20 +438,31 @@ func smdQueryEngine(ctx context.Context, engine Engine, pbReq *ctlpb.SmdQueryReq
...
+	}
 
-	listDevsResp, err := listSmdDevices(ctx, engine, new(ctlpb.SmdDevReq))
+	scanSmdResp, err := scanSmd(ctx, engine, &ctlpb.SmdDevReq{})
 	if err != nil {
-		return nil, errors.Wrapf(err, "rank %d", engineRank)
+		return nil, errors.Wrapf(err, "rank %d; scan smd", engineRank)
+	}
+	if scanSmdResp == nil {
+		return nil, errors.New("nil smd scan resp")
```
knard-intel, line 455: The rank info could also be useful ?

knard-intel (dismissed):
Mostly OK for me for what I understand, just some minor details.

src/control/pbin/exec.go
```diff
@@ -149,7 +150,7 @@ func ExecReq(parent context.Context, log logging.Logger, binPath string, req *Re
 	// ensure that /sbin and /usr/sbin are in child's $PATH
 	for i, env := range child.Env {
 		if strings.HasPrefix(env, "PATH=") {
-			child.Env[i] = env + ":/sbin:/usr/sbin"
+			child.Env[i] = common.AppendToPathEnv(env, "/sbin", "/usr/sbin")
```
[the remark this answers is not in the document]
tanabarr: I don't really want to change the behaviour here in this PR

src/control/server/instance_storage_rpc.go
tanabarr: done

src/control/provider/system/distribution.go
tanabarr: @mjmac @kjacque any thoughts on this? I will change if there is a consensus.

src/control/lib/hardware/pci.go
```diff
@@ -514,14 +518,19 @@ func (b *PCIBus) IsZero() bool {
 
 func (d *PCIDevice) String() string {
 	var speedStr string
-	if d.LinkSpeed > 0 {
-		speedStr = fmt.Sprintf(" @ %.2f GB/s", d.LinkSpeed)
+	if d.LinkNegSpeed > 0 {
```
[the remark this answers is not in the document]
tanabarr: I don't want to bloat the string representation for PCIDevice, the full link details including maximum and negotiated values will be reported in health which I think is the correct place to expose both values.

src/proto/ctl/smd.proto
```diff
@@ -73,6 +73,12 @@ message BioHealthResp {
 	uint64 cluster_size = 46;		// blobstore cluster size in bytes
 	uint64 meta_wal_size = 47;		// metadata WAL blob size
 	uint64 rdb_wal_size = 48;		// RDB WAL blob size
+	// PCIe config space link stats
```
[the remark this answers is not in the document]
tanabarr: done

src/control/provider/system/distribution.go
kjacque: Often I've seen the value added to the front of the PATH to ensure the right binary is picked up. That said, I don't have strong feelings about it, particularly in this case.

knard-intel (dismissed):

src/control/lib/hardware/pciutils/provider.go
tanabarr: reworking...

src/control/lib/hardware/pciutils/provider.go
tanabarr: reworking...

tanabarr:
reworking to address @kjacque 's comments about using the hardware lib interface model, converted back to draft

tanabarr:
The `PCIeLinkStatsProvider` interface approach is nice in many ways but from a top-level testing perspective there isn't a clean way to inject a mock provider dependency. The implemented approach works because by default nothing happens unless the `PciCfg` parameter is populated in the `ctrlr` struct. We probably need an `Engine` interface method to retrieve a hardware provider, like we do for the storage provider. Then mocking can be done by supplying a mock hardware provider in top-level (`ControlService` gRPC handler) server tests.

tanabarr:
I have been considering the error handling question. I was also concerned that some of the library function calls e.g. [`pci_scan_bus()`](https://github.com/pciutils/pciutils/blob/master/lib/pci.h#L103) don't return an RC and would probably result in a failed RPC if something unexpected happened. The mitigation against these concerns is that we are relying on the well tested SPDK and libpci APIs and simply passing output from one to the input of another and the likelihood of an incomplete or malformed output being produced is minimal. 

Given this thinking, would you like extra handling to be implemented and if so what would it look like?

mjmac:
> Thinking about the blast radius of a potential failure in `libpci`, the library calls are only made when health statistics are explicitly requested in a scan. This is not performed on server start-up and only through specific `dmg storage query list-devices --health` and `dmg storage scan --nvme-health` commands.

OK. Again, my main concern is avoiding a scenario where someone tries to use DAOS on a platform where it hasn't been run before, and it is completely unusable due to unexpected device configurations. It sounds like you're saying that the failure mode in this case would be graceful enough (e.g. device health queries might fail, but everything else should work OK) to allow basic functionality, even if some of the added features might need a software update to work properly.

kjacque (approved):
Thanks for the improvements. LGTM!

mjmac:
> The failure modes I was concerned about were specific to `libpci` failures that could not be handled within Go. I can definitely allow the health query to proceed in case of an unexpected configuration buffer content as I now see that's what you are getting at.
> 
> Looking at the extra checks we add before passing the configuration buffer to the `libpci` API, it's just checking the size is non-zero and that the content is new-line terminated. I think these are reasonable and not "rigid".
> 
> That said I think we have a good path forward. Do you think this is reasonable?

To be clear: My comments weren't intended as a criticism of the parser, just an observation in service of the question I had about what happens if the parser fails to read a configuration. As long as there is a graceful failure in this case, I'm fine with the patch as-is.

knard-intel (approved):
LGTM

tanabarr:
> > The failure modes I was concerned about were specific to `libpci` failures that could not be handled within Go. I can definitely allow the health query to proceed in case of an unexpected configuration buffer content as I now see that's what you are getting at.
> > Looking at the extra checks we add before passing the configuration buffer to the `libpci` API, it's just checking the size is non-zero and that the content is new-line terminated. I think these are reasonable and not "rigid".
> > That said I think we have a good path forward. Do you think this is reasonable?
> 
> To be clear: My comments weren't intended as a criticism of the parser, just an observation in service of the question I had about what happens if the parser fails to read a configuration. As long as there is a graceful failure in this case, I'm fine with the patch as-is.

I'm going to investigate more about what happens to the server process when the library calls fail during parsing. The `bus_scan` API call for example doesn't return an RC and therefore it isn't easy to catch the failure from the calling Go process. As a result of this I don't think there is a "graceful failure" but in the event that a corrupted configuration space is output by the SPDK API we would probably want the server to fail anyway. If you feel this is acceptable please approve the PR, otherwise we can look at further improvements.

mjmac (dismissed):

tanabarr:
As per discussion with @kris a failure in the library does indeed kill the server with this change as seen in the control log:
```
DEBUG 2024/06/30 00:15:25.676274 instance_drpc.go:86: dRPC to index 0 (rank 0): BioHealth/38B/784.99µs
pcilib: dump: line too long or unterminated
[tanabarr@wolf-313 daos]$
```
DMG output:
```
[tanabarr@wolf-313 daos]$ install-rocky/bin/dmg -i storage scan --nvme-health
ERROR: Errors:
  Hosts     Error
  -----     -----
  localhost rpc error: code = Unavailable desc = error reading from server: EOF


ERROR: dmg: 1 host has errors
```

src/control/provider/system/distribution.go
Michael-Hennecke: putting /sbin to the front of PATH (if it isn't already there) can have unintended side-effects. here we only want to make sure that we find the pci tools, so having it at the end should be fine and is safer.

Michael-Hennecke (dismissed):
it's very useful to have these PCIe infos, sissues in that area are otherwise easy to overlook and painful to identify...

tanabarr:
CI issues because `daos_server nvme reset` was failing has been resolved. lspci grep command when run returns rc=1 which wasn't previously been checked for, now the parent function returns empty list of VMD addresses on this situation.

knard-intel (approved):

tanabarr:
CI run https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-14395/43/tests/ failed with the following unrelated known issues:
- https://daosio.atlassian.net/browse/DAOS-16035
- dfuse DaosBuild "Failure to build over dfuse in mode writeback" ???

mjmac (approved):

kjacque (approved):

tanabarr:
Force landing requested based on identified CI failures in previous comment, rebasing is not ideal given this PR has been in CI for over 5 days.

PR #14412, opened 2024-05-21
DAOS-15874 control: Add optional credential cache to agent

mjmac:
On heavily-loaded client nodes where many processes are being
launched by the same user or users, the admin may optionally
enable the credential cache in the agent in order to lower
agent overhead caused by generating identical credentials
for each process owned by a user. The agent-generated
credential is presented by the client process during pool/container
connection and is used to evaluate ACL permissions for
that connection.

Example config:
credential_config:
  cache_expiration: 1m


src/control/cmd/daos_agent/mgmt_rpc.go
```diff
@@ -35,16 +35,13 @@ import (
...
-	fabricMutex     sync.RWMutex
-
 	log            logging.Logger
 	sys            string
 	ctlInvoker     control.Invoker
 	cache          *InfoCache
 	monitor        *procMon
 	cliMetricsSrc  *promexp.ClientSource
-	useDefaultNUMA bool
+	useDefaultNUMA atm.Bool
```
tanabarr, line 44: why change this to atomic?

tanabarr (approved):

src/control/cmd/daos_agent/mgmt_rpc.go
mjmac: I noticed this while working on the feature. As a standard bool, it's racy. One goroutine could flip it to true while others are reading it. I think you'd have to get slammed with a bunch of requests right at agent startup to see it.

mjmac:
Runs 3 & 4 both hit a timeout in the offline reintegration tests, which appears to be tracked in [DAOS-15608](https://daosio.atlassian.net/browse/DAOS-15608). I don't think it's related to any changes made in this patch.

src/control/cmd/daos_agent/infocache.go
```diff
@@ -155,6 +173,16 @@ func (ci *cachedAttachInfo) Refresh(ctx context.Context) error {
...
 }
 
+// Refresh contacts the remote management server and refreshes the GetAttachInfo cache.
+func (ci *cachedAttachInfo) Refresh(ctx context.Context) (func(), error) {
+	if ci == nil {
+		return cache.NoopRelease, errors.New("cachedAttachInfo is nil")
+	}
+
+	ci.Lock()
+	return ci.Unlock, ci.refresh(ctx)
```
kjacque, line 183: Hmm. Seeing the actual implementation of the `RefreshableItem` interface, it seems a little odd to me to have the Refresh functions taking the lock and having the caller release it. We are already exposing the `sync.Locker` interface as part of `cache.Item`, implying an expectation for the caller to be responsible for controlling the lock. Maybe none of these interface functions should be expected to lock themselves at all, and relegate all responsibility for locking to the caller.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -33,32 +53,106 @@ type (
...
 	}
 )
 
-// NewSecurityModule creates a new module with the given initialized TransportConfig
+// NewSecurityModule creates a new module with the given initialized TransportConfig.
 func NewSecurityModule(log logging.Logger, cfg *securityConfig) *SecurityModule {
+	var credCache *credentialCache
+	credSigner := auth.GetSignedCredential
+	if cfg.credentials.CacheLifetime > 0 {
+		cache := &credentialCache{
```
kjacque, line 67: This should be `credCache`, right?

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -33,32 +53,106 @@ type (
...
+
+func (cred *cachedCredential) Key() string {
+	if cred == nil {
+		return ""
+	}
+
+	return cred.key
+}
+
+func (cred *cachedCredential) IsExpired() bool {
```
kjacque, line 97: Need doc comment

kjacque (changes requested):

src/control/cmd/daos_agent/infocache.go
mjmac: > relegate all responsibility for locking to the caller

Hmm. The idea of returning the unlock closure is that the caller can retrieve a cached item that can be safely worked with in a multi-threaded environment. When the caller does not need thread safety anymore, it releases the lock. The pattern of returning the closure along with the item forces the caller to do something with the closure and reduces the chances of a caller forgetting to unlock the item, IMO.

An alternative approach would make a copy of the item while under a lock and return the copy, but this seems needlessly inefficient to me.

kjacque: > Hmm. The idea of returning the unlock closure is that the caller can retrieve a cached item that can be safely worked with in a multi-threaded environment. When the caller does not need thread safety anymore, it releases the lock. The pattern of returning the closure along with the item forces the caller to do something with the closure and reduces the chances of a caller forgetting to unlock the item, IMO.

All fair points, but this is already covered by the implementation of `GetOrCreate()` in `ItemCache`. `ItemCache` is the layer primarily interacting with the `Item` interface in the current implementation, taking care of all the locking and refreshing, and returning a closure for its own caller to release the item when finished with it. This is effectively what I was trying to describe, but didn't realize it until I reviewed the old code.

An advantage of sticking with this paradigm is that it keeps direct knowledge of locking and unlocking in the caller of `Item` interface methods, with no implicit locking that needs to be added in each implementation of `Item`/`RefreshableItem`. As I think about it, I think I prefer the old way for that reason. It is a bit easier to understand the expectations.

I do like breaking up the interface into multiple specific interfaces, though, and using that as a means to determine when to refresh.

utils/config/daos_agent.yml
```diff
@@ -61,10 +60,20 @@
...
+#  # Optionally cache generated credentials with the specified cache
+#  # lifetime. By default, a credential is generated for every client
+#  # process that connects to a pool. If the credential cache is
+#  # enabled, then local client processes connecting with stable
+#  # uid:gid associations may take advantage of the cached credential
+#  # and reduce some agent overhead. For heavily-loaded client nodes
+#  # with many frequent (e.g. hundreds per minute) client connections,
+#  # a lifetime of 1-5 minutes may be a reasonable tradeoff between
+#  # performance and responsiveness to user/group database updates.
+#  cred_cache_lifetime: 1m
```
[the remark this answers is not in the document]
mjmac: Updated.

src/control/cmd/daos_agent/security_rpc.go
mjmac: Oops... Missed this in a refactoring. Good catch. Updated the unit test to use the constructor, as that would have caught this.

src/control/cmd/daos_agent/security_rpc.go
mjmac: Done.

src/control/cmd/daos_agent/infocache.go
mjmac: Yep, I'm following you now. Adjusted to go back to the previous style of delegating lock management to the cache instead of the items. I agree that this makes more sense, and is a simpler implementation.

src/control/cmd/daos_agent/infocache_test.go
```diff
@@ -183,12 +195,15 @@ func TestAgent_cachedAttachInfo_NeedsRefresh(t *testing.T) {
...
 				},
+				rpcClient:    mockClient,
+				fetch:        noopGetAttachInfo,
 				lastResponse: &control.GetAttachInfoResp{},
 			},
 		},
 	} {
 		t.Run(name, func(t *testing.T) {
-			test.AssertEqual(t, tc.expResult, tc.ai.NeedsRefresh(), "")
+			refreshed, _ := tc.ai.RefreshIfNeeded(test.Context(t))
```
kjacque, line 205: Should probably verify the resulting error as well?

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -8,20 +8,40 @@ package main
...
+		cache        *cache.ItemCache
+		credLifetime time.Duration
+		cacheMissFn  credSignerFn
+	}
+
+	// cachedCredential wraps a cached credential and implements the cache.ExpirableItem interface.
+	cachedCredential struct {
+		cacheItem
+		key       string
+		expiredAt time.Time
```
kjacque, line 42:
```suggestion
		expiresAt time.Time
```
Just a nit but `expiredAt` reads to me as "already expired"

kjacque (approved):
A couple very minor nonblocking comments. LGTM.

tanabarr (approved):

PR #14455, opened 2024-05-28
DAOS-15920 object: check modification for CPD RPC

Nasf-Fan:
An CPD RPC may contain multiple sub-operations. On server side, simply checking the RPC opcode (DAOS_OBJ_RPC_CPD) cannot detect whether it contains some modification(s) or not. To resove that, related client logic will set ORF_CPD_RDONLY flag when pack the CPD RPC as long as there is one sub-modification in the CPD RPC. Then the server can easily to know related information via such flag.

The patch also fixes another issue that CPD RPC handler miss to set EC update timestamp as to misguide subsequent EC aggregation.

### Gatekeeper:

### DIFF
--- a/src/object/srv_obj.c
+++ b/src/object/srv_obj.c
@@ -2134,8 +2134,7 @@ obj_ioc_fini(struct obj_io_context *ioc, int err)
 	}
 
 	if (ioc->ioc_coc != NULL) {
-		if (obj_is_modification_opc(ioc->ioc_opc) && err == 0 &&
-		    daos_oclass_is_ec(&ioc->ioc_oca))
+		if (ioc->ioc_update_ec_ts && err == 0)
 			ds_cont_ec_timestamp_update(ioc->ioc_coc);
 		ds_cont_child_put(ioc->ioc_coc);
 		ioc->ioc_coc = NULL;
@@ -2281,7 +2280,7 @@ obj_ioc_end(struct obj_io_context *ioc, int err)
 }
 
 static int
-obj_ioc_init_oca(struct obj_io_context *ioc, daos_obj_id_t oid)
+obj_ioc_init_oca(struct obj_io_context *ioc, daos_obj_id_t oid, bool for_modify)
 {
 	struct daos_oclass_attr *oca;
 	uint32_t                 nr_grps;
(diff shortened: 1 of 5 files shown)


src/object/srv_obj.c
```diff
@@ -2313,7 +2314,7 @@ obj_inflight_io_check(struct ds_cont_child *child, uint32_t opc,
 		}
 	}
 
-	if (!obj_is_modification_opc(opc))
+	if (!obj_is_modification_opc(opc) && (opc != DAOS_OBJ_RPC_CPD || flags & ORF_CPD_RDONLY))
 		return 0;
```
liuxuezhao, line 2318: just confirm for CPD request, will it carry ORF_REINTEGRATING_IO/ORF_REBUILDING_IO flag which will be checked in following code section?
just checked the code seems oci_flags only carry ORF_REINTEGRATING_IO, no where carry ORF_REBUILDING_IO flag found. Not very sure how does it work for that checking?

liuxuezhao (approved):

src/object/srv_obj.c
Nasf-Fan: The flag ORF_REINTEGRATING_IO for CPD RPC was introduced via the commit 7010f2eb822f8b81547386230d7e84620c8f0c74. As my understand, CPR RPC with such flag only happens during reintegration caused rebuild instead of rebuild for target/rank exclusion. As the user reported, the new written data was lost because of discarding old data when rebuild, that is for reintegration. It is expected to be avoided via the following logic:

`2332 
2333         /* If the incoming I/O is during integration, then it needs to wait the
2334          * vos discard to finish, which otherwise might discard these new in-flight
2335          * I/O update.
2336          */
2337         if ((flags & ORF_REINTEGRATING_IO) &&
2338             (child->sc_pool->spc_pool->sp_need_discard &&
2339              child->sc_pool->spc_discard_done == 0)) {
2340                 D_ERROR("reintegrating "DF_UUID" retry.\n", DP_UUID(child->sc_pool->spc_uuid));
2341                 return -DER_UPDATE_AGAIN;
2342         }
2343 `

If without this patch, CPD RPC will always be regarded as non-modification operations, as to pass the obj_inflight_io_check() without returning the expected `-DER_UPDATE_AGAIN`.

liuxuezhao: I was thought about the code of checking ORF_REBUILDING_IO flag in obj_inflight_io_check, I think that code can be removed as now vos can support overwrite. I can do it in another patch. thx

NiuYawei (approved):

PR #14476, opened 2024-05-30
DAOS-15914 dtx: control DTX RPC to reduce network load

Nasf-Fan:
1. Send DTX batched commit RPCs step by step

Currently, for each DTX batched commit operation, it will handle at
most 512 DTX entries that may generate DTX commit RPCs to thousands
of DAOS targets. We will not send out the batched RPCs all together,
instead, we will send them step by step. After each step, the logic
will yield and wait until replied, and then next batched RPCs. That
can avoid holding too much system resources for relative long time.
It is also helpful to reduce the whole system network peak load and
the pressure on related targets.

2. Cleanup stale DTX based on global RPC timeout

Originally, DTX cleanup logic will be triggered if the life for some
stale DTX exceeds the fixed threshold DTX_CLEANUP_THD_AGE_UP (90 sec)
that maybe smaller than global default RPC timeout, as to related DTX
refresh RPC for cleanup logic maybe send out too early before related
modification RPC(s) timeout. It increases network load unnecessarily.

Then we adjust the DTX cleanup threshold based on global default RPC
timeout value, and give related DTX leader sometime after default RPC
timeout to commit or abort the DTX. If the DTX is still prepared after
that, then trigger DTX cleanup to handle potential stale DTX entries.

3. Reorg DTX CoS logic

Reduce the RPCs caused by potential repeated DTX commit. More clear
names for DTX CoS API.

### Gatekeeper:


daltonbohning:
FYI a fix for the DEB build just landed to master

src/dtx/dtx_internal.h
```diff
@@ -192,6 +192,14 @@ extern uint32_t dtx_batched_ult_max;
...
 #define DTX_COLL_TREE_WIDTH		8
 
+/*
+ * If a large transaction has sub-requests to dispatch to a lot of DTX participants,
+ * then we may have to split the dispatch process to multiple steps; otherwise, the
+ * dispatch process may trigger too many in-flight or in-queued RPCs that will hold
+ * too much resource as to server maybe out of memory.
+ */
+#define DTX_RPC_STEP_LENGTH	DTX_THRESHOLD_COUNT
+
```
daltonbohning, line 202: Is it feasible to have arbitrary limits like this in production?
At one end, what if the limit is too high and still floods the server?
At the other end, what if the limit is too low and we don't utilize the available bandwidth / performance?
Tuning things like this for each system seems difficult and error-prone. Should/are we looking at ways to handle this dynamically? And not just THIS limit, but many other arbitrary limits and thresholds that exist

Nasf-Fan: Here, the things to be resolved is that on a large system, such as Aurora, if a lot of engines send out massive RPCs to the same engine in a short time, it may cause the recv_buffer on the destination to be exhausted. So controlling the RPCs load peak will be helpful to reduce such possibility. On the other hand, if one engine sends out too many RPCs to the others concurrently, these in-flight RPCs will occupy a lot of DRAM as to DRAM pressure on the engine. That will affect others (on the same engine) to dynamically allocate DRAM. So the affect of RPC load peak is not only globally but also locally.

There are many factors to be considered. It is not expected that all the DAOS-15914 issues can be resolved if only applying this patch without other (network configuration) changes. On the other hand, this patch only controls DTX batched commit RPCs. That is server side background task, and will not direct affect client side IO bandwidth.

Honestly, this "limit" is some kinds of experience value instead of a strict calculated one. It will be great if we can automatically adjust this "limit" based on the system detailed configuration in the future (although that is not easy). But it does not conflict with current experience value based first-step.

daltonbohning: My concern is that every time we scale higher on Aurora, we hit new limits. We're seeing issues like this with less than 10% of the servers, and less than 5% of the computes. So this is still "small" scale.
I'm not picking on THIS PR specifically. It's more a general concern that repeatedly bumping limits and configuring envs might not be practical. Even if we arrive at a "good" config, it's very likely that new workloads in production will blow up the system again.

While we're debugging these issues now, and learning from them, this is okay. But in production we can't ask Admins to restart and try different patches and configs every time. So if we work towards more general and dynamic solutions, we can safeguard against future unknown workloads.

Nasf-Fan: This patch is not a debug patch to increase or decrease the "threshold", instead, such "threshold" has already been there for sometime. Originally, we used such "threshold" to restrict IO forwarded RPCs workload; now, we extend the restriction to cover DTX batched commit RPCs.

I can understand your concern. If running DAOS on more large scaled cluster, whether the "threshold" will be changed or not. Honestly, I do not know. But I tend to say "no". If we make the "threshold" to be configurable, it will be difficult for the user to determine how large the "threshold" will be a reasonable value. So be as the first step, keeping it as a fixed value will be the most simple solution. If we can automatically adjust such "threshold" based on the system (instead of only current DAOS engine) resource pressure in the future, then we can replace such fixed "threshold" without any compatibility issues. Such dynamical adjustment is more related with RPC flow control among DAOS engines. That is not small work.

src/dtx/dtx_rpc.c
```diff
@@ -369,37 +369,27 @@ struct dtx_common_args {
 
 /* If is_reentrance, this function ignores len. */
 static int
-dtx_req_list_send(struct dtx_common_args *dca, daos_epoch_t epoch, int len, bool is_reentrance)
+dtx_req_list_send(struct dtx_common_args *dca, bool is_reentrance)
```
liw, line 372: [Not requesting changes] Should the comment on new line 370 needs to be remove?

liw (dismissed):
The `crt` and `dtx_rpc.c` changes look good to me.

src/dtx/dtx_common.c
```diff
@@ -1812,6 +1820,19 @@ dtx_cont_register(struct ds_cont_child *cont)
 	if (dbca == NULL)
 		D_GOTO(out, rc = -DER_NOMEM);
 
+	rc = crt_context_get_timeout(dmi->dmi_ctx, &timeout);
```
frostedcmos, line 1823: I am not sure this will give you the right behavior today as the timeout returned will be 0. I grepped through daos code and no one calls crt_context_set_timeout().  Cart on its own leaves it as 0, i've verified it with my local app:

rpc  ERR  src/cart/crt_context.c:1634 crt_context_idx() ALEXMOD: the context timeout is 0

if you are interested in the global timeout used it will be stored in crt_gdata::cg_timeout ("src/cart/crt_internal_types.h")

src/dtx/dtx_rpc.c
Nasf-Fan: OK, I will update the comment.

src/dtx/dtx_common.c
Nasf-Fan: Sure, I will update the patch.

liw (dismissed):

src/dtx/dtx_common.c
frostedcmos: Thanks:) cart changes lgtm

Nasf-Fan:
> [2024-06-17T16:18:41.945Z] DEBUG 2024/06/17 16:18:36.027174 snapshot.go:206: snapshot complete up to: index=123
> [2024-06-17T16:58:57.243Z] Sending interrupt signal to process
NLT was aborted unexpectedly, restart

liuxuezhao (dismissed):

daltonbohning:
Sorry, looks like we gatekeepers missed this one. @Nasf-Fan just checking, we DO want this as a partial fix for https://daosio.atlassian.net/browse/DAOS-15914?

Nasf-Fan:
> Sorry, looks like we gatekeepers missed this one. @Nasf-Fan just checking, we DO want this as a partial fix for https://daosio.atlassian.net/browse/DAOS-15914?

Yes, we need it, although it does not completely resolve the root issue of DAOS-15914, it is still valuable for master.

daltonbohning:
@Nasf-Fan Could you merge latest master please since this is stale now?

Nasf-Fan:
> Test stage Functional on EL 8.8 completed with status FAILURE. https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-14476/33/execution/node/1229/log

test_telemetry_list failed for SRE-471 (DAOS-15337), not related with the patch.
test_ms_failover failed for DAOS-16103, not related with the patch.

The other tests were skipped because of above CI failures.
@daltonbohning , Skip-func-test-el8.8? any suggestion?

liw (approved):
The `crt` and dtx_rpc.c changes look good to me.

Nasf-Fan:
Ping reviewers, thanks!

src/dtx/dtx_common.c
```diff
@@ -336,6 +338,8 @@ dtx_cleanup(void *arg)
 	D_INIT_LIST_HEAD(&dcca.dcca_pc_list);
 	dcca.dcca_st_count = 0;
 	dcca.dcca_pc_count = 0;
+	/* Cleanup stale DTX entries within about 10 seconds windows each time. */
+	dcca.dcca_cleanup_thd = dbca->dbca_cleanup_thd - 10;
```
liuxuezhao, line 342: may assert dbca_cleanup_thd > 10 at some where

liuxuezhao (approved):

PR #14501, opened 2024-06-03
DAOS-15794 tools: Add --health-only flag to (dmg|daos) pool query (#14297)

mjmac:
As a convenience, provide a "streamlined" version of the pool
query that only performs the minimum amount of work to query
the pool's health. Practically speaking, this means that it
will query for disabled ranks and omit the space query, which
is expensive.


daltonbohning (commented):
ftest LGTM

kjacque (approved):

src/control/lib/daos/pool.go
```diff
@@ -87,8 +96,128 @@ type (
...
+	PoolQueryOptionSpace = "space"
+	// PoolQueryOptionRebuild retrieves pool rebuild status as part of the pool query.
+	PoolQueryOptionRebuild = "rebuild"
+	// PoolQueryOptionEnabledEngines retrieves enabled engines as part of the pool query.
+	PoolQueryOptionEnabledEngines = "enabled_engines"
+	// PoolQueryOptionDisabledEngines retrieves disabled engines as part of the pool query.
+	PoolQueryOptionDisabledEngines = "disabled_engines"
 )
 
+var poolQueryOptMap = map[C.int]string{
```
tanabarr, line 120: Just to note that a lot of this mask handling duplicates the implementation in https://github.com/daos-stack/daos/blob/master/src/control/server/storage/config.go#L838 of `OptionBits` (where you offered lots of guidance). It might be beneficial to consolidate at some point so we have a single way of handling masks.

tanabarr (approved):

src/control/lib/daos/pool.go
mjmac: > It might be beneficial to consolidate at some point so we have a single way of handling masks.

Good idea. I'd forgotten about that work.

PR #14503, opened 2024-06-04
DAOS-15960 tests: Improvements for io_sys_admin test

saurabhtandan:

### Gatekeeper:


src/tests/ftest/util/mdtest_test_base.py
```diff
@@ -134,7 +137,8 @@ def get_mdtest_job_manager_command(self, mpi_type):
             manager = get_job_manager(self, "Orterun", self.mdtest_cmd, self.subprocess)
         return manager
 
-    def run_mdtest(self, manager, processes, display_space=True, pool=None, out_queue=None):
+    def run_mdtest(self, manager, processes, display_space=True, pool=None, out_queue=None,
+                   intercept=None):
```
daltonbohning, line 141: We shouldn't add another option for this. The test itself could just do
```
mdtest_cmd.env.update(LD_PRELOAD=intercept, D_IL_REPORT='1')
```

src/tests/ftest/deployment/io_sys_admin.yaml
```diff
@@ -67,13 +67,13 @@ largefilecount:
     - HDF5-VOL
     - DFS
   object_class:
-    - SX
-    - EC_2P1GX
+    - [SX, EC_2P1GX]
+    - [S1, EC_2P1G1]
```
daltonbohning, line 71: Can we please split this into ior_object_class and mdtest_object_class so the config and logic is more clear?
And could we leave a comment saying something like `file oclass, dir oclass`?
Also, for mdtest we can use `*X` for the _**directory**_

daltonbohning (commented):
there are flake and pylint errors

src/tests/ftest/deployment/io_sys_admin.yaml
daltonbohning: Also, do we really intend to use SX/S1 for files and EC for directories?

daltonbohning: Or is this not file, dir? I think splitting into ior_object_class and mdtest_object_class will make this easier to understand

src/tests/ftest/util/file_count_test_base.py
```diff
@@ -49,21 +53,27 @@ def run_file_count(self):
...
         self.add_pool(connect=False)
 
-        for oclass in object_class:
-            self.ior_cmd.dfs_oclass.update(oclass)
-            self.mdtest_cmd.dfs_oclass.update(oclass)
-            self.ior_cmd.dfs_dir_oclass.update(oclass)
+        for idx, _ in enumerate(object_class):
+            self.ior_cmd.dfs_oclass.update(ior_oclass[idx])
+            self.mdtest_cmd.dfs_oclass.update(mdtest_oclass[idx])
+            self.ior_cmd.dfs_dir_oclass.update(ior_oclass[idx])
```
daltonbohning, line 67: This is enumerating `object_class` but then using the index like `ior_oclass[idx]` which I think is confusing. What if `object_class` is longer than `ior_oclass`?

src/tests/ftest/util/file_count_test_base.py
```diff
@@ -33,6 +34,8 @@ def add_containers(self, oclass=None):
             rd_fac = 'rd_fac:{}'.format(str(redundancy_factor))
             properties = (",").join(filter(None, [properties, rd_fac]))
             container.properties.update(properties)
+        if dir_oclass:
+            container.dir_oclass.update(dir_oclass)
```
daltonbohning, line 38: This might be bad if we do, e.g., `container create --oclass S1` because then the DFS root object will be `S1`. Which means ALL client ranks will read root from a single server target.

We should instead do
```
container create --file-oclass S1
```
If you just want to change the file oclass

daltonbohning (changes requested):

src/tests/ftest/util/file_count_test_base.py
```diff
@@ -8,84 +8,106 @@
...
     """
 
-    def add_containers(self, oclass=None):
+    def add_containers(self, file_oclass=None, dir_oclass=None):
         """Create a list of containers that the various jobs use for storage.
 
         Args:
-            oclass: object class of container
+            file_oclass: file object class of container
+            dir_oclass: dir object class of container
```
daltonbohning, line 24:
```suggestion
            file_oclass (str, optional): file object class of container. Defaults to None
            dir_oclass (str, optional): dir object class of container. Defaults to None
```

src/tests/ftest/util/file_count_test_base.py
```diff
@@ -8,84 +8,106 @@
...
         container.create()
 
         return container
 
+    def get_diroclass(self, rd_fac):
+        """
+        rd_fac: redundancy factor
+
+        Returns: value for dir_oclass
+        """
```
daltonbohning, line 49:
```suggestion
        """
        Args
            rd_fac (int): redundancy factor

        Returns:
            str: value for dir_oclass
        """
```

src/tests/ftest/util/file_count_test_base.py
```diff
@@ -8,84 +8,106 @@
...
+                    rd_fac = extract_redundancy_factor(oclass)
+                    dir_oclass = self.get_diroclass(rd_fac)
+                    self.mdtest_cmd.dfs_dir_oclass.update(dir_oclass)
+                    self.container = self.add_containers(oclass, dir_oclass)
                     try:
                         self.processes = mdtest_np
                         self.ppn = mdtest_ppn
-                        self.execute_mdtest()
+                        if self.mdtest_cmd.api.value == 'POSIX':
+                            self.mdtest_cmd.env.update(LD_PRELOAD=intercept, D_IL_REPORT='1')
```
daltonbohning, line 97: This is going to force all tests using this test base to use interception. Is that really what we want?

src/tests/ftest/util/file_count_test_base.py
```diff
@@ -95,6 +117,8 @@ def run_file_count(self):
                         self.ior_cmd.api.update('HDF5')
                         self.run_ior_with_pool(
                             create_pool=False, plugin_path=hdf5_plugin_path, mount_dir=mount_dir)
+                    elif self.ior_cmd.api.value == 'POSIX':
+                        self.run_ior_with_pool(create_pool=False, intercept=intercept)
```
daltonbohning, line 121: Similar - this is going to force all tests using this test base to use interception

daltonbohning (commented):
There are some lint errors that need to be resolved before merging

src/tests/ftest/util/file_count_test_base.py
saurabhtandan: Yeah that's what we want as of now.

src/tests/ftest/util/file_count_test_base.py
saurabhtandan: Yeah that's what we want as of now.

src/tests/ftest/util/file_count_test_base.py
```diff
@@ -15,77 +17,101 @@ class FileCountTestBase(IorTestBase, MdtestBase):
...
-                    self.mdtest_cmd.test_dir.update("/")
-                # run mdtest
-                if self.mdtest_cmd.api.value in ['DFS', 'POSIX']:
+        for api in apis:
+            self.ior_cmd.api.update(api)
+            self.mdtest_cmd.api.update(api)
+            if api == "DFS":
+                self.mdtest_cmd.test_dir.update("/")
+            if self.mdtest_cmd.api.value in ['DFS', 'POSIX']:
+                for oclass in mdtest_oclass:
```
daltonbohning, line 90: Maybe it would be better if the config did this?
```
    mdtest_oclass:
      # file, dir
      - [S1, SX]
      - [EC_2P1G1, RP_2GX]
```
And then the code could do
```
for file_oclass, dir_oclass in mdtest_oclass:
```

Which would 
1. Allow the dir oclass to be configurable without modifying the code
2. Make it easier to understand what is being ran when looking at the config. Right now you have to dig into the code because it's hardcoded

daltonbohning (approved):

daltonbohning (changes requested):
This introduces a test failure
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-14503/4/tests

Because this function is hardcoded to use `-F`
https://github.com/daos-stack/daos/blob/88968688bb1937afeb354d2a8d560466ab21c359/src/tests/ftest/util/data_mover_test_base.py#L1030-L1032
 
But this PR removed -F
https://github.com/daos-stack/daos/blob/9afa8824ca23debd9ef0f96918bdcf8131ebf860/src/tests/ftest/io/small_file_count.yaml#L43


One solution is to update that function to be dynamic instead of hardocded
```
# Original flags used for write
flags = self.ior_cmd.flags.value

# Remove read and write from flags if present
flags = re.sub(" *-r", "", flags)
flags = re.sub(" *-R", "", flags)
flags = re.sub(" *-w", "", flags)
flags = re.sub(" *-W", "", flags)

# Remove stonewall
flags = re.sub(" *-D [0-9]+", "", flags)

# Add read flags
flags += " -r -R"


self.run_ior_with_params(
    "DAOS", daos_path, read_back_pool, read_back_cont,
    flags=flags)
```

src/tests/ftest/util/file_count_test_base.py
daltonbohning: As we discussed, let's ignore this for now since you will be OOO soon

daltonbohning (approved):

mjean308 (approved):

PR #14609, opened 2024-06-18
DAOS-16070 tools: Include BuildInfo field in version output

mjmac:
When available, include more detailed build information in
the version output (most recent tag, etc).

Adds build details to the GetAttachInfo response message in
order to provide server build info to the dmg, daos_agent, and
daos tools (accessible via server-version subcommand).

Features: control
Required-githooks: true
Change-Id: Ia22a6f8007ece7fbf7f801cb9d48826aeae87485


src/control/SConscript
```diff
@@ -46,20 +47,32 @@ def gen_build_id():
     return '0x' + buildid.decode()
 
 
+def get_build_info():
+    """Attempt to retrieve commit/tag details from the build environment."""
+    try:
+        cmd = subprocess.run(['git', 'describe', '--tags', '--dirty', '--always'],
```
jolivier23, line 53: It would be useful if we got all build info here and and spit it to a file and exit when --build-info=write is set and read it from a file if --build-info=read and only automatically get it if --build-info=get (default) or something like that

jolivier23 (approved):

src/control/cmd/daos/util.go
```diff
@@ -62,6 +63,25 @@ func apiVersion() string {
 	)
 }
 
+func srvBuildInfo() (*build.Info, error) {
+	var major uint32
```
tanabarr, line 67: nit: could put all the uint32's on one line

tanabarr (approved):

src/control/cmd/daos/util.go
```diff
@@ -62,6 +63,25 @@ func apiVersion() string {
...
+	var minor uint32
+	var patch uint32
+	var tagPtr *C.char
+
+	rc := C.dc_mgmt_srv_version((*C.uint)(&major), (*C.uint)(&minor), (*C.uint)(&patch), &tagPtr)
+	if err := daosError(rc); err != nil {
+		return nil, err
+	}
```
kjacque, line 75: I wonder if it would be better to do this 100% in Go, instead of having a C function.

mjmac: How would that work? Remember that this is the `daos` utility, which relies on libdaos to talk to the agent...

mjmac:
Test failure is [DAOS-15608](https://daosio.atlassian.net/browse/DAOS-15608).

PR #14650, opened 2024-06-26
Handle human-friendly duration values in pool properties

mjmac:
Updates scrub_freq property handler.

Required-githooks: true

Change-Id: I268be61c8abc914d6cb91e81c800aaff25fa029e

### DIFF
--- a/src/control/lib/daos/pool_property.go
+++ b/src/control/lib/daos/pool_property.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2021-2023 Intel Corporation.
+// (C) Copyright 2021-2024 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -12,6 +12,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/dustin/go-humanize"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


tanabarr (approved):

src/control/lib/daos/pool_property.go
```diff
@@ -27,6 +28,42 @@ func numericMarshaler(v *PoolPropertyValue) ([]byte, error) {
...
+// parseDuration parses a duration string into a time.Duration.
+func parseDuration(s string) (time.Duration, error) {
+	if s == "" {
+		return 0, nil
+	}
+
+	// Append the duration unit if it is missing.
+	if s[len(s)-1] >= '0' && s[len(s)-1] <= '9' {
+		dduStr := defDurationUnit.String()
+		s += dduStr[len(dduStr)-1:]
```
kjacque, line 44: As I think about it, this isn't really durable if we changed the default unit to anything containing multiple letters (e.g. ms). Might be better to just define a string "s" as the default instead of deriving it from `time.Second`.

src/control/lib/daos/pool_property.go
```diff
@@ -27,6 +28,42 @@ func numericMarshaler(v *PoolPropertyValue) ([]byte, error) {
...
+	}
+
+	// For convenience, parse custom duration units beyond
+	// the standard set defined by time.ParseDuration().
+	unit := s[len(s)-1]
```
kjacque, line 49: Same problem here...

src/control/lib/daos/system_prop_test.go
```diff
@@ -177,3 +181,67 @@ func TestDaos_SystemPropertyKeys(t *testing.T) {
...
+			expDur: 0,
+		},
+		"no unit": {
+			input:  "300",
+			expDur: 5 * time.Minute,
+		},
+		"milliseconds": {
+			input:  "3000ms",
+			expDur: 3 * time.Second,
+		},
```
kjacque, line 214: Incorrect result! The code is only parsing out the `s` not the `ms`

kjacque (changes requested):

src/control/lib/daos/system_prop_test.go
mjmac: That result is from `time.ParseDuration()`... 3000ms is 3 seconds, no?

src/control/lib/daos/pool_property.go
mjmac: Yeah... I was just tacking on a "s" but tried to get fancy. I'll go back to the original approach.

src/control/lib/daos/pool_property.go
mjmac: Well, in this case, we're not overriding `time.ParseDuration()`, we're just looking for a trailing 'd' or 'w' to make it easier to specify longer scrubber intervals. If we don't see either of those single-letter custom units, then we fall back to the standard parsing behavior.

src/control/lib/daos/system_prop_test.go
kjacque: Ah wait, you're right.

src/control/lib/daos/pool_property.go
kjacque: Ah, I see. Thanks for clarifying.

kjacque (approved):

johannlombardi (approved):
Thanks!

PR #14665, opened 2024-06-28
DAOS-15967 control: Raise RAS event if link speed|width is downgraded

tanabarr:
Features: control
Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/control/events/ras.go
+++ b/src/control/events/ras.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2022 Intel Corporation.
+// (C) Copyright 2020-2024 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -55,6 +55,8 @@ const (
 	RASSystemStopFailed        RASID = C.RAS_SYSTEM_STOP_FAILED         // error
 	RASEngineJoinFailed        RASID = C.RAS_ENGINE_JOIN_FAILED         // error
 	RASSystemFabricProvChanged RASID = C.RAS_SYSTEM_FABRIC_PROV_CHANGED // info
+	RASNVMeLinkSpeedDown       RASID = C.RAS_DEVICE_LINK_SPEED_DOWN     // error
+	RASNVMeLinkWidthDown       RASID = C.RAS_DEVICE_LINK_WIDTH_DOWN     // error
 )
 
 func (id RASID) String() string {
--- a/src/control/server/instance.go
+++ b/src/control/server/instance.go
@@ -18,6 +18,7 @@ import (
 	mgmtpb "github.com/daos-stack/daos/src/control/common/proto/mgmt"
 	srvpb "github.com/daos-stack/daos/src/control/common/proto/srv"
 	"github.com/daos-stack/daos/src/control/drpc"
+	"github.com/daos-stack/daos/src/control/events"
 	"github.com/daos-stack/daos/src/control/lib/atm"
 	"github.com/daos-stack/daos/src/control/lib/control"
 	"github.com/daos-stack/daos/src/control/lib/ranklist"
@@ -42,6 +43,8 @@ type (
 // be used with EngineHarness to manage and monitor multiple instances
 // per node.
 type EngineInstance struct {
+	events.Publisher
+
 	log             logging.Logger
 	runner          EngineRunner
 	storage         *storage.Provider
@@ -69,9 +72,8 @@ type EngineInstance struct {
 	_lastErr    error // populated when harness receives signal
 }
 
-// NewEngineInstance returns an *EngineInstance initialized with
-// its dependencies.
-func NewEngineInstance(l logging.Logger, p *storage.Provider, jf systemJoinFn, r EngineRunner) *EngineInstance {
+// NewEngineInstance returns an *EngineInstance initialized with its dependencies.
+func NewEngineInstance(l logging.Logger, p *storage.Provider, jf systemJoinFn, r EngineRunner, ps *events.PubSub) *EngineInstance {
 	return &EngineInstance{
 		log:            l,
 		runner:         r,
@@ -80,6 +82,7 @@ func NewEngineInstance(l logging.Logger, p *storage.Provider, jf systemJoinFn, r
 		drpcReady:      make(chan *srvpb.NotifyReadyReq),
 		storageReady:   make(chan bool),
 		startRequested: make(chan bool),
+		Publisher:      ps,
 	}
 }
 
--- a/src/control/server/instance_storage_rpc.go
+++ b/src/control/server/instance_storage_rpc.go
@@ -13,12 +13,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/dustin/go-humanize"
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
 
 	"github.com/daos-stack/daos/src/control/build"
 	"github.com/daos-stack/daos/src/control/common/proto"
 	ctlpb "github.com/daos-stack/daos/src/control/common/proto/ctl"
+	"github.com/daos-stack/daos/src/control/events"
 	"github.com/daos-stack/daos/src/control/fault"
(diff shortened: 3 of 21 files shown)


src/control/events/ras.go
```diff
@@ -55,6 +55,8 @@ const (
 	RASSystemStopFailed        RASID = C.RAS_SYSTEM_STOP_FAILED         // error
 	RASEngineJoinFailed        RASID = C.RAS_ENGINE_JOIN_FAILED         // error
 	RASSystemFabricProvChanged RASID = C.RAS_SYSTEM_FABRIC_PROV_CHANGED // info
+	RASNVMeLinkSpeedDown       RASID = C.RAS_DEVICE_LINK_SPEED_DOWN     // error
+	RASNVMeLinkWidthDown       RASID = C.RAS_DEVICE_LINK_WIDTH_DOWN     // error
```
mjmac, line 59: I think it would be better to just have "Changed" events, as it's also relevant if one of these values is upgraded, right? Imagine there is a flaky bit of hardware that keeps upgrading/downgrading (maybe it needs to be re-seated or something).

With the current approach, the admin would only see the following:
  * 01:02:03 $device speed downgraded to 1Gbps
  * 01:02:05 $device speed downgraded to 1Gbps
  * 01:02:07 $device speed downgraded to 1Gbps
  
Whereas if these are changed events, the picture is a little clearer IMO:
  * 01:02:03 $device speed changed to 1Gbps
  * 01:02:04 $device speed changed to 10Gbps
  * 01:02:05 $device speed changed to 1Gbps
  * 01:02:06 $device speed changed to 10Gbps  
  * 01:02:07 $device speed changed to 1Gbps

src/control/server/instance_storage_rpc.go
```diff
@@ -185,6 +187,32 @@ func addLinkInfoToHealthStats(prov hardware.PCIeLinkStatsProvider, pciCfg string
...
 
+func newLinkEvent(id events.RASID, typ, addr, maxVal, negVal string, port uint32) *events.RASEvent {
+	msg := fmt.Sprintf("nvme ssd %q link %s downgraded (port: %d, max: %s, negotiated: %s)",
+		addr, typ, port, maxVal, negVal)
+	return events.NewGenericEvent(id, events.RASSeverityError, msg, "")
+}
+
+func publishLinkStatWarning(engine Engine, pciAddr string, health *ctlpb.BioHealthResp) {
+	if health.LinkMaxSpeed != health.LinkNegSpeed {
+		engine.Debugf("downgraded link speed on %s, publishing ras event", pciAddr)
```
mjmac, line 198: This assumes that changes in speed are always downgrades, which I don't think is the case, is it?

In any case, I think I agree that a downgrade probably warrants a higher-priority event than an upgrade. Is it possible to use the same event type with different severities, or is severity an intrinsic property of the event type?

src/control/server/instance.go
```diff
@@ -69,9 +72,8 @@ type EngineInstance struct {
 	_lastErr    error // populated when harness receives signal
 }
 
-// NewEngineInstance returns an *EngineInstance initialized with
-// its dependencies.
-func NewEngineInstance(l logging.Logger, p *storage.Provider, jf systemJoinFn, r EngineRunner) *EngineInstance {
+// NewEngineInstance returns an *EngineInstance initialized with its dependencies.
+func NewEngineInstance(l logging.Logger, p *storage.Provider, jf systemJoinFn, r EngineRunner, ps *events.PubSub) *EngineInstance {
```
mjmac, line 76: I don't love the ripple effects of this change, but that is probably more of a criticism of the test code than anything else. I don't see a better way of injecting this dependency.

src/control/events/ras.go
tanabarr: Agreed, the detailed can just be included in the message supplied to `NewGenericEvent()` with ID of a "Changed" event.

src/control/server/instance_storage_rpc.go
tanabarr: The problem here is that we don't store the previous state, i.e. currently there is no way of knowing what the previous health stat was when we do a scan. so we don't know when a transition from degraded to normal happens. It doesn't make sense to report whenever we have a normal state in this case. I will think about how we could store the degraded state.

src/control/server/instance.go
tanabarr: Neither do I, I was thinking about making it a vararg so it didn't need to be included in all usages but that isn't great either.

src/control/server/instance_storage_rpc.go
mjmac: Ah, I see... Hmm. Yeah, and the other issue I think is that couldn't it be normal that the negotiated speed isn't always the max? I don't know for certain, but I would imagine that it's not uncommon for the negotiated speed to be less than the max from this side's perspective. For example, when you have a newer-generation NIC connecting to an older switch.

In that case, maybe it's worth an initial event to indicate that the devices haven't negotiated max speed. I don't think you'd want to raise more events unless something has changed, though.

tanabarr: That would require storing the state somewhere though, right? Currently no state is stored.

tanabarr: I think you would generally want an event if negotiated speed isn't max, even if that's because of a switch bottleneck.

src/control/events/ras.go
tanabarr: Although there currently isn't a mechanism of storing the "downgraded" state, so the only reportable event, on health scan, is "negotiated != max". This single event type covers the original requirement for this change (DAOS-15967) and as I am not planning on offering to implement polling on link state (which was one of the proposals/requests) from either the engine or the control-plane side, it's likely that there will be some deployment specific polling of health stats via dmg commands.

That said I think you correct and the right approach is to provide changed events from any previous state. I propose to implement a `GetLastHealthStats()` engine interface method which we can use to implement the changed events. This result in a more sensible set of published events. Thoughts?

src/control/server/instance_storage_rpc.go
knard-intel: From my side, I agree that it should be relevant to have such events for support tasks: knowing that some interfaces are not working at full speed could be helpful for performance tuning.

knard-intel:
These new events should probable be added in the section "Event list" of the following documentation page https://github.com/daos-stack/daos/blob/master/docs/admin/administration.md

src/control/server/instance_storage_rpc.go
tanabarr: Implemented: events generated when either transitioning from or to negotiated==maximum state for either speed or width

src/control/events/ras.go
tanabarr: Implemented: changed events by referring to "last" health stats recorded which are access through Engine interface

knard-intel (dismissed):

kjacque (commented):
Thinking more about this, I find myself wondering, does raising a RAS event make sense if it can only be generated when a user triggers a scan? I'm wondering whether it would be better to include the speed/width information in the health scan output, if noticing the issue is contingent on the user running a command anyway.

kjacque:
> Thinking more about this, I find myself wondering, does raising a RAS event make sense if it can only be generated when a user triggers a scan? I'm wondering whether it would be better to include the speed/width information in the health scan output, if noticing the issue is contingent on the user running a command anyway.

That said, I'm not against the PR as-is. Just wanted to bring up the question, as generating a RAS event during the scan just feels strange way to notify about a result of a scan.

brianjmurrell (commented):
Packaging changes LGTM.

tanabarr:
reviews apologies for the force push, had trouble with merging master, no changes in the merge

tanabarr:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-14665/8/pipeline CI passed, I'm not sure why it is not being recognised on the GH page

kjacque (approved):

knard-intel (approved):

mjmac (approved):

PR #14672, opened 2024-06-29
DAOS-16038 pool: Enable pool list for clients

mjmac:
Revive parts of the old client mgmt API to allow
libdaos clients to list pools in the system, but
only allow a given client to see the pools that
it could connect to.

Required-githooks: true
Change-Id: I2b3c391ddf042b23811be8f3390ec290e92e4290


src/tests/ftest/pool/list_pools.py
```diff
@@ -35,47 +36,71 @@ def run_case(self, rank_lists, svcn=None):
...
-        # output as a dictionary of pool UUID keys with service replica list
-        # values.
-        detected_uuids = {}
+            expected_admin_uuids[self.pool[-1].uuid.lower()] = self.pool[-1].svc_ranks
+
+        # Remove the default ACLs that allow the creator to access the pool.
+        # These ACLs don't apply to dmg, but do apply to users.
+        offlimits = self.pool[0]
+        self.get_dmg_command().pool_delete_acl(offlimits.uuid, 'OWNER@')
+        self.get_dmg_command().pool_delete_acl(offlimits.uuid, 'GROUP@')
```
daltonbohning, line 48: We do have class methods for this
```suggestion
        offlimits.delete_acl('OWNER@')
        offlimits.delete_acl('GROUP@')
```

src/tests/ftest/pool/list_pools.py
```diff
@@ -35,47 +36,71 @@ def run_case(self, rank_lists, svcn=None):
...
+        # Compare the expected and detected pool information
+        self.assertEqual(
+            expected_user_uuids, detected_user_uuids,
+            "daos pool info does not list all expected pool UUIDs and their "
+            "service replicas")
+
+        # Destroy all the pools
+        if self.destroy_pools(self.pool):
+            self.fail("Error destroying pools")
+        self.pool = []
```
daltonbohning, line 91: This is pre-existing, but I'm not sure why we even need this at the end of the test since teardown will destroy all the pools. But I'm not 100% confident there isn't some implicit assumption somewhere

daltonbohning (commented):
one ftest suggestion if you need to push again

src/tests/ftest/pool/list_pools.py
mjmac: Lost the merge race and had to deal with a conflict, so I went ahead and made this change. Definitely cleaner this way.

daltonbohning (commented):
ftest LGTM

mjmac:
Ping reviewers... @mchaarawi, @kjacque, @liw, @kccain

mjmac:
The failure on the previous run appears to be an instance of [DAOS-15608](https://daosio.atlassian.net/browse/DAOS-15608). I restarted it in hopes of getting a clean run.

mjmac:
For the record, I've manually tested to see what happens when a new client attempts to list pools with an older (2.4.x) server. The client RPC fails with a `-DER_UNREG` error, which is reasonable. We could probably improve the UX a little by converting the raw DAOS error into a fault with more explanation, but that's work for another ticket.

src/mgmt/srv.c
```diff
@@ -429,6 +431,146 @@ void ds_mgmt_pool_find_hdlr(crt_rpc_t *rpc)
...
+	daos_prop_free(props);
+	return rc;
+}
+
+void
+ds_mgmt_pool_list_hdlr(crt_rpc_t *rpc)
+{
+	struct mgmt_pool_list_in   *in;
+	struct mgmt_pool_list_out  *out;
+	size_t                      n_mgmt = 0, n_rpc = 0;
```
kjacque, line 486: nit: in our [C coding standards](https://daosio.atlassian.net/wiki/spaces/DC/pages/4836655703/C+Style+Guide), the standard is to put each variable on its own line, even when they are the same type or uninitialized.

src/mgmt/srv.c
```diff
@@ -429,6 +431,146 @@ void ds_mgmt_pool_find_hdlr(crt_rpc_t *rpc)
...
+	daos_mgmt_pool_info_t      *mgmt_pools = NULL;
+	struct mgmt_pool_list_pool *rpc_pools  = NULL;
+	int                         i, rc, chk_rc;
```
kjacque, line 489: Same comment as above

src/mgmt/srv.c
```diff
@@ -429,6 +431,146 @@ void ds_mgmt_pool_find_hdlr(crt_rpc_t *rpc)
...
+	n_mgmt = in->pli_npools;
+	rc     = ds_get_pool_list(&n_mgmt, mgmt_pools);
+	if (rc != 0) {
+		DL_ERROR(rc, "ds_get_pool_list() failed");
+		D_GOTO(send_resp, rc);
+	}
+
+	/* caller just needs the number of pools */
+	if (in->pli_npools == 0) {
+		out->plo_npools = n_mgmt;
```
kjacque, line 514: Hmm, is it right to give them the total number of pools here, instead of figuring out which ones they have access to first?

src/mgmt/srv.c
```diff
@@ -429,6 +431,146 @@ void ds_mgmt_pool_find_hdlr(crt_rpc_t *rpc)
...
+	D_ALLOC_ARRAY(rpc_pools, n_mgmt);
+	if (rpc_pools == NULL) {
+		D_ERROR("failed to alloc response pools\n");
+		D_GOTO(send_resp, rc = -DER_NOMEM);
+	}
+
+	for (i = 0; i < n_mgmt; i++) {
+		daos_mgmt_pool_info_t      *mgmt_pool = &mgmt_pools[i];
+		struct mgmt_pool_list_pool *rpc_pool  = &rpc_pools[n_rpc];
+		chk_rc = check_cred_pool_access(mgmt_pool->mgpi_uuid, mgmt_pool->mgpi_svc,
```
kjacque, line 527: nit - In coding guidelines we are supposed to include a blank line after variable declaration section.

kjacque (commented):
Nice work, just have a few style comments and a question.

src/mgmt/rpc.h
```diff
@@ -130,6 +140,26 @@ CRT_RPC_DECLARE(mgmt_pool_find, DAOS_ISEQ_MGMT_POOL_FIND,
 
 #define MGMT_POOL_FIND_DUMMY_LABEL "NO LABEL, FINDING BY UUID"
 
+#define DAOS_SEQ_MGMT_POOL_LIST_POOL /* listed pool */           \
+	((uuid_t)		(plp_uuid)		CRT_VAR) \
+	((d_string_t)		(plp_label)		CRT_VAR) \
+	((d_rank_t)		(plp_svc_ldr)		CRT_VAR) \
+	((d_rank_list_t)	(plp_svc_list)		CRT_PTR)
```
liw, line 147: Do we need a uint32_t padding field between plp_svc_ldr and plp_svc_list?

src/mgmt/rpc.h
```diff
@@ -70,10 +72,18 @@ enum mgmt_profile_op {
...
 
+/* clang-format off */
+
 extern struct crt_proto_format mgmt_proto_fmt_v3;
 extern struct crt_proto_format mgmt_proto_fmt_v2;
 extern int dc_mgmt_proto_version;
 
+#define DAOS_OSEQ_MGMT_OP /* output fields */                    \
+	((int32_t)		(mo_rc)			CRT_VAR) \
+	((struct rsvc_hint)	(mo_hint)		CRT_VAR)
```
liw, line 83: Do we need a uint32_t padding field between mo_rc and mo_hint?

src/proto/srv/srv.proto
```diff
@@ -89,3 +89,19 @@ message CheckReportReq {
 message CheckReportResp {
 	int32 status = 1; // DAOS error code.
 }
+
+message ListPoolsReq
+{
+	bool includeAll = 1; // Include all pools in response, regardless of state
```
liw, line 95: [Minor] Should be written as `include_all`, which would give `includeAll` in Go and `include_all` in C. (See https://protobuf.dev/programming-guides/style/.)

src/tests/suite/daos_mgmt.c
```diff
@@ -256,13 +256,13 @@ list_pools_test(void **state)
 	/***** Test: retrieve number of pools in system *****/
 	npools = npools_orig = 0xABC0; /* Junk value (e.g., uninitialized) */
 	/* test only */
-	rc = dmg_pool_list(dmg_config_file, arg->group, &npools, NULL);
```
kccain, line 259: I don't feel too strongly about it, but would it make sense to keep at least one dmg_pool_list() invocation in this test (maybe for that instance, calling both dmg and libdaos API versions and verifying both)?

I see that dmg is exercised in catastrophic recovery tests, so there is still usage/coverage there.

src/mgmt/cli_mgmt.c
```diff
@@ -1271,6 +1272,158 @@ dc_mgmt_tm_register(const char *sys, const char *jobid, key_t shm_key, uid_t *ow
...
+			}
+		}
+	}
+
+out_free_args_pools:
+	if (args->pools && (rc != 0)) {
+		for (pidx = 0; pidx < out->plo_pools.ca_count; pidx++) {
+			daos_mgmt_pool_info_t *pool = &args->pools[pidx];
+
+			if (pool->mgpi_label)
```
knard-intel, line 1401: Is it sure that mgpi_lable and mgpi_svc are initialized to zero ?

src/tests/ftest/pool/list_pools.py
```diff
@@ -35,47 +36,71 @@ def run_case(self, rank_lists, svcn=None):
...
             "service replicas")
 
+        # Verify the 'daos pool list' command lists the correct created pool
+        # information.
+        detected_user_uuids = {}
+        try:
+            for data in self.get_daos_command().get_pool_list_all():
+                detected_user_uuids[data["uuid"]] = data["svc_reps"]
+        except KeyError as error:
+            self.fail("Error parsing dmg pool list output: {}".format(error))
```
knard-intel, line 77:
```suggestion
            self.fail("Error parsing daos pool list output: {}".format(error))
```

knard-intel (commented):
Mostly OK for what I understand.
My main concern is the question from @kjacque regarding the number of pool returned without removing the one not accessible.

src/mgmt/rpc.h
mchaarawi: yes i think so. otherwise I recall we get some corruption in certain cases in the network stack. there are some examples in this file that show that padding being added. to have things 64 bit aligned.

src/mgmt/srv.c
mjmac: Done.

src/mgmt/srv.c
mjmac: It's an upper bound... My thinking on this is that the client always has to make this first call to figure out how much space to allocate before requesting the actual pools. In this case, we're only returning a number, so it doesn't leak any information (aside from the total number of pools in the system, I suppose).

Maybe it's an unnecessary optimization? Just didn't make sense to me to have to go through and query all N pools for ACL access in this step. In a system with a large number of pools, doing the ACL queries twice would introduce more latency into the request.

knard-intel: Make sense for me :-)

src/proto/srv/srv.proto
mjmac: Thanks for catching this. I've updated it

src/mgmt/rpc.h
mjmac: Looks like yes... Didn't think about this. I've updated this and the other new macros where appropriate.

src/mgmt/rpc.h
mjmac: Done.

src/tests/ftest/pool/list_pools.py
mjmac: Fixed, thanks.

src/tests/ftest/pool/list_pools.py
mjmac: Unless you feel strongly about it, I'm inclined to leave well enough alone, at least in this PR.

src/tests/suite/daos_mgmt.c
mjmac: The `dmg_pool_list()` helper was added when we removed the old client-side pool list capability. The ftest stuff exercises dmg pretty thoroughly, so my inclination would actually be to completely remove the dmg helper in favor of going back to the client-only listing code in the suite.

src/mgmt/cli_mgmt.c
```diff
@@ -1271,6 +1272,158 @@ dc_mgmt_tm_register(const char *sys, const char *jobid, key_t shm_key, uid_t *ow
...
+
+	D_DEBUG(DB_MGMT, "req_npools=" DF_U64 " (pools=%p, *npools=" DF_U64 "\n", in->pli_npools,
+		args->pools, *args->npools);
+
+	crt_req_addref(rpc);
+	rc = daos_rpc_send_wait(rpc);
+	if (rc != 0) {
+		DL_ERROR(rc, "rpc send failed");
+		crt_req_decref(rpc);
+		goto rechoose;
```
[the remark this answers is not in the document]
mjmac: I implemented this client API to follow the pattern used by other rsvc clients, where the request is retried forever, or at least until the user decides to cancel it. As I recall from a conversation with @liw about this pattern, the idea is that in a large-scale distributed system, it may be difficult to distinguish from the client's perspective from a truly broken system vs. a system that is in the process of initializing or is otherwise not ready yet.

As we have seen on the `dmg` side of things, it's challenging to choose a timeout value that works reliably in all cases.

src/mgmt/cli_mgmt.c
mjmac: Hmm. I guess not, because the caller is responsible for allocating that array so we don't know for sure what they did. I added explicit NULL assignments prior to copying from the RPC response in the loop.

src/mgmt/cli_mgmt.c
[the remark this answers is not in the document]
knard-intel: Sounds logical.  Thanks for the explanation.

src/mgmt/cli_mgmt.c
```diff
@@ -1271,6 +1272,158 @@ dc_mgmt_tm_register(const char *sys, const char *jobid, key_t shm_key, uid_t *ow
...
+				d_rank_list_free(pool->mgpi_svc);
+		}
+	}
+out_put_req:
+	if (rc != 0)
+		DL_ERROR(rc, "failed to list pools");
+
+	crt_req_decref(rpc);
+
+	if (in != NULL) {
```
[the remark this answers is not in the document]
mjmac: Yeah, I suppose it is. You'd only get here after in has been assigned from `crt_req_get()`.

src/mgmt/cli_mgmt.c
```diff
@@ -1271,6 +1272,158 @@ dc_mgmt_tm_register(const char *sys, const char *jobid, key_t shm_key, uid_t *ow
...
+		/* Ensure credential memory is wiped clean */
+		explicit_bzero(in->pli_cred.iov_buf, in->pli_cred.iov_buf_len);
```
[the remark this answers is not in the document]
mjmac: I'm not sure, to be honest. This credential management stuff was added by @dpquigl, but maybe @kjacque has a better handle on it than I do. After looking at it a bit more, I think I agree that it probably needs to be done on the rechoose path too if it's going to be done at all, so I've added a static helper to do it in all three paths.

src/include/daos_task.h
```diff
@@ -191,6 +192,16 @@ struct daos_obj_register_class_t {
 	struct daos_oclass_attr	*cattr;
 };
 
+/** pool management pool list args */
+typedef struct {
+	/** Process set name of the DAOS servers managing the pool */
+	const char            *grp;
```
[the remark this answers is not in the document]
mjmac: Agreed that it should either be "the system" or "the pools"... Was a copy/paste from other struct defs. I've changed it to "the system". With regard to the "process set" terminology, it's consistent with the other doc strings. I'm inclined to have it match the others, and if someone wants to mass-update everything, this will be included in that effort.

kjacque (approved):

src/mgmt/cli_mgmt.c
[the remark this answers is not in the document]
kjacque: If we are requesting new creds with each rechoose, we should definitely clear the old creds first. Good catch.

kccain (approved):
C changes LGTM

src/control/cmd/daos/pool.go
```diff
@@ -89,61 +89,93 @@ func (cmd *poolBaseCmd) PoolID() ui.LabelOrUUIDFlag {
...
+		var err error
+		cmd.poolUUID, err = uuidFromC(poolInfo.pi_uuid)
+		if err != nil {
+			cmd.disconnectPool()
+			return err
 		}
 	case cmd.PoolID().HasUUID():
 		cmd.poolUUID = cmd.PoolID().UUID
 		cmd.Debugf("connecting to pool: %s", cmd.poolUUID)
 		cUUIDstr := C.CString(cmd.poolUUID.String())
```
tanabarr, line 163: nit: cUUIDstr isn't really consistent with the naming conventions used in our code base, cUUIDStr or cUuidStr would be

src/mgmt/srv.c
```diff
@@ -429,6 +431,150 @@ void ds_mgmt_pool_find_hdlr(crt_rpc_t *rpc)
...
+		daos_mgmt_pool_info_t      *mgmt_pool = &mgmt_pools[i];
+		struct mgmt_pool_list_pool *rpc_pool  = &rpc_pools[n_rpc];
+
+		chk_rc = check_cred_pool_access(mgmt_pool->mgpi_uuid, mgmt_pool->mgpi_svc,
+						DAOS_PC_RO, &in->pli_cred);
+		if (chk_rc != 0) {
+			if (chk_rc != -DER_NO_PERM)
+				DL_ERROR(chk_rc, DF_UUID ": failed to check pool access",
+					 DP_UUID(mgmt_pool->mgpi_uuid));
+			continue;
```
tanabarr, line 537: So just checking I am understanding this correctly, a NO_PERM is silently handled (not logging permission failure for security reasons?) and the pool is skipped. rpc_pools then may not be fully populated as n_mgmt is array size but only n_rpc copied. Would it be sensible to reduce the size of the RPC payload by only allocating space for the accessible pools or is that an unnecessary optimisation?

tanabarr (approved):

src/mgmt/cli_mgmt.c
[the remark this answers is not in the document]
liw: @knard-intel, for the engine side, we have introduced a deadline via `dsc_pool_svc_call`; for the client side, since the DAOS API does not have the concept of timeout and deadline, it's okay to continue using infinite retries, I think.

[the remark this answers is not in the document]
knard-intel: @liw thanks for the detailed explanation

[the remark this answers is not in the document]
liw: @knard-intel, np, it was a good question.

knard-intel (approved):

src/control/cmd/daos/pretty/pool.go
```diff
@@ -98,3 +98,187 @@ func PrintPoolQueryTargetInfo(pqti *daos.PoolQueryTargetInfo, out io.Writer) err
 
 	return w.Err
 }
+
+func poolListCreateRow(pool *daos.PoolInfo, upgrade bool) txtfmt.TableRow {
```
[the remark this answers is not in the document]
mjmac: The `dmg` command code pulls in protobuf and other dependencies, and its pretty-printers work with types from `lib/control` and `lib/daos`. These pretty-printers work only with types from `lib/daos`. The goal I'm pushing toward is for `lib/daos` to have no dependencies on `lib/control` or the rest of the control plane code, because it should be possible to use the client API without pulling in all of that other stuff. The code isn't there yet, but that's where I'm trying to go with it.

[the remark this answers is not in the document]
tanabarr: makes sense, thanks

src/mgmt/srv.c
mjmac: > (not logging permission failure for security reasons?)

The expectation is that there will usually be pools to which the requesting user does not have access, so it's not an error in this case and shouldn't be logged as one.

> Would it be sensible to reduce the size of the RPC payload by only allocating space for the accessible pools or is that an unnecessary optimisation?

I'm not particularly familiar with how the CaRT RPC serialization stuff works, but I'd imagine that setting the `ca_count` value takes care of properly sizing the RPC payload. Maybe @frostedcmos can elaborate?

PR #14702, opened 2024-07-05
DAOS-16164 pool: Update target status to UPIN for no_data_sync mode

wangshilong:
In data_sync mode, if any targets are down, a rebuild will be triggered, causing targets to enter a DOWNOUT state. During reintegration, data will be discarded, and targets will transition from DOWNOUT to UP, followed by another rebuild that changes the targets to UPIN.

In no_data_sync mode, if targets are down, the rebuild process will be skipped, and targets will remain in the DOWN state. Reintegration will skip data discard, bring targets from DOWN to UP directly, and then mark them as UPIN without triggering a rebuild.

Required-githooks: true

### Gatekeeper:


src/pool/srv_pool.c
```diff
@@ -6829,10 +6829,10 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
 
 	if (svc->ps_pool->sp_reint_mode == DAOS_REINT_MODE_NO_DATA_SYNC) {
 		D_DEBUG(DB_MD, "self healing is disabled for no_data_sync reintegration mode.\n");
-		if (opc == POOL_EXCLUDE || opc == POOL_DRAIN) {
-			rc = ds_pool_tgt_exclude_out(svc->ps_pool->sp_uuid, &target_list);
+		if (opc == MAP_REINT) {
```
liuxuezhao, line 6832: when exclude, it still will trigger rebuild?

wangshilong: No, rebuild will be skipped.

liuxuezhao: where is the code to skip it? I mean the rebuild for exclude. user first exclude, then reint right?

wangshilong: sp_reint_mode is no_data_sync, then nothing is done, and followng D_GOTO(out, rc) will be called? which skipped rebuild, am i missing something?

liuxuezhao: To avoid the problem, probably can set the tgt as DOWN when EXCLUDE, to avoid always read from it.
but later when REINT, cannot directly call ds_pool_tgt_finish_rebuild(), may define a new map_update_opc MAP_DOWN2UPIN specially for no_data_sync usage to update pool map.
How do you think?

wangshilong: You are right. thanks for good suggestions!

wangshilong: > To avoid the problem, probably can set the tgt as DOWN when EXCLUDE, to avoid always read from it. but later when REINT, cannot directly call ds_pool_tgt_finish_rebuild(), may define a new map_update_opc MAP_DOWN2UPIN specially for no_data_sync usage to update pool map. How do you think?

For reintegration, in pool_svc_update_map(), pool_svc_update_map_internal() is called first which will mark targets from DOWN to UP firstly(this is a bit like delay rebuild case, co_flags will set DOWN2UP). and then here we call ds_pool_tgt_finish_rebuild() which will mark targets from UP to UPIN. Looks simple for me.

NiuYawei (dismissed):

liuxuezhao (dismissed):

src/pool/srv_pool.c
```diff
@@ -6859,10 +6859,12 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
 
 	if (svc->ps_pool->sp_reint_mode == DAOS_REINT_MODE_NO_DATA_SYNC) {
 		D_DEBUG(DB_MD, "self healing is disabled for no_data_sync reintegration mode.\n");
-		if (opc == POOL_EXCLUDE || opc == POOL_DRAIN) {
-			rc = ds_pool_tgt_exclude_out(svc->ps_pool->sp_uuid, &target_list);
+		if (opc == MAP_EXCLUDE || opc == MAP_DRAIN) {
```
liuxuezhao, line 6862: hmm, why do same for DRAIN? 
user wants to DRAIN it first and then reint it back later?
why cannot be done just by exclude + reint.
I am fine to keep it as it now, seems not really necessary.  Mark it as DOWN for DRAIN also looks not very good.
Or for DRAIN, print an INFO log prompt user that in data sync mode, treat it same as EXCLUDE.

wangshilong: Remove.

liuxuezhao (dismissed):

src/pool/srv_pool.c
```diff
@@ -6863,6 +6863,15 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
 	daos_epoch_t			rebuild_eph = d_hlc_get();
 	uint64_t			delay = 2;
 
+	/*
+	 * For simplicity, deny drain and extend operations in
+	 * the no_data_sync reinteration mode.
+	 */
+	if (svc->ps_pool->sp_reint_mode == DAOS_REINT_MODE_NO_DATA_SYNC) {
+		if (opc != MAP_REINT && opc != MAP_EXCLUDE)
```
kccain, line 6871: maybe for future patch? add a test so that with no_data_sync mode and a pool extend or pool drain is attempted, this is confirmed to be returned to the client/test program?

kccain (dismissed):

NiuYawei (dismissed):

kccain:
As requested by @wangshilong I will add a commit here (and will probably also merge latest master branch).

The commit will address the scenario that the reintegration mode property value is not correctly gotten on PS leader engine startup:

- a pool starts out with the reintegration mode property set to "data_sync" default (0).
- dmg pool set-prop reintegration:no_data_sync (1) is run. 
- the engines are stopped and restarted.
- Upon restart, it can be observed in rebuild code on the leader engine, ds_rebuild_regenerate_task() reads the reintegration mode value from the DRAM-cached value in pool->sp_reint_mode, sees the (incorrect) value 0. And continues to run its logic (which is not desired in this case),. Whereas if it instead reads from the entry in the daos_prop *prop argument (DAOS_PROP_PO_REINT_MODE entry), it will see the correct value 1. The calling pool_svc_step_up_cb() just earlier has called ds_pool_iv_prop_update() that will (eventually) update pool->sp_reint_mode value - but this has not yet occurred by the time ds_rebuild_regenerate_task() runs.

wangshilong:
We had some discuss that we are going to remove this property finally(potential for 2.6.1 as well). since this is a small PR. before that I think still makes sense to land it firstly, thanks!

kccain:
I removed myself as reviewer because I contributed as a co-author to the code change.

NiuYawei (approved):

liuxuezhao (approved):

kccain:
@wangshilong has closed the corresponding ticket https://daosio.atlassian.net/browse/DAOS-16164 so I will close the PR as well.

PR #14730, opened 2024-07-09
DAOS-16127 tools: Add daos health check command

mjmac:
Perform basic system health checks from the client
perspective. Checks the following:

  * Client/Server versions
  * Key library versions and paths
  * Connected system information
  * Pool status for all pools to which the user
    has access
  * Container status for all containers in the
    checked pools

### DIFF
--- a/src/control/build/lib_versions.go
+++ b/src/control/build/lib_versions.go
@@ -1,0 +1,227 @@
+//
+// (C) Copyright 2024 Intel Corporation.
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+//
+
+package build
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/daos-stack/daos/src/control/lib/dlopen"
+)
+
+/*
+#include <stdint.h>
+
+int
+get_fi_version(void *fn, int *major, int *minor, int *patch)
+{
+	uint32_t version;
+
+	if (fn == NULL || major == NULL || minor == NULL || patch == NULL) {
+		return -1;
+	}
+	// uint32_t fi_version(void);
+	version = ((uint32_t (*)(void))fn)();
+
+	if (version == 0) {
+		return -1;
+	}
+
+	// FI_MAJOR(version);
+	*major = (version >> 16);
+	// FI_MINOR(version);
+	*minor = (version & 0xFFFF);
+	// No way to get at revision number via ABI. :(
+
+	return 0;
+}
+
+int
+get_hg_version(void *fn, int *major, int *minor, int *patch)
+{
+	if (fn == NULL || major == NULL || minor == NULL || patch == NULL) {
+		return -1;
+	}
+	// int HG_Version_get(int *, int *, int *);
+	return ((int (*)(int *, int *, int *))fn)(major, minor, patch);
+}
+
+int
+get_daos_version(void *fn, int *major, int *minor, int *fix)
+{
+	if (fn == NULL || major == NULL || minor == NULL || fix == NULL) {
+		return -1;
+	}
+	// int daos_version_get(int *, int *, int *);
+	return ((int (*)(int *, int *, int *))fn)(major, minor, fix);
+}
+*/
+import "C"
+
+// readMappedLibPath attempts to resolve the given library name to an on-disk path.
+// NB: The library must be loaded in order for it to be found!
+func readMappedLibPath(input io.Reader, libName string) (string, error) {
+	if libName == "" {
+		return "", nil
+	}
+
+	libs := make(map[string]struct{})
+	libRe := regexp.MustCompile(fmt.Sprintf(`\s([^\s]+/%s[\-\.\d]*\.so.*$)`, libName))
+	scanner := bufio.NewScanner(input)
+	for scanner.Scan() {
+		if matches := libRe.FindStringSubmatch(scanner.Text()); len(matches) > 1 {
+			libs[matches[1]] = struct{}{}
+		}
+	}
+
+	if len(libs) == 0 {
+		return "", errors.Errorf("unable to find path for %q", libName)
+	} else if len(libs) > 1 {
+		return "", errors.Errorf("multiple paths found for %q: %v", libName, libs)
+	}
+
+	var libPath string
+	for lib := range libs {
+		libPath = lib
+		break
+	}
+	return libPath, nil
+}
+
+func getLibPath(libName string) (string, error) {
+	f, err := os.Open("/proc/self/maps")
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to open %s", "/proc/self/maps")
+	}
+	defer f.Close()
+
+	return readMappedLibPath(f, libName)
+}
+
+func getLibHandle(libName string) (string, *dlopen.LibHandle, error) {
+	searchLib := libName + ".so"
+
+	// Check to see if it's already loaded, and if so, use that path
+	// instead of searching in order to make sure we're getting a
+	// handle to the correct library.
+	libPath, _ := getLibPath(libName)
+	if libPath != "" {
+		searchLib = libPath
+	}
+
+	hdl, err := dlopen.GetHandle(searchLib)
+	if err != nil {
+		return "", nil, err
+	}
+
+	if libPath == "" {
+		libPath, err = getLibPath(libName)
+		if err != nil {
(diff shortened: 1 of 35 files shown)


mjmac:
Tagging reviewers for early feedback. This will need to be rebased on master after #14672 lands, but I think it's pretty close to finished.

src/control/lib/daos/api/provider.go
```diff
@@ -0,0 +1,48 @@
...
+		log     debugTraceLogger
+		api     *api
+		cleanup func()
+	}
+)
+
+// NewProvider returns an initialized DAOS API provider.
+func NewProvider(log debugTraceLogger) (*Provider, error) {
+	api := &api{}
+	cleanup, err := api.Init()
```
kjacque, line 33: I find myself wondering if the `api` struct is an unnecessary layer. It doesn't seem that other API calls are going through it.

src/control/lib/daos/health.go
```diff
@@ -0,0 +1,27 @@
...
+package daos
+
+import (
+	"github.com/google/uuid"
+
+	"github.com/daos-stack/daos/src/control/build"
+)
+
+type (
+	ComponentBuild struct {
```
kjacque, line 16: Need doc comments on these structs

src/control/cmd/daos/health.go
```diff
@@ -0,0 +1,147 @@
...
+			queryMask.SetOptions(daos.PoolQueryOptionDisabledEngines)
+			tpi, err = queryPool(poolHdl, queryMask)
+			if err != nil {
+				cmd.Errorf("failed to query pool %s: %v", pool.Label, err)
+				continue
+			}
+			pool.DisabledRanks = tpi.DisabledRanks
+		}
+
+		poolConts, err := listContainers(poolHdl)
```
kjacque, line 96: I know pool list includes only pools that the user can access. Is the same true of container list?

src/control/cmd/daos/pretty/health.go
```diff
@@ -0,0 +1,174 @@
...
+	if si == nil {
+		return
+	}
+
+	fmt.Fprintf(out, "System Name: %s\n", si.Name)
+	fmt.Fprintf(out, "Fabric Provider: %s\n", si.Provider)
+	apStr := "[]"
+	if len(si.RankURIs) > 0 {
+		if apBytes, err := json.Marshal(si.AccessPoints()); err == nil {
+			apStr = string(apBytes)
```
kjacque, line 49: Marshaling JSON here seems a bit odd to me.

kjacque (commented):
Overall looks good.

src/control/lib/daos/system.go
```diff
@@ -0,0 +1,52 @@
...
+		Name     string     `json:"system_name"`
+		Provider string     `json:"fabric_provider"`
+		RankURIs []*RankURI `json:"rank_uris"`
+	}
+)
+
+// AccessPoints returns a string slice representation of the system access points.
+func (si *SystemInfo) AccessPoints() []string {
+	apSet := make(map[string]struct{})
+	// FIXME (DAOS-15064): Show the MS addresses, not all ranks in the system.
```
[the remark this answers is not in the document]
mjmac: Yep. I've already rebased locally on that one and made the updates. Will push once yours lands.

src/control/lib/daos/api/provider.go
mjmac: Yeah... I wasn't planning on adding the API stuff yet, but I realized that I needed the sub-package in order to move the `GetSystemInfo()` implementation. We don't want the `daos` package to have a linker dependency on `libdaos.so` because it provides types that are used by the server too.

Longer term, I think that I want the `api` type to provide a mocking point to allow for easier unit testing, but tbh I still haven't settled on an approach that I'm completely happy with.

src/control/lib/daos/health.go
mjmac: Yep, already updated locally.

src/control/cmd/daos/health.go
mjmac: Hmm... Not sure. If that's not the case, it probably wouldn't be a bad idea to make it work that way, though.

src/control/cmd/daos/pretty/health.go
mjmac: It's just a quick way to stringify the AP list, with each entry quoted. The output could be copy/pasted right into a config file. The alternative is building a new slice and iterating through it to quote each entry, then joining the new slice and manually enclosing it in brackets.

src/control/cmd/daos/health.go
mjmac: Looking at the implementation of [ds_pool_list_cont_handler()](https://github.com/daos-stack/daos/blob/master/src/pool/srv_pool.c#L4373), I don't see any evidence that the user's credential is checked against the container ACLs before returning the list, but maybe it's happening somewhere else?

I think @kccain added this feature. There is also the [ds_pool_filter_cont_handler()](https://github.com/daos-stack/daos/blob/master/src/pool/srv_pool.c#L4519) ... Maybe that can use the ACL check as a filtering criterion?

mjmac:
I think this is pretty much ready to go... Once #14764 lands, I'll re-rebase this back on master and take it out of draft.

src/control/cmd/daos/pretty/health.go
kjacque: That's fair. I hadn't thought of the config file use case (although wouldn't we then want to marshal it to yaml?).

kjacque (approved):
LGTM. Comments are minor and nonblocking.

src/control/lib/daos/api/provider.go
tanabarr: I think we should try to be consistent across the different bindings on how we handle API mocking, we have a number of different approaches and it would be good to settle on one. I'm biased but I liked the way the pciutils bindings ended up linking in with Kris' hardware library framework.

tanabarr (approved):
Is the vendor directory file change intended?
Is there an intention to keep feature parity between dmg and daos utilities where possible or  is there more effort going to daos utility development as part of a usage preference going forward?

src/control/cmd/daos/main.go
```diff
@@ -27,12 +27,14 @@ type cliOptions struct {
...
 	Verbose       bool             `long:"verbose" description:"Enable verbose output (when applicable)"`
 	JSON          bool             `long:"json" short:"j" description:"Enable JSON output"`
+	SysName       string           `long:"sys-name" short:"G" description:"DAOS system name (optional)"`
 	Container     containerCmd     `command:"container" alias:"cont" description:"Perform tasks related to DAOS containers"`
 	Pool          poolCmd          `command:"pool" description:"Perform tasks related to DAOS pools"`
 	Filesystem    fsCmd            `command:"filesystem" alias:"fs" description:"POSIX filesystem operations"`
 	Object        objectCmd        `command:"object" alias:"obj" description:"DAOS object operations"`
 	System        systemCmd        `command:"system" alias:"sys" description:"DAOS system operations"`
 	Version       versionCmd       `command:"version" description:"Print daos version"`
+	Health        healthCmds       `command:"health" description:"DAOS health operations"`
```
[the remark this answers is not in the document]
mjmac: I hadn't considered this idea... I'm not strictly opposed to it, but the motivation behind adding this feature to the `daos` tool is that it should give users somewhere to start when things don't seem to be working correctly from their perspective.

src/control/cmd/daos/pretty/health.go
mjmac: Technically, [YAML is a superset of JSON](https://en.wikipedia.org/wiki/JSON#YAML), so the YAML parser will accept JSON. 😄

daltonbohning: Wow. That's good info @mjmac :)

mchaarawi (commented):
my only concern is the change to daos_sys_info, but im no sure if there are any users or tools today that are use that. im not aware of any.
anyway, this is fine for me.

src/control/build/lib_versions.go
```diff
@@ -0,0 +1,227 @@
...
+		break
+	}
+	return libPath, nil
+}
+
+func getLibPath(libName string) (string, error) {
+	f, err := os.Open("/proc/self/maps")
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to open %s", "/proc/self/maps")
+	}
```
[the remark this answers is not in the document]
mjmac: I'm going to be doing follow-up work on the tooling here; I'll fix this later, thanks.

PR #14750, opened 2024-07-11
b/352550886 control: Allow Jumpbox to contact servers regardless of version

mjmac:
The upstream compatibility policy requires that admin clients
(e.g. dmg, Jumpbox) must be the same version as daos_server. In
our environment, this hard requirement may not make as much sense.

This commit adds a small workaround that allows the control client
to masquerade as whichever version the server reports.

Required-githooks: true

Change-Id: I2a170bddf3ee62ede082da10939510880a6702ac

### DIFF
--- a/src/control/lib/control/rpc.go
+++ b/src/control/lib/control/rpc.go
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2023 Intel Corporation.
+// (C) Copyright 2020-2024 Intel Corporation.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -8,8 +8,10 @@ package control
 
 import (
 	"context"
+	"fmt"
 	"math/rand"
 	"os"
+	"regexp"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


jolivier23 (approved):

src/control/lib/control/rpc.go
```diff
@@ -488,6 +490,29 @@ func invokeUnaryRPC(parentCtx context.Context, log debugLogger, c UnaryInvoker,
 				return nil, err
 			}
 
+			// b/352550886: Add a forced-compatibility shim for Jumpbox code to be able to
```
pavelgritsai, line 493: Should we submit buganizer references to GitHub?

src/control/lib/control/rpc.go
```diff
@@ -488,6 +490,29 @@ func invokeUnaryRPC(parentCtx context.Context, log debugLogger, c UnaryInvoker,
...
+						var ctxErr error
+						reqCtx, ctxErr = build.ToContext(reqCtx, c.GetComponent(), srvVer)
+						if ctxErr != nil {
+							log.Debugf("failed to set override version: %v", ctxErr)
+							return nil, errors.Wrapf(err, "unable to override %s version to match server's", c.GetComponent())
+						}
+						break
+					}
+				}
+			}
```
pavelgritsai, line 514: this seems like quite a lot of code compared to other cases - do consider extracting it into a separate method with a descriptive name? e.g. detectClinetServerIncompatibility() and move all the comments into it. This would simplify this method significantly.

jolivier23:
Does Go support any mechanism for conditional builds?  I was wondering if we could get it upstream if it wasn't enabled by default, for example.

src/control/lib/control/rpc.go
mjmac: I was intending that this code would only live on our branches, as there isn't really any need for it upstream.

src/control/lib/control/rpc.go
mjmac: Thanks for the review, Pavel! In general, I agree with you. I think there's a balance between reducing cyclomatic complexity and increasing readability, though. The logic here is specific to rpc error handling, and if I move it into a helper, then the reader has to jump to that helper to understand this error case.

That having been said, this section of the rpc processing code has grown organically over time, so it's probably due for a refactor at some point.

mjmac:
> Does Go support any mechanism for conditional builds? I was wondering if we could get it upstream if it wasn't enabled by default, for example.

It does, using [build tags](https://pkg.go.dev/go/build#hdr-Build_Constraints). There's precedent for using them for release builds and enabling the firmware management stuff that's disabled by default. I can float the idea of reworking this so that it's hidden behind a build tag by default.

PR #14772, opened 2024-07-15
DAOS-16010 control: add support to override rync binary and add args

cdavis28:
Add a new config (support_config->file_transfer_exec) to override the default of using rsync to copy files from servers as part of a "collect-log" command. * Added a new option "-T --transfer-args" to pass additional args to the executable defined in support_config->file_transfer_exec
This allows an example, to execute an external tool like gsutil to copy results to a GCS bucket.
Also a fix for allowing the collect command to work in insecure mode is included here.

### Gatekeeper:


ravalsam (commented):
Code looks good but can you add Unit test code for rsyncAlternateCopy() is possible? That will give some code coverage as not all functionality testing can be possible in Ci.

src/control/lib/support/log.go
```diff
@@ -47,14 +47,15 @@ const (
...
+	StopOnError           bool   `short:"s" long:"stop-on-error" description:"Stop the collect-log command on very first error"`
+	TargetFolder          string `short:"t" long:"target-folder" description:"Target Folder location where log will be copied"`
+	Archive               bool   `short:"z" long:"archive" description:"Archive the log/config files"`
+	ExtraLogsDir          string `short:"c" long:"extra-logs-dir" description:"Collect the Logs from given directory"`
+	LogStartDate          string `short:"D" long:"start-date" description:"Specify the start date, the day from log will be collected, Format: MM-DD"`
+	LogEndDate            string `short:"F" long:"end-date" description:"Specify the end date, the day till the log will be collected, Format: MM-DD"`
+	LogStartTime          string `short:"S" long:"log-start-time" description:"Specify the log collection start time, Format: HH:MM:SS"`
+	LogEndTime            string `short:"E" long:"log-end-time" description:"Specify the log collection end time, Format: HH:MM:SS"`
+	ExtraArgsCollectRsync string `short:"C" long:"cloud-storage" description:"Specify the destination in cloud storage"`
```
mjmac, line 58: From a UX perspective, it's a little awkward to require that the user both set an environment variable and specify a CLI flag. That's two different configuration paradigms. The [go-flags](https://github.com/jessevdk/go-flags) package does support setting flags via env var, so what might be better here is to add another flag that can be set either on the command line or the environment. For example:
```suggestion
         RsyncAltBin   string `short:"R" long:"rsync-alt-bin" description:"Use an alternate binary in place of rsync for transfers" env:"D_RSYNC_ALT"`
	RsyncAltArgs []string `short:"A" long:"rsync-alt-args" description:"Extra arguments for alternate transfer solution (e.g. cloud storage bucket)" env:"D_RSYNC_ALT_ARGS" env-delim:","`
```

Note that this suggestion also changes the extra arguments into a string slice instead of a single string. If we're building in flexibility to use solutions other than rsync, it would be nice to really make it flexible now.

The other thing to note here is that using `go-flags` for argument parsing keeps the responsibility for that in one place, and makes the configuration explicit. If you want to get fancy, you can do some validation on argument parsing by defining type aliases that implement the [Unmarshaler](https://pkg.go.dev/github.com/jessevdk/go-flags#Unmarshaler) interface. There are examples of this in the other commands.

Using raw environment variables for configuration in other places is harder to reason about and maintain.

mjmac (commented):
Added a few comments inline. It would probably be a good idea to add some documentation with examples showing how this would be used. @ravalsam: I didn't see any existing documentation of the support commands under `docs/` but maybe I missed it?

ravalsam:
> Added a few comments inline. It would probably be a good idea to add some documentation with examples showing how this would be used. @ravalsam: I didn't see any existing documentation of the support commands under `docs/` but maybe I missed it?

It's under src/control/lib/support/README.md. Do you want to move under docs?

mjmac:
> > Added a few comments inline. It would probably be a good idea to add some documentation with examples showing how this would be used. @ravalsam: I didn't see any existing documentation of the support commands under `docs/` but maybe I missed it?
> 
> It's under src/control/lib/support/README.md. Do you want to move under docs?

Ah, I see. Well, typically the `README.md` files are for developer documentation. Looks like a lot of the content of that file could probably be moved into the admin guide?

ravalsam:
> > > Added a few comments inline. It would probably be a good idea to add some documentation with examples showing how this would be used. @ravalsam: I didn't see any existing documentation of the support commands under `docs/` but maybe I missed it?
> > 
> > 
> > It's under src/control/lib/support/README.md. Do you want to move under docs?
> 
> Ah, I see. Well, typically the `README.md` files are for developer documentation. Looks like a lot of the content of that file could probably be moved into the admin guide?

Sure, I will move it under Admin guide (troubleshooting.md) with next PR.

src/control/lib/support/log.go
cdavis28: I had considered that solution. It is simpler. My concern was around security.
This would give the use the ability to do a dmg-execute command ability. T
he alternate binary could be "dd" with the args to overwrite a file. Or it could be to echo a command into a script.

src/control/lib/support/log.go
```diff
@@ -435,8 +437,28 @@ func getSysNameFromQuery(configPath string, log logging.Logger) ([]string, error
...
 }
 
+func rsyncAlternateCopy(log logging.Logger, opts ...CollectLogsParams) error {
+	cmd := strings.Join([]string{
+		os.Getenv("D_RSYNC_ALT"),
+		opts[0].TargetFolder,
+		opts[0].ExtraArgsCollectRsync},
+		" ")
+
+	out, err := exec.Command("sh", "-c", cmd).Output()
```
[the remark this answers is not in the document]
cdavis28: I'm just being consistent with the other code in that file.
The rsync is launched the same way.

src/control/lib/support/log.go
mjmac: > My concern was around security.
This would give the user the ability to do a dmg-execute command ability.

That's a very good point, but isn't that also a concern with the PR as it is? Seems like the best way to address that problem is to allow the user to specify the name of an alternate transfer method and then resolve that name into a known binary in a standard path.

cdavis28: Yes that would work. My original plan was to use a destination like "gcs://...." to imply that sort of thing, but that included a bunch of vendor deps.
I was thinking the name could be configurable via the ENV var, but I suppose it could be hard coded to a name like "daos_alt_rsync"

kjacque (changes requested):
In concurrence with the security discussion: allowing an environment variable to inject executable code is unsafe.

If we must allow a non-standard binary, perhaps it could be listed (requiring absolute path) in some config file param? Maybe something like:

```yaml
  support_config:
    file_transfer_exec: /home/myname/bin/myfiletransfer
```

cdavis28:
> In concurrence with the security discussion: allowing an environment variable to inject executable code is unsafe.
> 
> If we must allow a non-standard binary, perhaps it could be listed (requiring absolute path) in some config file param? Maybe something like:
> 
> ```yaml
>   support_config:
>     file_transfer_exec: /home/myname/bin/myfiletransfer
> ```



> In concurrence with the security discussion: allowing an environment variable to inject executable code is unsafe.
> 
> If we must allow a non-standard binary, perhaps it could be listed (requiring absolute path) in some config file param? Maybe something like:
> 
> ```yaml
>   support_config:
>     file_transfer_exec: /home/myname/bin/myfiletransfer
> ```

I wanted to do exactly that. But apparently that section of code does not have access to any type of configuration file.

kjacque:
> I wanted to do exactly that. But apparently that section of code does not have access to any type of configuration file.

Since it looks like the log collection is done on all three binaries, I think it would make sense to create a `support_config` subsection in all three config files, and add it as a parameter into the `CollectSupportLog` function that seems to be used by all three. All three binaries should have a config of some kind while running, so I think it's just a matter of passing it down into the function.

kjacque:
@cdavis28 Please avoid force-pushing once your patch is in the review stage. It makes it a bit harder for us to see what's changed. Adding new commits (and merging master in, if needing) keeps things a bit cleaner in the review process. It'll all be squashed and look like one clean commit when we actually land the PR.

cdavis28:
> @cdavis28 Please avoid force-pushing once your patch is in the review stage. It makes it a bit harder for us to see what's changed. Adding new commits (and merging master in, if needing) keeps things a bit cleaner in the review process. It'll all be squashed and look like one clean commit when we actually land the PR.

Sure, I'll keep that in mind in the future. Since the design changed a bit, I was thinking it was ok.

src/control/lib/support/log.go
mjmac, line 58: I still think that the flag name and description here are too specific to the single use case that you are envisioning. Since this could be used as a more general solution, I suggest a more general approach, e.g.
```suggestion
FileTransferExecArgs string `short:"T" long:"transfer-args" description:"Extra arguments for alternate file transfer tool"`
```

utils/config/daos_server.yml
```diff
@@ -601,3 +601,7 @@
 #
 #  bdev_auto_faulty:
 #    enable: false
+#
+## rather than use rsync to copy files. use the specify binary to copy files.
```
mjmac, line 605: This comment doesn't really explain why an admin would use the config option. As it's currently written, it's not clear that it's an optional override.
```suggestion
## Override the default file transfer mechanism for dmg support collect-log by supplying
## the path to an alternative script or binary that will be used to copy log off of the servers.
## Note that the --transfer-args flag to the collect-log command may be used to supply
## extra runtime arguments used by the copy tool (e.g. a cloud bucket name, etc.)
```

mjmac (commented):
I think this approach is more workable. I have a few comments and suggestions inline.

cdavis28:
FYI. recent force push was just to update signatures on commit messages to fix DCO.

src/control/lib/support/log.go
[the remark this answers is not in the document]
kjacque: Don't need to address it at this time, but it's probably better to execute directly than to depend on the shell.

src/control/lib/support/log.go
```diff
@@ -435,8 +438,28 @@ func getSysNameFromQuery(configPath string, log logging.Logger) ([]string, error
 	return hostNames, nil
 }
 
+func rsyncAlternateCopy(log logging.Logger, opts ...CollectLogsParams) error {
```
kjacque, line 441: The name is a little confusing since it's explicitly not using rsync. Maybe `customCopy`?

If the structure mirrors the rsync logic, could this function also serve as a helper for the rsync copy, to avoid duplication? I.e. Just have the rsync function call into this function with appropriate opts.

kjacque, line 441: From the function logic, it seems like we always expect exactly one `CollectLogsParams`.

```suggestion
func rsyncAlternateCopy(log logging.Logger, opts CollectLogsParams) error {
```

utils/config/daos_server.yml
```diff
@@ -601,3 +601,13 @@
...
+#
+## Override the default file transfer mechanism for dmg support collect-log by supplying
+## the path to an alternative script or binary that will be used to copy log off of the servers.
+## Note that the --transfer-args flag to the collect-log command may be used to supply
+## extra runtime arguments used by the copy tool (e.g. a cloud bucket name, etc.)
+##support_config
+##  file_transfer_exec: /usr/bin/copy_tool.sh
+#
+#support_config:
+#  file_transfer_exec:
```
kjacque, line 613: Maybe suggest rsync as that's the default:

```suggestion
#support_config:
#  file_transfer_exec: /usr/bin/rsync
```
It would be good to call that out in the documentation comment, too.

utils/config/daos_server.yml
```diff
@@ -601,3 +601,13 @@
 #
 #  bdev_auto_faulty:
 #    enable: false
+#
+## Override the default file transfer mechanism for dmg support collect-log by supplying
+## the path to an alternative script or binary that will be used to copy log off of the servers.
+## Note that the --transfer-args flag to the collect-log command may be used to supply
+## extra runtime arguments used by the copy tool (e.g. a cloud bucket name, etc.)
```
kjacque, line 608: I'd suggest moving this documentation section directly above `file_transfer_exec`.  Above the `support_config` section you could indicate that this is a configuration for support-related operations.

kjacque (changes requested):
Looking much better, thank you! Still a few small comments, but it is looking good.

utils/config/daos_server.yml
cdavis28: the problem with this file is that it is somehow used in the tests. I can change the example /usr/bin/copy_tool.sh to /usr/bin/rsync however.

src/control/lib/support/log.go
cdavis28: Yes, I found this puzzling and repeated throughout the file(s).
Maybe @ravalsam can chime in here on what he was thinking?

utils/config/daos_server.yml
cdavis28: Not sure what you mean. That's where it's currently located.

src/control/lib/support/log.go
cdavis28: renamed. I would have to think about rearranging this code to do what you suggest. But I would like to do that in a separate PR or when a bug fix is needed.

utils/config/daos_server.yml
kjacque: I mean that these documentation lines are specific to `file_transfer_exec`, not `support_config` overall. Right now this is the only configurable option in `support_config`, but that may not always be the case.

cdavis28: done

cdavis28:
I've finally been able to test this in a private build of a Parallelstore instance. Had to make some more tweaks, but this version is working as expected.

cdavis28:
@mjmac @kjacque are we good to go here?

src/control/cmd/daos_server/support.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/daos-stack/daos/src/control/common/cmdutil"
 	"github.com/daos-stack/daos/src/control/lib/support"
+	"github.com/pkg/errors"
```
mjmac, line 17: nit: The convention for imports is:

stdlib

vendored

internal

src/control/cmd/daos_server/support.go
```diff
@@ -53,6 +54,13 @@ func (cmd *collectLogCmd) Execute(_ []string) error {
 		}
 	}
 
+	if err := cmd.loadConfig(); err != nil {
+		return errors.Wrapf(err, "failed to load config from %s",
+			cmd.configPath())
+	} else if cmd.configPath() != "" {
+		cmd.Infof("DAOS Server config loaded from %s", cmd.configPath())
+	}
```
mjmac, line 62: I don't understand why this is needed. The `collectLogCmd` struct embeds the `cfgCmd` struct which implements the `cfgLoader` interface, so this should happen automatically over in [main.go](https://github.com/daos-stack/daos/blob/master/src/control/cmd/daos_server/main.go#L167).

mjmac (commented):
Overall LGTM. I don't think the duplicate code for loading the server config is needed, though.

src/control/cmd/daos_server/support.go
cdavis28: That's what I had thought at as well. But this is the general problem I've had with this change from Day 1.
From what I can tell there is no global variable with this information. It's my understanding of why https://github.com/daos-stack/daos/blob/5ea9557d84e2a62c8bbf1fcdea62bbfc5b5dca5d/src/control/lib/support/log.go#L277 code exists.

cdavis28: The entrypoint is here https://github.com/daos-stack/daos/blob/5ea9557d84e2a62c8bbf1fcdea62bbfc5b5dca5d/src/control/lib/support/log.go#L954
The call comes from another machine which doesn't have access to the config file.

mjmac: OK, I've checked this PR out to look at it locally. I still don't think this particular change is needed. I understand why you're loading the server config way down in `rsyncLog()`, but loading it here seems redundant. Have you verified that `cmd.configPath()`, which is used to set `params.Config` down around line 88, is actually wrong unless you load this here?

My concern with just letting this slide is that it sets a precedent. It shouldn't be necessary due to the config loading logic in main.go. If that logic is not working as expected, it should be fixed instead of working around it over here. If you have to work around it now, then the next developer who wants to add some stuff will see your changes and add another workaround, and eventually we've got a big(ger) pile of spaghetti code.

cdavis28: That's correct. The params.Config comes the client of this RPC call.
I need the server side to inject it's settings in here somehow.

ravalsam: The same function is getting called from dmg command too and we do not pass the server config to dmg command. 
https://github.com/daos-stack/daos/blob/12c4905b06ec13ab53375dbaaacce2031e772c8e/src/control/cmd/dmg/support.go#L162

kjacque (commented):
Nothing further to request beyond @mjmac's requested changes. I do think the issue with re-loading the config needs to be addressed.

src/control/cmd/daos_server/support.go
cdavis28: @kjacque @mjmac 
Ok, I think I see how this can/should be done.
The code in https://github.com/daos-stack/daos/blob/master/src/control/cmd/daos_server/config.go needs to move from the main package into a library. Then this code can access the config file loaded in main.go at server startup.
This is a lot bigger change than the one that's already in this PR. I would like to call that refactor out of scope for this change. (pretty please)

src/control/cmd/daos_server/support.go
cdavis28: done

src/control/cmd/daos_server/support.go
cdavis28: Ok, finally understood what was asked. Thanks @mjmac
Done.

kjacque (approved):

mjmac (approved):
@cdavis28: Please update the description to reflect what you want the commit body to say. It still references the environment variable stuff.

cdavis28:
@kjacque @mjmac seems I'm not authorized to merge this.

PR #14775, opened 2024-07-16
DAOS-16167 test: update soak test to use internal job scheduler

mjean308:
Skip-unit-tests: true
Test-tag: soak_smoke

Required-githooks: true

### Gatekeeper:


src/tests/ftest/util/soak_test_base.py
```diff
@@ -294,6 +299,77 @@ def harasser_job_done(self, args):
...
+        while True:
+            if time.time() > self.end_time or len(jobid_list) == 0:
+                break
+            jobs = []
+            job_results = {}
+            for job_dict in self.joblist:
+                job_id = job_dict["jobid"]
+                if job_id in jobid_list:
+                    node_count = job_dict["nodesperjob"]
+                    if len(node_list) >= node_count:
```
rpadma2, line 326: It will be nice to put a comment here. eg: Start a job if the available nodes is sufficient to start a particular job.

src/tests/ftest/util/soak_test_base.py
```diff
@@ -294,6 +299,77 @@ def harasser_job_done(self, args):
...
+                            f"DBG: node_list after launch_job {node_list}")
+            # run job scripts on all available nodes
+            for job in jobs:
+                job.start()
+            debug_logging(self.log, self.enable_debug_msg, "DBG: all jobs started")
+            for job in jobs:
+                job.join()
+            debug_logging(self.log, self.enable_debug_msg, "DBG: all jobs joined")
+            while not job_queue.empty():
+                job_results = job_queue.get()
```
rpadma2, line 362: If a job is hanging (or) kernel panic we may never get the job_results, Is it not? Is there something we can do to handle those situation?  I am not sure how the python threading mechanism handle any kernel panic, etc... Something to think about...

src/tests/ftest/util/soak_test_base.py
```diff
@@ -343,47 +415,70 @@ def job_setup(self, jobs, pool):
...
+        else:
+            for job_dict in self.joblist:
+                job_dict["jobid"] = get_id()
+                job_id_list.append(job_dict["jobid"])
+
+            # self.schedule_jobs()
+            method = self.schedule_jobs
+            name = "Job Scheduler"
+            scheduler = threading.Thread(target=method, name=name)
+            # scheduler = multiprocessing.Process(target=method, name=name)
```
rpadma2, line 479: Any idea why you decided using threading instead of process? Looks like multiprocessing can take less time to complete a job since it has its own memory, etc... Threading seems to take more time since it is spawned by a main process....

src/tests/ftest/util/soak_test_base.py
```diff
@@ -294,6 +299,77 @@ def harasser_job_done(self, args):
...
+        debug_logging(self.log, self.enable_debug_msg, "DBG: schedule_jobs ENTERED ")
+        job_queue = multiprocessing.Queue()
+        jobid_list = []
+        node_list = self.hostlist_clients
+        lib_path = os.getenv("LD_LIBRARY_PATH")
+        path = os.getenv("PATH")
+        v_env = os.getenv("VIRTUAL_ENV")
+        env = ";".join([f"export LD_LIBRARY_PATH={lib_path}",
+                        f"export PATH={path}",
+                        f"export VIRTUAL_ENV={v_env}"])
```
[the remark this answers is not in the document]
mjean308: No CPATH necessary.   Could pass the entire env but this is all that is needed

src/tests/ftest/util/soak_test_base.py
mjean308: Every job will either fail, timeout or complete.   If the job hangs/timeout we will get the logs up to that point,   If the node that is executing the job panics;  run_remote cmd should fail/hang/timeout and the put to the queue will have a failing status.

src/tests/ftest/util/soak_test_base.py
mjean308: I tried mutliprocessing but this thread that you reference is really a background thread.   I wanted to be able to monior the jobs similar to how soak worked with slurm.     In schedule jobs where it would have been nice to use multirocesing I could not use it with run_remote so I stuck with threads

src/tests/ftest/util/job_manager_utils.py
```diff
@@ -473,6 +473,7 @@ def __init__(self, job, subprocess=False, mpi_type="openmpi"):
         self.tmpdir_base = FormattedParameter("--mca orte_tmpdir_base {}", None)
         self.args = BasicParameter(None, None)
         self.mpi_type = mpi_type
+        self.hostlist = FormattedParameter("-hosts {}", None)
```
daltonbohning, line 476: Why not use a hostfile?

src/tests/ftest/util/soak_test_base.py
```diff
@@ -294,6 +299,79 @@ def harasser_job_done(self, args):
...
```
daltonbohning, line 313: I recommend using EnvironmentVariables here. It's just a dictionary with some helper methods
```
from command_utils_base import EnvironmentVariables

env = EnvironmentVariables()
env['FOO'] = 'bar'
env_str = env.to_export_str()  # export FOO=bar;

File: src/tests/ftest/util/soak_test_base.py [hunk 7]
```diff
@@ -294,6 +299,79 @@ def harasser_job_done(self, args):
...
+                            f"DBG: node_list after launch_job {node_list}")
+            # run job scripts on all available nodes
+            for job in jobs:
+                job.start()
+            # debug_logging(self.log, self.enable_debug_msg, "DBG: all jobs started")
+            # for job in jobs:
+            #     job.join()
+            # debug_logging(self.log, self.enable_debug_msg, "DBG: all jobs joined")
+            while not job_queue.empty():
+                job_results = job_queue.get()
```
daltonbohning, line 364: So this submits potentially multiple jobs and doesn't schedule anymore jobs until all of those are complete. I think you'd rather submit and complete asynchronously? I could help with that improvement if needed. There are a few approaches we could take

src/tests/ftest/util/soak_test_base.py
```diff
@@ -405,21 +503,28 @@ def job_completion(self, job_id_list):
...
-                            self.fail(f"Error canceling Job {job}")
+                    self.log.info("<< SOAK test timeout in Job Completion at %s >>", time.ctime())
+                    if self.job_scheduler == "slurm":
+                        for job in job_id_list:
+                            if not slurm_utils.cancel_jobs(self.log, self.control, int(job)).passed:
+                                self.fail(f"Error canceling Job {job}")
+                    else:
+                        # update soak_results to include job id NOT run and set state = CANCELLED
+                        for job in job_id_list:
+                            if job not in list(self.soak_results.keys()):
```
daltonbohning, line 523:
```suggestion
                            if job not in self.soak_results:
```

src/tests/ftest/util/soak_utils.py
```diff
@@ -233,7 +254,7 @@ def get_daos_server_logs(self):
         self (obj): soak obj
     """
     daos_dir = self.outputsoak_dir + "/daos_server_logs"
-    logs_dir = os.environ.get("DAOS_TEST_LOG_DIR", "/var/tmp/daos_testing/")
+    logs_dir = os.environ.get("DAOS_TEST_LOG_DIR", "/tmp/")
```
daltonbohning, line 257: This should always be defined now
```suggestion
    logs_dir = self.test_env.log_dir
```

src/tests/ftest/util/soak_utils.py
```diff
@@ -320,23 +369,127 @@ def wait_for_pool_rebuild(self, pool, name):
...
+        else:
+            state = "UNKNOWN"
+    # attempt to cleanup any leftover job processes on timeout
+    job_cleanup(log, hosts)
+    if time.time() >= test.end_time:
+        results = {"handle": job_id, "state": "CANCELLED", "host_list": host_list}
+        debug_logging(log, test.enable_debug_msg, f"DBG: JOB {job_id} EXITED launch_jobscript")
+        job_queue.put(results)
+        # give time to update the queue before exiting
+        time.sleep(0.5)
```
daltonbohning, line 472: Is this necessary? As far as I'm aware, the queue should be updated as soon as `put` returns

src/tests/ftest/util/soak_utils.py
```diff
@@ -1433,75 +1594,142 @@ def create_dm_cmdline(self, job_spec, pool, ppn, nodesperjob):
...
+                # write the slurm directives in the job script
+                script_file.write("#SBATCH --job-name={}\n".format(job))
+                script_file.write("#SBATCH --nodes={}\n".format(nodesperjob))
+                script_file.write("#SBATCH --distribution=cyclic\n")
+                script_file.write("#SBATCH --output={}\n".format(output))
+                if sbatch_params:
+                    for key, value in list(sbatch_params.items()):
+                        if value is not None:
+                            if key == "error":
+                                value = value
```
daltonbohning, line 1696:
```suggestion
```

src/tests/ftest/util/job_manager_utils.py
mjean308: The host file would need to be created at script time and unique to each jobscript;  was just easier to use -hosts within the script.  Besides -hosts should be a param for mpirun

src/tests/ftest/util/soak_utils.py
```diff
@@ -32,10 +35,11 @@
 from mdtest_utils import MdtestCommand
 from oclass_utils import extract_redundancy_factor
 from pydaos.raw import DaosApiError, DaosSnapshot
-from run_utils import run_remote
+from run_utils import run_local, run_remote
 from test_utils_container import add_container
 
 H_LOCK = threading.Lock()
+id_counter = count(start=1)
```
daltonbohning, line 42: Suggestion - you might want to start with a larger number so the job ids are easily sortable (if you need that). E.g. with `start=` you'll get
```
1
2
...
9
10  <-- throws off sorting / printing ids in a list format
```

But e.g. `start=1000`, numbers will be aligned from 1000 to 9999
```
1001
1002
...
1009
1010
...
9999
10000 # <-- finally not aligned
```

src/tests/ftest/util/soak_utils.py
```diff
@@ -340,6 +389,110 @@ def wait_for_pool_rebuild(self, pool, name):
...
+    node_results = []
+    down_nodes = NodeSet()
+    state = "UNKNOWN"
+    if time.time() >= test.end_time:
+        results = {"handle": job_id, "state": "CANCELLED", "host_list": host_list}
+        debug_logging(log, test.enable_debug_msg, f"DBG: JOB {job_id} EXITED launch_jobscript")
+        job_queue.put(results)
+        # give time to update the queue before exiting
+        time.sleep(0.5)
+        return
```
daltonbohning, line 444: When `job_queue.put` returns, it should be updated with no need to wait. Did you find otherwise?

src/tests/ftest/util/soak_utils.py
```diff
@@ -1428,75 +1592,140 @@ def create_dm_cmdline(self, job_spec, pool, ppn, nodesperjob):
...
+                script_file.write("    echo \"VIRTUAL_ENV not defined\" \n")
+                script_file.write("else \n")
+                script_file.write("    source $VIRTUAL_ENV/bin/activate \n")
+                script_file.write("fi \n")
+                script_file.write("exec 1> $JOB_LOG \n")
+                script_file.write("exec 2> $JOB_ERROR_LOG \n")
+
+            for cmd in list(job_cmds):
+                script_file.write(cmd + "\n")
+            script_file.close()
```
daltonbohning, line 1726: This shouldn't be necessary, since you opened like `with open(scriptfile, 'w') as script_file`
```suggestion
```

src/tests/ftest/util/soak_utils.py
mjean308: I don't think I need that because the id is added into the log names or it is used in a dictionary

src/tests/ftest/util/soak_test_base.py
```diff
@@ -97,30 +102,29 @@ def setUp(self):
...
-        # the nodes in the partition will be used for soak.
-        if not self.host_info.clients.partition.name:
-            raise SoakTestError(
-                "<<FAILED: Partition is not correctly setup for daos "
-                "slurm partition>>")
-        self.srun_params = {"partition": self.host_info.clients.partition.name}
-        if self.host_info.clients.partition.reservation:
-            self.srun_params["reservation"] = self.host_info.clients.partition.reservation
-        # Include test node for log cleanup; remove from client list
+        self.job_scheduler = self.params.get("job_scheduler", "/run/*", default="slurm")
```
daltonbohning, line 105: When this is more tested and stable I think the default should be the internal, and phase out slurm

src/tests/ftest/util/soak_test_base.py
```diff
@@ -284,6 +289,122 @@ def harasser_job_done(self, args):
...
+        jobs_not_done = []
+        # remove any nodes marked as DOWN
+        node_list.difference_update(self.down_nodes)
+        lib_path = os.getenv("LD_LIBRARY_PATH")
+        path = os.getenv("PATH")
+        v_env = os.getenv("VIRTUAL_ENV")
+        env = ";".join([f"export LD_LIBRARY_PATH={lib_path}",
+                        f"export PATH={path}"])
+        if v_env:
+            env = ";".join([env, f"export VIRTUAL_ENV={v_env}"])
```
daltonbohning, line 310: We should do for `lib_path` what we do for `v_env` since `LD_LIBRARY_PATH` might not be set

src/tests/ftest/util/soak_test_base.py
```diff
@@ -284,6 +289,122 @@ def harasser_job_done(self, args):
...
+                        node_count = job_dict["nodesperjob"]
+                        if len(node_list) < node_count:
+                            # cancel job
+                            self.soak_results.update({job_id: "CANCELLED"})
+                            self.log.info(
+                                "FINAL STATE: soak job %s completed with : %s at %s",
+                                job_id,
+                                "CANCELLED",
+                                time.ctime())
+                            jobs_not_done.remove(job_id)
```
daltonbohning, line 334: Why are these cancelled? Is this for the case, for example:
soak is asked to run a job with 32 nodes, but was given <32 nodes TOTAL, so it's not possible to run?

src/tests/ftest/util/soak_test_base.py
```diff
@@ -333,47 +453,72 @@ def job_setup(self, jobs, pool):
...
+                                              "nodesperjob": npj,
+                                              "taskspernode": ppn,
+                                              "hostlist": None,
+                                              "jobid": None,
+                                              "jobtimeout": jobtimeout,
+                                              "joblog": jobscript[1],
+                                              "joberrlog": jobscript[2]}])
+        # randomize job list
+        random.seed(4)
+        random.shuffle(self.joblist)
```
daltonbohning, line 471: We should use `self.random`, which supports setting a seed in the config
```suggestion
        # randomize job list
        self.random.shuffle(self.joblist)
```

src/tests/ftest/util/soak_utils.py
```diff
@@ -233,7 +254,7 @@ def get_daos_server_logs(self):
         self (obj): soak obj
     """
     daos_dir = self.outputsoak_dir + "/daos_server_logs"
-    logs_dir = os.environ.get("DAOS_TEST_LOG_DIR", "/var/tmp/daos_testing/")
+    logs_dir = self.test_env.log_dir + "/*log*"
```
daltonbohning, line 257: Better to let this handle the slashes
```suggestion
    logs_dir = os.path.join(self.test_env.log_dir, "*log*")
```

src/tests/ftest/util/soak_utils.py
```diff
@@ -244,6 +265,34 @@ def get_daos_server_logs(self):
...
+        cmd4 = f"/usr/bin/rm -rf {sharedscr_dir}"
+        if not run_local(self.log, cmd3, timeout=600).passed:
+            self.log.info("Script file copy failed with %s", cmd3)
+        if not run_local(self.log, cmd4, timeout=600).passed:
+            self.log.info("Script file copy failed with %s", cmd4)
+    # copy the local files; local host not included in hostlist_client
+    if not run_local(self.log, cmd, timeout=600).passed:
+        self.log.info("Local copy failed: %s", cmd)
+    if not run_local(self.log, cmd2, timeout=600).passed:
+        self.log.info("Local copy failed: %s", cmd2)
```
daltonbohning, line 293: Any reason to not use `self.log.error` instead?

src/tests/ftest/util/soak_utils.py
daltonbohning: Is this necessary? As far as I'm aware, the queue should be updated as soon as `put` returns

src/tests/ftest/util/soak_test_base.py
mjean308: agree;

src/tests/ftest/util/soak_test_base.py
mjean308: yes;  CANCELLED is a slurm term and I was trying to keep the FINAL states the same

src/tests/ftest/util/soak_test_base.py
daltonbohning: Sorry, I forgot setting in config is not implemented yet. Please ignore.

daltonbohning (approved):

src/tests/ftest/util/soak_test_base.py
```diff
@@ -144,7 +148,8 @@ def pre_tear_down(self):
         if self.all_failed_jobs:
             errors.append("SOAK FAILED: The following jobs failed {} ".format(
                 " ,".join(str(j_id) for j_id in self.all_failed_jobs)))
-
+        # cleanup any remaining jobs
+        job_cleanup(self.log, self.hostlist_clients)
```
phender, line 152: In the future it would be nice to remove the pre_tear_down method now that we can use register_cleanup to more effectively handle tearDown operations. Just a note; not a requested change.

src/tests/ftest/util/soak_test_base.py
```diff
@@ -284,6 +289,123 @@ def harasser_job_done(self, args):
...
+                            self.log, self.enable_debug_msg, f"DBG: node_count {node_count}")
+                        debug_logging(
+                            self.log,
+                            self.enable_debug_msg,
+                            f"DBG: node_list initial/queue {node_list}")
+                        job_node_list = node_list[:node_count]
+                        debug_logging(
+                            self.log,
+                            self.enable_debug_msg,
+                            f"DBG: node_list before launch_job {node_list}")
```
phender, line 351: Is this supposed to be logging `job_node_list`? Otherwise it doesn't seem to add anything different than the previous debug message.

src/tests/ftest/util/soak_utils.py
```diff
@@ -56,6 +60,27 @@ def ddhhmmss_format(seconds):
...
+def get_id():
+    """Increment a counter to generate job ids
+
+    Returns:
+        int : next counter value
+    """
+    return next(id_counter)
+
+
+def debug_logging(log, enable_debug_msg, log_msg):
```
phender, line 72: This may be better defined as a `SoakTestBase` method, where calls would look like `self.debug_logging("some message")` or `test.debug_logging("some message")`.

src/tests/ftest/util/soak_test_base.py
```diff
@@ -333,47 +454,72 @@ def job_setup(self, jobs, pool):
...
-                    job_id, script, time.ctime())
-                slurm_utils.register_for_job_results(job_id, self, max_wait=self.test_timeout)
-                # keep a list of the job_id's
-                job_id_list.append(int(job_id))
-            else:
-                # one of the jobs failed to queue; exit on first fail for now.
-                err_msg = f"Slurm failed to submit job for {script}"
-                job_id_list = []
-                raise SoakTestError(f"<<FAILED:  Soak {self.test_name}: {err_msg}>>")
+        if self.job_scheduler == "slurm":
```
phender, line 488: It would have been nice to have job scheduler classes to avoid multiple if/else statements.

src/tests/ftest/util/soak_test_base.py
```diff
@@ -559,7 +701,7 @@ def run_soak(self, test_param):
         resv_bytes = self.params.get("resv_bytes", test_param + "*", 500000000)
         ignore_soak_errors = self.params.get("ignore_soak_errors", test_param + "*", False)
         self.enable_il = self.params.get("enable_intercept_lib", test_param + "*", False)
-        self.sudo_cmd = "sudo" if enable_sudo else ""
+        self.sudo_cmd = "sudo -n" if enable_sudo else ""
```
phender, line 704: We should consider replacing uses of `self.sudo_cmd` with `command_as_user()` in the future to simplify maintenance like this.

src/tests/ftest/util/soak_utils.py
phender: fwiw I agree with Dalton, especially if the `id_counter` is used in file names.

src/tests/ftest/util/soak_utils.py
```diff
@@ -340,6 +389,108 @@ def wait_for_pool_rebuild(self, pool, name):
...
+        log (logger): logger for the messages produced by this method
+        job_queue (Queue): job queue to post status of job
+        job_id (int): unique job identifier
+        host_list (list): list of node to pass to job script
+        env (str): environment variables for job script
+        script (str): full path to job script
+        job_log (str): job std out
+        error_log (str): job std error
+        timeout (int): job timeout
+        test (TestObj): soak test obj
```
phender, line 430: Due to line 433 this should be:
```suggestion
        test (SoakTestBase): soak test obj
```

src/tests/ftest/util/soak_utils.py
```diff
@@ -340,6 +389,108 @@ def wait_for_pool_rebuild(self, pool, name):
...
+    state = "UNKNOWN"
+    if time.time() >= test.end_time:
+        results = {"handle": job_id, "state": "CANCELLED", "host_list": host_list}
+        debug_logging(log, test.enable_debug_msg, f"DBG: JOB {job_id} EXITED launch_jobscript")
+        job_queue.put(results)
+        return
+    if isinstance(host_list, str):
+        # assume one host in list
+        hosts = host_list
+        rhost = host_list
```
phender, line 446: Why do we support a list or string here?  We should keep all host information in a NodeSet object.

phender (approved):
Just some nits.

PR #14783, opened 2024-07-18
DAOS-16013 tools: Query all pool targets by default

mjmac:
Running `daos pool query-targets` with just a rank
argument should query all targets on that rank.

Required-githooks: true

Change-Id: Ib5fff358769679f8c282372fdfa8eaba2b5471cd


kjacque (approved):

kccain:
For symmetry should the dmg pool query-targets command support the same?

mjmac:
> For symmetry should the dmg pool query-targets command support the same?

Hmm. Missed that! Yes, it should. Thanks for pointing it out, will push an update.

tanabarr (changes requested):
Can you please add some unit tests to verify this change.

mjmac:
> Can you please add some unit tests to verify this change.

It's not easy to thoroughly test this, but I've added what I can without heroics.

tanabarr (approved):
LGTM

kjacque (approved):

knard-intel (approved):

PR #14785, opened 2024-07-18
DAOS-16250 control: Allow control client version override

mjmac:
Allow the control client to report a version other than the
static value embedded at build time. Enables some nonstandard
use cases where Control API users will take responsibility
for version interoperability without code changes.

Features: control
Required-githooks: true
Change-Id: I0ce6ddbf2c742ce9c8dab9e21e37cd5ea8c5f5b3

### DIFF
--- a/src/control/build/context.go
+++ b/src/control/build/context.go
@@ -53,6 +53,10 @@ func ToContext(parent context.Context, comp Component, verStr string) (context.C
 		return nil, errors.New("component cannot be ComponentAny")
 	}
 
+	if _, exists := metadata.FromOutgoingContext(parent); exists {
+		return nil, ErrCtxMetadataExists
+	}
+
 	version, err := NewVersion(verStr)
 	if err != nil {
 		return nil, err
--- a/src/control/build/context_test.go
+++ b/src/control/build/context_test.go
@@ -89,6 +89,19 @@ func TestBuild_ToContext(t *testing.T) {
 			verString: "x.y.z",
 			expErr:    errors.New("invalid major version"),
 		},
+		"already set": {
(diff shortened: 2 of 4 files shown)


kjacque (approved):
LGTM

mjmac:
Test failures are all in the UCX stage, which is apparently not supposed to run?

tanabarr (commented):
I don't understand the full context behind why you need to introduce a hack like this but I would like to understand a bit more detail on the potential security implementations and whether this could be abused?

mjmac:
> I don't understand the full context behind why you need to introduce a hack like this but I would like to understand a bit more detail on the potential security implementations and whether this could be abused?

I wouldn't characterize this implementation as a hack. It was, however, derived from a more extensive set of changes that could fairly be called hacky, and we'd like to drop those to do this properly.

This patch is simply providing an undocumented way for users of the control API to set the client version before calling an RPC. From a supportability standpoint, Intel's position is that dmg needs to match the servers, and this is reasonable and the default. There's no way to override the version when using dmg, nor should there be.

For advanced users of the control API, however, there are use cases for having a single binary that can talk to multiple versions of the server (e.g. 2.4.x/2.6.x). This is not supported by Intel, nor is there an expectation of support, or compatibility. It's up to the advanced user to ensure that their custom solution does what it needs to do by testing against the server versions in play.

There are no security considerations here. The worst thing that could happen is that a control RPC caller reports that it's a version that is compatible with the server, but it isn't actually compatible. This would be on the RPC caller to resolve.

tanabarr (approved):

mjmac:
Failures are the usual UCX suspects.

PR #14790, opened 2024-07-19
DAOS-10877 vos: gang allocation for huge SV

NiuYawei:
To avoid allocation failure on a fragmented system, huge SV allocation will
be split into multiple smaller allocations, each allocation size is capped
to 8MB (the DMA chunk size, that could avoid huge DMA buffer allocation).

The address of such scattered SV payload is represented by 'gang address'.

Removed io_allocbuf_failure() vos unit test, it's not applicable in gang
SV mode now.

Required-githooks: true

### Gatekeeper:


jolivier23 (dismissed):
I still wonder if we'd be better off with evtree for SV

NiuYawei:
> I still wonder if we'd be better off with evtree for SV

It looks to me that evtree might be too heavy for such scattered SV. Gang address uses 13 bytes (64 bits offset, 32 bits length, 8 bits type) to represent one segment of SV, however, evtree requires at least 64 bytes (evt_node_entry + evt_desc) for each segment. Using evtree will also introduce unnecessary complexities for SV updating & aggregation. (current evtree & aggregation algorithm can't handle the SV overwriting with a smaller value size, no matter if the two update epochs are same or not).

The gang address isn't introduced for scattered SV only, it's a generic feature and could be used for other purposes in the future, for example, to reduce internal fragmentation caused by unaligned I/O size, we could make an extent consists of multiple segments on different media. (Let's assume we support QLC and the block size of QLC is 64k, then a 65k update could be split into 64k on QLC and 1k on TLC).

src/vos/vos_io.c
```diff
@@ -533,9 +543,26 @@ vos_ioc_reserve_init(struct vos_io_context *ioc, struct dtx_handle *dth)
 	for (i = 0; i < ioc->ic_iod_nr; i++) {
 		daos_iod_t *iod = &ioc->ic_iods[i];
 
+		if (iod->iod_type == DAOS_IOD_SINGLE) {
+			gang_nr = vos_irec_gang_nr(ioc->ic_cont->vc_pool, iod->iod_size);
+			if (gang_nr > UINT8_MAX) {
+				D_ERROR("Too large SV:"DF_U64", gang_nr:%u\n",
+					iod->iod_size, gang_nr);
+				return -DER_REC2BIG;
```
Nasf-Fan, line 551: So we cannot support SV if its size is larger than 8MB * 2^8 = 2GB ? That seems not reasonable. Can we force to disable gang_addr (and try to allocate continuous large space) for the SV under such case?

src/vos/vos_io.c
```diff
@@ -2146,37 +2225,102 @@ iod_reserve(struct vos_io_context *ioc, struct bio_iov *biov)
...
+	gaddr.ba_gang_nr = gang_nr;
+	BIO_ADDR_SET_GANG(&gaddr);
+
+	iod_set_sv_addr(ioc, umoff, &gaddr);
+
+	for (i = 0; i < gang_nr; i++) {
+		D_ASSERT(size > 0);
+		alloc_sz = min(size, VOS_GANG_SIZE_THRESH);
+
+		rc = reserve_space(ioc, media, alloc_sz, &off);
```
Nasf-Fan, line 2272: Means that all the sub_addrs must have the same media type? Cannot be mixed?

src/vos/vos_io.c
NiuYawei: I think this number (2GB) is far more larger than the restrictions of current implementation. Without this gang SV, we can't even support 200MB sized SV.

I think we shouldn't try to allocate 2GB and hope it won't fail, furthermore, there is only by default 1GB DMA buffer per-xstream, it can't support DMA transfer for a 2GB SV.

src/vos/vos_io.c
NiuYawei: Gang address itself supports mixed media type, but here VOS SV uses same media type for this moment. If we need to support mixed media SV in the future, we could simply change this function then.

Nasf-Fan: If we do not support mixed media types, then we do not need store the type in gang address to reduce metadata a bit?

src/vos/vos_io.c
Nasf-Fan: I am not sure how the real use case will be, such as for HDF5, hopes 2GB is large enough. Maybe we can add something in our document to describe related restriction.

src/vos/vos_io.c
NiuYawei: See my comment on this PR, the gang address is a common feature which could be used for other purpose in the future.

src/vos/vos_io.c
NiuYawei: Yes, I agree that we'd document all the DAOS API restrictions well, but I think that's not a trivial work and should be done in a separate PR.

NiuYawei:
ping reviewers.

src/vos/vos_io.c
```diff
@@ -2146,37 +2225,102 @@ iod_reserve(struct vos_io_context *ioc, struct bio_iov *biov)
...
+	irec = (struct vos_irec_df *)umem_off2ptr(vos_ioc2umm(ioc), umoff);
+	gaddr_ptr = vos_irec2data(irec);
+
+	bio_addr_set(&gaddr, DAOS_MEDIA_SCM, umem_ptr2off(vos_ioc2umm(ioc), gaddr_ptr));
+	gaddr.ba_gang_nr = gang_nr;
+	BIO_ADDR_SET_GANG(&gaddr);
+
+	iod_set_sv_addr(ioc, umoff, &gaddr);
+
+	for (i = 0; i < gang_nr; i++) {
```
liuxuezhao, line 2268: do we need to consider to yield if gang_nr is large? same for the loop in iod_gang_fetch().
MAX gang_nr is 255 now

liuxuezhao (approved):

src/vos/vos_io.c
NiuYawei: Let's keep it simple for this moment, if it turns out 255 reserve consumes too much CPU, we could add yield then.

NiuYawei:
@jolivier23 @Nasf-Fan could you take look? nothing was changed since your last review.

jolivier23 (approved):

Nasf-Fan (approved):

PR #14804, opened 2024-07-22
DAOS-15800 client: create cart context on specific interface

mchaarawi:
Cart has added the ability to select network interface on context creation. The daos_agent also added a numa-fabric map that can be queried at init time. Update the DAOS client to query from the agent a map of numa to network interface on daos_init(), and on EQ creation, select the best interface for the network context based on the numa of the calling thread.

Required-githooks: true

### Gatekeeper:


src/mgmt/cli_mgmt.c
```diff
@@ -625,14 +641,51 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
...
+				strcat(crt_info->cio_interface, numa_ifaces->ifaces[j]->interface);
+				strcat(crt_info->cio_domain, numa_ifaces->ifaces[j]->domain);
+			}
+			/*
+			 * If we have multiple interfaces per numa node, we want to randomize the
+			 * first interface selected in case we have multiple processes running
+			 * there. So initialize the index array at that interface to -1 to know that
+			 * this is the first selection later.
+			 */
+			if (numa_ifaces->n_ifaces)
```
frostedcmos, line 676: shouldnt this be (if numa_ifaces->n_ifaces > 1 ) to match comment above?

src/mgmt/cli_mgmt.c
```diff
@@ -625,14 +641,51 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
...
+		if (crt_info->cio_domain == NULL)
+			D_GOTO(cleanup, rc = -DER_NOMEM);
+
+		for (i = 0; i < resp->n_numa_fabric_interfaces; i++) {
+			Mgmt__FabricInterfaces *numa_ifaces = resp_g->numa_fabric_interfaces[i];
+			int                     j;
+
+			for (j = 0; j < numa_ifaces->n_ifaces; j++) {
+				if (i != 0 || j != 0) {
+					strcat(crt_info->cio_interface, ",");
```
frostedcmos, line 664: you allocate crt_info->cio_interface and crt_info->cio_domain on line 649 to be of size resp->n_numa_fabric_interfaces * [string_max], however here we can fill at max  (resp->n_numa_fabric_interaces * numa_ifaces->n_ifaces) * [string_max] amount. 
if any numa_ifaces->n_ifaces > 1 then you might overflow the array.

frostedcmos (commented):
inline comments

src/mgmt/cli_mgmt.c
mchaarawi: indeed. i probably got confused with all the names. will fix.

src/mgmt/cli_mgmt.c
mchaarawi: correct. good catch, will fix. thx

src/cart/crt_internal_types.h
```diff
@@ -14,7 +14,7 @@
 #define CRT_CONTEXT_NULL         (NULL)
 
 #ifndef CRT_SRV_CONTEXT_NUM
-#define CRT_SRV_CONTEXT_NUM (64)	/* Maximum number of contexts */
+#define CRT_SRV_CONTEXT_NUM (128) /* Maximum number of contexts */
```
frostedcmos, line 17: as a note this will increase a client side memory footprint. Clients contain 2 cache tables - uri and hg addr; each at most contains  num_ranks * CRT_SRV_CONTEXT_NUM entries. CRT_SRV_CONTEXT_NUM entries allocated at once (as an array) per each discovered rank. In case of a uri cache, each entry is a string such as (ofi+tcp://127.0.0.1:12345) while hg_addr cache the entry is a mercury handle.

frostedcmos: yup 128 should be ok, as we used to have 256 limit for a long time. still, if we had any scaled up test that was on the border of oom, this might cause it to go over, so something to be aware of.

src/mgmt/cli_mgmt.c
```diff
@@ -667,6 +720,64 @@ int dc_mgmt_net_cfg_check(const char *name)
...
+		D_ERROR("No fabric interfaces initialized.\n");
+		return -DER_INVAL;
+	}
+
+	for (i = 0; i < resp_g->n_numa_fabric_interfaces; i++) {
+		Mgmt__FabricInterfaces *numa_ifaces = resp_g->numa_fabric_interfaces[i];
+		int                     idx;
+
+		if (numa_ifaces->numa_node != numa)
+			continue;
```
kjacque, line 752: On the last patch to these structures, I filled out the array with all NUMA nodes, even those with no interfaces, so you should be able to index directly to the desired node now, e.g. `resp_g->numa_fabric_interfaces[numa]`

src/client/api/event.c
```diff
@@ -656,7 +669,23 @@ daos_eq_create(daos_handle_t *eqh)
...
+		char iface[DAOS_SYS_INFO_STRING_MAX];
+
+		rc = dc_mgmt_get_iface(&iface[0]);
+		if (rc) {
+			D_ERROR("failed to get iface: " DF_RC "\n", DP_RC(rc));
+			daos_eq_free(&eqx->eqx_hlink);
+			return rc;
+		}
+		/** if no interface returned, use the default */
+		if (rc == -DER_NONEXIST)
```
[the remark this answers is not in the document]
mchaarawi: hmm, i think i messed up something here when i pushed an update. i will fix that. thanks

src/mgmt/cli_mgmt.c
mchaarawi: im not sure this the case..  i tried on boro that has 2 numa nodes, 1 interface and it reports just 1 entry:

NUMA/Fabric list has 1 entries:
         Numa 0: 1 interfaces
                 numa_node = 0
                 interface = ib0
                 domain = ib0
                 provider = ofi+tcp

kjacque: Ah, we only fill in the gaps up to the highest NUMA node _that has a fabric interface associated with it_. That still allows a shortcut: if numa < resp_g->nr_numa_fabric_interfaces, index directly, otherwise numa does not have a local iface.

mchaarawi: OK let's keep this as is for now then and we can optimize later.
in reality the loop bounds here are like 2-4 numas and 4 interfaces at most per numa, so optimization wise, it's less than negligible :-)

daltonbohning (commented):
ftest LGTM

src/mgmt/cli_mgmt.c
```diff
@@ -644,14 +644,15 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
 
 	d_getenv_bool("D_DYNAMIC_CTX", &d_dynamic_ctx_g);
 	if (d_dynamic_ctx_g) {
-		int i;
+		int i, cnt = 0;
```
kjacque, line 647: style nit - our C style guide says each variable should be on its own line, even when they are the same type.

src/mgmt/cli_mgmt.c
```diff
@@ -625,14 +641,52 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
...
+			Mgmt__FabricInterfaces *numa_ifaces = resp_g->numa_fabric_interfaces[i];
+			int                     j;
+
+			for (j = 0; j < numa_ifaces->n_ifaces; j++) {
+				if (i != 0 || j != 0) {
+					strcat(crt_info->cio_interface, ",");
+					strcat(crt_info->cio_domain, ",");
+				}
+				strcat(crt_info->cio_interface, numa_ifaces->ifaces[j]->interface);
+				strcat(crt_info->cio_domain, numa_ifaces->ifaces[j]->domain);
```
kjacque, line 669: Just occurred to me, we probably should use `strncat` here, as we know the length of the destination buffer. Coverity will probably ding us for `strcat`.

src/mgmt/cli_mgmt.c
```diff
@@ -625,14 +641,52 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
...
-	if (NULL == crt_info->cio_domain)
-		D_GOTO(cleanup, rc = -DER_NOMEM);
 
-	D_INFO("Network interface: %s, Domain: %s\n", info->interface, info->domain);
+	d_getenv_bool("D_DYNAMIC_CTX", &d_dynamic_ctx_g);
+	if (d_dynamic_ctx_g) {
+		int i, cnt = 0;
+
+		for (i = 0; i < resp->n_numa_fabric_interfaces; i++)
+			cnt += resp_g->numa_fabric_interfaces[i]->n_ifaces + 1;
```
frostedcmos, line 650: is it intended to add +1 for each repetition of loop rather than at the end once? I don't quite understand what +1 is for

frostedcmos (commented):
looks good to me, 1 question inline

src/mgmt/cli_mgmt.c
frostedcmos: I thought a bit more about this and i think the loop should be instead something like

```
size = 0;
for (i = 0; i < resp->n_numa_fabric_interfaces; i++) {
	/* DAOS_SYS_INFO_STRING_MAX for interface name and 1 for a separator ',' between them */
	size += resp_g->numa_fabric_interfaces[i]->n_ifaces * (DAOS_SYS_INFO_STRING_MAX + 1);
}

D_ALLOC(crt_info->cio_interface, size);
```
otherwise today for each comma you allocate DAOS_SYS_INFO_STRING_MAX
...

mchaarawi: yea sorry too much multitasking, the +1 was wrong. i mixed up count and size. ill fix this again

src/mgmt/cli_mgmt.c
mchaarawi: no need because we should have already verified the in strings since they are coming from the resp. so it can be marked as false positive in coverity.

mchaarawi: since im already repushing ill just incorporate that

frostedcmos (approved):

kjacque (approved):

mchaarawi:
the failure is because of DAOS-16315
but since i would like this test to run with the new mode, i will hold off on landing this till we update mercury in our CI.

mchaarawi:
I have verified now that the test i wanted to run (pil4dfs_fio) with the new ENV now works after mercury has been updated. the only failure is unreltated: FTEST_aggregation.DaosAggregationBasic
so this is good to go finally.

PR #14812, opened 2024-07-24
DAOS-15583 client: introduce whitelist mode into libpil4dfs

wiliamhuang:
Since daos_init() is expensive to call, it could be beneficial to bypass interception and forward IO requests to dfuse in short-lived applications. Whitelist mode is enabled as default. Users can set env "D_IL_BYPASS_LIST" to disable interception in individual applications, 
export D_IL_BYPASS_LIST="app_a:app_b:app_c:app_d"
Consequently, IO requests are handled by dfuse. User supplied list is appended to a built-in list. Whitelist mode could be disabled by set env "D_IL_NO_BYPASS" 1. Interception is forced on in current process and all children processes. This env is only for testing. Important libpil4dfs env variables are always inherited in children processes now and env variable "D_IL_ENFORCE_EXEC_ENV" is not used anymore.

Features: pil4dfs

Whitelist mode allows users bypass function interception in specific applications.

Required-githooks: true

### Gatekeeper:


wiliamhuang:
The failed test does not use libpil4dfs at all. I could not find exactly same issue. I will rerun this test later in CI.

docs/user/filesystem.md
```diff
@@ -1021,6 +1021,14 @@ libpil4dfs intercepting summary for ops on DFS:
 [op_sum ]  5003
 ```
 
+### Turn on whitelist mode in libpil4dfs
+libpil4dfs could boost IO performance considerably over using dfuse for IO intensive workloads. However, initializing daos environment in libpil4dfs is expensive. To alleviate the overhead of libpil4dfs, user could set env to disable function interception in specific applications, i.e., let dfuse handle all IO requests. "D_IL_WHITELIST=1" turns on whitelist mode. A list of common applications are hardcoded to be skipped in whitelist mode. User can set env "D_IL_BYPASS_ALL_LIST" to bypass function interceptions in target processes and their child processes. "D_IL_BYPASS_LIST" can be set to bypass function interceptions in target processes only. Setting "D_IL_BYPASS_ALL_LIST" or "D_IL_BYPASS_LIST" enables whitelist mode too even "D_IL_WHITELIST=1" is not set. Whitelist mode is recommended for short-lived processes that run a large number (more than 1,000) of times. Example to set env,
```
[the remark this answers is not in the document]
wiliamhuang: Fixed as you suggested. Thank you!

src/tests/ftest/dfuse/il_whitelist.py
```diff
@@ -0,0 +1,97 @@
...
+        """Jira ID: DAOS-15583.
+
+        Test Description:
+            Mount a DFuse mount point
+            Run whitelist_test.
+            Check the number of daos_init() with interception library
+        """
+        if il_lib is not None:
+            lib_path = os.path.join(self.prefix, "lib64", il_lib)
+            env_str = f"export LD_PRELOAD={lib_path}; "
```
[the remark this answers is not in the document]
wiliamhuang: Fixed as you suggested. Thank you!

src/tests/ftest/dfuse/il_whitelist.py
```diff
@@ -0,0 +1,97 @@
...
+            env_str += 'export D_IL_BYPASS_LIST=whitelist_test; '
+
+        result = run_remote(self.log, dfuse_hosts, env_str + exe_path)
+        output = "\n".join(result.all_stdout.values())
+        lines = output.split('\n')
+        num_daos_init = 0
+        for line in lines:
+            if "called daos_init()" in line:
+                num_daos_init += 1
+        print(f'num_daos_init = {num_daos_init}')
```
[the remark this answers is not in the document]
wiliamhuang: Fixed as you suggested. Thank you!

src/tests/ftest/dfuse/il_whitelist.py
```diff
@@ -0,0 +1,97 @@
...
+        """Jira ID: DAOS-15583.
+
+        :avocado: tags=all,daily_regression
+        :avocado: tags=vm
+        :avocado: tags=pil4dfs,dfuse
+        :avocado: tags=ILWhiteList,test_whitelist_pil4dfs
+        """
+        num_daos_init = self.run_test(il_lib='libpil4dfs.so')
+        if num_daos_init != 10:
+            self.fail(f"Test failed: num_daos_init = {num_daos_init}. Expected 10.\n")
```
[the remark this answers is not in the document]
wiliamhuang: New line is removed. Thank you!

src/tests/ftest/dfuse/il_whitelist.py
```diff
@@ -0,0 +1,97 @@
...
+        """Jira ID: DAOS-15583.
+
+        :avocado: tags=all,daily_regression
+        :avocado: tags=vm
+        :avocado: tags=pil4dfs,dfuse
+        :avocado: tags=ILWhiteList,test_whitelist_pil4dfs_bypass
+        """
+        num_daos_init = self.run_test(il_lib='libpil4dfs.so', bypass=True)
+        if num_daos_init != 5:
+            self.fail(f"Test failed: num_daos_init = {num_daos_init}. Expected 5.\n")
```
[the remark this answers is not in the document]
wiliamhuang: New line is removed. Thank you!

src/tests/ftest/dfuse/il_whitelist.py
```diff
@@ -0,0 +1,97 @@
...
+        """Jira ID: DAOS-15583.
+
+        :avocado: tags=all,daily_regression
+        :avocado: tags=vm
+        :avocado: tags=pil4dfs,dfuse
+        :avocado: tags=ILWhiteList,test_whitelist_pil4dfs_bypassall
+        """
+        num_daos_init = self.run_test(il_lib='libpil4dfs.so', bypass_all=True)
+        if num_daos_init != 0:
+            self.fail(f"Test failed: num_daos_init = {num_daos_init}. Expected 0.\n")
```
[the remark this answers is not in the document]
wiliamhuang: New line is removed. Thank you!

src/tests/ftest/dfuse/il_whitelist.yaml
```diff
@@ -0,0 +1,21 @@
...
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos
+  system_ram_reserved: 1
+pool:
+  scm_size: 1000000000
+  control_method: dmg
```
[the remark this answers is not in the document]
wiliamhuang: This line is removed.

src/tests/ftest/dfuse/utest/whitelist_test.c
```diff
@@ -0,0 +1,96 @@
...
+static char  exe_path[PATH_MAX];
+
+static int
+task_child(void)
+{
+	int   rc, fd, status;
+	pid_t pid;
+	char *str_preload;
+	char *argv[3] = {"/usr/bin/ls", "", NULL};
+	char *envp[5] = {"LD_PRELOAD=/usr/lib64/libpil4dfs.so", "D_LOG_MASK=DEBUG", "DD_SUBSYS=il",
```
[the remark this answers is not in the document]
wiliamhuang: Actually, "LD_PRELOAD=..." here is only a place holder. "envp[0] = str_preload;" in line 39 overrides the string pointer.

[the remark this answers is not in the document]
daltonbohning: Right. I just mean it's misleading to see `LD_PRELOAD=/usr/lib64/libpil4dfs.so` when it's always overridden

daltonbohning (approved):

src/tests/ftest/dfuse/utest/whitelist_test.c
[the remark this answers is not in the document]
wiliamhuang: You are right. I will change it to "LD_PRELOAD=/dummy_path/libpil4dfs.so" and add comment there in next commit. Thank you!

docs/user/filesystem.md
```diff
@@ -1021,6 +1021,14 @@ libpil4dfs intercepting summary for ops on DFS:
 [op_sum ]  5003
 ```
 
+### Turn on whitelist mode in libpil4dfs
+libpil4dfs could boost IO performance considerably over using dfuse for IO intensive workloads. However, initializing daos environment in libpil4dfs is expensive. To alleviate the overhead of libpil4dfs, users could set env to disable function interception in specific applications. I.e., let dfuse handle all IO requests. "D_IL_WHITELIST=1" turns on whitelist mode. A list of common applications are hardcoded to be skipped in whitelist mode. Users can set env "D_IL_BYPASS_ALL_LIST" to bypass function interceptions in target processes and their child processes. "D_IL_BYPASS_LIST" can be set to bypass function interceptions in target processes only. Setting "D_IL_BYPASS_ALL_LIST" or "D_IL_BYPASS_LIST" enables whitelist mode too even "D_IL_WHITELIST=1" is not set. Whitelist mode is recommended for short-lived processes that run a large number (more than 1,000) of times. Example to set env:
```
mchaarawi, line 1025: i thought we want the whitelist mode ON by default for such applications?
it's always handy to add env variables but most of those are going to be for dev use only and in production i bet no one would use them.
The other comment here is that we need to use some other term than whitelist. it's too vague for use here.

as for the other 2 env variables, im not sure why there is any use case where we want to have them separated. again if you think this is valuable for development, let's just hide 1 and expose just 1 of those.

Another thing that is not clear from this description. does the bypass list override the whitlist applications? or does it append to them?

wiliamhuang: My only concern to set whitelist mode as default is that the applications in the built-in bypass list will not be fully tested with libpil4dfs. The bugs in libpil4dfs will not be exposed in those applications. We probably will make he built-in bypass list shorter if we decide to make whitelist mode as default.

You are right. Most regular users do not read user manual carefully and they may be not aware of available env variables. Maybe we can add some description about compatible mode and whitelist mode in frequently asked question section. In case libpil4dfs breaks an application, users can compatible mode and whitelist mode.

> The other comment here is that we need to use some other term than whitelist. it's too vague for use here.
ok. I will rephrase the words here to be more clear.

> as for the other 2 env variables, im not sure why there is any use case where we want to have them separated. again if you think this is valuable for development, let's just hide 1 and expose just 1 of those.

You are right. We can expose only one env variable "D_IL_BYPASS_ALL_LIST" to make it simpler.

> Another thing that is not clear from this description. does the bypass list override the whitlist applications? or does it append to them?
The built-in bypass list can not be overridden. The user provided will be appended. I will make it more clear in document. Thank you!

mchaarawi: > My only concern to set whitelist mode as default is that the applications in the built-in bypass list will not be fully tested with libpil4dfs. The bugs in libpil4dfs will not be exposed in those applications. We probably will make he built-in bypass list shorter if we decide to make whitelist mode as default.
> 
> You are right. Most regular users do not read user manual carefully and they may be not aware of available env variables. Maybe we can add some description about compatible mode and whitelist mode in frequently asked question section. In case libpil4dfs breaks an application, users can compatible mode and whitelist mode.

im going to push back on this.
what is the purpose of this PR? is whitelisting to avoid interception recommended or not? if yes, then this should be the default, if not then this PR should really not be merged or it can be a dev only thing.
No matter how much you publicize env variables, they are not a good option for production. no one is using them unless the admin makes it inherit in user environments, which is the same really as making it default in the first place.

if your concern is testing, then why not run testing with the env variable set (to turn that off)?
in production your goal should be to make things faster and avoid hitting bugs / issues and not finding bugs. the goal for regression and system tests is to find bugs.

wiliamhuang: I agree. I will enable whitelist mode as default and disable it in our testing in CI. Thank you very much!

wiliamhuang:
Now whitelist mode is set as default. Env "D_IL_INTERCEPTION_ON" is set in most CI tests to force function interception enabled. Since fio, io, and mdtest are not in the bypass list, "D_IL_INTERCEPTION_ON" is not added in several related tests.

docs/user/filesystem.md
```diff
@@ -1021,6 +1021,12 @@ libpil4dfs intercepting summary for ops on DFS:
 [op_sum ]  5003
 ```
 
+### Bypassing function interception in libpil4dfs
+libpil4dfs could boost IO performance considerably over using dfuse for IO intensive workloads. However, initializing daos environment in libpil4dfs is expensive. To alleviate the overhead of libpil4dfs, users could set env to disable function interception in specific applications. I.e., let dfuse handle all IO requests. Users can set env "D_IL_BYPASS_LIST" to bypass function interceptions in target processes and their child processes. This is recommended for short-lived processes (shorter than 100 milliseconds) that run a large number (more than 1,000) of times. Example to set env:
```
jolivier23, line 1025: Do you really want to disable for child processes?   Just thinking you may get a situation where client runs their application via a shell script and they want to ignore everything but the application.   In fact, it might be better to have a D_IL_INTERCEPT_LIST (which by default is "all")

wiliamhuang: If interception is turned on only for applications provided by D_IL_INTERCEPT_LIST, users need to have enough knowledge about their workflow. Some users may just take scripts from their colleagues and they are not very familiar which applications have intensive IO. Currently interception is default on for applications except those in "D_IL_BYPASS_LIST" and built-in bypass list.

wiliamhuang: Considering cases like "configure", disabling interception in all child processes would be easier to set up. Otherwise, we need to add all applications involved into the bypass list.

wiliamhuang: @jolivier23 I updated this PR to bypass the interception in individual process only. Children processes are not affected now. Thank you!

src/tests/ftest/dfuse/utest/whitelist_test.c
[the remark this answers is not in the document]
wiliamhuang: Comments were added and "/dummy_path/libpil4dfs.so" is used now. Thank you!

wiliamhuang:
@mchaarawi @jolivier23 @daltonbohning Could you please review this PR? Thank you very much!

src/tests/ftest/dfuse/il_whitelist.py
```diff
@@ -0,0 +1,86 @@
...
+        Test Description:
+            Mount a DFuse mount point
+            Run whitelist_test.
+            Check the number of daos_init() with interception library
+        """
+        env = EnvironmentVariables()
+        if il_lib is not None:
+            env["LD_PRELOAD"] = os.path.join(self.prefix, "lib64", il_lib)
+
+        exe_path = os.path.join(self.prefix, "lib", 'daos/TESTING/tests/whitelist_test')
```
daltonbohning, line 34: I think either
```suggestion
        exe_path = os.path.join(self.prefix, 'lib/daos/TESTING/tests/whitelist_test')
```
or
```suggestion
        exe_path = os.path.join(self.prefix, "lib", 'daos', 'TESTING', 'tests', 'whitelist_test')
```

src/tests/ftest/dfuse/il_whitelist.py
```diff
@@ -0,0 +1,86 @@
...
+        if interception:
+            env["D_IL_INTERCEPTION_ON"] = "1"
+
+        result = run_remote(self.log, dfuse_hosts, env.to_export_str() + exe_path)
+        output = "\n".join(result.all_stdout.values())
+        lines = output.split('\n')
+        num_daos_init = 0
+        for line in lines:
+            if "called daos_init()" in line:
+                num_daos_init += 1
```
daltonbohning, line 59: Not necessary but FYI you could use `re`
```suggestion
        num_daos_init = len(re.findall('called daos_init()', output))
```

src/tests/ftest/util/soak_utils.py
```diff
@@ -1005,6 +1005,7 @@ def create_ior_cmdline(self, job_spec, pool, ppn, nodesperjob, oclass_list=None,
             if api == "POSIX-LIBPIL4DFS":
                 env["LD_PRELOAD"] = os.path.join(self.prefix, 'lib64', 'libpil4dfs.so')
                 env["D_IL_REPORT"] = "1"
+                env["D_IL_INTERCEPTION_ON"] = "1"
```
daltonbohning, line 1008: I'm not sure if we want to do this in soak testing. It might be better for soak to run similar to production/users

src/tests/ftest/dfuse/mu_perms.py
```diff
@@ -301,7 +301,8 @@ def run_test_il(self, il_lib=None):
         env_with_il = env_without_il.copy()
         env_with_il.update({
             'LD_PRELOAD': os.path.join(self.prefix, 'lib64', il_lib),
-            'D_IL_REPORT': -1  # Log all intercepted calls
+            'D_IL_REPORT': -1,  # Log all intercepted calls
+            'D_IL_INTERCEPTION_ON': '1'
```
daltonbohning, line 305: If it's possible for permissions to be reported differently, maybe we shouldn't add it for this test?

daltonbohning (commented):
I'm not sure what it means to set `D_IL_INTERCEPTION_ON` for testing purposes. Should our tests run more like production? Otherwise we aren't testing the same thing users will do

docs/user/filesystem.md
```diff
@@ -1021,17 +1021,16 @@ libpil4dfs intercepting summary for ops on DFS:
...
```
mchaarawi, line 1025: I would reword this a bit. it starts of with 2 contradicting statements:
1) libpil4dfs boost io performance over dfuse
2) libpil4dfs has high overhead

after that you explain shortlive applications, etc. so you should move the explanation of the exception before 2.
so something like (including some other syntax fixes):

libpil4dfs boosts IO performance by bypassing the fuse kernel when going over dfuse for IO intensive workloads. In some scenarios however, for short running applications or simple processes like linux commands, the IO workload does not justify initializing the DAOS environment in user space with libpil4dfs, since this is relatively expensive. These scenarios can be quantified for example with processes that complete in order of tens or hundreds of milliseconds and run for a large number of times.
To alleviate this problem, DAOS avoid function interception by libpil4dfs for specific executables or commands:
**_List them here please_**
In addition, DAOS provides an environment variable (`D_IL_BYPASS_LIST`) to disable function interception by libpil4dfs for specific applications that are set in that env with the following syntax:
```
$ export D_IL_BYPASS_LIST="app_a:app_b:app_c:app_d"
```

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -140,8 +140,6 @@ static _Atomic uint32_t        daos_init_cnt;
  * true.
  */
 static bool             report;
-/* always load libpil4dfs related env variables in exec() */
-static bool             enforce_exec_env;
```
mchaarawi, line 144: can you please explain why you are removing this now?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -1166,6 +1180,7 @@ query_path(const char *szInput, int *is_target_path, struct dcache_rec **parent,
 			}
 
 			rc = daos_init();
+			D_DEBUG(DB_ANY, "called daos_init().\n");
```
mchaarawi, line 1183: this debug statement looks unnecessary, please remove.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -4077,12 +4092,24 @@ new_readdir(DIR *dirp)
 	return &mydir->ents[mydir->num_ents];
 }
 
+static char env_str_int_on[]  = "D_IL_INTERCEPTION_ON=1";
+static char env_str_int_off[] = "D_IL_INTERCEPTION_ON=0";
```
mchaarawi, line 4096: i do not get what this D_IL_INTERCEPTION_ON env is about and why we need it?
when was this added?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -4292,6 +4330,10 @@ pre_envp(char *const envp[], char ***new_envp)
 	return rc;
 }
 
+/* Set this flag to avoid setup_fd_0_1_2() executed more than one time in case exec() fails due to
+ * wrong path.
+ */
+static bool setup_fd_0_1_2_done;
```
mchaarawi, line 4336: are these changes unrelated to this PR? can they be moved to a different PR?

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6675,38 +6739,174 @@ register_handler(int sig, struct sigaction *old_handler)
...
+static char *bypass_bash_cmd_list[] = {"autoconf", "configure", "libtool", "libtoolize",
+				       "lsb_release"};
 
-	exe_short = basename(exe_path);
-	if (exe_short == NULL) {
-		DS_ERROR(errno, "basename() failed");
-		goto out;
+static char *bypass_python3_cmd_list[] = {"scons", "scons-3", "dnf", "dnf-3", "meson"};
+
+static char *bypass_app_list[] = {"arch", "as", "awk", "basename", "bc", "cal", "cat", "chmod",
```
mchaarawi, line 6827: as i mentioned this list should be documented better above.

src/client/dfuse/pil4dfs/int_dfs.c
```diff
@@ -6675,38 +6739,174 @@ register_handler(int sig, struct sigaction *old_handler)
...
+				      "head", "link", "ln", "ls", "kill", "m4", "make", "mkdir",
+				      "mktemp", "mv", "nasm", "yasm", "nm",  "numactl", "patchelf",
+				      "ping", "pkg-config", "ps", "pwd", "ranlib", "readelf",
+				      "readlink", "rename", "rm", "rmdir", "rpm", "sed", "seq",
+				      "size", "sleep", "sort", "ssh", "stat", "strace", "strip",
+				      "su", "sudo", "tail", "tee", "telnet", "time", "top", "touch",
+				      "tr", "truncate", "uname", "vi", "vim", "whoami", "yes"};
+
+static void
+check_whitelist(void)
```
mchaarawi, line 6841: can you please confirm that interception is turned off for everything in that list and user's list for both parent and child processes?

mchaarawi (changes requested):

docs/user/filesystem.md
wiliamhuang: Thank you very much for your rewording and example! I will revise them soon.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: We provide two modes with this PR, whitelist mode and interception-on all time, when libpil4dfs is loaded. Both need that children processes inherit parent process's pil4dfs related env variables. It makes more sense to always set "enforce_exec_env" on. If we expose this variable with env, the state is not well-defined if user sets "enforce_exec_env" off.
I do not see any obvious drawback to append pil4dfs related env variables in children processes if parent process preloads libpil4dfs.so already.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. I will remove it in next commit. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: You previously suggested that we should make whitelist mode as default. Users always run in whitelist mode.
I added "D_IL_INTERCEPTION_ON" as a new env in recent revision of this PR. This is mainly used for testing to make interception all the time. We do not expose this env to users in documentation.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: This seems needed for daos build test in my impression. I can double-check.
execve() could fail when trying different paths to execute commands without absolute path.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: How much information we will provide about these applications? Would a list of them in document sufficient?

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Yes. I did confirm it.
I can add debug log if D_IL_REPORT is true and check the log in ftest.

wiliamhuang:
> I'm not sure what it means to set `D_IL_INTERCEPTION_ON` for testing purposes. Should our tests run more like production? Otherwise we aren't testing the same thing users will do

For production, the interception would be on for most user applications except those linux commands in our bypass list. The involved code are quite small when the interception is bypassed. They are tested by "test_bashcmd_pil4dfs". We allocate major tests for interception enabled (D_IL_INTERCEPTION_ON set 1).

src/tests/ftest/dfuse/il_whitelist.py
wiliamhuang: Thank you very much! I will revise it with the first suggested form then.

src/tests/ftest/dfuse/il_whitelist.py
wiliamhuang: Thank you very much! This is much shorter now. I will update it in next commit.

src/tests/ftest/util/soak_utils.py
wiliamhuang: What applications does soak run? fio, ior, mdtest, and lammps etc? This env is supposed to make no difference. I will remove it as you suggested. Thank you!

src/tests/ftest/dfuse/mu_perms.py
wiliamhuang: Setting 'D_IL_INTERCEPTION_ON' as '1' in this PR actually behaves same as the code in current master in this case (interception is on as default). I think this env will not affect the permissions reported.

daltonbohning:
> > I'm not sure what it means to set `D_IL_INTERCEPTION_ON` for testing purposes. Should our tests run more like production? Otherwise we aren't testing the same thing users will do
> 
> For production, the interception would be on for most user applications except those linux commands in our bypass list. The involved code are quite small when the interception is bypassed. They are tested by "test_bashcmd_pil4dfs". We allocate major tests for interception enabled (D_IL_INTERCEPTION_ON set 1).

If we set `D_IL_INTERCEPTION_ON` in our testing, but not production, then we aren't testing what will happen in production. I don't know if that's an issue here, but just pointing it out.

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: please add a comment here to explain clearly this case.
i still don't get why we need that for testing. when we test, we want to use the whitelist and bypass interception of those, no?
or are you saying you also want to add some tests to override the whitelist just to check those work even if performance is not good?
If the latter, the name of the env is very confusing. to me, if one set LD_PRELOAD that means interception is on with some exceptions. this sounds exactly the same, but the behavior is different. since it's test only, maybe just call it:
D_IL_NO_WHITELIST
making it more clear?

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: ok just wondering if they are related or not to the PR as it would be better to have them separate.

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: IMO just list them above inthe md file where i a dded the bold comment

mchaarawi:
> > > I'm not sure what it means to set `D_IL_INTERCEPTION_ON` for testing purposes. Should our tests run more like production? Otherwise we aren't testing the same thing users will do
> > 
> > 
> > For production, the interception would be on for most user applications except those linux commands in our bypass list. The involved code are quite small when the interception is bypassed. They are tested by "test_bashcmd_pil4dfs". We allocate major tests for interception enabled (D_IL_INTERCEPTION_ON set 1).
> 
> If we set `D_IL_INTERCEPTION_ON` in our testing, but not production, then we aren't testing what will happen in production. I don't know if that's an issue here, but just pointing it out.

indeed. i think this env name is confusing in the first place.
but in most of our testing for pil4dfs we should be only setting LD_PRELOAD. any other env should be just for some targeted testing.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. I will add comments as you suggested.

Yes. We expect users to use whitelist mode all the time to bypass interception in common Linux commands without intensive IO. Interception would be on in whitelist mode if the application is not in the bypass list.

The code of interception disabled are relatively much simpler. A few tests should be sufficient for testing.
If we do not force interception on with the new env, the interception is not enabled in many tests we have. Running more tests with interception is desired to discover bugs, so major tests are allocated for interception enabled.

> or are you saying you also want to add some tests to override the whitelist just to check those work even if performance is not good?
Yes. This is the case. OK. I will rename the env variable to D_IL_NO_WHITELIST. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: ok. Thank you!

mchaarawi:
> Regarding the name of env for testing purpose only, D_NO_IL_BYPASSING, D_IL_INTERCEPTION_ON, D_IL_NO_WHITELIST. I prefer "D_NO_IL_BYPASSING". This sounds more clear. What do you think? Thank you very much!

for consistency, i would say D_IL_*
so if you prefer we can call it D_IL_NO_BYPASS, or D_IL_BYPASS_IGNORE (to ignore the bypass list and internal white list commands)

wiliamhuang:
> for consistency, i would say D_IL_* so if you prefer we can call it D_IL_NO_BYPASS, or D_IL_BYPASS_IGNORE (to ignore the bypass list and internal white list commands)

Thank you very much! We will use "D_IL_NO_BYPASS" then.

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: The debug message is used by a ftest. I added a comment. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: The list is added in document. Thank you!

src/tests/ftest/dfuse/il_whitelist.py
wiliamhuang: Fixed as you suggested. Thank you!

src/tests/ftest/dfuse/il_whitelist.py
wiliamhuang: Fixed as you suggested. Thank you very much!

src/tests/ftest/util/soak_utils.py
wiliamhuang: The env is removed in soak test. Thank you!

docs/user/filesystem.md
wiliamhuang: The document is revised. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Env variable is renamed. Comments are added. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: You are right. The changes here are not necessary. I removed them. Thank you!

src/client/dfuse/pil4dfs/int_dfs.c
wiliamhuang: Interception status (ON/OFF) is checked in the test inside src/tests/ftest/dfuse/il_whitelist.py. "ls" is in built-in bypass list and "whitelist_test" is in user supplied bypass list.

src/client/dfuse/pil4dfs/int_dfs.c
mchaarawi: This debug message for users is useless and will just cause clutter in the debug logs.
if you have a test relying on this message to see if daos_init() is called, you need to update the test.
How about you check what currently daos_init() prints when it's called in INFO and use those in the test?
If there is nothing clear there you can add something as an INFO message in daos_init() itself, not here, towards printing something like:
libdaos-"version" initializing
this does provide something meaningful for users (the version).

wiliamhuang: ok. Thank you! I will remove this line and check for existing debug output, e.g., "crt_init_opt() libcart".

wiliamhuang: Fixed as you suggested. Thank you very much!

wiliamhuang:
> > Now env "D_IL_NO_BYPASS" is used for testing in CI.
> 
> I would like to understand and know which tests in CI are using that env and which are testing the normal way (what users would use usually - using whitelist mode)

White mode is tested in soak, "test_bashcmd_pil4dfs" in src/tests/ftest/dfuse/bash.py, and "test_whitelist_pil4dfs" in src/tests/ftest/dfuse/il_whitelist.py.
All rest tests in CI run with env "D_IL_NO_BYPASS".

mchaarawi:
> > > Now env "D_IL_NO_BYPASS" is used for testing in CI.
> > 
> > 
> > I would like to understand and know which tests in CI are using that env and which are testing the normal way (what users would use usually - using whitelist mode)
> 
> White mode is tested in soak, "test_bashcmd_pil4dfs" in src/tests/ftest/dfuse/bash.py, and "test_whitelist_pil4dfs" in src/tests/ftest/dfuse/il_whitelist.py. All rest tests in CI run with env "D_IL_NO_BYPASS".

IMO we should add all the io tests (fio, ior, mdtest) to the former, not just soak.

wiliamhuang:
> IMO we should add all the io tests (fio, ior, mdtest) to the former, not just soak.

ok. I will make the changes although they are supposed to make not difference since those applications are not in our bypass list.

mchaarawi:
> > IMO we should add all the io tests (fio, ior, mdtest) to the former, not just soak.
> 
> ok. I will make the changes although they are supposed to make any difference since those applications are not in our bypass list.

sure, so why use the env then there ;-)

wiliamhuang:
> > > IMO we should add all the io tests (fio, ior, mdtest) to the former, not just soak.
> > 
> > 
> > ok. I will make the changes although they are supposed to make any difference since those applications are not in our bypass list.
> 
> sure, so why use the env then there ;-)

You are right.
I checked all tests using libpil4dfs. All IO tests (fio, ior, mdtest) currently tested in whitelist mode (not using env "D_IL_NO_BYPASS"). Sorry for the confusion.

Updated list of tests tested in whitelist mode,
fio
ior
mdtest
soak
test_bashcmd_pil4dfs
test_whitelist_pil4dfs

mchaarawi:
> > > > IMO we should add all the io tests (fio, ior, mdtest) to the former, not just soak.
> > > 
> > > 
> > > ok. I will make the changes although they are supposed to make any difference since those applications are not in our bypass list.
> > 
> > 
> > sure, so why use the env then there ;-)
> 
> You are right. I checked all tests using libpil4dfs. All IO tests (fio, ior, mdtest) currently tested in whitelist mode (not using env "D_IL_NO_BYPASS"). Sorry for the inaccurate information in previous reply.
> 
> Updated list of tests tested in whitelist mode, fio ior mdtest soak test_bashcmd_pil4dfs test_whitelist_pil4dfs

ok this is fine with me then. i will leave the test changes review for Dalton to check / approve.

mchaarawi (dismissed):

wiliamhuang:
@daltonbohning @jolivier23 Could you please review? Thank you very much!

daltonbohning (dismissed):

mchaarawi:
i see some linting failures. you want to fix those?

wiliamhuang:
> i see some linting failures. you want to fix those?

I noticed those format related issues previously. I think most of them are not needed to fix. I will not update this PR unless there is anything necessary to be changed. Thank you!

mchaarawi:
> > i see some linting failures. you want to fix those?
> 
> I noticed those format related issues previously. I think most of them are not needed to fix. I will not update this PR unless there is anything necessary to be changed. Thank you!

i see there is one error where a line exceeds 100 characters. sorry but you will need to fix that. thanks in advance

mchaarawi:
> > i see some linting failures. you want to fix those?
> 
> I noticed those format related issues previously. I think most of them are not needed to fix. I will not update this PR unless there is anything necessary to be changed. Thank you!

you should make your life easier and install git-clang-format that does all this automatically for you.

wiliamhuang:
> i see there is one error where a line exceeds 100 characters. sorry but you will need to fix that. thanks in advance

ok. I will fix it. Thank you!

wiliamhuang:
> you should make your life easier and install git-clang-format that does all this automatically for you.

I have never used "git-clang-format" previously. Good to know. I will try it! Thank you very much!

wiliamhuang:
Two places were updated to avoid lines with more than 100 characters.
@mchaarawi @daltonbohning Could you please review again? Thank you very much!

mchaarawi (approved):

mchaarawi:
tests passed in build 19:
https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14812/19/pipeline-graph/
further changes were just clang format issues.

PR #14832, opened 2024-07-27
DAOS-16282 control: Fix some test assumptions about JSON

mjmac:
The commit landed for DAOS-16127 resulted in some differences in JSON output for pool query. Several tests were written in such a way that the code expected the (en|dis)abled_ranks keys to always be set, even if those arrays were NULL. This isn't very idiomatic and is awkward to work with. The test code has been updated to instead use the get() operator which will return None if the response dict does not have the requested key.

Also fixes a problem reported in DAOS-16283, where the JSON output of `dmg pool query` differed from the JSON output of `daos pool query` because it didn't include the usage array.

Features: pool control
Required-githooks: true


tanabarr (approved):

src/tests/ftest/util/pool_create_all_base.py
```diff
@@ -119,7 +119,7 @@ def check_pool_full_storage(self, scm_delta_bytes, nvme_delta_bytes=None, ranks=
         if ranks is not None:
             wait_ranks = sorted(ranks)
             data = self.dmg.pool_query(self.pool[pool_idx].identifier, show_enabled=True)
-            got_ranks = sorted(data['response']['enabled_ranks'])
+            got_ranks = sorted(data['response'].get('enabled_ranks'))
```
daltonbohning, line 122: If `enabled_ranks` is not returned this will raise an exception
```
sorted(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not iterable
```

We could default it to an empty list so `sorted` works
```suggestion
            got_ranks = sorted(data['response'].get('enabled_ranks', []))
```

src/tests/ftest/pool/create_query.py
```diff
@@ -51,7 +51,7 @@ def test_create_and_query(self):
 
         # Query pool created
         resp = self.get_dmg_command().pool_query(pool.identifier, show_enabled=True)["response"]
-        nb_ranks = len(resp["enabled_ranks"])
+        nb_ranks = len(resp.get("enabled_ranks"))
```
daltonbohning, line 54: Similar
```
len(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object of type 'NoneType' has no len()
```

src/tests/ftest/deployment/server_rank_failure.py
```diff
@@ -184,7 +184,7 @@ def verify_rank_failure(self, ior_namespace):
 
         # 9. Call dmg pool query -b to find the disabled ranks.
         output = self.get_dmg_command().pool_query(pool=self.pool.identifier, show_disabled=True)
-        disabled_ranks = output["response"]["disabled_ranks"]
+        disabled_ranks = output["response"].get("disabled_ranks")
```
daltonbohning, line 187: Below, this is used as
```
for disabled_rank in disabled_ranks:
```

And you can't iterate over `None`
```
>>> for i in {}.get('foo'): pass
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not iterable
```

src/tests/ftest/deployment/network_failure.py
```diff
@@ -249,7 +249,7 @@ def verify_network_failure(self, ior_namespace, container_namespace):
 
         # 6. Call dmg pool query -b to find the disabled ranks.
         output = dmg_cmd.pool_query(pool=self.pool.identifier, show_disabled=True)
-        disabled_ranks = output["response"]["disabled_ranks"]
+        disabled_ranks = output["response"].get("disabled_ranks")
```
daltonbohning, line 252: Below uses as
```
for disabled_rank in disabled_ranks:
```

daltonbohning:
@mjmac To clarify on my comments, was the previous behavior
1. To return null/None
    If yes, then my comments don't matter because the code would have raised exceptions before
3. To return an empty list
    If yes, then I think we should handle this at those code points

mjmac:
> 1. To return null/None
>    If yes, then my comments don't matter because the code would have raised exceptions before

That was my understanding... I can test it to confirm, though.

daltonbohning:
@mjmac I did confirm that it returned null/None previously so please disregard my comments. This actually restores the ftest behavior/values to previous `None` logic. Thanks

daltonbohning (approved):

mjmac:
> @mjmac I did confirm that it returned null/None previously so please disregard my comments. This actually restores the ftest behavior/values to previous `None` logic. Thanks

Looks like I missed a couple of pool tests that need to have the expected JSON updated, so I'm going to have to re-push. I could update those tests to get an empty slice instead of `None`, although I'm a little wary of changing existing behavior because it seems like it often turns into a thread-pulling exercise of unintended consequences.

daltonbohning:
> > @mjmac I did confirm that it returned null/None previously so please disregard my comments. This actually restores the ftest behavior/values to previous `None` logic. Thanks
> 
> Looks like I missed a couple of pool tests that need to have the expected JSON updated, so I'm going to have to re-push. I could update the other tests that you pointed at to get an empty slice instead of `None`, although I'm a little wary of changing existing behavior because it seems like it often turns into a thread-pulling exercise of unintended consequences.

Yeah, I'd recommend keeping the behavior as-is :)

kjacque (approved):

kjacque (approved):

daltonbohning (approved):

PR #14849, opened 2024-07-31
DAOS-15849 control: Add client uid map to agent config (#14381)

mjmac:
Allow daos_agent to optionally handle unresolvable client
uids via custom mapping. In deployments where the agent
may not have access to the same user namespace as client
applications (e.g. in containerized deployments), the
client_user_map can provide a fallback mechanism for
resolving the client uids to known usernames for the
purpose of applying ACL permissions tests.

Example agent config:

credential_config:
  client_user_map:
    default:
      user: nobody
      group: nobody
    1000:
      user: joe
      group: blow


tanabarr (approved):

kjacque:
As far as I know, this wasn't approved for 2.6.x.

PR #14852, opened 2024-07-31
DAOS-16013 tools: Query all pool targets by default (#14783)

mjmac:
Running `(daos|dmg) pool query-targets` with just a rank
argument should query all targets on that rank.


mjmac:
Gah, forgot that this can't land until the health check PR lands.

PR #14889, opened 2024-08-07
DAOS-16313 build: Add explicit linker flags to lib/daos/api

mjmac:
Apparently these are needed on Ubuntu.

Required-githooks: true

Change-Id: Ieb0446760f0b53e2f09feeae0226ea26dd455d58

### DIFF
--- a/src/control/lib/daos/api/api.go
+++ b/src/control/lib/daos/api/api.go
@@ -15,7 +15,7 @@ import (
 /*
 #include <daos.h>
 
-#cgo LDFLAGS: -ldaos
+#cgo LDFLAGS: -lcart -lgurt -ldaos -ldaos_common
 */
 import "C"
 
--- a/src/control/lib/daos/api/system.go
+++ b/src/control/lib/daos/api/system.go
@@ -17,7 +17,7 @@ import (
 /*
 #include <daos_mgmt.h>
 
-#cgo LDFLAGS: -ldaos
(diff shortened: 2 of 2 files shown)


.github/workflows/landing-builds.yml
```diff
@@ -1,5 +1,6 @@
 name: Landings
 
+# mjmac test (don't land)
```
tanabarr, line 3: is this not ready for review yet?

tanabarr (changes requested):

.github/workflows/landing-builds.yml
daltonbohning: He modified this file so it would trigger these GHA. They normally only run on landing builds

mjmac:
@daltonbohning: Probably makes sense to just force-land this one. I just reverted the workflow change. The previous build successfully passed all of the normal and landings builds.

daltonbohning (approved):

daltonbohning:
> @daltonbohning: Probably makes sense to just force-land this one. I just reverted the workflow change. The previous build successfully passed all of the normal and landings builds.

Agree

.github/workflows/landing-builds.yml
tanabarr: ah, thanks

tanabarr (approved):

PR #14905, opened 2024-08-09
DAOS-16330 cart: Reduce error logs during timeouts

frostedcmos:
- Reduce burst timeout error prints when multiple rpcs fail at the same time
- Replace error prints with info during rpc state processing, as error is already logged above
- Few cleanups, added crt_gdata_dump() at the end of init.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/cart/crt_context.c
+++ b/src/cart/crt_context.c
@@ -1287,7 +1287,7 @@ crt_context_timeout_check(struct crt_context *crt_ctx)
 			left_to_print--;
 
 			if (err_to_print == 0 && left_to_print > 0)
-				D_ERROR("[%d more rpcs timed out. rest logged at INFO level]\n",
+				D_ERROR(" %d more rpcs timed out. rest logged at INFO level\n",
 					left_to_print);
 
 		} else {
--- a/src/cart/crt_context.c
+++ b/src/cart/crt_context.c
@@ -1059,9 +1059,7 @@ crt_req_timeout_track(struct crt_rpc_priv *rpc_priv)
 	if (rc == 0) {
 		rpc_priv->crp_in_binheap = 1;
 	} else {
-		RPC_ERROR(rpc_priv,
-			  "d_binheap_insert failed, rc: %d\n",
-			  rc);
+		RPC_ERROR(rpc_priv, "d_binheap_insert failed, rc: %d\n", rc);
 		RPC_DECREF(rpc_priv);
 	}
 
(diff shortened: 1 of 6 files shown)


src/cart/crt_context.c
```diff
@@ -1059,9 +1059,7 @@ crt_req_timeout_track(struct crt_rpc_priv *rpc_priv)
 	if (rc == 0) {
 		rpc_priv->crp_in_binheap = 1;
 	} else {
-		RPC_ERROR(rpc_priv,
-			  "d_binheap_insert failed, rc: %d\n",
-			  rc);
+		RPC_ERROR(rpc_priv, "d_binheap_insert failed, rc: %d\n", rc);
```
jolivier23, line 1062: May I suggest we start using DL_ERROR internally here (or add another version of RPC_ERROR that takes rc as a parameter).  It will print string version of error

jolivier23 (dismissed):

src/cart/crt_internal_types.h
```diff
@@ -119,16 +119,13 @@ struct crt_gdata {
...
-				cg_use_sensors		: 1,
-				/** whether we are on a primary provider */
-				cg_provider_is_primary	: 1;
+	unsigned int cg_inited : 1, cg_grp_inited : 1, cg_swim_inited : 1, cg_auto_swim_disable : 1,
+	    /** whether it is a client or server */
+	    cg_server              : 1,
+	    /** whether metrics are used */
+	    cg_use_sensors         : 1,
+	    /** whether we are on a primary provider */
+	    cg_provider_is_primary : 1;
```
soumagne, line 128: it would be great if you were using `unsigned int` on each field to fix the format or just start using bool instead :)

frostedcmos: > it would be great if you were using `unsigned int` on each field to fix the format or just start using bool instead :)

we use bitfields here as we have to be careful about the size of this rpc_priv struct. its annoying for formatting though

src/cart/crt_context.c
frostedcmos: I dont think we ever hit this particular error; this would be something fatal.  I agree though we should use DL_ERROR in places where it makes sense

src/cart/crt_internal_types.h
soumagne: maybe I was not clear, what I'm saying is that you should not separate each field with `,` but instead repeat `unsigned int` for each field so that they remain aligned.
```C
	    /** whether metrics are used */
	    unsigned int cg_use_sensors         : 1;
	    /** whether we are on a primary provider */
	    unsigned int cg_provider_is_primary : 1;
```
does that make sense ?

soumagne: also fwiw this is the `struct crt_gdata` not the rpc_priv struct...

frostedcmos: > maybe I was not clear, what I'm saying is that you should not separate each field with `,` but instead repeat `unsigned int` for each field so that they remain aligned.
> 
> ```c
> 	    /** whether metrics are used */
> 	    unsigned int cg_use_sensors         : 1;
> 	    /** whether we are on a primary provider */
> 	    unsigned int cg_provider_is_primary : 1;
> ```
> 
> does that make sense ?

Neat, i didnt know it allowed such style

jolivier23 (approved):

jgmoore-or (approved):

PR #14937, opened 2024-08-15
DAOS-16220 test: Include username in default common test directory

phender:
To support multiple users testing in the same environment without sudo access the default common test directory will be made unique by including the user ID in the path.

Skip-unit-tests: true

Required-githooks: true

### Gatekeeper:


src/tests/ftest/launch.py
```diff
@@ -733,10 +763,26 @@ def main():
...
         args.clear_mounts.append("/mnt/daos1")
 
+    elif args.mode == "custom_a":
+        if args.agent_user is None:
+            # Run the agent with the current user by default
+            args.agent_user = getpass.getuser()
+        if os.environ.get("DAOS_TEST_LOG_DIR", args.test_log_dir) is None:
+            # Use a user-specific test log dir by default
+            args.test_log_dir = os.path.join(
+                os.sep, "var", "tmp", f"daos_testing_{args.agent_user}")
```
daltonbohning, line 778: I think it's better to always use `getpass.getuser()` here. If for whatever reason `agent_user` is different, I think the directory should be named for the running user
```suggestion
            args.test_log_dir = os.path.join(
                os.sep, "var", "tmp", f"daos_testing_{getpass.getuser()}")
```

phender:
Confirmed running the daos_agent as the user in https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14937/29/

src/tests/ftest/launch.py
```diff
@@ -744,10 +774,26 @@ def main():
...
         args.clear_mounts.append("/mnt/daos1")
 
+    elif args.mode == "custom_a":
+        if args.agent_user is None:
+            # Run the agent with the current user by default
+            args.agent_user = getpass.getuser()
+        if os.environ.get("DAOS_TEST_LOG_DIR", args.test_log_dir) is None:
+            # Use a user-specific test log dir by default
+            args.test_log_dir = os.path.join(
+                os.sep, "var", "tmp", f"daos_testing_{getpass.getuser()}")
```
daltonbohning, line 789: Because we instantiate TestEnvironment before setting this, the config paths aren't under this directory

src/tests/ftest/util/command_utils.py
```diff
@@ -894,6 +893,9 @@ def __init__(self, namespace, command, path="", yaml_cfg=None, timeout=10, run_u
         # Default owner of the certificate files
         self.certificate_owner = getuser()
 
+        # Support disabling verifying the socket directory (runtime_dir) for tests
+        self.verify_socket_dir = True
```
daltonbohning, line 897: Looks like this is always True?

src/tests/ftest/util/file_utils.py
```diff
@@ -68,6 +68,31 @@ def change_file_owner(logger, hosts, filename, owner, group, timeout=15, verbose
     return __run_command(logger, hosts, command, verbose, timeout)
 
 
+def create_owned_directory(logger, hosts, directory, timeout=15, verbose=True, owner=None,
```
daltonbohning, line 71: Couldn't we add an argument to create_directory to handle this? This is the same function but conditionally calls chown afterwards

src/tests/ftest/util/server_utils_params.py
```diff
@@ -501,6 +501,7 @@ def __init__(self, base_namespace, index, provider=None, max_storage_tiers=MAX_S
             if name in self.REQUIRED_ENV_VARS:
                 default_env_vars.extend(self.REQUIRED_ENV_VARS[name])
         self.env_vars = BasicParameter(None, default_env_vars)
+        self.env_pass_through = BasicParameter(None, None)
```
daltonbohning, line 504: If we update the systemctl env with LD_LIBRARY_PATH, then we need to pass through LD_LIBRARY_PATH here or otherwise it won't work as expected

src/tests/ftest/util/command_utils.py
phender: Normally, yes, unless `SubprocessManager.verify_socket_dir = False` is called, like in `test_daos_agent_config_basic`

daltonbohning: I missed that this was just moved from one class to another

src/tests/ftest/util/file_utils.py
phender: Yes, that's a good idea.

src/tests/ftest/util/command_utils.py
```diff
@@ -1062,38 +1064,41 @@ def copy_configuration(self, hosts):
...
-                result = change_file_owner(
-                    self.log, nodes, directory, user, get_primary_group(user), user="root")
-                if not result.passed:
-                    raise CommandFailure(
-                        f"{self.command}: error setting socket directory {directory} owner for "
-                        f"user {user} on {result.failed_hosts}")
+                "Creating %s socket directory %s for user %s on %s",
+                self.command, directory, user, nodes)
+            result = create_directory(
+                self.log, nodes, directory, user=user, owner="root" if privileged else None)
```
daltonbohning, line 1097: I think this. The owner should always be the user. But we need to run as root if the directory is privileged, or if the owner is not the current user
```suggestion
            result = create_directory(
                self.log, nodes, directory,
                user="root" if (privileged or user != getuser()) else None,
                owner=user)
```

daltonbohning: Actually, I think this.
```suggestion
            result = create_directory(
	                self.log, nodes, directory,
	                user="root" if privileged else user,
	                owner=user)
```

phender:
The errors seen in https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14937/33/testReport/ are known issues:
- https://daosio.atlassian.net/browse/DAOS-16495
- https://daosio.atlassian.net/browse/DAOS-16496

src/tests/ftest/util/command_utils.py
phender: Thanks

src/tests/ftest/launch.py
phender: This has been addressed in https://github.com/daos-stack/daos/blob/46b0765eebb7f99463c0fcb8417e5ccea69344cc/src/tests/ftest/util/environment_utils.py#L164

src/tests/ftest/launch.py
phender: Addressed by no longer setting defaults when instantiating the `TestEnvironment` object.

src/tests/ftest/util/launch_utils.py
```diff
@@ -614,9 +613,13 @@ def _setup_test_directory(self, logger, test):
             directories.append(os.path.join(test_env.log_dir, directory))
         commands.append(f"mkdir -p {' '.join(directories)}")
         commands.append(f"ls -al {test_env.log_dir}")
-        for command in commands:
-            if not run_remote(logger, hosts, command).passed:
-                message = "Error setting up the common test directory on all hosts"
+        for host in "client", "server":
```
phender, line 616: Probably should have used `role` here instead of `host`

src/tests/ftest/util/cart_ctl_utils.py
```diff
@@ -82,4 +84,8 @@ def run(self):
         """
         self.log.info('Starting CartCtl')
 
+        drpc_dir = os.environ.get("DAOS_AGENT_DRPC_DIR", None)
+        if 'DAOS_AGENT_DRPC_DIR' not in self.env and drpc_dir is not None:
+            self.env['DAOS_AGENT_DRPC_DIR'] = str(drpc_dir)
```
daltonbohning, line 89: We shouldn't need to do this since it's ultimately ran with `run_command`, which uses `process.run`, which uses `os.environ`

src/tests/ftest/util/job_manager_utils.py
```diff
@@ -143,6 +143,18 @@ def job(self, value):
...
+    def command_regex(self):
+        """Get the regular expression to use to search for the command.
+
+        Typical use would include combining with pgrep to verify a subprocess is running.
+
+        Returns:
+            str: regular expression to use to search for the command
+
+        """
+        return "'({})'".format("|".join(self._exe_names + self.job._exe_names))
```
daltonbohning, line 156: > protected-access, Access to a protected member _exe_names of a client class

src/tests/ftest/util/server_utils.py
```diff
@@ -45,7 +46,7 @@ def get_server_command(group, cert_dir, bin_dir, config_file, config_temp=None):
     transport_config = DaosServerTransportCredentials(cert_dir)
     common_config = CommonConfig(group, transport_config)
     config = DaosServerYamlParameters(config_file, common_config)
-    command = DaosServerCommand(bin_dir, config, None)
+    command = DaosServerCommand(bin_dir, config, None, run_user)
```
daltonbohning, line 49: Not a big deal, but the server fundamentally cannot be ran as a non-privileged user so this will probably never be used

src/tests/ftest/util/cart_ctl_utils.py
phender: Removed

src/tests/ftest/util/job_manager_utils.py
phender: Removed, will address in DAOS-16100

src/tests/ftest/util/server_utils.py
phender: Understood, this is more to keep the code aligned and to avoid any special handling where an attribute may not be set.

phender:
Failures in https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-14937/44/testReport/ are due to known issues:
- https://daosio.atlassian.net/browse/DAOS-16674

daltonbohning (approved):

src/tests/ftest/util/general_utils.py
```diff
@@ -444,7 +448,7 @@ def get_host_data(hosts, command, text, error, timeout=None):
     data_error = "[ERROR]"
 
     # Find the data for each specified servers
-    log.info("  Obtaining %s data on %s", text, hosts)
+    log.info("  Obtaining %s data on %s using: %s", text, hosts, command)
```
phender, line 451: This is to compensate for the loss of logging the journalctl command in soak's updated get_journalctl().

PR #14984, opened 2024-08-22
DAOS-16329 chk: maintenance mode after checking pool with dryrun

Nasf-Fan:
Sometimes, after system shutdown unexpectedly, the users may expect to check their critical data under some kind of maintenance mode. Under such mode, no user data can be modified or moved or aggregated. That will guarantee no further potential (DAOS logic caused) damage can happen during the check.

For such purpose, we will enhance current DAOS CR logic with --dryrun option to allow the pool (after check) to be opened as immutable with disabling some mechanism that may potentially cause data modification or movement (such as rebuild or aggregation).

Under such mode, if client wants to connect to the pool, the read-only option must be specified. Similarly for opening container in such pool.

Test-tag: pr cat_recov
Allow-unstable-test: true

### Gatekeeper:

### DIFF
--- a/src/chk/chk_engine.c
+++ b/src/chk/chk_engine.c
@@ -1797,10 +1797,8 @@ cont:
 	}
 
 	rc = chk_engine_cont_cleanup(cpr, svc, &aggregator);
-	if (rc != 0)
-		goto out;
-
-	rc = ds_pool_svc_schedule_reconf(svc);
+	if (rc == 0 && !cpr->cpr_immutable)
+		rc = ds_pool_svc_schedule_reconf(svc);
 
 out:
 	chk_engine_cont_list_fini(&aggregator);
@@ -2112,6 +2110,11 @@ chk_engine_start_post(struct chk_instance *ins)
 
 		if (pool_cbk->cb_phase == CHK__CHECK_SCAN_PHASE__CSP_DONE)
 			continue;
+
+		if (ins->ci_prop.cp_flags & CHK__CHECK_FLAG__CF_DRYRUN)
+			cpr->cpr_immutable = 1;
+		else
+			cpr->cpr_immutable = 0;
 
 		if (phase > pool_cbk->cb_phase)
 			phase = pool_cbk->cb_phase;
@@ -2950,7 +2953,7 @@ chk_engine_pool_start(uint64_t gen, uuid_t uuid, uint32_t phase, uint32_t flags)
 	cbk = &cpr->cpr_bk;
 	chk_pool_get(cpr);
 
-	rc = ds_pool_start(uuid, false);
+	rc = ds_pool_start(uuid, false, cpr->cpr_immutable);
 	if (rc != 0)
 		D_GOTO(put, rc = (rc == -DER_NONEXIST ? 1 : rc));
 
--- a/src/security/srv_acl.c
+++ b/src/security/srv_acl.c
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright 2019-2023 Intel Corporation.
+ * (C) Copyright 2019-2024 Intel Corporation.
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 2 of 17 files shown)


src/tests/ftest/cart/util/cart_logtest.py
```diff
@@ -209,7 +209,7 @@ def subtract(self, val):
                "DER_NOTLEADER(-2008): 'Not service leader'")
 
 # Functions that are never reported as errors.
-IGNORED_FUNCTIONS = ('sched_watchdog_post', 'rdb_timerd')
+IGNORED_FUNCTIONS = ('sched_watchdog_post', 'rdb_timerd', 'cont_open')
```
daltonbohning, line 212: Why are we ignoring `cont_open` here instead of fixing the errors? And what are the errors we're ignoring?

daltonbohning (dismissed):

src/security/srv_acl.c
```diff
@@ -499,8 +498,20 @@ ds_sec_pool_get_capabilities(uint64_t flags, d_iov_t *cred,
 
 	rc = get_sec_capas_for_token(token, ownership, acl, 0 /* no special owner perms */,
 				     pool_capas_from_perms, capas);
-	if (rc == 0)
+	if (rc == 0) {
 		filter_pool_capas_based_on_flags(flags, capas);
+		if (immutable) {
```
kjacque, line 503: simpler than this, I wonder about checking this at the handler level:
```c
if (immutable && (flags != DAOS_PC_RO)) {
        D_ERROR("caller requested RW permissions on immutable pool\n");
        return -DER_NO_PERM;
}
```
Then I don't think any change would be needed in the security capas code at all. If you make the simpler check early in the RPC handlers, only RO requests will get through to this point, so the capas won't need to be further filtered.

kjacque (changes requested):

src/chk/chk_engine.c
```diff
@@ -1757,7 +1757,8 @@ chk_engine_pool_ult(void *args)
 	 * Cleanup all old connections. It is no matter even if we cannot evict some
 	 * old connections. That is also independent from former check phases result.
 	 */
-	ds_pool_svc_evict_all(svc);
+	if (!cpr->cpr_immutable)
+		ds_pool_svc_evict_all(svc);
```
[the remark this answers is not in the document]
Nasf-Fan: You are right, seems that we have to evict the old handles. I will fix it.

src/security/srv_acl.c
Nasf-Fan: Yes, I have ever implemented like that, it can work. But as former discussion, we hoped to unify the check `if (immutable && (flags != DAOS_PC_RO))` into `ds_sec_pool_can_connect()`. That will change `ds_sec_pool_can_connect()`. So either `ds_sec_pool_get_capabilities()` or `ds_sec_pool_can_connect()` needs some enhancement to support `immutable` feature. I prefer to change `ds_sec_pool_get_capabilities()`. Because if still want to implement my original idea about the security capability `ds_sec_pool_get_capabilities()` in future, then we do not need to re-change related security interfaces.

kjacque: On seeing the code, I guess the question for me is whether it is useful to unify the immutable check into the capas calculation at all. We don't even need to get the ACL/owner props to know the user has made an invalid request, asking for RW permissions to an immutable pool. It feels almost like an input consistency check, rather than a capability calculation.

Nasf-Fan: OK, I will move `immutable` check out side of security module, instead, related RPC handler (for pool connection and container open) will handle that explicitly.

wangshilong (dismissed):

kjacque (approved):

wangshilong (approved):

daltonbohning (commented):
Will reserve approval for those familiar with core code, I think we will need to force land for the existing NLT failure

daltonbohning:
I'm not familiar enough to merge myself without reviews from relevant @daos-stack/metadata-owners and @daos-stack/client-api-owners. Will leave this to another gatekeeper or let those owners approve first.

Nasf-Fan:
> I'm not familiar enough to merge myself without reviews from relevant @daos-stack/metadata-owners and @daos-stack/client-api-owners. Will leave this to another gatekeeper or let those owners approve first.

Ping @gnailzenh @NiuYawei @daos-stack/daos-gatekeeper , thanks!

gnailzenh (approved):

PR #14999, opened 2024-08-24
DAOS-16445 client: Add function to cycle OIDs non-sequentially

jolivier23:
We've noticed that with sequential order, object placement is poor.

We get 40% fill for 8GiB files with 25 ranks and 16 targets per rank
with EC_2P1G8.  With this patch, we get a much better distribution.

This patch adds the following:

1. A function for cycling oid.hi incrementing by a large prime
2. For DFS, randomize the starting value
3. Modify DFS to cycle OIDs using the new function.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/client/dfs/dfs_internal.h
+++ b/src/client/dfs/dfs_internal.h
@@ -102,6 +102,18 @@
 /** MAX value for the HI OID */
 #define MAX_OID_HI         ((1UL << 32) - 1)
 
+/** Use a large prime for cycling through all oids */
+#define OID_INC            (999999937)
+
+static inline void
+oid_inc(daos_obj_id_t *oid)
+{
+	oid->hi = (oid->hi + OID_INC) & MAX_OID_HI;
+	/** Skip reserved entries */
+	if (oid->lo == RESERVED_LO && oid->hi <= 1)
+		oid->hi += OID_INC;
+}
+
 typedef uint64_t dfs_magic_t;
 typedef uint16_t dfs_sb_ver_t;
 typedef uint16_t dfs_layout_ver_t;
@@ -164,6 +176,8 @@ struct dfs {
 	daos_handle_t        coh;
 	/** refcount on cont handle that through the DFS API */
 	uint32_t             coh_refcount;
+	/** The last oid.hi in the sequence */
+	uint32_t             last_hi;
 	/** Transaction handle epoch. DAOS_EPOCH_MAX for DAOS_TX_NONE */
 	daos_epoch_t         th_epoch;
 	/** Transaction handle */
@@ -343,7 +357,7 @@ oid_gen(dfs_t *dfs, daos_oclass_id_t oclass, bool file, daos_obj_id_t *oid)
 
 	D_MUTEX_LOCK(&dfs->lock);
 	/** If we ran out of local OIDs, alloc one from the container */
-	if (dfs->oid.hi >= MAX_OID_HI) {
+	if (dfs->oid.hi == dfs->last_hi) {
 		/** Allocate an OID for the namespace */
 		rc = daos_cont_alloc_oids(dfs->coh, 1, &dfs->oid.lo, NULL);
 		if (rc) {
@@ -351,12 +365,14 @@ oid_gen(dfs_t *dfs, daos_oclass_id_t oclass, bool file, daos_obj_id_t *oid)
 			D_MUTEX_UNLOCK(&dfs->lock);
 			return daos_der2errno(rc);
 		}
-		dfs->oid.hi = 0;
+		/** Start such that dfs->last_hi will be final value */
+		dfs->oid.hi = dfs->last_hi;
 	}
 
 	/** set oid and lo, bump the current hi value */
 	oid->lo = dfs->oid.lo;
-	oid->hi = dfs->oid.hi++;
+	oid_inc(&dfs->oid);
+	oid->hi = dfs->oid.hi;
 	D_MUTEX_UNLOCK(&dfs->lock);
 
 	/** if a regular file, use UINT64 typed dkeys for the array object */
--- a/src/client/dfs/dfs_internal.h
+++ b/src/client/dfs/dfs_internal.h
@@ -99,21 +99,6 @@
 /** Max recursion depth for symlinks */
 #define DFS_MAX_RECURSION  40
 
-/** MAX value for the HI OID */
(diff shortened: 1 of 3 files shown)


src/client/dfs/dfs_internal.h
```diff
@@ -337,20 +346,21 @@ oid_gen(dfs_t *dfs, daos_oclass_id_t oclass, bool file, daos_obj_id_t *oid)
 
 	D_MUTEX_LOCK(&dfs->lock);
 	/** If we ran out of local OIDs, alloc one from the container */
-	if (dfs->oid.hi >= MAX_OID_HI) {
+	if (dfs->oid.hi == 0) {
```
jolivier23, line 349: I guess this isn't necessarily better...also need to start at a random place potentially

jolivier23:
@mchaarawi In our simulations, this does quite a bit better.

mchaarawi:
I am uncertain as to why this resolves things.
it sounds to me like this issue needs to be fixed at a lower level and not in the POSIX or other middleware layers.

jolivier23:
> I am uncertain as to why this resolves things. it sounds to me like this issue needs to be fixed at a lower level and not in the POSIX or other middleware layers.

I agree to some extent.   It sounds like the placement algorithm doesn't work very well with mostly sequential oid allocation (e.g. not even close to balanced placement).   I'm not sure how to fix the allocation order in any other way than in the client logic that cycles through them.   If we want to fix placement hash such that it works better with sequentially allocated oids, that is probably best but also a layout change.

jolivier23:
> > I am uncertain as to why this resolves things. it sounds to me like this issue needs to be fixed at a lower level and not in the POSIX or other middleware layers.
> 
> I agree to some extent. It sounds like the placement algorithm doesn't work very well with mostly sequential oid allocation (e.g. not even close to balanced placement). I'm not sure how to fix the allocation order in any other way than in the client logic that cycles through them. If we want to fix placement hash such that it works better with sequentially allocated oids, that is probably best but also a layout change.

It may be good to have this be more configurable so we can play with settings.

I suppose one other thing we could do is to do something similar with oid.lo but I suspect that may still have this issue where sequential allocation doesn't place well.

mchaarawi:
> > > I am uncertain as to why this resolves things. it sounds to me like this issue needs to be fixed at a lower level and not in the POSIX or other middleware layers.
> > 
> > 
> > I agree to some extent. It sounds like the placement algorithm doesn't work very well with mostly sequential oid allocation (e.g. not even close to balanced placement). I'm not sure how to fix the allocation order in any other way than in the client logic that cycles through them. If we want to fix placement hash such that it works better with sequentially allocated oids, that is probably best but also a layout change.
> 
> It may be good to have this be more configurable so we can play with settings.
> 
> I suppose one other thing we could do is to do something similar with oid.lo but I suspect that may still have this issue where sequential allocation doesn't place well.

if you touch oid.lo then you are introducing a dfs layout change and will have to add compatibility code.
im more worried about non DFS use cases.. it sounds like this is a more generic problem that needs a better solution IMO.

jolivier23:
> > > > I am uncertain as to why this resolves things. it sounds to me like this issue needs to be fixed at a lower level and not in the POSIX or other middleware layers.
> > > 
> > > 
> > > I agree to some extent. It sounds like the placement algorithm doesn't work very well with mostly sequential oid allocation (e.g. not even close to balanced placement). I'm not sure how to fix the allocation order in any other way than in the client logic that cycles through them. If we want to fix placement hash such that it works better with sequentially allocated oids, that is probably best but also a layout change.
> > 
> > 
> > It may be good to have this be more configurable so we can play with settings.
> > I suppose one other thing we could do is to do something similar with oid.lo but I suspect that may still have this issue where sequential allocation doesn't place well.
> 
> if you touch oid.lo then you are introducing a dfs layout change and will have to add compatibility code. im more worried about non DFS use cases.. it sounds like this is a more generic problem that needs a better solution IMO.

Only option I can think of is making a generic version of oid_gen, something like

daos_cont_oid_gen(coh, &oid);

which basically does what oid_gen does.  I can take a crack at it.  We may need to cherry-pick this patch in the near term though.

and maybe something like
daos_cont_oid_init(coh, num_reserved) to reserve the first few oids in oid.lo==0

jolivier23:
We get 80% utilization before ENOSPC with 64GiB files with this patch.  Previously, 8GiB files were hitting ENOSPC at around 40% utilization.   We are testing 8GiB files next but this looks promising.

mchaarawi (approved):
it's still surprising to me why the sequential allocation does such a bad job, but this is OK for me and does not introduce layout compatibility issues with existing containers.

jolivier23:
> it's still surprising to me why the sequential allocation does such a bad job, but this is OK for me and does not introduce layout compatibility issues with existing containers.

Agreed, should probably open a ticket on this at some point though maybe not that important with a workaround.  I suppose we should call this out in the docs somewhere that this is important.  @Michael-Hennecke do you have a place that would fit?

src/include/daos_obj.h
```diff
@@ -564,6 +564,21 @@ daos_obj_generate_oid(daos_handle_t coh, daos_obj_id_t *oid,
 		      enum daos_otype_t type, daos_oclass_id_t cid,
 		      daos_oclass_hints_t hints, uint32_t args);
 
+
+/**
+ * This function, if called 2^32 times will set oid->hi to every unique 32-bit
```
karthjyojay, line 568: I'm confused by the first sentence of this method. Are oids 32 bit integers which means there are 2^32 possible combinations? Is this function just getting us a unique `hi` every time it is called? When does it get called?

karthjyojay (approved):

src/include/daos_obj.h
jolivier23: The oid has 2 64-bit integers, lo and hi.   The lower 96 bits (all of lo and the lo 32-bits of hi) are supposed to be unique within a container for a given object.   DAOS provides a method that allocates a globally unique lo oid upon request for the container.   The user, in this case DFS, then has 2^32 unique oids it can allocate to produce a globally unique oid within the container.   When it runs out, it will request a new lo (it probably won't run out as that would require the process to create 2^32 + 1 objects).

This function is just a helper function to cycle through the range of available oids.

jolivier23: The upper 32-bits of hi are used by DAOS to encode things like the object class

PR #15037, opened 2024-08-29
DAOS-16467 rebuild: add DAOS_POOL_RF ENV for massive failure case

liuxuezhao:
Allow user to set DAOS_POOL_RF as pool RF.
    If SWIM detected engine failure is going to break pool RF, don't change
    pool map, also don't trigger rebuild.
    With critical log message to ask administrator to bring back those
    engines in top priority (just "system start --ranks=xxx", need not to
    reintegrate those engines).
    
    a few functions renamed to avoid confuse -
    pool_map_find_nodes() -> pool_map_find_ranks()
    pool_map_find_node_by_rank() -> pool_map_find_dom_by_rank()
    pool_map_node_nr() -> pool_map_rank_nr()
    
### Gatekeeper:


src/pool/srv_pool.c
```diff
@@ -6500,6 +6500,43 @@ pool_svc_schedule_reconf(struct pool_svc *svc, struct pool_map *map, uint32_t ma
...
+	primary_grp = crt_group_lookup(NULL);
+	D_ASSERT(primary_grp != NULL);
+
+	D_CRIT("!!! Please check and bring back these engines in top priority -\n");
+	D_CRIT("!!! To-be-fixed engine rank %d\n", rank);
+	for (i = 0; i < doms_cnt; i++) {
+		struct swim_member_state state;
+
+		if (!(doms[i].do_comp.co_status & PO_COMP_ST_UPIN))
+			continue;
```
Nasf-Fan, line 6524: Skip `healthy` or `unhealthy` ?

src/pool/srv_pool.c
```diff
@@ -6653,6 +6691,31 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc,
...
+		d_rank_t	rank;
+		int		failed_cnt;
+
+		rc = pool_map_update_failed_cnt(map);
+		if (rc) {
+			DL_ERROR(rc, "pool_map_update_failed_cnt failed.");
+			goto out_map;
+		}
+
+		D_ASSERT(tgt_addrs->pta_number == 1);
```
Nasf-Fan, line 6707: So swim can guarantee to exclude ranks one by one? For example, in Aurora configuration, if some server node crashed, then two ranks will be excluded in order or together?

src/pool/srv_pool.c
```diff
@@ -6653,6 +6691,31 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc,
...
+		rank = tgt_addrs->pta_addrs->pta_rank;
+		failed_cnt = pool_map_get_failed_cnt(map, PO_COMP_TP_NODE);
+		D_INFO("SWIM exclude rank %d, failed_cnt %d\n", rank, failed_cnt);
+		if (failed_cnt > pw_rf) {
```
Nasf-Fan, line 6711: So `pw_rf` is based on `rank` or `node` ?

src/pool/srv_target.c
```diff
@@ -1887,6 +1887,7 @@ ds_pool_tgt_map_update(struct ds_pool *pool, struct pool_buf *buf,
 			/* Ignore DTX resync failure that is not fatal. */
 			D_WARN("dtx_resync_ult failure %d\n", ret);
 			D_FREE(arg);
+			goto out;
```
Nasf-Fan, line 1890: Seems unnecessary.

Nasf-Fan (commented):
On some degree, to make the semantics to be more clear and avoid confusing with container RF, we can use some new concept instead of pool wise RF, for example: max concurrent (include both in-flight and pending) rebuild or DAOS Fuse Threshold. Then original container RF semantics will not be changed, and can be set as any value as the user wants; the MCR (or DFT) is not related with whether lost data or not, instead, it is some kinds of threshold for DAOS fuse mechanism to avoid massive damage.

If the use specifies smaller container RF than DFT, then himself/herself will take the risk of losing data. If he/she sets larger container RF than the DFT, then we (DAOS) guarantee that there will be no data lost (unless the unrecoverable hardware failure exceeds such RF) and the redundancy that exceeds DFT part will be read only under extreme case.

How do you think?

src/pool/srv_pool.c
liuxuezhao: this is to check UPIN rank in poolmap, if it is lost/down in primary group.
if it is not UPIN (unhealthy), need not check it.

src/pool/srv_pool.c
liuxuezhao: it is in the path of handle_event() -> pool_svc_exclude_rank() -> pool_svc_update_map() -> pool_svc_update_map_internal().
In pool_svc_exclude_rank() it always set list.pta_number = 1 to call pool_svc_update_map now.
So I think here it is safe to assert it for current code.

src/pool/srv_pool.c
liuxuezhao: on NODE.

src/pool/srv_target.c
liuxuezhao: hmm,  I was not intended to change it, maybe some failure when rebasing, I'll remove it.

liuxuezhao:
> On some degree, to make the semantics to be more clear and avoid confusing with container RF, we can use some new concept instead of pool wise RF, for example: max concurrent (include both in-flight and pending) rebuild or DAOS Fuse Threshold. Then original container RF semantics will not be changed, and can be set as any value as the user wants; the MCR (or DFT) is not related with whether lost data or not, instead, it is some kinds of threshold for DAOS fuse mechanism to avoid massive damage.
> 
> If the use specifies smaller container RF than DFT, then himself/herself will take the risk of losing data. If he/she sets larger container RF than the DFT, then we (DAOS) guarantee that there will be no data lost (unless the unrecoverable hardware failure exceeds such RF) and the redundancy that exceeds DFT part will be read only under extreme case.
> 
> How do you think?

One point to I heard we agreed to change RF to be - all container will take pool's RF or 0 when creating. So likely we will have pool wise RF in future. But I am not good at naming, we can change it later with discuss with others, I'll refresh the patch to only remove the "goto" currently.

src/pool/srv_pool.c
```diff
@@ -6500,6 +6500,45 @@ pool_svc_schedule_reconf(struct pool_svc *svc, struct pool_map *map, uint32_t ma
...
+		if (rc != 0 && rc != -DER_NONEXIST) {
+			D_ERROR("failed to get status of rank %u: %d\n",
+				doms[i].do_comp.co_rank, rc);
+			break;
+		}
+
+		D_DEBUG(DB_MD, "rank/state %d/%d\n", doms[i].do_comp.co_rank,
+			rc == -DER_NONEXIST ? -1 : state.sms_status);
+		if (rc == -DER_NONEXIST || state.sms_status == SWIM_MEMBER_DEAD)
+			D_CRIT("!!! To-be-fixed engine rank %d\n", doms[i].do_comp.co_rank);
```
liw, line 6538: Should log `DF_UUID": intolerable unavailability: engine rank %u\n"`, followed by an explanation in the docs what that means and how to deal with it. I think these would require too many words to explain clearly in log messages. On the other hand, having a short, uncommon phrase in the log message would make searching simple: For instance, googling "daos intolerable unavailability" finds the doc, and grepping "intolerable unavailability" finds the log messages.

src/pool/srv_pool.c
```diff
@@ -6653,6 +6693,31 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc,
...
+
+		D_ASSERT(tgt_addrs->pta_number == 1);
+		rank = tgt_addrs->pta_addrs->pta_rank;
+		failed_cnt = pool_map_get_failed_cnt(map, PO_COMP_TP_NODE);
+		D_INFO("SWIM exclude rank %d, failed NODE %d\n", rank, failed_cnt);
+		if (failed_cnt > pw_rf) {
+			D_CRIT("exclude rank %d will break pw_rf %d, failed_cnt %d\n",
+			       rank, pw_rf, failed_cnt);
+			pool_map_crit_prompt(svc, map, rank);
+			goto out_map;
```
liw, line 6717: Should we set `rc` to some error?

liw (commented):
Thanks, Xuezhao. The "[Nit]s" are not blockers.

src/pool/srv_pool.c
```diff
@@ -6500,6 +6500,45 @@ pool_svc_schedule_reconf(struct pool_svc *svc, struct pool_map *map, uint32_t ma
 	return 0;
 }
 
+static void
+pool_map_crit_prompt(struct pool_svc *svc, struct pool_map *map, d_rank_t rank)
```
NiuYawei, line 6504: Shouldn't a RAS event being raised? And I think the ras event function will print local error logs accordingly.

liuxuezhao: TBH I am not sure if current handling for pw_rf will be changed in 2.6.2. We will simply the container RF/rd_lvl things later. maybe prompt with critical log is fine for 2.6.1's temporarily solution, thx.

liw (dismissed):

liuxuezhao:
two failed test cases due to https://daosio.atlassian.net/browse/DAOS-16498, and https://daosio.atlassian.net/browse/DAOS-16499

wangshilong (dismissed):
It might make more sense to just change CI python scripts to set env by default to 4? otherwise there might be some other rebuild tests failure(when running full test...)

Nasf-Fan (dismissed):

daltonbohning:
> It might make more sense to just change CI python scripts to set env by default to 4? otherwise there might be some other rebuild tests failure(when running full test...)

Ideally, if DAOS (control plane, server, engine) can figure out a default automatically, that would be preferred.
But if not, and if this should be set for all CI, we could put it here in `common` so every test uses it. Then we don't need to update individual tests, and new tests will also use it.
https://github.com/daos-stack/daos/blob/46e638339b9b1ae9e0c7dc2cc13a15fad6124fbc/src/tests/ftest/util/server_utils_params.py#L435-L437

liuxuezhao:
> > It might make more sense to just change CI python scripts to set env by default to 4? otherwise there might be some other rebuild tests failure(when running full test...)
> 
> Ideally, if DAOS (control plane, server, engine) can figure out a default automatically, that would be preferred. But if not, and if this should be set for all CI, we could put it here in `common` so every test uses it. Then we don't need to update individual tests, and new tests will also use it.
> 
> https://github.com/daos-stack/daos/blob/46e638339b9b1ae9e0c7dc2cc13a15fad6124fbc/src/tests/ftest/util/server_utils_params.py#L435-L437

OK, will update in following PR, thx

docs/admin/env_variables.md
```diff
@@ -53,6 +53,7 @@ Environment variables in this section only apply to the server side.
 |DAOS\_DTX\_RPC\_HELPER\_THD|DTX RPC helper threshold. The valid range is [18, unlimited). The default value is 513.|
 |DAOS\_DTX\_BATCHED\_ULT\_MAX|The max count of DTX batched commit ULTs. The valid range is [0, unlimited). 0 means to commit DTX synchronously. The default value is 32.|
 |DAOS\_FORWARD\_NEIGHBOR|Set to enable I/O forwarding on neighbor xstream in the absence of helper threads.|
+|DAOS\_PW\_RF|Set it as pool-wise redundancy factor. The valid range is [1, 4]. The default value is 2.|
```
mchaarawi, line 56: can we please rename this to something more clear:
DAOS_POOL_RF
update comment to:
Redundancy factor for the pool. The valid range is [1, 4]. The default value is 2.

docs/admin/pool_operations.md
```diff
@@ -916,6 +916,26 @@ and possibly repair a pmemobj file. As discussed in the previous section, the
 rebuild status can be consulted via the pool query and will be expanded
 with more information.
 
+## Pool-Wise Redundancy Factor (PW_RF)
+
+When there are cascading multiple failures, if the number of failed fault
+domains exceeds the PW_RF, there could be unrecoverable errors and applications
+could suffer from data loss. Especially when there is power or network outage,
+it could cause massive node/engine failures. In most cases those failures can
+be recovered and DAOS engines can be restarted and work again.
```
mchaarawi, line 925: If the DAOS system experiences cascading failures, where the number of failed fault domains exceeds a pool's redundancy factor, there could be unrecoverable errors and applications could suffer from data loss. This can happen in cases of power or network outages and would cause node/engine failures. In most cases those failures can be recovered and DAOS engines can be restarted and the system can function again.

docs/admin/pool_operations.md
```diff
@@ -916,6 +916,26 @@ and possibly repair a pmemobj file. As discussed in the previous section, the
...
+
+Administrator can set PW_RF by environment variable "DAOS_PW_RF". When SWIM
```
mchaarawi, line 927: This needs to be imo a server yaml setting and not an env variable. something like:
```
port: 10001
provider: ofi+tcp
socket_dir: /tmp/daos_server
min_pool_rf: 2
```

docs/admin/pool_operations.md
```diff
@@ -916,6 +916,26 @@ and possibly repair a pmemobj file. As discussed in the previous section, the
...
+reports an engine is dead, if the number of failed fault domain exceeds or
+is going to exceed PW_RF, it will not change pool map immediately. Instead,
+it will give critical log message "intolerable unavailability: engine rank x",
+so administrator should check and try to recover those engines, and bring
+back those engines by "dmg system start --ranks=x" one by one, needs not
+reintegrate it back.
```
mchaarawi, line 933: Administrator can set the default pool-wise redundancy factor by environment variable "DAOS_PW_RF" in the server yaml file. If SWIM detects and reports an engine is dead and the number of failed fault domain exceeds or
is going to exceed the pool-wise redundancy factor, it will not change pool map immediately. Instead,
it will give critical log message: 
`intolerable unavailability: engine rank x`
In this case, the system administrator should check and try to recover those failed engines and bring
them back with:
`dmg system start --ranks=x `
one by one. A reintegrate call is not needed.

docs/admin/pool_operations.md
```diff
@@ -916,6 +916,26 @@ and possibly repair a pmemobj file. As discussed in the previous section, the
...
+
+For unrecoverable failures, administrator can still exclude it explicitly
+that is not affected by PW_RF. However, data loss is expected as the number
+of unrecoverable failures exceeds PW_RF.
```
mchaarawi, line 937: For true unrecoverable failures, the administrator can still exclude engines / server nodes. However, data loss is expected as the number of unrecoverable failures exceeds the pool-wise redundancy factor.

mchaarawi (commented):
Made some suggestions on the documentation and naming. please consider to make it more clear.
is there any concern on bwd compatibility? for systems with existing pool and daos upgrades, is this properly handled?

docs/admin/env_variables.md
liuxuezhao: ok

docs/admin/pool_operations.md
liuxuezhao: ok

docs/admin/pool_operations.md
liuxuezhao: in 2.6.2 will add pool property for that, so need not yml file setting later.

docs/admin/pool_operations.md
liuxuezhao: ok

docs/admin/pool_operations.md
liuxuezhao: ok

liuxuezhao:
> Made some suggestions on the documentation and naming. please consider to make it more clear. is there any concern on bwd compatibility? for systems with existing pool and daos upgrades, is this properly handled?

now it is used as ENV setting with default setting, should no bwd compatibility issue. 
later if use pool property will need to handle it.

Nasf-Fan (approved):

gnailzenh (approved):

PR #15064, opened 2024-09-04
DAOS-16486 object: return proper error on stale pool map

NiuYawei:
Client with stale pool map may try to send RPC to a DOWN target, if the target was brought DOWN due to faulty NVMe device, the ds_pool_child could have been stopped on the NVMe faulty reaction, We'd ensure proper error code is returned for such case.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/src/object/srv_obj.c
+++ b/src/object/srv_obj.c
@@ -2161,8 +2161,37 @@ obj_ioc_begin_lite(uint32_t rpc_map_ver, uuid_t pool_uuid,
 	int			rc;
 
 	rc = obj_ioc_init(pool_uuid, coh_uuid, cont_uuid, rpc, ioc);
-	if (rc)
+	if (rc) {
+		DL_ERROR(rc, "Failed to initialize object I/O context.");
+
+		/*
+		 * Client with stale pool map may try to send RPC to a DOWN target, if the
+		 * target was brought DOWN due to faulty NVMe device, the ds_pool_child could
+		 * have been stopped on the NVMe faulty reaction, then above obj_io_init()
+		 * will fail with -DER_NO_HDL.
+		 *
+		 * We'd ensure proper error code is returned for such case.
+		 */
+		poc = ds_pool_child_find(pool_uuid);
+		if (poc == NULL) {
+			D_ERROR("Failed to find pool:"DF_UUID"\n", DP_UUID(pool_uuid));
+			return rc;
+		}
+
+		if (rpc_map_ver < poc->spc_pool->sp_map_version) {
+			D_ERROR("Stale pool map version %u < %u from client.\n",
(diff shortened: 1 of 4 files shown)


src/pool/srv_target.c
```diff
@@ -88,6 +88,21 @@ pool_child_lookup_noref(const uuid_t uuid)
 	return NULL;
 }
 
+struct ds_pool_child *
+ds_pool_child_find(const uuid_t uuid)
```
tanabarr, line 92: given the logic for find and lookup is mostly the same why not just add a bool opt to lookup to return all?

tanabarr (approved):

wangshilong (approved):

src/object/srv_obj.c
```diff
@@ -2161,8 +2161,37 @@ obj_ioc_begin_lite(uint32_t rpc_map_ver, uuid_t pool_uuid,
...
+
+		/*
+		 * Client with stale pool map may try to send RPC to a DOWN target, if the
+		 * target was brought DOWN due to faulty NVMe device, the ds_pool_child could
+		 * have been stopped on the NVMe faulty reaction, then above obj_io_init()
+		 * will fail with -DER_NO_HDL.
+		 *
+		 * We'd ensure proper error code is returned for such case.
+		 */
+		poc = ds_pool_child_find(pool_uuid);
```
liuxuezhao, line 2175: just confirm why not just use ds_pool_child_lookup()?
the difference with ds_pool_child_find() is ds_pool_child_find() will not return NULL for POOL_CHILD_STOPPING, so you want it continue to serve IO request when POOL_CHILD_STOPPING?

liuxuezhao (approved):

src/object/srv_obj.c
NiuYawei: As the comment explained, If the ds_pool_child is already stopped, it's state will be NEW or STOPPING, then ds_pool_child_lookup() will return NULL.

We'll return error instead of continue server I/O request, but we need to ensure -DER_STALE is returned when pool map is stale (instead of -DER_NO_HDL).

src/pool/srv_target.c
NiuYawei: that would cause more code changes. (every caller of ds_pool_child_lookup() needs be changed).

tanabarr (approved):

NiuYawei:
NLT result is unstable since this PR introduced a new error message. @daos-stack/daos-gatekeeper , this PR requires force landing.

src/pool/srv_target.c
gnailzenh, line 92: This is a improper name, I'd rather call it ds_pool_child_lookup_force(), it's confusing to have both find() and lookup()

gnailzenh (approved):

PR #15130, opened 2024-09-14
DAOS-16559 container: fix race of concurrent container destroy

gnailzenh:
If there are more than one caller of destroy against a container, the current container destory handler will hit an assertion.

This patch fixes the race of concurrent destroy against a same container.

### Gatekeeper:

### DIFF
--- a/src/common/lru.c
+++ b/src/common/lru.c
@@ -37,9 +37,12 @@ lru_hop_rec_decref(struct d_hash_table *htable, d_list_t *link)
 	D_ASSERT(llink->ll_ref > 0);
 	llink->ll_ref--;
 
-	/* eviction waiter is the last one holds refcount */
-	if (llink->ll_wait_evict &&
-	    llink->ll_ops->lop_wakeup && daos_lru_is_last_user(llink))
+	/* someone is waiting for eviction of this element, no one else takes refcount
+	 * except the hash table.
+	 */
+	if (llink->ll_wait_evict > 0 &&
+	    llink->ll_wait_evict + 1 == llink->ll_ref &&
+	    llink->ll_ops->lop_wakeup)
 		llink->ll_ops->lop_wakeup(llink);
 
 	/* Delete from hash only if no more references */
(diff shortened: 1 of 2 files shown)


jolivier23 (approved):

src/common/lru.c
```diff
@@ -289,15 +292,17 @@ daos_lru_ref_evict_wait(struct daos_lru_cache *lcache, struct daos_llink *llink)
...
-		 */
-		D_ASSERT(!llink->ll_wait_evict);
-		llink->ll_wait_evict = 1;
+	/* - hash table has +1 ll_ref, myself also has +1 ll_ref (reason of 2).
+	 * - each eviction caller has +1 ll_ref, also has +1 ll_wait_evict.
+	 * - each active user has +1 ll_ref.
+	 *
+	 * if (ll_ref - ll_wait_evict) == 2, it means there is no active user on
+	 * container, I don't need to wait anymore.
+	 */
```
NiuYawei, line 301: That reminds me why we had a "ll_evicting" state before.

Current implementation will make the cont_child as "evicted" on the first daos_lru_ref_evict_wait() caller, so that the second one who tries to destroy container will fail to find the cont_child and continue the destroy. (see cont_child_destroy_one() -> cont_child_lookup()).

I'm wondering why do we want to support concurrent container destroy? Even if two concurrent both hold cont_child and wait in daos_lru_ref_evict_wait(), one of it will eventually fail on later container destroy operations, isn't it?

Shouldn't we make it simpler to fail the second container destroy on daos_lru_ref_hold() (by checking the "evicting" state)? (daos_lru_ref_evict_wait() needs be slightly modified to ignore or fail the second caller).

NiuYawei (changes requested):

src/common/lru.c
gnailzenh: I thought about this, but I didn't  choose this way because the second destroy() will return success even the first one is still in progress, which is confusing because the actual destroy is not done yet. We are not supposed to return success to either of callers.

NiuYawei: The second destroy should fail on cont_child_lookup() & daos_lru_ref_evict_wait(), since it'll find the item is in 'evicting' state.

Your current approach will allow the second destroy moving forward to destroy the container even when the container is still in use. (since the first destroy blindly set the item as 'evicted' despite of the refcount)

gnailzenh:
Supporting concurrent destroy is more complex than I thought, let's just return EBUSY for now, I pushed a new PR: https://github.com/daos-stack/daos/pull/15154

PR #15150, opened 2024-09-18
DAOS-16585 tests: Fix NLT handling of __fxstat detection

techbasset:
Use strace to determine whether calls to __fxstat actually happen when using a utility/command the IL is being tested on, and stop treating it as an error to not see __fxstat when it's not used.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/ci/unit/required_packages.sh
+++ b/ci/unit/required_packages.sh
@@ -30,7 +30,7 @@ pkgs="argobots                         \
       pmix                             \
       protobuf-c                       \
       spdk-devel                       \
-      strace                           \  
+      strace                           \
       valgrind-devel"
 
 # output with trailing newline suppressed
--- a/ci/unit/required_packages.sh
+++ b/ci/unit/required_packages.sh
@@ -30,6 +30,7 @@ pkgs="argobots                         \
       pmix                             \
       protobuf-c                       \
       spdk-devel                       \
+      strace                           \  
       valgrind-devel"
 
 # output with trailing newline suppressed
--- a/utils/node_local_test.py
+++ b/utils/node_local_test.py
@@ -6328,15 +6328,15 @@ def server_fi(args):
 
 
 def look_for_library_call(conf, cmd, library_str):
+    # pylint: disable=wrong-spelling-in-comment
     """Look for library_str in the strace call stack of running cmd."""
-    tmpfile = tempfile.NamedTemporaryFile(mode='r',
-                                          prefix='dnt_assess_',
-                                          suffix='.strace',
-                                          dir=conf.tmp_dir)
-    strace_cmd = ['strace', '--stack-traces', '-v', '-o', tmpfile.name] + cmd
-    subprocess.run(strace_cmd, check=True)
-    lines = tmpfile.readlines()
-    tmpfile.close()
+    with tempfile.NamedTemporaryFile(mode='r',
+                                     prefix='dnt_assess_',
+                                     suffix='.strace',
+                                     dir=conf.tmp_dir) as tmpfile:
+        strace_cmd = ['strace', '--stack-traces', '-v', '-o', tmpfile.name] + cmd
+        subprocess.run(strace_cmd, check=True)
+        lines = tmpfile.readlines()
     for line in lines:
         if re.search(library_str, line):
             return True
--- a/utils/node_local_test.py
+++ b/utils/node_local_test.py
@@ -1521,8 +1521,8 @@ class DFuse():
         print(f'Logged il to {log_name}')
         print(ret)
 
-        if self.caching:
-            check_fstat = False
+        check_fstat = check_fstat and not self.caching and \
+            look_for_library_call(self.conf, cmd, '__fxstat')
 
         try:
             log_test(self.conf, log_name, check_read=check_read, check_write=check_write,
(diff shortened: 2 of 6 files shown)


ci/unit/required_packages.sh
```diff
@@ -30,6 +30,7 @@ pkgs="argobots                         \
       pmix                             \
       protobuf-c                       \
       spdk-devel                       \
+      strace                           \
```
daltonbohning, line 33: There is a space after `\` which messes up the linebreaks
```suggestion
      strace                           \
```

mchaarawi, line 33: @daltonbohning  found an extra space at the end of this line which seems to be causing the CI issues.

techbasset: Ty! Fixed.

techbasset: Ty! Fixed.

utils/node_local_test.py
```diff
@@ -6327,6 +6327,23 @@ def server_fi(args):
         server.set_fi(probability=0)
 
 
+def look_for_library_call(conf, cmd, library_str):
+    """Look for library_str in the strace call stack of running cmd."""
+    tmpfile = tempfile.NamedTemporaryFile(mode='r',
```
jolivier23, line 6332: I don't recall if pylint will still complain on every PR thereafter but it used to be the case.  You could either remove strace from the comment or add a comment such as https://github.com/daos-stack/daos/blob/b5d2047a7283b25d378562fcb8152d3e72b6d6e5/utils/node_local_test.py#L3402

utils/node_local_test.py
```diff
@@ -6327,6 +6327,23 @@ def server_fi(args):
...
+                                          prefix='dnt_assess_',
```
jolivier23, line 6333: this one is simple enough to address, then you don't need explicit close

jolivier23 (approved):

utils/node_local_test.py
```diff
@@ -1521,8 +1521,8 @@ def il_cmd(self, cmd, check_read=True, check_write=True, check_fstat=True):
         print(f'Logged il to {log_name}')
         print(ret)
 
-        if self.caching:
-            check_fstat = False
+        check_fstat = check_fstat and not self.caching and \
+            look_for_library_call(self.conf, cmd, '__fxstat')
```
ashleypittman, line 1525: This would be called for every il_cmd invocation and there are probably dozens so it could/should be saved in `conf`

utils/scripts/install-ubuntu.sh
```diff
@@ -14,6 +14,13 @@ set -e
 
 arch=$(uname -i)
 
+# hack to install 24.04's golang-go on 22.04:
+apt-get install -y software-properties-common
+add-apt-repository "deb http://archive.ubuntu.com/ubuntu noble main"
+apt-get update
+apt-get install -y golang-go
+add-apt-repository -r "deb http://archive.ubuntu.com/ubuntu noble main"
+
```
ashleypittman, line 23: Is this a part of the fix or something else?

ashleypittman (commented):
This would work in as much as if glibc is behaving differently then the specific check that's failing will be disabled.  What we should be testing however is if fstat is being intercepted properly, historically and what the code currently does is check the logs for the wrapper function however dfuse now has per operation statistics available to the client, a more comprehensive solution would be to sample the fstat count before and after the command is invoked as a way of knowing if it had been intercepted or not.

One complexity here is that the first fstat of every file is forwarded so that the st_dev value can be loaded/cached so in order to properly fix this il_stat may need to be passed in the number of files which are accessed.  I'll see if I can get a PR together on this basis.

utils/scripts/install-ubuntu.sh
jolivier23: > Is this a part of the fix or something else?

there was a recent change that made it such that made go 1.22 a requirement and ubuntu 22 has 1.18.   Not sure why the builds didn't fail on that patch.

utils/node_local_test.py
techbasset: I thought about that and wanted to not assume different executables ended up using the same libraries. And, I don't think it's going to affect overall runtime to just do this every time. Do you feel differently?

ashleypittman:
> FWIW I kind of like the strace approach as a general solution as it lets you establish a ground truth about what's actually happening without making assumptions. One can imagine extending this to have the strace dictate all the calls you should be intercepting. @jolivier23 pointed out, for instance, that "newfstat" shows up and is probably not being intercepted right now? _shrug_

I think we're for different points but with the same end.  The current code checks is a particular glibc implementation of fstat is being intercepted, if it's not then is that because the interception is broken or because a different implementation is in use?  Using strace will hide the second failure mode.

Overall this is really just a quick smoke test and tracking for filenames in the log file is a bit of a hack, to do this properly we'd write custom `c` code and call it from ftest with an appropriate harness.  For now, and to make progress here I'd be happy to simply disable the `check_fstat` check, I can re-work NLT to re-enable this smoke test using more modern dfuse/ioil features and having it disabled for a short period of time won't be an issue IMO.

> Meantime we (Google) would like some fix here ASAP to unblock our own client testing. So: request to separate a short term fix from a longer term more complete solution?

utils/scripts/install-ubuntu.sh
ashleypittman: This would probably be better landing as part of https://github.com/daos-stack/daos/pull/15174 where the go developers can review it.

That said, this script normally just installs packages, https://github.com/daos-stack/daos/blob/master/utils/docker/Dockerfile.ubuntu would be a better place for this code, or perhaps a `utils/scripts/helpers/repo-helper-debian.sh` script to match what rocky does.

PR #15174, opened 2024-09-23
DAOS-16621 build: Fix Go versions in rpm/deb packaging

kjacque:
- Set Go minimum version to 1.21 in rpm and debian packaging spec files.
- Update scons Go version check to use version in go.mod.
- Add a reminder in go.mod file so we remember the packaging files when bumping the minimum Go version in the future.
- Update Ubuntu 22.04 Dockerfile to get an appropriate version of Go.

Required-githooks: true

### Gatekeeper:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -25,7 +25,7 @@ Build-Depends: debhelper (>= 10),
                dpdk-dev (>= 21.11.2),
                libisal-crypto-dev,
                libcunit1-dev,
-               golang-go (>= 1.21),
+               golang-go (>= 2:1.21),
                libboost-dev,
                libspdk-dev (>= 22.01.2),
                libipmctl-dev,
@@ -117,7 +117,7 @@ Depends: python (>=3.8), python3, python-yaml, python3-yaml,
          ${shlibs:Depends}, ${misc:Depends},
          daos-client (= ${binary:Version}),
          daos-admin (= ${binary:Version}),
-         golang-go (>=1.21),
+         golang-go (>= 2:1.21),
          libcapstone-dev
 Description: The Distributed Asynchronous Object Storage (DAOS) is an open-source
  software-defined object store designed from the ground up for
--- a/debian/control
+++ b/debian/control
@@ -25,7 +25,7 @@ Build-Depends: debhelper (>= 10),
                dpdk-dev (>= 21.11.2),
                libisal-crypto-dev,
                libcunit1-dev,
-               golang-go (>= 1.18),
+               golang-go (>= 1.21),
                libboost-dev,
                libspdk-dev (>= 22.01.2),
                libipmctl-dev,
@@ -117,7 +117,7 @@ Depends: python (>=3.8), python3, python-yaml, python3-yaml,
          ${shlibs:Depends}, ${misc:Depends},
          daos-client (= ${binary:Version}),
          daos-admin (= ${binary:Version}),
-         golang-go (>=1.18),
+         golang-go (>=1.21),
          libcapstone-dev
 Description: The Distributed Asynchronous Object Storage (DAOS) is an open-source
  software-defined object store designed from the ground up for
--- a/site_scons/site_tools/go_builder.py
+++ b/site_scons/site_tools/go_builder.py
@@ -48,7 +48,7 @@ def _scan_go_file(node, env, _path):
     return includes
 
 
-def get_min_go_version(env):
+def get_min_go_version():
     """Get go minimum version from go.mod"""
     go_mod_path = os.path.join(Dir('#').abspath, "src", "control", "go.mod")
     with open(go_mod_path, 'r') as f:
@@ -92,7 +92,7 @@ def generate(env):
             return 0
 
         context.Display(f'Getting minimum {env.d_go_bin} version... ')
-        min_go_version = get_min_go_version(env)
+        min_go_version = get_min_go_version()
         if min_go_version is None:
             context.Result('failed to extract minimum version from go.mod')
             return 0
--- a/site_scons/site_tools/go_builder.py
+++ b/site_scons/site_tools/go_builder.py
@@ -5,7 +5,7 @@ import os
 import re
 import subprocess  # nosec B404
 
-from SCons.Script import Configure, Exit, File, GetOption, Glob, Scanner
+from SCons.Script import Configure, Dir, Exit, File, GetOption, Glob, Scanner
 
 GO_COMPILER = 'go'
 include_re = re.compile(r'\#include [<"](\S+[>"])', re.M)
@@ -48,6 +48,17 @@ def _scan_go_file(node, env, _path):
     return includes
 
 
+def get_min_go_version(env):
+    """Get go minimum version from go.mod"""
+    go_mod_path = os.path.join(Dir('#').abspath, "src", "control", "go.mod")
+    with open(go_mod_path, 'r') as f:
+        for line in f:
+            if line.startswith('go '):  # e.g. "go 1.21"
+                parts = line.split()
+                return parts[1] + ".0"  # e.g. "1.21.0"
+    return None
+
+
 def get_go_version(output):
     """Capture only the version after 'go'"""
     ver_re = re.compile(r'go([0-9\.]+)')
@@ -80,6 +91,13 @@ def generate(env):
             context.Result(0)
             return 0
 
+        context.Display(f'Getting minimum {env.d_go_bin} version... ')
+        min_go_version = get_min_go_version(env)
+        if min_go_version is None:
+            context.Result('failed to extract minimum version from go.mod')
+            return 0
+        context.Display(min_go_version + '\n')
+
         context.Display(f'Checking {env.d_go_bin} version... ')
         cmd_rc = subprocess.run([env.d_go_bin, 'version'], check=True, stdout=subprocess.PIPE)
         out = cmd_rc.stdout.decode('utf-8').strip()
@@ -92,6 +110,11 @@ def generate(env):
         if go_version is None:
             context.Result(f'failed to get version from "{out}"')
             return 0
+        if len([x for x, y in zip(go_version.split('.'), min_go_version.split('.'))
+                if int(x) < int(y)]) > 0:
+            context.Result(f'{out} is too old (min supported: {min_go_version}) ')
+            return 0
+
         context.Result(go_version)
         return 1
 
--- a/site_scons/site_tools/go_builder.py
+++ b/site_scons/site_tools/go_builder.py
@@ -55,8 +55,8 @@ def get_min_go_version():
         for line in f:
             if line.startswith('go '):  # e.g. "go 1.21"
                 parts = line.split()
-                return parts[1] + ".0"  # e.g. "1.21.0"
-    return None
+                return get_go_version("go" + parts[1])
+    raise Exception("no minimum go version found in go.mod")
 
 
 def get_go_version(output):
@@ -93,9 +93,6 @@ def generate(env):
 
         context.Display(f'Getting minimum {env.d_go_bin} version... ')
         min_go_version = get_min_go_version()
-        if min_go_version is None:
-            context.Result('failed to extract minimum version from go.mod')
-            return 0
         context.Display(min_go_version + '\n')
 
         context.Display(f'Checking {env.d_go_bin} version... ')
--- a/site_scons/site_tools/go_builder.py
+++ b/site_scons/site_tools/go_builder.py
@@ -8,7 +8,6 @@ import subprocess  # nosec B404
 from SCons.Script import Configure, Exit, File, GetOption, Glob, Scanner
 
 GO_COMPILER = 'go'
-MIN_GO_VERSION = '1.18.0'
 include_re = re.compile(r'\#include [<"](\S+[>"])', re.M)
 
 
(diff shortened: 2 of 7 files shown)


kjacque:
No need to run feature tests -- these changes affect the build only.

debian/control
```diff
@@ -25,7 +25,7 @@ Build-Depends: debhelper (>= 10),
                dpdk-dev (>= 21.11.2),
                libisal-crypto-dev,
                libcunit1-dev,
-               golang-go (>= 1.18),
+               golang-go (>= 1.21),
```
ashleypittman, line 28: This has come up before and got dropped but I think the golang-go version on debian uses an epoch so you need to specify the version as I've indicated here.
```suggestion
               golang-go (>= 2:1.21),
```

site_scons/site_tools/go_builder.py
ashleypittman, line 1: There is a lot more code that can be removed from this file if you're removing the go version check.  As this PR stands it'll load the go version but then not do anything with it.

ashleypittman (changes requested):
I don't know if you're aware but docker builds on Ubuntu are failing, as per the build log at https://github.com/daos-stack/daos/actions/runs/10962467337/job/30441777476

What is happening there is the default go version is 1.18, the scons check is detecting this and passing, then there's a non-obvious build failure later on.  In light of that perhaps we need to keep the scons check for now (but update the target version obviously) but also we need to update the packages for that build somehow.

debian/control
ashleypittman: https://packages.debian.org/bookworm/golang-go for the upstream package, and yes it does have/require an epoch number.

site_scons/site_tools/go_builder.py
kjacque: Given the weirdness in the failed build you pointed out, maybe I'll leave the scons check in. The reason I left some of that code in was because it looked like it was printing it to the console, which is useful for debugging build issues.

kjacque:
> In light of that perhaps we need to keep the scons check for now (but update the target version obviously) but also we need to update the packages for that build somehow.

Ubuntu LTS versions tend to stick at a very old version of Go, to the point where I think it's more common to either download/install Go manually or use a third-party PPA. The longsleep/go-backports PPA is the one I've seen used most often, and is updated on a regular cadence. Do we need to get approval for that? @brianjmurrell what do you think?

mjmac:
> Ubuntu LTS versions tend to stick at a very old version of Go, to the point where I think it's more common to either download/install Go manually or use a third-party PPA. The longsleep/go-backports PPA is the one I've seen used most often, and is updated on a regular cadence. Do we need to get approval for that? @brianjmurrell what do you think?

A better option than the PPA is to just install the official golang-go packages from the next LTS release. There is some discussion about the best way to do this in another PR: https://github.com/daos-stack/daos/pull/15150/files#diff-44f4cb066190d53259ef4c6c84dbac0f295f7e229ddb213000deadb5b9f7cb9eR17

utils/rpms/daos.spec
```diff
@@ -592,6 +592,9 @@ getent passwd daos_agent >/dev/null || useradd -s /sbin/nologin -r -g daos_agent
 # No files in a shim package
 
 %changelog
+* Mon Sep 23 2024 Kris Jacque <kris.jacque@intel.com> 2.3.103-6
```
mjmac, line 595: Looks like the version got goofed up here.

site_scons/site_tools/go_builder.py
```diff
@@ -8,7 +8,6 @@
 from SCons.Script import Configure, Exit, File, GetOption, Glob, Scanner
 
 GO_COMPILER = 'go'
-MIN_GO_VERSION = '1.18.0'
```
mjmac, line 11: May be overengineering, but in the interest of minimizing the number of places that need to be updated, maybe this could be based on the version in `go.mod`?

Incidentally, for the record, I do agree that we should keep this scons check for the clear failure when building from source with an outdated Go toolchain.

utils/rpms/daos.spec
kjacque: Good catch. Fixed.

site_scons/site_tools/go_builder.py
kjacque: I did this and it seems to work. Thanks, this should make things a little easier to maintain in the future.

site_scons/site_tools/go_builder.py
kjacque: I ended up taking Mike's suggestion below and automating the minimum version detection based on go.mod.

kjacque:
> Ubuntu LTS versions tend to stick at a very old version of Go, to the point where I think it's more common to either download/install Go manually or use a third-party PPA. The longsleep/go-backports PPA is the one I've seen used most often, and is updated on a regular cadence. Do we need to get approval for that? @brianjmurrell what do you think?

Latest commits use the external PPA I mentioned here.

site_scons/site_tools/go_builder.py
```diff
@@ -48,6 +48,17 @@ def _scan_go_file(node, env, _path):
...
 
 
+def get_min_go_version():
+    """Get go minimum version from go.mod"""
+    go_mod_path = os.path.join(Dir('#').abspath, "src", "control", "go.mod")
+    with open(go_mod_path, 'r') as f:
+        for line in f:
+            if line.startswith('go '):  # e.g. "go 1.21"
+                parts = line.split()
+                return parts[1] + ".0"  # e.g. "1.21.0"
```
mjmac, line 58: Hmm. I don't think that it's safe to unconditionally append the `.0` suffix here. Per [the spec](https://go.dev/ref/mod#go-mod-file-go), the `go` directive may contain any valid [Go version](https://go.dev/doc/toolchain#version).

After looking at little more closely at the version comparison, I remembered that it doesn't rely on the versions being balanced, i.e. it works fine if one is `x.y.z` and one is `x.y`, so you could just return `parts[1]` here and be done with it, except for the fact that the version might contain an `rc` string. Sigh. You've got the `get_go_version()` helper below that almost does what you need, except that it won't quite match on either the line or just the version. Were it me, I'd probably just do something like `return get_go_version("go" + parts[1])` and call it a day.

The other thing is that if this function somehow fails to get a version from that file, returning `None` will make the version comparison fail with a mysterious error like `AttributeError: 'NoneType' object has no attribute 'split'`. Might be best to just raise a specific exception instead.

kjacque: Good idea. Thanks. I don't know how likely it is for us to use more complicated go version strings in the go.mod file, but simple enough to harden it up.

site_scons/site_tools/go_builder.py
```diff
@@ -49,6 +48,17 @@ def _scan_go_file(node, env, _path):
...
 
+def get_min_go_version():
+    """Get go minimum version from go.mod"""
+    go_mod_path = os.path.join(Dir('#').abspath, "src", "control", "go.mod")
+    with open(go_mod_path, 'r') as f:
+        for line in f:
+            if line.startswith('go '):  # e.g. "go 1.21"
+                parts = line.split()
+                return get_go_version("go" + parts[1])
+    raise Exception("no minimum go version found in go.mod")
```
ashleypittman, line 59: @jolivier23 might remember quicker than me but there's a `scons` exception we cal call here I think.  Either way the error needs to be resolved before this can be landed.

ashleypittman (commented):
Without having Ubuntu in the support matrix (which I don't think it is) then Ubuntu dockerfiles are for convenience only and I see no reason why we shouldn't be updating to the latest stable or LTS release.  This dockerfile is only invoked from GHA that I'm aware of.

site_scons/site_tools/go_builder.py
ashleypittman: Reading this through again, there's a scons configure check going on here but the configure check itself is also only called from this file and if the configure check fails then the build stops with `Exit(1)` (a scons function).

What I think you need to do here is return `None` here, and then when calling this function later on check for `None` and call `context.Result` with the error string and return 0.

site_scons/site_tools/go_builder.py
```diff
@@ -81,6 +91,10 @@ def _check_go_version(context):
             context.Result(0)
             return 0
 
+        context.Display(f'Getting minimum {env.d_go_bin} version... ')
+        min_go_version = get_min_go_version()
```
ashleypittman, line 95:
```suggestion
        min_go_version = get_min_go_version()
        if min_go_version is None:
            context.Result("no minimum go version found in go.mod")
            return 1
```
This will need testing and adjusting for whitespace I assume.

site_scons/site_tools/go_builder.py
```diff
@@ -81,6 +91,10 @@ def _check_go_version(context):
             context.Result(0)
             return 0
 
+        context.Display(f'Getting minimum {env.d_go_bin} version... ')
```
ashleypittman, line 94: You're not checking the minimum version of the installed binary, you're checking the binary against a minimum version of the language.
```suggestion
        context.Display('Getting minimum go version... ')
```
As this code stands it's outputting the below which feels incorrect.
```
Checking for Go compiler... /bin/go
Getting minimum /bin/go version... 1.21
Checking /bin/go version... 1.21.13
```

site_scons/site_tools/go_builder.py
kjacque: Yeah, this was similar to what I had before. I'll change it back.

site_scons/site_tools/go_builder.py
kjacque: Fair point, fixing.

ashleypittman:
The other thing on the scons check front is that it might be possible to just get the `go` command to do "something" with the go.mod file and see if fails or not.  I'm not sure if there is a suitable go command for this that exists in 1.21 though.  It would simplify all the code we have if we could just do a test compile.  Out of scope for this PR though.

site_scons/site_tools/go_builder.py
ashleypittman: I missed that, sorry, but I think this is the correct way to fail the build if the check cannot be run.

utils/scripts/install-ubuntu.sh
```diff
@@ -14,6 +14,11 @@ set -e
 
 arch=$(uname -i)
 
+# DAOS requires newer golang version than the one available in core ubuntu repo
+apt-get install gpg-agent software-properties-common
+add-apt-repository ppa:longsleep/golang-backports
+apt update
```
ashleypittman, line 20: I thought I'd made a comment on this change but I can't see it anywhere, apologies if I'm repeating myself here.

One of the intentions of these scripts is that an admin should be able to call them to install required packages and to do so safely, so therefore they install packages without the -y option for example.  The package manager/repo configuration is all done in the dockerfile/helper scripts and this script then just installs the named packages.

This code breaks that model so the script will re-configure the node as well as install code and in addition it's also possibly not idempotent any more.

kjacque: I tested moving the Dockerfile Ubuntu version forward to 24.04 LTS and ran into a warning with the interception library (src/client/dfuse/il), I think due to an upgraded version of gcc.

```
In file included from src/client/dfuse/il/int_posix.c:32:
src/client/dfuse/il/intercept.h:38:21: error: ‘fseek64’ specifies less restrictive attribute than its target ‘fseek’: ‘nonnull’ [-Werror=missing-attributes]
   38 |         ACTION(int, fseek, (FILE *, long, int))                                                    \
      |                     ^~~~~
src/client/dfuse/il/intercept.h:106:27: note: in definition of macro ‘IOIL_DECLARE_ALIAS64’
  106 |         DFUSE_PUBLIC type name##64 params __attribute__((weak, alias(#name)));
      |                           ^~~~
src/client/dfuse/il/int_posix.c:3023:1: note: in expansion of macro ‘FOREACH_ALIASED_INTERCEPT’
 3023 | FOREACH_ALIASED_INTERCEPT(IOIL_DECLARE_ALIAS64)
      | ^~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/client/dfuse/il/int_posix.c:21:
/usr/include/stdio.h:779:12: note: ‘fseek64’ target declared here
  779 | extern int fseek (FILE *__stream, long int __off, int __whence)
      |            ^~~~~
cc1: all warnings being treated as errors
```

I think pushing up the Ubuntu version is out of scope for this task, given that it'll require production code changes.

I could shift the PPA addition to the Dockerfile if that would be an acceptable compromise.

mjmac (approved):

tanabarr (approved):

mjmac (approved):

utils/scripts/install-ubuntu.sh
ashleypittman: If there's build errors then we definitely should update, that's what we're looking for with this test.  Outside the scope of this PR though but I'll file a ticket and take it on.

src/control/go.mod
```diff
@@ -1,9 +1,11 @@
 module github.com/daos-stack/daos/src/control
 
+// NB: When updating minimum Go build version, don't forget to update:
+// - rpm packaging version checks: utils/rpms/daos.spec
+// - debian packaging version checks: debian/control
+// Scons uses this file to extract the minimum version.
 go 1.21
 
-toolchain go1.22.3
-
```
ashleypittman, line 6: I'm not familiar with go, what does this part of the change do?

ashleypittman (approved):

src/control/go.mod
kjacque: Yeah, the go mod tools added that line automatically, but I don't think we need it. It seems that it's only relevant if we want to use a different toolchain version than the minimum go version.

PR #15196, opened 2024-09-25
DAOS-16477 mgmt: return suspect engines for pool healthy query

wangshilong:
After significant failures, the system may leave behind some suspect
engines that were marked as DEAD by the SWIM protocol, but were not
excluded from the system to prevent data loss. An administrator
can bring these ranks back online by restarting them.

This PR aims to provide an administrative interface for querying
suspect engines following a massive failure. These suspect engines
can be retrieved using the daos/dmg --health-only command.

An example of output of dmg pool query --health-only:

Pool 6f450a68-8c7d-4da9-8900-02691650f6a2, ntarget=8, disabled=2, leader=3, version=4, state=Degraded
Pool health info:
- Disabled ranks: 1
- Suspect ranks: 2
- Rebuild busy, 0 objs, 0 recs

Required-githooks: true

### Gatekeeper:


johannlombardi:
@wangshilong reviewing the terminology at the TSC working group and we are suggesting to replace "hanging" in the output of the tool with "suspect".

wangshilong:
> @wangshilong reviewing the terminology at the TSC working group and we are suggesting to replace "hanging" in the output of the tool with "suspect".

@johannlombardi Sure

tanabarr (changes requested):
Changes look okay but we need some updates to documentation to explain what the new parameter means in pool query output and how and why the administrator should use this information.

Also please provide some control-plane unit tests and preferably update ftest for dmg pool query coverage.

src/control/cmd/daos/pool.go
```diff
@@ -329,6 +329,9 @@ func queryPool(poolHdl C.daos_handle_t, queryMask daos.PoolQueryMask) (*daos.Poo
 		if queryMask.HasOption(daos.PoolQueryOptionDisabledEngines) {
 			poolInfo.DisabledRanks = rs
 		}
+		if queryMask.HasOption(daos.PoolQueryOptionSuspectEngines) {
+			poolInfo.SuspectRanks = rs
```
mjmac, line 333: You haven't added the SuspectEngines option to the check up at line 302, so `rlPtr` will always be `nil`.

I know that you're just following the pattern that was here before (which came from a quick refactoring to support pool queries from different parts of the `daos` tool code), but it would be ideal to rework `queryPool()` so that a caller can invoke it once with all of the desired bits set, and then inside of `queryPool()` we can take care of performing multiple pool queries to get the various rank lists. It shouldn't be up to callers of `queryPool()` to do all of this work to build the full response.

Longer-term, I think we need to add a new client-side pool query API that allows all of these rank lists to be fetched at once, similar to how `ds_mgmt_pool_query()` works.

liw (dismissed):
The C changes look good to me.

kjacque (changes requested):
Agreed with @tanabarr that documentation is needed here to tell users what "suspect" ranks are, and understanding what this means for their system. From reading the code, I get the idea that they are ranks we think may be down, but aren't sure yet for some reason (is this correct?).

I noticed that the suspect ranks were added to the default flags. Do we really want to show these "maybe up, maybe down" ranks by default, if this state is probably transient?

Also, at minimum I think all the Go changes should have matching unit test cases demonstrating the added logic.

src/mgmt/tests/srv_drpc_tests.c
```diff
@@ -1408,12 +1408,13 @@ test_drpc_pool_query_success(void **state)
 	Drpc__Response		resp = DRPC__RESPONSE__INIT;
 	uuid_t			exp_uuid;
 	daos_pool_info_t	exp_info = {0};
+	uint64_t flags = DPI_ENGINES_ENABLED | DPI_ENGINES_DISABLED | DPI_ENGINES_SUSPECT;
```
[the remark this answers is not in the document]
wangshilong: I tried, clang formatted it back to this.

src/pool/srv_cli.c
```diff
@@ -367,16 +368,65 @@ pool_query_init(uuid_t pool_uuid, crt_rpc_t *rpc, void *varg)
...
+		rc = crt_rank_state_get(primary_grp, doms[i].do_comp.co_rank, &state);
+		if (rc != 0 && rc != -DER_NONEXIST) {
+			D_ERROR("failed to get status of rank %u: %d\n", doms[i].do_comp.co_rank,
+				rc);
+			break;
+		}
+
+		D_DEBUG(DB_MD, "rank/state %d/%d\n", doms[i].do_comp.co_rank,
+			rc == -DER_NONEXIST ? -1 : state.sms_status);
+		if (rc == -DER_NONEXIST || state.sms_status == SWIM_MEMBER_DEAD) {
```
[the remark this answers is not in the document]
wangshilong: To be honest, i don't quite understand this too, i copied this logic from Xuezhao's codes, @liuxuezhao would you help answer this question please?

wangshilong:
> Agreed with @tanabarr that documentation is needed here to tell users what "suspect" ranks are, and understanding what this means for their system. From reading the code, I get the idea that they are ranks we think may be down, but aren't sure yet for some reason (is this correct?).

I added documents in the following commits.

> 
> I noticed that the suspect ranks were added to the default flags. Do we really want to show these "maybe up, maybe down" ranks by default, if this state is probably transient?

Yes, it is transient. Notice this is only added to health query default flags.

> 
> Also, at minimum I think all the Go changes should have matching unit test cases demonstrating the added logic.

Done

src/control/cmd/daos/pool.go
wangshilong: Thanks @mjmac I tried to refactor queryPool(), I agreed with longer-term, let's address that in the separate PR.

wangshilong:
> Changes look okay but we need some updates to documentation to explain what the new parameter means in pool query output and how and why the administrator should use this information.

I added example and documents in the "pool operation"

> 
> Also please provide some control-plane unit tests and preferably update ftest for dmg pool query coverage.

Added, control-plane unit tests, I am trying to update ftests which might cause more time(I might test commits in the PR or in a separate PR) Thanks.

src/control/cmd/daos/pool.go
```diff
@@ -337,6 +338,55 @@ func queryPool(poolHdl C.daos_handle_t, queryMask daos.PoolQueryMask) (*daos.Poo
 	return poolInfo, nil
 }
 
+func queryPool(poolHdl C.daos_handle_t, queryMask daos.PoolQueryMask) (*daos.PoolInfo, error) {
```
tanabarr, line 341: do these really need to be done sequentially, can't the mask handle multiple options set in the same query?

tanabarr (changes requested):
changes look good just some minor updates required for doc updates

src/control/cmd/daos/pool.go
wangshilong: We could do that in the long-term goal, introduce daos_pool_query_v2() to return three rank lists at the same time.

docs/admin/pool_operations.md
```diff
@@ -286,6 +286,28 @@ The example below shows a rebuild in progress and NVMe space allocated.
     Rebuild busy, 75 objs, 9722 recs
 ```
 
+After experiencing significant failures, the system may retain some suspect
+engines that have been marked as DEAD by the SWIM protocol but were not excluded
+from the system to prevent data loss. An administrator can bring these engines back
```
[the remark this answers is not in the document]
liw: Agreed. Plus, two nit-picking comments below, just for consideration.

"The system may retain" and "from the system" are technically incorrect. Such an engine is not excluded **from the pool**, but may have been excluded from the system. So "the pool" might be more accurate.

"To prevent data loss" could be more accurate as "to prevent potential data inconsistency" or "to preserve data consistency". The foremost purpose of not excluding above RF is to prevent applications from reading potential data that would be impacted if we were to exclude all unavailable engines. The second purpose is to avoid having to reintegrate too many engines to work around certain aspects of the current reintegration protocol.

liw (approved):

src/pool/srv_cli.c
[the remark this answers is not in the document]
liw: For example, the pool comprises rank 0, 1, and 2. Rank 2, however, has been excluded from the system by the MS, and has therefore been deleted from the current engine's local primary group. In this case, we'll get `-DER_NONEXIST` here. In other words, both `-DER_NONEXIST` and "exist but dead" imply that the rank is "unavailable".

liw (approved):

src/tests/ftest/control/dmg_pool_query_ranks.py
```diff
@@ -80,58 +87,77 @@ def test_pool_query_ranks_mgmt(self):
...
+            data['response'].get('disabled_ranks'), disabled_ranks,
+            "Invalid disabled_ranks field: want={}, got={}".format(
+                disabled_ranks, data['response'].get('disabled_ranks')))
+
+        self.log.debug("Waiting for pool to be rebuild")
+        self.pool.wait_for_rebuild_to_start()
+
+        # kill second rank.
+        self.server_managers[0].stop_ranks([suspect_rank], self.d_log)
+        time.sleep(20)
```
tanabarr, line 114: Isn't there a better way than using fixed sleeps?? seems like we might be introducing intermittent failures by using them

src/tests/ftest/control/dmg_pool_query_ranks.py
```diff
@@ -80,58 +87,77 @@ def test_pool_query_ranks_mgmt(self):
...
+            data['response'].get('suspect_ranks'), suspect_ranks,
+            "Invalid enabled_ranks field: want={}, got={}".format(
+                suspect_ranks, data['response'].get('suspect_ranks')))
+        self.assertListEqual(
+            data['response'].get('disabled_ranks'), disabled_ranks,
+            "Invalid disabled_ranks field: want={}, got={}".format(
+                disabled_ranks, data['response'].get('disabled_ranks')))
+
+        self.server_managers[0].start_ranks([suspect_rank], self.d_log)
+        time.sleep(10)
```
tanabarr, line 128: as above

tanabarr (approved):
nice work!

src/tests/ftest/control/dmg_pool_query_ranks.py
```diff
@@ -23,7 +23,7 @@ def setUp(self):
         super().setUp()
 
         # Init the pool
-        self.add_pool(connect=False)
+        self.add_pool(connect=False, properties='rd_fac:1')
```
daltonbohning, line 26: We should instead put the properties in the config `dmg_pool_query_ranks.yaml`. Under the existing `pool` key:
```
pool:
  properties: rd_fac:1
```

src/tests/ftest/control/dmg_pool_query_ranks.py
daltonbohning: I agree with Tom. It would be better if we can deterministically decide when the rank has stopped. It's possible to check the corresponding log file, but I would need to dig up the relevant functions. Or do we think 20 seconds is almost always enough, even for various configs?

src/tests/ftest/control/dmg_pool_query_ranks.py
```diff
@@ -80,58 +87,77 @@ def test_pool_query_ranks_mgmt(self):
...
+            data['response'].get('enabled_ranks'), enabled_ranks,
+            "Invalid enabled_ranks field: want={}, got={}".format(
+                enabled_ranks, data['response'].get('enabled_ranks')))
+        self.assertListEqual(
+            data['response'].get('disabled_ranks'), disabled_ranks,
+            "Invalid disabled_ranks field: want={}, got={}".format(
+                disabled_ranks, data['response'].get('disabled_ranks')))
+
+        self.log.debug("Waiting for pool to be rebuild")
+        self.pool.wait_for_rebuild_to_start()
```
daltonbohning, line 110: Only wait for rebuild to start? And don't wait for it to finish?

src/tests/ftest/control/dmg_pool_query_ranks.py
daltonbohning: I think this is why the previous code used
```
self.pool.wait_for_rebuild_to_start()
self.pool.wait_for_rebuild_to_end()
```

src/tests/ftest/control/dmg_pool_query_ranks.py
```diff
@@ -80,58 +87,77 @@ def test_pool_query_ranks_mgmt(self):
...
+        for _ in range(3):
+            try:
+                result = self.pool.reintegrate(exclude_rank)
+                cmd_succeed = True
+                break
+            except CommandFailure:
+                self.log.debug("dmg command failed retry")
+            time.sleep(3)
+
+        self.assertTrue(cmd_succeed, "pool reintegrate failed: {}".format(result))
```
daltonbohning, line 146: If `self.pool.reintegrate(exclude_rank)` failed above and raised an exception, `result` will not be set. I think in this case we don't need to log it, since the output will be in the test log
```suggestion
        self.assertTrue(cmd_succeed, "pool reintegrate failed")
```

src/tests/ftest/control/dmg_pool_query_ranks.yaml
```diff
@@ -1,20 +1,26 @@
...
+        - DAOS_POOL_RF=1
+        - DD_MASK=io,epc,rebuild
       storage:
         0:
           class: ram
           scm_mount: /mnt/daos
   system_ram_reserved: 1
 pool:
   size: 4GB
+  svcn: 5
```
daltonbohning, line 26: I think better to not set svcn, since the default will be 5 with 5 servers? This has been an issue before, so we've started removing svcn from pool create and letting DAOS handle it

src/tests/ftest/util/dmg_utils_base.py
```diff
@@ -6,8 +6,10 @@
 from socket import gethostname
 
 from ClusterShell.NodeSet import NodeSet
+
 from command_utils import CommandWithSubCommand, YamlCommand
-from command_utils_base import BasicParameter, CommandWithParameters, FormattedParameter
+from command_utils_base import (BasicParameter, CommandWithParameters,
+                                FormattedParameter)
 from environment_utils import TestEnvironment
```
daltonbohning, line 13: > Imports are incorrectly sorted and/or formatted.

I think it wants this
```suggestion
from ClusterShell.NodeSet import NodeSet
from command_utils import CommandWithSubCommand, YamlCommand
from command_utils_base import BasicParameter, CommandWithParameters, FormattedParameter
```

src/tests/ftest/util/server_utils_params.py
```diff
@@ -5,7 +5,8 @@
 """
 import os
 
-from command_utils_base import BasicParameter, LogParameter, TransportCredentials, YamlParameters
+from command_utils_base import (BasicParameter, LogParameter,
+                                TransportCredentials, YamlParameters)
```
daltonbohning, line 9:
```suggestion
from command_utils_base import BasicParameter, LogParameter, TransportCredentials, YamlParameters
```

daltonbohning, line 9: You might have a different version of isort if the githooks are changing this.

You can check with
```
python3 -m isort --version
```

And the current you should have is 5.13.2, from `utils/cq/requirements.txt`

src/tests/ftest/control/dmg_pool_query_ranks.py
wangshilong: Added a new helper for that.

src/tests/ftest/control/dmg_pool_query_ranks.py
wangshilong: ..

src/tests/ftest/control/dmg_pool_query_ranks.py
wangshilong: Fixed.

src/tests/ftest/control/dmg_pool_query_ranks.py
wangshilong: Yes, that is written on purpose to see suspect ranks.

src/tests/ftest/control/dmg_pool_query_ranks.py
wangshilong: Fixed.

src/tests/ftest/control/dmg_pool_query_ranks.yaml
wangshilong: This is written on purpose, i want to simplify codes, otherwise i need take care in the codes don't break 1/2 of PS replica, other pool query won't work.

src/tests/ftest/util/dmg_utils_base.py
wangshilong: Sure.

tanabarr (approved):

daltonbohning (changes requested):
ftest changes LGTM but since this touches more that just ftest, we should use `Features` instead of `Test-tag`
```
Features: DmgPoolQueryRanks
```
`Test-tag` tells CI to run ONLY that one test. Also, this test is automatically included because you modified it so you don't even need `Features`

src/mgmt/tests/srv_drpc_tests.c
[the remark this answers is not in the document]
kjacque: Huh, okay. It's weird how clang-format seems to defy its own rules sometimes.

src/pool/srv_cli.c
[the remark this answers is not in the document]
kjacque: I see. Thanks for the clarification, @liw!

src/tests/ftest/util/server_utils_params.py
```diff
@@ -495,6 +494,7 @@ def __init__(self, base_namespace, index, provider=None, max_storage_tiers=MAX_S
             "ABT_ENV_MAX_NUM_XSTREAMS=100",
             "ABT_MAX_NUM_XSTREAMS=100",
             "DAOS_MD_CAP=1024",
+            "DAOS_POOL_RF=4",
```
wangshilong, line 497: @daltonbohning would you help point me how i could modify this parts? I notice if i moved DAOS_POOL_RF=4 to here, some tests, eg rebuild won't inherit this proper：
https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-15196/16/artifact/Functional%20Hardware%20Medium%20Verbs%20Provider/daos_test/suite.py/daos_configs.wolf-217/daos_server.yml/*view*/

But i want to specify this in dmg pool query test, all other tests should be DAOS_POOL_RF=4, it looks a messy to me.

daltonbohning: I think the problem is this is the default set of `env_vars`, but since https://github.com/daos-stack/daos/blob/master/src/tests/ftest/daos_test/suite.yaml#L51-L57 defines `env_vars` these defaults aren't used.

@phender Is this the intended behavior?

src/control/cmd/daos/pool.go
```diff
@@ -329,6 +330,58 @@ func queryPool(poolHdl C.daos_handle_t, queryMask daos.PoolQueryMask) (*daos.Poo
...
+		queryMask.SetOptions(daos.PoolQueryOptionDisabledEngines)
+		origQueryMask.ClearOptions(daos.PoolQueryOptionDisabledEngines)
+	} else if origQueryMask.HasOption(daos.PoolQueryOptionSuspectEngines) {
+		queryMask.SetOptions(daos.PoolQueryOptionSuspectEngines)
+		origQueryMask.ClearOptions(daos.PoolQueryOptionSuspectEngines)
+	}
+	poolInfo, err := queryPoolHelper(poolHdl, queryMask)
+	if err != nil {
+		return nil, err
+	}
```
kjacque, line 358: I don't really understand why this part is necessary given that we go through the sequence of all three flags below.

src/control/cmd/daos/pool.go
```diff
@@ -329,6 +330,58 @@ func queryPool(poolHdl C.daos_handle_t, queryMask daos.PoolQueryMask) (*daos.Poo
...
+		poolInfo.DisabledRanks = poolInfo1.DisabledRanks
+	}
+	if origQueryMask.HasOption(daos.PoolQueryOptionSuspectEngines) {
+		queryMask.ClearAll()
+		queryMask.SetOptions(daos.PoolQueryOptionSuspectEngines)
+		poolInfo1, err := queryPoolHelper(poolHdl, queryMask)
+		if err != nil {
+			return nil, err
+		}
+		poolInfo.SuspectRanks = poolInfo1.SuspectRanks
```
kjacque, line 384: This seems like it could be a loop or a helper function, as these three blocks are very similar.

src/control/cmd/daos/pool.go
```diff
@@ -295,11 +295,12 @@ func convertPoolInfo(pinfo *C.daos_pool_info_t) (*daos.PoolInfo, error) {
 	return poolInfo, nil
 }
 
-func queryPool(poolHdl C.daos_handle_t, queryMask daos.PoolQueryMask) (*daos.PoolInfo, error) {
+func queryPoolHelper(poolHdl C.daos_handle_t, queryMask daos.PoolQueryMask) (*daos.PoolInfo, error) {
```
kjacque, line 298: naming suggestion - I'd leave `queryPool` as-is, as this is the function that calls `daos_pool_query`. Maybe the other function could be called `queryPoolRankLists`, since its main goal is to collect the different rank lists from multiple queries.

kjacque (commented):
Thanks for the fixes. I have a couple more comments on the repetition in the pool query logic. Would be nice to clean this up a bit before landing.

src/control/cmd/daos/pool.go
wangshilong: It is possible that we don't have all three flags enabled at all.. Notice if we only have one flag enabled, after we run this query, the following query won't be executed since we cleared original flag.

src/control/cmd/daos/pool.go
wangshilong: Sure.

src/control/cmd/daos/pool.go
wangshilong: OK

src/control/cmd/daos/pool.go
kjacque: Ah, I see. Thanks for the explanation. I think refactoring this area to reduce repetition will probably help make things clearer.

wangshilong:
ping reviewers..

src/control/cmd/daos/pool.go
```diff
@@ -337,51 +337,61 @@ func queryPoolHelper(poolHdl C.daos_handle_t, queryMask daos.PoolQueryMask) (*da
...
-	}
-	poolInfo, err := queryPoolHelper(poolHdl, queryMask)
-	if err != nil {
-		return nil, err
-	}
-	if origQueryMask.HasOption(daos.PoolQueryOptionEnabledEngines) {
+	poolInfo := &daos.PoolInfo{}
+	originalMask := queryMask // Save the original queryMask
+
+	// Function to handle the query and return a single RankList
```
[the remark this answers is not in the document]
tanabarr: nit: In order to avoid unintended consequences, in-line functions like this should accept the object they are acting on as a parameter rather than mutating objects in an outer scope.

tanabarr (approved):

liw (approved):
The C changes look good to me.

src/tests/ftest/util/server_utils_params.py
liuxuezhao: Just curious why CRT_EVENT_DELAY and DAOS_POOL_RF set at different places?

liuxuezhao (approved):

kjacque (approved):
OK with me. Thanks.

src/control/cmd/daos/pool.go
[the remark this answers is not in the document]
kjacque: This is a good point. I think in this case it is safe, but it is easier to reason about if it accepts the object to be acted upon.

PR #15229, opened 2024-10-01
DAOS-13938 dfuse: remove drc if statx fails

johannlombardi:
Fix another occurence where statx after readdir can fail.

### Gatekeeper:


src/client/dfuse/ops/readdir.c
```diff
@@ -471,7 +471,21 @@ dfuse_do_readdir(struct dfuse_info *dfuse_info, fuse_req_t req, struct dfuse_obj
 					    O_RDWR | O_NOFOLLOW, &obj, &stbuf.st_mode, &stbuf, 1,
 					    &duns_xattr_name, (void **)&outp, &attr_len);
 
-					if (rc != 0) {
+					if (rc == ENOENT) {
```
ashleypittman, line 474: This all looks reasonable.  This loop is called with either read and write locks held depending on context but if the code is doing a lookup then it should be a read lock so this should be safe.

The things I'd be concerned about is what if this is the last drc on the list, or does the relasedir function check that the offsets in adjacent drcs are contagious, they obviously won't be after this but if it does check that then we could just remove the check.

Do you still need help creating a reliable reproducer for accessing this code?

johannlombardi: Thanks Ashley. I have been struggling with using fault inject in fuse. If you have an example where it is used, i would welcome it.

ashleypittman: I've found it really tricky to use fault injection in fuse because of the kernel involvement, that's the reason I've been focussed on running it with `ioil` to validate the underlying functions.

I've still got an idea for producing a reproducer for this code, I should have an update on that later this week depending on priorities.

johannlombardi: i have added some checks for when the drc is last on the list.

PR #15230, opened 2024-10-02
DAOS-16653 pool: Batch crt events

liw:
When multiple engines become unavailable around the same time, if a pool
cannot tolerate the unavailability of those engines, it is sometimes
desired that the pool would not exclude any of the engines. Hence, this
patch introduces a CaRT event delay, tunable via the server-side
environment variable, CRT_EVENT_DELAY, so that the events signaling the
unavailability of those engines will be handled in hopefully one batch,
giving pool_svc_update_map_internal a chance to reject the pool map
update based on the RF check.

When the RF check rejects a pool map change, we should revisit the
corresponding events later, rather than simply throwing them away. This
patch improves this case by returning the events back to the event
queue, and pause the queue handling until next new event or pool map
update.

  - Introduce event sets: pool_svc_event_set. Now the event queue can be
    simplified to just one event set.

  - Add the ability to pause and resume the event handling: pse_paused.

  - Track the time when the latest event was queued: pse_time.

### Gatekeeper:

### DIFF
--- a/docs/admin/env_variables.md
+++ b/docs/admin/env_variables.md
@@ -44,6 +44,7 @@ Environment variables in this section only apply to the server side.
 |DAOS\_MD\_CAP         |Size of a metadata pmem pool/file in MBs. INTEGER. Default to 128 MB.|
 |DAOS\_START\_POOL\_SVC|Determines whether to start existing pool services when starting a daos\_server. BOOL. Default to true.|
 |CRT\_DISABLE\_MEM\_PIN|Disable memory pinning workaround on a server side. BOOL. Default to 0.|
+|CRT\_EVENT\_DELAY|Delay in seconds before handling each CaRT event. INTEGER. Default to 0 s. A longer delay enables batching of successive CaRT events, leading to fewer pool map changes when multiple engines become unavailable at around the same time.|
 |DAOS\_SCHED\_PRIO\_DISABLED|Disable server ULT prioritizing. BOOL. Default to 0.|
 |DAOS\_SCHED\_RELAX\_MODE|The mode of CPU relaxing on idle. "disabled":disable relaxing; "net":wait on network request for INTVL; "sleep":sleep for INTVL. STRING. Default to "net"|
 |DAOS\_SCHED\_RELAX\_INTVL|CPU relax interval in milliseconds. INTEGER. Default to 1 ms.|
--- a/docs/admin/env_variables.md
+++ b/docs/admin/env_variables.md
@@ -44,7 +44,7 @@ Environment variables in this section only apply to the server side.
 |DAOS\_MD\_CAP         |Size of a metadata pmem pool/file in MBs. INTEGER. Default to 128 MB.|
 |DAOS\_START\_POOL\_SVC|Determines whether to start existing pool services when starting a daos\_server. BOOL. Default to true.|
 |CRT\_DISABLE\_MEM\_PIN|Disable memory pinning workaround on a server side. BOOL. Default to 0.|
-|CRT\_EVENT\_DELAY|Delay in seconds before handling each CaRT event. INTEGER. Default to 0 s. A longer delay enables batching of successive CaRT events, leading to fewer pool map changes when multiple engines become unavailable at around the same time.|
+|CRT\_EVENT\_DELAY|Delay in seconds before handling each CaRT event. INTEGER. Default to 10 s. A longer delay enables batching of successive CaRT events, leading to fewer pool map changes when multiple engines become unavailable at around the same time.|
 |DAOS\_SCHED\_PRIO\_DISABLED|Disable server ULT prioritizing. BOOL. Default to 0.|
 |DAOS\_SCHED\_RELAX\_MODE|The mode of CPU relaxing on idle. "disabled":disable relaxing; "net":wait on network request for INTVL; "sleep":sleep for INTVL. STRING. Default to "net"|
 |DAOS\_SCHED\_RELAX\_INTVL|CPU relax interval in milliseconds. INTEGER. Default to 1 ms.|
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -1189,7 +1189,7 @@ pool_svc_locate_cb(d_iov_t *id, char **path)
 static unsigned int
 get_crt_event_delay(void)
 {
-	unsigned int t = 0 /* s */;
+	unsigned int t = 10 /* s */;
 
 	d_getenv_uint("CRT_EVENT_DELAY", &t);
 	return t;
@@ -6965,7 +6965,7 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc,
 		D_INFO(DF_UUID": SWIM exclude %d ranks, failed NODE %d\n",
 		       DP_UUID(svc->ps_uuid), tgt_addrs->pta_number, failed_cnt);
 		if (failed_cnt > pw_rf) {
-			D_CRIT(DF_UUID": exclude %d ranks will break pw_rf %d, failed_cnt %d\n",
+			D_CRIT(DF_UUID": exclude %d ranks will break pool RF %d, failed_cnt %d\n",
 			       DP_UUID(svc->ps_uuid), tgt_addrs->pta_number, pw_rf, failed_cnt);
 			ABT_rwlock_rdlock(svc->ps_pool->sp_lock);
 			rc = pool_map_crit_prompt(svc, svc->ps_pool->sp_map);
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -62,7 +62,6 @@ ds_pool_get_vos_df_version(uint32_t pool_global_version)
 
 /* Pool service crt event */
 struct pool_svc_event {
-	d_list_t		psv_link;
 	d_rank_t		psv_rank;
 	uint64_t		psv_incarnation;
 	enum crt_event_source	psv_src;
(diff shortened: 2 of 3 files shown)


src/pool/srv_pool.c
```diff
@@ -1300,98 +1333,221 @@ ds_pool_enable_exclude(void)
...
+			D_REALLOC_ARRAY(buf, event_set->pss_buf, event_set->pss_cap, cap);
+			if (buf == NULL)
+				return -DER_NOMEM;
+			event_set->pss_buf = buf;
+			event_set->pss_cap = cap;
+		}
+		event_set->pss_len++;
+	}
+
+	event_set->pss_buf[i].psv_rank        = rank;
```
wangshilong, line 1382: Do we need compare incarnation if in existed case?

liw: We do not, I think. The `crt` module should have ordered the events properly (`swim` compares incarnations before updating the SWIM status of a rank). Also, some `GRPMOD` events have `0` incarnations:

```
09/26-01:31:55.86 [...] queue_event() 7db09892: queuing event: rank=6 inc=1885680475287322624 src=1 type=1
09/26-01:31:56.22 [...] queue_event() 7db09892: queuing event: rank=6 inc=0 src=2 type=1
```

And the `0`-incarnation event really occurred after, and should override, the one with a proper incarnation.

wangshilong (dismissed):

src/pool/srv_pool.c
```diff
@@ -1403,31 +1559,83 @@ events_handler(void *arg)
...
 			}
+
+			/* A simple timed cond_wait without polling. */
+			if (time_left > 0) {
+				rc = start_event_timer(&timer_arg);
+				if (rc != 0) {
+					/* No delay then. */
+					DL_ERROR(rc, DF_UUID ": failed to start event timer",
+						 DP_UUID(svc->ps_uuid));
+					events->pse_time = 0;
```
liuxuezhao, line 1596: minor, "events->pse_timeout = 0" here maybe more readable.

src/pool/srv_pool.c
```diff
@@ -6757,15 +6961,19 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc,
...
 		failed_cnt = pool_map_get_failed_cnt(map, PO_COMP_TP_NODE);
-		D_INFO(DF_UUID": SWIM exclude rank %d, failed NODE %d\n",
-		       DP_UUID(svc->ps_uuid), rank, failed_cnt);
+		D_INFO(DF_UUID": SWIM exclude %d ranks, failed NODE %d\n",
+		       DP_UUID(svc->ps_uuid), tgt_addrs->pta_number, failed_cnt);
 		if (failed_cnt > pw_rf) {
-			D_CRIT(DF_UUID": exclude rank %d will break pw_rf %d, failed_cnt %d\n",
-			       DP_UUID(svc->ps_uuid), rank, pw_rf, failed_cnt);
-			rc = pool_map_crit_prompt(svc, map, rank);
+			D_CRIT(DF_UUID": exclude %d ranks will break pw_rf %d, failed_cnt %d\n",
```
liuxuezhao, line 6968: minor, if you need to refresh the PR, could you please change the "pw_rf" in the output log to "pool RF", thx

liuxuezhao (dismissed):
LGTM, just a few points to confirm, thx

src/pool/srv_pool.c
```diff
@@ -1162,6 +1186,15 @@ pool_svc_locate_cb(d_iov_t *id, char **path)
 	return 0;
 }
 
+static unsigned int
+get_crt_event_delay(void)
+{
+	unsigned int t = 0 /* s */;
+
+	d_getenv_uint("CRT_EVENT_DELAY", &t);
```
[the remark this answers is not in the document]
liw: Yes, the feature is currently disabled by default, mostly because I don't know how to modify the test framework yet. (All previous CI runs of this PR were using 10 s delays.) May be changed in a follow-on PR.

Is there any strong reason for introducing a maximum-allowed value? If we have a particular max value in mind, I'm happy to consider it.

src/pool/srv_pool.c
```diff
@@ -1300,98 +1333,221 @@ ds_pool_enable_exclude(void)
...
+		events->pse_paused = false;
+	}
 
-	ABT_mutex_lock(events->pse_mutex);
-	d_list_add_tail(&event->psv_link, &events->pse_queue);
 	ABT_cond_broadcast(events->pse_cv);
+
+out:
+	if (rc != 0 && allocated)
+		free_event_set(&events->pse_pending);
```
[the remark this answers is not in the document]
liw: Actually it simplifies the do-we-have-pending-event checks. If we don't free `pse_pending` here, it may be an empty event set, and we will need to not only see if `pse_pending` is `NULL`, but also see if it's nonempty. So I think the model is simpler if we free the event set here. (We will free only when there's not enough memory---a pathological case anyway.)

src/pool/srv_pool.c
liw: We can't modify `pse_timeout`, because it is supposed to cache the constant value of the `CRT_EVENT_DELAY` environment variable. Setting `pse_time` to zero means "the event set is very old, so next time it won't be delayed". (This is again a `ENOMEM` scenario, I think.)

src/pool/srv_pool.c
```diff
@@ -1403,31 +1559,83 @@ events_handler(void *arg)
...
+				rc = start_event_timer(&timer_arg);
+				if (rc != 0) {
+					/* No delay then. */
+					DL_ERROR(rc, DF_UUID ": failed to start event timer",
+						 DP_UUID(svc->ps_uuid));
+					events->pse_time = 0;
+					continue;
+				}
+			}
 			sched_cond_wait(events->pse_cv, events->pse_mutex);
```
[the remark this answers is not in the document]
liw: Right, `ABT_cond_signal` would also work, I guess. My conditional variable coding style is: always use `broadcast` unless there is performance concern, because there will be no correctness concern this way. In the past I have fixed correctness bugs caused by `signal` being used incorrectly where `broadcast` was required.

src/pool/srv_pool.c
liw: Will do, either in this PR or in a follow-on PR.

src/pool/srv_pool.c
```diff
@@ -7213,28 +7424,51 @@ ds_pool_update_handler_v5(crt_rpc_t *rpc)
...
+	for (i = 0; i < event_set->pss_len; i++) {
+		struct pool_svc_event *event = &event_set->pss_buf[i];
+
+		if (event->psv_type != CRT_EVT_DEAD)
+			continue;
+		if (event->psv_src == CRT_EVS_GRPMOD && event->psv_rank == self_rank) {
+			D_DEBUG(DB_MD, DF_UUID ": ignore exclusion of self\n",
+				DP_UUID(svc->ps_uuid));
+			continue;
+		}
```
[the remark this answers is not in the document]
liw: Looks like it's safe. This check was moved from old line 1356, which I added in commit 47dd7a828bd241b49f09e7c77e84a4b5e274adf0. Unfortunately, I don't recall why I specially wrote `GRPMOD`. One possible reason is that `swim` never trigger a `DEAD` event for oneself.

liw (commented):
Thanks, @xuezhao. Please let me know if your thoughts on my replies.

src/pool/srv_pool.c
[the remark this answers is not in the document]
liuxuezhao: can add "CRT_EVENT_DELAY=0" to src/tests/ftest/util/server_utils_params.py, similar as "DAOS_POOL_RF=4", just FYI.

"Is there any strong reason for introducing a maximum-allowed value?" just avoid admin set it to a unreasonable large number. looks fine to leave it as now if we don't know which max number is reasonable.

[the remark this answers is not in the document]
liw: Thanks for the helpful pointer. I've changed the default to 10 s, and set this delay to 1 s for CI runs, to get some coverage while adding minimum extra delay. Hopefully that will work.

src/pool/srv_pool.c
liw: Done.

liuxuezhao (approved):

src/tests/ftest/util/server_utils_params.py
```diff
@@ -435,6 +435,7 @@ class EngineYamlParameters(YamlParameters):
         "common": [
             "D_LOG_FILE_APPEND_PID=1",
             "DAOS_POOL_RF=4",
+            "CRT_EVENT_DELAY=1",
```
daltonbohning, line 438: The default is 10 but we are setting 1 for all CI testing. Are there concerns we won't be testing what will happen in production?

daltonbohning: I.e. do we expect all DAOS administrators to also override the default?

liw: @daltonbohning, each exclusion will take `CRT_EVENT_DELAY` seconds longer. So, I think it makes little sense to prolong CI runs by causing more idle waiting, even if this means we are not testing the default.

wangshilong (approved):

src/pool/srv_pool.c
```diff
@@ -1300,98 +1333,221 @@ ds_pool_enable_exclude(void)
...
+
+		if (event->psv_src != CRT_EVS_SWIM || event->psv_type != CRT_EVT_ALIVE)
+			continue;
+		if (ds_pool_map_rank_up(svc->ps_pool->sp_map, event->psv_rank)) {
+			/*
+			 * The rank is up in the pool map. Request a pool map
+			 * distribution just in case the rank has recently
+			 * restarted and does not have a copy of the pool map.
+			 */
+			ds_rsvc_request_map_dist(&svc->ps_rsvc);
```
kccain, line 1498: no change requested, food for thought for optimization: if there are multiple alive rank events, could a single ds_rsvc_request_map_dist() be done? I guess for a "mass start" event?

src/pool/srv_pool.c
```diff
@@ -1300,98 +1333,221 @@ ds_pool_enable_exclude(void)
...
-		D_DEBUG(DB_MD, "ignore exclusion of self\n");
-		goto out;
+	/*
+	 * Check if the alive ranks are up in the pool map. If in the future we
+	 * add automatic reintegration below, for instance, we may need
+	 * to not only take svc->ps_lock, but also employ an RDB TX by
+	 * the book.
+	 */
+	ABT_rwlock_rdlock(svc->ps_pool->sp_lock);
+	for (i = 0; i < event_set->pss_len; i++) {
```
kccain, line 1487: no change requested, food for thought. If the struct pool_svc_event_set also kept track of how many of the events (out of pss_len total) are "alive" events, then maybe some optimizations are possible? For example skipping this lock/loop/unlock logic if number of alive events == 0?

Or in the opposite scenario (all events are "alive" events) then pool_svc_exclude_ranks wouldn't have to loop over pss_len items?

kccain (approved):
LGTM. I added some questions/ideas as comments.

src/pool/srv_pool.c
liw: Both are true, indeed. Because `event_set` is an array, not a linked list, I think (admittedly without benchmarking evidence) the looping over at most a couple of thousands of events is likely not a problem. Probably do the optimization if I turn out to be wrong, or when some new requirement also requires the tracking?

src/pool/srv_pool.c
liw: I _think_ we call `ds_rsvc_request_map_dist` at most once here, via the `break` on new line 1501. Does that do what you have in mind?

docs/admin/env_variables.md
```diff
@@ -44,6 +44,7 @@ Environment variables in this section only apply to the server side.
 |DAOS\_MD\_CAP         |Size of a metadata pmem pool/file in MBs. INTEGER. Default to 128 MB.|
 |DAOS\_START\_POOL\_SVC|Determines whether to start existing pool services when starting a daos\_server. BOOL. Default to true.|
 |CRT\_DISABLE\_MEM\_PIN|Disable memory pinning workaround on a server side. BOOL. Default to 0.|
+|CRT\_EVENT\_DELAY|Delay in seconds before handling each CaRT event. INTEGER. Default to 10 s. A longer delay enables batching of successive CaRT events, leading to fewer pool map changes when multiple engines become unavailable at around the same time.|
```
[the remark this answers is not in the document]
liw: Yes, that is correct. Suggestion accepted: https://github.com/liw/daos/tree/liw/ps-events-m1. I will submit it either into this PR or as a follow-up one.

liw (commented):
Thanks, @kccain. Replies in line.

gnailzenh (approved):

src/pool/srv_pool.c
kccain: Oh yes, you are right. I guess in my mind I was thinking about looping over all events and then coming to the determination that a ds_rsvc_request_map_dist() is needed. But here, the code already does a single invocation, and stops the loop after finding the first "alive" event. My bad.

src/pool/srv_pool.c
kccain: Yes, sounds good to me.

PR #15290, opened 2024-10-10
DAOS-16674 vos: create sysdb lock with recursive attribute

Nasf-Fan:
Then up layer caller can hold sysdb lock recursively and avoid self-deadlock.

Features: PoolCleanupTest PoolListConsolidationTest CatRecovCoreTest
Allow-unstable-test: true

### Gatekeeper:

### DIFF
--- a/src/vos/sys_db.c
+++ b/src/vos/sys_db.c
@@ -349,8 +349,9 @@ vos_db_init(const char *db_path)
 int
 vos_db_init_ex(const char *db_path, const char *db_name, bool force_create, bool destroy_db_on_fini)
 {
-	int	create;
-	int	rc;
+	ABT_mutex_attr	attr = ABT_MUTEX_ATTR_NULL;
+	int		create;
+	int		rc;
 
 	D_ASSERT(db_path != NULL);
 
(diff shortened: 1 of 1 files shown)


src/chk/chk_vos.c
```diff
@@ -78,17 +78,7 @@ chk_db_delete(char *key, int key_size)
 static int
 chk_db_traverse(sys_db_trav_cb_t cb, void *args)
 {
-	int	rc;
-
-	if (chk_db->sd_lock)
-		chk_db->sd_lock(chk_db);
```
NiuYawei, line 84: This locking is necessary, it can't be simply removed here. ABT_mutex does support nested locking, so it's ok even if the traverse callback lock it again.
The deadlock can only happen when traverse callback create a new ULT and take the same lock in the new ULT, we should avoid such use case.

daltonbohning (commented):
Since this is supposed to fix DAOS-16674, we should run the tests reported failing
```
Features: PoolCleanupTest PoolListConsolidationTest CatRecovCoreTest
```

src/chk/chk_vos.c
Nasf-Fan: It is hope that the `sysdb` lock can be held recursively, but the current implementation does not. I have refreshed the patch to change the sysdb lock attribute.

src/vos/sys_db.c
```diff
@@ -416,6 +429,8 @@ vos_db_init_ex(const char *db_path, const char *db_name, bool force_create, bool
 	}
 	return 0;
 failed:
+	if (attr != ABT_MUTEX_ATTR_NULL)
+		ABT_mutex_attr_free(&attr);
```
NiuYawei, line 433: Looks the 'attr' is leaked on success case, the 'attr' should be immediately freed after the mutex is created.

Nasf-Fan: Right, refreshed the patch for fix.

NiuYawei (approved):

NiuYawei (approved):

PR #15296, opened 2024-10-11
DAOS-16687 control: Handle missing PCIe caps in storage query usage

tanabarr:
Missing PCIe capabilities when querying a NVMe SSD's configuration
space is unusual but should be handled gracefully by the control-plane
and shouldn't cause a failure to return usage statistics when calling
dmg storage query usage.

Features: control
Required-githooks: true

### Gatekeeper:


knard-intel (approved):
LGTM for what I understand.

src/control/server/instance_storage_rpc.go
```diff
@@ -326,18 +387,29 @@ func scanEngineBdevsOverDrpc(ctx context.Context, engine Engine, pbReq *ctlpb.Sc
...
+				bhReq.RdbSize = pbReq.RdbSize
+			}
+
+			chReq := ctrlrHealthReq{
+				engine: engine,
+				bhReq:  bhReq,
+				ctrlr:  c,
+			}
+			if !pbReq.Meta {
+				// Add link stats to health only if health requested w/o usage flag.
```
kjacque, line 402: This feels a little strange to me. If the user is requesting health info and usage info together, that implies they don't want the link statistics? Why?

src/control/server/instance_storage_rpc.go
```diff
@@ -326,18 +387,29 @@ func scanEngineBdevsOverDrpc(ctx context.Context, engine Engine, pbReq *ctlpb.Sc
...
-				pciutils.NewPCIeLinkStatsProvider())
+
+			healthUpdated, err = populateCtrlrHealth(ctx, chReq)
 			if err != nil {
 				return nil, err
 			}
-			healthUpdated = upd
 		}
+		// Used to update health with link stats, now redundant.
+		c.PciCfg = ""
```
kjacque, line 412: A bit confused by this. Maybe populateCtrlrHealth should be the one clearing this, if it needs to be cleared.

src/control/server/instance_storage_rpc.go
```diff
@@ -510,12 +582,20 @@ func smdQueryEngine(ctx context.Context, engine Engine, pbReq *ctlpb.SmdQueryReq
...
+				ctrlr:         dev.Ctrlr,
+				linkStatsProv: linkStatsProv,
+			}
+
+			if _, err = populateCtrlrHealth(ctx, chReq); err != nil {
 				return nil, err
 			}
 		}
+		// Used to update health with link stats, now redundant.
+		dev.Ctrlr.PciCfg = ""
```
kjacque, line 597: Same question.

src/control/server/instance_storage_rpc.go
```diff
@@ -28,9 +28,14 @@ import (
 )
 
 var (
-	scanSmd                       = listSmdDevices
-	getCtrlrHealth                = getBioHealth
+	// Function pointers to enable mocking.
+	scanSmd         = listSmdDevices
+	getCtrlrsHealth = getBioHealth
```
kjacque, line 33: This is very confusing now that there is a regular function named `getCtrlrHealth`

tanabarr: I thought it wasn't too bad given one is plural, getCtrlrsHealth and getCtrlrHealth?

src/control/server/instance_storage_rpc.go
tanabarr: User can't explicitly request both usage and health in a single command. The `dmg storage query usage` command does set both flags though as we use the health query as a side channel to gather metadata size specifics to be used when evaluating potential metadata overheads associated with a pool.

The case we are trying to handle here is `dmg storage query usage` because we don't want to call into pciutils library if we are not going to be displaying the link info health statistics (like we do for `dmg storage query list-devices --health` and `dmg storage scan --nvme-health`.

src/control/server/instance_storage_rpc.go
tanabarr: I moved this out of the populate function because we want it to be cleared from here on whether or not it gets used and it didn't seem sensible doing it in 2 places.

src/control/server/instance_storage_rpc.go
tanabarr: as above

knard-intel (approved):

src/control/server/instance_storage_rpc.go
kjacque: This makes it easier to tell the difference at a glance, thanks.

src/control/server/instance_storage_rpc.go
kjacque: I wonder if it would be better to add a different flag for the link stats, rather than relying on a non-obvious assumption about implementation details.

src/control/server/instance_storage_rpc.go
kjacque: Understood, although we are still doing it in two places here. I'm OK with it as-is though.

kjacque (commented):
Good fixes, thanks. The one issue remaining open for me is the assumption about the interaction between the incoming flags (meta vs. health). IMO needing to make an assumption like that is a sign that we need more granular flags coming in.

tanabarr:
> Good fixes, thanks. The one issue remaining open for me is the assumption about the interaction between the incoming flags (meta vs. health). IMO needing to make an assumption like that is a sign that we need more granular flags coming in.

To me this Assumption is based on an internal coupling that is acceptable and understood given how the backend is used in multiple front-end calls. If this is a blocking request please explain exactly what the problem is with the assumption and how you want it addressed. The more detail the better please.

src/control/server/instance_storage_rpc.go
tanabarr: done

kjacque (approved):

knard-intel (approved):

PR #15298, opened 2024-10-11
DAOS-16686 dfuse: Fix overlapping chunk reads.

ashleypittman:
Handle concurrent read in the chunk_read code.  Rather than assuming
each slot only gets requested once save the slot number as part of the
request and handle multiple requests.

This corrects the behaviour and avoids a crash when multiple readers read
the same file concurrently and improves the performance in this case.


src/client/dfuse/dfuse.h
```diff
@@ -474,6 +481,10 @@ struct dfuse_pool {
 	ACTION(OPEN)                                                                               \
 	ACTION(PRE_READ)                                                                           \
 	ACTION(READ)                                                                               \
+	ACTION(READ_EOF_M)                                                                         \
+	ACTION(READ_CON)                                                                           \
+	ACTION(READ_BUCKET)                                                                        \
+	ACTION(READ_BUCKET_M)                                                                      \
```
ashleypittman, line 487: Extra metrics in dfuse is a good thing but I'm thinking of removing these from this PR.  They were certainly useful to detect/different code-paths but I think I'd rather extend these outside of a bug-fix PR.

src/tests/ftest/dfuse/read.py
```diff
@@ -23,7 +23,9 @@ def test_dfuse_pre_read(self):
         Read one large file entirely using pre-read.  Read a second smaller file to ensure that
         the first file leaves the flag enabled.
 
-        :avocado: tags=all,full_regression
+        TODO: Check this with 128k I/O size and non-128K I/O size.
```
phender, line 26: I'm surprised this isn't be called out by the linting check.  We should either do this or remove the comment before landing and add a ticket as a reminder.

src/tests/ftest/dfuse/read.py
```diff
@@ -23,7 +23,9 @@ def test_dfuse_pre_read(self):
...
+
+        :avocado: tags=all,daily_regression
```
phender, line 28: Just confirming that moving this from weekly to daily is intentional (I thought I had asked this before, but couldn't find the question in the ticket).  According to the [last weekly](https://build.hpdd.intel.com/job/daos-stack/job/daos/job/weekly-testing/411/testReport/FTEST_dfuse/DFusePreReadTest/) it only takes a minute to run, so the move should be fine.

phender (changes requested):

ashleypittman:
This is consistently failing one of the tests, taking back to draft whilst I investigate.

src/client/dfuse/ops/read.c
```diff
@@ -173,186 +173,112 @@ pick_eqt(struct dfuse_info *dfuse_info)
...
-			if (cd->reqs[i]) {
-				req         = cd->reqs[i];
-				cd->reqs[i] = 0;
-				break;
-			}
-		}
+	D_SPIN_UNLOCK(&ia->lock);
 
-		D_SPIN_UNLOCK(&ia->lock);
+	d_list_for_each_entry_safe(cr, crn, &cd->req_list, req_list) {
```
ashleypittman, line 257: Note to self: There's no reference on `cd` held here after the last call to `DFUSE_REPLY../.` so the list needs to be spliced onto the stack before the list is iterated.

wangdi1 (dismissed):

jolivier23 (dismissed):

phender (dismissed):
Removing -1.

mchaarawi (changes requested):
change looks ok to me. but since the description mentions an issue with a crash and not just a perf improvement, a test needs to be added.

Also please push with Features: dfuse

ashleypittman:
> change looks ok to me. but since the description mentions an issue with a crash and not just a perf improvement, a test needs to be added.

I actually wrote a PR to test this issue, it's easy enough to trigger locally but the timing in CI meant that I wasn't able to reproduce it via ftest, possibly because reads are just faster than running over loopback.

https://github.com/daos-stack/daos/pull/15312

ashleypittman:
> Yea i see how this can be a race that is not always reproducible. but at least having that test in would still give someone a way to run that.

> curious though if putting that case in NLT would trigger that issue since NLT runs on 1 host?

Yes, it's reproducible in NLT every time.  In that case dfuse is run under valgrind so will be much slower and hence it's reproducible every time.  I didn't want to do that though as NLT runs with full debug on so I've typically only used that for very small datasets/iteration counts.  I don't see another way to get this into CI though.

I'll paste the script I've been using tomorrow when I'm at my work laptop but it's the same script I copied to make that test, essentially create a 1M file in dfuse, evict it and then read it in parallel in at least 128k chunks.  The bug/race of course is when the same 128k bunk of a file is requested multiple times concurrently before the first request has been replied to.

ashleypittman:
I can't seem to push to this PR today for some reason but it want's [c7a870377d632454fd99c97dc7cdb4042b01d444](https://github.com/daos-stack/daos/pull/15680/commits/c7a870377d632454fd99c97dc7cdb4042b01d444) added to it.

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15298/7/testReport/

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15298/8/testReport/

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15298/9/testReport/

PR #15306, opened 2024-10-12
DAOS-16693 telemetry: Avoid race between init/read

mjmac:
In rare cases, a reader may attempt to access a telemetry
node after it has been added to the tree, but before it
has been fully initialized. Use an atomic to prevent
reads before the initialization has completed. Unlucky
readers will get a -DER_AGAIN instead of crashing.

Features: telemetry
Required-githooks: true
Change-Id: I12281a7db7ef70d5f827547af1233b7a6619ae6d


tanabarr (approved):

kjacque (approved):

kjacque:
IMO this is a good candidate for backport to 2.6.x.

PR #15321, opened 2024-10-15
DAOS-15971 dfs: add support for mounting a snapshot

johannlombardi:
Add new dfs interface to mount a specific snapshot.
    Provide dfuse command line options to use this feature.

    Signed-off-by: Johann Lombardi <johann.lombardi@gmail.com>


ashleypittman (commented):
Are snapshots still immutable?  If so then as well as setting read-only mode then dfuse should not check the container attributes for caching data but just enable caching with some reasonably long timeout.

johannlombardi:
Yes, snapshots are immutable. Your proposal makes sense. I will implement it in another PR.

johannlombardi:
adding @mjmac for the daos cmd output changes.

src/control/cmd/daos/container.go
```diff
@@ -962,8 +962,8 @@ func printContainerInfo(out io.Writer, ci *daos.ContainerInfo, verbose bool) err
 			{"Pool UUID": ci.PoolUUID.String()},
 			{"Container redundancy factor": fmt.Sprintf("%d", ci.RedundancyFactor)},
 			{"Number of open handles": fmt.Sprintf("%d", ci.NumHandles)},
-			{"Latest open time": fmt.Sprintf("%#x (%s)", ci.OpenTime, daos.HLC(ci.OpenTime))},
-			{"Latest close/modify time": fmt.Sprintf("%#x (%s)", ci.CloseModifyTime, daos.HLC(ci.CloseModifyTime))},
+			{"Latest open time": fmt.Sprintf("%s (%#x) (%s)", daos.HLC(ci.OpenTime), ci.OpenTime)},
```
mjmac, line 965: There are three formatting directives, but only two values, unless I'm missing something...

src/control/cmd/daos/snapshot.go
```diff
@@ -233,13 +233,14 @@ func (cmd *containerSnapListCmd) Execute(args []string) error {
 		return cmd.OutputJSON(snaps, nil)
 	}
 
-	cmd.Info("Container's snapshots :")
+	cmd.Infof("%s\t\t\t%s\t\t\t%s", "Timestamp" , "Epoch", "Name")
+	cmd.Infof("---------\t\t\t-----\t\t\t----")
```
mjmac, line 237: For consistency with other commands, it would be better to use the `txtfmt` package to format these. See [printAttributes()](https://github.com/daos-stack/daos/blob/master/src/control/cmd/daos/attribute.go#L49) for an example.

src/control/cmd/daos/snapshot.go
```diff
@@ -233,13 +233,14 @@ func (cmd *containerSnapListCmd) Execute(args []string) error {
...
 	if len(snaps) == 0 {
 		cmd.Info("no snapshots")
 		return nil
 	}
 	for _, snap := range snaps {
-		cmd.Infof("0x%x %s", snap.Epoch, snap.Name)
+		cmd.Infof("%s\t0x%x\t%s", snap.Timestamp, snap.Epoch, snap.Name)
```
mjmac, line 243: What does the JSON output look like? Is there any processing that needs to happen to make that output useful?

daltonbohning:
> looks good to me. but what is the plan for testing this feature?

Quick suggestion if we want to use ftest.
1. IOR, e.g. to write some data
2. mount dfuse on snapshot
3. IOR read-verify data in dfuse
4. Write more data to container
5. IOR read-verify data from dfuse is still the same


Related, if someone uses the datamover tools with a snapshot-dfuse mount as the source, I think the datamover utils will open the pool/cont NOT at the snapshot. Would need to verify

src/include/daos_fs.h
```diff
@@ -162,6 +162,35 @@ int
...
+ * time). In this case, an epoch of 0 should be passed and \a name should be set to the snapshot
+ * name. Please note that \a name is valid only w
+ *
+ * \param[in]	pool	Pool label.
+ * \param[in]	sys	DAOS system name to use for the pool connect.
+ *			Pass NULL to use the default system.
+ * \param[in]	cont	Container label.
+ * \param[in]	flags	Mount flags for future use, will be O_RDONLY by definition.
+ * \param[in]	epoch	Epoch associated with the snapshot to mount.
+ *			If a null epoch is passed, then the snapshot is looked up by \a name.
```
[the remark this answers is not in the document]
johannlombardi: I am using DAOS_EPOCH_MAX internally already to mean the tip of the container.

johannlombardi:
the plan for the testing is to extend the dfs test to have some coverage for the new API and then add a new ftest to run tests like suggested by dalton. I just would like to do that in a different PR because i know it is going to take me forever to complete due to the lack of jenkins access.

src/control/cmd/daos/snapshot.go
johannlombardi: The json output looks good and already has the timestamp, epoch and name.

src/control/cmd/daos/snapshot.go
johannlombardi: fixed.

ashleypittman:
> Related, if someone uses the datamover tools with a snapshot-dfuse mount as the source, I think the datamover utils will open the pool/cont NOT at the snapshot. Would need to verify

As this is still work-in-progress it would be an idea to just disable the ioctl in this PR, the ioctl will need updating to support this anyway and disabling it would at least prevent any incorrect behaviour by tools built on top of it.
https://github.com/daos-stack/daos/blob/master/src/client/dfuse/ops/ioctl.c#L28

mjmac (commented):
Go changes look good to me, thanks.

src/include/daos_fs.h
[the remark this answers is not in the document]
mchaarawi: no i mean the comment here "If a null epoch is passed". did you mean if 0 epoch is passed?

src/client/dfs/mnt.c
```diff
@@ -717,6 +745,74 @@ dfs_mount(daos_handle_t poh, daos_handle_t coh, int flags, dfs_t **_dfs)
...
+					break;
+				}
+			}
+
+			if (ep != 0)
+				break;
+		}
+
+		if (ep == 0) {
+			D_DEBUG(DB_ALL, "No matching snapshot name found.\n");
```
kjacque, line 803: Could be D_ERROR or D_INFO

kjacque (dismissed):

src/control/cmd/daos/snapshot.go
```diff
@@ -211,6 +215,32 @@ func listContainerSnapshots(ap *C.struct_cmd_args_s, containerID string) ([]*sna
 	return snapshots, nil
 }
 
+func printSnaps(out io.Writer, snaps []*snapshot) {
```
kjacque, line 218: This function looks good, but would be nice to have a unit test for this output. I would suggest shifting this function into the cmd/daos/pretty subpackage and reference the existing unit tests for verifying output.

kjacque (approved):
My other comment is nonblocking. Thanks!

PR #15322, opened 2024-10-15
DAOS-16693 telemetry: Avoid race between init/read (#15306)

mjmac:
In rare cases, a reader may attempt to access a telemetry
node after it has been added to the tree, but before it
has been fully initialized. Use an atomic to prevent
reads before the initialization has completed. Unlucky
readers will get a -DER_AGAIN instead of crashing.


kjacque (approved):
LGTM. Is this a clean cherry-pick?

mjmac:
> LGTM. Is this a clean cherry-pick?

Yep, just added the label.

jolivier23 (approved):

PR #15345, opened 2024-10-18
DAOS-16729 dfuse: Remove deprecated single-threaded option.

ashleypittman:
This only serves to add confusion at this point.


src/tests/ftest/util/dfuse_utils.py
```diff
@@ -30,7 +30,6 @@ def __init__(self, namespace, command, path=""):
         self.sys_name = FormattedParameter("--sys-name {}")
         self.thread_count = FormattedParameter("--thread-count {}")
         self.eq_count = FormattedParameter("--eq-count {}")
-        self.singlethreaded = FormattedParameter("--singlethread", False)
```
daltonbohning, line 33: Luckily no tests are using this

daltonbohning (approved):

src/tests/ftest/util/dfuse_utils.py
ashleypittman: I did check, although only by grepping.

daltonbohning: Searching for `singlethreaded` is 99% sufficient. We don't know if someone somewhere has a different config that uses this, but it's pretty unlikely

knard-intel (dismissed):
LGTM

daltonbohning (approved):

jolivier23 (approved):

utils/node_local_test.py
```diff
@@ -1969,9 +1969,16 @@ class needs_dfuse_with_opt():
...
 
-    def __init__(self, caching=None, wbcache=True, dfuse_inval=True,
-                 ro=False):
-        self.caching = caching
+    # dict of tests that have been excluded at runtime (so we know what to skip)
+    # values are the versions of the test to skip
+    excluded_name_dict = {}
+
+    # ensure thread safety; not sure this is actually necessary, but meh...
+    wrapping_lock = threading.Lock()
```
daltonbohning, line 1977: I think all dict operations are atomic. So if this code were changed to use a single `get`, it would be thread-safe
https://github.com/daos-stack/daos/blob/875b99cfebf6a56ab230c286932c7291ad91dbf0/utils/node_local_test.py#L2030-L2034

I.e.
```
     return list(needs_dfuse_with_opt.excluded_name_dict.get(name, [])
```

daltonbohning (approved):

PR #15427, opened 2024-10-29
Restore patches after upstream fixes

jolivier23:
- **Restore mercury to v2.4.0rc5**
- **Reapply "DAOS-15914 cart: add D_MRECV_BUF env var to control number of multi-r… (#14662) (#14683)"**
- **DAOS-16716 ci: Set reference build for PRs (#15337)**
- **DAOS-16716 ci: Set reference build for PRs (#15379)**
- **DAOS-14408 common: ensure NDCTL not used for storage class ram (#15203)**


cdavis28 (commented):
I see new libraries. Is this going to be a repeat of pci drivers?

cdavis28 (approved):
I suppose we need to worry about the drivers when it's merged into GOB.

jolivier23:
> I see new libraries. Is this going to be a repeat of pci drivers?

Unfortunately, probably.  I don't see a good way around this.  We can avoid taking the pmdk update (which is really removing all stuff we don't need from pmdk) but we will want it eventually, I think

PR #14389, opened 2024-05-16
DAOS-15698 test: Fix nvme/fault.py KeyError

phender:
<Solution pending>

Skip-func-hw-test-large-md-on-ssd: false
Test-tag: NvmeFault

Required-githooks: true

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate owners.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
  * [ ] Quick-build or Quick-functional is not used.
* [ ] Fix the commit message upon landing. Check the standard [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments). Edit it to create a single commit. If necessary, ask submitter for a new summary.

PR #15430, opened 2024-10-30
DAOS-15698 test: nvme_utils.py - Fix code to collect ranks from a host

shimizukko:
test.server_managers[-1].ranks returns a dictionary of rank to host, but the original code was trying to use hostname as key. Fix it so that we create a list of ranks from the dictionary.

Also update the reset faulty device logic.
MD-on-SSD cluster: If we set storage faulty, the corresponding engine dies (becomes excluded). Tell the server manager the rank is excluded and don't reset faulty device.
Non-MD-on-SSD cluster: Reset by calling `dmg storage led identify --reset`. We expect the command to work and all engines to be joined.

Skip-unit-tests: true
Skip-func-hw-test-large-md-on-ssd: false
Test-tag: test_nvme_fault

### Gatekeeper:


dinghwah (dismissed):

src/tests/ftest/util/nvme_utils.py
```diff
@@ -44,19 +44,19 @@ def set_device_faulty(test, dmg, server, uuid, pool=None, has_sys_xs=False, **kw
...
         uuid (str): the device UUID
         pool (TestPool, optional): pool used to wait for rebuild to start/complete if specified.
             Defaults to None.
         has_sys_xs (bool, optional): the device's has_sys_xs property value. Defaults to False.
         kwargs (dict, optional): named arguments to pass to the DmgCommand.storage_set_faulty.
 
     Returns:
-        dict: the json response from the dmg storage set-faulty command.
-
+        dict: the json response from the dmg storage set-faulty command. None if has_sys_xs is True.
```
daltonbohning, line 55: It's always None if has_sys_xs is True? In that case, what's the point in calling this function if it's just expected to fail every time?

phender: We want to call the function to set the device faulty.  The behavior afterward is what differs with a sys_xs device.  In that case we need to tell the test that we expect specific ranks to be stopped/excluded.  Otherwise we need to register a cleanup step to reset the faulty device - which we can do because the ranks are still up with a non sys_xs device.

Technically this method doesn't need to return the response as its stored in the `dmg.result`.  I had made that change in https://github.com/daos-stack/daos/pull/14389/files.

phender (changes requested):
This change needs to be run with any test calling `set_device_faulty`.  Which would include `test_vmd_led_faulty`, `test_disk_failure_recover`, `test_nvme_fault_reintegration` (all of which require un-skipping the Functional Hardware Medium VMD stage to run), and the already run `test_nvme_fault` test.

src/tests/ftest/util/nvme_utils.py
daltonbohning: I think it's better practice to use the returned response and not the `dmg.result`

daltonbohning: So if `has_sys_xs is True` this will call `dmg.storage_set_faulty`, which may or may not work. But we ignore the failure.

So
`if has_sys_xs is False` -> call `dmg.storage_set_faulty` and fail the test if it fails. But if it works return the response
`if has_sys_xs is True` -> call `dmg.storage_set_faulty` and don't care if it works or not. But if it did work return the response, even though the docstring says it will return `None`

phender: > I think it's better practice to use the returned response and not the `dmg.result`

I agree, there are problems with `dmg.result` in terms of knowing when it was last updated.  I just wasn't sure if your concern was with the None return value.  Ideally ExecutableCommand.run would use run_local/run_remote and we'd always have an object to return, but that's way out of the scope of this PR.

shimizukko:
> This change needs to be run with any test calling `set_device_faulty`. Which would include `test_vmd_led_faulty`, `test_disk_failure_recover`, `test_nvme_fault_reintegration` (all of which require un-skipping the Functional Hardware Medium VMD stage to run), and the already run `test_nvme_fault` test.

How can I run HW Medium VMD stage? I tried `Skip-func-hw-test-medium-vmd: false`, but didn't work.

phender:
> > This change needs to be run with any test calling `set_device_faulty`. Which would include `test_vmd_led_faulty`, `test_disk_failure_recover`, `test_nvme_fault_reintegration` (all of which require un-skipping the Functional Hardware Medium VMD stage to run), and the already run `test_nvme_fault` test.
> 
> How can I run HW Medium VMD stage? I tried `Skip-func-hw-test-medium-vmd: false`, but didn't work.

In build 7 the Functional Hardware Medium VMD stage did not run because only the `test_nvme_fault` test tag was used:
```
[2024-11-07T05:45:55.049Z] Detecting tests matching tags: avocado list --filter-by-tags=test_nvme_fault,hw_vmd,medium,-faults ./
[2024-11-07T05:46:16.911Z] None of the following tests matched the tags:
```
and this is a `hw,large` tagged test.

shimizukko:
> > > This change needs to be run with any test calling `set_device_faulty`. Which would include `test_vmd_led_faulty`, `test_disk_failure_recover`, `test_nvme_fault_reintegration` (all of which require un-skipping the Functional Hardware Medium VMD stage to run), and the already run `test_nvme_fault` test.
> > 
> > 
> > How can I run HW Medium VMD stage? I tried `Skip-func-hw-test-medium-vmd: false`, but didn't work.
> 
> In [build 7](https://%5Bbuild.hpdd.intel.com/job/daos-stack/job/daos/job/PR-15430/7/pipeline-console/log?nodeId=805%5D(https://build.hpdd.intel.com/job/daos-stack/job/daos/job/PR-15430/7/pipeline-console/log?nodeId=805)) the Functional Hardware Medium VMD stage did not run because only the `test_nvme_fault` test tag was used:
> 
> ```
> [2024-11-07T05:45:55.049Z] Detecting tests matching tags: avocado list --filter-by-tags=test_nvme_fault,hw_vmd,medium,-faults ./
> [2024-11-07T05:46:16.911Z] None of the following tests matched the tags:
> ```
> 
> and this is a `hw,large` tagged test.
> 
> The reason only the `test_nvme_fault` test tag was used is because of this typo in the commit message:
> 
> ```
> Test-tag: test_nvme_fault, test_nvme_fault_reintegration, test_vmd_led_faulty, test_disk_failure_recover
> ```
> 
> it should be:
> 
> ```
> Test-tag: test_nvme_fault test_nvme_fault_reintegration test_vmd_led_faulty test_disk_failure_recover
> ```

Thanks for pointing out. I fixed it and all tests passed.

phender (approved):

src/tests/ftest/util/nvme_utils.py
daltonbohning: > dict: the json response from the dmg storage set-faulty command. None if has_sys_xs is True.

If this is correct, then the code should verify that `storage_set_faulty` always fails when `has_sys_xs is True`

daltonbohning (changes requested):

src/tests/ftest/util/nvme_utils.py
```diff
@@ -44,19 +44,19 @@ def set_device_faulty(test, dmg, server, uuid, pool=None, has_sys_xs=False, **kw
...
     """
     kwargs['host'] = server
     kwargs['uuid'] = uuid
+    response = None
     try:
         response = get_dmg_response(dmg.storage_set_faulty, **kwargs)
```
phender, line 61: This should cover having to verify the None return value:
```suggestion
    response = None
    try:
        response = get_dmg_response(dmg.storage_set_faulty, **kwargs)
        if has_sys_xs:
            test.fail("Setting a sys_xs device faulty should fail.")
```

phender (changes requested):

phender (approved):

daltonbohning (approved):

dinghwah (approved):

PR #15437, opened 2024-11-03
DAOS-623 client: Reflow comments in src/client.

ashleypittman:
Signed-off-by: Ashley Pittman <ashley@pittman.co.uk>


ashleypittman:
This was done against and old commit but is only intended as a demonstration for @jxiong

PR #15470, opened 2024-11-07
DAOS-16791 control: Add include_fabric_ifaces to agent config

mjmac:
Provide an inverse to the existing exclude_fabric_ifaces
directive. In some cases, a given environment will only have
a small number of valid interfaces, so it is simpler to
specify that rather than having to exclude all of the
invalid interfaces.

Features: control
Required-githooks: true
Change-Id: I0d8445fcc2da267df76856d7f9f9b43eed290599

### DIFF
--- a/src/control/cmd/daos_agent/config_test.go
+++ b/src/control/cmd/daos_agent/config_test.go
@@ -142,7 +142,7 @@ exclude_fabric_ifaces: ["ib3"]
 		},
 		"bad filter config": {
 			path:   badFilterCfg,
-			expErr: errors.New("agent config validation failed"),
+			expErr: errors.New("cannot specify both exclude_fabric_ifaces and include_fabric_ifaces"),
 		},
 		"all options": {
 			path: optCfg,
--- a/src/control/cmd/daos_agent/config_test.go
+++ b/src/control/cmd/daos_agent/config_test.go
@@ -88,6 +88,18 @@ transport_config:
   allow_insecure: true
 `)
 
+	badFilterCfg := test.CreateTestFile(t, dir, `
+name: shire
+access_points: ["one:10001", "two:10001"]
+port: 4242
+runtime_dir: /tmp/runtime
+log_file: /home/frodo/logfile
+transport_config:
+  allow_insecure: true
+include_fabric_ifaces: ["ib0"]
+exclude_fabric_ifaces: ["ib3"]
+`)
+
 	for name, tc := range map[string]struct {
 		path      string
 		expResult *Config
@@ -127,6 +139,10 @@ transport_config:
 		"bad log mask": {
 			path:   badLogMaskCfg,
 			expErr: errors.New("not a valid log level"),
+		},
+		"bad filter config": {
+			path:   badFilterCfg,
+			expErr: errors.New("agent config validation failed"),
 		},
 		"all options": {
 			path: optCfg,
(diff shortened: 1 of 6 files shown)


src/control/cmd/daos_agent/config_test.go
```diff
@@ -128,6 +140,10 @@ transport_config:
 			path:   badLogMaskCfg,
 			expErr: errors.New("not a valid log level"),
 		},
+		"bad filter config": {
+			path:   badFilterCfg,
+			expErr: errors.New("agent config validation failed"),
```
kjacque, line 145: I think it would be better to test for the specific error here ("cannot specify both exclude_fabric_ifaces and include_fabric_ifaces").

kjacque (approved):
Nice. I wouldn't block on the unit test comment, I do think it should be as rigorous as the other cases though.

cdavis28 (dismissed):

src/control/cmd/daos_agent/fabric.go
```diff
@@ -92,9 +92,10 @@ type NUMAFabric struct {
 
 	numaMap NUMAFabricMap
 
-	currentNumaDevIdx map[int]int // current device idx to use on each NUMA node
-	currentNUMANode   int         // current NUMA node to search
-	ignoreIfaces      common.StringSet
+	currentNumaDevIdx  map[int]int      // current device idx to use on each NUMA node
+	currentNUMANode    int              // current NUMA node to search
+	ifaceFilterSet     common.StringSet // set of interface names for filtering
+	ifaceFilterExclude bool             // true: exclude filtered interfaces, false: include only filtered interfaces
```
pavelgritsai, line 98: bool is not the best design here (false meaning include? when the size of the set above is not 0?)

consider instead defining a new type (ifaceFilter) with mode (include/exclude as const/enum? and then not needing to check the size of the set but check if filter is nil or not.

src/control/cmd/daos_agent/fabric_test.go
```diff
@@ -723,6 +724,46 @@ func TestAgent_NUMAFabric_GetDevice(t *testing.T) {
...
+					NetDevClass: hardware.Ether,
+				},
+				{
+					Name:        "t2",
+					Domain:      "t2",
+					NetDevClass: hardware.Ether,
+				},
+			},
+		},
 		"ignore interface": {
```
pavelgritsai, line 767: using ignore and exclude to mean the same thing may seem confusing - should we just use one term?

pavelgritsai (dismissed):

src/control/cmd/daos_agent/fabric.go
mjmac: Hmm. I didn't want to overengineer this as it's all internal to the agent, but I can take another look at it. I didn't really like this either, tbh.

src/control/cmd/daos_agent/config_test.go
mjmac: Sure, I can do that.

kjacque: I lean toward the latter approach, as it verifies that we are actually testing what we think we are testing. I have definitely written unit tests that initially fail in a different location than I expected, because I overlooked some element of the setup.

src/control/cmd/daos_agent/fabric.go
kjacque: An alternative could be to pass in the desired filter function as a parameter directly.

pavelgritsai (approved):

PR #15488, opened 2024-11-11
DAOS-16686 dfuse: Move pre-read code to inode from file handle.

ashleypittman:
Attach pre-read buffers to the inode rather than open file handle.  This code
was written with the idea that a client would open and read the file and that
would populate the kernel cache however in practice there are often concurrent
readers for the same file and what was happening was that the first-to-open
was doing the pre-read but this was often not the first process to perform a read
and furthermore often there would be multiple reads for the same regions, all
of which would hit the network.

Use the "active" entry on the inode to launch a pre-read on first open and have
any request on any file handle use the pre-read buffer for replies if possible.

In addition, when a read is to be serviced from the pre-read buffer but the
data is not yet in memory rather than spinning on a lock consuming a fuse
thread add a descriptor to a callback list so that the thread is released and
the reply is made sooner when the data is available.

This greatly reduces the number of duplicate network round-trip reads for
workloads where multiple clients are trying to fetch the same data, something
that we see a lot in some applications.


src/client/dfuse/ops/open.c
```diff
@@ -67,6 +67,7 @@ dfuse_cb_open(fuse_req_t req, fuse_ino_t ino, struct fuse_file_info *fi)
 		 * which pre-existed in the container.
 		 */
 
+		/* TODO: This probably wants reflowing to not reference ie_open_count */
```
ashleypittman, line 70: I have two other PRs that depend on this one and cover the same area, I can look at this issue in the round once the three of them have all landed.

jolivier23 (approved):

mchaarawi (approved):

PR #15525, opened 2024-11-20
DAOS-xxx client: Dump metrics to container

mjmac:
If configured, dump metrics to a container instead of
a directory.


src/client/dfs/io.c
```diff
@@ -85,11 +97,14 @@ dfs_read_int(dfs_t *dfs, dfs_obj_t *obj, daos_off_t off, dfs_iod_t *iod, d_sg_li
...
 		D_GOTO(err_params, rc);
 
+	DFS_OP_STAT_INCR(dfs, DOS_READ);
 	/*
 	 * dc_task_schedule() calls tse_task_complete() even on error (which also calls the
 	 * completion cb that frees params in this case, so we can just ignore the rc here.
 	 */
 	dc_task_schedule(task, true);
+
+	dfs_update_file_metrics(dfs, *params->read_size, 0);
```
mchaarawi, line 107: i think those metrics update calls should be in the completion callback when the read completes

src/client/dfs/obj.c
```diff
@@ -412,6 +412,12 @@ open_stat(dfs_t *dfs, dfs_obj_t *parent, const char *name, mode_t mode, int flag
 
 out:
 	if (rc == 0) {
+		if (flags & O_CREAT) {
+			DFS_OP_STAT_INCR(dfs, DOS_CREATE);
```
mchaarawi, line 416: i  think it would be better to not have a CREATE metric, but re-use the mkdir (because that is what open with O_CREAT for dir is) and File_Create or just CREATE for creating a file.

src/client/dfs/obj.c
```diff
@@ -412,6 +412,12 @@ open_stat(dfs_t *dfs, dfs_obj_t *parent, const char *name, mode_t mode, int flag
...
+		} else {
+			DFS_OP_STAT_INCR(dfs, DOS_OPEN);
```
mchaarawi, line 418: i would also split this into opendir and open

src/client/dfs/obj.c
```diff
@@ -1662,6 +1679,7 @@ dfs_punch(dfs_t *dfs, dfs_obj_t *obj, daos_off_t offset, daos_size_t len)
 		return daos_der2errno(rc);
 	}
 
+	DFS_OP_STAT_INCR(dfs, DOS_PUNCH);
```
mchaarawi, line 1682: punch is not a posix term, so maybe better one would be truncate?

src/client/dfs/obj.c
```diff
@@ -1708,6 +1726,7 @@ dfs_sync(dfs_t *dfs)
 	if (dfs->amode != O_RDWR)
 		return EPERM;
 
+	DFS_OP_STAT_INCR(dfs, DOS_SYNC);
```
mchaarawi, line 1729: since this is a no-op, I would not increment the metric for it as it can cause confusion.

docs/user/container.md
```diff
@@ -507,6 +507,12 @@ corruption proactively. (see data_integrity.md for more details). This can be
 disabled per container using the `DAOS_PROP_CO_SCRUBBER_DISABLED` container
 property.
 
+### Per-Container Metrics Collection
+
+If enabled, per-container metrics (currently only POSIX containers are supported)
+will be collected on the client. Set the `DAOS_PROP_CO_METRICS_ENABLED` container
```
mchaarawi, line 513: i don't think this is a good idea. we can discuss offline

mjmac:
@mchaarawi: This is the combined patch for playing around with the feature. I'll incorporate your feedback into the standalone PR for DFS metrics: https://github.com/daos-stack/daos/pull/15544

src/client/dfs/io.c
mjmac: Fixed.

PR #15526, opened 2024-11-21
DAOS-16822 client: Add mkdir_p variant to dfs_sys API

mjmac:
As a usability enhancement, add a mkdir variant with
the following properties:
  * If the directory or any of its parents already exist
    and are directories, don't return an error
  * If any of the parent directories need to be created,
    automatically do so as a prerequisite step to creating
    the final child directory

Features: dfs
Required-githooks: true

### DIFF
--- a/src/client/dfs/common.c
+++ b/src/client/dfs/common.c
@@ -629,29 +629,13 @@ entry_stat(dfs_t *dfs, daos_handle_t th, daos_handle_t oh, const char *name, siz
 }
 
 /*
- * create a dir object. If caller passes parent obj, we check for existence of
- * object first.
+ * Create a dir object. If caller passes parent obj, and cid is not set,
+ * the child oclass is taken from the parent.
  */
 int
 create_dir(dfs_t *dfs, dfs_obj_t *parent, daos_oclass_id_t cid, dfs_obj_t *dir)
 {
-	bool             exists;
-	struct dfs_entry entry = {0};
-	size_t           len   = strnlen(dir->name, DFS_MAX_NAME);
 	int rc;
-
-	if (parent != NULL) {
-		rc = fetch_entry(dfs->layout_v, parent->oh, dfs->th, dir->name, len, false, &exists,
-				 &entry, 0, NULL, NULL, NULL);
-		if (rc != 0) {
-			DL_ERROR(rc, "fetch_entry() %s failed.", dir->name);
-			return rc;
-		} else if (exists) {
-			if (S_ISDIR(entry.mode))
-				return EEXIST;
-			return ENOTDIR;
-		}
-	}
 
 	/** set oclass for dir. order: API, parent dir, cont default */
 	if (cid == 0) {
--- a/src/client/dfs/common.c
+++ b/src/client/dfs/common.c
@@ -635,7 +635,23 @@ entry_stat(dfs_t *dfs, daos_handle_t th, daos_handle_t oh, const char *name, siz
 int
 create_dir(dfs_t *dfs, dfs_obj_t *parent, daos_oclass_id_t cid, dfs_obj_t *dir)
 {
+	bool             exists;
+	struct dfs_entry entry = {0};
+	size_t           len   = strnlen(dir->name, DFS_MAX_NAME);
 	int rc;
+
+	if (parent != NULL) {
+		rc = fetch_entry(dfs->layout_v, parent->oh, dfs->th, dir->name, len, false, &exists,
+				 &entry, 0, NULL, NULL, NULL);
+		if (rc != 0) {
+			DL_ERROR(rc, "fetch_entry() %s failed.", dir->name);
+			return rc;
+		} else if (exists) {
+			if (S_ISDIR(entry.mode))
+				return EEXIST;
+			return ENOTDIR;
+		}
+	}
 
 	/** set oclass for dir. order: API, parent dir, cont default */
 	if (cid == 0) {
--- a/src/client/dfs/dfs_sys.c
+++ b/src/client/dfs/dfs_sys.c
@@ -1400,43 +1400,32 @@ out_free_path:
 }
 
 int
-dfs_sys_mkdir_p(dfs_sys_t *dfs_sys, const char *dir, mode_t mode, daos_oclass_id_t cid)
+dfs_sys_mkdir_p(dfs_sys_t *dfs_sys, const char *dir_path, mode_t mode, daos_oclass_id_t cid)
 {
-	int         rc      = 0;
-	int         dir_len = strnlen(dir, PATH_MAX);
-	struct stat st      = {0};
-	char       *_dir    = NULL;
+	int         path_len = strnlen(dir_path, PATH_MAX);
+	char       *_path    = NULL;
 	char       *ptr     = NULL;
+	int         rc       = 0;
 
 	if (dfs_sys == NULL)
 		return EINVAL;
-	if (dir == NULL)
+	if (dir_path == NULL)
 		return EINVAL;
-	if (dir_len == PATH_MAX)
-		return EINVAL;
+	if (path_len == PATH_MAX)
+		return ENAMETOOLONG;
 
-	/* quick check -- if it already exists we can exit early */
-	rc = dfs_sys_stat(dfs_sys, dir, 0, &st);
-	if (rc == 0) {
-		if ((st.st_mode & S_IFMT) != S_IFDIR) {
-			D_ERROR("%s exists and is not a directory", dir);
-			return ENOTDIR;
-		}
-		return 0;
-	}
-
-	D_STRNDUP(_dir, dir, dir_len);
-	if (_dir == NULL)
+	D_STRNDUP(_path, dir_path, path_len);
+	if (_path == NULL)
 		return ENOMEM;
 
 	/* iterate through the parent directories and create them if necessary */
-	for (ptr = _dir + 1; *ptr != '\0'; ptr++) {
+	for (ptr = _path + 1; *ptr != '\0'; ptr++) {
 		if (*ptr != '/')
 			continue;
 
 		/* truncate the string here to create the parent */
 		*ptr = '\0';
-		rc   = dfs_sys_mkdir(dfs_sys, _dir, mode, cid);
+		rc   = dfs_sys_mkdir(dfs_sys, _path, mode, cid);
 		if (rc != 0 && rc != EEXIST)
 			D_GOTO(out_free, rc);
 		/* reset to keep going */
@@ -1444,12 +1433,12 @@ dfs_sys_mkdir_p(dfs_sys_t *dfs_sys, const char *dir, mode_t mode, daos_oclass_id
 	}
 
 	/* create the final directory */
-	rc = dfs_sys_mkdir(dfs_sys, _dir, mode, cid);
+	rc = dfs_sys_mkdir(dfs_sys, _path, mode, cid);
 	if (rc == EEXIST)
 		rc = 0;
 
 out_free:
-	D_FREE(_dir);
+	D_FREE(_path);
 	return rc;
 }
 
--- a/src/client/dfs/dfs_sys.c
+++ b/src/client/dfs/dfs_sys.c
@@ -1399,12 +1399,32 @@ out_free_path:
 	return rc;
 }
 
+static int
+check_existing_dir(dfs_sys_t *dfs_sys, const char *dir_path)
+{
+	struct stat st = {0};
+	int         rc;
+
+	rc = dfs_sys_stat(dfs_sys, dir_path, 0, &st);
+	if (rc != 0) {
+		D_DEBUG(DB_TRACE, "failed to stat %s: (%d)\n", dir_path, rc);
+		return rc;
+	}
+
+	/* if it's not a directory, fail */
+	if (!S_ISDIR(st.st_mode))
+		return EEXIST;
+
+	/* if it is a directory, then it's not an error */
+	return 0;
+}
+
 int
 dfs_sys_mkdir_p(dfs_sys_t *dfs_sys, const char *dir_path, mode_t mode, daos_oclass_id_t cid)
 {
 	int         path_len = strnlen(dir_path, PATH_MAX);
 	char       *_path    = NULL;
-	char       *ptr     = NULL;
+	char       *ptr      = NULL;
 	int         rc       = 0;
 
 	if (dfs_sys == NULL)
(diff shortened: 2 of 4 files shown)


src/client/dfs/dfs_sys.c
```diff
@@ -1399,6 +1399,60 @@ dfs_sys_mkdir(dfs_sys_t *dfs_sys, const char *dir, mode_t mode,
...
+
+	if (dfs_sys == NULL)
+		return EINVAL;
+	if (dir == NULL)
+		return EINVAL;
+	if (dir_len == PATH_MAX)
+		return EINVAL;
+
+	/* quick check -- if it already exists we can exit early */
+	rc = dfs_sys_stat(dfs_sys, dir, 0, &st);
```
daltonbohning, line 1419: It's unfortunate we need this extra stat. In the future we could optimize around this by adding support to the helper functions or something to use `dfs_open_stat` to create + stat if existing, or maybe some other mechanism

daltonbohning (commented):
Mostly LGTM. Just nitpicks on errno and I think we should add a simple test here:
https://github.com/daos-stack/daos/blob/3cb9b576fd9a649ddaf6db4dcd1c80741c1daae2/src/tests/suite/dfs_sys_unit_test.c#L145

src/client/dfs/dfs_sys.c
```diff
@@ -1399,6 +1399,60 @@ dfs_sys_mkdir(dfs_sys_t *dfs_sys, const char *dir, mode_t mode,
...
+	struct stat st      = {0};
+	char       *_dir    = NULL;
+	char       *ptr     = NULL;
+
+	if (dfs_sys == NULL)
+		return EINVAL;
+	if (dir == NULL)
+		return EINVAL;
+	if (dir_len == PATH_MAX)
+		return EINVAL;
```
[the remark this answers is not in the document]
mjmac: Done.

src/client/dfs/dfs_sys.c
mjmac: Yeah, it's only happening for the target directory though, so it seems like a minor inefficiency.

src/include/daos_fs_sys.h
```diff
@@ -542,6 +542,19 @@ int
...
+ *
+ * \param[in]	dfs_sys Pointer to the mounted file system.
+ * \param[in]	dir	Link path of new dir.
+ * \param[in]	mode	mkdir mode.
+ * \param[in]	cid	DAOS object class id (pass 0 for default MAX_RW).
+ *
+ * \return		0 on success, errno code on failure.
+ */
+int
+dfs_sys_mkdir_p(dfs_sys_t *dfs_sys, const char *dir, mode_t mode, daos_oclass_id_t cid);
```
mchaarawi, line 556: what is the reasoning for creating this at the sys level and not at the lower level dfs api level and then just add the sys wrapper?

src/client/dfs/dfs_sys.c
mchaarawi: im not sure why this is required though. you would need to recreate the parent dirs if they don't exist anyway, so you will need to parse every entry anyway. so let's say you are creating: /a/b/c/d
/a/b/c already exisits.
you are walking /a/b/c/ here, and then walking it again in the mkdir loop at the bottom.

mchaarawi (changes requested):
a dfs test is required for any new api in:
src/test/suite/dfs_*

I also see no reason why the api is added at dfs_sys and not the dfs level. the former is only a wrapper over the latter.

src/client/dfs/dfs_sys.c
mchaarawi: > Yeah, it's only happening for the target directory though, so it seems like a minor inefficiency.

it's not actually. stat will need to walk the entire path.

src/include/daos_fs_sys.h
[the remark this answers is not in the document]
mjmac: OK, I'll update it.

mchaarawi: indeed you are correct. probably just checking dfs_mkdir EEXIST error should be sufficient and not necessary to add dfs_mkdir_p. sorry :-)

daltonbohning: The only public DFS function that takes a path is `dfs_lookup`, which doesn't do any creates. The main motivation for DFS Sys was that it accepts a path instead of parent object handle.
It would be a bit more work but @mchaarawi should we modify `lookup_rel_path` (internal function) to conditionally create as it walks the path?

mchaarawi: we can explore sure, but way out of scope for this PR.
i always forget that we have the sys api for these lookup traversals :-)

daltonbohning: Yes, agree it's out of scope here. We can always rework some internals later to optimize

src/client/dfs/dfs_sys.c
mjmac: OK, coming back to this... I removed the stat, but now my new unit test fails because it seems like mkdir() with an existing file doesn't fail:

```C
	/* this shouldn't work */
	rc = dfs_sys_mkdir_p(dfs_sys_mt, file, S_IWUSR | S_IRUSR, 0);
	assert_int_equal(rc, EEXIST);
```

Looking at [the comment](https://github.com/daos-stack/daos/blob/master/src/client/dfs/common.c#L632) for `create_dir()`, I'm wondering if maybe some code was lost along the way? There's no existence check, unless I'm missing something. The only thing the parent object is used for is getting the dir oclass.

daltonbohning: I'm seeing the same. `create_dir` just creates the object, then `dfs_mkdir` tries to insert the metadata with `DAOS_COND_DKEY_INSERT`. So if the dkey (name of the file or dir) already exists then it returns success
https://github.com/daos-stack/daos/blob/b20a41dafec49582dcdb27d917b23bccea4468db/src/client/dfs/dir.c#L58


Side note @mchaarawi what happens when `create_dir` succeeds, which does `daos_obj_open`, and then `dfs_mkdir` calls `daos_obj_close`? Does that new obj still exist? with no reference to it?

mjmac: I just restored (or added?) the existence check. If the entry exists and is a directory, then the check will return EEXIST. If it exists and is not a directory, it will return ENOTDIR. I didn't see a better way to distinguish between these states.

mchaarawi: the DAOS_COND_DKEY_INSERT check should return DER_EXISTS if the entry exists.
@daltonbohning objects in DAOS have no "existence" so just opening / closing an object with the obj API does not consume anything. they are just local operations on the server.

src/client/dfs/common.c
```diff
@@ -635,8 +635,24 @@ entry_stat(dfs_t *dfs, daos_handle_t th, daos_handle_t oh, const char *name, siz
...
+				 &entry, 0, NULL, NULL, NULL);
+		if (rc != 0) {
+			DL_ERROR(rc, "fetch_entry() %s failed.", dir->name);
+			return rc;
+		} else if (exists) {
+			if (S_ISDIR(entry.mode))
+				return EEXIST;
+			return ENOTDIR;
+		}
+	}
```
mchaarawi, line 654: the whole point of conditional inserts is to avoid doing an existence check to improve dfs IOPS performance. so this is not going to be acceptable.
the original comment above create_dir i believe was just left over before we had implemented conditional operations at the object level.

src/client/dfs/dfs_sys.c
```diff
@@ -1399,6 +1399,49 @@ dfs_sys_mkdir(dfs_sys_t *dfs_sys, const char *dir, mode_t mode,
...
+		rc   = dfs_sys_mkdir(dfs_sys, _path, mode, cid);
+		if (rc != 0 && rc != EEXIST)
+			D_GOTO(out_free, rc);
+		/* reset to keep going */
+		*ptr = '/';
+	}
+
+	/* create the final directory */
+	rc = dfs_sys_mkdir(dfs_sys, _path, mode, cid);
+	if (rc == EEXIST)
```
mchaarawi, line 1437: i think the main issue why your test fails is this line here. dfs is returning EEXIST saying there is an entry (a file) called whoops and you set that to success. i think this is a bug in your PR

mjmac: But if you've already got a file named `/a/b/whoops` and you do `mkdir(/a/b/whoops)`, it shouldn't succeed, right? Without the existence check that I re-added, running mkdir on a file doesn't fail. Conversely, running regular mkdir on a directory that does exist also doesn't fail. Those both seem pretty broken to me.

I'm guessing that handling this in the conditional insert may be difficult or impossible because at that level it's just objects and there's no notion of file vs. directory, right?

daltonbohning: So even C `mkdir()` sets errno=EEXIST whether it is an existing file or directory.
But the behavior @mjmac is after is similar to what bash `mkdir -p` does.
Existing file will error
```
$ touch file
$ mkdir -p file
mkdir: cannot create directory ‘file’: File exists
```
Existing dir will not
```
$ mkdir dir
$ mkdir -p dir
```

daltonbohning: I think `dfs_mkdir` returning EEXIST for both cases makes sense, so then `dfs_sys_mkdir_p` should handle this?

src/client/dfs/dfs_sys.c
mchaarawi: i just verified that DFS does return EEXISTS in this case, but the new mkdir_p function coverts that to success..  i think the main issue is that we need to distinguish between EEXISTS and ENOTDIR for this function. so in that case i would suggest that an extra check is added here in case EEXIST is returned ot verify it's a dir and return ENOTDIR other wise. does that sound sensible?

src/client/dfs/dfs_sys.c
mchaarawi: > running mkdir on a file doesn't fail.

well it doesn't fail with your function but dfs does fail it. you are just converting the error to success

mchaarawi: > Conversely, running regular mkdir on a directory that does exist also doesn't fail. Those both seem pretty broken to me.

i also challenge you on that. dfs_mkdir() will still fail.. again it depends on how you are handling the failure

mchaarawi: > I think `dfs_mkdir` returning EEXIST for both cases makes sense, so then `dfs_sys_mkdir_p` should handle this?

yes agreed. this is what my suggestion was.. we will **not** add an extra fetch on the happy path given we have conditional ops.

src/client/dfs/dfs_sys.c
```diff
@@ -1399,6 +1399,49 @@ dfs_sys_mkdir(dfs_sys_t *dfs_sys, const char *dir, mode_t mode,
...
+		return ENOMEM;
+
+	/* iterate through the parent directories and create them if necessary */
+	for (ptr = _path + 1; *ptr != '\0'; ptr++) {
+		if (*ptr != '/')
+			continue;
+
+		/* truncate the string here to create the parent */
+		*ptr = '\0';
+		rc   = dfs_sys_mkdir(dfs_sys, _path, mode, cid);
```
mchaarawi, line 1428: i think there is another inefficiency here with this loop. every iteration is passing the full path right?
for example if dir_path is /a/b/c/d/e
we are calling dfs_sys_mkdir with
/a
/a/b
/a/b/c
...

This is also doing unnecessary path walks every iteration. (please correct me if i understood your parsing wrong).
If this is the case, then it would be much more efficient to not use dfs_sys_mkdir() here and use dfs_mkdir() instead or even dfs_open with O_CREAT and DIR type to create the directories and use the parent from the previous iteration to avoid the intermediate lookups in every iteration.

src/client/dfs/dfs_sys.c
mjmac: > > Conversely, running regular mkdir on a directory that does exist also doesn't fail. Those both seem pretty broken to me.
> 
> i also challenge you on that. dfs_mkdir() will still fail.. again it depends on how you are handling the failure

I must have misunderstood what I was seeing. I just added a new unit test for the regular mkdir and it fails as expected.

mjmac: > > I think `dfs_mkdir` returning EEXIST for both cases makes sense, so then `dfs_sys_mkdir_p` should handle this?
> 
> yes agreed. this is what my suggestion was.. we will **not** add an extra fetch on the happy path given we have conditional ops.

OK, this makes sense. I think I'm going to need to do it for each parent, though. Sigh, it's never as easy as it first seems.

src/client/dfs/dfs_sys.c
mjmac: I was thinking about this... But perhaps I misunderstood the point of the hash table in dfs_sys. @daltonbohning: Shouldn't the parent objects be pulled out of that instead of having to do the lookups?

src/client/dfs/common.c
mjmac: I'll re-remove the check and also update the comment to avoid future confusion. :)

src/client/dfs/dfs_sys.c
daltonbohning: The dfs sys hash table is to avoid looking up the same path again. But `hash_lookup` doesn't recursively cache. So this works okay
```
dfs_sys_mkdir /a     # lookup a, cache a
dfs_sys_mkdir /a/b   # a from cache, lookup b, cache b
dfs_sys_mkdir /a/b/c # b from cache, lookup c, cache c
```
but the reverse is less optimal
```
dfs_sys_mkdir /a/b/c  # none cached, dfs_lookup traverses a,b,c
                      # cache c - ONLY c is cached!
dfs_sys_mkdir /a/b    # none cached! dfs_lookup traverses a,b
                      # cache b
```

daltonbohning: But if caching is turned off, Mohamad's point is correct.
I guess it's possible to optimize for both cases, but that's more work and probably out of scope :)

mjmac: I added some debugging lines and it seems like it's doing what we expect:
```
[ RUN      ] DFS_SYS_UNIT_TEST14: DFS Sys mkdir_p
daos ERR  src/client/dfs/dfs_sys.c:204 hash_lookup() hash hit: /
daos ERR  src/client/dfs/dfs_sys.c:208 hash_lookup() hash miss: /a
daos ERR  src/client/dfs/dfs_sys.c:204 hash_lookup() hash hit: /
daos ERR  src/client/dfs/dfs_sys.c:204 hash_lookup() hash hit: /a/b
daos ERR  src/client/dfs/dfs_sys.c:204 hash_lookup() hash hit: /
daos ERR  src/client/dfs/dfs_sys.c:204 hash_lookup() hash hit: /a
daos ERR  src/client/dfs/dfs_sys.c:204 hash_lookup() hash hit: /a/b
```

Given that, does it alleviate your performance concern, @mchaarawi?

mchaarawi: As Dalton indicated, caching is optional. so you will still see unnecessary lookups when you have nested dirs here. but we can improve that later, so i won't hold this up bec of that.
The path parsing and existence check + type check can be combined together if we implement something similar within dfs itself. I will make a note to add something along that lines when i get a chance.

mchaarawi (approved):

PR #15528, opened 2024-11-21
DAOS-16686 dfuse: Detect matching reads to avoid network access.

ashleypittman:
If a read matches a current outstanding read then simply
connect the two and when there's a reply from the network
then respond to both requests.

Skip-func-test-vm: true
Test-tag: DaosBuild


src/client/dfuse/dfuse.h
```diff
@@ -1017,6 +1024,9 @@ struct dfuse_inode_entry {
 
 struct active_inode {
 	d_list_t               chunks;
+	d_list_t               open_reads;
+	bool                   seen_eof;
+	size_t                 file_size;
```
ashleypittman, line 1029: These two are not used in this PR, I'll remove them on re-push but I don't want to loose testing.

ashleypittman:
This is a self-contained subset of https://github.com/daos-stack/daos/pull/15298 and contains just the performance improvement for regular reads rather than any bug-fix to the chunk code.  For now it's just running the tests which are failing in the other PR, the hope being that it passes and we can land this one as a way of moving forward whilst I debug why the others failing.

jolivier23 (dismissed):

wangdi1 (approved):

PR #15529, opened 2024-11-21
DAOS-16824 cart: lower error messages to warn level

jxiong:
Lower some chatty messages from error level to warn level.

Ideally the error messages should be printed by callers because they should know how serious the error is.

This PR also removes some error messages because they are being printed repeatedly due to the same error.

Signed-off-by: Jinshan Xiong <jinshanx@google.com>

Allow-unstable-test: true
Change-Id: I71966a235c068abbd3c74c422458a68e3a86154a

### Gatekeeper:

### DIFF
--- a/src/cart/crt_context.c
+++ b/src/cart/crt_context.c
@@ -535,7 +535,7 @@ crt_rpc_complete_and_unlock(struct crt_rpc_priv *rpc_priv, int rc)
 
 		if (cbinfo.cci_rc != 0)
 			RPC_CWARN(crt_quiet_error(cbinfo.cci_rc), DB_NET, rpc_priv,
-				 "failed, " DF_RC "\n", DP_RC(cbinfo.cci_rc));
+				  "failed, " DF_RC "\n", DP_RC(cbinfo.cci_rc));
 
 		RPC_TRACE(DB_TRACE, rpc_priv,
 			  "Invoking RPC callback (rank %d tag %d) rc: "
--- a/src/cart/crt_context.c
+++ b/src/cart/crt_context.c
@@ -1227,6 +1227,7 @@ crt_context_timeout_check(struct crt_context *crt_ctx)
 	struct d_binheap_node		*bh_node;
 	d_list_t			 timeout_list;
 	uint64_t			 ts_now;
+	bool				 print_once = false;
 
 	D_ASSERT(crt_ctx != NULL);
 
@@ -1260,11 +1261,21 @@ crt_context_timeout_check(struct crt_context *crt_ctx)
 		/* NB: The reason that the error message is printed at INFO
 		 * level is because the user should know how serious the error
 		 * is and they will print the RPC error at appropriate level */
-		RPC_INFO(rpc_priv,
-			 "ctx_id %d, (status: %#x) timed out (%d seconds), "
-			 "target (%d:%d)\n",
-			 crt_ctx->cc_idx, rpc_priv->crp_state, rpc_priv->crp_timeout_sec,
-			 rpc_priv->crp_pub.cr_ep.ep_rank, rpc_priv->crp_pub.cr_ep.ep_tag);
+		if (!print_once) {
+			print_once = true;
+
+			RPC_WARN(rpc_priv,
+				 "ctx_id %d, (status: %#x) timed out (%d seconds), "
+				 "target (%d:%d)\n",
+				 crt_ctx->cc_idx, rpc_priv->crp_state, rpc_priv->crp_timeout_sec,
+				 rpc_priv->crp_pub.cr_ep.ep_rank, rpc_priv->crp_pub.cr_ep.ep_tag);
+		} else {
+			RPC_INFO(rpc_priv,
+				 "ctx_id %d, (status: %#x) timed out (%d seconds), "
+				 "target (%d:%d)\n",
+				 crt_ctx->cc_idx, rpc_priv->crp_state, rpc_priv->crp_timeout_sec,
+				 rpc_priv->crp_pub.cr_ep.ep_rank, rpc_priv->crp_pub.cr_ep.ep_tag);
+		}
 
 		crt_req_timeout_hdlr(rpc_priv);
 		RPC_DECREF(rpc_priv);
--- a/src/cart/crt_context.c
+++ b/src/cart/crt_context.c
@@ -534,8 +534,8 @@ crt_rpc_complete_and_unlock(struct crt_rpc_priv *rpc_priv, int rc)
 			cbinfo.cci_rc = rpc_priv->crp_reply_hdr.cch_rc;
 
 		if (cbinfo.cci_rc != 0)
-			RPC_CERROR(crt_quiet_error(cbinfo.cci_rc), DB_NET, rpc_priv,
-				   "failed, " DF_RC "\n", DP_RC(cbinfo.cci_rc));
+			RPC_WARN(rpc_priv,
+				 "failed, " DF_RC "\n", DP_RC(cbinfo.cci_rc));
 
 		RPC_TRACE(DB_TRACE, rpc_priv,
 			  "Invoking RPC callback (rank %d tag %d) rc: "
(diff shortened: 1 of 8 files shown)


src/cart/crt_context.c
```diff
@@ -1256,48 +1254,17 @@ crt_context_timeout_check(struct crt_context *crt_ctx)
...
-
-		} else {
-			RPC_INFO(rpc_priv,
-				 "ctx_id %d, (status: %#x) timed out (%d seconds), "
-				 "target (%d:%d)\n",
-				 crt_ctx->cc_idx, rpc_priv->crp_state, rpc_priv->crp_timeout_sec,
-				 rpc_priv->crp_pub.cr_ep.ep_rank, rpc_priv->crp_pub.cr_ep.ep_tag);
-			left_to_print--;
-		}
+		/* NB: The reason that the error message is printed at INFO
```
frostedcmos, line 1260: in many instances we dont have info level logging on by default. i would prefer the first message to remain as warn, while subsequent ones can be downgraded to info.

src/cart/crt_hg.c
```diff
@@ -1154,9 +1154,9 @@ crt_rpc_handler_common(hg_handle_t hg_hdl)
 	else
 		rc = crt_corpc_common_hdlr(rpc_priv);
 	if (unlikely(rc != 0)) {
-		RPC_ERROR(rpc_priv,
-			  "failed to invoke RPC handler, rc: "DF_RC"\n",
-			  DP_RC(rc));
+		RPC_WARN(rpc_priv,
+			 "failed to invoke RPC handler, rc: "DF_RC"\n",
+			 DP_RC(rc));
```
frostedcmos, line 1159: i expect either crp_rpc_common_hdlr() or crt_corpc_common_hdlr() to print reason for failure if they return rc != 0, so probably we can just nuke message here altogether

src/cart/crt_hg.c
```diff
@@ -1364,8 +1364,8 @@ crt_hg_req_send_cb(const struct hg_cb_info *hg_cbinfo)
 		crt_cbinfo.cci_rc = rc;
 
 		if (crt_cbinfo.cci_rc != 0)
-			RPC_CERROR(crt_quiet_error(crt_cbinfo.cci_rc), DB_NET, rpc_priv,
-				   "RPC failed; rc: " DF_RC "\n", DP_RC(crt_cbinfo.cci_rc));
+			RPC_WARN(rpc_priv,
+				 "RPC failed; rc: " DF_RC "\n", DP_RC(crt_cbinfo.cci_rc));
```
frostedcmos, line 1368: we need something like RPC_CWARN here that takes the same condition. you dont want to print warnings for all group version mismatches either, and those are very frequent

src/cart/crt_hg_proc.c
```diff
@@ -712,14 +712,9 @@ crt_proc_out_common(crt_proc_t proc, crt_rpc_output_t *data)
...
-				RPC_ERROR(rpc_priv,
-					  "RPC failed to execute on target. "
-					  "error code: "DF_RC"\n", DP_RC(rc2));
-			else
-				RPC_TRACE(DB_NET, rpc_priv,
-					  "RPC failed to execute on target. "
-					  "error code: "DF_RC"\n", DP_RC(rc2));
+			RPC_WARN(rpc_priv,
+				 "RPC failed to execute on target. "
+				 "error code: "DF_RC"\n", DP_RC(rc2));
```
frostedcmos, line 717: same as above, you dont really want to print these as warns for DER_GRPVER either

frostedcmos (commented):
comments inline

src/cart/crt_hg.c
jxiong: Yes, I kept the error message over there. How about we change this to INFO just in case it will miss thing?

src/cart/crt_context.c
jxiong: I can do that.

src/cart/crt_hg.c
jxiong: For sure.

src/cart/crt_context.c
```diff
@@ -534,8 +534,8 @@ crt_rpc_complete_and_unlock(struct crt_rpc_priv *rpc_priv, int rc)
 			cbinfo.cci_rc = rpc_priv->crp_reply_hdr.cch_rc;
 
 		if (cbinfo.cci_rc != 0)
-			RPC_CERROR(crt_quiet_error(cbinfo.cci_rc), DB_NET, rpc_priv,
-				   "failed, " DF_RC "\n", DP_RC(cbinfo.cci_rc));
+			RPC_WARN(rpc_priv,
```
frostedcmos, line 537: this needs to be RPC_CWARN as well

frostedcmos (commented):
looks good. 1 more place to change, noted inline

frostedcmos (approved):
looks good, thanks:)

jolivier23 (approved):

jxiong:
NLT is broken - I have seen this from multiple of my PRs.

mchaarawi:
i don't have a strong opinion one way or the other for this change.
we do have a lot of chatty error logging in several areas, especially timeouts.
lowering to info does not really do much, because 99% of users actually do use info logging, not ERR or WARN. 

my main concern on this is what if there are actual errors on the fabric (not just timeouts). are these also downgraded to warn / info messages?

jxiong:
> i don't have a strong opinion one way or the other for this change. we do have a lot of chatty error logging in several areas, especially timeouts. lowering to info does not really do much, because 99% of users actually do use info logging, not ERR or WARN.

Well, at least the users would have a choice to adjust the log error. Right now, the rate error message is at ~400 messages per second at error level.

On the server side, one of the major problem is that the same error now is printed thrice, in crt_rpc_handler_common(),  crt_rpc_common_hdlr(), and crt_proc_out_common() respectively. This PR lowers the message level in crt_rpc_handler_common() and crt_proc_out_common() but keeps the error message on crt_rpc_common_hdlr().

> 
> my main concern on this is what if there are actual errors on the fabric (not just timeouts). are these also downgraded to warn / info messages?

This PR is not intended to change any internal errors in fabric. I tend to think the framework should only mind its own internal errors. The users will know how serious an error is then decide to print out something. As my example shows in the ticket, evetually the error will be printed by CLI, why would the framework bother printing those errors? 

Another issue is the same as I mentioned previously - the same error is being printed over and over again in the stack. This PR does its best to remove those duplicated messages.

jolivier23:
> i don't have a strong opinion one way or the other for this change. we do have a lot of chatty error logging in several areas, especially timeouts. lowering to info does not really do much, because 99% of users actually do use info logging, not ERR or WARN.
> 
> my main concern on this is what if there are actual errors on the fabric (not just timeouts). are these also downgraded to warn / info messages?

@mchaarawi our logs explorer allows us to filter messages based on the log level during or after the logs have been produced.  Also, ERROR messages are marked "red" in the timeline and it would be useful to only see red when it's an actual error to make it easier to spot the root cause of real problems without having to sift through noise.  Additionally, there is a psychological issue for customer who sees a bunch of errors vs warnings or informational messages.   If we flag everything as an error, especially when it's not fatal, it gives a worse impression, IMO

soumagne (dismissed):

jolivier23 (dismissed):

frostedcmos (approved):

jolivier23:
Looks like it hit some NLT issues.  Try merging latest and running with 

Allow-unstable-test: true in final commit message to make sure it runs full tests regardless.

src/tests/simple_dfs.c
```diff
@@ -48,8 +48,16 @@ main(int argc, char **argv)
 	rc = dfs_init();
```
mchaarawi, line 48: changes in this file are unrelated to the PR. a wrong merge perhaps?

jxiong: oops

frostedcmos (approved):

jolivier23 (approved):

PR #15539, opened 2024-11-26
DAOS-16830 pool: Deprecate DAOS_POOL_RF

jolivier23:
Rename DAOS_POOL_RF to DAOS_POOL_EXCLUDE_LIMIT.

The old variable will still work.  Also changed description and allow setting to a value lower than the RF.

Tested this locally and behaves as expected.

Required-githooks: true

### Gatekeeper:


phender (changes requested):
I think you're missing https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/server_utils_params.py#L437

jolivier23:
> I think you're missing https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/server_utils_params.py#L437

I left that one on purpose to test that it still works

phender:
> > I think you're missing https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/server_utils_params.py#L437
> 
> I left that one on purpose to test that it still works

The problem is the `REQUIRED_ENV_VARS` list ensures that every server config file will specify `DAOS_POOL_RF=4` unless the test's yaml defines a different value for `DAOS_POOL_RF`, so when `src/tests/ftest/control/dmg_pool_query_ranks.yaml` is modified to specify `DAOS_EVICTION_THRESHOLD` the code will generate a server config file with `DAOS_POOL_RF=4` and `DAOS_EVICTION_THRESHOLD=1`.

jolivier23:
> > > I think you're missing https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/server_utils_params.py#L437
> > 
> > 
> > I left that one on purpose to test that it still works
> 
> The problem is the `REQUIRED_ENV_VARS` list ensures that every server config file will specify `DAOS_POOL_RF=4` unless the test's yaml defines a different value for `DAOS_POOL_RF`, so when `src/tests/ftest/control/dmg_pool_query_ranks.yaml` is modified to specify `DAOS_EVICTION_THRESHOLD` the code will generate a server config file with both `DAOS_POOL_RF=4` and `DAOS_EVICTION_THRESHOLD=1` defined.

This is actually another good thing to test too.   We want DAOS_EVICTION_THRESHOLD to take precedent when both are specified.

jolivier23:
> > > > I think you're missing https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/server_utils_params.py#L437
> > > 
> > > 
> > > I left that one on purpose to test that it still works
> > 
> > 
> > The problem is the `REQUIRED_ENV_VARS` list ensures that every server config file will specify `DAOS_POOL_RF=4` unless the test's yaml defines a different value for `DAOS_POOL_RF`, so when `src/tests/ftest/control/dmg_pool_query_ranks.yaml` is modified to specify `DAOS_EVICTION_THRESHOLD` the code will generate a server config file with both `DAOS_POOL_RF=4` and `DAOS_EVICTION_THRESHOLD=1` defined.
> 
> This is actually another good thing to test too. We want DAOS_EVICTION_THRESHOLD to take precedent when both are specified. That should be fine to have both. It will see that the first one is set and not even check DAOS_POOL_RF which is the desired behavior.

So really the only issue is that for all other tests, we will have a deprecation warning in the logs but I assume that's not checked anywhere

docs/admin/env_variables.md
```diff
@@ -54,7 +54,7 @@ Environment variables in this section only apply to the server side.
 |DAOS\_DTX\_RPC\_HELPER\_THD|DTX RPC helper threshold. The valid range is [18, unlimited). The default value is 513.|
 |DAOS\_DTX\_BATCHED\_ULT\_MAX|The max count of DTX batched commit ULTs. The valid range is [0, unlimited). 0 means to commit DTX synchronously. The default value is 32.|
 |DAOS\_FORWARD\_NEIGHBOR|Set to enable I/O forwarding on neighbor xstream in the absence of helper threads.|
-|DAOS\_POOL\_RF|Redundancy factor for the pool. The valid range is [1, 4]. The default value is 2.|
+|DAOS\_EVICTION\_THRESHOLD|Number of failures after which we stop evicting ranks from the pool map. The valid range is [0, 4]. The default value is 2.|
```
liw, line 57: "Evict" refers to "evicting pool/container handles" in our current UI and code; the right term should be "exclude", I believe. Also, the "pool" is an important word in the current name. I'm not very sure why we want to rename the env var.

jolivier23: Ok, I'll change it

jolivier23: Will change it.  The reason to rename is because it's very confusing.

We already have a pool rf property which defines the minimum rf for all containers.  Also, what this controls really has nothing to do with the rf but rather the threshold where we switch off exclusion.  For the Google usage, we want to just set it to zero because we can quickly restore the service and otherwise would entail a full rebuild and map change.

jolivier23: I'm open to better ideas. But given that we have a pool property that controls the minimum rf and that container rf can be even greater, this is a little bit weird to associate with the rf.  On the other hand, probably don't need a critical message unless it actually does exceed the actual rf.  But for our use case, the new name makes more sense and I think it also works for the existing use case.

liw: Being able to set the environment variable to 0 seems like a reasonable request to me.

The pool RF property (unfortunately) does not define the minimum RF for its containers. It should. Moreover, supporting different container RFs turns out to be very hard. We introduced the environment variable as quick, temporary workaround for 2.6. It doesn't fit well with other RF-related concepts (which may not be as consistent as we would like they to be). @liuxuezhao, @gnailzenh, anything you would like to correct or add here?

jolivier23: If pool rf property doesn't define the minimum, why can't I create a container with a lower rf that what is set by that property?

liw: Not implying that this is the correct behavior, but I just tried master:
```
[liw2@wolf-410 daos]$ dmg -o ~/daos_control.yml pool create --scm-size=4g --nvme-size=4g --properties=rd_fac:2 test-pool-0
Creating DAOS pool with manual per-engine storage allocation: 4.0 GB SCM, 4.0 GB NVMe (100.00% ratio)
Pool created with 50.00%,50.00% storage tier ratio
--------------------------------------------------
  UUID                 : 485017cb-3268-4bea-b079-6d2df85865e3
  Service Leader       : 0
  Service Ranks        : [0-3,6]
  Storage Ranks        : [0-7]
  Total Size           : 64 GB
  Storage tier 0 (SCM) : 32 GB (4.0 GB / rank)
  Storage tier 1 (NVMe): 32 GB (4.0 GB / rank)

[liw2@wolf-410 daos]$ daos container create -t posix --properties=rd_fac:1 test-pool-0 test-cont-2
Successfully created container 1ad90506-153e-45fb-b4fd-dfa7f0c12330
  Container UUID : 1ad90506-153e-45fb-b4fd-dfa7f0c12330
  Container Label: test-cont-2                         
  Container Type : POSIX                               

[liw2@wolf-410 daos]$ daos container get-prop test-pool-0 test-cont-2
Properties for container test-cont-2
Name                                             Value                               
----                                             -----                               
[...]
Redundancy Factor (rd_fac)                       rd_fac1                             
[...]
```

phender:
> > > > > I think you're missing https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/server_utils_params.py#L437
> > > > 
> > > > 
> > > > I left that one on purpose to test that it still works
> > > 
> > > 
> > > The problem is the `REQUIRED_ENV_VARS` list ensures that every server config file will specify `DAOS_POOL_RF=4` unless the test's yaml defines a different value for `DAOS_POOL_RF`, so when `src/tests/ftest/control/dmg_pool_query_ranks.yaml` is modified to specify `DAOS_EVICTION_THRESHOLD` the code will generate a server config file with both `DAOS_POOL_RF=4` and `DAOS_EVICTION_THRESHOLD=1` defined.
> > 
> > 
> > This is actually another good thing to test too. We want DAOS_EVICTION_THRESHOLD to take precedent when both are specified. That should be fine to have both. It will see that the first one is set and not even check DAOS_POOL_RF which is the desired behavior.
> 
> So really the only issue is that for all other tests, we will have a deprecation warning in the logs but I assume that's not checked anywhere

I'm fairly certain we don't check for deprecation warnings in the logs.

While I agree it's a good idea to have tests to verify that DAOS_EVICTION_THRESHOLD takes precedent over DAOS_POOL_RF in the server config and verifying that DAOS_POOL_RF still works if specified by itself (and probably verifying the deprecation warning as well), I feel this testing should be its own isolated test for a few reasons:
1) If someone down the line removes support for DAOS_POOL_RF, only one test would fail - not all of them
2) For a test not validating the DAOS_EVICTION_THRESHOLD/DAOS_POOL_RF entry it adds confusion if the two are used interchangeably - especially where one is always specified, but overridden by a different variable name.

docs/admin/env_variables.md
jolivier23: You are right.   I'm not sure what issue I ran into before now but I recall not being able to create a container with rd 0.  Maybe it changed after 2.4?  Or it's possible it's just user error when I tried it before.

jolivier23: I guess I would argue that the existing pool_rf should be changed to be a minimum rf.  Is there an issue with doing that?

liw: @liuxuezhao, @gnailzenh, could you comment on what our plan is for making the RF-related UI (i.e., the pool `rd_fac` property, the container `rd_fac` property, and the `DAOS_POOL_RF` workaround) more consistent, please?

jolivier23:
I changed the 

> > > > > > I think you're missing https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/server_utils_params.py#L437
> > > > > 
> > > > > 
> > > > > I left that one on purpose to test that it still works
> > > > 
> > > > 
> > > > The problem is the `REQUIRED_ENV_VARS` list ensures that every server config file will specify `DAOS_POOL_RF=4` unless the test's yaml defines a different value for `DAOS_POOL_RF`, so when `src/tests/ftest/control/dmg_pool_query_ranks.yaml` is modified to specify `DAOS_EVICTION_THRESHOLD` the code will generate a server config file with both `DAOS_POOL_RF=4` and `DAOS_EVICTION_THRESHOLD=1` defined.
> > > 
> > > 
> > > This is actually another good thing to test too. We want DAOS_EVICTION_THRESHOLD to take precedent when both are specified. That should be fine to have both. It will see that the first one is set and not even check DAOS_POOL_RF which is the desired behavior.
> > 
> > 
> > So really the only issue is that for all other tests, we will have a deprecation warning in the logs but I assume that's not checked anywhere
> 
> I'm fairly certain we don't check for deprecation warnings in the logs.
> 
> While I agree it's a good idea to have tests to verify that DAOS_EVICTION_THRESHOLD takes precedent over DAOS_POOL_RF in the server config and verifying that DAOS_POOL_RF still works if specified by itself (and probably verifying the deprecation warning as well), I feel this testing should be its own isolated test for a few reasons:
> 
> 1. If someone down the line removes support for DAOS_POOL_RF, only one test would fail - not all of them
> 2. For a test not validating the DAOS_EVICTION_THRESHOLD/DAOS_POOL_RF entry it adds confusion if the two are used interchangeably - especially where one is always specified, but overridden by a different variable name.

Ok, I changed the default one

jolivier23:
Reviewers: Do you want me to back out the name change parts and just commit the bit that allows us to set it to a lower value?  This is something we need.  I can always commit it just to our branch but I prefer to go upstream first and that is the most important part of this change.  Otherwise, if you're OK with the name change, I'd appreciate some reviews if you get a chance amidst the churn of company transitions

docs/admin/env_variables.md
mchaarawi: i did attempt a PR to rename it after it was first introduced:
https://github.com/daos-stack/daos/pull/15090

but I believe we decided since this is all changing sometime to just have per property setting (i.e. this env will be removed), we don't need to rename. @liuxuezhao can correct me if my understanding is wrong.

I do agree with Li Wei that evict is probably not the right name.

PR #15544, opened 2024-11-27
DAOS-16837 container: Add client-side DFS metrics

mjmac:
If metrics are enabled for a POSIX container, create
a new pool/$UUID/container/$UUID/dfs metrics root in
the client telemetry to provide DFS-oriented metrics
(POSIX ops, file I/Os, etc).

Also fixes a bug in the agent code for pruning unused
client telemetry segments.

Features: container telemetry

### DIFF
--- a/src/client/dfs/io.c
+++ b/src/client/dfs/io.c
@@ -15,7 +15,20 @@
 
 #include "dfs_internal.h"
 
+static void
+dfs_update_file_metrics(dfs_t *dfs, daos_size_t read_bytes, daos_size_t write_bytes)
+{
+	if (dfs == NULL || dfs->metrics == NULL)
+		return;
+
+	if (read_bytes > 0)
+		d_tm_inc_gauge(dfs->metrics->dm_read_bytes, read_bytes);
+	if (write_bytes > 0)
+		d_tm_inc_gauge(dfs->metrics->dm_write_bytes, write_bytes);
+}
+
 struct dfs_read_params {
+	dfs_t           *dfs;
 	daos_size_t     *read_size;
 	daos_array_iod_t arr_iod;
 	daos_range_t     rg;
@@ -35,6 +48,8 @@ read_cb(tse_task_t *task, void *data)
 		D_GOTO(out, rc);
 	}
 
+	DFS_OP_STAT_INCR(params->dfs, DOS_READ);
+	dfs_update_file_metrics(params->dfs, params->arr_iod.arr_nr_read, 0);
 	*params->read_size = params->arr_iod.arr_nr_read;
 out:
 	D_FREE(params);
@@ -61,6 +76,7 @@ dfs_read_int(dfs_t *dfs, dfs_obj_t *obj, daos_off_t off, dfs_iod_t *iod, d_sg_li
 	if (params == NULL)
 		D_GOTO(err_task, rc = -DER_NOMEM);
 
+	params->dfs       = dfs;
 	params->read_size = read_size;
 
 	/** set array location */
@@ -90,6 +106,7 @@ dfs_read_int(dfs_t *dfs, dfs_obj_t *obj, daos_off_t off, dfs_iod_t *iod, d_sg_li
 	 * completion cb that frees params in this case, so we can just ignore the rc here.
 	 */
 	dc_task_schedule(task, true);
+
 	return 0;
 
 err_params:
@@ -125,6 +142,7 @@ dfs_read(dfs_t *dfs, dfs_obj_t *obj, d_sg_list_t *sgl, daos_off_t off, daos_size
 			daos_event_launch(ev);
 			daos_event_complete(ev, 0);
 		}
+		DFS_OP_STAT_INCR(dfs, DOS_READ);
 		return 0;
 	}
 
@@ -146,7 +164,9 @@ dfs_read(dfs_t *dfs, dfs_obj_t *obj, d_sg_list_t *sgl, daos_off_t off, daos_size
 			return daos_der2errno(rc);
 		}
 
+		DFS_OP_STAT_INCR(dfs, DOS_READ);
 		*read_size = iod.arr_nr_read;
+		dfs_update_file_metrics(dfs, iod.arr_nr_read, 0);
 		return 0;
 	}
 
@@ -173,6 +193,7 @@ dfs_readx(dfs_t *dfs, dfs_obj_t *obj, dfs_iod_t *iod, d_sg_list_t *sgl, daos_siz
 			daos_event_launch(ev);
 			daos_event_complete(ev, 0);
 		}
+		DFS_OP_STAT_INCR(dfs, DOS_READ);
 		return 0;
 	}
 
@@ -189,7 +210,9 @@ dfs_readx(dfs_t *dfs, dfs_obj_t *obj, dfs_iod_t *iod, d_sg_list_t *sgl, daos_siz
 			return daos_der2errno(rc);
 		}
 
+		DFS_OP_STAT_INCR(dfs, DOS_READ);
 		*read_size = arr_iod.arr_nr_read;
+		dfs_update_file_metrics(dfs, arr_iod.arr_nr_read, 0);
 		return 0;
 	}
 
@@ -223,6 +246,7 @@ dfs_write(dfs_t *dfs, dfs_obj_t *obj, d_sg_list_t *sgl, daos_off_t off, daos_eve
 			daos_event_launch(ev);
 			daos_event_complete(ev, 0);
 		}
+		DFS_OP_STAT_INCR(dfs, DOS_WRITE);
 		return 0;
 	}
 
@@ -238,8 +262,12 @@ dfs_write(dfs_t *dfs, dfs_obj_t *obj, d_sg_list_t *sgl, daos_off_t off, daos_eve
 		daos_event_errno_rc(ev);
 
 	rc = daos_array_write(obj->oh, DAOS_TX_NONE, &iod, sgl, ev);
-	if (rc)
+	if (rc == 0) {
+		DFS_OP_STAT_INCR(dfs, DOS_WRITE);
+		dfs_update_file_metrics(dfs, 0, buf_size);
+	} else {
 		D_ERROR("daos_array_write() failed, " DF_RC "\n", DP_RC(rc));
+	}
 
 	return daos_der2errno(rc);
 }
@@ -248,6 +276,8 @@ int
 dfs_writex(dfs_t *dfs, dfs_obj_t *obj, dfs_iod_t *iod, d_sg_list_t *sgl, daos_event_t *ev)
 {
 	daos_array_iod_t arr_iod;
+	daos_size_t      buf_size;
+	int              i;
 	int              rc;
 
 	if (dfs == NULL || !dfs->mounted)
@@ -266,6 +296,7 @@ dfs_writex(dfs_t *dfs, dfs_obj_t *obj, dfs_iod_t *iod, d_sg_list_t *sgl, daos_ev
 			daos_event_launch(ev);
 			daos_event_complete(ev, 0);
 		}
+		DFS_OP_STAT_INCR(dfs, DOS_WRITE);
 		return 0;
 	}
 
@@ -276,9 +307,18 @@ dfs_writex(dfs_t *dfs, dfs_obj_t *obj, dfs_iod_t *iod, d_sg_list_t *sgl, daos_ev
 	if (ev)
 		daos_event_errno_rc(ev);
 
+	buf_size = 0;
+	if (dfs->metrics != NULL && sgl != NULL)
+		for (i = 0; i < sgl->sg_nr; i++)
+			buf_size += sgl->sg_iovs[i].iov_len;
+
 	rc = daos_array_write(obj->oh, DAOS_TX_NONE, &arr_iod, sgl, ev);
-	if (rc)
+	if (rc == 0) {
+		DFS_OP_STAT_INCR(dfs, DOS_WRITE);
+		dfs_update_file_metrics(dfs, 0, buf_size);
+	} else {
 		D_ERROR("daos_array_write() failed (%d)\n", rc);
+	}
 
 	return daos_der2errno(rc);
 }
--- a/src/client/dfs/metrics.c
+++ b/src/client/dfs/metrics.c
@@ -14,6 +14,7 @@
 #include <daos/common.h>
 #include <daos/container.h>
 #include <daos/metrics.h>
+#include <daos/pool.h>
 #include <daos/job.h>
 #include <gurt/telemetry_common.h>
 #include <gurt/telemetry_producer.h>
@@ -28,11 +29,12 @@
 #define FILE_METRICS_SIZE (((D_TM_METRIC_SIZE * NR_SIZE_BUCKETS) * 2) + D_TM_METRIC_SIZE * 2)
 #define DFS_METRICS_SIZE  (STAT_METRICS_SIZE + FILE_METRICS_SIZE)
 
-#define SPRINTF_CONT_PATH(buf, cont_uuid, path)                                                    \
-	snprintf(buf, sizeof(buf), "container/" DF_UUIDF "/%s", DP_UUID(cont_uuid), path);
+#define SPRINTF_TM_PATH(buf, pool_uuid, cont_uuid, path)                                           \
+	snprintf(buf, sizeof(buf), "pool/" DF_UUIDF "/container/" DF_UUIDF "/%s",                  \
+		 DP_UUID(pool_uuid), DP_UUID(cont_uuid), path);
 
 #define ADD_STAT_METRIC(name, ...)                                                                 \
-	SPRINTF_CONT_PATH(tmp_path, cont_uuid, DFS_METRICS_ROOT "/ops/" #name);                    \
+	SPRINTF_TM_PATH(tmp_path, pool_uuid, cont_uuid, DFS_METRICS_ROOT "/ops/" #name);           \
 	rc = d_tm_add_metric(&metrics->dm_op_stats[i], D_TM_COUNTER, "Count of " #name " calls",   \
 			     "calls", tmp_path);                                                   \
 	if (rc != 0) {                                                                             \
@@ -42,7 +44,7 @@
 	i++;
 
 static void
-op_stats_init(struct dfs_metrics *metrics, uuid_t cont_uuid)
+op_stats_init(struct dfs_metrics *metrics, uuid_t pool_uuid, uuid_t cont_uuid)
 {
 	char tmp_path[D_TM_MAX_NAME_LEN] = {0};
 	int  i                           = 0;
@@ -55,7 +57,7 @@ op_stats_init(struct dfs_metrics *metrics, uuid_t cont_uuid)
 }
 
 static void
-cont_stats_init(struct dfs_metrics *metrics, uuid_t cont_uuid)
+cont_stats_init(struct dfs_metrics *metrics, uuid_t pool_uuid, uuid_t cont_uuid)
 {
 	char tmp_path[D_TM_MAX_NAME_LEN] = {0};
 	int  rc                          = 0;
@@ -63,7 +65,7 @@ cont_stats_init(struct dfs_metrics *metrics, uuid_t cont_uuid)
 	if (metrics == NULL)
 		return;
 
-	SPRINTF_CONT_PATH(tmp_path, cont_uuid, "mount_time");
+	SPRINTF_TM_PATH(tmp_path, pool_uuid, cont_uuid, "mount_time");
 	rc = d_tm_add_metric(&metrics->dm_mount_time, D_TM_TIMESTAMP, "container mount time", NULL,
 			     tmp_path);
 	if (rc != 0)
@@ -71,7 +73,7 @@ cont_stats_init(struct dfs_metrics *metrics, uuid_t cont_uuid)
 }
 
 static void
-file_stats_init(struct dfs_metrics *metrics, uuid_t cont_uuid)
+file_stats_init(struct dfs_metrics *metrics, uuid_t pool_uuid, uuid_t cont_uuid)
 {
 	char tmp_path[D_TM_MAX_NAME_LEN] = {0};
 	int  rc                          = 0;
@@ -79,7 +81,7 @@ file_stats_init(struct dfs_metrics *metrics, uuid_t cont_uuid)
 	if (metrics == NULL)
 		return;
 
-	SPRINTF_CONT_PATH(tmp_path, cont_uuid, DFS_METRICS_ROOT "/read_bytes");
+	SPRINTF_TM_PATH(tmp_path, pool_uuid, cont_uuid, DFS_METRICS_ROOT "/read_bytes");
 	rc = d_tm_add_metric(&metrics->dm_read_bytes, D_TM_STATS_GAUGE, "dfs read bytes", "bytes",
 			     tmp_path);
 	if (rc != 0)
@@ -89,7 +91,7 @@ file_stats_init(struct dfs_metrics *metrics, uuid_t cont_uuid)
 	if (rc)
 		DL_ERROR(rc, "Failed to init dfs read size histogram");
 
-	SPRINTF_CONT_PATH(tmp_path, cont_uuid, DFS_METRICS_ROOT "/write_bytes");
+	SPRINTF_TM_PATH(tmp_path, pool_uuid, cont_uuid, DFS_METRICS_ROOT "/write_bytes");
 	rc = d_tm_add_metric(&metrics->dm_write_bytes, D_TM_STATS_GAUGE, "dfs write bytes", "bytes",
 			     tmp_path);
 	if (rc != 0)
@@ -110,6 +112,7 @@ dfs_metrics_enabled()
 void
 dfs_metrics_init(dfs_t *dfs)
 {
+	uuid_t pool_uuid;
 	uuid_t cont_uuid;
 	char   root_name[D_TM_MAX_NAME_LEN];
 	pid_t  pid       = getpid();
@@ -118,6 +121,12 @@ dfs_metrics_init(dfs_t *dfs)
 
 	if (dfs == NULL)
 		return;
+
+	rc = dc_pool_hdl2uuid(dfs->poh, NULL, &pool_uuid);
+	if (rc != 0) {
+		DL_ERROR(rc, "failed to get pool UUID");
+		goto error;
+	}
 
 	rc = dc_cont_hdl2uuid(dfs->coh, NULL, &cont_uuid);
 	if (rc != 0) {
@@ -139,16 +148,16 @@ dfs_metrics_init(dfs_t *dfs)
 		goto error;
 	}
 
-	SPRINTF_CONT_PATH(root_name, cont_uuid, DFS_METRICS_ROOT);
+	SPRINTF_TM_PATH(root_name, pool_uuid, cont_uuid, DFS_METRICS_ROOT);
 	rc = d_tm_add_ephemeral_dir(NULL, DFS_METRICS_SIZE, root_name);
 	if (rc != 0) {
 		DL_ERROR(rc, "failed to add DFS metrics dir");
 		goto error;
 	}
 
-	cont_stats_init(dfs->metrics, cont_uuid);
-	op_stats_init(dfs->metrics, cont_uuid);
-	file_stats_init(dfs->metrics, cont_uuid);
+	cont_stats_init(dfs->metrics, pool_uuid, cont_uuid);
+	op_stats_init(dfs->metrics, pool_uuid, cont_uuid);
+	file_stats_init(dfs->metrics, pool_uuid, cont_uuid);
 
 	d_tm_record_timestamp(dfs->metrics->dm_mount_time);
 	return;
--- a/src/client/dfs/metrics.c
+++ b/src/client/dfs/metrics.c
@@ -1,0 +1,165 @@
+/**
+ * (C) Copyright 2024 Intel Corporation.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause-Patent
+ */
+#define D_LOGFAC DD_FAC(dfs)
+
+#include <uuid/uuid.h>
+#include <fcntl.h>
+
+#include <daos.h>
+#include <daos_fs.h>
+#include <daos_fs_sys.h>
+#include <daos/common.h>
+#include <daos/container.h>
+#include <daos/metrics.h>
+#include <daos/job.h>
+#include <gurt/telemetry_common.h>
+#include <gurt/telemetry_producer.h>
+#include <gurt/telemetry_consumer.h>
+
+#include "metrics.h"
+#include "dfs_internal.h"
+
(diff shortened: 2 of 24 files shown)


src/client/dfs/obj.c
```diff
@@ -412,6 +412,12 @@ open_stat(dfs_t *dfs, dfs_obj_t *parent, const char *name, mode_t mode, int flag
 
 out:
 	if (rc == 0) {
+		if (flags & O_CREAT) {
+			DFS_OP_STAT_INCR(dfs, DOS_CREATE);
+		} else {
+			DFS_OP_STAT_INCR(dfs, DOS_OPEN);
+		}
```
mjmac, line 419: @mchaarawi: I will rework this to increment the counters in the relevant `open_` helpers. Not sure what I was thinking here, tbh...

mjmac:
@mchaarawi: I've removed the dependency on the new container property for now, as I'd prefer not to hold this patch up if that becomes a sticking point. Please take a look at let me know if there are any significant changes that you'd like. I can add some ftest coverage for the new DFS metrics if it looks OK.

@ashleypittman: I based this patch on the stats you added to dfuse. It's a bit heavier-weight because it relies on the telemetry library instead of atomics, but maybe there's an opportunity to merge the approaches?

src/client/dfs/io.c
```diff
@@ -85,11 +100,13 @@ dfs_read_int(dfs_t *dfs, dfs_obj_t *obj, daos_off_t off, dfs_iod_t *iod, d_sg_li
 	if (rc)
 		D_GOTO(err_params, rc);
 
+	DFS_OP_STAT_INCR(dfs, DOS_READ);
```
mchaarawi, line 103: is this counting the number of read operations?
should you then move it the completion cb too after the op succeeds?
for other operations like mkdir, etc you only increment after the op succeeds not if the op fails.

src/client/dfs/io.c
```diff
@@ -146,7 +164,9 @@ dfs_read(dfs_t *dfs, dfs_obj_t *obj, d_sg_list_t *sgl, daos_off_t off, daos_size
 			return daos_der2errno(rc);
 		}
 
+		DFS_OP_STAT_INCR(dfs, DOS_READ);
```
mchaarawi, line 167: so for instance here you increment the nmber of read ops only on success, just FYI.

src/client/dfs/obj.c
```diff
@@ -786,6 +790,8 @@ dfs_stat(dfs_t *dfs, dfs_obj_t *parent, const char *name, struct stat *stbuf)
 		oh = parent->oh;
 	}
 
+	DFS_OP_STAT_INCR(dfs, DOS_STAT);
```
mchaarawi, line 793: it sounds like we are incrementing the stat counter twice? once here and once in entry_stat

src/client/dfs/obj.c
```diff
@@ -810,6 +816,9 @@ dfs_ostat(dfs_t *dfs, dfs_obj_t *obj, struct stat *stbuf)
 		D_GOTO(out, rc);
 
 out:
+	if (rc == 0)
+		DFS_OP_STAT_INCR(dfs, DOS_OSTAT);
```
mchaarawi, line 820: same here

src/client/dfs/obj.c
```diff
@@ -1013,6 +1022,9 @@ statx_task(tse_task_t *task)
 err1_out:
 	D_FREE(op_args);
 	daos_obj_close(args->parent_oh, NULL);
+
+	if (rc == 0)
+		DFS_OP_STAT_INCR(args->dfs, DOS_OSTATX);
```
mchaarawi, line 1027: i would not make a statx, just stat is fine in this case. statx is just an async version of the API

mchaarawi (changes requested):
just some minor issues that need to be addressed

src/client/dfs/obj.c
mjmac: So that comment would probably apply to the other -X calls? I was wondering about that... On the one hand, I can see the simplicity of consolidating them. On the other, I wonder if it's useful to see when a client is using the async versions (performance improvements?). I'll defer to your guidance on that.

mchaarawi: yea i think users care less about the type of the call vs the actual metrics of what is being done. so yes i agree consolidating them with removing the X would be better and would avoid confusion for users between 2 different but really the same metric

src/client/dfs/io.c
mjmac: Oops. I must have screwed up a rebase. I had moved this into the callback. Sigh. I'll go back through and check it all again more carefully.

src/client/dfs/obj.c
mjmac: Hmm, yep. Removed this one. The increment only happens in entry_stat() if it succeeded.

src/client/dfs/obj.c
mjmac: Ah, I see what you mean. I'll remove the OSTAT metrics.

src/client/dfs/obj.c
mjmac: Done, thanks.

src/client/dfs/io.c
```diff
@@ -276,9 +307,18 @@ dfs_writex(dfs_t *dfs, dfs_obj_t *obj, dfs_iod_t *iod, d_sg_list_t *sgl, daos_ev
 	if (ev)
 		daos_event_errno_rc(ev);
 
+	buf_size = 0;
+	if (sgl)
+		for (i = 0; i < sgl->sg_nr; i++)
+			buf_size += sgl->sg_iovs[i].iov_len;
+
```
mchaarawi, line 314: I would rather you add a condition around this if metrics are enabled only.

mchaarawi (dismissed):
looks good to me.
It would be great if we can enable metrics for something like our existing ior_small.py and mdtest_small.py ci tests and do some verifications there. whether in this PR or other PR is fine with me.

src/client/dfs/io.c
mjmac: Done.

daltonbohning:
I suspect the copyright GHA is failing because the workflow is coming from a merge of this PR + master, but the source tree used is just this PR.. Something I'll need to consider in the future for new GHA. Anyway, the copyright GHA is not required so can be ignored

mjmac:
> looks good to me. It would be great if we can enable metrics for something like our existing ior_small.py and mdtest_small.py ci tests and do some verifications there. whether in this PR or other PR is fine with me.

I added a simple test to verify that the DFS metrics are incremented as expected after an IOR run. We can add more extensive testing as we get more experience with the feature.

src/tests/ftest/telemetry/dfs_client_telemetry.py
```diff
@@ -0,0 +1,66 @@
...
+            'client_dfs_write_bytes'
+        ]
+
+        self.log_step('Reading dfs telemetry')
+        after_metrics = self.telemetry.collect_client_data(val_metric_names + bkt_metric_names)
+        for metric in val_metric_names:
+            msum = 0
+            for value in after_metrics[metric].values():
+                msum += value
+            self.assertGreater(msum, 0)
```
phender, line 59: It would be useful to have a description to go with the potential raised exception:
```suggestion
            self.assertGreater(msum, 0, f'{metric} value not greater than zero after I/O')
```
Do we want to verify all metric values before reporting an error?

src/tests/ftest/telemetry/dfs_client_telemetry.py
```diff
@@ -0,0 +1,66 @@
...
+        for metric in bkt_metric_names:
+            msum = 0
+            for hist in after_metrics[metric].values():
+                msum += hist['sample_sum']
+            self.assertGreater(msum, 0)
```
phender, line 64: Same thing here:
```suggestion
            self.assertGreater(msum, 0, f'{msum} sample_sum not greater than zero after I/O')
```

src/tests/ftest/telemetry/dfs_client_telemetry.yaml
```diff
@@ -0,0 +1,46 @@
...
+  telemetry_port: 9191
+  telemetry_retain: 30s
+  telemetry_enabled: true
+
+pool:
+  scm_size: 2G
+
+container:
+  type: POSIX
+  control_method: daos
```
phender, line 30: FYI, this is now the default.

src/tests/ftest/util/telemetry_utils.py
```diff
@@ -1217,13 +1248,21 @@ def _get_data(self, names, info):
...
-                    data[name][label_key] = metric['value']
+                    if 'value' in metric:
+                        data[name][label_key] = metric['value']
+                    else:
+                        data[name][label_key] = {
+                            'sample_count': metric['sample_count'],
+                            'sample_sum': metric['sample_sum'],
+                            'buckets': metric['buckets'],
+                        }
         return data
```
phender, line 1266: Maybe we should handle it this way:
```suggestion
                for metric in metrics[name]['metrics']:
                    if 'labels' not in metric:
                        continue
                    labels = [f'host:{host}']
                    for key, value in metric['labels'].items():
                        labels.append(":".join([str(key), str(value)]))
                    label_key = ",".join(sorted(labels))
                    if 'value' in metric:
                        data[name][label_key] = metric['value']
                    elif 'sample_count' in metric and 'sample_sum' in metric and 'buckets' in metric:
                        data[name][label_key] = {
                            'sample_count': metric['sample_count'],
                            'sample_sum': metric['sample_sum'],
                            'buckets': metric['buckets'],
                        }
                    else:
                        continue
        return data
```
Or will there always be an accompanying `value` entry for each `buckets` entry?

phender (changes requested):
@mjmac @daltonbohning thoughts on failing the test at the first non-expected metric value vs. reporting all non-expected values and then raising a single error? As described in DAOS-15542 (so I could handle this later), the `TelemetryUtils` class has `collect_data()`, `display_data()`, and `verify_data()` methods for reporting and verifying all the telemetry values.  It would probably require a slight modification to the `MetricData` class to support displaying/verifying `sample_sum` instead of `value`.

src/tests/ftest/telemetry/dfs_client_telemetry.py
```diff
@@ -0,0 +1,66 @@
...
+            'client_dfs_read_bytes',
+            'client_dfs_write_bytes'
+        ]
+
+        self.log_step('Reading dfs telemetry')
+        after_metrics = self.telemetry.collect_client_data(val_metric_names + bkt_metric_names)
+        for metric in val_metric_names:
+            msum = 0
+            for value in after_metrics[metric].values():
+                msum += value
```
daltonbohning, line 58: FYI python has a built-in for this
```suggestion
            msum = sum(after_metrics[metric].values())
```

src/tests/ftest/util/telemetry_utils.py
daltonbohning: I disagree with the `else..continue` because that could silently mask errors. I think keeping the implicit assumption is better because it will raise an exception if something is unexpected

src/tests/ftest/telemetry/dfs_client_telemetry.py
daltonbohning: Side note: @phender I think we could enforce the description by overriding the function in our base class and making the argument non-optional. Then lint checks would complain about the missing argument - hopefully in the githooks before going through CI

src/tests/ftest/telemetry/dfs_client_telemetry.yaml
mjmac: The container type and control_method? Or just the control_method?

src/tests/ftest/util/telemetry_utils.py
mjmac: > I think keeping the implicit assumption is better because it will raise an exception if something is unexpected

Yeah, that's where I was going with this... The default/standard metric is the `value` type, and I've added a new `histogram` type to allow tests to have access to the histograms that are in the DFS metrics (and will probably be added elsewhere now that they're supported properly).

mjmac:
> > @mjmac @daltonbohning thoughts on failing the test at the first non-expected metric value vs. reporting all non-expected values and then raising a single error? As described in DAOS-15542 (so I could handle this later), the `TelemetryUtils` class has `collect_data()`, `display_data()`, and `verify_data()` methods for reporting and verifying all the telemetry values. It would probably require a slight modification to the `MetricData` class to support displaying/verifying `sample_sum` instead of `value`.
> 
> I think it would be better to handle that separately since this PR is primarily non-ftest. If it is simple to modify the existing function, then I think it's a good idea. But if the function becomes too complex then IMO it's overkill

I'd also prefer to tackle that separately. I think we could probably flesh out the telemetry collection utils a bit more, e.g. create `ValueMetric` and `HistMetric` types with validation, etc. It might also be interesting to add instrumentation to the performance-oriented tests to catch obvious regressions.

src/tests/ftest/telemetry/dfs_client_telemetry.py
mjmac: Done.

src/tests/ftest/telemetry/dfs_client_telemetry.py
```diff
@@ -0,0 +1,62 @@
...
+        After performing some I/O, the DFS-level metrics should look reasonable.
+
+        Test steps:
+        1) Create a pool and container
+        2) Perform some I/O with IOR
+        3) Verify that the DFS metrics are sane
+
+        :avocado: tags=all,daily_regression
+        :avocado: tags=vm
+        :avocado: tags=telemetry
```
daltonbohning, line 29: Sorry, didn't notice this before. We should add the `dfs` feature tag
```suggestion
        :avocado: tags=dfs,telemetry
```

mjmac: Done.

mchaarawi:
unfortunately there was a force-push along the way after my last +1 review and my review diff is not visible anymore. please avoid force-push after someone had reviewed.

mjmac:
> unfortunately there was a force-push along the way after my last +1 review and my review diff is not visible anymore. please avoid force-push after someone had reviewed.

Well, it was still in draft then, so I figured that I'd squash it all down before moving it out of draft and requesting final reviews. In any case, there haven't been any changes to the code outside of ftest since then, other than the check you asked me to add.

mchaarawi:
> > unfortunately there was a force-push along the way after my last +1 review and my review diff is not visible anymore. please avoid force-push after someone had reviewed.
> 
> Well, it was still in draft then, so I figured that I'd squash it all down before moving it out of draft and requesting final reviews. In any case, there haven't been any changes to the code outside of ftest since then, other than the check you asked me to add.

to be fair, you did ask for a review on this PR and i did not jump in to review this myself while it was in draft state ;-)
anyway thanks for the note on the C changes so i don't have to go through them again.

src/tests/ftest/telemetry/dfs_client_telemetry.py
```diff
@@ -0,0 +1,62 @@
...
+        ]
+
+        self.log_step('Reading dfs telemetry')
+        after_metrics = self.telemetry.collect_client_data(val_metric_names + bkt_metric_names)
+        for metric in val_metric_names:
+            msum = sum(after_metrics[metric].values())
+            self.assertGreater(msum, 0, f'{metric} value not greater than zero after I/O')
+        for metric in bkt_metric_names:
+            msum = sum([hist['sample_sum'] for hist in after_metrics[metric].values()])
+            self.assertGreater(msum, 0, f'{metric} sample_sum not greater than zero after I/O')
```
mchaarawi, line 60: sounds like the test just verifies that the metrics are > 0 ?
to me that sounds pretty useless test if we are not verifying what is the expected exact value from a deterministic workload.
but again I mentioned that the ftest changes can be done in a future PR. please create a Jira ticket for this so we can tack it.

phender (changes requested):
Need to address the `consider-using-generator` pylint error.

src/tests/ftest/telemetry/dfs_client_telemetry.py
```diff
@@ -0,0 +1,62 @@
...
+            'client_dfs_write_bytes'
+        ]
+
+        self.log_step('Reading dfs telemetry')
+        after_metrics = self.telemetry.collect_client_data(val_metric_names + bkt_metric_names)
+        for metric in val_metric_names:
+            msum = sum(after_metrics[metric].values())
+            self.assertGreater(msum, 0, f'{metric} value not greater than zero after I/O')
+        for metric in bkt_metric_names:
+            msum = sum([hist['sample_sum'] for hist in after_metrics[metric].values()])
```
daltonbohning, line 59: I think just don't put brackets, which would make this a generator instead of list
```suggestion
            msum = sum(hist['sample_sum'] for hist in after_metrics[metric].values())
```

src/tests/ftest/telemetry/dfs_client_telemetry.py
mjmac: I think that this is better than nothing, but I agree that it would be better to validate that the collected metrics are within expected tolerances. Created https://daosio.atlassian.net/browse/DAOS-16885 to capture this.

src/tests/ftest/telemetry/dfs_client_telemetry.py
daltonbohning: E.g. notice the difference in type here
```
>>> type(i for i in range(3))
<class 'generator'>
>>> type([i for i in range(3)])
<class 'list'>
```

mjmac: Ah, I see. Duh. I guess it's more efficient because the consumer (sum()) can do its thing as the values come in vs. waiting for the whole list to be built first and then thrown away.

daltonbohning: Right. It doesn't matter much in this case, but if you were using something like `any` where early exit is possible, and a super large "list", it would help.

src/tests/ftest/util/telemetry_utils.py
daltonbohning: I see what you mean now. Should we handle this as part of the other ftest updates?

src/client/dfs/metrics.c
```diff
@@ -0,0 +1,165 @@
...
+#include "dfs_internal.h"
+
+#define DFS_METRICS_ROOT  "dfs"
+
+#define STAT_METRICS_SIZE (D_TM_METRIC_SIZE * DOS_LIMIT)
+#define FILE_METRICS_SIZE (((D_TM_METRIC_SIZE * NR_SIZE_BUCKETS) * 2) + D_TM_METRIC_SIZE * 2)
+#define DFS_METRICS_SIZE  (STAT_METRICS_SIZE + FILE_METRICS_SIZE)
+
+#define SPRINTF_CONT_PATH(buf, cont_uuid, path)                                                    \
+	snprintf(buf, sizeof(buf), "container/" DF_UUIDF "/%s", DP_UUID(cont_uuid), path);
```
kjacque, line 32: Would it make sense to include the pool ID as part of this path?

src/tests/ftest/util/telemetry_utils.py
```diff
@@ -1217,13 +1248,21 @@ def _get_data(self, names, info):
...
                     label_key = ",".join(sorted(labels))
-                    data[name][label_key] = metric['value']
+                    if 'value' in metric:
+                        data[name][label_key] = metric['value']
+                    else:
+                        data[name][label_key] = {
+                            'sample_count': metric['sample_count'],
+                            'sample_sum': metric['sample_sum'],
+                            'buckets': metric['buckets'],
+                        }
```
phender, line 1265: This will cause problems with https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/telemetry_utils.py#L1247-L1250 where the `self._data[metric][combined_label]` (i.e. `data[name][label_key]`) value is now a dictionary.

Since the new src/tests/ftest/telemetry/dfs_client_telemetry.py test is not making use of the `self.telemetry.verify_data()` method it wouldn't be an immediate problem, but if future tests verify metrics with sample_count/sample_sum/buckets (or if the metrics for existing tests change to use sample_count/sample_sum/buckets instead of value) will have a problem.

kjacque (approved):
I didn't look as deeply into the DFS changes. Go and telemetry changes LGTM.

phender (changes requested):

src/tests/ftest/util/telemetry_utils.py
mjmac: OK, so what do you suggest as a solution?

src/client/dfs/metrics.c
mjmac: > Would it make sense to include the pool ID as part of this path?

... yes. Good catch. I had a concern about whether or not it would work depending on whether or not full client metrics (vs. DFS-only metrics) were enabled, but I tested and it works fine either way. Had to add a new helper to get the pool UUID from the handle.

src/tests/ftest/util/telemetry_utils.py
mjmac: @daltonbohning, @phender: Would you prefer that I just remove the ftest changes from this PR and we can come up with a more complete solution in DAOS-16885?

daltonbohning:
@mjmac Is this correct? Total ` "sample_count": 2,` but all 4 buckets - 1048575, 2097151, 4194303, +Inf - have count 2?
```
2024-12-15 02:49:18,255 process          L0416 DEBUG| [stdout]             "labels": {
2024-12-15 02:49:18,255 process          L0416 DEBUG| [stdout]               "container": "6981025d-695a-4839-ac13-1e64bd7db15c",
2024-12-15 02:49:18,255 process          L0416 DEBUG| [stdout]               "jobid": "wolf-233vm3.wolf.hpdd.intel.com-84216",
2024-12-15 02:49:18,255 process          L0416 DEBUG| [stdout]               "pid": "84216"
2024-12-15 02:49:18,255 process          L0416 DEBUG| [stdout]             },
2024-12-15 02:49:18,255 process          L0416 DEBUG| [stdout]             "sample_count": 2,
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]             "sample_sum": 1048576,
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]             "buckets": [
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]                 "upper_bound": 255
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,256 process          L0416 DEBUG| [stdout]                 "upper_bound": 511
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]                 "upper_bound": 1023
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]                 "upper_bound": 2047
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,257 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]                 "upper_bound": 4095
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]                 "upper_bound": 8191
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,258 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]                 "upper_bound": 16383
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]                 "upper_bound": 32767
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,259 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]                 "upper_bound": 65535
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]                 "upper_bound": 131071
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,260 process          L0416 DEBUG| [stdout]                 "upper_bound": 262143
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]                 "cumulative_count": 0,
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]                 "upper_bound": 524287
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]                 "cumulative_count": 2,
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]                 "upper_bound": 1048575
2024-12-15 02:49:18,261 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "cumulative_count": 2,
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "upper_bound": 2097151
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "cumulative_count": 2,
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "upper_bound": 4194303
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]               },
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]               {
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "label": "",
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "cumulative_count": 2,
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]                 "upper_bound": "+Inf"
2024-12-15 02:49:18,262 process          L0416 DEBUG| [stdout]               }
2024-12-15 02:49:18,263 process          L0416 DEBUG| [stdout]             ]
2024-12-15 02:49:18,263 process          L0416 DEBUG| [stdout]           },
```

src/tests/ftest/util/telemetry_utils.py
daltonbohning: IMO, assuming what you have now is functionally correct, let's keep it for basic code path coverage. Then can expand on it in DAOS-16885

daltonbohning:
> > @mjmac Is this correct? Total ` "sample_count": 2,` but all 4 buckets - 1048575, 2097151, 4194303, +Inf - have count 2?
> 
> Yep. It's a [Cumulative Histogram](https://en.wikipedia.org/wiki/Histogram#Cumulative_histogram). That's [the format](https://www.robustperception.io/why-are-prometheus-histograms-cumulative/) that Prometheus likes.

Interesting :) I don't think I've seen that before

mjmac:
@phender: Nudge... Please advise on how you want me to address your -1 review.

src/tests/ftest/util/telemetry_utils.py
phender: I have some ideas that could support histogram type metrics as part DAOS-15542.  Since it doesn't seem like it will cause an immediate problem its probably ok.

phender (approved):
Approving ftest changes.

kjacque (approved):

mchaarawi (approved):

PR #15545, opened 2024-11-29
DAOS-16838 control: Fix dmg storage query usage with emulated NVMe

tanabarr:
Fix a regression which prevents commands that scan storage from enumerating devices backed with emulated (AIO file or kdev) NVMe. 

Required-githooks: true

### Gatekeeper:


tanabarr:
NLT failing on unrelated dfuse valgrind issue

knard-intel (dismissed):
LGTM

src/control/server/instance_storage_rpc.go
```diff
@@ -340,12 +341,16 @@ func scanEngineBdevsOverDrpc(ctx context.Context, engine Engine, pbReq *ctlpb.Sc
...
-	for _, sd := range scanSmdResp.Devices {
+	for i, sd := range scanSmdResp.Devices {
 		if sd.Ctrlr == nil {
 			return nil, errors.Errorf("smd %q has no ctrlr ref", sd.Uuid)
 		}
 
 		addr := sd.Ctrlr.PciAddr
+		if addr == "" {
+			// Mock identifier for emulated NVMe mode where devices have no PCI-address.
+			addr = fmt.Sprintf("0000:00:0.%d", i)
```
kjacque, line 352: My only question is whether this could possibly conflict with a real device, since it is used in the `seenCtrlrs` map. I wonder if it would be better to use a "name" which could be either a PCI address for the real device, or the file name for emulated nvme.

kjacque (approved):
Overall looks good. The comment is more of a question/suggestion.

src/control/server/instance_storage_rpc.go
mjmac: In DPDK, `PCI_ANY_ID` is defined as `0xffff`: https://doc.dpdk.org/api/rte__pci_8h.html#a53aca768a081fcf56089353d805ab77c

That also seems to match the kernel: https://elixir.bootlin.com/linux/v6.12.1/source/include/linux/mod_devicetable.h#L18

Might be better to use that value, as there is zero possibility of conflicting with a real vendor ID that way?

tanabarr: so are you suggesting `FFFF:00:0.%d` as the format string?

mjmac: I think so, if the goal is to avoid any conflict with a device that might be found via the topology scan. I can't find an official PCI spec to cite, but from what I found about how linux and DPDK do things, that value is a "can't happen" value, i.e. there's no way that a valid device could show up with that vendor ID.

tanabarr: aren't we getting confused here between PCI-address and PCI-ID? https://wiki.debian.org/HowToIdentifyADevice/PCI

tanabarr: the first segment of the PCI address is the domain not the vendor ID afais

tanabarr: I've improved the resilience against collisions by using `0xffff` PCI-domain segment and `0xf` func value (neither would be used in a "real" NVMe device address as domain segment is either `0x10000` or `0x0000` in almost all the addresses and BDF func value is 0-7).

knard-intel (approved):

kjacque (approved):

PR #15548, opened 2024-12-02
DAOS-16840 control: Deprecate access_points in server config

mjmac:
The access_points name in the server configuration is
an ongoing source of confusion. Rename it to the more
descriptive mgmt_svc_replicas and emit deprecation
notices for older configurations.

Required-githooks: true

### DIFF
--- a/docs/admin/troubleshooting.md
+++ b/docs/admin/troubleshooting.md
@@ -313,7 +313,7 @@ sudo ipcrm -M 0x10242049
 1. Format the SCMs defined in the config file.
 1. Generate the config file using `dmg config generate`. The various requirements will be populated without a syntax error.
 1. Try starting with `allow_insecure: true`. This will rule out the credential certificate issue.
-1. Verify that the `access_points` host is accessible and the port is not used.
+1. Verify that the `mgmt_svc_replicas` host is accessible and the port is not used.
 1. Check the `provider` entry. See the "Network Scan and Configuration" section of the admin guide for determining the right provider to use.
 1. Check `fabric_iface` in `engines`. They should be available and enabled.
 1. Check that `socket_dir` is writable by the daos_server.
@@ -327,7 +327,7 @@ sudo ipcrm -M 0x10242049
 1. When the server configuration is changed, it's necessary to restart the agent.
 1. `DER_UNREACH(-1006)`: Check the socket ID consistency between PMem and NVMe. First, determine which socket you're using with `daos_server network scan -p all`. e.g., if the interface you're using in the engine section is eth0, find which NUMA Socket it belongs to. Next, determine the disks you can use with this socket by calling `daos_server nvme scan` or `dmg storage scan`. e.g., if eth0 belongs to NUMA Socket 0, use only the disks with 0 in the Socket ID column.
 1. Check the interface used in the server config (`fabric_iface`) also exists in the client and can communicate with the server.
-1. Check the access_points of the agent config points to the correct server host.
+1. Check the `access_points` of the agent config points to the correct server hosts.
 1. Call `daos pool query` and check that the pool exists and has free space.
 
 ### Applications run slow
@@ -512,7 +512,7 @@ fabric providers.
 
 After starting `daos_server`, ranks will be unable to join if their configuration's fabric provider
 does not match that of the system. The system configuration is determined by the management service
-(MS) leader node, which may be arbitrarily chosen from the configured access points.
+(MS) leader node, which may be arbitrarily chosen from the configured MS replicas.
 
 The error message will include the string: `fabric provider <provider1> does not match system provider <provider2>`
 
--- a/src/tests/ftest/util/agent_utils_params.py
+++ b/src/tests/ftest/util/agent_utils_params.py
@@ -72,6 +72,10 @@ class DaosAgentYamlParameters(YamlParameters):
         #        Enable client telemetry for all client processes.
         #   - telemetry_retain: <str>, e.g. 5m
         #        Time to retain per-client telemetry data.
+        #   - access_points: <list>, e.g.  ["hostname1:10001"]
+        #       Hosts can be specified with or without port, default port below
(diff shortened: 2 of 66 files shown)


daltonbohning (commented):
ftest LGTM but would be good for @phender to review too

kjacque (commented):
Overall looks good. This PR needs to be run through CI with `Features: control` though.

src/tests/ftest/util/agent_utils_params.py
```diff
@@ -72,6 +72,10 @@ def __init__(self, filename, common_yaml):
         #        Enable client telemetry for all client processes.
         #   - telemetry_retain: <str>, e.g. 5m
         #        Time to retain per-client telemetry data.
+        #   - access_points: <list>, e.g.  ["hostname1:10001"]
```
phender, line 75: Doesn't leaving `access_points` in the agent config and confusion when the same list is called `mgmt_svc_replicas` in the server and control configs, or can the `access_points` be a subset of the `mgmt_svc_replicas` list?

utils/config/examples/daos_server_mdonssd.yml
```diff
@@ -1,7 +1,7 @@
 # Example configuration file for Metadata on SSD.
 
 name: daos_server           # sys group daos_server
-access_points: ['example']  # management service leader (bootstrap)
+mgmt_svc_replicas: ['example']  # management service leader (bootstrap)
 # port: 10001               # control listen port, default 10001
```
phender, line 5: Seeing as this is an example, shouldn't we cleanup the comment string spacing?
```suggestion
name: daos_server               # sys group daos_server
mgmt_svc_replicas: ['example']  # management service leader (bootstrap)
# port: 10001                   # control listen port, default 10001
```
Same comment applies to other `utils/config/examples/daos_server_*.yml` files

phender (changes requested):
Minor cleanup of example configs.

src/tests/ftest/util/agent_utils_params.py
mjmac: The control config only has a hostlist. A long time ago (pre-1.0), there used to be separate hostlist and access_point lists. Now, the control config just has a hostlist including all of the servers for convenience with operations like storage scan and format.

One of the major drivers for this change is to codify that the `access_points` parameter on the agent does not have to be in sync with the value of `mgmt_svc_replicas` on the server. It just needs to have some server addresses in the list. Doesn't matter which ones. Having the same set as `mgmt_svc_replicas` is a slight optimization, but it will work even if you have none of the replicas in the list.

Down the road, we will likely design a more dynamic MS replica selection similar to how the PS works, and then it will be really clear that the `access_points` list is independent of which servers are the current replicas.

utils/config/examples/daos_server_mdonssd.yml
mjmac: Sure, I can do that... Given how big the PR was already, though, I was trying to minimize changes that aren't directly related to the rename.

docs/admin/troubleshooting.md
```diff
@@ -327,7 +327,7 @@ sudo ipcrm -M 0x10242049
 1. When the server configuration is changed, it's necessary to restart the agent.
 1. `DER_UNREACH(-1006)`: Check the socket ID consistency between PMem and NVMe. First, determine which socket you're using with `daos_server network scan -p all`. e.g., if the interface you're using in the engine section is eth0, find which NUMA Socket it belongs to. Next, determine the disks you can use with this socket by calling `daos_server nvme scan` or `dmg storage scan`. e.g., if eth0 belongs to NUMA Socket 0, use only the disks with 0 in the Socket ID column.
 1. Check the interface used in the server config (`fabric_iface`) also exists in the client and can communicate with the server.
-1. Check the access_points of the agent config points to the correct server host.
+1. Check the `access_points` of the agent config points to the correct server hosts.
```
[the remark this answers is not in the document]
mjmac: I don't think we want to put those implementation details in the user documentation. In practice, it really doesn't matter which server addresses go in the agent config. If anything, I'd be inclined to change the guidance to just use a hostlist that covers all of the servers.

utils/config/examples/daos_server_mdonssd.yml
mjmac: Done.

docs/admin/troubleshooting.md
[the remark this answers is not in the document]
mjmac: To expand a bit on this comment, as I've noted elsewhere, longer-term I think we want to move away from exposing a fixed set of server addresses as "special" to the agent. Keeping this somewhat vague now gives us flexibility to change the server side of things without requiring changes to the agent config process.

I've pushed the latest commit with `Features: control`. I was under the impression that the test harness was smart enough to automatically test all changed .py files now, but maybe not?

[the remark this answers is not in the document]
daltonbohning: It's only smart enough when you modify specific functional tests.
- modify specific functional test -> run that test
- modify ftest/util -> run `pr` - not smart!
- modify dfs -> run `dfs` - slightly smart
- anything else -> run `pr` - not smart!

The mapping is here
https://github.com/daos-stack/daos/blob/master/src/tests/ftest/tags.yaml

This hasn't gotten much attention since I originally added it (I haven't had time), but we could add something like this e.g. Or whatever you think makes sense for the code you regularly edit.
But keep in mind with this mapping, even if you modify a single line comment, all `control` tests will be included. So devs should address this for their feature areas.
```
# Run "control" for files under "src/control/"
# Run "control" and "security" for files under "src/control/security"
src/control/: control
src/control/security/: security
```

The way to override this behavior would be to use `Test-tag: <specific tests I want to run>`
`Test-tag` overrides whatever CI auto-determined.
`Features` is on top of whatever CI auto-determined.

phender (approved):

kjacque (approved):

PR #15550, opened 2024-12-02
DAOS-16822 client: Add mkdir_p variant to dfs_sys API (#15526)

mjmac:
As a usability enhancement, add a mkdir variant with
the following properties:
  * If the directory or any of its parents already exist
    and are directories, don't return an error
  * If any of the parent directories need to be created,
    automatically do so as a prerequisite step to creating
    the final child directory


mjmac:
@jolivier23: This will need a forced landing due to the NLT/ubuntu issue.

jolivier23 (approved):

PR #15584, opened 2024-12-11
DAOS-16477 pool: Rename Suspect state to Dead

mjmac:
Change the name to more closely reflect the underlying
SWIM status, and reduce user confusion. An engine that
has been marked DEAD by SWIM cannot participate in pool
services, and has most likely already SIGKILL-ed itself.

Features: pool
Required-githooks: true


src/include/daos_pool.h
```diff
@@ -162,8 +162,8 @@ enum daos_pool_info_bit {
 	DPI_ENGINES_ENABLED = 1ULL << 2,
 	/** true to include (in \a ranks) engines with some or all targets disabled (down). */
 	DPI_ENGINES_DISABLED = 1ULL << 3,
-	/** true to include (in \a ranks) suspect engines. */
-	DPI_ENGINES_SUSPECT = 1ULL << 4,
+	/** true to include (in \a ranks) engines marked DEAD by SWIM. */
```
liw, line 165: [Nit; not requesting changes] Maybe this degree of inaccuracy is acceptable, but I just want to clarify that these "dead" engines are actually the set of pool engines who are not "alive". In other words, they may also include engines absent from the SWIM membership.

liw (approved):

daltonbohning (commented):
ftest LGTM

src/include/daos_pool.h
tanabarr: if that is the case then maybe suspect is a more accurate term?

tanabarr (approved):

mjmac:
Test failure appears to be https://daosio.atlassian.net/issues/DAOS-16829, which was introduced with the PR that added the suspect state. I could re-run without `Features: pool`, but probably best to wait until the fix is merged and then re-run after merging.

src/include/daos_pool.h
mjmac: @liw: When would we have engines in the pool map that are absent from the SWIM membership, but aren't dead?

liw: When the MS has excluded the engines from the system (and the new system map has been distributed out), but the PS hasn't excluded the engines from the pool. In this state, from the pool's point of view, these engines are not strictly SWIM-DEAD, just unavailable---loosely speaking, "dead".

Hmm, personally I don't think "suspect" is more accurate. :)

mjmac:
I've merged master and re-pushed with the `Allow-unstable-test: true` pragma set.

mjmac:
Two known failures:
  * [DAOS-16874](https://daosio.atlassian.net/browse/DAOS-16874)
  * [DAOS-16829](https://daosio.atlassian.net/browse/DAOS-16829)

PR #14713, opened 2024-07-08
DAOS-16073 iosrv: Invalid ULT stack size with MMAP_STACK feature

knard-intel:
### Description

This PR is fixing the following issues:
- The stack size allocated with mmap()'ed ULT was lower than what it was requested.  Indeed, it was the asked size minus the size of the stack_descriptor.
- Access to the btree holding the free stack was racy.  Even if there is one btree per xstream, the xstream creating the ULT thread and the one calling the callback recycling the stack can eventually not be the same.  Indeed, this last callback is respectively called for a unnamed and named ULT by the xstream running the ULT thread and the xstream calling ABT_thread_free() function.
- using the GROW_DOWN is not safe as the size of the stack is not tracked. Thus the unmap() of a stack can eventually lead to some mapped pages leak: mmaped page automatically allocated by the kernel.

This PR is providing the following enhancement:
- Add a guard page at the end of the allocated stack.  This guard page allow to  detect stack overflow.  Without this guard page, the mmap()'ed stack will grow until it reach the high end of the next lower mapping.
- Enabling and disabling mmap()'ed stack ULT features dynamically with setting environment variables.
- Do not use wrapper for starting named thread.
- Add a stack overflow  management test.

The following features will be done in a follow-up PR #14782 :
- Take into account available physical ressources when mmap()'ed stack are created and recycled.
- Add a functional test

### Notes

The main goal of using such stack is to be able to easily find ULT overusing its stack and eventually corrupting the stack of another ULT.  It is also aimed for running the non regression test to check that no ULTs are overusing their stacks.

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [x] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [x] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [x] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate owners.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
[...]

PR #15613, opened 2024-12-13
DAOS-16877 client: implement utilities for shared memory

wiliamhuang:
Features: shm

1. use tlsf as memory allocator
2. shared memory create/destroy
3. robust mutex in shared memory
4. hash table in shared memory

Required-githooks: true
Skipped-githooks: codespell

### Gatekeeper:


wiliamhuang:
Some component (e.g., hash table record reference count) will be revised later when we add more use cases.

wiliamhuang:
Currently we use tlsf memory allocator. It will be replaced by our own allocator in future.

src/tests/ftest/daos_test/shm.yaml
```diff
@@ -0,0 +1,20 @@
...
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos
+  system_ram_reserved: 1
+pool:
+  scm_size: 1G
+container:
+  type: POSIX
+  control_method: daos
```
phender, line 20: This isn't used by the test.  A typical cmocka test would use the `pool` entry information, but only when the test is run via `DaosCoreBase.run_subtest()`:
- https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/daos_core_base.py#L68-L69
- https://github.com/daos-stack/daos/blob/master/src/tests/ftest/util/daos_core_base.py#L89-L90

phender (changes requested):

src/tests/ftest/daos_test/shm.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        daos_test = os.path.join(self.bin, 'shm_test')
+        cmocka_utils = CmockaUtils(
+            self.hostlist_clients, "shm", self.outputdir, self.test_dir, self.log)
+        daos_test_env = cmocka_utils.get_cmocka_env()
+        job = get_job_manager(self, "Clush", get_cmocka_command(daos_test))
+        job.assign_hosts(cmocka_utils.hosts)
+        job.assign_environment(daos_test_env)
+
+        cmocka_utils.run_cmocka_test(self, job)
```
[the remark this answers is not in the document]
wiliamhuang: @phender Thank you very much! I wrote shm.py and shm.yaml with dfuse.py and dfuse.yaml as templates. "shm_test" does not need to configure dmg command. Additional environment variables might be in added in futures tests with "daos_test_env" here.
Thank you for your tip of using "faults-enabled:  false". I will use it in future. I used "Features: shm" previously to run the new test.

src/tests/ftest/daos_test/shm.yaml
wiliamhuang: The new test "shm_test" does not need scm_size and nvme_size information.

daltonbohning: I think what @phender means is since this test does not use `DaosCoreBase.run_subtest()`, the entire `pool` and `container` keys here are not used and can be removed

daltonbohning: ... or if we use `DaosCoreBase.run_subtest()` we can keep it

wiliamhuang: @daltonbohning @phender Thank you very much! I will try removing pool and container keys locally to make sure it works. I thought they are required.

wiliamhuang: @daltonbohning @phender You are right. We can remove pool and container keys in yaml file as you suggested. I will update it in next commit. Thank you!

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+
+void
+shm_fini(void)
+{
+	if (!shm_inited())
+		return;
+
+	atomic_fetch_add_relaxed(&(d_shm_head->ref_count), -1);
+	if (pid != pid_shm_creator)
+		munmap(d_shm_head, d_shm_head->size);
```
knard38, line 294: From my understanding, all the process should `unmap` and then `close` the file.
The shared mmaped file content should be keept until the `unlink()` and `shm_unlink()` will be called.
Thus, not unmapping for the process creating the shared memory file seems to be useless.
I am also concerned that it could be seen as memoy leak by valgrind or other memory checker tools.

knard38 (commented):
I have not yet finish the review process, but I still have several concerns and questions regarding this PR.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+	D_ASSERT(rc == 0);
+
+	/* the shared memory only accessible for individual user for now */
+	sprintf(daos_shm_name_buf, "%s_%d", daos_shm_name, getuid());
+open_rw:
+	shm_ht_fd = shm_open(daos_shm_name_buf, O_RDWR, shmopen_perm);
+	/* failed to open */
+	if (shm_ht_fd == -1) {
+		if (errno == ENOENT) {
+			goto create_shm;
```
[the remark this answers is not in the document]
wiliamhuang: You are right. I will create a function for shm creation. Thank you!

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+				DS_ERROR(errno, "failed to open shared memory after %d retries",
+					 RETRY);
+				goto err;
+			}
+		}
+	}
+
+	/* map existing shared memory */
+	shm_addr = mmap(FIXED_SHM_ADDR, shm_size, PROT_READ | PROT_WRITE,
+			MAP_SHARED | MAP_FIXED, shm_ht_fd, 0);
```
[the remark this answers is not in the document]
wiliamhuang: I agree. Fixed memory location is a strong limitation. It comes from the memory allocator we use. We could eliminate this limitation later once we have our own memory allocator supporting shared memory management.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+		usleep(1);
+	if (((struct d_shm_hdr *)shm_addr)->size != shm_size) {
+		/* EBADRQC - Invalid request code */
+		errno = EBADRQC;
+		DS_ERROR(errno, "unexpected shared memory size. Multiple versions of daos or env?");
+		goto err_unmap;
+	}
+	atomic_fetch_add_relaxed(&(((struct d_shm_hdr *)shm_addr)->ref_count), 1);
+	d_shm_head = (struct d_shm_hdr *)shm_addr;
+	close(shm_ht_fd);
```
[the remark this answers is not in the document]
wiliamhuang: The fd is not needed any more after calling mmap() in my impression, so you can close fd and still access shared memory.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+	return 0;
+
+create_shm:
+	shm_ht_fd = shm_open(daos_shm_name_buf, O_RDWR | O_CREAT | O_EXCL, shmopen_perm);
```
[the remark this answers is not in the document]
wiliamhuang: You are right. I will fix it. Thank you!

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	if (shm_mutex_init(&(d_shm_head->g_lock)) != 0) {
+		DS_ERROR(errno, "shm_mutex_init() failed");
+		goto err_unmap;
+	}
+	for (i = 0; i < N_SHM_POOL; i++) {
+		if (shm_mutex_init(&(d_shm_head->mem_lock[i])) != 0) {
+			DS_ERROR(errno, "shm_mutex_init() failed");
+			goto err_unmap;
```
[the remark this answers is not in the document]
wiliamhuang: I think it is not needed to destroy mutexes in shared memory since shared memory is not set up successfully. They will not be used.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+		}
+	}
+	if (shm_mutex_init(&(d_shm_head->ht_lock)) != 0) {
+		DS_ERROR(errno, "shm_mutex_init() failed");
```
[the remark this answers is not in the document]
wiliamhuang: Same here. It is not needed to destroy mutexes in shared memory since shared memory is not set up successfully. They will not be used.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+shm_alloc_comm(size_t align, size_t size)
+{
+	int      idx_allocator;
+	void    *buf;
+	int      tid;
+	uint32_t hash;
+	uint64_t oldref;
+
+	if (idx_small < 0) {
+		tid  = syscall(SYS_gettid);
```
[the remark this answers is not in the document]
wiliamhuang: I never used "pthread_self" before.
https://man7.org/linux/man-pages/man3/pthread_self.3.html

**The thread ID returned by pthread_self() is not the same thing as
 the kernel thread ID returned by a call to gettid.**

This syscall will be called only once. I would not worry about the overhead.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+	shm_mutex_unlock(&(d_shm_head->mem_lock[idx_allocator]));
+}
+
+void
+shm_destroy(void)
+{
+	char daos_shm_file_name[128];
+
+	sprintf(daos_shm_file_name, "/dev/shm/%s_%d", daos_shm_name, getuid());
+	unlink(daos_shm_file_name);
```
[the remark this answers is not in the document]
wiliamhuang: You are right. I will replace unlink() with shm_unlink(). Thank you!

src/tests/ftest/daos_test/shm.py
```diff
@@ -0,0 +1,44 @@
+"""
```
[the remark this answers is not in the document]
daltonbohning: In general, yes. If the same test can be ran as a unit test (low cost, quick) instead of a functional test (higher cost, slower) then we should run it as a unit test

src/gurt/shm_alloc.c
wiliamhuang: Considering the cache in kernel space, I thought we may want to keep our cache persistent too. Otherwise, shared memory needs to be initialized again and again. Ideally, the space for caching would be freed after the content expires. We need our own shared memory allocator to dynamically expand/shrink shared memory region. It would be a long way to get there.
Yes. I did have a little concern about whether valgrind can detect the memory leak in shared memory usage here. I could play with valgrind to find out with simple test.

src/tests/ftest/daos_test/shm.py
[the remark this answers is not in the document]
wiliamhuang: @knard38 @daltonbohning Thank you very much! I will look into running this test as a unit test.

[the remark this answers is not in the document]
wiliamhuang: Yes. Current tests are simple. They could fit as a unit test. More complex tests will be added in future. We can add shared memory related ftest later when we need.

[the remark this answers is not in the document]
wiliamhuang: > @wiliamhuang , If it can help, I recently added some unit tests in the following PR: https://github.com/daos-stack/daos/pull/14713/files#diff-294ea4ccb7880cabe2a9a4ffadd3c709916da304a39a819c82f23bfe06197a61

@knard38 Thank you very much! It's very helpful.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	rc = d_getenv_uint64_t("DAOS_SHM_SIZE", &shm_size);
+	if (rc != -DER_NONEXIST) {
+		/* set parameter from env */
+		shm_pool_size = shm_size / N_SHM_POOL;
+		if (shm_pool_size % 4096)
+			/* make shm_pool_size 4K aligned */
+			shm_pool_size += (4096 - (shm_pool_size % 4096));
+		shm_size = shm_pool_size * N_SHM_POOL + sizeof(struct d_shm_hdr);
```
knard38, line 75: In fact my remarks on aligning the `struct d_shm_addr` was to have a sizeof struct which will be a multiple of the `PAGE_SIZE` otherwise `shm_size` could be not a multiple of `PAGE_SIZE` from my understanding. Then the shared pools (following the header), could be not aligned on PAGE_SIZE (from my understanding).
However, I have not check that the align attribute will properly change the sizeof. I will check this asap.

src/tests/suite/SConscript
```diff
@@ -13,6 +13,15 @@ def scons():
...
 
+    shm_test_env = base_env.Clone()
+    shm_test_env.compiler_setup()
+    shm_test_env.AppendUnique(LIBPATH=[Dir('../../gurt')])
+    shm_test_env.AppendUnique(LIBPATH=[Dir('../../common')])
+    shm_test_env.AppendUnique(LIBPATH=[Dir('../../cart')])
+    shmtest = shm_test_env.d_program(File("shm_test.c"), LIBS=['gurt', 'daos_common', 'cart',
+                                     'cmocka', 'rt', 'pthread'])
+    denv.Install('$PREFIX/bin/', shmtest)
+
```
mchaarawi, line 24: i don't think you need a DAOS server to run those tests, right?
so probably adding those as unit tests in gurt will be more appropriate

src/include/gurt/shm_alloc.h
```diff
@@ -0,0 +1,134 @@
+/**
+ * (C) Copyright 2024 Intel Corporation.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause-Patent
+ */
+
+#ifndef __DAOS_SHM_ALLOC_H__
+#define __DAOS_SHM_ALLOC_H__
+
+#include <stdint.h>
```
mchaarawi, line 10: it's probably better to create 1 header for this rather than multiple public headers that users of this module need to always include.

mchaarawi (commented):
just some quick comments.
still need to review more closely. will do that soon

src/gurt/shm_alloc.c
wiliamhuang: "shm_size" does not have be a multiple of PAGE_SIZE. mmap() does not require size to be a multiple of PAGE_SIZE. The memory allocator will use some space in the pool too. I am not sure making shm_size a multiple of PAGE_SIZE will bring noticeable benefit. Maybe some performance tests could help to clarify later.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,296 @@
...
+	rc = pthread_mutexattr_settype(&d_shm_mutex_attr, PTHREAD_MUTEX_NORMAL);
+	D_ASSERT(rc == 0);
+	rc = pthread_mutexattr_setpshared(&d_shm_mutex_attr, PTHREAD_PROCESS_SHARED);
+	D_ASSERT(rc == 0);
+	pthread_mutexattr_setrobust(&d_shm_mutex_attr, PTHREAD_MUTEX_ROBUST);
+	D_ASSERT(rc == 0);
+
+	/* the shared memory only accessible for individual user for now */
+	sprintf(daos_shm_name_buf, "%s_%d", daos_shm_name, getuid());
+open_rw:
```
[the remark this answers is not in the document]
wiliamhuang: I used "O_CREAT | O_EXCL" to make sure only one process will initialize shared memory. Without "O_EXCL" would allow more than one concurrent processes to initialize shared memory.

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
wiliamhuang: You are right.
The requirement of using fix same address across processes is due to memory allocator we use. This is a quick and dirty way to allow us to use existing memory allocator for now. In future we need to implement our own memory allocator to natively support shared memory management, then the limitation could be removed.

src/tests/suite/SConscript
wiliamhuang: Right. Current tests are quite simple. DAOS server is not needed. I will change the tests as unit tests. We will add ftest later when necessary. Thank you!

src/include/gurt/shm_alloc.h
wiliamhuang: Thank you very much! Just fixed it. Now only necessary APIs are exposed in this header file.

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
knard38: Fair enough

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
knard38: Fair enough

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
knard38: Not fully agree with you on this, this function should be called for each memory allocation and thus its performance have some impact.  However, this point is not a blocker for me.

src/gurt/shm_alloc.c
knard38: Not sure that it will change something for the kernel cache to keep a dangling pointer when the process will die.
From my understanding what will make the difference is to unlink the file.
@mchaarawi do you have an opinion on this ?

src/gurt/shm_alloc.c
knard38: From what I know it is indeed always better to have aligned memory for performance.
Moreover, from what I understand, the size really allocated by `mmap()` will be the same.
The only difference is that the padding will be added at the end of allocated memory instead of being between the `struct d_shm_hdr` and the first shm_pool.
In any case, this is not a blocker from my side.

src/include/gurt/shm_internal.h
knard38, line 1: From what I understand of the best DAOS practice, the internal header should not be located in the `src/include directory` to not be visible by the end user.  But I could be wrong on this point.

src/include/gurt/shm_internal.h
```diff
@@ -0,0 +1,72 @@
...
+/* Head of shared memory region */
+struct d_shm_hdr {
+	/* magic not equal DSM_MAGIC means shared memory is not initialized yet */
+	int              magic;
+	d_shm_mutex_t    g_lock;
+	/* the count of how many processes are mapping the shared memory region */
+	_Atomic int      ref_count;
+	/* global counter used for round robin picking memory allocator for large memory request */
+	_Atomic uint64_t large_mem_count;
+	/* array of pointors to memory allocators */
```
knard38, line 48: NIT
```suggestion
	/* array of pointers to memory allocators */
```

src/gurt/shm_dict.c
knard38, line 1: Instead of creating a new hash map, could it not be possible to update the current DAOS htable implementation with your shared memory new features ?

wiliamhuang: I agree. I inclined to take current DAOS htable implementation and modify it to fit shared memory at the beginning. I decided to implement hash table in shared memory from scratch after I realized many parts are not compatible.

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
wiliamhuang: I tested pthread_self() and syscall(SYS_gettid).

Hello from thread 1! tid_pthread_self = 140335788001024 SYS_gettid = 3023160
Hello from thread 2! tid_pthread_self = 140335779608320 SYS_gettid = 3023161
Hello from thread 3! tid_pthread_self = 140335771215616 SYS_gettid = 3023162
Hello from thread 4! tid_pthread_self = 140335762822912 SYS_gettid = 3023163

It looks like all tid_pthread_self are even number here. They are not evenly distributed to me. syscall(SYS_gettid) is only called once per thread instead of every memory allocation.

src/gurt/shm_alloc.c
wiliamhuang: One extreme case, the caching does not have any benefit at all if a user runs job serially in case we destruct caching once application ends.

src/include/gurt/shm_internal.h
wiliamhuang: ok. I will move it to src/gurt. Thank you!

src/include/gurt/shm_internal.h
wiliamhuang: A good catch. Thank you! Will fix it.

src/gurt/shm_dict.c
knard38: Fair enough, I will thus have a more in depth look to this file.

src/gurt/shm_alloc.c
knard38: Maybe, I am missing something, but if you remove the shared file then keeping a dangling pointer will not help if you have unlink the file.  On the other hand, if you do not unlink the file, the cache will still be available even if you do not have a dangling pointer at the end your applications.
However, it is perfectly that I am missing something.

wiliamhuang: shm_destroy() is not called by regular applications. It was called only in the shm_test. The file associated with shared memory will not be unlinked.
I talked to Mohamad recently. He suggested daos_agent will initialize and destruct shared memory region. We will update it later.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,331 @@
...
+			return errno;
+		DS_ERROR(errno, "shm_open() failed to create shared memory");
+		return errno;
+	}
+	/* set the size of shared memory region */
+	if (ftruncate(shm_ht_fd, shm_size) != 0) {
+		DS_ERROR(errno, "ftruncate() failed for shm_ht_fd");
+		goto err;
+	}
+	/* map the shared memory at a fixed address for now. We will remove this limit later. */
```
mchaarawi, line 75: can we add some more details for this in the comment as to why FIXED is needed for the allocator.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,331 @@
...
+	/* failed to open */
+	if (shm_ht_fd == -1) {
+		if (errno == ENOENT) {
+			rc = create_shm_region(shm_size, shm_pool_size);
+			if (rc == EEXIST)
+				goto open_rw;
+			else
+				return rc;
+		} else {
+			DS_ERROR(errno, "unexpected error shm_open()");
```
mchaarawi, line 183: im curious why the need for retry in this case?
also this should be a warn not an error in this case.

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
mchaarawi: i am wondering if we should block on this before landing. im worried about this limitation and what issues can arise because of that.

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
mchaarawi: if the descriptor is not needed after the mmap call, there is no need to keep it open. it's best close it to avoid leaks.

src/gurt/shm_alloc.c
mchaarawi: not sure i quite understood the discussion. the cache needs to be persistent beyond one process lifetime and can be destroyed when the agent is killed.
but as i understand each procecss needs to call shm_open(), mmap(), close(), munmap() on the shared memory region, and only the agent calls shm_unlink().
the agent part can be done later.

src/gurt/shm_dict.c
mchaarawi: i did attempt to do that myself too and got the same conclusion as Lei before.

src/gurt/shm_alloc.c
wiliamhuang: ok. I will add more details as suggested. Thank you!

src/gurt/shm_alloc.c
wiliamhuang: I thought retry might make the code more robust.
I just went through all possible errno returned by shm_open(). Retry should be not needed. I will change DS_ERROR to DS_WARN. Thank you!

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
wiliamhuang: I agree that this fixed mapping address limit should be removed before landing to master. Maybe ok for feature branch? This would allow us to work on caching while we implement our own memory allocator.

src/gurt/shm_alloc.c
wiliamhuang: The retry code was removed. DS_ERROR is kept to log the error of shm_open().

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
wiliamhuang: Fixed as suggested. Thank you!

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
wiliamhuang: This is fixed. Thank you!

src/include/gurt/shm_internal.h
wiliamhuang: Fixed as suggested. Thank you!

src/include/gurt/shm_internal.h
wiliamhuang: This is fixed. Thank you!

src/gurt/shm_dict.c
```diff
@@ -0,0 +1,430 @@
...
+
+	/* This hash table does not exist, then create it. */
+	*ht_head = shm_alloc(sizeof(struct d_shm_ht_head) + (sizeof(d_shm_mutex_t) * n_lock) +
+			     (sizeof(long int) * n_bucket));
+	if (*ht_head == NULL)
+		return ENOMEM;
+	ht_head_loc = *ht_head;
+
+	memcpy(ht_head_loc->ht_name, name, len_name + 1);
+	ht_head_loc->n_bucket = n_bucket;
```
knard38, line 77: At least we should check if `n_bucket` is greater or equal than `n_lock`.
We could also check if `n_bucket % n_lock == 0` as it will ensure an even number of buckets per lock and also not need to use float to find the lock index.

src/gurt/shm_dict.c
```diff
@@ -0,0 +1,430 @@
...
+	long int          *p_off_list;
+	struct shm_ht_rec *rec;
+	struct shm_ht_rec *rec_prev = NULL;
+	struct shm_ht_rec *rec_next = NULL;
+
+	hash       = d_hash_string_u32(key, ksize);
+	idx        = hash & (ht_head->n_bucket - 1);
+	idx_lock   = (unsigned int)(idx * ht_head->n_lock * 1.0f / ht_head->n_bucket);
+	p_ht_lock  = (d_shm_mutex_t *)((char *)ht_head + sizeof(struct d_shm_ht_head));
+	p_off_list = (long int *)((char *)p_ht_lock + sizeof(d_shm_mutex_t) * ht_head->n_lock);
```
knard38, line 117: seems to have concurrency issue with `shm_ht_destroy()` function.
The same apply for most of the other functions working on records and `get_shm_ht_with_name()` functions.

src/gurt/shm_alloc.c
wiliamhuang: Thank you!
To be clear, line 209,
"close(shm_ht_fd);"
It only frees the file descriptor.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,323 @@
...
+	pid_shm_creator         = pid;
+	d_shm_head->off_ht_head = INVALID_OFFSET;
+
+	atomic_store_relaxed(&(d_shm_head->ref_count), 1);
+	atomic_store_relaxed(&(d_shm_head->large_mem_count), 0);
+	d_shm_head->size          = shm_size;
+	d_shm_head->shm_pool_size = shm_pool_size;
+	d_shm_head->magic         = DSM_MAGIC;
+	/* initialization is finished now. */
+	return 0;
```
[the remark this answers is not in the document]
wiliamhuang: Thank you! Now fd is closed here. After creating shared memory region, it returns immediately instead of calling munmap and mmap.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	pid_shm_creator         = pid;
+	d_shm_head->off_ht_head = INVALID_OFFSET;
+
+	atomic_store_relaxed(&(d_shm_head->ref_count), 1);
+	atomic_store_relaxed(&(d_shm_head->large_mem_count), 0);
+	d_shm_head->size          = shm_size;
+	d_shm_head->shm_pool_size = shm_pool_size;
+	d_shm_head->magic         = DSM_MAGIC;
```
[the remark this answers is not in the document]
wiliamhuang: Thank you! This is fixed as you suggested.

src/gurt/shm_dict.c
```diff
@@ -0,0 +1,430 @@
...
+	}
+
+	if (*ht_head) {
+		/* found existing hash table */
+		shm_mutex_unlock(&(d_shm_head->ht_lock));
+		return 0;
+	}
+
+	/* This hash table does not exist, then create it. */
+	*ht_head = shm_alloc(sizeof(struct d_shm_ht_head) + (sizeof(d_shm_mutex_t) * n_lock) +
```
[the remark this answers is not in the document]
wiliamhuang: Now error (EINVAL) is returned if hash table exists but with different parameters. Thank you!

src/gurt/shm_dict.c
wiliamhuang: Thank you very much! I just made some improvements about possible issue of accessing hash table head.  I am working on mitigating the concurrency of hash table record. I will push my commit soon.

src/gurt/shm_alloc.c
wiliamhuang: Fixed mapping address is not needed with updated tlsf memory alllocator.

src/gurt/shm_dict.c
wiliamhuang: This is fixed as suggested. Thank you!

src/gurt/shm_dict.c
wiliamhuang: Could you please review again? Thank you very much!
Now the memory of hash table head and records are not freed as long as the reference count is not zero.

knard38 (commented):
The overall PR is OK for me.  The only blocker from my side is to check if the maximal value of the htable reference counter is acceptable for our use case.

src/gurt/shm_alloc.c
```diff
@@ -0,0 +1,327 @@
...
+	 * implement a memory allocator supporting shared memory natively and remove this limit
+	 * later.
+	 */
+	shm_addr = mmap(NULL, shm_size, PROT_READ | PROT_WRITE, MAP_SHARED, shm_ht_fd, 0);
+	if (shm_addr == MAP_FAILED) {
+		DS_ERROR(errno, "mmap() failed");
+		goto err;
+	}
+	d_shm_head        = (struct d_shm_hdr *)shm_addr;
+	d_shm_head->magic = 0;
```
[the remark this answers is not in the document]
wiliamhuang: Now memset() is used to zero the whole mapped region as you suggested. Thank you!

src/gurt/shm_dict.c
```diff
@@ -0,0 +1,803 @@
...
+	int64_t old_nref_htid, old_nref, saved_ht_id, *addr_old_nref_htid;
+	int64_t new_nref_htid, ht_id;
+
+	saved_ht_id        = ht_loc->ht_id;
+	addr_old_nref_htid = (int64_t *)&(ht_loc->ht_head->nref_htid);
+
+	if (!force) {
+		/* invalidate htid only if reference count is 0 */
+		old_nref_htid = saved_ht_id;
+		new_nref_htid = 0;
```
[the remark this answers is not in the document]
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_dict.c
```diff
@@ -0,0 +1,803 @@
...
+	old_nref_htid =
+	    atomic_load_explicit((_Atomic int64_t *)addr_old_nref_htid, memory_order_relaxed);
+	ht_id = GET_HTID(old_nref_htid);
+	if (saved_ht_id != ht_id)
+		/* hash table is not valid any more */
+		return SHM_HT_INVALID_HT;
+
+	old_nref = GET_NREF(old_nref_htid);
+	/* sanity check */
+	if (old_nref <= 0 && change < 0) {
```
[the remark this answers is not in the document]
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_dict.c
```diff
@@ -0,0 +1,803 @@
...
+		} else {
+			/* failed to exchange */
+			ht_id = GET_HTID(old_nref_htid);
+			if (saved_ht_id != ht_id) {
+				/* hash table is not valid any more */
+				return SHM_HT_INVALID_HT;
+			} else {
+				/* hash table is still valid, but nref was updated by other user */
+				old_nref = GET_NREF(old_nref_htid);
+				if (old_nref <= 0 && change < 0) {
```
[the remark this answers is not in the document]
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_alloc.c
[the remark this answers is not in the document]
knard38: In fact, it was an error from my side as I was expecting `mmap` to fill the mapped memory with zero.
After more investigation , it appears that it is done only with the `MAP_ANONYMOUS` flag which is not compatible with using a file descriptor.
At the end, using `memset()` or your assignment is OK for me.
However, to avoid some race conditions I will switch the lines 85 and 86:
```c
	memset(shm_addr, 0, shm_size);
	d_shm_head = (struct d_shm_hdr *)shm_addr;
```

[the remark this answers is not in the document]
wiliamhuang: Thank you very much! These lines are reordered.

src/gurt/shm_dict.c
```diff
@@ -0,0 +1,803 @@
...
+#include <errno.h>
+
+#include "shm_internal.h"
+#include <gurt/shm_dict.h>
+
+/* the address of shared memory region */
+extern struct d_shm_hdr *d_shm_head;
+
+#define INVALID_HT_ID (0)
+#define NREF_MASK     (0xFFFF000000000000L)
```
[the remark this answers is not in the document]
wiliamhuang: Now the lower 40 bits are allocated for ht id and 24 bits are for reference count.

knard38 (approved):
LGTM

mchaarawi (approved):
looks good to me overall. 1 thing i still like to see done, perhaps in a follow on PR, is move all the shm headers in gurt to 1 directory to not have 4 shm* headers in the gurt main dir. so i would suggest either:
1) 1 header in gurt
2) 1 dir in gurt called shm/; with all the headers in there (similar for the utilities implementation in the .c files).

wiliamhuang:
> looks good to me overall. 1 thing i still like to see done, perhaps in a follow on PR, is move all the shm headers in gurt to 1 directory to not have 4 shm* headers in the gurt main dir. so i would suggest either:
> 
> 1. 1 header in gurt
> 2. 1 dir in gurt called shm/; with all the headers in there (similar for the utiliti

Thank you! I will do this in the robust readers-writer lock PR.

mchaarawi:
even though this is to a feature branch, the PR must be green (all linting and builds, tests passing).

mchaarawi:
> even though this is to a feature branch, the PR must be green (all linting and builds, tests passing).

we know there is an issue with GHA arm builds since there are no runners, but others should be pass i believe.

mjmac (commented):
Is there any plan to unify this work with the shm management in the telemetry library? https://github.com/daos-stack/daos/blob/master/src/gurt/telemetry.c

Not suggesting that should be a hard requirement, but it seems like it would be a miss to wind up with two completely separate designs and implementations in the codebase.

cc @kjacque

mchaarawi:
> > > even though this is to a feature branch, the PR must be green (all linting and builds, tests passing).
> > 
> > 
> > we know there is an issue with GHA arm builds since there are no runners, but others should be pass i believe.
> 
> codespell check: the issues reported are not from this PR. copyright check: The major code of src/gurt/shm_tlsf.c and src/include/gurt/shm_tlsf.h are from https://github.com/mattconte/tlsf, so I did not add the HPE copyright. Clang format check: Most of them seems not necessary.

ok for the first 2. Clang is required on all PRs, so unfortunately you have to fix it.
note that if you think a code block looks bad with clang formatting you can add the clang format off / on around the code block to disable clang checks. but it has to pass on the PR.

wiliamhuang:
> ok for the first 2. Clang is required on all PRs, so unfortunately you have to fix it. note that if you think a code block looks bad with clang formatting you can add the clang format off / on around the code block to disable clang checks. but it has to pass on the PR.

ok. I will fix the clang format issues. Thank you!

wiliamhuang:
Pushed two commits to fix clang format issues. Now clang format is green.

johannlombardi (approved):
lgtm

PR #15616, opened 2024-12-13
DAOS-16829 client: Fix daos pool query regression

mjmac:
The patch landed for DAOS-16477 introduced a regression
in the pool query functionality of the daos tool. This
patch fixes the regression and introduces a v2 of the
pool query API that will get both enabled and disabled
ranks in one call.

Features: pool
Required-githooks: true


jolivier23 (approved):

src/include/daos_pool.h
```diff
@@ -341,6 +341,52 @@ int
...
+ * \return		These values will be returned by \a ev::ev_error in
+ *			non-blocking mode:
+ *			0		Success
+ *			-DER_INVAL	Invalid parameter
+ *			-DER_UNREACH	Network is unreachable
+ *			-DER_NO_HDL	Invalid pool handle
+ */
+int
+daos_pool_query_v2(daos_handle_t poh, d_rank_list_t **enabled_ranks, d_rank_list_t **disabled_ranks,
+		   daos_pool_info_t *info, daos_prop_t *pool_prop, daos_event_t *ev);
```
mchaarawi, line 388: I wonder if we can have a better name for this.
daos_pool_query_ranks() ?

mchaarawi (changes requested):

src/include/daos_pool.h
mjmac: I was thinking that we could deprecate the original pool query, similar to how the pool connect API was converted to accept a UUID or label, and that this would replace it eventually. But if that's not practical I'm happy to rename it.

mchaarawi: pool query is VERY common and almost no one is using the ranks field (passing NULL) and most just want the info field. so deprecating the API just for the ranks sake would be pretty extensive change for all users and all our middleware to make.

mjmac: OK, sounds like `daos_pool_query_ranks()` makes a lot more sense, then. I'll update it, thanks.

src/include/daos_pool.h
```diff
@@ -341,6 +341,52 @@ int
...
+ *			The caller is responsible for freeing the list with d_rank_list_free().
+ * \param[out]	disabled_ranks
+ * 			Optional, returned disabled pool storage engine ranks.
+ *			If #info is NULL, this list will be populated with the ranks of
+ * 			all engines with any targets disabled. If #info is not NULL, this
+ * 			list will be populated if #pi_bits has DPI_ENGINES_DISABLED set.
+ *			The caller is responsible for freeing the list with d_rank_list_free().
+ * \param[in,out]
+ *		info	Optional, returned pool information,
+ *			see daos_pool_info_bit.
```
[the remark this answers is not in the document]
mjmac: Seems like it would be less surprising to keep the same logic as the v1 query, IMO.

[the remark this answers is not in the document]
mchaarawi: I thought now that you have a separate array for enabled and disabled the check is just if the array is NULL or not whether to return those or not.  did i misunderstand?

src/include/daos_task.h
```diff
@@ -206,8 +206,10 @@ typedef struct {
 typedef struct {
 	/** Pool open handle. */
 	daos_handle_t		poh;
-	/** Optional, returned storage ranks in this pool. */
-	d_rank_list_t	      **ranks;
+	/** Optional, returned enabled storage ranks in this pool. */
+	d_rank_list_t         **enabled_ranks;
+	/** Optional, returned disabled storage ranks in this pool. */
+	d_rank_list_t         **disabled_ranks;
```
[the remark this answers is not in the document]
mjmac: Ugh. I'm thinking that I may just abandon this approach, actually... It looked like all the uses of this struct were internal to the client, so I thought it was safe to modify. Creating a new struct means that we can't share the query processing callback between these functions.

PR #15621, opened 2024-12-16
Revert "DAOS-16783 client: move daos_hhash_fini ahead"

mchaarawi:
Reverts daos-stack/daos#15452


wangdi1 (changes requested):
Please do not revert this. otherwise it will cause segfault if client side metrics is enabled. @mchaarawi sorry, I did not see your comments on the other ticket. let's discuss there.

PR #15624, opened 2024-12-16
DAOS-16886 control: Update deprecated code

mjmac:
The ioutil package has been deprecated since Go 1.16
and may be removed in the future. Update the code to
use the standard interfaces.

Also fix a linter warning that is now a compile error
in Go 1.24+ (non-constant format string in call
to *f functions).

Features: control
Required-githooks: true


kjacque (approved):
Nice cleanup, thanks.

tanabarr (approved):

PR #15632, opened 2024-12-17
DAOS-16881 control: Fix daos_server scm prep for single missing ns

tanabarr:
When single socket is missing a pmem namespace on dual-socket host a
confusing no-space error can be returned from daos_server scm prepare.
The previously required workaround is to specify --socket. Fix this
issue by adding NumaNode in fall-back case where ndctl region idset
overflow requires matching of numa/socket via ipmctl region info
instead. Also add unit test cases to cover the situation.

Features: control
Required-githooks: true

### Gatekeeper:


src/control/server/storage/scm/ipmctl.go
```diff
@@ -201,27 +201,24 @@ func (cr *cmdRunner) handleFreeCapacity(sockSelector int, nrNsPerSock uint, regi
 		return nil, nil, errors.Wrap(err, "createNamespaces")
 	}
 
-	numaSelector := sockAny
 	switch len(numaIDs) {
 	case 0:
-		return nil, nil, errors.New("no numa nodes were processed")
+		return nil, nil, errors.New("no namespaces created on regions with free capacity")
 	case 1:
```
knard-intel, line 207: Nit: not sure it still needed.
If yes, the switch could probably be replace by a simple `if len(numaIDs) == 0`

knard-intel, line 207: Nit: not sure the case 1 still needed.
If yes, the switch could probably be replace by a simple `if len(numaIDs) == 0`

kjacque: I think `case 1` is still needed, although it confused me for a minute, too. Maybe clearer:

```go
if len(numaIDs) == 0 {
    return return nil, nil, errors.New("no namespaces created on regions with free capacity")
}

if len(numaIDs) > 1 && sockSelector != sockAny {
    return nil, nil, errors.Errorf("unexpected number of numa nodes processed, want 1 got %d", len(numaIDs))
}
```

kjacque (commented):
Otherwise looks OK, comment is small but I think important for readability.

src/control/server/storage/scm/ipmctl.go
tanabarr: I'm not clear why an empty switch statement is problematic, it is just allowing a drop through much like a break, I'm requesting reviews again because I don't see the need to change it.

kjacque: I won't block on it. It is technically correct. I just thought it might be worth changing if it confused two of us on the first read.

Personally I find it harder to parse and understand greater-than and less-than checks when using `switch` in this way, as opposed to either phrasing the cases as conditionals (e.g. `case len(numaIDs) > 1`), or using if-statements. I effectively had to translate the switch cases into conditionals for my own understanding.

kjacque (dismissed):

src/control/server/storage/scm/ipmctl.go
tanabarr: fixed

knard-intel (approved):

kjacque (approved):

PR #15638, opened 2024-12-18
DAOS-16829 client: Fix daos pool query regression

mjmac:
The patch landed for DAOS-16477 introduced a regression
in the pool query functionality of the daos tool. This
patch simplifies the pool query logic to retrieve
enabled and disabled ranks. Dead ranks are not available
to the client without pool query protocol changes.

Features: daos_cmd
Required-githooks: true


src/control/cmd/daos/pool.go
```diff
@@ -295,21 +295,36 @@ func convertPoolInfo(pinfo *C.daos_pool_info_t) (*daos.PoolInfo, error) {
...
+		C.d_rank_list_free(disabledRanks)
+	}()
 
+	var rc C.int
 	cPoolInfo := C.daos_pool_info_t{
 		pi_bits: C.uint64_t(queryMask),
 	}
+	if queryMask.HasOption(daos.PoolQueryOptionEnabledEngines) && queryMask.HasOption(daos.PoolQueryOptionDisabledEngines) {
+		enaQm := queryMask
+		enaQm.ClearOptions(daos.PoolQueryOptionDisabledEngines)
```
tanabarr, line 312: how does this work? `queryMask` type `daos.PoolQueryMask` is alias of  `C.uint64_t`. is that somehow a reference? will clearing options here somehow update `cPoolInfo.pi_bits`, I just don't see how? thanks in advance

docs/admin/troubleshooting.md
```diff
@@ -556,7 +556,7 @@ Alternately, the administrator may erase and re-format the DAOS system to start
...
+Engines may become unavailable due to server power losses and reboots, network switch failures, etc. After staying unavailable for a certain period of time, these engines may become "excluded" or "errored" in `dmg system query` output. Once the states of all engines stabilize (see [`CRT_EVENT_DELAY`](env_variables.md)), each pool will check whether there is enough redundancy (see [Pool RF](pool_operations.md#pool-redundancy-factor)) to tolerate the unavailability of the "excluded" or "errored" engines. If there is enough redundancy, these engines will be excluded from the pool ("Disabled ranks" in `dmg pool query --health-only` output); otherwise, the pool will perform no exclusion ("Dead ranks" in `dmg pool query --health-only` output as described in [Querying a Pool](pool_operations.md#querying-a-pool)) and may become temporarily unavailable (as seen by timeouts of `dmg pool query`, `dmg pool list`, etc.). Similarly, when engines become available, whenever the states of all engines stabilize, each pool will perform the aforementioned check for any unavailable engines that remain.
```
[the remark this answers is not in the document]
mjmac: This is just a fix to s/suspect/Dead/ in existing documentation. I believe @liw was the original author. If there are specific suggestions please make them, or even better, submit them in a new PR. :)

src/control/cmd/daos/pool.go
mjmac: D'oh, good catch. I rushed this last night after abandoning the other approach. You're right; this won't do what I want it to do. I need to update the `pi_bits` field with the new mask.

mjmac: The funny thing is, it worked OK as it is in testing, but that's probably by accident. Oh, well. Definitely best to update it so that we're not relying on implicit behavior.

wangshilong (approved):

tanabarr (changes requested):
don't we need a unit test to verify the need for this change?

mjmac:
> don't we need a unit test to verify the need for this change?

There is a functional test, which is currently failing on master and will continue to fail until this lands. Writing unit tests for cgo calls in the daos tool is close to impossible with the current code organization.

tanabarr (approved):
approving because unit tests are to be covered in a future piece of work

knard-intel (commented):
A little bit late, but seems good to me :-)

PR #15640, opened 2024-12-18
DAOS-16837 container: Add client-side DFS metrics (#15544)

mjmac:
If metrics are enabled for a POSIX container, create
a new pool/$UUID/container/$UUID/dfs metrics root in
the client telemetry to provide DFS-oriented metrics
(POSIX ops, file I/Os, etc).

Also fixes a bug in the agent code for pruning unused
client telemetry segments.

Required-githooks: true

### DIFF
--- a/src/client/dfs/metrics.c
+++ b/src/client/dfs/metrics.c
@@ -1,0 +1,174 @@
+/**
+ * (C) Copyright 2024 Intel Corporation.
+ *
+ * SPDX-License-Identifier: BSD-2-Clause-Patent
+ */
+#define D_LOGFAC DD_FAC(dfs)
+
+#include <uuid/uuid.h>
+#include <fcntl.h>
+
+#include <daos.h>
+#include <daos_fs.h>
+#include <daos_fs_sys.h>
+#include <daos/common.h>
+#include <daos/container.h>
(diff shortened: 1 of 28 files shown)


jolivier23 (approved):

src/client/dfs/metrics.c
```diff
@@ -0,0 +1,174 @@
...
+		goto error;
+	}
+
+	snprintf(root_name, sizeof(root_name), "%d", pid);
+	/* if only container-level metrics are enabled; this will init a root for them */
+	rc = d_tm_init_with_name(d_tm_cli_pid_key(pid), root_size, D_TM_OPEN_OR_CREATE, root_name);
+	if (rc != 0 && rc != -DER_ALREADY) {
+		DL_ERROR(rc, "failed to init DFS metrics");
+		goto error;
+	}
```
wangdi1, line 143: will all threads share the same root here? I wonder if all there are a lot dfuse service threads, will it reach to the root_size maximum? or it will create a share memory for each pid?

mjmac: They all share the same segment. I think this is OK, since we don't care about tracking per-thread metrics.

PR #15646, opened 2024-12-19
DAOS-16896 common: Spill Over Evictable Buckets Implementation

sherintg:
The DAV_v2 allocator now includes support for Spill Over Evictable Buckets (SOEMB). All global allocations will continue to utilize the standard non-evictable memory buckets, while spillover allocations from evictable memory buckets will be directed to SOEMB. In the current implementation, SOEMB remains locked in the memory cache, similar to the behavior of non-evictable memory buckets.

### Gatekeeper:

### DIFF
--- a/src/common/dav_v2/heap.c
+++ b/src/common/dav_v2/heap.c
@@ -63,6 +63,8 @@ struct mbrt {
 	struct recycler      *recyclers[MAX_ALLOCATION_CLASSES];
 	bool                  laf[MAX_ALLOCATION_CLASSES]; /* last allocation failed? */
 	bool                  laf_updated;
+	bool                  is_global_mbrt;
+	bool                  is_evictable;
 };
 
 enum mb_usage_hint {
@@ -78,7 +80,24 @@ enum mb_usage_hint {
 #define MB_U30         (ZONE_MAX_SIZE * 3 / 10)
 #define MB_USAGE_DELTA (ZONE_MAX_SIZE / 20)
 
-size_t mb_usage_byhint[MB_UMAX_HINT] = {0, MB_U30 + 1, MB_U75 + 1, MB_U90 + 1};
+size_t mb_usage_byhint[MB_UMAX_HINT] = {1, MB_U30 + 1, MB_U75 + 1, MB_U90 + 1};
+
+struct mbrt_qbs {
+	struct mbrt_q mb_u90;
+	struct mbrt_q mb_u75;
+	struct mbrt_q mb_u30;
+	struct mbrt_q mb_u0;
+	struct mbrt_q mb_ue;
+};
+
+#define SOEMB_ACTIVE_CNT 3
+
+struct soemb_rt {
+	struct mbrt    *svec[SOEMB_ACTIVE_CNT];
+	int             cur_idx;
+	int             fur_idx;
+	struct mbrt_qbs qbs;
+};
 
 struct heap_rt {
 	struct alloc_class_collection *alloc_classes;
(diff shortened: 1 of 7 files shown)


src/common/dav_v2/tx.c
```diff
@@ -559,6 +559,7 @@ dav_tx_begin_v2(dav_obj_t *pop, jmp_buf env, ...)
 			sizeof(struct tx_range_def));
 		tx->first_snapshot = 1;
 		tx->pop = pop;
+		heap_soemb_reserve(pop->do_heap);
```
NiuYawei, line 562: This might needs be done before the umem_cache_reserve() call in the future?

When we support turning SOE bucket to evict-able (once a SOE bucket isn't qualified as a SOE anymore), allocator might need to pass the SOE set to umem_cache_reserve(), so that we can ensure that all SOE buckets are loaded in umem_cache_reserve().

src/common/dav_v2/heap.c
```diff
@@ -600,6 +772,85 @@ heap_mbrt_mb_reclaim_garbage(struct palloc_heap *heap, uint32_t zid)
...
+		if (mb)
+			mbrt_qbs_insertmb(&smbrt->qbs, mb);
+
+		for (i = 1; i < SOEMB_ACTIVE_CNT; i++) {
+			smbrt->svec[i - 1] = smbrt->svec[i];
+		}
+
+		smbrt->svec[SOEMB_ACTIVE_CNT - 1] = NULL;
+		smbrt->fur_idx                    = 0;
+	}
```
NiuYawei, line 829: I don't quite follow the logic here (and the SOE selecting in heap_soemb_active_get()), are we trying to use the buckets in SOE set in a round-robin manner?

src/common/dav_v2/heap.c
```diff
@@ -600,6 +772,85 @@ heap_mbrt_mb_reclaim_garbage(struct palloc_heap *heap, uint32_t zid)
...
+			smbrt->svec[i] = heap_mbrt_get_mb(heap, mb_id);
+			break;
+		}
+		mb = mbrt_qbs_getmb(&smbrt->qbs, 1);
+		if (mb) {
+			smbrt->svec[i] = mb;
+			break;
+		}
+		break;
+	}
```
NiuYawei, line 850: Above loop is to create ensure there is always 1 available SOE bucket or SOEMB_ACTIVE_CNT buckets? (The loop breaks if any bucket setup successfully).

I think the goal here is to ensure enough free space being pinned in memory for the potential spilling over happened in next transaction, right? So we'd replace any unqualified bucket (in 'svec') with qualified one to satisfy the space requirement. It looks to me it's too late to remove the unqualified bucket from 'svec' in heap_recycle_soembs().

src/common/dav_v2/tx.c
sherintg: Yes,  this will be moved when SOEMB starts using the evictable MB pool. For this release SOEMB will use the cache pages from umem_cache_reserve() for creating a new non-evictable SOEMB. Hence this code is moved to the end of tx_begin() so that the page can be initialized in the same way as non-evictable memory bucket.

src/common/dav_v2/heap.c
sherintg: The allocator will maintain SOEMB_ACTIVE_CNT == 3 active SOEMBs. The allocator will first attempt spill over to svec[0] and if it fails svec[1] and finally svec[2]. If the allocation still fails then it will spill over to the global non-evictable memory buckets.
fur_idx indicates the furthest index of active SOEMB list that was used to do the spill over within a TX. If its value is greater than 1 then in the next tx_begin(), heap_soemb_reserve() will mark svec[0] as passive and left shifts other active SOEMBs.

src/common/dav_v2/heap.c
sherintg: Outside of the early boot phase, this condition primarily occurs when fur_idx is greater than 1, which triggers a left shift of the active SOEMBs by one position.  This implies that only svec[2] has to be populated. 

The left shift happen when atleast one allocation within a TX failed to spill over to svec[0] and svec[1] and allocator ends up using svec[2]. Hence svec[2] will be very sparsely populated. After the left shift svec[0] will be almost full, svec[1] sparsely populated and svec[2] not populated (or least populated if obtained from passive list). The assumption made here is that the free space in svec[1] and svec[2] is sufficient to satisfy the next TX.

NiuYawei (approved):

wangshilong (approved):

PR #15657, opened 2024-12-21
DAOS-16876 vos: set cont parameter when deregister modification from DTX

Nasf-Fan:
As long as the container is not destroyed, then anytime want to deregister a modification from related active DTX entry (that is usually triggered for vos discard or aggregation), the caller needs to offer container handle to vos_dtx_deregister_record() for locating the DTX entry in active DTX table. Otherwise, if the caller offers empty container handle, then it will cause dangling reference in related DTX entry as to data corruption in subsequent DTX commit or abort.

On the other hand, if the container will be destroyed, then all related DTX entries for such container will be useless any more. We need to destroy DTX table firstly to avoid generating dangling DTX references during destroying the container.

### Gatekeeper:

### DIFF
--- a/src/vos/vos_dtx.c
+++ b/src/vos/vos_dtx.c
@@ -1584,7 +1584,6 @@ vos_dtx_deregister_record(struct umem_instance *umm, daos_handle_t coh,
 			  uint32_t entry, daos_epoch_t epoch, umem_off_t record)
 {
 	struct dtx_handle		*dth = vos_dth_get(false);
-	struct vos_container		*cont;
 	struct vos_dtx_act_ent		*dae;
 	struct vos_dtx_act_ent_df	*dae_df;
 	umem_off_t			*rec_df;
@@ -1593,20 +1592,19 @@ vos_dtx_deregister_record(struct umem_instance *umm, daos_handle_t coh,
 	int				 rc;
 	int				 i;
 
+	/*
+	 * If @coh is empty handle, then we are destroying the container. Under such case,
+	 * both the in-DRAM and on-dish DTX entry have already been released or destroyed.
+	 */
+	if (daos_handle_is_inval(coh))
+		return 0;
+
 	if (!vos_dtx_is_normal_entry(entry))
 		return 0;
 
 	D_ASSERT(entry >= DTX_LID_RESERVED);
 
-	cont = vos_hdl2cont(coh);
-	/* If "cont" is NULL, then we are destroying the container.
-	 * Under such case, the DTX entry in DRAM has been removed,
-	 * The on-disk entry will be destroyed soon.
-	 */
-	if (cont == NULL)
-		return 0;
-
-	found = lrua_lookupx(cont->vc_dtx_array, entry - DTX_LID_RESERVED,
+	found = lrua_lookupx(vos_hdl2cont(coh)->vc_dtx_array, entry - DTX_LID_RESERVED,
 			     epoch, &dae);
 	if (!found) {
 		D_WARN("Could not find active DTX record for lid=%d, epoch="
--- a/src/vos/vos_gc.c
+++ b/src/vos/vos_gc.c
@@ -296,6 +296,15 @@ gc_drain_cont(struct vos_gc *gc, struct vos_pool *pool, daos_handle_t coh,
 	int			 i;
 	int			 rc;
 
+	/*
+	 * When we prepaer to drain the container, we do not need DTX entry any long.
+	 * Then destroy DTX table firstly to avoid dangling DXT records during drain
+	 * the container (that may yield).
+	 */
+	rc = vos_dtx_table_destroy(&pool->vp_umm, cont);
+	if (rc != 0)
+		return rc;
+
 	/** Move any leftover bags to the pool gc */
 	for (i = GC_AKEY; i < GC_CONT; i++) {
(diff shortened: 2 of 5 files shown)


src/vos/vos_gc.c
```diff
@@ -296,6 +296,15 @@ gc_drain_cont(struct vos_gc *gc, struct vos_pool *pool, daos_handle_t coh,
 	int			 i;
 	int			 rc;
 
+	/*
+	 * When we prepaer to drain the container, we do not need DTX entry any long.
+	 * Then destroy DTX table firstly to avoid dangling DXT records during drain
+	 * the container (that may yield).
+	 */
+	rc = vos_dtx_table_destroy(&pool->vp_umm, cont);
```
NiuYawei, line 304: I don't see why it's moved here, then it'll called on draining every object? The original place looks more proper to me.

src/vos/vos_obj_index.c
```diff
@@ -945,7 +945,7 @@ oi_iter_aggregate(daos_handle_t ih, bool range_discard)
 		if (rc != 0)
 			D_ERROR("Could not evict object "DF_UOID" "DF_RC"\n",
 				DP_UOID(oid), DP_RC(rc));
-		rc = dbtree_iter_delete(oiter->oit_hdl, NULL);
+		rc = dbtree_iter_delete(oiter->oit_hdl, oiter->oit_cont);
```
NiuYawei, line 948: This is one of the reasons I suggested reverting the "special OI deletion for layout upgrading" (see the oi_delete_arg), this was for EC rotation layout upgrade, it's not useful for 2.8 anymore. There are lots of potential issues in this PR, and it'll likely cause more issues in the future. (your current PR is an example), please revert that change before your fix.

src/vos/vos_dtx.c
```diff
@@ -1565,20 +1564,19 @@ vos_dtx_deregister_record(struct umem_instance *umm, daos_handle_t coh,
...
-	 * Under such case, the DTX entry in DRAM has been removed,
-	 * The on-disk entry will be destroyed soon.
-	 */
-	if (cont == NULL)
-		return 0;
-
-	found = lrua_lookupx(cont->vc_dtx_array, entry - DTX_LID_RESERVED,
+	found = lrua_lookupx(vos_hdl2cont(coh)->vc_dtx_array, entry - DTX_LID_RESERVED,
 			     epoch, &dae);
 	if (!found) {
```
NiuYawei, line 1581: Return error?

src/vos/vos_gc.c
Nasf-Fan: Because in gc_drain_cont(), there is assertion:
`        D_ASSERT(daos_handle_is_inval(coh));
        return gc_drain_btr(gc, pool, coh, item, &cont->cd_obj_root, credits, empty);
`
Means that the `coh` parameter for dtx_deregister will always be zero for draining the container. That will cause related DTX entry to be left as dangling. To avoid such case, since it will be drained, the DTX table is useless any longer, we can destroyed the DTX table firstly. On the other hand, even if the `coh` is not zero, deregistering DTX records one by one is also inefficient for drain case, destroying the table will be more efficient.

src/vos/vos_obj_index.c
Nasf-Fan: This patch will be back-ported to release/2.6, not sure whether it will affect. I am thinking that if Di's patch is really useless any longer, we can directly revert it? Or via another independent patch?

src/vos/vos_gc.c
```diff
@@ -791,9 +796,6 @@ gc_get_container(struct vos_pool *pool)
 	 */
 	cont = d_list_pop_entry(&pool->vp_gc_cont, struct vos_container,
 				vc_gc_link);
-	if (DAOS_FAIL_CHECK(DAOS_VOS_GC_CONT_NULL))
-		D_ASSERT(cont == NULL);
-
```
NiuYawei, line 796: Why this test code is removed?

Nasf-Fan: Why we need it?

NiuYawei (approved):

Nasf-Fan:
Please help to review the patch with higher priority, that is for a P1 ticket, thanks!

src/vos/vos_gc.c
```diff
@@ -296,6 +296,15 @@ gc_drain_cont(struct vos_gc *gc, struct vos_pool *pool, daos_handle_t coh,
 	int			 i;
 	int			 rc;
 
+	/*
+	 * When we prepaer to drain the container, we do not need DTX entry any long.
```
jolivier23, line 300: typo s/prepaer/prepare

src/vos/vos_obj.c
```diff
@@ -2599,7 +2599,7 @@ vos_obj_iter_aggregate(daos_handle_t ih, bool range_discard)
 		 *	be aborted. Then it will be added and handled via GC when ktr_rec_free().
 		 */
 
-		rc = dbtree_iter_delete(oiter->it_hdl, NULL);
+		rc = dbtree_iter_delete(oiter->it_hdl, obj->obj_cont);
```
jolivier23, line 2602: this seems like the main bug here

jolivier23 (approved):

PR #15659, opened 2024-12-22
DAOS-10028 client: Add Go bindings for libdaos (Pool)

mjmac:
Start the work of converting the raw cgo in the daos
tool into proper Go bindings for libdaos. This patch
covers pool functionality and adds some new infrastructure
common to both pools and containers.

Features: daos_cmd pool
Required-githooks: true


mjmac:
Note to reviewers... Yeah, this is big. It's the first of several patches, though, and contains a lot of the infrastructure needed to start the change over to using the client API instead of raw cgo in the daos tool. There is no new functionality here, and my focus will be on maintaining parity with the existing implementation for now. Most of the "new" code is just the existing cgo moved over into the API, along with extensive unit tests. Imagine that, unit tests!

src/control/lib/daos/api/attribute.go
```diff
@@ -0,0 +1,268 @@
...
+
+func listDaosAttributes(hdl C.daos_handle_t, at attrType) ([]string, error) {
+	var rc C.int
+	expectedSize, totalSize := C.size_t(0), C.size_t(0)
+
+	switch at {
+	case poolAttr:
+		rc = daos_pool_list_attr(hdl, nil, &totalSize, nil)
+	/*case contAttr:
+	rc = daos_cont_list_attr(hdl, nil, &totalSize, nil)*/
```
kjacque, line 52: Does this need to be enabled? Or removed?

src/control/lib/daos/api/attribute.go
```diff
@@ -0,0 +1,268 @@
...
+	attrNames := []string{}
+	expectedSize = totalSize
+	cNamesBuf := C.malloc(totalSize)
+	defer C.free(cNamesBuf)
+
+	switch at {
+	case poolAttr:
+		rc = daos_pool_list_attr(hdl, (*C.char)(cNamesBuf), &totalSize, nil)
+	/*case contAttr:
+	rc = daos_cont_list_attr(hdl, (*C.char)(buf), &totalSize, nil)*/
```
kjacque, line 73: Same question

src/control/lib/daos/api/attribute.go
```diff
@@ -0,0 +1,268 @@
...
+	// Next, create a slice of C.size_t entries to hold the sizes of the values.
+	// We have to do this first in order to know the buffer sizes to allocate
+	// before fetching the actual values.
+	cAttrSizes := make([]C.size_t, numAttr)
+	var rc C.int
+	switch at {
+	case poolAttr:
+		rc = daos_pool_get_attr(hdl, C.int(numAttr), &cAttrNames[0], nil, &cAttrSizes[0], nil)
+	/*case contAttr:
+	rc = daos_cont_get_attr(hdl, C.int(numAttr), &attrNames[0], nil, &attrSizes[0], nil)*/
```
kjacque, line 129: Same question -- if these are being left as placeholders, it would be better to add a TODO with a ticket number.

src/control/lib/daos/api/attribute.go
```diff
@@ -0,0 +1,268 @@
...
+		attrNames[i] = C.CString(attr.Name)
+	}
+	defer func(nameSlice []*C.char) {
+		for _, name := range nameSlice {
+			freeString(name)
+		}
+	}(attrNames)
+
+	// Next, create a slice of C.size_t entries to hold the sizes of the values,
+	// and a slice of pointers to the actual values.
```
kjacque, line 205: Any reason this step couldn't be done at the same time as the previous one? I understand why we couldn't do that for the getter, but the setter should be able to build all of these arrays at the same time.

src/control/lib/daos/api/handle.go
```diff
@@ -0,0 +1,92 @@
...
+	ch.daosHandle.cookie = 0
+}
+
+// FillHandle copies the handle to the supplied pointer,
+// which must be a reference to a C.daos_handle_t.
+// NB: Caller is responsible for keeping the copy in sync with
+// this handle -- use of this method should be discouraged as
+// it is provided for compatibility with older code that calls
+// into libdaos directly.
+func (ch *connHandle) FillHandle(cHandle unsafe.Pointer) error {
```
kjacque, line 59: The caller is a CGo function, right? Any reason `cHandle` shouldn't explicitly be a `*C,daos_handle_t`?

src/control/lib/daos/api/api_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// TestMain defines package-level setup and teardown logic.
+// NB: Any packages that run tests which depend on the API stubs
+// should copy this function in order to avoid interference
+// between parallel tests. Go parallelizes testing across
+// packages (but not within a package) by default.
+//
+// Long-term, this should be phased out as API users should mock
+// the API instead of relying on the stubs.
```
kjacque, line 21: IMO we probably need a task to change any existing code at the higher level so people don't imitate it.

src/control/lib/daos/api/util.h
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/*static inline uint32_t
+get_rebuild_state(struct daos_rebuild_status *drs)
+{
+	if (drs == NULL)
+		return 0;
+
+	return drs->rs_state;
+}*/
```
kjacque, line 99: Should this be commented out?

kjacque (commented):
A nice cleanup overall. I do find myself worrying about the rapidly-expanding stubs, though. I'm also curious how people are supposed to mock the API without the stubs, given that they are functions and not interfaces.

src/control/lib/daos/api/attribute.go
mjmac: This was code moved over from the tool... The next PR (#15721) migrates a lot of the container logic into the API, and these are uncommented.

src/control/lib/daos/api/attribute.go
mjmac: Next PR (https://github.com/daos-stack/daos/pull/15721) is coming in hot right behind this one. :) Trying to make them a little easier to review in chunks.

src/control/lib/daos/api/attribute.go
mjmac: This code has been working as-is for a couple of years now. My inclination is to get it all moved inside the API with proper unit tests first, and then work on cleaning up inefficiencies with confidence that the unit tests will find any mistakes.

src/control/lib/daos/api/handle.go
mjmac: No, this function is for users outside of the API boundary (e.g. the daos tool). Can't use C types in public function signatures. Eventually we should be able to remove this entirely once the API is complete and all of the tool code is cleaned up.

src/control/lib/daos/api/util.h
mjmac: Hmm, I think that's cruft that I missed. I'll remove it if I need to make changes to this PR, or in the next one.

src/control/lib/daos/api/api_test.go
mjmac: Yeah, good point. I see these as temporary scaffolding for the construction of the API. I'll update the comment to make that more explicit.

mjmac:
> A nice cleanup overall. I do find myself worrying about the rapidly-expanding stubs, though.
I really don't see a better way to make the code testable, but I'm open to suggestions. Personally, I think it's fine for the long term because all of the stub stuff should eventually be hidden behind the API, so API users and external tests don't need to deal with them.

>  I'm also curious how people are supposed to mock the API without the stubs, given that they are functions and not interfaces.
I was thinking about this the other day. I started playing around with the api.Provider thing when I added the health check stuff. One option would be to expand that to include methods for all of the API functions. At the moment, though, I think that would be a maintenance headache.

I think a better way to do this is to delegate mocking to users of the API. So, e.g. out in the daos tool we can define interfaces for the parts of the API that we want to test, and just create wrappers to implement those interfaces. This seems like a more maintainable and cleaner solution. I think it will be clearer as more of the tool code is moved into the API, and the code that remains under `cmd/daos` is mostly presentation layer, which means that the tests should be pretty simple.

tanabarr (approved):
No blocking issues but I really dislike the "shoehorning" of target-indexes into RankSetFlag just because it's a numerical list.

src/control/cmd/dmg/pool.go
```diff
@@ -700,20 +701,16 @@ func (cmd *poolQueryCmd) Execute(args []string) error {
 type poolQueryTargetsCmd struct {
 	poolCmd
 
-	Rank    uint32 `long:"rank" required:"1" description:"Engine rank of the targets to be queried"`
-	Targets string `long:"target-idx" description:"Comma-separated list of target idx(s) to be queried"`
+	Rank    uint32         `long:"rank" required:"1" description:"Engine rank of the target(s) to be queried"`
+	Targets ui.RankSetFlag `long:"target-idx" description:"Comma-separated list of target index(es) to be queried (default: all)"`
```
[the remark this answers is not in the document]
mjmac: Yeah, I agree that it would probably be better to create a new flag type. The nice thing about `RankSetFlag` is that it handles individual numbers, comma-separated numbers, or ranges (with or without enclosing brackets). There are also helpers already for converting rank slices to uint32 slices and vice-versa. Probably better to do that cleanup in a standalone PR, though.

src/control/lib/daos/api/attribute.go
kjacque: Okay, fair point.

src/control/lib/daos/api/handle.go
kjacque: How does the caller end up with a C handle if it's not using C types? Just a piece that hasn't been translated to the API yet?

src/control/lib/daos/api/attribute.go
kjacque: Okay, just wanted to make sure it wasn't forgotten. Sounds okay to me.

src/control/lib/daos/api/handle.go
mjmac: > Just a piece that hasn't been translated to the API yet?

Yes, exactly. There's still a ton of code in the daos tool that does raw libdaos calls and those need pool/container handles. This helper lets us take advantage of the new pool and container connection management code while still working with the stuff that hasn't been migrated yet.

kjacque:
> I really don't see a better way to make the code testable, but I'm open to suggestions. Personally, I think it's fine for the long term because all of the stub stuff should eventually be hidden behind the API, so API users and external tests don't need to deal with them.

I think that's the piece that's missing for me. The stub usage climbs up out of the layer where it should be isolated.

To be honest I'd prefer the stubbing to be done at runtime in the test cases themselves, similar to what we do elsewhere in the code, by having a thin wrapper type that can have interfaces written around it, passing around functions as parameters, setting them as members of structs, etc. Any of those methods seem better to me than having build-time stubs similar to what we are forced to do in the C code.

For the time being I think it's fine, any coverage is better than nothing. But philosophically, I do think tests are more readable if you have most of the information in the test itself, regardless of the method used.

> I think a better way to do this is to delegate mocking to users of the API. So, e.g. out in the daos tool we can define interfaces for the parts of the API that we want to stub, and just create wrappers to implement those interfaces. This seems like a more maintainable and cleaner solution.

I'm inclined toward the API package defining the wrapper struct itself, but otherwise I agree. One big interface seems guaranteed to be a maintenance nightmare (and leads to more bloated mock structures that try to be everything for every test, and become hard to understand). Makes sense to have the callers define their own interfaces, and mock those small interfaces in their tests.

All that is to say I think this is fine to land and continue to iterate. We can debate (or not) about approaches, but it's all an improvement over the current untestable tool. Nice work.

kjacque (approved):
Any remaining issues can be improved upon in the coming follow-on patches. Nice work.

src/control/lib/daos/api/handle.go
kjacque: Got it, thanks.

knard38 (approved):
Mostly OK for me for what I understand.
If my two comments are relevant, I have no issues that they will be fix in a followup PR.

src/control/cmd/daos/pool.go
```diff
@@ -67,147 +67,72 @@ type PoolID struct {
...
-		}
-	}
 
-	return func() {
-		cmd.disconnectPool()
-	}, nil
-}
+		pLabel := C.CString(cmd.pool.Label)
+		defer freeString(pLabel)
+		C.strncpy(&ap.pool_str[0], pLabel, C.DAOS_PROP_LABEL_MAX_LEN)
```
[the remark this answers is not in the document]
mjmac: OK, I'll add that in the next PR.

PR #15671, opened 2024-12-31
DAOS-14011 rebuild: incremental reintegration

liuxuezhao:
incremental reintegration, set pool property "reintegration:incremental" to
enable it, disabled by default.
1. dmg pool set-prop pool_label "reintegration:incremental", or
2. dmg pool create pool_label--properties="reintegration:incremental"

Some internals -
1. Maintain global stable epoch through IV, it is the globally minimal
    stable epoch of all engines' all targets for the container. 
2. Reintegrate will rebuild from that stable epoch.
3. Some containers/objects possibly be destroyed between exclude and
     reint, so during reintegration should try to discard those containers
     or objects. Add reint post process for that purpose.

Signed-off-by: Xuezhao Liu <xuezhao.liu@hpe.com>

### Gatekeeper:

### DIFF
--- a/src/container/srv_container.c
+++ b/src/container/srv_container.c
@@ -1847,7 +1847,6 @@ ds_cont_tgt_refresh_track_eph(uuid_t pool_uuid, uuid_t cont_uuid,
 	} else {
 		D_ALLOC_ARRAY(arg.tgt_status, dss_tgt_nr);
 		if (arg.tgt_status == NULL) {
-			ds_pool_put(pool);
 			return -DER_NOMEM;
 		}
 	}
@@ -1954,6 +1953,14 @@ cont_track_eph_leader_ult(void *arg)
 			new_eph = d_hlc2sec(min_ec_agg_eph);
 			if (cur_eph && new_eph > cur_eph && (new_eph - cur_eph) >= 600)
 				D_WARN(DF_CONT": Sluggish EC boundary reporting. "
+				       "cur:"DF_U64" new:"DF_U64" gap:"DF_U64"\n",
+				       DP_CONT(svc->cs_pool_uuid, eph_ldr->ea_cont_uuid),
+				       cur_eph, new_eph, new_eph - cur_eph);
+
+			cur_eph = d_hlc2sec(eph_ldr->ea_current_stable_eph);
+			new_eph = d_hlc2sec(min_stable_eph);
+			if (cur_eph && new_eph > cur_eph && (new_eph - cur_eph) >= 600)
+				D_WARN(DF_CONT": Sluggish stable epoch reporting. "
 				       "cur:"DF_U64" new:"DF_U64" gap:"DF_U64"\n",
 				       DP_CONT(svc->cs_pool_uuid, eph_ldr->ea_cont_uuid),
 				       cur_eph, new_eph, new_eph - cur_eph);
--- a/src/container/srv_container.c
+++ b/src/container/srv_container.c
@@ -1,5 +1,6 @@
 /*
  * (C) Copyright 2016-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -21,6 +22,7 @@
 #include <daos/pool.h>
 #include <daos_srv/pool.h>
 #include <daos_srv/rdb.h>
+#include <daos_srv/vos.h>
 #include <daos_srv/security.h>
 #include "rpc.h"
 #include "srv_internal.h"
@@ -185,8 +187,8 @@ ds_cont_svc_fini(struct cont_svc **svcp)
 	*svcp = NULL;
 }
 
-static int cont_svc_ec_agg_leader_start(struct cont_svc *svc);
-static void cont_svc_ec_agg_leader_stop(struct cont_svc *svc);
+static int cont_svc_eph_track_leader_start(struct cont_svc *svc);
+static void cont_svc_eph_track_leader_stop(struct cont_svc *svc);
 
 int
 ds_cont_svc_step_up(struct cont_svc *svc)
@@ -202,7 +204,7 @@ ds_cont_svc_step_up(struct cont_svc *svc)
 	}
 	D_ASSERT(svc->cs_pool != NULL);
 
-	rc = cont_svc_ec_agg_leader_start(svc);
+	rc = cont_svc_eph_track_leader_start(svc);
 	if (rc != 0)
 		D_ERROR(DF_UUID": start ec agg leader failed: "DF_RC"\n",
 			DP_UUID(svc->cs_pool_uuid), DP_RC(rc));
@@ -213,7 +215,7 @@ ds_cont_svc_step_up(struct cont_svc *svc)
 void
 ds_cont_svc_step_down(struct cont_svc *svc)
 {
-	cont_svc_ec_agg_leader_stop(svc);
+	cont_svc_eph_track_leader_stop(svc);
 	D_ASSERT(svc->cs_pool != NULL);
 	ds_pool_put(svc->cs_pool);
 	svc->cs_pool = NULL;
@@ -1518,7 +1520,7 @@ out:
 }
 
 static void
-cont_ec_agg_delete(struct cont_svc *svc, uuid_t cont_uuid);
+cont_track_eph_leader_delete(struct cont_svc *svc, uuid_t cont_uuid);
 
 static int
 cont_destroy(struct rdb_tx *tx, struct ds_pool_hdl *pool_hdl, struct cont *cont, crt_rpc_t *rpc,
@@ -1571,7 +1573,7 @@ cont_destroy(struct rdb_tx *tx, struct ds_pool_hdl *pool_hdl, struct cont *cont,
 	if (rc != 0)
 		goto out_prop;
 
-	cont_ec_agg_delete(cont->c_svc, cont->c_uuid);
+	cont_track_eph_leader_delete(cont->c_svc, cont->c_uuid);
 
         if (pool_hdl->sph_global_ver >= DAOS_POOL_GLOBAL_VERSION_WITH_OIT_OID_KVS) {
                 need_destroy_oid_oit_kvs = true;
@@ -1647,32 +1649,32 @@ out:
 	return rc;
 }
 
-struct cont_ec_agg *
-cont_ec_agg_lookup(struct cont_svc *cont_svc, uuid_t cont_uuid)
+struct cont_track_eph_leader *
+cont_track_eph_leader_lookup(struct cont_svc *cont_svc, uuid_t cont_uuid)
 {
-	struct cont_ec_agg *ec_agg;
+	struct cont_track_eph_leader *eph_ldr;
 
-	d_list_for_each_entry(ec_agg, &cont_svc->cs_ec_agg_list, ea_list) {
-		if (ec_agg->ea_deleted)
+	d_list_for_each_entry(eph_ldr, &cont_svc->cs_cont_ephs_leader_list, ea_list) {
+		if (eph_ldr->ea_deleted)
 			continue;
-		if (uuid_compare(ec_agg->ea_cont_uuid, cont_uuid) == 0)
-			return ec_agg;
+		if (uuid_compare(eph_ldr->ea_cont_uuid, cont_uuid) == 0)
+			return eph_ldr;
 	}
 	return NULL;
 }
 
 static int
-cont_ec_agg_alloc(struct cont_svc *cont_svc, uuid_t cont_uuid,
-		  struct cont_ec_agg **ec_aggp)
+cont_track_eph_leader_alloc(struct cont_svc *cont_svc, uuid_t cont_uuid,
+			    struct cont_track_eph_leader **leader_p)
 {
-	struct cont_ec_agg	*ec_agg = NULL;
-	struct pool_domain	*doms;
-	int			rank_nr;
-	int			rc = 0;
+	struct cont_track_eph_leader	*eph_ldr = NULL;
+	struct pool_domain		*doms;
+	int				 rank_nr;
+	int				 rc = 0;
 	int			i;
 
-	D_ALLOC_PTR(ec_agg);
-	if (ec_agg == NULL)
+	D_ALLOC_PTR(eph_ldr);
+	if (eph_ldr == NULL)
 		return -DER_NOMEM;
 
 	D_ASSERT(cont_svc->cs_pool->sp_map != NULL);
@@ -1680,42 +1682,43 @@ cont_ec_agg_alloc(struct cont_svc *cont_svc, uuid_t cont_uuid,
 	if (rank_nr < 0)
 		D_GOTO(out, rc = rank_nr);
 
-	D_ALLOC_ARRAY(ec_agg->ea_server_ephs, rank_nr);
-	if (ec_agg->ea_server_ephs == NULL)
+	D_ALLOC_ARRAY(eph_ldr->ea_server_ephs, rank_nr);
+	if (eph_ldr->ea_server_ephs == NULL)
 		D_GOTO(out, rc = -DER_NOMEM);
 
-	uuid_copy(ec_agg->ea_cont_uuid, cont_uuid);
-	ec_agg->ea_servers_num = rank_nr;
-	ec_agg->ea_current_eph = 0;
+	uuid_copy(eph_ldr->ea_cont_uuid, cont_uuid);
+	eph_ldr->ea_servers_num = rank_nr;
+	eph_ldr->ea_current_ec_agg_eph = 0;
 	for (i = 0; i < rank_nr; i++) {
-		ec_agg->ea_server_ephs[i].rank = doms[i].do_comp.co_rank;
-		ec_agg->ea_server_ephs[i].eph = 0;
+		eph_ldr->ea_server_ephs[i].re_rank = doms[i].do_comp.co_rank;
+		eph_ldr->ea_server_ephs[i].re_ec_agg_eph = 0;
+		eph_ldr->ea_server_ephs[i].re_stable_eph = 0;
 	}
-	d_list_add(&ec_agg->ea_list, &cont_svc->cs_ec_agg_list);
-	*ec_aggp = ec_agg;
+	d_list_add(&eph_ldr->ea_list, &cont_svc->cs_cont_ephs_leader_list);
+	*leader_p = eph_ldr;
 out:
 	if (rc) {
-		if (ec_agg)
-			D_FREE(ec_agg->ea_server_ephs);
-		D_FREE(ec_agg);
+		if (eph_ldr)
+			D_FREE(eph_ldr->ea_server_ephs);
+		D_FREE(eph_ldr);
 	}
 
 	return rc;
 }
 
 static void
-cont_ec_agg_delete(struct cont_svc *svc, uuid_t cont_uuid)
+cont_track_eph_leader_delete(struct cont_svc *svc, uuid_t cont_uuid)
 {
-	struct cont_ec_agg	*ec_agg;
+	struct cont_track_eph_leader	*eph_ldr;
 
-	ec_agg = cont_ec_agg_lookup(svc, cont_uuid);
-	if (ec_agg == NULL)
+	eph_ldr = cont_track_eph_leader_lookup(svc, cont_uuid);
+	if (eph_ldr == NULL)
 		return;
 
-	/* Set ea_deleted flag to destroy it inside cont_agg_eph_leader_ult()
+	/* Set ea_deleted flag to destroy it inside cont_track_eph_leader_ult()
 	 * to avoid list iteration broken.
 	 */
-	ec_agg->ea_deleted = 1;
+	eph_ldr->ea_deleted = 1;
 }
 
 /**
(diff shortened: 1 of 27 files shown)


liuxuezhao:
the 1st commit is from @Nasf-Fan 's PR https://github.com/daos-stack/daos/pull/15605 
other commits is incremental reint code
will rebase after fanyong's PR landed later.

daltonbohning:
> removed request for https://github.com/orgs/daos-stack/teams/ftest-watchers, https://github.com/orgs/daos-stack/teams/ftest-owners, https://github.com/orgs/daos-stack/teams/object-owners and https://github.com/orgs/daos-stack/teams/object-watchers

Why are you manually removing review requests from codeowners? The whole point is to make sure there are reviews for each area of code

src/tests/ftest/util/server_utils_params.py
```diff
@@ -441,6 +441,7 @@ class EngineYamlParameters(YamlParameters):
             "D_LOG_FILE_APPEND_PID=1",
             "DAOS_POOL_RF=4",
             "CRT_EVENT_DELAY=1",
+            "DAOS_VOS_AGG_GAP=25",
```
daltonbohning, line 444: We need to consider whether this actually belongs in test code because it means we will not be testing what is used in production

liuxuezhao:
> > removed request for https://github.com/orgs/daos-stack/teams/ftest-watchers, https://github.com/orgs/daos-stack/teams/ftest-owners, https://github.com/orgs/daos-stack/teams/object-owners and https://github.com/orgs/daos-stack/teams/object-watchers
> 
> Why are you manually removing review requests from codeowners? The whole point is to make sure there are reviews for each area of code

the test code was added by Fanyong's PR, this PR just add it temporally and will rebase it lafter after this PR landed. 
My PR did not add such code. it was explained above.

daltonbohning:
> > > removed request for https://github.com/orgs/daos-stack/teams/ftest-watchers, https://github.com/orgs/daos-stack/teams/ftest-owners, https://github.com/orgs/daos-stack/teams/object-owners and https://github.com/orgs/daos-stack/teams/object-watchers
> > 
> > 
> > Why are you manually removing review requests from codeowners? The whole point is to make sure there are reviews for each area of code
> 
> the test code was added by Fanyong's PR, this PR just add it temporally and will rebase it lafter after this PR landed. My PR did not add such code. it was explained above.

Sorry for missing that - and I didn't mean to sound so harsh, now that I reread it...

src/container/srv_container.c
```diff
@@ -1647,196 +1649,248 @@ cont_destroy(struct rdb_tx *tx, struct ds_pool_hdl *pool_hdl, struct cont *cont,
...
-		cont_child->sc_ec_agg_eph_boundary = arg->min_eph;
+		cont_child->sc_ec_agg_eph_boundary < arg->min_ec_agg_eph ? "update" : "ignore",
+		cont_child->sc_ec_agg_eph_boundary, arg->min_ec_agg_eph,
+		cont_child->sc_stable_epoch < arg->min_stable_eph ? "update" : "ignore",
+		cont_child->sc_stable_epoch, arg->min_stable_eph);
+
+	if (cont_child->sc_ec_agg_eph_boundary < arg->min_ec_agg_eph)
+		cont_child->sc_ec_agg_eph_boundary = arg->min_ec_agg_eph;
+
+	/* Only should update local stable epoch if the target is in UPIN status */
```
Nasf-Fan, line 1817: For the target with `PO_COMP_ST_UP` status, should we consider its local stable epoch?

src/container/srv_container.c
```diff
@@ -1647,196 +1649,248 @@ cont_destroy(struct rdb_tx *tx, struct ds_pool_hdl *pool_hdl, struct cont *cont,
...
 	uuid_copy(arg.cont_uuid, cont_uuid);
-	arg.min_eph = eph;
+	arg.min_ec_agg_eph = ec_agg_eph;
+	arg.min_stable_eph = stable_eph;
+	if (likely(dss_tgt_nr <= EPH_ARG_TGT_INLINE)) {
+		arg.tgt_status = arg.tgt_status_inline;
+	} else {
+		D_ALLOC_ARRAY(arg.tgt_status, dss_tgt_nr);
+		if (arg.tgt_status == NULL) {
+			ds_pool_put(pool);
```
Nasf-Fan, line 1850: `pool` is NULL here.

src/container/srv_container.c
```diff
@@ -1853,59 +1907,64 @@ cont_agg_eph_leader_ult(void *arg)
...
-			cur_eph = d_hlc2sec(ec_agg->ea_current_eph);
-			new_eph = d_hlc2sec(min_eph);
+			D_DEBUG(DB_MD, DF_CONT" min_ec_agg_eph "DF_X64" current "DF_X64
+				", min_stable_eph "DF_X64" current "DF_X64".\n",
+				DP_CONT(svc->cs_pool_uuid, eph_ldr->ea_cont_uuid),
+				min_ec_agg_eph, eph_ldr->ea_current_ec_agg_eph,
+				min_stable_eph, eph_ldr->ea_current_stable_eph);
+
+			cur_eph = d_hlc2sec(eph_ldr->ea_current_ec_agg_eph);
+			new_eph = d_hlc2sec(min_ec_agg_eph);
```
Nasf-Fan, line 1954: Does it only affect ec_agg_epoch or also stable_epoch?

src/container/srv_target.c
```diff
@@ -2378,131 +2396,229 @@ ds_cont_oid_alloc_handler(crt_rpc_t *rpc)
...
+}
+
+static int
+cont_tgt_track_eph_init(struct ds_cont_child *cont_child)
+{
+	int rc;
+
+	ds_cont_child_get(cont_child);
+
+	rc = dss_ult_create(cont_tgt_track_eph_init_ult, cont_child, DSS_XS_SELF,
```
Nasf-Fan, line 2603: cont_tgt_track_eph_init_ult() will create another ULT cont_track_eph_init_ult(), why need two ULTs for that?

src/container/srv_target.c
```diff
@@ -2548,39 +2667,54 @@ ds_cont_tgt_ec_eph_query_ult(void *data)
...
-					D_ERROR("ignore for now "DF_X64" < "DF_X64
-						" "DF_UUID"\n", min_eph, ec_eph->ce_last_eph,
-						DP_UUID(ec_eph->ce_cont_uuid));
+			if (min_ec_agg_eph == 0 || min_ec_agg_eph == DAOS_EPOCH_MAX ||
+			    min_stable_eph == 0 || min_stable_eph == DAOS_EPOCH_MAX ||
+			    (min_ec_agg_eph <= ec_eph->cte_last_ec_agg_epoch &&
+			     min_stable_eph <= ec_eph->cte_last_stable_epoch)) {
+				if (min_ec_agg_eph > 0 && min_stable_eph > 0 &&
+				    min_ec_agg_eph <= ec_eph->cte_last_ec_agg_epoch &&
+				    min_stable_eph <= ec_eph->cte_last_stable_epoch)
```
Nasf-Fan, line 2684: Why EC_agg_epoch/stable_epoch cannot keep unchanged between two cycle checks? I think that is normal instead or ERROR().

src/container/srv_target.c
```diff
@@ -2548,39 +2667,54 @@ ds_cont_tgt_ec_eph_query_ult(void *data)
...
+				       DP_CONT(pool->sp_uuid, ec_eph->cte_cont_uuid), rc);
+			}
 		}
 		D_FREE(failed_tgts);
 yield:
 		if (dss_ult_exiting(pool->sp_ec_ephs_req))
 			break;
 
-		sched_req_sleep(pool->sp_ec_ephs_req, EC_TGT_AGG_INTV);
+		sched_req_sleep(pool->sp_ec_ephs_req, EC_TGT_EPH_QUERY_INTV);
```
Nasf-Fan, line 2717: That means per-5 seconds, current engine will report its local stable epoch (per-ppol) to the leader, right?

src/vos/vos_container.c
```diff
@@ -934,6 +934,13 @@ vos_cont_get_globla_stable_epoch(daos_handle_t coh)
 	cont = vos_hdl2cont(coh);
 	D_ASSERT(cont != NULL);
 
+	if (cont->vc_pool->vp_pool_df->pd_version < VOS_POOL_DF_2_8) {
+		D_DEBUG(DB_MD, DF_CONT" return 0 stable epoch for lower pool version %d\n",
+			DP_CONT(cont->vc_pool->vp_pool_df->pd_id, cont->vc_id),
+			cont->vc_pool->vp_pool_df->pd_version);
+		return 0;
+	}
```
Nasf-Fan, line 942: That seems unnecessary since for old pool either `cont_ext` is NULL or `cont_ext->ced_global_stable_epoch` is not set by default.

Nasf-Fan (changes requested):

src/container/srv_container.c
liuxuezhao: here it is setting local's persistent global stable epoch, only UPIN node can set it. 
For reint it should rebuild from the local stored stable epoch before exclude it (still be UPIN).

The UP tgt reports its local epoch by query vos_cont_get_local_stable_epoch() in cont_aggregate_interval().

src/container/srv_container.c
liuxuezhao: maybe an issue when rebase, will fix it soon, thx

src/container/srv_container.c
liuxuezhao: this is to throw a warning if the EC agg epoch boundary is updated too slow (> 600S).

Both ec_agg_epoch and stable_epoch are considered by other code in this func

src/container/srv_target.c
liuxuezhao: cont_tgt_track_eph_init_ult() execute on vos tgt main XS, it will execute cont_track_eph_init_ult on XS0. The pool->sp_ec_ephs_list only maintained (insert/lookup) on XS0

src/container/srv_target.c
liuxuezhao: The D_ERROR() is just kept from original code, commonly should not get smaller epoch,  but better to change the <= to < same as before.

src/container/srv_target.c
liuxuezhao: yes, reports each container's local stable epoch and ec agg epoch to leader per container.

src/vos/vos_container.c
liuxuezhao: that need some knowledge for example suing umem_zalloc() to allocate the vos_cont_ext_df. adding such explicit code looks harmless?

liuxuezhao (commented):
thanks for review @Nasf-Fan , will rebase later after your patch landed.

src/object/srv_obj_migrate.c
```diff
@@ -3271,8 +3272,70 @@ migrate_obj_ult(void *data)
 		}
 	}
 
+	if (!tls->mpt_reintegrating_check) {
+		struct pool_target *target;
+		struct ds_pool *pool = tls->mpt_pool->spc_pool;
+
+		ABT_rwlock_rdlock(pool->sp_lock);
```
Nasf-Fan: Is it possible during current ULT being blocked on the lock, another ULT do the same check `if (!tls->mpt_reintegrating_check)` ?

src/object/srv_obj_migrate.c
```diff
@@ -3271,8 +3272,70 @@ migrate_obj_ult(void *data)
...
+	}
+
+	if (tls->mpt_reintegrating) {
+		struct ds_cont_child *cont_child = NULL;
+
+		/* check again to see if the container is being destroyed. */
+		migrate_get_cont_child(tls, arg->cont_uuid, &cont_child, false);
+		if (cont_child != NULL && !cont_child->sc_stopping) {
+			if (vos_oi_exist(cont_child->sc_hdl, arg->oid)) {
+				rc = vos_dtx_cmt_reindex(cont_child->sc_hdl);
```
Nasf-Fan: It is unnecessary to reindex DTX if the opening container only for get the global stable epoch. On the other hand, since the global stable epoch is (at least) per-container, then we can pass it as one of the parameters for `migrate_obj_ult()` instead of loading it from VOS for every object again and again.

src/object/srv_obj_migrate.c
```diff
@@ -3271,8 +3272,70 @@ migrate_obj_ult(void *data)
...
 	for (i = 0; i < arg->snap_cnt; i++) {
-		epr.epr_lo = i > 0 ? arg->snaps[i - 1] + 1 : 0;
+		daos_epoch_t lower_epoch = 0;
+
+		if (arg->snaps[i] < stable_epoch) {
+			D_DEBUG(DB_REBUILD, DF_CONT " obj " DF_UOID " skip snap "DF_X64
+				" < stable " DF_X64 "\n", DP_CONT(arg->pool_uuid, arg->cont_uuid),
+				DP_UOID(arg->oid), arg->snaps[i], stable_epoch);
+			continue;
+		} else {
```
Nasf-Fan, line 3347: unnecessary `else` since above `continue`.

src/object/srv_obj_migrate.c
liuxuezhao: seems you are checking an older version, the call to vos_dtx_cmt_reindex removed already.
migrate_get_cont_child() is the original code maybe useful to detect if the cont destroy during rebuild WIP.

src/object/srv_obj_migrate.c
liuxuezhao: right, it can be removed.

src/object/srv_obj_migrate.c
```diff
@@ -3853,49 +3934,302 @@ ds_obj_migrate_handler(crt_rpc_t *rpc)
...
+		       void *data, unsigned *acts)
+{
+	struct reint_post_iter_arg	*arg = data;
+	struct migrate_pool_tls		*tls = arg->ria_tls;
+	daos_unit_oid_t			 uoid = entry->ie_oid;
+	int				 rc = 0;
+
+	/* for non-replica obj don't try to delete it */
+	if (daos_obj_id2ord(uoid.id_pub) == OR_RP_1)
+		goto out;
```
Nasf-Fan, line 3996: How to handle the following case: a 2-way object, related servers on which the two shards reside are both offline for some reason almost at the same time, then nobody update such object after that, and then when we reint the two servers, we should keep such 2-way object, right? Does current logic support that? Here, we need to discard such object, right?

liuxuezhao: right, this is a known issue to me. I wanted to discuss with you at weekly meeting but did not get time.
One explanation is with RP2 obj then RF <= 1, then losing 2 engines cannot ensure no data lose. But we should try the best to not lost data.
Anyway, we cannot handle everything in one patch, let's discuss and resolve it in following PR.

src/object/srv_obj_migrate.c
```diff
@@ -3853,49 +3934,302 @@ ds_obj_migrate_handler(crt_rpc_t *rpc)
...
+		}
+
+		DL_ERROR(rc, DF_RB " obj_tree_lookup_cont "DF_UUID" failed",
+			 DP_RB_MPT(tls), DP_UUID(entry->ie_couuid));
+		goto out;
+	}
+
+	D_ASSERT(daos_handle_is_valid(arg->ria_cont_toh));
+
+	rc = vos_cont_open(iter_param->ip_hdl, entry->ie_couuid, &coh);
```
NiuYawei, line 4066: Here we intentionally call VOS interface to open the container? I think the ds_cont_child in container level should be in started state? Why don't we just lookup & use  ds_cont_child directly?

src/container/srv_container.c
liuxuezhao: yes, there is a time window between excluding and every tgt server get the new pool map.
But as long as this tgt is still UPIN in current poolmap, it will report its local stable epoch and on PS leader will ensure the global stable epoch will not exceed every UPIN tgt.
So here even in the time window that this engine with stable pool map still safe to update its local stored global stable epoch, right?

src/container/srv_container.c
```diff
@@ -1647,196 +1649,246 @@ cont_destroy(struct rdb_tx *tx, struct ds_pool_hdl *pool_hdl, struct cont *cont,
...
+		cont_child->sc_ec_agg_eph_boundary = arg->min_ec_agg_eph;
+
+	/* Only should update local stable epoch if the target is in UPIN status */
+	if (cont_child->sc_stable_epoch < arg->min_stable_eph &&
+	    (arg->tgt_status[idx] & PO_COMP_ST_UPIN)) {
+		rc = vos_cont_set_global_stable_epoch(cont_child->sc_hdl, arg->min_stable_eph);
+		if (rc == 0)
+			cont_child->sc_stable_epoch = arg->min_stable_eph;
+		else
+			rc = 0;
```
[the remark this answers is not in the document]
liuxuezhao: the refresh callback returned err code seems will cause some trouble at IV path.
the vos_cont_set_global_stable_epoch()'s err cases are -
1) -DER_NOTSUPPORTED on old pool, this can be ignored as will not change anyghing.
2) -DER_NO_PERM for invalid epoch for very cornel case, also can be ignored with err/warn log
3) umm tx fail, that will not change persistent value, seems err/warn log enough?
these cases seems adding log is enough, I don't want cart IV path ivo_on_refresh()'s failure case affect too much to whole IV path's handling.

src/container/srv_container.c
```diff
@@ -1647,196 +1649,246 @@ cont_destroy(struct rdb_tx *tx, struct ds_pool_hdl *pool_hdl, struct cont *cont,
...
+		D_ALLOC_ARRAY(arg.tgt_status, dss_tgt_nr);
+		if (arg.tgt_status == NULL) {
+			return -DER_NOMEM;
+		}
+	}
+
+	rc = ds_pool_lookup(pool_uuid, &pool);
+	if (rc != 0) {
+		D_DEBUG(DB_MD, DF_UUID" lookup pool failed: %d\n",
+			DP_UUID(pool_uuid), rc);
```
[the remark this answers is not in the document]
liuxuezhao: ok, will change to err

src/container/srv_target.c
liuxuezhao: the pool->sp_ec_ephs_list only maintained on XS0, so create ULT and execute on XS0 need not take lock.
hmm, maybe need not create cont_tgt_track_eph_init_ult, just create cont_track_eph_init_ult on XS0, is this what you prefer?

src/object/srv_obj_migrate.c
liuxuezhao: I'll see if the ds_cont_child can be used, thx

src/container/srv_container.c
NiuYawei: hmm, on leader side, if a rank is DOWN, I think it'll be ignored when leader finding out the lowest stable epoch, right?

src/container/srv_target.c
NiuYawei: Yes, I just don't see why we need to create a ULT for creating a ULT on sys xstream. This dss_ult_create() looks not necessary to me, it also makes cont_track_eph_insert() not being executed before cont_tgt_track_eph_init() returns, that looks not quite right to me.

src/container/srv_container.c
liuxuezhao: right, that case is - leader with latest pool map that tgt A is DOWN, will ignore its reports then global stable epoch can exceed A's local stable epoch.
But A itself with stable poolmap still think itself is UPIN.
But when leader try to refresh the stable epoch by cont_iv_track_eph_refresh(), the IV msg propogate will find group version mismatch and return -DER_GRPVER, and will retry the IV OP until grp ver matched on all nodes.
So far looks fine?

src/container/srv_target.c
liuxuezhao: OK, i see, will change it. thx

src/container/srv_target.c
```diff
@@ -2572,39 +2691,55 @@ ds_cont_tgt_ec_eph_query_ult(void *data)
...
+						", or min_stable_eph "DF_X64" <"DF_X64
+						", "DF_UUID"\n",
+						min_ec_agg_eph, ec_eph->cte_last_ec_agg_epoch,
+						min_stable_eph, ec_eph->cte_last_stable_epoch,
+						DP_UUID(ec_eph->cte_cont_uuid));
 				else
 					D_DEBUG(DB_MD, "Skip eph "DF_X64"/"DF_X64
-						" "DF_UUID"\n", min_eph, ec_eph->ce_last_eph,
-						DP_UUID(ec_eph->ce_cont_uuid));
+						", "DF_X64"/"DF_X64", "DF_UUID"\n",
```
kccain, line 2717: suggest output of the log message includes description of what the values are, such as "Skip ec_agg_eph " for the first two, and then "stable_eph" for the second two values

kccain (commented):
This patch will take some time to fully review. Got through the container/ code and some other files. Not so much the object/ code yet. I will take a look at the rest of the patch in a second chunk of work. For now, some preliminary comments since they may be helpful sooner as the patch may be revised soon.

src/container/srv_container.c
NiuYawei: I see, I think we are good as long as the IV request with different pool map version can be rejected, thanks for the explanation.

src/container/srv_target.c
```diff
@@ -2372,131 +2397,229 @@ ds_cont_oid_alloc_handler(crt_rpc_t *rpc)
...
-	daos_epoch_t	*ce_ephs;
-	daos_epoch_t	ce_last_eph;
-	uint32_t	ce_ephs_cnt;
-	int		ce_ref;
+/* Track each container EC aggregation Epoch and stable epoch under ds_pool */
+struct cont_track_eph {
+	uuid_t		cte_cont_uuid;
+	d_list_t	cte_list;
+	/* each target's stable epoch */
+	daos_epoch_t	*cte_stable_ephs;
```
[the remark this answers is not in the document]
liuxuezhao: ok

src/container/srv_target.c
```diff
@@ -2402,131 +2420,229 @@ ds_cont_oid_alloc_handler(crt_rpc_t *rpc)
...
+	d_list_add(&new_ec->cte_list, ec_list);
+	new_ec->cte_ref = 0;
 	return new_ec;
 }
 
-int
-ds_cont_ec_eph_insert(struct ds_pool *pool, uuid_t cont_uuid, int tgt_idx,
-		      uint64_t **epoch_p)
+static int
+cont_track_eph_insert(struct ds_pool *pool, uuid_t cont_uuid, int tgt_idx,
```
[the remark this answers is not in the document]
liuxuezhao: the leader side was named as cont_track_eph_leader_alloc()/cont_track_eph_leader_lookup()/cont_track_eph_leader_delete() to differenciate.

src/container/srv_target.c
```diff
@@ -2402,131 +2420,229 @@ ds_cont_oid_alloc_handler(crt_rpc_t *rpc)
...
+struct track_eph_ult_arg {
+	struct ds_pool	*pool;
+	uuid_t		cont_uuid;
+	uint32_t	tgt_idx;
+	daos_epoch_t	*ec_agg_eph;
+	daos_epoch_t	*stable_eph;
+};
+
+static	int
+cont_track_eph_fini_ult(void *data)
```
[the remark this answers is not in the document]
liuxuezhao: no "_leader_" in func name means tgt

src/container/srv_target.c
```diff
@@ -2402,131 +2420,229 @@ ds_cont_oid_alloc_handler(crt_rpc_t *rpc)
...
+{
+	struct track_eph_ult_arg	*arg = data;
+
+	cont_track_eph_delete(arg->pool, arg->cont_uuid, arg->tgt_idx);
+	return 0;
+}
+
+static void
+cont_tgt_track_eph_fini(struct ds_cont_child *cont_child)
```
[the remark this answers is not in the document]
liuxuezhao: ditto

src/container/srv_target.c
liuxuezhao: sure

src/include/daos_srv/container.h
```diff
@@ -111,6 +112,12 @@ struct ds_cont_child {
 	/* Last timestamp when EC aggregation reports -DER_INPROGRESS. */
 	uint64_t		 sc_ec_agg_busy_ts;
 
+
+	/* The global minimum stable epoch. All data @lower epoch should has been globally
+	 * stable (committed or aborted). Used as the start epoch for incremental reintegration.
+	 */
+	uint64_t		sc_stable_epoch;
```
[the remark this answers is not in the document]
liuxuezhao: sure

src/pool/srv.c
```diff
@@ -71,6 +73,11 @@ init(void)
 	if (unlikely(ec_agg_disabled))
 		D_WARN("EC aggregation is disabled.\n");
 
+	daos_incr_reint_enabled = false;
```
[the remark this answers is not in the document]
liuxuezhao: in cont_svc_eph_track_leader_start(), the ec_agg_disabled checked looks can be removed.
and I am going to remove DAOS_INCR_REINT_ENABLE and replace it by pool property

src/vos/vos_container.c
```diff
@@ -814,3 +835,193 @@ struct vos_iter_ops vos_cont_iter_ops = {
...
+	umm = vos_cont2umm(cont);
+	cont_ext = umem_off2ptr(umm, cont->vc_cont_df->cd_ext);
+
+	/* Do not allow to set global stable epoch against old container without extension. */
+	if (cont_ext == NULL)
+		D_GOTO(out, rc = -DER_NOTSUPPORTED);
+
+	/*
+	 * Either the leader gives wrong global stable epoch or current target does not participant
+	 * in the calculating new globle stable epoch. Then do not allow to set globle stable epoch.
```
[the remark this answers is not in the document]
liuxuezhao: fixed

src/vos/vos_container.c
```diff
@@ -814,3 +835,193 @@ struct vos_iter_ops vos_cont_iter_ops = {
...
+	 */
+	if (unlikely(cont->vc_local_stable_epoch < epoch)) {
+		D_WARN("Invalid global stable epoch: " DF_X64" vs " DF_X64 " for container "
```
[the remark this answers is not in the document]
liuxuezhao: ok

liuxuezhao:
comments from @Nasf-Fan , @NiuYawei and @kccain should all have been addressed, thx
the last commit is for adding a pool property to enable incr reint (the ENV removed).

src/vos/vos_container.c
Nasf-Fan: It is harmless but redundant, not fatal.

Nasf-Fan (dismissed):

tanabarr (dismissed):
Go changes LGTM

src/container/srv_target.c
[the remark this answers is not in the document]
kccain: OK I guess the code can remain this way. Based on the below analysis(which I did again now to make sure I have it right), it is why I had suggested a renaming of the insert/delete functions, but I guess it is OK.

- ds_cont_track_eph_free() is actually operating at the levbel of a pool (iterate over a list of containers, each represented by a struct cont_track_eph). But naming the function to begin with ds_pool might seem strange in this file.
- cont_track_eph_alloc() and cont_track_eph_destroy() seem pretty straightforward naming and operate at a container level for this engine. They create or destroy a struct cont_track_eph (this is at the level of the container in this engine).
- cont_track_eph_insert() is not straightfoward. It refers to both the whole struct cont_track_eph (a container in this engine), and also a specific target index within the engine. It might alloc/create a whole struct cont_track_eph, increments a cte_ref count for the struct (that was looked up or created), and returns pointers to ec_agg_eph and stable_eph values for a specific target index.
- cont_track_eph_delete() also not totally straightforward. It refers to a particular target index, but only manipulates struct cont_track_eph.cte_ref count at the container/engine level.

src/container/srv_target.c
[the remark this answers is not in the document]
kccain: The reason for the suggestion is that the ULT is executing for a specific target index. The concern was not about leader vs. other engine. Instead, the concern is distinguishing between whole engine and a specific target.

The function that creates this ULT has "tgt" in its name, so it probably makes sense for this function to have "tgt" too.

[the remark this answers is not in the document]
kccain: oh but it's running on system XS, that's why tgt is not in the function name.

kjacque (commented):
I only inspected the Go changes -- LGTM.

Leaving it to reviewers of the C code to officially approve.

liuxuezhao:
the codespell warning is all not related with this PR
Error: ./src/dtx/tests/dts_local_rdb.c:48: re-using ==> reusing
Error: ./src/control/server/mgmt_system.go:634: synthesise ==> synthesize
Error: ./src/gurt/examples/telem_producer_example.c:14: incremements ==> increments
Error: ./src/vea/vea_internal.h:189: aggergate ==> aggregate

NiuYawei (dismissed):

Nasf-Fan (dismissed):

NiuYawei (approved):

wangshilong (approved):

liuxuezhao:
the "Linting / Codespell " check failed, but it is totally unrelated with my PR.
So may need force landing @gnailzenh @daos-stack/daos-gatekeeper

gnailzenh (approved):

daltonbohning:
> the "Linting / Codespell " check failed, but it is totally unrelated with my PR. So may need force landing @gnailzenh @daos-stack/daos-gatekeeper

Yep. Those snuck in somehow on files that haven't even changed. It's fixed now in master and 2.6

PR #15714, opened 2025-01-09
DAOS-16931 container: fix oid iv race with using invalid on_update ptr

mchaarawi:
the src pointer for on_update() callback is on the stack and should not be saved beyond that call to detect the same current request, because in some cases we can end up with 2 different requests sharing getting the same pointer value. Use the private pointer allocated by the oid iv on_alloc operation instead.

### Gatekeeper:

### DIFF
--- a/src/container/oid_iv.c
+++ b/src/container/oid_iv.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2017-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -133,10 +134,10 @@ oid_iv_ent_update(struct ds_iv_entry *ns_entry, struct ds_iv_key *iv_key,
 	entry = ns_entry->iv_value.sg_iovs[0].iov_buf;
 	rc = ABT_mutex_trylock(entry->lock);
 	/** For retry requests, from _iv_op(), the lock may not be released in some cases. */
-	if (rc == ABT_ERR_MUTEX_LOCKED && entry->current_req != src)
+	if (rc == ABT_ERR_MUTEX_LOCKED && entry->current_req != priv)
 		return -DER_BUSY;
 
-	entry->current_req = src;
+	entry->current_req = priv;
 	avail = &entry->rg;
 
 	oids = src->sg_iovs[0].iov_buf;


src/container/oid_iv.c
```diff
@@ -133,10 +134,10 @@ oid_iv_ent_update(struct ds_iv_entry *ns_entry, struct ds_iv_key *iv_key,
 	entry = ns_entry->iv_value.sg_iovs[0].iov_buf;
 	rc = ABT_mutex_trylock(entry->lock);
 	/** For retry requests, from _iv_op(), the lock may not be released in some cases. */
-	if (rc == ABT_ERR_MUTEX_LOCKED && entry->current_req != src)
+	if (rc == ABT_ERR_MUTEX_LOCKED && entry->current_req != priv)
```
frostedcmos, line 137: Just to double-check -- does on_get() implementation always return a valid priv or are there cases when priv returned can be NULL? (iv framework just passes whatever priv was filled with, we dont enforce it to be non-null). if latter, then two requests might have same null priv causing issues with this logic.

src/container/oid_iv.c
```diff
@@ -133,10 +134,10 @@ oid_iv_ent_update(struct ds_iv_entry *ns_entry, struct ds_iv_key *iv_key,
 	entry = ns_entry->iv_value.sg_iovs[0].iov_buf;
 	rc = ABT_mutex_trylock(entry->lock);
 	/** For retry requests, from _iv_op(), the lock may not be released in some cases. */
```
frostedcmos, line 136: is it guranteed that priv going to be the same for retry requests? i am not sure what daos stores in it or how it manages it across requests

mchaarawi: yes i believe so. the priv is on the on-alloc request. so it will not change on retry afaik.

src/container/oid_iv.c
mchaarawi: for the oid iv, the on get returns a valid ptr always. see here:
https://github.com/daos-stack/daos/pull/15714/files#diff-9396dfee0370217c952706fe6a29b126a7ec9f73f9d23078a7b82659bc94df41R214

frostedcmos (approved):

wangdi1 (approved):
good catch!

src/container/oid_iv.c
liuxuezhao: looks like the "entry->current_req == priv" can never be true, as every time it will allocate a new one though oid_iv_ent_get(). So why need this check in the original place?

mchaarawi: I am not sure TBH.  @wangdi1 had added this one a while back. according to wangdi's comment, on an IV retry operation, we can re-enter the on-refresh call. but are you saying on a retry, we are calling on_get() again and so we have a new priv pointer allocated?

liuxuezhao: If the "retry" you mean here -
static int                                                                                              
_iv_op(struct ds_iv_ns *ns, struct ds_iv_key *key, d_sg_list_t *value,                                  
       crt_iv_sync_t *sync, unsigned int shortcut, bool retry, int opc)                                 
{                                                                                                       
>       int>    sleep_ms, total_ms = 0, rc;                                                             
                                                                                                        
>       if (ns->iv_stop)                                                                                
>       >       return -DER_SHUTDOWN;                                                                   
                                                                                                        
>       /* Sleep 1 ms before retry for IV_OID, sleep 1 sec before retry for other IV */                 
>       sleep_ms = key->class_id == IV_OID ? 1 : 1000;                                                  
retry:                                                                                                  
>       rc = iv_op_internal(ns, key, value, sync, shortcut, opc);  

Then yes, I think the retry of iv_op_internal() -> crt_iv_update()/crt_iv_fetch(), it will cause the oid_iv_ent_get() be called in the path and provide a new priv pointer.
So the check "entry->current_req != priv" will be always true.

@wangdi1 is above correct?

mchaarawi: in that case it's not the same request anyway. but yea maybe i misunderstood the original condition for this check.

liuxuezhao: the original code was added by @wangdi1 's PR https://github.com/daos-stack/daos/pull/13632 
the original code's "src" pointer is ds_cont_oid_alloc_handler() ->  cont_oid_alloc()'s sgl pointer, 
using src instead of priv looks more reasonable to me, because the src will not be changed when retry the IV operation, but priv will be changed.

mchaarawi: using src is just wrong.. it causes assertions. check the ticket.
src is a pointer on the stack. a totally different request can get the same src pointer value.

frostedcmos: hmm perhaps you may want to use "entry->current_req = src->sg_iovs[0].iov_buf"

while 'src' (temp iv value) can be on the stack, depending where iv calls it from, the underlying iov buffer will be provided by on_get(), so should remain valid until a corresponding on_put()

mchaarawi: yea that was my thought as well. thanks Alex, i'll post a follow on PR.

liuxuezhao: "while 'src' (temp iv value) can be on the stack, depending where iv calls it from, the underlying iov buffer will be provided by on_get(),"
I think it is not true. let's discuss it more on the new PR https://github.com/daos-stack/daos/pull/15727

PR #15727, opened 2025-01-13
DAOS-16931 container: fix the oid iv check for same request

mchaarawi:
on retry the priv pointer is always freed/allocated, so use the sgl iov pointer instead.

### Gatekeeper:

### DIFF
--- a/src/container/oid_iv.c
+++ b/src/container/oid_iv.c
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright 2017-2024 Intel Corporation.
+ * (C) Copyright 2017-2025 Intel Corporation.
  * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -32,7 +32,8 @@ struct oid_iv_entry {
 	struct oid_iv_range	rg;
 	/** protect the entry */
 	ABT_mutex		lock;
-	void                   *current_req;
+	d_rank_t                current_req_rank;
+	void                   *current_req_ptr;
 };
 
 /** Priv data in the iv layer */
@@ -86,12 +87,14 @@ oid_iv_ent_refresh(struct ds_iv_entry *iv_entry, struct ds_iv_key *key,
 
 	D_ASSERT(priv);
 	num_oids = priv->num_oids;
-	D_DEBUG(DB_MD, "%u: ON REFRESH: num_oids = %zu, REF_RC = " DF_RC "\n", dss_self_rank(),
-		num_oids, DP_RC(ref_rc));
 	D_ASSERT(num_oids != 0);
-
 	entry = iv_entry->iv_value.sg_iovs[0].iov_buf;
 	D_ASSERT(entry != NULL);
+
+	D_DEBUG(DB_MD,
+		"%u: ON REFRESH: num_oids = %zu, REF_RC = " DF_RC " ENTRY rank %u Ptr = %p\n",
+		dss_self_rank(), num_oids, DP_RC(ref_rc), entry->current_req_rank,
+		entry->current_req_ptr);
 
 	/** if iv op failed, just release the entry lock acquired in update */
 	if (ref_rc != 0)
@@ -130,17 +133,28 @@ oid_iv_ent_update(struct ds_iv_entry *ns_entry, struct ds_iv_key *iv_key,
 	d_rank_t		myrank = dss_self_rank();
 	int			rc;
 
+	if (src == NULL) {
+		D_DEBUG(DB_MD, "%u: ON UPDATE delete entry iv_entry %p\n", myrank, ns_entry);
+		ns_entry->iv_to_delete = 1;
+		return 0;
+	}
+
 	D_ASSERT(priv != NULL);
 	entry = ns_entry->iv_value.sg_iovs[0].iov_buf;
+	oids  = src->sg_iovs[0].iov_buf;
+
 	rc = ABT_mutex_trylock(entry->lock);
-	/** For retry requests, from _iv_op(), the lock may not be released in some cases. */
-	if (rc == ABT_ERR_MUTEX_LOCKED && entry->current_req != priv)
-		return -DER_BUSY;
+	if (rc == ABT_ERR_MUTEX_LOCKED) {
+		if (entry->current_req_rank == oids->req_rank &&
+		    entry->current_req_ptr == oids->req_ptr)
+			D_DEBUG(DB_MD, "%u: ON UPDATE - SAME REQ %p \n", myrank, src);
+		else
+			return -DER_BUSY;
+	}
 
-	entry->current_req = priv;
-	avail = &entry->rg;
-
-	oids = src->sg_iovs[0].iov_buf;
+	entry->current_req_rank = oids->req_rank;
+	entry->current_req_ptr  = oids->req_ptr;
+	avail                   = &entry->rg;
 
 	if (myrank == oids->req_rank)
 		num_oids = oids->req_num_oids;
@@ -209,8 +223,6 @@ oid_iv_ent_get(struct ds_iv_entry *entry, void **_priv)
 {
 	struct oid_iv_priv	*priv;
 
-	D_DEBUG(DB_MD, "%u: OID GET\n", dss_self_rank());
-
 	D_ALLOC_PTR(priv);
 	if (priv == NULL)
 		return -DER_NOMEM;
@@ -222,7 +234,6 @@ static void
 oid_iv_ent_put(struct ds_iv_entry *entry, void *priv)
 {
 	D_ASSERT(priv != NULL);
-	D_DEBUG(DB_MD, "%u: ON PUT\n", dss_self_rank());
 	D_FREE(priv);
 }
 
@@ -331,6 +342,7 @@ oid_iv_reserve(void *ns, uuid_t po_uuid, uuid_t co_uuid, uint64_t num_oids, d_sg
 	oids->num_oids = num_oids;
 	oids->req_rank     = dss_self_rank();
 	oids->req_num_oids = num_oids;
+	oids->req_ptr      = value;
 
 	rc = ds_iv_update(ns, &key, value, 0, CRT_IV_SYNC_NONE,
 			  CRT_IV_SYNC_BIDIRECTIONAL, true /* retry */);
--- a/src/container/srv_internal.h
+++ b/src/container/srv_internal.h
@@ -1,5 +1,6 @@
 /*
  * (C) Copyright 2016-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -110,6 +111,7 @@ struct oid_iv_range {
 	daos_size_t	num_oids;
 	daos_size_t     req_num_oids;
 	d_rank_t        req_rank;
+	void           *req_ptr;
 };
 
 /* Container IV structure */


src/container/oid_iv.c
```diff
@@ -134,12 +134,11 @@ oid_iv_ent_update(struct ds_iv_entry *ns_entry, struct ds_iv_key *iv_key,
 	entry = ns_entry->iv_value.sg_iovs[0].iov_buf;
 	rc = ABT_mutex_trylock(entry->lock);
 	/** For retry requests, from _iv_op(), the lock may not be released in some cases. */
-	if (rc == ABT_ERR_MUTEX_LOCKED && entry->current_req != priv)
+	if (rc == ABT_ERR_MUTEX_LOCKED && entry->current_req != src->sg_iovs[0].iov_buf)
```
frostedcmos, line 137: unless it's done in the layers above this, might want to check for src != NULL here. "iv invalidate" operation internally is an update with NULL iv_value

mchaarawi: we do not have an oid invalidate operation in this case. would this use case still be triggered somehow?

mchaarawi: it's interesting though because it should have failed on existing master in other parts of the on_update if src was NULL. since we dereference it elsewhere

liuxuezhao: src->sg_iovs[0].iov_buf is also in stack, it is cont_oid_alloc()'s "struct oid_iv_range rg".
not allocated by on_get() callback. I think this PR is wrong.
I'll check the ticket what's the problem later

frostedcmos (dismissed):

src/container/oid_iv.c
liuxuezhao: I checked the assertion in the ticket, not fully understood the reason. Is it suppose to be the case that two ds_cont_oid_alloc_handler() be called and happened with the same stack address? but cont_oid_alloc() will not release its stack before oid_iv_reserve() finish. So I am still not sure how can the assertion happen.

but one method maybe can try - can do the check like this PR, but also inside src/container/srv_target.c's cont_oid_alloc(), allocate the "struct oid_iv_range *rg" in heap rather than use stack variable.

liuxuezhao: I thought it again, looks to me that the above suggestion or the current code cannot fix the problem.

I can think one case possibly trigger the assertion:
the IV path is 
A  -> B -> C ... -> root
On A it calls oid_iv_reserve(), first time it get timeout (or other network error) and retry, both the original request and retried request sent to parent B, on B the timedout request possibly be processed finally so may cause the oid_iv_ent_refresh() be called twice (one for original timedout request and another for retired request), and the 2nd oid_iv_ent_refresh() found the lock is not locked so assert it.

To fix that case the method could be -
1. in oid_iv_ent_update() still check like original code "if (rc == ABT_ERR_MUTEX_LOCKED && entry->current_req != src", and 
2. in oid_iv_ent_refresh() check if the lock is locked first before calling ABT_mutex_unlock(). Or use a kind of flag rather than mutex.
what do you think?

liuxuezhao (changes requested):
leave -1 before we are clear about the root cause

src/container/oid_iv.c
mchaarawi: I have added debug before these patches were pushed and i think the case happens on some systems (boro) and not others like CI. maybe because of constrained memory or something.
the issue to me was clear and there was no timeout, but there were 2 separate requests. there was 1 request where it entered the on_update, forwarded the request, then another request on the same node that got into on_update, got ABT_ERR_MUTEX_LOCKED on the mutex, but the src pointer was the same of the other req. so it went ahead and also forward.
then on refresh was called twice in a row.

i started thinking first as you and had created a PR like here:
https://github.com/daos-stack/daos/pull/15711/files
this prevents the assertion, but i think that is wrong because this loop may never return in some cases:
	while (entry->in_flight)
		dss_sleep(1);

liuxuezhao: "but the src pointer was the same of the other req. so it went ahead and also forward."
the first request has not complete, then the cont_oid_alloc() function has not release its stack frame,
how can the 2nd request use the same src pointer? it calls same function cont_oid_alloc() but in different stack place right? how can the 2 requests use the same src pointer, this is what I understood.

mchaarawi: they can use the same src pointer value because the pointer value itself is not valid anymore after the on_update call and can be re-used by the OS. src is just an on stack value no? see here:
https://github.com/daos-stack/daos/blob/master/src/cart/crt_iv.c#L3154

mchaarawi: > the first request has not complete, then the cont_oid_alloc() function has not release its stack frame, how can the 2nd request use the same src pointer? it calls same function cont_oid_alloc() but in different stack place right? how can the 2 requests use the same src pointer, this is what I did not understand.

this src value is not coming from the cont_oid_alloc(). it's coming from the crt iv handler for on_update():
https://github.com/daos-stack/daos/blob/master/src/cart/crt_iv.c#L3154
which becomes invalid as soon as the on_update call is finished.

here is a pr that makes the assertion happen quicker on wolf (just using tcp, tmpfs):
https://github.com/daos-stack/daos/pull/15732
i test with 4 server, 8 engines. yaml file here: /home/mschaara/config/srv_tcp_dual.yml

run the multi threaded connect dfs test in this pr in a loop:
'''for n in {0..100}; do POOL_NVME_SIZE=0 dfs_test -n ~/config/daos.yml -u &> >(tee -a out); done'''

at sometime one of the engines will assert.

mchaarawi: looking at the log for the asserted engine, here is the debug output:

```
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9156] container DBUG src/container/oid_iv.c:308 oid_iv_alloc() 2: IV ALLOC: oids = 0x7faecddce730
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9156] container DBUG src/container/oid_iv.c:227 oid_iv_ent_get() 2: OID GET
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9156] container DBUG src/container/oid_iv.c:151 oid_iv_ent_update() 2: ON UPDATE src 0x7facd568e5c0; priv 0x7faecdff7af0; oids 0x7faecddce730; UNLOCKED Mutex
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9156] container DBUG src/container/oid_iv.c:165 oid_iv_ent_update() 2: ON UPDATE, num_oids = 32, priv = 0
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9156] container DBUG src/container/oid_iv.c:167 oid_iv_ent_update() 2: ENTRY NUM OIDS = 0, oid = 0
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9156] container DBUG src/container/oid_iv.c:212 oid_iv_ent_update() 2: IDs not available, FORWARD 64 oids
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9157] container DBUG src/container/oid_iv.c:308 oid_iv_alloc() 2: IV ALLOC: oids = 0x7faecde53290
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9157] container DBUG src/container/oid_iv.c:227 oid_iv_ent_get() 2: OID GET
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9157] container DBUG src/container/oid_iv.c:147 oid_iv_ent_update() 2: ON UPDATE src 0x7facd568e5c0; priv 0x7faecdf5df80; oids 0x7faecde53290; SAME req detected
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9157] container DBUG src/container/oid_iv.c:165 oid_iv_ent_update() 2: ON UPDATE, num_oids = 32, priv = 0
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9157] container DBUG src/container/oid_iv.c:167 oid_iv_ent_update() 2: ENTRY NUM OIDS = 0, oid = 0
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9157] container DBUG src/container/oid_iv.c:212 oid_iv_ent_update() 2: IDs not available, FORWARD 64 oids
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9160] container DBUG src/container/oid_iv.c:89 oid_iv_ent_refresh() 2: ON REFRESH: num_oids = 32, REF_RC = DER_SUCCESS(0): 'Success'
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9160] container DBUG src/container/oid_iv.c:113 oid_iv_ent_refresh() 2: ON REFRESH 0/32 avail 32/32
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9160] container DBUG src/container/oid_iv.c:240 oid_iv_ent_put() 2: ON PUT
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9161] container DBUG src/container/oid_iv.c:89 oid_iv_ent_refresh() 2: ON REFRESH: num_oids = 32, REF_RC = DER_SUCCESS(0): 'Success'
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9161] container DBUG src/container/oid_iv.c:113 oid_iv_ent_refresh() 2: ON REFRESH 96/32 avail 128/32
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:46 print_backtrace() *** Process 38266 (daos_engine) received signal 6 (Aborted) ***
01/14-21:12:19.35 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:50 print_backtrace() Associated errno: Success (0)
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /lib64/libc.so.6(+0x54df0) [0x7faee6e54df0]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /lib64/libc.so.6(+0xa154c) [0x7faee6ea154c]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /lib64/libc.so.6(raise+0x16) [0x7faee6e54d46]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /lib64/libc.so.6(abort+0xd3) [0x7faee6e287f3]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /lib64/libc.so.6(+0x2871b) [0x7faee6e2871b]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /lib64/libc.so.6(+0x4dce6) [0x7faee6e4dce6]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/../prereq/debug/argobots/lib/libabt.so.1(ABT_mutex_unlock+0x331) [0x7faee756d141]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/../lib64/daos_srv/libcont.so(+0x59e8b) [0x7faeddfa3e8b]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/daos_engine(+0x5b418) [0x55e6e219d418]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/daos_engine(+0x60297) [0x55e6e21a2297]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/daos_engine(+0x60a0e) [0x55e6e21a2a0e]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/../lib64/libcart.so.4(+0xbbf66) [0x7faee7a95f66]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/../lib64/libcart.so.4(+0xd171a) [0x7faee7aab71a]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/../lib64/libcart.so.4(+0xd27f5) [0x7faee7aac7f5]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/../prereq/debug/argobots/lib/libabt.so.1(+0x2bd2c) [0x7faee7589d2c]
01/14-21:12:19.36 wolf-167 DAOS[38266/0/9161] misc ERR  src/gurt/signals.c:87 print_backtrace() Frame /home/mschaara/install/daos/bin/../prereq/debug/argobots/lib/libabt.so.1(+0x16849) [0x7faee7574849]
```

as you can tell, there are 2 requests that happen at the same time (get oid_iv_alloc). both are for 32 oids (so it means these requests are in an intermediate rank). So rankk B in this case:
X->B->root
Y->B->root
everything is different except the src pointer value. OID IV think they are the same request, and let's both in. the on-refresh call for the second asserts.

mchaarawi: so after looking into this more, i guess you are right that this iv_value is coming from the sgl in the cont_oid_alloc() and not the cart iv handler for on_get (this is not invoked in this case).
but I guess still confused why this is happening. the patch that landed on master that wangdi had approved that used priv probably need to be reverted. but the segault issue is still there. i will debug more to see why this value is being seen on 2 requests.

mchaarawi: @liuxuezhao please disregard my last comment. i was actually right that since this a middle node (not the originating one); the iv_value src pointer is coming from crt_iv_update_inline_hdlr. so this one cannot be saved to detect the same request on a middle node.
so as far as i can tell, we need a different way to detect the same req

mchaarawi: @liuxuezhao i made an update that i believe would work. please check and let me know if you see something wrong.

src/container/oid_iv.c
```diff
@@ -331,6 +342,7 @@ oid_iv_reserve(void *ns, uuid_t po_uuid, uuid_t co_uuid, uint64_t num_oids, d_sg
 	oids->num_oids = num_oids;
 	oids->req_rank     = dss_self_rank();
 	oids->req_num_oids = num_oids;
+	oids->req_ptr      = value;
```
frostedcmos, line 345: the 'value' passed here is from https://github.com/daos-stack/daos/blob/0be2d10c2917257b3d1ca6d8af9fdfc0891d9aa2/src/container/srv_target.c#L2357 and it is an addr of a variable on a stack.

as such req_ptr can end up be the same for different iv values, which is not what we want.
i think instead oid->req_ptr should store 'oids' as the iov_buf will be different for different iv values and combined with req_rank should uniquely identify the 'iv transaction'

mchaarawi: this one is synchronous.. you cannot reuse a stack variable unless the call exits, right?

frostedcmos: well, not sure what you mean by synchronous here. 
If i am understanding this whole thing correctly:
you are storing it here in oid (iv value contents) and then those contents will possibly get shipped around to different engine and examined asynchronously there (e.g. in oid_iv_end_update() on that engine)

some intermediate engine can then get 2 oid_iv_end_update() calls for different iv values from the same rank, but those iv value contents will have same oid->req_rank and same oid->req_ptr (as req_ptr will both contain &sgl on a stack on when it was created on that remote engine). 

Where as if you stored "sg_iovs[0].iov_buf" during reserve then it would be unique on that rank for each iv value that is being worked on.

frostedcmos: After discussing over slack with @mchaarawi  I understand now that the remote stack for iv update will remain intact ensuring another iv_update happening from same rank ending up using a different value. 
LGTM as such.

src/container/oid_iv.c
liuxuezhao: thanks, I guess you are right. Although I may need to check some details later, within this week.

src/container/oid_iv.c
```diff
@@ -130,17 +133,28 @@ oid_iv_ent_update(struct ds_iv_entry *ns_entry, struct ds_iv_key *iv_key,
 	d_rank_t		myrank = dss_self_rank();
 	int			rc;
 
+	if (src == NULL) {
```
liuxuezhao, line 136: did you see such case that with NULL src in oid_iv_ent_update() in test? may come from oid_iv_invalidate()

liuxuezhao (approved):

liuxuezhao:
The patch looks fine to me. But I suspect it possibly not helpful to fix the original assertion that the lock is held in oid_iv_ent_refresh().
For example if the request A1 send out, for some reason it retried A2 (maybe due to A1 timeout), but later both A1 and A2's request replied and both triggered oid_iv_ent_refresh(). Then the 2nd oid_iv_ent_refresh will triger the assertion.

src/container/oid_iv.c
```diff
@@ -130,17 +133,28 @@ oid_iv_ent_update(struct ds_iv_entry *ns_entry, struct ds_iv_key *iv_key,
...
 
-	entry->current_req = priv;
-	avail = &entry->rg;
+	rc = ABT_mutex_trylock(entry->lock);
+	if (rc == ABT_ERR_MUTEX_LOCKED) {
+		if (entry->current_req_rank == oids->req_rank &&
+		    entry->current_req_ptr == oids->req_ptr)
+			D_DEBUG(DB_MD, "%u: ON UPDATE - SAME REQ %p \n", myrank, src);
+		else
+			return -DER_BUSY;
```
liuxuezhao, line 152: On the middle node, if both child X, Y forward request to A -
X -> A -> P
Y -> A -> P
Assume X's request first arrive and goes up to P, then Y's request arrives
In that case looks should make it wait first request's reply, but current code will return EBUSY to Y, and Y will retry the req. This causes some unnecessary RPC + retry, it is worse if there are many layers between X and A.
We probably should think more for it and refine it later.

mchaarawi: i thought with ebusy, the retry is just local on A.. so it will be:
Y->A(busy spin)->P
or am i wrong?

liuxuezhao: I think that is incorrect.
If the iv update initiated on A, then will retry on A.
But in the example case, the IV update initiated on Y, then A returns EBUSY will cause it reply X's request as EBUSY, and then on X will retry the IV update.
The retry only on initiated node.

mchaarawi: whatever the flow is, This PR does not change the behavior for this. the same thing happens before, right?

src/container/oid_iv.c
mchaarawi: no, but i thought we can just add it for now to be sure

src/container/oid_iv.c
liuxuezhao: right, I think can land it first and consider how to refine it later.

frostedcmos (approved):

src/container/oid_iv.c
mchaarawi: > I think that is incorrect. If the iv update initiated on A, then will retry on A. But in the example case, the IV update initiated on Y, then A returns EBUSY will cause it reply X's request as EBUSY, and then on X will retry the IV update that will be forwarded to A again. So if there are many layers between Y and A, it will need to go through every layers for reply + retry on Y + forward The retry only on initiated node.

if that is the case, yes it's very bad.. it's better to block on A rather than send the ebusy back. sigh. we should have caught this when this ebusy was introduced.

mchaarawi:
> The patch looks fine to me. But I suspect it possibly cannot fully fix the original assertion that the lock is held in oid_iv_ent_refresh(). For example if the request A1 send out, for some reason it retried A2 (maybe due to A1 timeout), but later both A1 and A2's request replied and both triggered oid_iv_ent_refresh(). Then the 2nd oid_iv_ent_refresh will triger the assertion.

discussed in slack. the timeout reply should not invoke an on-refresh.. i will create a ticket to address the massive RPC caused by EBUSY for 2.8

PR #15731, opened 2025-01-14
DAOS-6611 control: Enable dmg pool exclude,drain,reint on multiple ranks

tanabarr:
Enable --ranks option on excluded drain and reintegrate dmg pool cmds.
To reduce change in dRPC code, iterate through ranks in request from
within the control API layer which reduces complexity at the cost of
per-rank (rather than batched) MS-gRPC and dRPC requests. Uses common
request and response type for pool-ranks operations where possible.
Reuse common display format for exclude/drain/reint system and pool
cmds. Improvement of workflow and unit test coverage.

### Gatekeeper:


knard38 (commented):
Is it planned to have one or more functional tests using these new features ?
Do we have some protocol version information to update as this new feature will change the RPC content and not just extend it ?

src/control/lib/control/pool.go
```diff
@@ -744,80 +744,176 @@ func PoolGetProp(ctx context.Context, rpcClient UnaryInvoker, req *PoolGetPropRe
...
+	TargetIdx []uint32        `json:"target_idx"`
+	Force     bool            `json:"force"`
 }
 
-// ExcludeResp has no other parameters other than success/failure for now.
+// PoolRanksResp struct contains response from operation on multiple pool-ranks.
+type PoolRanksResp struct {
+	Status         int32           `json:"status"`
+	ID             string          `json:"id"`
+	FailedRank     ranklist.Rank   `json:"failed_rank"`
```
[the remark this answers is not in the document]
tanabarr: the design chosen is to iterate through ranklist in the engine dRPC handler to update each rank's state. the first failure will break out of the loop and return

tanabarr:
> Is it planned to have one or more functional tests using these new features ? Do we have some protocol version information to update as this new feature will change the RPC content and not just extend it ?

I imagine there will be more than one functional test added, I think @rpadma2 will be adding these. The protocol changes are dmg <-> daos_server <-> engine which at the moment all have to be at the same version IIUC, am I understanding that correct @mjmac @kjacque @NiuYawei @liw or is there some compatibility related work to be done?

src/control/cmd/dmg/pool.go
```diff
@@ -531,72 +531,104 @@ func (cmd *poolEvictCmd) Execute(args []string) error {
...
 	}
 
-	err := control.PoolDrain(cmd.MustLogCtx(), cmd.ctlInvoker, req)
+	resp, err := control.PoolDrain(cmd.MustLogCtx(), cmd.ctlInvoker, req)
+
+	if cmd.JSONOutputEnabled() {
+		return cmd.OutputJSON(resp, err)
+	}
+
+	cmd.Debugf("%T: %+v, %T: %+v", req, req, resp, resp)
```
mjmac, line 611: Are these debug lines necessary? I think you should already have debug printing from the control library.

mjmac:
> The protocol changes are dmg <-> daos_server <-> engine which at the moment all have to be at the same version IIUC, am I understanding that correct @mjmac @kjacque @NiuYawei @liw or is there some compatibility related work to be done?

That's correct. There's no interoperability concern here because all of those components must be the same version. At some point, when rolling upgrades are supported, we'll need to ensure that dmg can always talk to the -1/+1 versions of daos_server, though.

src/control/lib/control/pool.go
```diff
@@ -831,9 +927,9 @@ type PoolExtendReq struct {
 // This should automatically start the rebalance process.
 // Returns an error (including any DER code from DAOS).
 func PoolExtend(ctx context.Context, rpcClient UnaryInvoker, req *PoolExtendReq) error {
```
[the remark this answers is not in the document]
knard38: Sounds reasonable to me.

knard38 (commented):
LGTM

src/control/cmd/dmg/pool.go
tanabarr: I wasn't getting any debug for response from `interceptors.go` but  maybe I can just move this to trace. this also prints the native types rather than proto which is what we get from the control library from what I can see

tanabarr:
@mjmac @kjacque can I have a quick approval on the approach in the last commit please?

mjmac:
It's kind of confusing to have `PoolRanksResult` and `PoolRanksResp` types which are very similar. Seems like it would be better to figure out a way to reduce the redundancy.

Also, while I see that you have added a `FailedRanks` slice, the intent of my feedback was to move away from the all-or-nothing response in order to accommodate a more flexible design that will allow for individual rank results.

The current implementation has made some pretty significant changes in the drpc handlers to populate the multi-rank response, but this seems unnecessary to me. Why do we need to do all of that down in the engine? The current approach seems to be trying to conceptualize a multi-rank request as a single pool operation that happens to be performed on multiple ranks. This is not really what's happening, and I think it's a mistake to look at and present it this way. What's really happening is that we're making a UX improvement to allow the admin to call a single command (or RPC) in order to perform multiple independent operations across several ranks.

A less invasive (and therefore less risky) approach would be to keep most of the changes in the gRPC handlers at the daos_server level. So the request/response messages from control<->daos_server would handle multiple ranks, but the messages from daos_server<->daos_engine would remain the same: 1 req/resp per rank operation. Then you take care of aggregating the responses in the gRPC handler before sending the composite response back to the control client.

This does imply multiple dRPC invocations for a single gRPC request. I think that's fine. In the future, if we do happen change the pool service to support concurrent operations on multiple ranks, then we won't need to change the gRPC messages or UI. I don't think that's likely to happen, though. Thinking down the road in another direction, keeping the iteration logic up in daos_server may enable other UX improvements, such as an automatic reintegrate command (e.g. `dmg pool reintegrate --auto`) instead of requiring the admin to go and figure out which ranks were excluded and need to be reintegrated.

So, just to be explicitly clear, my recommendation for how to do this is:
  * Add a new response message that is a list of results, not a single message that contains lists of ranks and a single result -- this will give you the flexibility to describe each result, and at the UI level, the results can be coalesced as we do with the various scan operations
  * Back out most or all of the changes at the dRPC layer and keep the logic for iterating over request ranks in the daos_server gRPC handlers

tanabarr:
> It's kind of confusing to have `PoolRanksResult` and `PoolRanksResp` types which are very similar. Seems like it would be better to figure out a way to reduce the redundancy.
> 
> Also, while I see that you have added a `FailedRanks` slice, the intent of my feedback was to move away from the all-or-nothing response in order to accommodate a more flexible design that will allow for individual rank results.
> 
> The current implementation has made some pretty significant changes in the drpc handlers to populate the multi-rank response, but this seems unnecessary to me. Why do we need to do all of that down in the engine? The current approach seems to be trying to conceptualize a multi-rank request as a single pool operation that happens to be performed on multiple ranks. This is not really what's happening, and I think it's a mistake to look at and present it this way. What's really happening is that we're making a UX improvement to allow the admin to call a single command (or RPC) in order to perform multiple independent operations across several ranks.
> 
> A less invasive (and therefore less risky) approach would be to keep most of the changes in the gRPC handlers at the daos_server level. So the request/response messages from control<->daos_server would handle multiple ranks, but the messages from daos_server<->daos_engine would remain the same: 1 req/resp per rank operation. Then you take care of aggregating the responses in the gRPC handler before sending the composite response back to the control client.
> 
> This does imply multiple dRPC invocations for a single gRPC request. I think that's fine. In the future, if we do happen change the pool service to support concurrent operations on multiple ranks, then we won't need to change the gRPC messages or UI. I don't think that's likely to happen, though. Thinking down the road in another direction, keeping the iteration logic up in daos_server may enable other UX improvements, such as an automatic reintegrate command (e.g. `dmg pool reintegrate --auto`) instead of requiring the admin to go and figure out which ranks were excluded and need to be reintegrated.
> 
> So, just to be explicitly clear, my recommendation for how to do this is:
> 
> * Add a new response message that is a list of results, not a single message that contains lists of ranks and a single result -- this will give you the flexibility to describe each result, and at the UI level, the results can be coalesced as we do with the various scan operations
> * Back out most or all of the changes at the dRPC layer and keep the logic for iterating over request ranks in the daos_server gRPC handlers

Thanks for being explicit about your request, I did find your initial feedback rather confusing.

I think by not batching the requests over dRPC we are missing a trick in terms of reducing potential XStream and network traffic load but I will defer to your judgement. The suggested simplifications will definitely make the change smaller and is similar to the solution for system drain/reintegrate.

mjmac:
> I think by not batching the requests over dRPC we are missing a trick in terms of reducing potential XStream and network traffic load

In my mind that's a future optimization opportunity, and low priority compared to the usability improvements that can be made now. If that change does get made down the road, then it would ideally result in minimal or no changes to the UI or control<->daos_server messages.

Another thought that occurs to me with regard to the composite response message is that it establishes a pattern for more complex operations potentially involving multiple steps across multiple ranks. For example, an operation that drains a rank's targets and then reintegrates it, as a way of forcing a resilvering. Or rolling upgrades, where a rank is drained and then restarted, then reintegrated, etc.

I'm not suggesting that your solution now should be over-engineered to try and support any and all hypothetical future use cases, but I think that having those possibilities in mind may inform your decisions about how to implement the solution now.

src/control/lib/control/pool.go
```diff
@@ -744,80 +744,183 @@ func PoolGetProp(ctx context.Context, rpcClient UnaryInvoker, req *PoolGetPropRe
...
+		// Add one result for failed ranks. Note same status applies to all failed ranks.
+		results = append(results, &PoolRanksResult{
+			ID:     poolID,
+			Ranks:  resp.FailedRanks,
+			Status: resp.Status,
+			Msg:    daos.Status(resp.Status).Error(),
+		})
+	} else {
+		if len(resp.FailedRanks) > 0 {
+			return nil, errors.New("failed rank returned with zero status")
```
kjacque, line 841: IMO we should handle these seemingly-inconsistent responses more gracefully. It's a surprising output but seems more like something we should log as strange (maybe NOTICE level?) and return an appropriate result anyway. An error like this isn't really actionable for the user, since it indicates an engine bug in writing the response. The operation may have succeeded or failed, and the lists of successful and failed ranks seem potentially still useful.

src/control/lib/control/pool.go
```diff
@@ -744,80 +744,183 @@ func PoolGetProp(ctx context.Context, rpcClient UnaryInvoker, req *PoolGetPropRe
...
-// PoolDrain will set a pool target for a specific rank in to the drain state which should
-// automatically start the rebuildiing process. Returns an error (including any DER code from DAOS).
-func PoolDrain(ctx context.Context, rpcClient UnaryInvoker, req *PoolDrainReq) error {
-	pbReq := &mgmtpb.PoolDrainReq{
-		Sys:       req.getSystem(rpcClient),
-		Id:        req.ID,
-		Rank:      req.Rank.Uint32(),
-		TargetIdx: req.TargetIdx,
+// PoolDrain will set pool targets on specified ranks in to the drain state which should
+// automatically start the rebuildiing process. Returns pool-ranks response and error.
```
kjacque, line 893: spelling nit
```suggestion
// automatically start the rebuilding process. Returns pool-ranks response and error.
```

src/control/server/faults.go
```diff
@@ -110,8 +111,8 @@ func FaultPoolInvalidRanks(invalid []ranklist.Rank) *fault.Fault {
 
 	return serverFault(
 		code.ServerPoolInvalidRanks,
-		fmt.Sprintf("pool request contains invalid ranks: %s", strings.Join(rs, ",")),
-		"retry the request with a valid set of ranks",
+		fmt.Sprintf("pool request contains ranks with invalid state: %s", strings.Join(rs, ",")),
+		"retry the request with a set of ranks that have valid states",
```
kjacque, line 115: Feels a little weird and potentially confusing to lump in "bad state" (is just just excluded ranks?) with "nonexistent."

src/tests/ftest/deployment/network_failure.py
```diff
@@ -234,7 +235,7 @@ def verify_network_failure(self, ior_namespace, container_namespace):
         # 7. Call dmg pool reintegrate one rank at a time to enable all ranks.
         self.log_step("Reintegrate one rank at a time to enable all ranks.")
         for disabled_rank in disabled_ranks:
-            self.pool.reintegrate(rank=disabled_rank)
+            self.pool.reintegrate(ranks=disabled_rank)
```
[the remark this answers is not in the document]
daltonbohning: I think it depends on what the test is verifying but IMO doing multiple ranks should be preferred.
@shimizukko This test was added for https://daosio.atlassian.net/browse/DAOS-10003.
The description specifically says "Call dmg pool reintegrate --rank=<rank> one rank at a time".
But the test plan doesn't say "one rank at a time": https://daosio.atlassian.net/wiki/spaces/DC/pages/11098619936/Fault+management+Test+Plan
So any reason we shouldn't just reintegrate all at once?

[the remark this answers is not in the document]
rpadma2: @daltonbohning / @shimizukko : I am creating some tests under OSA tests to verify multiple ranks reintegration, drain. It is planned as part of my PR: https://github.com/daos-stack/daos/pull/15788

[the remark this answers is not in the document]
rpadma2: All I think --rank parameter is replaced --ranks. I think it accepts both single or multiple ranks. @tanabarr : please confirm.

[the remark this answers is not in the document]
tanabarr: correct, single rank is a valid rankset character string, as is a range or comma-separated list

src/control/lib/control/pool.go
```diff
@@ -1,5 +1,5 @@
 //
-// (C) Copyright 2020-2024 Intel Corporation.
+//
```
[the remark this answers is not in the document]
tanabarr: correct I will fix

src/mgmt/srv_drpc.c
```diff
@@ -751,18 +751,90 @@ pool_change_target_state(char *id, d_rank_list_t *svc_ranks, size_t n_target_idx
...
+	if (svc_rl == NULL)
+		D_GOTO(out_tgt, rc = -DER_NOMEM);
+
+	success_rl = d_rank_list_alloc(n_ranks);
+	if (success_rl == NULL)
+		D_GOTO(out_svc, rc = -DER_NOMEM);
+
+	failed_rl = d_rank_list_alloc(n_ranks);
+	if (failed_rl == NULL)
+		D_GOTO(out_success, rc = -DER_NOMEM);
```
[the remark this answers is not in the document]
tanabarr: changes to this file will be reverted

src/mgmt/srv_drpc.c
```diff
@@ -751,18 +751,90 @@ pool_change_target_state(char *id, d_rank_list_t *svc_ranks, size_t n_target_idx
...
+
+	for (i = 0; i < n_ranks; i++) {
+		D_INFO("change_target_state: %d\n", ranks[i]);
+		rc = pool_change_target_state(id, svc_rl, n_target_idx, target_idx, ranks[i], state,
+					      scm_size, nvme_size, meta_size, skip_rf_check);
```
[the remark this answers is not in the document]
tanabarr: changes to this file will be reverted

src/mgmt/srv_drpc.c
```diff
@@ -751,18 +751,90 @@ pool_change_target_state(char *id, d_rank_list_t *svc_ranks, size_t n_target_idx
...
+		if (rc != 0) {
+			D_ASSERT(j < failed_rl->rl_nr);
+			failed_rl->rl_ranks[j++] = ranks[i];
+			D_INFO("nr failures: %d (%d)\n", j, ranks[i]);
```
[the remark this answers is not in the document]
tanabarr: changes to this file will be reverted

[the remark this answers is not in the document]
tanabarr: (this was just debug that I forgot to remove)

src/mgmt/srv_drpc.c
```diff
@@ -751,18 +751,90 @@ pool_change_target_state(char *id, d_rank_list_t *svc_ranks, size_t n_target_idx
...
+			continue;
+		}
+
+		D_ASSERT(j < success_rl->rl_nr);
+		success_rl->rl_ranks[j++] = ranks[i];
+		D_INFO("nr successe: %d (%d)\n", j, ranks[i]);
```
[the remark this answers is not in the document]
tanabarr: changes to this file will be reverted, again inadvertently remaining debug

src/mgmt/srv_drpc.c
```diff
@@ -751,18 +751,90 @@ pool_change_target_state(char *id, d_rank_list_t *svc_ranks, size_t n_target_idx
...
+	}
+
+	rc = rank_list_to_uint32_array(failed_rl, &resp->failed_ranks, &resp->n_failed_ranks);
+	if (rc != 0) {
+		D_ERROR("Failed to convert failed rank list: rc=%d\n", rc);
+		D_GOTO(out_failed_resp, rc = -DER_INVAL);
```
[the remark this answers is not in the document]
tanabarr: should have, changes to this file will be reverted

src/mgmt/srv_drpc.c
```diff
@@ -751,18 +751,90 @@ pool_change_target_state(char *id, d_rank_list_t *svc_ranks, size_t n_target_idx
...
+		}
+
+		D_ASSERT(j < success_rl->rl_nr);
+		success_rl->rl_ranks[j++] = ranks[i];
+		D_INFO("nr successe: %d (%d)\n", j, ranks[i]);
+	}
+
+	rc = rank_list_to_uint32_array(failed_rl, &resp->failed_ranks, &resp->n_failed_ranks);
+	if (rc != 0) {
+		D_ERROR("Failed to convert failed rank list: rc=%d\n", rc);
```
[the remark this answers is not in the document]
tanabarr: correct, changes to this file will be reverted

src/control/lib/control/pool.go
```diff
@@ -744,80 +744,183 @@ func PoolGetProp(ctx context.Context, rpcClient UnaryInvoker, req *PoolGetPropRe
 	return resp, nil
 }
 
-// PoolExcludeReq struct contains request
-type PoolExcludeReq struct {
+// PoolRanksResult describes the result of an OSA operation on a pool's ranks.
```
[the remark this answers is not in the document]
tanabarr: removed

src/control/lib/control/pool.go
```diff
@@ -744,80 +744,183 @@ func PoolGetProp(ctx context.Context, rpcClient UnaryInvoker, req *PoolGetPropRe
...
+	if resp.Status == int32(daos.Success) {
+		return nil
+	}
+
+	err := daos.Status(resp.Status)
+
+	if len(resp.FailedRanks) > 0 {
+		rs := ranklist.RankSetFromRanks(resp.InitialRanks)
+		return errors.Wrapf(err, "pool %s ranks %s", resp.ID, rs.String())
+	}
```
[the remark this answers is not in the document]
tanabarr: will be addressed in refactor

src/control/lib/control/pool.go
tanabarr: will be addressed in refactor

src/control/lib/control/pool.go
tanabarr: done

src/control/server/faults.go
tanabarr: agreed , will fix

knard38 (dismissed):
OK from my side

mjmac (dismissed):
Looks much better. Lots of duplicate code eliminated. Would be nice to further reduce that duplication in dmg as well, but I wouldn't block on it. The main thing that I would like to see addressed in this PR or in a follow-on is the debug statements that have snuck back into dmg.

src/control/server/mgmt_system.go
```diff
@@ -1147,141 +1142,61 @@ func (svc *mgmtSvc) getPoolsRanks(ranks *ranklist.RankSet) ([]string, poolRanksM
...
-		return pbResp.Status, daos.Status(pbResp.Status)
-	}
-
-	svc.log.Tracef("pool-reint triggered from system-reint: %+v (req: %+v)", pbResp, pbReq)
-
-	return int32(daos.Success), nil
+	return resps, nil
 }
 
 // SystemDrain marks specified ranks on all pools as being in a drain state.
```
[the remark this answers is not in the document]
tanabarr: Will address in follow-up

src/control/cmd/dmg/pool.go
```diff
@@ -36,10 +36,10 @@ type PoolCmd struct {
...
 	Evict        poolEvictCmd        `command:"evict" description:"Evict all pool connections to a DAOS pool"`
 	List         poolListCmd         `command:"list" alias:"ls" description:"List DAOS pools"`
-	Extend       poolExtendCmd       `command:"extend" description:"Extend a DAOS pool to include new ranks."`
-	Exclude      poolExcludeCmd      `command:"exclude" description:"Exclude targets from a rank"`
-	Drain        poolDrainCmd        `command:"drain" description:"Drain targets from a rank"`
-	Reintegrate  poolReintegrateCmd  `command:"reintegrate" alias:"reint" description:"Reintegrate targets for a rank"`
+	Extend       poolExtendCmd       `command:"extend" description:"Extend a DAOS pool to include new ranks"`
+	Exclude      poolExcludeCmd      `command:"exclude" description:"Exclude targets from a set of ranks"`
+	Drain        poolDrainCmd        `command:"drain" description:"Drain targets from a set of ranks"`
+	Reintegrate  poolReintegrateCmd  `command:"reintegrate" alias:"reint" description:"Reintegrate targets for a set of rank"`
```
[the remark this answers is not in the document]
tanabarr: Will address in follow-up

mjmac:
I'd like to give @kjacque time to add any feedback. If you're good to go with this Kris, just let us know...

src/control/cmd/dmg/pool.go
```diff
@@ -532,72 +532,96 @@ func (cmd *poolEvictCmd) Execute(args []string) error {
...
-		msg = errors.WithMessage(err, "failed").Error()
+		return err
+	}
+
+	if cmd.JSONOutputEnabled() {
+		return cmd.OutputJSON(resp, resp.Errors())
 	}
 
-	cmd.Infof("Exclude command %s\n", msg)
+	cmd.Debugf("%T: %+v, %T: %+v", req, req, resp.Results, resp.Results)
```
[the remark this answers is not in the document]
tanabarr: Will address in follow-up

src/control/cmd/dmg/pool_test.go
```diff
@@ -614,13 +614,14 @@ func TestPoolCommands(t *testing.T) {
 			}, " "),
 			nil,
 		},
+		// Exclude testing with multiple ranks is verified at the control API layer.
```
kjacque, line 617: Does it hurt anything to verify it here as well? There seems to be value to me in ensuring rank lists are parsed correctly at this level.

src/control/cmd/dmg/pretty/pool.go
```diff
@@ -195,3 +197,79 @@ func PrintPoolProperties(poolID string, out io.Writer, properties ...*daos.PoolP
...
+
+		for _, res := range resp.Results {
+			msg := "" // Key used for successful ranks,
+			if res.Errored {
+				msg = res.Msg
+			}
+
+			if _, exists := seenRanks[res.Rank]; exists {
+				return errors.Errorf("multiple PoolRankResults for rank %d", res.Rank)
+			}
```
kjacque, line 236: This is a type of case where I feel like returning an error is incorrect. If it's duplicated it could be logged and skipped (instead of ensuring the user doesn't get to see the rest of the command output).

src/control/lib/control/pool.go
```diff
@@ -776,94 +777,73 @@ func (resp *PoolRanksResp) Errors() error {
 	if resp == nil {
 		return errors.Errorf("nil %T", resp)
 	}
-	if resp.Status == int32(daos.Success) {
-		return nil
+	if resp.ID == "" {
+		return errors.New("empty id in response")
```
kjacque, line 781: Similar comment about erroring on the response returned to us from the engine. If the engine returns something bad, I think we should still make an attempt to return/display as much info as possible to the user. I.e. if the ID is unknown, we should still try to return the rank count errors for an "unknown pool".

src/control/lib/control/pool.go
```diff
@@ -776,94 +777,73 @@ func (resp *PoolRanksResp) Errors() error {
...
-		if len(resp.SuccessRanks) == 0 {
-			// Expected that at least one result will be generated for each pool.
-			return nil, errors.New("no ranks were operated on")
+	results := []*PoolRankResult{}
+	for _, rank := range req.Ranks {
+		result, err := poolRankOp(ctx, rpcClient, req, rank)
+		if err != nil {
+			return nil, errors.Wrapf(err, "pool %s rank %d", req.ID, rank)
 		}
+		results = append(results, result)
```
kjacque, line 820: I think this is a case where we'd want to save the error in the results, so the caller can see if some ranks succeeded?

src/control/lib/control/pool.go
```diff
@@ -909,18 +892,23 @@ func PoolDrain(ctx context.Context, rpcClient UnaryInvoker, req *PoolRanksReq) (
 		return nil, err
 	}
 	if msErr := ur.getMSError(); err != nil {
```
kjacque, line 894: Found one actual bug. I think this issue needs to be fixed before landing.

```suggestion
	if msErr := ur.getMSError(); msErr != nil {
```

src/control/server/mgmt_system.go
```diff
@@ -1264,20 +1208,38 @@ func (svc *mgmtSvc) SystemDrain(ctx context.Context, req *mgmtpb.SystemDrainReq)
...
-		Reint:   req.Reint,
-		Results: results,
-	}, nil
+	if len(resps) == 0 {
+		return nil, errors.New("no pool-ranks responses received")
+	}
+	if len(resps) != len(poolIDs) {
+		return nil, errors.Errorf("unexpected number of pool-ranks responses received, "+
+			"want %d got %d", len(poolIDs), len(resps))
+	}
```
kjacque, line 1234: IMO should log and return the surprising response anyway

kjacque (changes requested):
One tiny issue that is blocking. I think everything else can be addressed in follow-ons.

src/control/cmd/dmg/pool_test.go
tanabarr: currently this would require significant reworking of the command_test.go code because we have multiple layers of Control API calls now. I don't want to do that as part of this PR

src/control/server/mgmt_system.go
tanabarr: done

src/control/lib/control/pool.go
tanabarr: done

src/control/lib/control/pool.go
tanabarr: The reason I haven't done that is because most of the failure causes here are  "should not happen" rather than sometimes expected. Happy to change though. Done.

src/control/lib/control/pool.go
```diff
@@ -815,7 +816,12 @@ func getPoolRanksResp(ctx context.Context, rpcClient UnaryInvoker, req *PoolRank
 	for _, rank := range req.Ranks {
 		result, err := poolRankOp(ctx, rpcClient, req, rank)
 		if err != nil {
-			return nil, errors.Wrapf(err, "pool %s rank %d", req.ID, rank)
+			rpcClient.Debugf("pool %s rank %d failed: %s", req.ID, rank, err.Error())
```
kjacque, line 819: Could log this at Error, since it is documenting a failure.

src/control/cmd/dmg/pool_test.go
kjacque: Oh, I didn't realize. Yeah, if it requires some refactoring, let's leave it for a future improvement.

kjacque (approved):

src/control/server/mgmt_system.go
kjacque: Doesn't look like this one made it into the update. I think it's fine to include in a follow-on though.

knard38 (approved):

tanabarr:
from what I can see NLT failure looks unrelated

tanabarr:
No NLT failure on most recent run

daltonbohning:
https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-15731/22/tests
online_rebuild_mdtest.py failure is

https://daosio.atlassian.net/browse/DAOS-17161

tanabarr:
thanks @daltonbohning

PR #15736, opened 2025-01-15
DAOS-16115 2.6RC1: Jar file label need to be updated to 2.6. Still re…

jiafuzha:
…ferring 2.4

### DIFF
--- a/src/client/java/daos-java/pom.xml
+++ b/src/client/java/daos-java/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>io.daos</groupId>
     <artifactId>daos-java-root</artifactId>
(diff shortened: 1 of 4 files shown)


jolivier23 (approved):

jolivier23:
title should probably use something like "client" as the label rather than 2.6RC1

jolivier23:
> title should probably use something like "client" as the label rather than 2.6RC1

also, could put the 2nd sentence in the body

saurabhtandan (changes requested):
This PR is for master. So in this case the version should be 2.7 instead of 2.6.
For 2.6 there need to be a separate branch on top of `release/2.6` branch.

PR #15747, opened 2025-01-20
DAOS-16876 vos: Improve error reporting on corruption crash

jolivier23:
If the records referenced by a DTX have been removed from the tree, the data will no longer match.
We can't detect this in all cases but we can avoid overwriting when we do detect it.

### Gatekeeper:


src/vos/vos_dtx.c
```diff
@@ -646,6 +658,14 @@ do_dtx_rec_release(struct umem_instance *umm, struct vos_container *cont,
 		break;
 	}
 
+	if (rc == -DER_NONEXIST) {
+		D_WARN("record no longer exists, may indicate some corruption: " DF_DTI
+		       " type %u\n",
+		       DP_DTI(&DAE_XID(dae)), dtx_umoff_flag2type(rec));
+		d_tm_inc_counter(vos_tls_get(false)->vtl_dtx_rec_missing, 1);
+		rc = 0;
+	}
```
Nasf-Fan, line 667: The user may miss the warning message. As we discussed, there is risk for silent data corruption. So when we hit recognisable corruption, I would suggest to trigger assertion that will make the user to be aware of potential data corruption in stead of make the system to go ahead without explicitly blocking the system.

On the other hand, in your current patch, if we do not access some corrupted DTX entry, the corruption will be always there, longer time then more dangerous since related area maybe reused by others as time going.

We can move your patch logic into some new DDB functionality to repair the dangling DTX records via removing them. Since it is  inconvenient to directly access "epoch" via DTX record reference, then we can do that as your patch does with checking lid+magic (if have). That is some kind of "quick mode" or "quick scan".

src/vos/vos_obj_index.c
```diff
@@ -321,6 +322,7 @@ vos_oi_find_alloc(struct vos_container *cont, daos_unit_oid_t oid,
 	if (log) {
 		vos_ilog_desc_cbs_init(&cbs, vos_cont2hdl(cont));
 		rc = ilog_open(vos_cont2umm(cont), &obj->vo_ilog, &cbs, dth == NULL, &loh);
+		D_ASSERTF(rc == -DER_NONEXIST, "Uncorrectable incarnation log corruption detected");
```
Nasf-Fan, line 325: rc != -DER_NONEXIST

Nasf-Fan (changes requested):

src/vos/vos_obj_index.c
jolivier23: Ah, right, I fixed the other one when a unit test failed

src/vos/vos_dtx.c
jolivier23: In our case, we would monitor the metric and alert someone that it happened.   I still don't think quick scan buys us anything.  A full scan would be able to find all of the missing entries but a quick scan can only find entries like the ones in this patch.

I think maybe having this behavior be configurable would be better.  Until we have full scan, this patch is really all we can really do

Nasf-Fan: One of the main concern for this patch is that we passively handle the corrupted DTX entries when we access them, that may leave more windows as to more risks of data corruption. @janekmi is working on the "quick scan" via ddb as offline mode. Maybe we need both.

src/tests/ftest/util/telemetry_utils.py
```diff
@@ -207,6 +208,7 @@ class TelemetryUtils():
         _gen_stats_metrics("engine_io_dtx_committable")
     ENGINE_IO_DTX_COMMITTED_METRICS = \
         _gen_stats_metrics("engine_io_dtx_committed")
+    ENGINE_IO_DTX_RECORD_MISSING_METRICS=["engine_io_dtx_record_missing"]
```
daltonbohning, line 211: > E225 missing whitespace around operator

```suggestion
    ENGINE_IO_DTX_RECORD_MISSING_METRICS = ["engine_io_dtx_record_missing"]
```

src/tests/ftest/util/telemetry_utils.py
```diff
@@ -310,6 +312,7 @@ class TelemetryUtils():
     ENGINE_IO_METRICS = ENGINE_IO_DTX_ASYNC_CMT_LAT_METRICS +\
         ENGINE_IO_DTX_COMMITTABLE_METRICS +\
         ENGINE_IO_DTX_COMMITTED_METRICS +\
+        ENGINE_IO_DTX_RECORD_MISSING_METRIC +\
```
daltonbohning, line 315: > undefined-variable, Undefined variable 'ENGINE_IO_DTX_RECORD_MISSING_METRIC'

Missing a list definition?

jolivier23: missed the S at the end

janekmi (dismissed):

src/vos/ilog.c
```diff
@@ -1213,7 +1221,7 @@ ilog_fetch(struct umem_instance *umm, struct ilog_df *root_df,
 	int			 rc = 0;
 	bool			 retry;
 
-	ILOG_ASSERT_VALID(root_df);
+	ILOG_CHECK_VALID(root_df);
```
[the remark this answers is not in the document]
NiuYawei: yeah, I think we have to assert.

src/vos/vos_dtx.c
NiuYawei: Indeed, I tend to agree with @janekmi and @Nasf-Fan . 

- If we don't know exactly the root case, It's not safe to keep engine running, otherwise, more severe data corruption could be introduced;
- If we know the root case for sure, we should fix the bug;

If the original assert can only be triggered by bug (it was a proper assert), it looks to me there is no way of getting round of it (I don't think software is able to gracefully handle it's own bug).

src/vos/vos_dtx.c
```diff
@@ -646,6 +658,14 @@ do_dtx_rec_release(struct umem_instance *umm, struct vos_container *cont,
 		break;
 	}
 
+	if (rc == -DER_NONEXIST) {
+		D_WARN("record no longer exists, may indicate some corruption: " DF_DTI
+		       " type %u\n",
+		       DP_DTI(&DAE_XID(dae)), dtx_umoff_flag2type(rec));
+		d_tm_inc_counter(vos_tls_get(false)->vtl_dtx_rec_missing, 1);
```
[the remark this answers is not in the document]
jolivier23: Will probably remove the metric since we decided abort is better but, in general, metrics are for admins. We use them all the time to try to diagnose issues with running systems.

src/vos/vos_dtx.c
```diff
@@ -571,7 +572,7 @@ dtx_ilog_rec_release(struct umem_instance *umm, struct vos_container *cont,
 
 	ilog_close(loh);
 
-	if (rc != 0)
+	if (rc != 0 && rc != -DER_NONEXIST)
```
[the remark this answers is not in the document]
Nasf-Fan: That is good point. Considering the scenarios of generating the dangling DTX record: the modification was made against some akey, but before its committed/aborted, related dkey was removed; or the modification was made against some dkey, but related object was removed. So under these scenarios, the subsequent modifications under related {d,a}key became meaningless, but we may cannot detect all dangling DTX records if there is no magic information (such as for SV) or "lid" maybe reused. So once we detect and discard the dangling DTX records for ilog, then all the subsequent modifications after such ilog DTX records should be discarded (until another valid ilog DTX record was found for distributed transaction case in the future).

src/vos/ilog.c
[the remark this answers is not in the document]
jolivier23: It will assert in the caller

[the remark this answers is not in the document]
jolivier23: nevermind, not sure why I did that.

NiuYawei (approved):

jolivier23:
@Nasf-Fan @janekmi can you have another look?

Nasf-Fan:
Ping reviewers, Thanks!

gnailzenh (approved):

gnailzenh (approved):

Nasf-Fan:
Ping @NiuYawei , thanks!

src/vos/ilog.c
```diff
@@ -363,7 +366,8 @@ ilog_open(struct umem_instance *umm, struct ilog_df *root,
 	struct ilog_context	*lctx;
 	int			 rc;
 
-	ILOG_ASSERT_VALID(root);
+	if (!ILOG_CHECK_VALID(root))
+		return -DER_NONEXIST;
```
NiuYawei, line 370: Better to add an error message.

src/vos/ilog.c
```diff
@@ -939,8 +943,12 @@ ilog_modify(daos_handle_t loh, const struct ilog_id *id_in,
 		"%s in incarnation log " DF_X64 " status: rc=" DF_RC " tree_version: %d\n",
 		opc_str[opc], id_in->id_epoch, DP_RC(rc), ilog_mag2ver(lctx->ic_root->lr_magic));
 
-	if (rc == 0 && version != ilog_mag2ver(lctx->ic_root->lr_magic) &&
-	    (opc == ILOG_OP_PERSIST || opc == ILOG_OP_ABORT)) {
+	if (rc == 0 && opc != ILOG_OP_UPDATE) {
+		if (version == ilog_mag2ver(lctx->ic_root->lr_magic)) {
+			D_WARN("ilog entry on %s doesn't exist\n", opc_str[opc]);
```
NiuYawei, line 948: D_ERROR() instead of D_WARN().

src/vos/vos_dtx.c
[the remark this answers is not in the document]
NiuYawei: I don't see why we want to avoid error messages on DER_NONEXIST errors.

NiuYawei (approved):

src/vos/vos_dtx.c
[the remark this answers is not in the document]
Nasf-Fan: > Isn't it too late to check DER_NONEXIST here? It should have returned error on ilog_open() above, right?

The caller - `do_dtx_rec_release()` will print the log message.

Nasf-Fan (approved):

PR #15752, opened 2025-01-21
DAOS-16951 dtx: add invalid records discard capability

janekmi:
# Guide's TLDR

- Introduce a new `vos_dtx_discard_invalid()` API
- Introduce a few accessory functions validating ILOG, SVT and EVT records
- Make the `vos_dtx_discard_invalid()` API available via DDB
- New unit test binaries (`dtx_ut` and `ddb_ut`)

# Guide (full)

- Introduce a new `vos_dtx_discard_invalid()` API
  - `src/include/daos_srv/vos.h` - add to the header file
  - `src/vos/vos_dtx.c`
    - core of the implementation goes here
    - **Note**: a few types and a helper function from this file have been moved. Please see `src/vos/ilog_internal.h` for details.
  - unit tests
    - `src/dtx/tests/dts_discard_invalid.c` - here you find all the unit tests for the `vos_dtx_discard_invalid()` API
    - `src/dtx/tests/SConscript` - because of the mocking requirements a new `dtx_ut` unit test binary has been introduced
    - `src/dtx/tests/dtx_ut.c` - greatly simplified copy of the `dtx_tests.c` - the entry point for all unit tests compiled in the binary
    - `src/dtx/tests/dts_structs.c` - because of the minimal requirements these tests fitted more to the new `dtx_ut` binary rather than to `dtx_tests`
- Introduce a few accessory functions validating ILOG, SVT and EVT records
  - ILOG:
    - `src/vos/ilog.h` - header
    - `src/vos/ilog.c` - implementation
      - **Note**: A few things have been moved from here. Please see `src/vos/ilog_internal.h` for details.
    - `src/vos/tests/vts_ilog.c` - unit tests
      - `src/vos/ilog_internal.h` - introduced to easier access ILOG internals for testing purposes.
  - SVT:
    - `src/vos/vos_internal.h` - header
    - `src/vos/vos_tree.c` - implementation
    - `src/vos/tests/vts_tree.c` - unit tests
  - EVT:
    - `src/include/daos_srv/evtree.h` - header file
    - `src/vos/evtree.c` - impementation
    - `src/vos/tests/vts_evtree.c` - unit tests
- Make the `vos_dtx_discard_invalid()` API available via DDB
  - `src/utils/ddb/ddb_commands.c`
    - `ddb_run_dtx_act_discard_invalid()` implementation
    - extend `dtx_modify_init()` to skip DTX ID initialization if `dti_all` is requested
    - `src/utils/ddb/tests/ddb_commands_tests.c` - simple tests
  - `src/utils/ddb/ddb.h`:
    - add a new command type and a new entry point - `ddb_run_dtx_act_discard_invalid()`
    - replace `struct dtx_act_commit_options` and `struct dtx_act_abort_options` with a single `struct dtx_act_options` which works for both `dtx_act_commit` and `dtx_act_abort` commands as well as the newly introduced `dtx_act_discard_invalid`
      - `src/utils/ddb/tests/ddb_cmd_options_tests.c` - update tests accordingly
  - `src/utils/ddb/ddb.c`:
    - `dtx_act_option_parse()` instead of `dtx_act_commit_option_parse()` and `dtx_act_abort_option_parse()` which works for both old and new `dtx_act_*` commands
    - connect command parsing and the new entry point
  - `dv_dtx_active_entry_discard_invalid()` proxy
    - `src/utils/ddb/ddb_vos.h` and `src/utils/ddb/ddb_vos.c`
    - `src/utils/ddb/tests/ddb_vos_ut.c` - unit tests
    - `src/utils/ddb/tests/SConscript` - because of the mocking requirements a new `ddb_ut` unit test binary has been introduced
    - `src/utils/ddb/tests/ddb_ut.c` - greatly simplified copy of the `ddb_test_driver.c` - the entry point for all unit tests compiled in the binary
- New unit test binaries (`dtx_ut` and `ddb_ut`)
  - `utils/utest.yaml` - add them to standard execution outlets
  - `utils/rpms/daos.spec` - add them to the RPM package
    - `utils/rpms/daos.rpmlintrc` - add them as yet another binaries without manuals
  - `debian/daos-server-tests.install` - add them to the DEB package
    - `debian/changelog` - report above in the changelog

---

### Gatekeeper:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+daos (2.7.101-5) unstable; urgency=medium
+  [ Jan Michalski ]
+  * Add ddb_ut and dtx_ut to the server-tests package
+
+ -- Jan Michalski <jan-marian.michalski@hpe.com>  Wed, 22 Jan 2025 14:31:00 +0000
+
 daos (2.7.101-3) unstable; urgency=medium
   [ Jeff Olivier ]
   * Switch from libfuse3 to libfused
--- a/debian/daos-server-tests.install
+++ b/debian/daos-server-tests.install
@@ -1,4 +1,5 @@
 usr/bin/dtx_tests
+usr/bin/dtx_ut
 usr/bin/jump_pl_map
 usr/bin/ring_pl_map
 usr/bin/evt_ctl
@@ -9,6 +10,7 @@ usr/bin/bio_ut
 usr/bin/vea_ut
 usr/bin/vos_tests
 usr/bin/ddb_tests
+usr/bin/ddb_ut
 usr/bin/vea_stress
 usr/bin/vos_perf
 usr/bin/obj_ctl
--- a/src/control/cmd/ddb/ddb_commands.go
+++ b/src/control/cmd/ddb/ddb_commands.go
@@ -342,7 +342,7 @@ the path must include the extent, otherwise, it must not.`,
 		HelpGroup: "vos",
 		Args: func(a *grumble.Args) {
 			a.String("path", "VOS tree path to a container.")
-			a.String("dtx_id", "DTX id of the entry to validate.")
+			a.String("dtx_id", "DTX id of the entry to validate or 'all' to validate all active DTX entries.")
 		},
 		Run: func(c *grumble.Context) error {
 			return ddbDtxActDiscardInvalid(ctx, c.Args.String("path"), c.Args.String("dtx_id"))
--- a/src/control/cmd/ddb/ddb_commands.go
+++ b/src/control/cmd/ddb/ddb_commands.go
@@ -1,5 +1,6 @@
 //
 // (C) Copyright 2022-2024 Intel Corporation.
+// (C) Copyright 2025 Hewlett Packard Enterprise Development LP.
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -331,5 +332,21 @@ the path must include the extent, otherwise, it must not.`,
 			return ddbRmPool(ctx, c.Args.String("path"))
 		},
 		Completer: rmPoolCompleter,
+	})
+	// Command: dtx_act_discard_invalid
+	app.AddCommand(&grumble.Command{
+		Name:      "dtx_act_discard_invalid",
+		Aliases:   nil,
+		Help:      "Discard the active DTX entry's records if invalid.",
+		LongHelp:  "Committing or aborting the DTX entry with discarded records may result in inconsistencies. Allow DTX to resync instead.",
+		HelpGroup: "vos",
+		Args: func(a *grumble.Args) {
+			a.String("path", "VOS tree path to a container.")
+			a.String("dtx_id", "DTX id of the entry to validate.")
+		},
+		Run: func(c *grumble.Context) error {
+			return ddbDtxActDiscardInvalid(ctx, c.Args.String("path"), c.Args.String("dtx_id"))
+		},
+		Completer: nil,
 	})
 }
--- a/src/control/cmd/ddb/ddb_commands.go
+++ b/src/control/cmd/ddb/ddb_commands.go
@@ -338,7 +338,7 @@ the path must include the extent, otherwise, it must not.`,
 		Name:      "dtx_act_discard_invalid",
 		Aliases:   nil,
 		Help:      "Discard the active DTX entry's records if invalid.",
-		LongHelp:  "Committing or aborting the DTX entry with discarded records may result in inconsistencies. Allow DTX to resync instead.",
+		LongHelp:  "",
 		HelpGroup: "vos",
 		Args: func(a *grumble.Args) {
 			a.String("path", "VOS tree path to a container.")
--- a/src/utils/ddb/tests/SConscript
+++ b/src/utils/ddb/tests/SConscript
@@ -26,24 +26,16 @@ def scons():
     libs = ['vos', 'daos_common_pmem', 'abt', 'gurt', 'uuid', 'bio', 'cart',
             'cmocka', 'ddb']
     # spdk libraries
-    spdk_libs = ['spdk_event', 'spdk_log']
-    spdk_libs += ['spdk_bdev', 'spdk_blob', 'spdk_blob_bdev', 'spdk_json']
-    spdk_libs += ['spdk_nvme', 'spdk_init', 'spdk_thread', 'spdk_log']
-    spdk_libs += ['spdk_env_dpdk', 'spdk_thread', 'spdk_bdev', 'rte_mempool']
-    spdk_libs += ['rte_mempool_ring', 'rte_bus_pci', 'rte_pci', 'rte_ring']
-    spdk_libs += ['rte_mbuf', 'rte_eal', 'rte_kvargs', 'spdk_bdev_aio']
-    spdk_libs += ['spdk_bdev_nvme', 'spdk_blob', 'spdk_nvme', 'spdk_util']
-    spdk_libs += ['spdk_json', 'spdk_jsonrpc', 'spdk_rpc', 'spdk_trace']
-    spdk_libs += ['spdk_sock', 'spdk_log', 'spdk_notify', 'spdk_blob_bdev']
-    spdk_libs += ['spdk_vmd', 'spdk_event_bdev', 'spdk_init', 'rte_power']
-    src = ['ddb_cmd_options_tests.c',
-           'ddb_commands_tests.c',
-           'ddb_main_tests.c',
-           'ddb_parse_tests.c',
-           'ddb_path_tests.c',
-           'ddb_test_driver.c',
-           'ddb_vos_tests.c',
-           'ddb_commands_print_tests.c']
+    spdk_libs = ['rte_bus_pci', 'rte_eal', 'rte_kvargs', 'rte_mbuf', 'rte_mempool_ring',
+                 'rte_mempool', 'rte_pci', 'rte_power', 'rte_ring', 'spdk_bdev_aio',
+                 'spdk_bdev_nvme', 'spdk_bdev', 'spdk_bdev', 'spdk_blob_bdev', 'spdk_blob_bdev',
+                 'spdk_blob', 'spdk_blob', 'spdk_env_dpdk', 'spdk_event_bdev', 'spdk_event',
+                 'spdk_init', 'spdk_init', 'spdk_json', 'spdk_json', 'spdk_jsonrpc', 'spdk_log',
+                 'spdk_log', 'spdk_log', 'spdk_notify', 'spdk_nvme', 'spdk_nvme', 'spdk_rpc',
+                 'spdk_sock', 'spdk_thread', 'spdk_thread', 'spdk_trace', 'spdk_util', 'spdk_vmd']
+    src = ['ddb_cmd_options_tests.c', 'ddb_commands_print_tests.c', 'ddb_commands_tests.c',
+           'ddb_main_tests.c', 'ddb_parse_tests.c', 'ddb_path_tests.c', 'ddb_test_driver.c',
+           'ddb_vos_tests.c']
     ddb_tests = denv.d_program('ddb_tests', [src], LIBS=libs + spdk_libs)
 
     denv.Install('$PREFIX/bin/', ddb_tests)
@@ -64,18 +56,9 @@ def scons():
     vos_src = Glob('../../../vos/*.c')
 
     mock_src = Glob('../../../dtx/tests/*_mock.c')
-    src = ['ddb_ut.c',
-           'ddb_vos_ut.c',
-           '../ddb_commands.c',
-           '../ddb_vos.c',
-           '../ddb_main.c',
-           '../ddb_printer.c',
-           '../ddb_tree_path.c',
-           '../ddb_parse.c',
-           '../ddb.c',
-           '../ddb_spdk.c',
-           ]
-    ddb_ut = denv.d_program('ddb_ut', src + vos_src + mock_src, LIBS=libs + spdk_libs)
+    ddb_src = Glob('../*.c', exclude=['../ddb_entry.c'])
+    src = ['ddb_ut.c', 'ddb_vos_ut.c']
+    ddb_ut = denv.d_program('ddb_ut', src + vos_src + mock_src + ddb_src, LIBS=libs + spdk_libs)
 
     denv.Install('$PREFIX/bin/', ddb_ut)
 
--- a/src/utils/ddb/tests/SConscript
+++ b/src/utils/ddb/tests/SConscript
@@ -56,6 +56,8 @@ def scons():
     vos_src = Glob('../../../vos/*.c')
 
     mock_src = Glob('../../../dtx/tests/*_mock.c')
+    # Pylint does not like the exclude keyword below, but it works.
+    # pylint: disable=unexpected-keyword-arg
     ddb_src = Glob('../*.c', exclude=['../ddb_entry.c'])
     src = ['ddb_ut.c', 'ddb_vos_ut.c']
     ddb_ut = denv.d_program('ddb_ut', src + vos_src + mock_src + ddb_src, LIBS=libs + spdk_libs)
--- a/src/utils/ddb/tests/SConscript
+++ b/src/utils/ddb/tests/SConscript
@@ -26,16 +26,16 @@ def scons():
     libs = ['vos', 'daos_common_pmem', 'abt', 'gurt', 'uuid', 'bio', 'cart',
             'cmocka', 'ddb']
     # spdk libraries
-    libs += ['spdk_event', 'spdk_log']
-    libs += ['spdk_bdev', 'spdk_blob', 'spdk_blob_bdev', 'spdk_json']
-    libs += ['spdk_nvme', 'spdk_init', 'spdk_thread', 'spdk_log']
-    libs += ['spdk_env_dpdk', 'spdk_thread', 'spdk_bdev', 'rte_mempool']
-    libs += ['rte_mempool_ring', 'rte_bus_pci', 'rte_pci', 'rte_ring']
-    libs += ['rte_mbuf', 'rte_eal', 'rte_kvargs', 'spdk_bdev_aio']
-    libs += ['spdk_bdev_nvme', 'spdk_blob', 'spdk_nvme', 'spdk_util']
-    libs += ['spdk_json', 'spdk_jsonrpc', 'spdk_rpc', 'spdk_trace']
-    libs += ['spdk_sock', 'spdk_log', 'spdk_notify', 'spdk_blob_bdev']
-    libs += ['spdk_vmd', 'spdk_event_bdev', 'spdk_init', 'rte_power']
+    spdk_libs = ['spdk_event', 'spdk_log']
+    spdk_libs += ['spdk_bdev', 'spdk_blob', 'spdk_blob_bdev', 'spdk_json']
+    spdk_libs += ['spdk_nvme', 'spdk_init', 'spdk_thread', 'spdk_log']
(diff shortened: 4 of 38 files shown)


Nasf-Fan (commented):
The framework and the basic process are fine. The main concern is that for most of cases, the user does not know which DTX entry is corrupted. Means that related DTX ID is known. So for conveniency, the logic itself needs to iterate the active DTX table, and for each active DTX entry, calls ddb_run_dtx_act_discard_invalid().

On the other hand, please check the patch from Jeff https://github.com/daos-stack/daos/pull/15747, he does similar repair with online mode. Currently, no decision about which one will be used finally, need more discussion.

janekmi:
> The framework and the basic process are fine. The main concern is that for most of cases, the user does not know which DTX entry is corrupted. Means that related DTX ID is known. So for conveniency, the logic itself needs to iterate the active DTX table, and for each active DTX entry, calls ddb_run_dtx_act_discard_invalid().
> 
> On the other hand, please check the patch from Jeff #15747, he does similar repair with online mode. Currently, no decision about which one will be used finally, need more discussion.

I added an option to just set `all` or a specific DTX id in this case.

Nasf-Fan:
The basic logic for discarding dangling DTX reference is fine. But my concern is that whether all the changes are related with our main purpose of resolving existing dangling DTX reference? We are so closed to the 2.6.3 deadline, I am afraid of landing such a huge patch maybe some challenged, not sure, but if possible, please split unrelated parts into subsequent patch(es). That will be helpful.

janekmi:
> The basic logic for discarding dangling DTX reference is fine. But my concern is that whether all the changes are related with our main purpose of resolving existing dangling DTX reference? We are so closed to the 2.6.3 deadline, I am afraid of landing such a huge patch maybe some challenged, not sure, but if possible, please split unrelated parts into subsequent patch(es). That will be helpful.

I believe here is no unrelated changes. Just a VOS/DTX API + DDB command + tests for all of these.

debian/daos-server-tests.install
```diff
@@ -9,6 +10,7 @@ usr/bin/bio_ut
 usr/bin/vea_ut
 usr/bin/vos_tests
 usr/bin/ddb_tests
+usr/bin/ddb_ut
```
jolivier23, line 13: so long as we are adding new unit tests to the package, why don't we run them?  Add entries to utils/utest.yaml ?

Alternatively, if these are just unit tests, you don't necessarily have to install them.  We really should settle on that.  Some tests we don't install, some we do and I don't think it's all that useful for most of them.  Anyway, the alternative here is to forego install and add to ci/test_files_to_stash.txt

src/utils/ddb/tests/SConscript
```diff
@@ -26,16 +26,16 @@ def scons():
...
-    libs += ['spdk_bdev', 'spdk_blob', 'spdk_blob_bdev', 'spdk_json']
-    libs += ['spdk_nvme', 'spdk_init', 'spdk_thread', 'spdk_log']
-    libs += ['spdk_env_dpdk', 'spdk_thread', 'spdk_bdev', 'rte_mempool']
-    libs += ['rte_mempool_ring', 'rte_bus_pci', 'rte_pci', 'rte_ring']
-    libs += ['rte_mbuf', 'rte_eal', 'rte_kvargs', 'spdk_bdev_aio']
-    libs += ['spdk_bdev_nvme', 'spdk_blob', 'spdk_nvme', 'spdk_util']
-    libs += ['spdk_json', 'spdk_jsonrpc', 'spdk_rpc', 'spdk_trace']
-    libs += ['spdk_sock', 'spdk_log', 'spdk_notify', 'spdk_blob_bdev']
-    libs += ['spdk_vmd', 'spdk_event_bdev', 'spdk_init', 'rte_power']
+    spdk_libs = ['spdk_event', 'spdk_log']
```
jolivier23, line 29: Why is this not one statement such as here?  So long as you're changing it, that would be more efficient

https://github.com/daos-stack/daos/blob/d38cd11a1fcfa175ce73a881051f20a04efae0cf/src/SConscript#L5-L10

src/vos/ilog_internal.h
```diff
@@ -0,0 +1,61 @@
+/**
```
jolivier23, line 1: Not opposed to this change necessarily but this is only used by ilog.c, right?

src/vos/vos_tree.c
```diff
@@ -1354,3 +1355,13 @@ obj_tree_find_attr(unsigned tree_class, int flags)
 			return NULL;
 	}
 }
+
+bool
+vos_irec_is_valid(const struct vos_irec_df *svt, uint32_t dtx_lid)
+{
+	if (svt == NULL) {
```
jolivier23, line 1362: Would need to check but if we use zalloc, we may also be able to check that ir_pad8 and ir_pad16 are zero.  Could possibly hijack one of them  to check if zero || magic

jolivier23 (changes requested):
Just a few minor comments.  Otherwise, looks fine to me

debian/daos-server-tests.install
janekmi: > why don't we run them? Add entries to utils/utest.yaml ?

Strange. I do add them to `utils/utest.yaml` with this patch. Please double-check.

src/vos/ilog_internal.h
janekmi: No. It was necessary to write a unit test for newly created `ilog_is_valid()`. Please see `src/vos/tests/vts_ilog.c`.

src/vos/vos_tree.c
janekmi: According to my quick survey, it is zalloc'ed. However, I am a bit worried about using a side effect as a basis for a new validation rule for this type of record. So, I would not go this way.

src/utils/ddb/tests/SConscript
janekmi: IMHO these lists are so long it doesn't look right no matter what. xD Done.

knard38 (dismissed):
LGTM for what I understand.
Just a minor questions regarding changelog management.

For the validation I was able to use this new version of the ddb tool to fix the corrupted VOS files from LRZ.
image

However, there is still some issues for detecting corrupted VOS files.
The only way that I know until now is to crash ddb with running a dtx_act command on a corrupted DTX entry before discarding it.
image

I will create a new JIRA ticket to fix/improve this.

debian/changelog
```diff
@@ -1,3 +1,9 @@
+daos (2.7.101-5) unstable; urgency=medium
```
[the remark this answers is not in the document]
janekmi: IMHO it is better to keep both version in sync. I do not want to have a discussion in which we try to determine which RPM version is an equivalent of a certain DEB version. It would be bad.

tanabarr (dismissed):
go and build changes LGTM

knard38 (dismissed):

knard38 (dismissed):

src/control/cmd/ddb/ddb_commands.go
```diff
@@ -332,4 +333,20 @@ the path must include the extent, otherwise, it must not.`,
 		},
 		Completer: rmPoolCompleter,
 	})
+	// Command: dtx_act_discard_invalid
+	app.AddCommand(&grumble.Command{
+		Name:      "dtx_act_discard_invalid",
+		Aliases:   nil,
+		Help:      "Discard the active DTX entry's records if invalid.",
+		LongHelp:  "Committing or aborting the DTX entry with discarded records may result in inconsistencies. Allow DTX to resync instead.",
```
Nasf-Fan, line 341: This is some misguided, if the DTX record is discarded, then related DTX record will be set as "UMOFF_NULL", that will allow the subsequent DTX_resync logic to commit or abort related DTX entry. Here, when discarding the invalid DTX record, we cannot locally decide whether the DTX entry can be committed or aborted. It is the DTX resync logic to globally make such decision.

src/vos/ilog.c
```diff
@@ -1621,3 +1576,35 @@ ilog_version_get(daos_handle_t loh)
...
+	}
+
+	if (ilog_empty(root)) {
+		return false;
+	}
+
+	if (root->lr_tree.it_embedded) {
+		id = &root->lr_id;
+		return (id->id_tx_id == dtx_lid && id->id_epoch == epoch);
+	} else {
```
Nasf-Fan, line 1599: `else` is redundant.

src/vos/vos_dtx.c
```diff
@@ -2706,6 +2678,144 @@ vos_dtx_abort(daos_handle_t coh, struct dtx_id *dti, daos_epoch_t epoch)
...
+		}
+
+		count = DTX_INLINE_REC_CNT;
+	} else {
+		count = DAE_REC_CNT(dae);
+	}
+
+	/* go through the inlined records */
+	for (i = 0; i < count; i++) {
+		do_dtx_rec_discard_invalid(umm, dae, &DAE_REC_INLINE(dae)[i], &discarded_inline);
```
Nasf-Fan, line 2755: Considering the scenarios of generating the dangling DTX reference: the modification was made against some akey, but before its committed/aborted, related dkey was removed; or the modification was made against some dkey, but related object was removed. So under these scenarios, the subsequent modifications under related {d,a}key became meaningless, but we may cannot detect all dangling DTX records if there is no magic information (such as for SV) or "lid" maybe reused. So once we detect and discard the dangling DTX record for ilog, then all the subsequent modification(s) after such ilog DTX record should be discarded (until another valid ilog DTX record was found for distributed transaction case in the future). So here, if current DTX record is invalid, then it can be used as the hint for the next DTX record in the same DTX entry. Usually, ilog DTX record will be registered before EV/SV records, so we need to check inline DTX records firstly, then non-inline ones.

src/vos/vos_dtx.c
```diff
@@ -2706,6 +2678,144 @@ vos_dtx_abort(daos_handle_t coh, struct dtx_id *dti, daos_epoch_t epoch)
...
+	/* lookup the DTX entry */
+	d_iov_set(&kiov, dti, sizeof(*dti));
+	d_iov_set(&riov, NULL, 0);
+	rc = dbtree_lookup(cont->vc_dtx_active_hdl, &kiov, &riov);
+	if (rc != 0) {
+		return rc;
+	}
+	dae = riov.iov_buf;
+
+	if (vos_dae_is_abort(dae)) {
```
Nasf-Fan, line 2798: Why skip `vos_dae_is_abort`?

src/control/cmd/ddb/ddb_commands.go
knard38: Not sure to understand all the details, but from the end user point of view, I had to manually abort the discarded DTX entries.  @Nasf-Fan are you suggesting that this should be done automatically by the engine or the ddb tool during the next resync ?

src/vos/vos_dtx.c
janekmi: Do you see an option to improve the invalid record detection here but it would require to process inline records first, is it correct?

At the same time, the order does not make any difference considering the current algorithm where each of the records is processed separately.

Just to be clear, the order of processing was directly derived from `dtx_rec_release()`. But in that case the records are processed separately as well.

Do you insist to change the order of processing despite it does not change a thing for the current algorithm or am I missing something?

src/vos/vos_dtx.c
janekmi: Considering this excerpt from the `vos_dtx_abort()`:

```c
	if (vos_dae_is_abort(dae)) {
		/*
		 * The DTX has been aborted before, but failed to be removed from the active
		 * table at that time, then need to be removed again via vos_dtx_post_handle.
		 */
		if (dae->dae_aborted)
			vos_dtx_post_handle(cont, &dae, NULL, 1, true, false, false);

		D_GOTO(out, rc = -DER_ALREADY);
	}
```

When `vos_dae_is_abort(dae) == true` its records won't be processed anymore. So, there is no point of discarding them. My understanding is that this particular DAE can be simply removed from the active table.

Am I correct?

src/control/cmd/ddb/ddb_commands.go
janekmi: @Nasf-Fan can you recommend a direction/proper wording here?

Nasf-Fan: Sorry, I may did not express the things clearly. Discarding invalid DTX records has nothing related with DTX commit or abort. Even if for a valid DTX entry without any invalid DTX records, it is also risky when manually commit or abort locally without global knowledge. For this case, after discarding the invalid DTX records, it will be more safe to make related DTX entry with left valid DTX record (or even if the DTX entry became empty) to be handled via subsequent DTX resync globally.

src/vos/vos_tree.c
jolivier23: Sure, my assertion is that we could take one of those fields, rename it to "magic", set a magic number and then allow 0 or magic.  We are essentially preventing corruption here so I think using whatever we can for validation is reasonable.  @NiuYawei @Nasf-Fan have an opinion here?

jolivier23 (dismissed):

src/vos/vos_tree.c
jolivier23: Along similar lines, I wonder if there is anything else we can use to validate the record(s).   Such as whether the bio_addr value makes sense (e.g. not out of bounds).

osalyk (dismissed):
:lgtm:

src/vos/vos_dtx.c
Nasf-Fan: For example, for insert a SV record via stand-along (non-distributed transaction that is not officially supported by 2.6.x) modification, related DTX registration order maybe as the followings:

1. register ilog rec for object if the object shard does not exist
2. register ilog rec for dkey if the dkey does not exist
3. register ilog rec for akey if the akey does not exist
4. register SV rec for the record

Generally, it is more easy to detect the ilog rec related corruption since there is the magic check, but relative difficult for SV rec. So if we found the ilog rec corruption, then that can be used as the hint to detect corruption for subsequent DTX records belong to the same DTX entry. That bases on the scenarios of how those dangling DTX records were generated.

> Do you see an option to improve the invalid record detection here but it would require to process inline records first, is it correct?

For the discard logic, inline DTX records needs to be checked firstly.

> At the same time, the order does not make any difference considering the current algorithm where each of the records is processed separately.

That is different, as explained above, there is DTX registration order among those DTX records belong to the same DTX entry. If the former DTX record is corrupted, then for stand-alone transaction, the subsequent DTX records belong to the same DTX entry will be also invalid and needs to be discard.

src/vos/vos_dtx.c
Nasf-Fan: In spite of DTX commit or DTX abort, the logic will set the DTX lid that is stored inside related target record (modified via current DTX) as a special LID (DTX_LID_COMMITTED or DTX_LID_ABORTED).

So both the commit and abort logic will parse DTX record and locate the target record. If the DTX record is invalid, then such DTX record needs to be discard in spite of it is for commit or abort.

src/control/cmd/ddb/ddb_commands.go
janekmi: Done.

src/vos/ilog.c
janekmi: Done.

src/vos/vos_dtx.c
janekmi: > If the former DTX record is corrupted, then for stand-alone transaction, the subsequent DTX records belong to the same DTX entry will be also invalid and needs to be discard.

@Nasf-Fan is this a thing that is missing from this feature because otherwise it is incorrect? Or is it an improvement which be nice to have in the future?

src/control/cmd/ddb/ddb_commands.go
janekmi: It is great we clarified this. The misleading comment is gone. Thanks!

src/vos/vos_dtx.c
janekmi: Done. I changed the order. Now the inline records go first.

src/vos/vos_tree.c
janekmi: I just filed a ticket to refine this idea and possibly implement it in the next round: https://daosio.atlassian.net/browse/DAOS-17005. Feel free to add further ideas there. I believe there is a greater chance to discuss them properly and make an informed decision.

src/control/cmd/ddb/ddb_commands.go
knard38: > > Not sure to understand all the details, but from the end user point of view, I had to manually abort the discarded DTX entries. @Nasf-Fan are you suggesting that this should be done automatically by the engine or the ddb tool during the next resync ?
> 
> "I had to manually abort the discarded DTX entries.", why for this? it is more safe to make related DTX entry with left valid DTX record (or even if the DTX entry became empty) to be handled via subsequent DTX resync that will have global knowledge and make the decision automatically.

TBH, I just followed the same workaround which was provided for solving the hot LRZ issues.
I was thinking, that after discarding invalid DTX it was needed to abort it in order to keep the system consistent.
From my understanding of your sentence, the abort is not needed and the choice between abort and commit will be done by the engine will restarted during the next resin() of the discorded entries.

src/control/cmd/ddb/ddb_commands.go
```diff
@@ -332,4 +333,20 @@ the path must include the extent, otherwise, it must not.`,
 		},
 		Completer: rmPoolCompleter,
 	})
+	// Command: dtx_act_discard_invalid
+	app.AddCommand(&grumble.Command{
+		Name:      "dtx_act_discard_invalid",
+		Aliases:   nil,
+		Help:      "Discard the active DTX entry's records if invalid.",
+		LongHelp:  "",
```
knard38, line 341: Sound strange to not have LongHelp at all.
At least, the same short help could be set.

Nasf-Fan: The original LongHelp message was some confused or say that it was not related with current changing for discarding invalid DTX record. But I did not suggest to remove the LongHelp message, instead, I am fine to reuse the shard help for that.

jolivier23 (approved):

src/control/cmd/ddb/ddb_commands.go
janekmi: > Sound strange to not have LongHelp at all.

It may seem strange, but it's not odd at all. To be exact, all other DDB commands have empty `LongHelp,` and we were okay with that until now. IMHO, there is no need to make an exception unless we have something extra to write here. I am not aware of anything specific we would like to write here, but I am open to suggestions.

Nasf-Fan (approved):
We can consider to improve the ability of detection SVT record properly in subsequent patch(es).

src/control/cmd/ddb/ddb_commands.go
knard38: TBH, I am not fully agree with you as there is some other commands which have some long helps:
- https://github.com/daos-stack/daos/blob/9e6042fec749fdd6b5aa0b3fbc6e97e39a5f0446/src/control/cmd/ddb/ddb_commands.go#L38
- https://github.com/daos-stack/daos/blob/9e6042fec749fdd6b5aa0b3fbc6e97e39a5f0446/src/control/cmd/ddb/ddb_commands.go#L94
- https://github.com/daos-stack/daos/blob/9e6042fec749fdd6b5aa0b3fbc6e97e39a5f0446/src/control/cmd/ddb/ddb_commands.go#L129

Moreover, such critical command should have some detailed explanation from my point of view.
For example, it could be very valuable, for the end user, to have some explanations of the --all option and/or the fact that it is not needed to abort/commit the fixed DTX entries.

However, it is not a blocker for me but it should probably be a good thing to work on this for the follow-up PR.

knard38 (approved):

tanabarr (approved):
Go and build changes LGTM

PR #15770, opened 2025-01-23
DAOS-16968 test: Fix pool query tests

mjmac:
The backport from master missed a couple of pool tests that
needed to be updated for the new JSON output from pool query.
Query results always include a dead_ranks array, even when
it's empty.

### DIFF
--- a/src/tests/ftest/control/dmg_pool_query_ranks.py
+++ b/src/tests/ftest/control/dmg_pool_query_ranks.py
@@ -53,6 +53,7 @@ class DmgPoolQueryRanks(ControlTestBase):
         self.log_step("Checking pool query with dead ranks state information")
         data = self.dmg.pool_query(self.pool.identifier, health_only=True)
         self._verify_ranks([], data, "dead_ranks")
+        self._verify_ranks([], data, "disabled_ranks")
 
         data = self.dmg.pool_query(self.pool.identifier, show_disabled=True)
         self._verify_ranks(None, data, "enabled_ranks")
@@ -119,13 +120,13 @@ class DmgPoolQueryRanks(ControlTestBase):
         self.log_step(f"Stopping rank:{dead_rank} all_ranks={all_ranks}")
         self.server_managers[0].stop_ranks([dead_rank], self.d_log)
 
-        self.log_step(f"Waiting for pool rank {dead_rank} to be deaded")
+        self.log_step(f"Waiting for pool rank {dead_rank} to be dead")
         self.pool.wait_pool_dead_ranks([dead_rank], timeout=30)
 
         self.log_step(f"Starting rank {dead_rank}")
         self.server_managers[0].start_ranks([dead_rank], self.d_log)
 
-        self.log_step("Waiting for pool ranks to no longer be deaded")
+        self.log_step("Waiting for pool ranks to no longer be dead")
         self.pool.wait_pool_dead_ranks([], timeout=30)
 
         self.log_step("Waiting for rebuild to complete")
--- a/src/tests/ftest/control/dmg_pool_query_test.py
+++ b/src/tests/ftest/control/dmg_pool_query_test.py
@@ -1,5 +1,6 @@
 """
   (C) Copyright 2020-2024 Intel Corporation.
+  (C) Copyright 2025 Hewlett Packard Enterprise Development LP
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 """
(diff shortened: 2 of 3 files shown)


src/tests/ftest/control/dmg_pool_query_ranks.py
```diff
@@ -53,6 +53,7 @@ def test_pool_query_ranks_basic(self):
         self.log_step("Checking pool query with dead ranks state information")
         data = self.dmg.pool_query(self.pool.identifier, health_only=True)
         self._verify_ranks([], data, "dead_ranks")
+        self._verify_ranks([], data, "disabled_ranks")
```
phender, line 56: Shouldn't we be making this change to master before implementing it on the release/2.6 branch?  Also lines 58-60 of this file don't exist on https://github.com/daos-stack/daos/blob/master/src/tests/ftest/control/dmg_pool_query_ranks.py

src/tests/ftest/control/dmg_pool_query_ranks.py
```diff
@@ -119,13 +120,13 @@ def test_pool_query_ranks_mgmt(self):
         self.log_step(f"Stopping rank:{dead_rank} all_ranks={all_ranks}")
         self.server_managers[0].stop_ranks([dead_rank], self.d_log)
 
-        self.log_step(f"Waiting for pool rank {dead_rank} to be deaded")
+        self.log_step(f"Waiting for pool rank {dead_rank} to be dead")
```
phender, line 123: Should this step also be verifying the disabled ranks after line 124 and after line 156 like the master branch version does [here](https://github.com/daos-stack/daos/blob/master/src/tests/ftest/control/dmg_pool_query_ranks.py#L97) and [here](https://github.com/daos-stack/daos/blob/master/src/tests/ftest/control/dmg_pool_query_ranks.py#L130)?

src/tests/ftest/control/dmg_pool_query_test.py
```diff
@@ -84,6 +85,7 @@ def test_pool_query_basic(self):
                     "total": self.params.get("total", path="/run/exp_vals/nvme/*")
                 }
             ],
+            "dead_ranks": [],
```
phender, line 88: This also seems like we are making these changes in the reverse order as this entry doesn't exist in https://github.com/daos-stack/daos/blob/master/src/tests/ftest/control/dmg_pool_query_test.py

phender (changes requested):

src/tests/ftest/control/dmg_pool_query_ranks.py
mjmac: @phender: I'm confused. Are these tests broken on master, too? Or just release/2.6? There are significant differences between the branches (2.6 does not have @knard38's changes to allow for enabled and disabled ranks to be queried simultaneously, in particular).

wangshilong: Sorry, I looked at my original back-port PRs, I missed "Feature: pool" tags in the following commits which missed this test. it was included in the first rename commit.

wangshilong (approved):

wangshilong (dismissed):

src/tests/ftest/control/dmg_pool_query_ranks.py
phender: Yes, the test is only failing in the release/2.6 branch, but it is not clear to me with which commits the tests deviated and where the cherry-picks went wrong to cause this.  I just want to be sure we're not making this more confusing by just patching the test failure and possibly removing test functionality or coverage.

mjmac: For this particular change, I suspect that @wangshilong just made a small mistake in resolving a conflict. The health-only pool query is supposed to return a `disabled_ranks` array that can be empty on this branch. As the query also always returns a `dead_ranks` array on this branch, the test should check both.

In any case, I think that this is all working as intended on this branch.

daltonbohning (dismissed):

phender (changes requested):
We should also fix pool/list_verbose.py which has the same `dead_ranks` problem

phender (approved):

daltonbohning (approved):

mjmac:
@daltonbohning, @phender: This has been waiting for > 48hrs for the remaining hw stages. It's already passed the following tests that were fixed in the PR, though:
  * pool/list_verbose.py
  * control/dmg_pool_query_ranks.py
  * control/dmg_pool_query_test.py

I can continue to monitor and land it myself when it finally finishes, or if you guys don't see any benefit in waiting, you could force-land it.

daltonbohning:
> @daltonbohning, @phender: This has been waiting for > 48hrs for the remaining hw stages. It's already passed the following tests that were fixed in the PR, though:
> 
> * pool/list_verbose.py
> * control/dmg_pool_query_ranks.py
> * control/dmg_pool_query_test.py
> 
> I can continue to monitor and land it myself when it finally finishes, or if you guys don't see any benefit in waiting, you could force-land it.

It's misleading, but no other stages will run. This PR only modified specific ftest files, so only those specific tests are ran, and only the stages where those tests run.
The gap here is the landing requirements are not dynamic, so the checks still are expecting HW Large, etc.

PR #15803, opened 2025-01-28
DAOS-16312 control: Always use --force for dmg system stop

tanabarr:
Whenever stopping an engine process from within the control-plane, use
SIGKILL rather than asking nicely (SIGTERM). This has been requested
to try to avoid situations that could result in dataloss.

A side effect of this change is that `ds_mgmt_drpc_prep_shutdown()` and
then `ds_pool_disable_exclude()` will not be called during a controlled
shutdown where `dmg system stop` is called without other arguments.

Allow-unstable-test: true
Features: control

### Gatekeeper:

### DIFF
--- a/src/control/lib/control/system.go
+++ b/src/control/lib/control/system.go
@@ -474,7 +474,8 @@ func SystemStop(ctx context.Context, rpcClient UnaryInvoker, req *SystemStopReq)
 		Hosts: req.Hosts.String(),
 		Ranks: req.Ranks.String(),
 		Sys:   req.getSystem(rpcClient),
-		Force: req.Force,
+		// DAOS-16312: Always force stop engine processes.
+		Force: true,
 	}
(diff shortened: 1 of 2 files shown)


tanabarr:
CI run  2 started with allow unstable pragma set and P2 priority

tanabarr:
Not all hardware stages started in run 2, restart from stage hardware test -> run 3.

kjacque (commented):
I appreciate the simplicity. I guess the one question is whether we need the prepare stage at all anymore, if it's okay to just kill the engines.

If the intention is to bring back the original behavior someday soon, with this being a temporary hack to fix an urgent problem, I think this is the best approach.

However if this is going to be permanent, or as close to it as software gets, I think more major surgery is in order (I believe you had another PR taking that approach?). If that's the case, I think we should remove all of the obsolete clean shutdown code. If we need to reimplement in a year or two, then so be it--the parameters of the problem may have changed by then.

PR #15809, opened 2025-01-29
DAOS-17001 rebuild: when self_heal is set to delay_rebuild, do not re…

cdavis28:
…build on exclude

delay_rebuild mode should delay the rebuild in all scenarios and not have an exception for target exclusion. Also changed an error message to warn on shard update failure. Shard update failure is normal during a failure, and the message was too frequent.

Testing:
`dmg pool exclude default-pool --rank 0 --target-idx 4` while write/read workflow was running against a cluster

### Gatekeeper:

### DIFF
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2016-2024 Intel Corporation.
+ * (C) Copyright 2025 Google LLC
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -5816,7 +5817,7 @@ dc_obj_update(tse_task_t *task, struct dtx_epoch *epoch, uint32_t map_ver,
 
 	rc = obj_update_shards_get(obj, args, map_ver, obj_auxi, &shard, &shard_cnt);
 	if (rc != 0) {
-		D_ERROR(DF_OID" get update shards failure %d\n", DP_OID(obj->cob_md.omd_id), rc);
+		D_WARN(DF_OID " get update shards failure %d\n", DP_OID(obj->cob_md.omd_id), rc);
 		D_GOTO(out_task, rc);
 	}
 
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -5817,7 +5817,8 @@ dc_obj_update(tse_task_t *task, struct dtx_epoch *epoch, uint32_t map_ver,
 
 	rc = obj_update_shards_get(obj, args, map_ver, obj_auxi, &shard, &shard_cnt);
 	if (rc != 0) {
-		D_WARN(DF_OID " get update shards failure %d\n", DP_OID(obj->cob_md.omd_id), rc);
+		D_CDEBUG(rc == -DER_STALE, DLOG_INFO, DLOG_ERR,
+			 DF_OID " get update shards failure %d\n", DP_OID(obj->cob_md.omd_id), rc);
 		D_GOTO(out_task, rc);
 	}
 
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -5817,7 +5817,7 @@ dc_obj_update(tse_task_t *task, struct dtx_epoch *epoch, uint32_t map_ver,
 
 	rc = obj_update_shards_get(obj, args, map_ver, obj_auxi, &shard, &shard_cnt);
 	if (rc != 0) {
-		D_CDEBUG(rc == -DER_STALE, DLOG_INFO, DLOG_ERR,
+		D_CDEBUG(rc == -DER_STALE, DLOG_DBG, DLOG_ERR,
 			 DF_OID " get update shards failure %d\n", DP_OID(obj->cob_md.omd_id), rc);
 		D_GOTO(out_task, rc);
 	}
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -7344,7 +7344,7 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
 
 	entry = daos_prop_entry_get(&prop, DAOS_PROP_PO_SELF_HEAL);
 	D_ASSERT(entry != NULL);
-	if (!(entry->dpe_val & DAOS_SELF_HEAL_AUTO_REBUILD)) {
+	if (!(entry->dpe_val & (DAOS_SELF_HEAL_AUTO_REBUILD | DAOS_SELF_HEAL_DELAY_REBUILD))) {
 		D_DEBUG(DB_MD, "self healing is disabled\n");
 		D_GOTO(out, rc);
 	}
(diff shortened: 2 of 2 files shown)


src/pool/srv_pool.c
```diff
@@ -7343,7 +7344,7 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
 
 	entry = daos_prop_entry_get(&prop, DAOS_PROP_PO_SELF_HEAL);
 	D_ASSERT(entry != NULL);
-	if (!(entry->dpe_val & (DAOS_SELF_HEAL_AUTO_REBUILD | DAOS_SELF_HEAL_DELAY_REBUILD))) {
+	if (!(entry->dpe_val & DAOS_SELF_HEAL_AUTO_REBUILD)) {
```
wangdi1, line 7347: Delay rebuild mean delaying rebuild, instead of disabling rebuild, i.e. the rebuild job should still be scheduled though with delay = -1, i.e. the rebuild job will stay in the queue until there are other reintegrate/extend job being issued.

Hmm, it seems this check should be only for exclude, other operation like, extend/drain/reintegrate should not check this.  Though this is not for this patch.

cdavis28: Actually, maybe I didn't. let me try again.

cdavis28: Actually lets talk @wangdi1. I would like to better understand why we would want to schedule a repair on any pool map update with delay_rebuild.

jolivier23 (dismissed):

src/pool/srv_pool.c
```diff
@@ -7360,7 +7361,7 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
 		D_GOTO(out, rc);
 	}
 
-	if ((entry->dpe_val & DAOS_SELF_HEAL_DELAY_REBUILD) && exclude_rank)
+	if (entry->dpe_val & DAOS_SELF_HEAL_DELAY_REBUILD)
```
jolivier23, line 7364: Only question I have....do we need any extra check here?   For stuff like reintegrate or drain or extend, we actually do want rebuild.  But I will defer to Di who is infinitely more familiar with the logic here

wangdi1 (dismissed):

src/pool/srv_pool.c
kccain: Yeah, I am also curious about this, too. It seems another check is needed.
So the goal of the patch is to make sure to have a delayed rebuild for any type of exclusion whether initiated by SWIM detection of a lost engine (already handled), or administratively excluded targets (e.g., via dmg pool exclude command). 

Maybe logic such as the following would do what is needed here, to not impact reintegrate, drain, and extend?

if ((entry->dpe_val & DAOS_SELF_HEAL_DELAY_REBUILD) && (opc == MAP_EXCLUDE))

Now that I look at this, it seems maybe there is a bug in the DAOS_REINT_MODE_NO_DATA_SYNC branch above when it tests opc to see if it is POOL_EXCLUDE or POOL_DRAIN. I think instead the test there should be for MAP_EXCLUDE or MAP_DRAIN, since the caller converts the RPC opcodes (POOL_*) into pool map opcodes (MAP_*) that have different values (!).

jolivier23 (dismissed):

src/pool/srv_pool.c
```diff
@@ -7350,7 +7351,7 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
```
wangdi1, line 7351: probably this(L7346)  should be changed as  this?

if (opc == MAP_EXCLUDE && !(entry->dpe_val & (DAOS_SELF_HEAL_AUTO_REBUILD | DAOS_SELF_HEAL_DELAY_REBUILD)) {
               
}
 So only exclude operation will be checked by this property. Though this is not this patch issue, probably can be done in a different patch.

wangdi1 (dismissed):

jolivier23:
@liuxuezhao can you please review this?

src/object/cli_obj.c
```diff
@@ -5816,7 +5817,7 @@ dc_obj_update(tse_task_t *task, struct dtx_epoch *epoch, uint32_t map_ver,
 
 	rc = obj_update_shards_get(obj, args, map_ver, obj_auxi, &shard, &shard_cnt);
 	if (rc != 0) {
-		D_ERROR(DF_OID" get update shards failure %d\n", DP_OID(obj->cob_md.omd_id), rc);
+		D_WARN(DF_OID " get update shards failure %d\n", DP_OID(obj->cob_md.omd_id), rc);
```
mchaarawi, line 5820: no quite sure i understand the reasoning behind this. the main comment says:
"Shard update failure is normal during a failure, and the message was too frequent."
which is not clear to me.
frequency does not mean the update is not a legitimate error.

cdavis28: The issue with this message is that it's an expected condition during a target exclusion. 
The 2nd part of the problem is that we log the message a few thousand times in a tight loop during an exclusion.

mchaarawi: - what about outside of a target exclusion?
- can you post an example of a log? i thought retries on client happen every CRT_TIMEOUT.

ultimately, if this error can be expected, then WARN is also not the right level. it would be DEBUG or at least INFO IMO. but i am not convinced that in some scenarios the return error is not legitimate.

cdavis28: @jolivier23 do you have a comment here? I originally did have it at INFO, but you thought it should be WARN.

cdavis28: At the time you wrote:
> I think WARN might be more appropriate but we are also potentially ignoring errors here. We typically use D_CDEBUG for such cases to make sure the error is retryable or such before we lower its log level

kccain (dismissed):
srv_pool.c changes LGTM

src/object/cli_obj.c
jolivier23: I think D_CDEBUG is probably what we want but I really want to create a better solution for those checks to avoid too many branches (such as retryable error checks).

liuxuezhao (dismissed):

src/object/cli_obj.c
```diff
@@ -5816,7 +5817,8 @@ dc_obj_update(tse_task_t *task, struct dtx_epoch *epoch, uint32_t map_ver,
 
 	rc = obj_update_shards_get(obj, args, map_ver, obj_auxi, &shard, &shard_cnt);
 	if (rc != 0) {
-		D_ERROR(DF_OID" get update shards failure %d\n", DP_OID(obj->cob_md.omd_id), rc);
+		D_CDEBUG(rc == -DER_STALE, DLOG_INFO, DLOG_ERR,
```
liuxuezhao, line 5820: minor, DLOG_INFO looks can be replaced by DLOG_DBG

cdavis28: since we are here. Done.

liuxuezhao (approved):

kccain (approved):

jolivier23 (approved):

PR #15846, opened 2025-02-05
DAOS-16927 pool: Store server handles in RDBs

liw:
See the Jira ticket for the background. In the upgrade case, we try to use the existing volatile server handles generated by the old method, so that hopefully problems will be less likely to occur.

Features: pool

### Gatekeeper:


src/pool/srv_pool.c
```diff
@@ -5912,6 +5967,49 @@ pool_upgrade_props(struct rdb_tx *tx, struct pool_svc *svc, uuid_t pool_uuid, cr
...
+	 * their existing values instead of generating new UUIDs.
+	 */
+	d_iov_set(&value, valuuid, sizeof(uuid_t));
+	rc = rdb_tx_lookup(tx, &svc->ps_root, &ds_pool_prop_srv_handle, &value);
+	if (rc && rc != -DER_NONEXIST) {
+		D_GOTO(out_free, rc);
+	} else if (rc == -DER_NONEXIST) {
+		if (uuid_is_null(svc->ps_pool->sp_srv_pool_hdl)) {
+			D_ERROR(DF_UUID ": server pool handle unstable\n", DP_UUID(pool_uuid));
+			rc = -DER_AGAIN;
```
liw, line 5982: [Note] My manual testing has revealed a problem in this case. I think it's an existing bug with the upgrade mechanism, and will address it elsewhere.

liw:
Previous Ubuntu build failed due to unable to resolve dependencies; rebase and try again.

liw:
Due to extremely slow CI responses, I'm requesting reviews before CI results come in.

src/pool/srv_pool.c
```diff
@@ -2358,27 +2407,33 @@ pool_svc_step_up_cb(struct ds_rsvc *rsvc)
 		goto out;
 	}
 
-	if (!uuid_is_null(svc->ps_pool->sp_srv_cont_hdl)) {
-		uuid_copy(pool_hdl_uuid, svc->ps_pool->sp_srv_pool_hdl);
-		uuid_copy(cont_hdl_uuid, svc->ps_pool->sp_srv_cont_hdl);
+	if (svc->ps_global_version >= DAOS_POOL_GLOBAL_VERSION_WITH_SRV_HDLS) {
+		/* See the is_pool_from_srv comment in the "else" branch. */
```
NiuYawei, line 2411: I'm bit worried about this "delay the sp_srv_cont_hdl initialization in ds_pool_iv_refresh_hdl()", what if an IV or whatever request reached leader and called ds_cont_hdl_rdb_lookup() before the sp_srv_cont_hdl is initialized?

I also don't quite see why we need to compare the request handle with this global sp_srv_cont_hdl in ds_cont_hdl_rdb_lookup(), this function is only called by cont_iv_ent_fetch() which is for IV_CONT_CAPA, probably because we mistakenly regard global handle as a normal open handle when fetching IV_CONT_CAPA? (see cont_iv_hdl_fetch(), it's called by common request handler and doesn't distinguish global handle from normal handle).

Clearing the global handles in pool_iv_ent_invalid() looks not safe to me too, it looks to me this IV invalidate function could be called when IV request failed? Given that global handles are immutable now, I'm wondering if we still need to invalidate them?

Just raise some concerns here, they could be addressed in a separate ticket.

NiuYawei (dismissed):

src/pool/srv_pool.c
liw: @NiuYawei, agreed. When writing this patch, this IV code was the most difficult part to investigate. I tried to 1) maintain the current algorithm for the "2.8 code hosting 2.6 pool" case, and 2) avoid making big changes to code I don't confidently understand.

One change I'm confident about is that we need a new IV operation semantics: Update/invalidate an IV on the local node always synchronously, while update/invalidate it synchronously or asynchronously based on flags. Then, this trick here would no longer be necessary. Such semantics would also benefit other cases on a PS leader.

In 3.0, we will be able to drop the current algorithm because 2.6 pools will no longer be supported. We can gradually remove the complexities during (or even before) the 3.0 development.

kccain (approved):
If practical, it may be worth exercising this patch with the full regression tests in server/metadata.py since there is some checking against a specific number of resources (daos containers) being created before rdb gets an out of space error. Maybe Features: metadata? Might need to consider running that in a separate debug pr rather than changing this one though. Or (if it's even manageable with the latest code), running the server/metadata.py tests on a developer system manually with launch.py.

liw:
> If practical, it may be worth exercising this patch with the full regression tests in server/metadata.py since there is some checking against a specific number of resources (daos containers) being created before rdb gets an out of space error. Maybe Features: metadata? Might need to consider running that in a separate debug pr rather than changing this one though. Or (if it's even manageable with the latest code), running the server/metadata.py tests on a developer system manually with launch.py.

@kccain, I've created https://github.com/daos-stack/daos/pull/15916 for this purpose. By the way, do you know if this set of tests are passing consistently or otherwise?

kccain:
> > If practical, it may be worth exercising this patch with the full regression tests in server/metadata.py since there is some checking against a specific number of resources (daos containers) being created before rdb gets an out of space error. Maybe Features: metadata? Might need to consider running that in a separate debug pr rather than changing this one though. Or (if it's even manageable with the latest code), running the server/metadata.py tests on a developer system manually with launch.py.
> 
> @kccain, I've created #15916 for this purpose. By the way, do you know if this set of tests are passing consistently or otherwise?

Sounds good. I wasn't familiar with the latest pass/fail status, so I checked over here https://daosio.atlassian.net/wiki/spaces/DC/pages/11389534250/Master+Weekly+Test+Results and found there is one outstanding ticket for the test_metadata_addremove test in that .py file https://daosio.atlassian.net/browse/DAOS-16619 . I think the most important tests here would be the test_metadata_fillup_* ones.

liw:
> Sounds good. I wasn't familiar with the latest pass/fail status, so I checked over here https://daosio.atlassian.net/wiki/spaces/DC/pages/11389534250/Master+Weekly+Test+Results and found there is one outstanding ticket for the test_metadata_addremove test in that .py file https://daosio.atlassian.net/browse/DAOS-16619 . I think the most important tests here would be the test_metadata_fillup_* ones.

I see. Thank you, Ken.

liw:
Merge with master to fix a conflict in one log message.

kccain (approved):

src/pool/srv_pool.c
```diff
@@ -2358,27 +2407,33 @@ pool_svc_step_up_cb(struct ds_rsvc *rsvc)
...
+			/* Only copy server handle to make is_pool_from_srv() check correctly, and
+			 * container server handle will not be copied here, otherwise
+			 * ds_pool_iv_refresh_hdl will not open the server container handle.
+			 */
+			uuid_copy(svc->ps_pool->sp_srv_pool_hdl, srv_pool_hdl);
+		}
 	}
 
-	rc = ds_pool_iv_srv_hdl_update(svc->ps_pool, pool_hdl_uuid, cont_hdl_uuid);
+	rc = ds_pool_iv_srv_hdl_update(svc->ps_pool, srv_pool_hdl, srv_cont_hdl);
```
liuxuezhao, line 2429: @liw if change ds_pool_iv_srv_hdl_update()'s pool_iv_update(..., CRT_IV_SYNC_EAGER, ...) (change CRT_IV_SYNC_LAZY to CRT_IV_SYNC_EAGER), can the problem mentioned by @NiuYawei be avoided?

liuxuezhao (dismissed):

src/pool/srv_pool.c
liw: @liuxuezhao, if we use an eager update, then the PS will finish stepping up only after the update completes successfully. In this time window, the PS will remain inaccessible, that is, even `dmg pool query --health-only` will not work, I think. Is that right?

liw: As mentioned above, a new locally synchronous but remotely asynchronous semantics would be ideal.

liuxuezhao: the PS leader switch already with a period time, EAGER will adds one sync bcast time that seems probably will not change too much if considering the switch time?

if the case that after leader stepup, "what if an IV or whatever request reached leader and called ds_cont_hdl_rdb_lookup() before the sp_srv_cont_hdl is initialized"? Seems it possibly will cause rebuild fail?

"a new locally synchronous but remotely asynchronous semantics would be ideal."
before that available, do you think EAGER could be a temporary workaround?

liw: I'm not worrying about one sync bcast time; I'm thinking about the case where this bcast times out (Or hang? If we don't retry it, then no hang.). E.g., the system map is not fully in sync, or the pool map is not fully in sync.

liw: Has this caused rebuild issues already? If so we can discuss if there's short-term way to change the IV code to address the issues?

liuxuezhao: > Has this caused rebuild issues already? If so we can discuss if there's short-term way to change the IV code to address the issues?

No, just by analysis that once the PS leader step up, rebuild or EC agg possibly lookup/use the handle and fail.

src/pool/srv_pool.c
wangdi1: @NiuYawei I think even if sp_srv_cont_hdl is not initialized in time on the ps leader, then this check in pool_iv_ent_fetch should return NOTleader to let it retry. or I miss your question?

        if (dss_self_rank() == entry->ns->iv_master_rank) {
                /* The PS leader might still in initialization phase, the IV entry may
                 * not be fully initialized yet.
                 */
                if (!entry->iv_valid) {
                        D_INFO(DF_UUID" master %u is still stepping up: %d.\n",
                               DP_UUID(entry->ns->iv_pool_uuid), entry->ns->iv_master_rank,
                               -DER_NOTLEADER);
                        return -DER_NOTLEADER;
                }
        }

liuxuezhao: @liw for the IV problem, I pushed a small PR https://github.com/daos-stack/daos/pull/15978 based on your branch. I think it can resolve the IV semantic problem, please check if it is fine can just merge to your branch?

kccain (approved):

NiuYawei (approved):

PR #15848, opened 2025-02-05
DAOS-16979 control: Reduce frequency of hugepage allocation at runtime

tanabarr:
Test-tag-hw-medium: pr daily_regression
Allow-unstable-test: true

### Gatekeeper:

### DIFF
--- a/src/control/server/config/server.go
+++ b/src/control/server/config/server.go
@@ -1,5 +1,6 @@
 //
 // (C) Copyright 2020-2024 Intel Corporation.
+// (C) Copyright 2025 Hewlett Packard Enterprise Development LP
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -509,6 +510,8 @@ func (cfg *Server) SetNrHugepages(log logging.Logger, mi *common.MemInfo) error 
 	if cfg.DisableHugepages {
 		return FaultConfigHugepagesDisabledWithBdevs
 	}
+
+	log.Infof("%d total hugepages in system", mi.HugepagesTotal)
 
 	// Calculate minimum number of hugepages for all configured engines.
 	minHugepages, err := storage.CalcMinHugepages(mi.HugepageSizeKiB, cfgTargetCount+sysXSCount)


tanabarr:
@phender as discussed in order to try to reproduce the DMA grow failure (DAOS-16979) related to hugepage fragmentation I ran this PR with `Test-tag-hw-medium: pr daily_regression` to try to trigger the failure. Unfortunately https://build.hpdd.intel.com/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-15848/1/pipeline/ didn't hit the issue. any other ideas on how to get a baseline to prove a fix? or any other approaches like just landing a fix and seeing if it has the desired result? I'm going to try to generate a local reproducer as well.

src/control/server/config/server.go
```diff
@@ -502,37 +508,112 @@ func (cfg *Server) SetNrHugepages(log logging.Logger, mi *common.MemInfo) error
...
+
+	var msgSysXS string
+	if sysXSCount > 0 {
+		msgSysXS = fmt.Sprintf(" and %d/%d sys-xstreams", sysXSCount, largeSysXSCount)
+	}
+	log.Tracef("calculated min/max %d/%d nr_hugepages based on %d/%d targets%s",
+		minHugepages, maxHugepages, cfgTargetCount, largeTargetCount, msgSysXS)
+
+	if minHugepages > maxHugepages {
+		log.Debugf("config hugepage requirements exceed normal maximum")
```
mjmac, line 544: Should this be logged at NOTICE level? Who is it for?

src/control/server/config/server.go
```diff
@@ -502,37 +508,112 @@ func (cfg *Server) SetNrHugepages(log logging.Logger, mi *common.MemInfo) error
...
+
+	if cfg.DisableHugepages {
+		if cfg.NrHugepages != 0 {
+			return FaultConfigHugepagesDisabledWithNrSet
+		}
+		if cfg.GetBdevConfigs().HaveRealNVMe() {
+			return FaultConfigHugepagesDisabledWithBdevs
+		}
+		if minHugepages != 0 {
+			log.Noticef("hugepages disabled but targets will be assigned to bdevs, " +
```
mjmac, line 570: I don't disagree with doing something here, but logging "caution is advised" is not particularly helpful, IMO. Is it an error or not? What is the admin supposed to do if/when they happen to notice this message in the server log?

mjmac (commented):
If I understand the ticket and the PR changes correctly, this patch aims to reduce CI instability by reserving a hard-coded maximum amount of hugepages at startup, regardless of the number of bdevs in the configuration.

If that is correct, then I'm not sure I agree with the approach. Changing the product code to work around quirks of the CI testing process seems wrong to me. Instead, it seems like it would be best to ensure that the first configuration of the server has the maximum number of hugepages allocated for the rest of the run.

src/control/server/config/server.go
tanabarr: this is just a debug message because the user doesn't need to do anything about it, it is an indication that the configuration requires more huge pages than the normal maximum

src/control/server/config/server.go
tanabarr: this is to indicate that the server is operating in an unusual mode, the administrator should be aware of that

tanabarr:
> If I understand the ticket and the PR changes correctly, this patch aims to reduce CI instability by reserving a hard-coded maximum amount of hugepages at startup, regardless of the number of bdevs in the configuration.
> 
> If that is correct, then I'm not sure I agree with the approach. Changing the product code to work around quirks of the CI testing process seems wrong to me. Instead, it seems like it would be best to ensure that the first configuration of the server has the maximum number of hugepages allocated for the rest of the run. This is a CI problem rather than a product problem, and therefore the solution should be implemented in the test harness, IMO.

I don't think this issue is restricted just to CI, IIRC this has been seen outside of our test infrastructure. @NiuYawei requested this change so maybe it's appropriate that he responds to your objection.

mjmac:
> > > If I understand the ticket and the PR changes correctly, this patch aims to reduce CI instability by reserving a hard-coded maximum amount of hugepages at startup, regardless of the number of bdevs in the configuration.
> > > If that is correct, then I'm not sure I agree with the approach. Changing the product code to work around quirks of the CI testing process seems wrong to me. Instead, it seems like it would be best to ensure that the first configuration of the server has the maximum number of hugepages allocated for the rest of the run. This is a CI problem rather than a product problem, and therefore the solution should be implemented in the test harness, IMO.
> > 
> > 
> > I don't think this issue is restricted just to CI, IIRC this has been seen outside of our test infrastructure. @NiuYawei requested this change so maybe it's appropriate that he responds to your objection.
> 
> FWIW I've seen similar on Aurora after a fresh reboot: https://daosio.atlassian.net/browse/DAOS-16921?focusedCommentId=135440 And I've only seen that with master, not 2.6.

I don't doubt that there is a problem... My concern is more that it seems like the actual problem is not yet understood, and the proposed approach in this PR is a solution for a very specific set of scenarios. Adding a hard-coded configuration for hugepages kind of defeats the purpose of having a configuration mechanism, and it seems likely to cause unintended problems for configurations that are outside of what's being hard-coded in this PR.

NiuYawei:
> > > > If I understand the ticket and the PR changes correctly, this patch aims to reduce CI instability by reserving a hard-coded maximum amount of hugepages at startup, regardless of the number of bdevs in the configuration.
> > > > If that is correct, then I'm not sure I agree with the approach. Changing the product code to work around quirks of the CI testing process seems wrong to me. Instead, it seems like it would be best to ensure that the first configuration of the server has the maximum number of hugepages allocated for the rest of the run. This is a CI problem rather than a product problem, and therefore the solution should be implemented in the test harness, IMO.
> > > 
> > > 
> > > I don't think this issue is restricted just to CI, IIRC this has been seen outside of our test infrastructure. @NiuYawei requested this change so maybe it's appropriate that he responds to your objection.
> > 
> > 
> > FWIW I've seen similar on Aurora after a fresh reboot: https://daosio.atlassian.net/browse/DAOS-16921?focusedCommentId=135440 And I've only seen that with master, not 2.6.
> 
> I don't doubt that there _is_ a problem... My concern is more that it seems like the actual problem is not yet understood, and the proposed approach in this PR is a potential solution for a very specific set of scenarios. Adding a hard-coded configuration for hugepages kind of defeats the purpose of having a configuration mechanism, and it seems likely to cause unintended problems for configurations that are outside of what's being hard-coded in this PR.

Yes, there could be other unknown issues to be solved (As @daltonbohning mentioned that allocation failure was seen after a fresh reboot when the memory isn't supposed to be fragmented), but allocating hugepages at run time (setting nr_hugepages) is believed likely generating fragmentations.

I think our goal is to avoid allocating hugepages at run time when possible, no matter for production or testing system.

src/control/server/config/server.go
```diff
@@ -502,37 +509,112 @@ func (cfg *Server) SetNrHugepages(log logging.Logger, mi *common.MemInfo) error
...
+				"this is an atypical situation and caution is advised")
+		}
+
+		// Hugepages disabled and so zero requested in config.
+		return nil
+	} else if minHugepages == 0 {
+		// Enable minimum needed for scanning NVMe on host in discovery mode.
+		if cfg.NrHugepages < scanMinHugepageCount && mi.HugepagesTotal < scanMinHugepageCount {
```
knard38, line 579: I am probably missing something but I was thinking that `mi.HugepagesTotal` was the number of available Huge Pages, and thus `cfg.NrHugePages` could not be greater than this first value.

src/control/server/server_utils.go
```diff
@@ -354,42 +345,40 @@ func prepBdevStorage(srv *server, iommuEnabled bool) error {
...
-			// request.
-			prepReq.HugepageCount = scanMinHugepageCount
+			srv.log.Debugf("allocating %d hugepages on each of these numa nodes: %v",
+				prepReq.HugepageCount, numaNodes)
 		} else {
-			// If nr_hugepages has been set manually but no bdevs in config then
-			// allocate on numa node 0 (for example if a bigger number of hugepages are
-			// required in discovery mode for an unusually large number of SSDs).
-			prepReq.HugepageCount = srv.cfg.NrHugepages
+			srv.log.Debugf("skip allocating hugepages, no change is required")
```
knard38, line 375: Should it not be an error to have bdev without huge pages ?

tanabarr: After chatting with @NiuYawei decided that we should support emulated NVMe with or without hugepages as some usage models may not require them.

knard38: Make sense.
NIT, the error `FaultConfigHugepagesDisabledWithBdevs` raised at line should probably be renamed to something such as `FaultConfigHugepagesDisabledWithNvmeBdevs`.

tanabarr: fixed

src/control/server/config/server.go
tanabarr: `mi.HugepagesTotal` refers to the number of hugepages already allocated on the system. `cfg.NrHugepages` refers to the configured and required number. This line (L579) is checking whether the configured number needs to be increased to account for a minimum value that will allow NVMe devices to be discovered in the case where the service started with no bdevs  listed in the server config file. If the number of system (already) allocated hugepages is already enough to satisfy this minimum then we don't need to set the configured value as no further allocation is necessary.

tanabarr:
I think tests have probably finished but Jenkins seems to be inaccessible because of certificate issue.

src/control/server/config/server.go
knard38: Thanks for the clarification :-)

knard38 (approved):

tanabarr:
> So if I'm understanding the motivation correctly, the problem is that people sometimes change the storage/engine config in a way that affects hugepages after booting up the system, and then restart the `daos_server`? Is that correct, or am I missing something?

Yes memory can be fragmented causing DMA buffer allocations to fail if frequent reallocation of hugepages happens. Reallocation occurs on `daos_server` start-up (currently every time) and advice from kernel focus is to allocate once to reduce the chance of fragmentation https://docs.kernel.org/admin-guide/mm/hugetlbpage.html .

> Always allocating a large number of hugepages immediately after reboot feels a bit weird to me. And it seems peculiar for a user to be twiddling those configuration knobs beyond the initial system setup/tuning--is this something that really happens?

I guess that the counterpoint to this question is that this may well be what the user does for one reason or another and if so then we want to try to prevent memory fragmentation from causing the engine to fail. One way to do this is as implemented where we allocate a large amount on first start in order to prevent reallocation requests in the majority of cases, I can't see any major downsides to doing this vs. what we have currently.

src/control/server/config/server.go
```diff
@@ -502,37 +509,112 @@ func (cfg *Server) SetNrHugepages(log logging.Logger, mi *common.MemInfo) error
...
+				humanize.IBytes(hugePageBytes(cfg.NrHugepages, mi.HugepageSizeKiB)))
+		}
+	} else {
+		log.Infof("nr_hugepages requested manually-set to %d (%s)", cfg.NrHugepages,
+			humanize.IBytes(hugePageBytes(cfg.NrHugepages, mi.HugepageSizeKiB)))
+		if cfg.NrHugepages < minHugepages {
+			log.Noticef("configured nr_hugepages %d is less than recommended %d, "+
+				"if this is not intentional update the 'nr_hugepages' config "+
+				"parameter or remove and it will be automatically calculated",
+				cfg.NrHugepages, minHugepages)
```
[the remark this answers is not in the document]
tanabarr: I don't think we want to mandate the value that the user can set. There maybe resource constrained environments where operation with lower DMA buffer values are required. The 1G per target guidelines are only recommendations based on expected workloads, the actual usage is variable.

src/control/server/config/server.go
tanabarr: The numbers are included in the trace level message, it doesn't indicate a problem only that the calculated value based on the config exceeds the normal upper bound used for the initial allocation.

kjacque: In theory someone could see this debug-level message without seeing the trace-level message though. I think you could either make both messages the same log level, or include the min and max values in this message.

src/control/server/config/server.go
kjacque: If emulated NVMe can go either way, and we can't detect specifics, we could log a more specific, helpful message, for example: "Hugepages are explicitly disabled in this configuration file. Hugepages may be required depending on the underlying storage technology for the emulated SSDs. If you experience problems with this configuration, try...", etc.

daltonbohning:
> Sure, I get that part. Is the fragmentation potentially happening even if the user didn't change their config? I guess that's what I'm trying to understand, what user behavior (if any) triggers this issue.

FWIW I've seen the DMA allocation failures on Aurora on a fresh cluster, fresh reboot of the nodes. And on master but not 2.6

kjacque:
> FWIW I've seen the DMA allocation failures on Aurora on a fresh cluster, fresh reboot of the nodes. And on master but not 2.6

That implies to me that there's a bug unrelated to the issue this patch addresses. Also odd that it only happens in master, when 2.6 is also handling hugepages the old way.

Maybe there are multiple issues causing similar symptoms. I think the basic idea of this patch (only allocate hugepages if they haven't already been allocated by a previous DAOS run) makes sense regardless. We don't want to reallocate hugepages if daos_server needs to be restarted. I just don't like the high default. In odd cases where the number of targets could change (testing/tuning/experimenting), I think the user should probably set `nr_hugepages` to give themselves wiggle room.

NiuYawei:
> > FWIW I've seen the DMA allocation failures on Aurora on a fresh cluster, fresh reboot of the nodes. And on master but not 2.6
> 
> That implies to me that there's a bug unrelated to the issue this patch addresses. Also odd that it only happens in master, when 2.6 is also handling hugepages the old way.
> 
> Maybe there are multiple issues causing similar symptoms. I think the basic idea of this patch (only allocate hugepages if they haven't already been allocated by a previous DAOS run) makes sense regardless. We don't want to reallocate hugepages if daos_server needs to be restarted. I just don't like the high default. In odd cases where the number of targets could change (testing/tuning/experimenting), I think the user should probably set `nr_hugepages` to give themselves wiggle room.

Yeah, I agree there could be some other issues besides the potential fragmentation caused by the unnecessary hugepages re-allocation. One wild guess is that kernel might allocate hugepages for the two NUMA nodes in sort of interleaved mode? So that allocated pages for certain NUMA could be fragmented? @tanabarr , any thoughts? I remembered that we switched to use NUMA aware hugepage allocation in setup.sh (engine was also changed to use NUMA aware DMA alloc API) in an earlier version (probably in 2.4? Or 2.6?), I'm wondering if that's something we need to investigate further?

The reason it happens in master only is that I bumped initial DMA allocation from a fixed number (24 chunks per xstream) to 50% of total DMA buffer (64 chunks per xstream by default) in master (it's not backported to 2.6), I made this change because we observed run time DMA allocation failures on both Aurora and CI testing, the failure won't interrupt application but it does hurt performance on high workload which requires more DMA buffers, so I decided to pre-allocate more on engine start.

@tanabarr and I discussed this offline, his proposal is that we just prevents allocation if total system number is more than either calculated or what is specified in the config, and we ask admin (or change CI test/image) to allocate enough hugepages on boot to avoid fragmentations. @kjacque Does it sound reasonable to you?

kjacque:
> @tanabarr and I discussed this offline, his proposal is that we just prevents allocation if total system number is more than either calculated or what is specified in the config, and we ask admin (or change CI test/image) to allocate enough hugepages on boot to avoid fragmentations. @kjacque Does it sound reasonable to you?

This makes sense to me.

We should update the admin guide and any quickstart documentation, and maybe the server config file as well, to make sure users are aware of the potential issue with hugepage fragmentation and our recommendations to mitigate it.

tanabarr:
> > > FWIW I've seen the DMA allocation failures on Aurora on a fresh cluster, fresh reboot of the nodes. And on master but not 2.6
> > 
> > 
> > That implies to me that there's a bug unrelated to the issue this patch addresses. Also odd that it only happens in master, when 2.6 is also handling hugepages the old way.
> > Maybe there are multiple issues causing similar symptoms. I think the basic idea of this patch (only allocate hugepages if they haven't already been allocated by a previous DAOS run) makes sense regardless. We don't want to reallocate hugepages if daos_server needs to be restarted. I just don't like the high default. In odd cases where the number of targets could change (testing/tuning/experimenting), I think the user should probably set `nr_hugepages` to give themselves wiggle room.
> 
> Yeah, I agree there could be some other issues besides the potential fragmentation caused by the unnecessary hugepages re-allocation. One wild guess is that kernel might allocate hugepages for the two NUMA nodes in sort of interleaved mode? So that allocated pages for certain NUMA could be fragmented? @tanabarr , any thoughts? I remembered that we switched to use NUMA aware hugepage allocation in setup.sh (engine was also changed to use NUMA aware DMA alloc API) in an earlier version (probably in 2.4? Or 2.6?), I'm wondering if that's something we need to investigate further?
> 
> The reason it happens in master only is that I bumped initial DMA allocation from a fixed number (24 chunks per xstream) to 50% of total DMA buffer (64 chunks per xstream by default) in master (it's not backported to 2.6), I made this change because we observed run time DMA allocation failures on both Aurora and CI testing, the failure won't interrupt application but it does hurt performance on high workload which requires more DMA buffers, so I decided to pre-allocate more on engine start.
> 
> @tanabarr and I discussed this offline, his proposal is that we just prevents allocation if total system number is more than either calculated or what is specified in the config, and we ask admin (or change CI test/image) to allocate enough hugepages on boot to avoid fragmentations. @kjacque Does it sound reasonable to you?

I guess we have to see  whether we can advocate across NUMA on boot, I imagine we can. let's see the results of the PR first

phender:
What I'm seeing in https://daosio.atlassian.net/browse/DAOS-16921 is - as previously stated - it only occurs on master, but it also only occurs when running the tests on the wolf-[52,122-125] cluster (Label wolf-52_nvme5).  Due its hardware configuration this cluster runs with the ofi+tcp provider, instead of verbs and the same 4 tests always fail when run on this cluster.

knard38 (approved):

tanabarr:
@mjmac @kjacque @knard38 please take a look/review when you can, apologies for force push as the base changed.

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/17/execution/node/1436/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/17/execution/node/1422/log

daosbuild3:
Test stage Unit Test on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15848/18/testReport/

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15848/19/testReport/

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15848/19/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15848/19/testReport/

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/21/execution/node/1419/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/21/execution/node/1394/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/21/execution/node/1427/log

daosbuild3:
Test stage Unit Test on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15848/23/testReport/

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-15848/24/testReport/

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/24/execution/node/1069/log

src/control/server/config/server.go
```diff
@@ -503,42 +506,130 @@ func (cfg *Server) SetNrHugepages(log logging.Logger, smi *common.SysMemInfo) er
...
+		}
 	}
 
 	return nil
 }
 
+// GetNumaNodes returns in use NUMA nodes based on engine configurations. Detects the number of
+// engine configs assigned to each NUMA node and return error if engines are distributed unevenly
```
mjmac, line 620: Hmm. I'm a little concerned about hard-coding the requirement that engines must be evenly distributed across NUMA nodes. There are some hardware configurations where an imbalance may be unavoidable. I agree that it is a good idea to detect and call this out as a suboptimal configuration, but if the default behavior is to return an error, then there ought to be a way to override it in cases where the drawbacks are understood and accepted.

tanabarr:
> Taking a look at this again for the first time in a couple of months... Putting it bluntly, if it works, it works. It's pretty hard to follow and review from a correctness standpoint, though. One thing that strikes me about the whole architecture is that there are so many layers involved that any changes always result in wide ripples across the codebase. Server logic, server config, server CLI, storage layer, etc.
> 
> It occurs to me that a better architecture might result from redesigning and refactoring to compartmentalize as much of the responsibility for hugepage management in one spot. So the server doesn't necessarily need to know any details about hugepages. The server config is fed into the hugepage manager so that it knows what the configuration is, and then the storage layers accept an instance of the hugepage manager which can tell them what they need to know.
> 
> With that architecture in place, future changes to hugepage management should be fairly self-contained, easier to reason about, and therefore easier to review and maintain. I understand that direction may not be preferable, but it seems like this hugepage management logic is being revisited frequently enough to justify the investment in cleaning things up now.

I would like to land this to attempt to resolve the long-standing issues then look at re-factoring as a subsequent piece of work. It's a high-priority issue that needs to be addressed urgently.

src/control/server/config/server.go
tanabarr: sure, will implement an override

tanabarr:
@mjmac @kjacque I have tidied up the PR to focus on the area of concern. I think it will now be easier to review and reason.

Main purpose:

Avoid shrinking and growing per-NUMA allocations on daos_server start which causes memory fragmentation over server boot lifetime.

Main changes:

SetNrHugepages
- sets the value in the server configuration
- uses existing number if already set in config
- otherwise calculates the value based on DisableHugepages flag, number of targets and/or minimum number of pages required to scan ssds

SetHugeNodes
- ensures sufficient pages exist on each NUMA-node engines are assigned to
- maintains allocations; if there are already sufficient pages on a NUMA-node, request the existing count to avoid shrinking allocation
- supplies HUGENODE string to spdk/scripts/setup.sh
- used once in cmd/daos_server:prepareNVMe and once in server:prepBdevStorage

setEngineMemSize
- verifies that free hugepage count satisfies what is required based on configured targets
- does not enforce numa alignment of pages, uses global count
- sets mem_size cap passed to engine on fork

I will address @mjmac 's concern about allowing NUMA imbalance.

Please review again, many thanks

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/28/execution/node/1394/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/28/execution/node/1338/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/28/execution/node/1534/log

src/control/server/config/server.go
```diff
@@ -503,42 +513,132 @@ func (cfg *Server) SetNrHugepages(log logging.Logger, smi *common.SysMemInfo) er
...
+		return nil
+	} else if minHugepages == 0 {
+		if cfg.NrHugepages < ScanMinHugepageCount {
+			log.Infof("No hugepages required as zero configured engine targets, setting "+
+				"minimum (%d) in config to enable NVMe device discovery",
+				ScanMinHugepageCount)
+			cfg.NrHugepages = ScanMinHugepageCount
+		} else {
+			log.Infof("No hugepages required as zero configured engine targets, "+
+				"configurd value (%d) will be used to enable NVMe device discovery",
```
kjacque: nit - typo
```suggestion
				"configured value (%d) will be used to enable NVMe device discovery",
```

src/control/server/config/server.go
kjacque: I wonder if it would be better just to log it as a Notice, rather than refuse to start up.

kjacque (approved):
I'm fine with this approach. Comments are minor.

tanabarr:
To try to verify PR-15848 fixes DAOS-16921:

```
Test-tag: PoolCreateAllHwTests PoolCreateCapacityTests RbldNoCapacity RbldWithIO
FUNCTIONAL_HARDWARE_MEDIUM_LABEL = opa-10_nvme5
```

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/29/execution/node/1071/log

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-15848/31/display/redirect

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/32/execution/node/1021/log

knard38 (approved):
LGTM

tanabarr:
Despite some strange issue with merge conflicts, PR looks like it's addressing the issue at hand.

Details being added to ticket.

src/control/server/config/server.go
tanabarr: Notices get ignored and on bare-metal we really don't want to be running with NUMA imbalance as it causes all sorts of DAOS performance issues.

src/control/server/config/server.go
tanabarr: 87566307f956a7de60b44e9154fcec651d84f0fa contains this update

src/control/server/server_utils.go
```diff
@@ -271,37 +263,109 @@ func getFabricNetDevClass(cfg *config.Server, fis *hardware.FabricInterfaceSet)
...
+			}
+			// Ensure that if there is already sufficient number allocated then the
+			// script will request the existing number which results in a no-op.
+			//
+			// FIXME DAOS-16921: SPDK https://review.spdk.io/c/spdk/spdk/+/25831 adds
+			//                   SKIP_HUGE which can be used to simplify this logic.
+			if nn.HugepagesTotal >= perNumaNrWant {
+				nodeNrs[nID] = nn.HugepagesTotal // Maintain
+			} else {
+				nodeNrs[nID] = perNumaNrWant // Grow
```
tanabarr, line 293: 87566307f956a7de60b44e9154fcec651d84f0fa contains an update to log notice when the number of hugepages on a NUMA is being increased.

phender:
> Test stage Functional Hardware Large MD on SSD completed with status FAILURE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/32/execution/node/1021/log

This looks like https://daosio.atlassian.net/browse/DAOS-17731

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/36/execution/node/754/log

src/control/server/config/server.go
tanabarr: corrected here instead

tanabarr:
@kjacque @knard38 apologies for the force push, I tried to merge master without it but the commits from the previous base branch would not go away. The only way to get GH UI to show a clean apply of this PRs commits was to rebase and force push.
No changes were made since last review, simply resolved conflicts without any functional code changes. please rereview, TIA.

tanabarr:
CI run with priority: 2 and full_regression control pr test tags
OPA cluster verification performed in previous runs

kjacque (dismissed):

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/42/execution/node/990/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/42/execution/node/900/log

phender:
> Test stage Functional Hardware Medium MD on SSD completed with status FAILURE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/42/execution/node/900/log

The test failures with known issues:
- 1-./container/full_pool_container_create.py:FullPoolContainerCreate.test_no_space_cont_create - known issue https://daosio.atlassian.net/browse/DAOS-17777
- 1-./daos_vol/bigio.py:DaosVolBigIO.test_daos_vol_bigio - known issue https://daosio.atlassian.net/browse/DAOS-16618
- 1-./dfuse/sparse_file.py:SparseFile.test_sparsefile - known issue https://daosio.atlassian.net/browse/SRE-3152


Test failures w/o known issues:
- 4-./container/per_server_fault_domain.py:PerServerFaultDomainTest.test_rf2_unclean
- 1-./rebuild/mdtest.py:RebuildMdtest.test_rebuild_mdtest
but these two failures where test timeouts waiting for rebuild to complete.

NiuYawei (dismissed):

tanabarr:
> > Test stage Functional Hardware Medium MD on SSD completed with status FAILURE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/42/execution/node/900/log
> 
> The test failures with known issues:
> 
> * 1-./container/full_pool_container_create.py:FullPoolContainerCreate.test_no_space_cont_create - known issue https://daosio.atlassian.net/browse/DAOS-17777
> * 1-./daos_vol/bigio.py:DaosVolBigIO.test_daos_vol_bigio - known issue https://daosio.atlassian.net/browse/DAOS-16618
> * 1-./dfuse/sparse_file.py:SparseFile.test_sparsefile - known issue https://daosio.atlassian.net/browse/SRE-3152
> 
> Test failures w/o known issues:
> 
> * 4-./container/per_server_fault_domain.py:PerServerFaultDomainTest.test_rf2_unclean
> * 1-./rebuild/mdtest.py:RebuildMdtest.test_rebuild_mdtest
>   but these two failures where test timeouts waiting for rebuild to complete.

Requesting force landing based on the following rationale:

- https://jenkins.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-15848/32/ was run with
`Test-tag: PoolCreateAllHwTests PoolCreateCapacityTests RbldNoCapacity RbldWithIO` and `FUNCTIONAL_HARDWARE_MEDIUM_LABEL = opa-10_nvme5`
to verify PR-15848 fixes DAOS-16921 which occurs on opa clusters with tcp provider and passed all.

- @phender 's analysis on the failures which are either known issues or rebuild related and unlikely to be directly related to hugepage allocation changes. https://jenkins.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-15848/42 was run with `Test-tag: full_regression control pr` to get wide test coverage.

knard38 (dismissed):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-15848/43/execution/node/781/log

tanabarr:
https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/view/change-requests/job/PR-15848/43/

Of the failures, Hdf5 seems unrelated to me and MSMembershipTest is ramdisk sizing not been met by available memory which probably needs system_ram_reserved server config file parameter update for that specific test, this might be related to the hugepage allocation algorithm changes in the PR but can possibly be addressed by a follow on PR. @phender do you think the coverage is sufficient to land the PR?

phender:
> https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/view/change-requests/job/PR-15848/43/
> 
> Of the failures, Hdf5 seems unrelated to me and MSMembershipTest is ramdisk sizing not been met by available memory which probably needs system_ram_reserved server config file parameter update for that specific test, this might be related to the hugepage allocation algorithm changes in the PR but can possibly be addressed by a follow on PR. @phender do you think the coverage is sufficient to land the PR?

Can you create that ticket please.

Also, for the `1-./mpiio/hdf5.py:Hdf5.test_hdf5` test failure I've reopened https://daosio.atlassian.net/browse/DAOS-17654

tanabarr:
> > https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/view/change-requests/job/PR-15848/43/
> > Of the failures, Hdf5 seems unrelated to me and MSMembershipTest is ramdisk sizing not been met by available memory which probably needs system_ram_reserved server config file parameter update for that specific test, this might be related to the hugepage allocation algorithm changes in the PR but can possibly be addressed by a follow on PR. @phender do you think the coverage is sufficient to land the PR?
> 
> Can you create that ticket please.
> 
> Also, for the `1-./mpiio/hdf5.py:Hdf5.test_hdf5` test failure I've reopened https://daosio.atlassian.net/browse/DAOS-17654

https://daosio.atlassian.net/browse/DAOS-17790

daltonbohning:
This conflicts with the now-merged #16264

tanabarr:
conflicts resolved, no functional change, requesting gatekeeper

tanabarr:
Approvals got wiped from resolving merge conflicts but given no functional code changes were made, requesting landing: https://hpe-internal.slack.com/archives/C08PU15E971/p1752499015932589

PR #15870, opened 2025-02-07
DAOS-17006 cart: Publish Mercury counters as metrics

mjmac:
When Mercury has been built with diagnostic RPC counters
enabled, CaRT will periodically republish the counters as DAOS
telemetry for consumption by monitoring infrastructure.

Change-Id: I3b0bcb260ad970798ac1cd838f8469c4cfbede55


mjmac:
@soumagne: FYI... Do you see any major problems with the approach? I think the additional overhead for this should be pretty small.

src/cart/crt_hg.c
```diff
@@ -903,6 +903,76 @@ crt_hg_class_init(crt_provider_t provider, int ctx_idx, bool primary, int iface_
...
+		D_ERROR("hg_ctx is NULL.\n");
+		return;
+	}
+
+	prov    = crt_provider_name_get(hg_ctx->chc_provider);
+	metrics = &hg_ctx->chc_metrics;
+
+	rc = d_tm_add_metric(&metrics->chm_bulks, D_TM_COUNTER,
+			     "Mercury-layer count of bulk transfers", "bulks",
+			     "net/%s/hg/bulks/ctx_%u", prov, idx);
```
frostedcmos, line 923: Would need to either handle multiprovider case here or add todo for it.
for multiprovider situation the name would need to have a form of  'net/[provider secondary idx]/[provider_name]/hg/bulks/[ctx_idx]' as the same provider can be used both as a primary and a secondary provider.

frostedcmos (commented):
Based on the skype discussions I was under the impression that we also require 'reset counters' call to be performed after each query -- is the plan different now on how counters will be consumed in the end?

One other comment inline.
LGTM otherwise.

src/cart/crt_hg.c
mjmac: Hmm. I guess that comment also applies to the [existing CaRT-level metrics](https://github.com/daos-stack/daos/blob/master/src/cart/crt_context.c#L280)?

Can we handle all of that in a standalone PR that is specifically for multiprovider metrics?

mjmac: @kjacque: FYI

frostedcmos: Indeed it would also apply to already existing cart level metrics, so yeah probably best to handle it all in a standalone pr later

frostedcmos (approved):

src/cart/crt_hg.c
kjacque: This is [DAOS-8704](https://daosio.atlassian.net/issues/DAOS-8704) btw. I will address them all with that ticket.

frostedcmos (dismissed):

PR #15871, opened 2025-02-08
DAOS-16940 common: Update pool map levels

kjacque:
- Rename pool map levels to allow for optional fault domain level above
  node.
- Rename any performance domain "group" terminology.
- Update compressed fault domain tree parsing to include metadata about
  whether to include optional fault or performance domains.
- Update compressed domain tree format in control plane.

Signed-off-by: Kris Jacque <kris.jacque@hpe.com>

### Gatekeeper:

### DIFF
--- a/src/common/fault_domain.c
+++ b/src/common/fault_domain.c
@@ -1,8 +1,10 @@
 /**
  * (C) Copyright 2021-2023 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
+#define D_LOGFAC DD_FAC(common)
 
 #include "fault_domain.h"
 
@@ -16,26 +18,40 @@
 #define FD_TREE_TUPLE_LEN	3
 
 /**
+ * A compressed domain tree starts with values that encode metadata about the tree.
+ */
+#define FD_TREE_MD_LEN          1
+
+/**
+ * The compressed tree must contain metadata and at least a root node.
+ */
+#define FD_TREE_MIN_LEN         (FD_TREE_TUPLE_LEN + FD_TREE_MD_LEN)
+
+/**
  * The fault domain tree level numbering begins at 0 (rank level) and increases
  * as you approach the root. The level above the ranks will always have the same
  * number.
  *
  * This assumes the tree has a uniform depth throughout.
  */
-#define LAST_DOMAIN_LEVEL	1
+#define NODE_DOMAIN_LEVEL       1
+
+/**
+ * The root node always has a static ID number.
+ */
+#define ROOT_ID                 1
 
 int
-d_fd_tree_init(struct d_fd_tree *tree, const uint32_t *compressed,
-	       const uint32_t compressed_len)
+d_fd_tree_init(struct d_fd_tree *tree, const uint32_t *compressed, const uint32_t compressed_len)
 {
 	if (compressed == NULL) {
 		D_ERROR("null compressed fd tree\n");
 		return -DER_INVAL;
 	}
 
-	if (compressed_len < FD_TREE_TUPLE_LEN) {
-		D_ERROR("compressed len=%u, less than minimum %u\n",
-			compressed_len, FD_TREE_TUPLE_LEN);
+	if (compressed_len < FD_TREE_MIN_LEN) {
+		D_ERROR("compressed len=%u, less than minimum %u\n", compressed_len,
+			FD_TREE_MIN_LEN);
 		return -DER_INVAL;
 	}
 
@@ -50,6 +66,58 @@ d_fd_tree_init(struct d_fd_tree *tree, const uint32_t *compressed,
 	return d_fd_tree_reset(tree);
 }
 
+static bool
+tree_has_fault_domain(struct d_fd_tree *tree)
+{
+	return (tree->fdt_compressed[0] & D_FD_TREE_HAS_FAULT_DOMAIN) != 0;
+}
+
+static bool
+tree_has_perf_domain(struct d_fd_tree *tree)
+{
+	return (tree->fdt_compressed[0] & D_FD_TREE_HAS_PERF_DOMAIN) != 0;
+}
+
+static bool
+need_perf_dom(struct d_fd_tree *tree)
+{
+	return tree_has_perf_domain(tree) && tree->fdt_perf_dom_level < 0;
+}
+
+static bool
+need_fault_dom(struct d_fd_tree *tree)
+{
+	return tree_has_fault_domain(tree) && tree->fdt_fault_dom_level < 0;
+}
+
+static bool
+domain_is_root(struct d_fault_domain *dom)
+{
+	return dom->fd_id == ROOT_ID;
+}
+
+static bool
+domain_is_node(struct d_fault_domain *dom)
+{
+	return dom->fd_level == NODE_DOMAIN_LEVEL;
+}
+
+static bool
+domain_is_fault(struct d_fd_tree *tree, struct d_fault_domain *dom)
+{
+	/* Performance domain must be higher in tree than fault domain */
+	if (need_perf_dom(tree))
+		return false;
+
+	return need_fault_dom(tree) || tree->fdt_fault_dom_level == dom->fd_level;
+}
+
+static bool
+domain_is_perf(struct d_fd_tree *tree, struct d_fault_domain *dom)
+{
+	return need_perf_dom(tree) || tree->fdt_perf_dom_level == dom->fd_level;
+}
+
 static int
 get_next_domain(struct d_fd_tree *tree, struct d_fd_node *next)
 {
@@ -61,16 +129,32 @@ get_next_domain(struct d_fd_tree *tree, struct d_fd_node *next)
 	}
 
 	fd = (struct d_fault_domain *)&tree->fdt_compressed[tree->fdt_idx];
-	next->fdn_type = D_FD_NODE_TYPE_DOMAIN;
+	if (domain_is_root(fd)) {
+		next->fdn_type = D_FD_NODE_TYPE_ROOT;
+	} else if (domain_is_node(fd)) {
+		next->fdn_type = D_FD_NODE_TYPE_NODE;
+	} else if (domain_is_fault(tree, fd)) {
+		next->fdn_type = D_FD_NODE_TYPE_FAULT_DOM;
+		if (need_fault_dom(tree))
+			tree->fdt_fault_dom_level = fd->fd_level;
+	} else if (domain_is_perf(tree, fd)) {
+		next->fdn_type = D_FD_NODE_TYPE_PERF_DOM;
+		if (need_perf_dom(tree))
+			tree->fdt_perf_dom_level = fd->fd_level;
+	} else {
+		D_ERROR("fault domain tree has a node of unknown type (level=%d, id=%d, "
+			"children=%d)",
+			fd->fd_level, fd->fd_id, fd->fd_children_nr);
+		return -DER_INVAL;
+	}
 	next->fdn_val.dom = fd;
-
 	tree->fdt_domains_found++;
 
 	/*
-	 * At the final level, we expect the children to be ranks,
+	 * At the node level, we expect the children to be ranks,
 	 * not domains.
 	 */
-	if (fd->fd_level == LAST_DOMAIN_LEVEL)
+	if (domain_is_node(fd))
 		tree->fdt_ranks_expected += fd->fd_children_nr;
 	else
 		tree->fdt_domains_expected += fd->fd_children_nr;
@@ -148,11 +232,13 @@ d_fd_tree_reset(struct d_fd_tree *tree)
 		return -DER_UNINIT;
 	}
 
-	tree->fdt_idx = 0;
+	tree->fdt_idx              = FD_TREE_MD_LEN; /* skip the metadata values */
 	tree->fdt_domains_expected = 1; /* at least the root */
 	tree->fdt_domains_found = 0;
 	tree->fdt_ranks_expected = 0;
 	tree->fdt_ranks_found = 0;
+	tree->fdt_perf_dom_level   = -1;
+	tree->fdt_fault_dom_level  = -1;
 
 	return 0;
 }
@@ -165,7 +251,7 @@ d_fd_get_exp_num_domains(uint32_t compressed_len, uint32_t exp_num_ranks,
 	uint32_t domain_len;
 
 	/* Minimal tree must contain at least the root domain */
-	min_len = FD_TREE_TUPLE_LEN + exp_num_ranks;
+	min_len = FD_TREE_MD_LEN + FD_TREE_TUPLE_LEN + exp_num_ranks;
 
 	if (compressed_len < min_len) {
 		D_ERROR("len = %u, needed minimum = %u)\n", compressed_len,
@@ -173,7 +259,7 @@ d_fd_get_exp_num_domains(uint32_t compressed_len, uint32_t exp_num_ranks,
 		return -DER_INVAL;
 	}
 
-	domain_len = compressed_len - exp_num_ranks;
+	domain_len = compressed_len - exp_num_ranks - FD_TREE_MD_LEN;
 	if (domain_len % FD_TREE_TUPLE_LEN != 0) {
 		D_ERROR("domain_len = %u is not a multiple of %u\n",
 			domain_len, FD_TREE_TUPLE_LEN);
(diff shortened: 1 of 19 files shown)


kjacque:
@liuxuezhao @gnailzenh This begins to implement the changes we discussed a few months ago. Does it all make sense to you? I am adding "node" as a hard coded level, as well as allowing another "fault" level to be defined above that, if the fault domain for the pool is not node. I will need some help to finish adding support for the FAULT level as implementation progresses.

While I am making the node level explicit here, it is worth noting that in 2.6 and earlier, "node" does not always mean "node"... It is user defined. I still feel a bit mixed about this representation as a result... A simpler version of this patch could just rename NODE to FAULT, limiting containers to using either the pool's fault domain or RANK (without NODE as a third option).

@johannlombardi Originally we talked about passing the whole arbitrary fault domain tree down into the pool map, but that required a lot of complex changes. Essentially the goal of this approach is to minimize changes to the engine's pool map representation, while allowing choice of fault domain during pool create. In the final feature, the user will select a fault level for the pool during pool create, and the control plane will pass down a domain tree representation for the pool map that includes: root, perf domain (optional), fault domain above node (optional), and node.

kjacque:
I've created an alternate version of this PR, #15965, to show what it looks like to simply rename the existing levels rather than add a new level in the engine.

@liuxuezhao @gnailzenh @johannlombardi Please use that PR as a comparison and let me know which approach you prefer.

src/include/daos/pool_map.h
```diff
@@ -30,14 +31,15 @@
...
-	PO_COMP_TP_NODE		= 2, /** for test only */
-	PO_COMP_TP_GRP		= 3, /** group, commonly used for performance domain */
-	PO_COMP_TP_MAX		= 254, /** last user-defined domain */
-	PO_COMP_TP_ROOT		= 255,
-	PO_COMP_TP_END		= 256,
+	PO_COMP_TP_TARGET = 0,   /** reserved, hard-coded */
+	PO_COMP_TP_RANK   = 1,   /** reserved, hard-coded */
+	PO_COMP_TP_NODE   = 2,   /** reserved, hard-coded */
+	PO_COMP_TP_MIN    = 3,   /** first user-defined domain */
+	PO_COMP_TP_FAULT  = 3,   /** user-defined fault domain (if not node) */
```
mjmac, line 38: Should this be 4?

src/include/daos_prop.h
```diff
@@ -190,8 +190,8 @@ daos_svc_rf_is_valid(uint64_t svc_rf)
  * Level of perf_domain, should be same value as PO_COMP_TP_xxx (enum pool_comp_type).
  */
 enum {
-	DAOS_PROP_PERF_DOMAIN_ROOT = 255,
-	DAOS_PROP_PERF_DOMAIN_GROUP = 3,
+	DAOS_PROP_PERF_DOMAIN_ROOT  = 255,
+	DAOS_PROP_PERF_DOMAIN_GROUP = 200,
```
mjmac, line 194: Should these use the enum names from pool_map.h instead of raw values?

src/include/daos_prop.h
```diff
@@ -479,15 +479,17 @@ enum {
...
 enum {
-	DAOS_PROP_CO_REDUN_MIN	= 1,
+	DAOS_PROP_CO_REDUN_MIN = 1,
 	/* server rank (engine) level */
-	DAOS_PROP_CO_REDUN_RANK	= 1,
+	DAOS_PROP_CO_REDUN_RANK = 1,
 	/* server node level */
-	DAOS_PROP_CO_REDUN_NODE	= 2,
-	DAOS_PROP_CO_REDUN_MAX	= 254,
+	DAOS_PROP_CO_REDUN_NODE = 2,
```
mjmac, line 486: Ditto

mjmac (commented):
Took a quick pass through -- looks fine to me, but I don't fully grok the context.

src/include/daos/pool_map.h
kjacque: No, the minimum user-defined level is now FAULT in this PR.

src/include/daos_prop.h
kjacque: Yeah, I felt that way too. I think they are defined this way to avoid circular dependencies. daos/pool_map.h includes daos/common.h, which includes daos_prop.h. Not sure if it's an over-broad include for convenience or if there is really something in daos_prop.h that is needed by daos/pool_map.h.

I usually take stuff like this as a sign that a header file needs to be broken up (daos/common.h maybe shouldn't exist). Feels out of scope of this PR though.

tanabarr (commented):
LGTM from a cursory perspective. Some details related to fault-domain specifics aren't within my grasp yet.

src/common/fault_domain.c
```diff
@@ -50,6 +66,58 @@ d_fd_tree_init(struct d_fd_tree *tree, const uint32_t *compressed,
 	return d_fd_tree_reset(tree);
 }
 
+static bool
+tree_has_fault_domain(struct d_fd_tree *tree)
+{
+	return (tree->fdt_compressed[0] & D_FD_TREE_HAS_FAULT_DOMAIN) != 0;
```
liuxuezhao, line 72: just confirm that the D_FD_TREE_HAS_FAULT_DOMAIN bit only set when the DAOS_PROP_PO_REDUN_LVL property is not on NODE and also not on RANK?
i.e. if DAOS_PROP_PO_REDUN_LVL on NODE or RANK, should not carry D_FD_TREE_HAS_FAULT_DOMAIN bit in the first byte?

liuxuezhao:
> I've created an alternate version of this PR, #15965, to show what it looks like to simply rename the existing levels rather than add a new level in the engine.
> 
> @liuxuezhao @gnailzenh @johannlombardi Please use that PR as a comparison and let me know which approach you prefer.

So this PR #15871 , the pool map with "ROOT, [PERF], [FAULT], NODE, RANK, TGT", the [] is optional, other layers are must needed. If DAOS_PROP_PO_REDUN_LVL on NODE/RANK then no FAULT layer.
PR #15965 , the pool map with "ROOT, [PERF], FAULT, RANK, TGT". This one looks a confuse is if DAOS_PROP_PO_REDUN_LVL on RANK, will the FAULT layer still present and what is it?

To consider which one is better, may need to consider placement algorithm as we need to keep backward compatibility that the new code should be able to open/access old pool/cont right?
Current placement code will firstly select one fault domain from ROOT level to fault dom level, jump hash select one child dom, each time can only jump one level - need to traverse each domain level from ROOT to fault domain level (fault domain now container property specified, in future will change to pool property DAOS_PROP_PO_REDUN_LVL).
Once the fault domain selected, no need to traverse each layer, will jump hash to select target inside that fault domain.
One thing is that currently if the fault domain set on RANK, we still have NODE layer in the pool map, so the placement code still will traverse NODE layer that cannot bypass it.

So to keep placement backward compatibility, seems that PR#15871 is a little bit clearer?
may see the opinion from @gnailzenh and @johannlombardi also, thanks.

src/common/fault_domain.c
```diff
@@ -15,27 +17,41 @@
...
+
+/**
+ * The compressed tree must contain metadata and at least a root node.
+ */
+#define FD_TREE_MIN_LEN         (FD_TREE_TUPLE_LEN + FD_TREE_MD_LEN)
+
 /**
  * The fault domain tree level numbering begins at 0 (rank level) and increases
  * as you approach the root. The level above the ranks will always have the same
  * number.
```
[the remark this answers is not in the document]
kjacque: Perhaps just an unclear comment. I initially meant that the "node" level will always == 1, and will not change. I can clarify the comment.

src/include/daos_prop.h
```diff
@@ -479,15 +479,17 @@ enum {
...
+	/* pool fault domain level */
+	DAOS_PROP_CO_REDUN_FAULT = 3,
+	DAOS_PROP_CO_REDUN_MAX   = 254,
```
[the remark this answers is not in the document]
kjacque: Makes sense to me, agreed.

src/pool/srv_pool.c
```diff
@@ -6998,9 +6998,9 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc, bool exclud
 	if (opc == MAP_EXTEND) {
 		D_ASSERT(extend_rank_list != NULL);
 		map_version = pool_map_get_version(map) + 1;
-		rc = gen_pool_buf(map, &map_buf, map_version, extend_domains_nr,
-				  extend_rank_list->rl_nr, extend_rank_list->rl_nr * dss_tgt_nr,
-				  extend_domains, dss_tgt_nr);
+		rc          = gen_pool_buf(map, &map_buf, map_version, extend_domains_nr,
```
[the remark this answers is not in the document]
kjacque: Yes, this is because of the automatic formatting.

src/common/fault_domain.c
kjacque: Correct, if NODE is selected as the pool redundancy level, the bit will be unset. Only if a higher level is chosen during pool create will this bit be set.

src/common/fault_domain.c
```diff
@@ -61,16 +129,32 @@ get_next_domain(struct d_fd_tree *tree, struct d_fd_node *next)
 	}
 
 	fd = (struct d_fault_domain *)&tree->fdt_compressed[tree->fdt_idx];
-	next->fdn_type = D_FD_NODE_TYPE_DOMAIN;
+	if (domain_is_root(fd)) {
+		next->fdn_type = D_FD_NODE_TYPE_ROOT;
+	} else if (domain_is_node(fd)) {
+		next->fdn_type = D_FD_NODE_TYPE_NODE;
+	} else if (domain_is_fault(tree, fd)) {
```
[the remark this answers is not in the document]
kjacque: Yes, that is right. So the idea in this patch is to always have NODE, with FAULT_DOM as optional above NODE.

kjacque:
@liuxuezhao Thanks for your comments. I think maybe I got confused about requirements -- your comment sounds a little different from what I had been thinking.

Here are my assumptions based on previous email discussions:
- The pool fault level could be any level between performance domain and rank, non-inclusive (pool_perf_lvl > pool_redun_lvl > rank_lvl). Pool fault level cannot be rank.
- For containers, user can select any level in the pool map at or below pool_redun_lvl as the container redun_lvl (pool_redun_lvl >= cont_redun_lvl >= rank_lvl). This doesn't really change anything about current placement behavior.

Please let me know if I got the above assumptions wrong.

Based on these assumptions, the big question for me is whether to always include nodes in the pool map, or include a fault level only (as seen in #15965). One advantage of #15965 is no extra work is needed in engine code to maintain backward compatibility in placement. The main changes would be at the control plane and daos tool levels.

The main disadvantage of #15871, keeping NODE always present even if there is an additional FAULT level, is ensuring placement chooses between these levels correctly. But maybe this is not very hard if it's just another option for container redun_lvl?

src/include/daos_prop.h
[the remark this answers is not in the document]
liuxuezhao: ok, can change it in a following PR, thx

liuxuezhao:
> Here are my assumptions based on previous email discussions:
> 
> * The pool fault level could be any level between performance domain and rank, non-inclusive (pool_perf_lvl > pool_redun_lvl > rank_lvl). Pool fault level cannot be rank.
> * For containers, user can select any level in the pool map at or below pool_redun_lvl as the container redun_lvl (pool_redun_lvl >= cont_redun_lvl >= rank_lvl). This doesn't really change anything about current placement behavior.
> 
> Please let me know if I got the above assumptions wrong.
> 

I think fault domain (redundant level) on RANK is what we need to support, this is a normal usage/config actually.

We ever discussed before that for container create, does not allow to assign the DAOS_PROP_CO_REDUN_LVL anymore, all containers just use pool's redundant level (need to add pool property DAOS_PROP_PO_REDUN_LVL).
To keep backward compatible, still keep the DAOS_PROP_CO_REDUN_LVL container property internally, for newly created container just inherit from pool property DAOS_PROP_PO_REDUN_LVL), for old existed container still keep the previous set container property (only possible to be NODE or RANK). For obj layout calculation still read the container property same as now.

> Based on these assumptions, the big question for me is whether to always include nodes in the pool map, or include a fault level only (as seen in #15965). One advantage of #15965 is no extra work is needed in engine code to maintain backward compatibility in placement. The main changes would be at the control plane and daos tool levels.
> 

Not quite sure which work needed to maintain backward compatibility in placement in your mind?

> The main disadvantage of #15871, keeping NODE always present even if there is an additional FAULT level, is ensuring placement chooses between these levels correctly. But maybe this is not very hard if it's just another option for container redun_lvl?

I thought about another point of the necessary to always include NODE in pool map. For example if later we define the FAULT level at RACK (higher than NODE), if with NODE layer in pool map it is easier to control the data balance across different NODE for placement algorithm.

let's keep discuss to clarify all not clear points, thanks.

kjacque:
> I think fault domain (redundant level) on RANK is what we need to support, this is a normal usage/config actually.

OK. Thanks for the clarification. I am still deciding on how to present this choice to the user in the pool create process.

> We ever discussed before that for container create, does not allow to assign the DAOS_PROP_CO_REDUN_LVL anymore, all containers just use pool's redundant level (need to add pool property DAOS_PROP_PO_REDUN_LVL). To keep backward compatible, still keep the DAOS_PROP_CO_REDUN_LVL container property internally, for newly created container just inherit from pool property DAOS_PROP_PO_REDUN_LVL), for old existed container still keep the previous set container property (only possible to be NODE or RANK). For obj layout calculation still read the container property same as now.

Ah, I recall this idea, but it was suggested (maybe by @gnailzenh?) to let users continue to customize redun_lvl at the container level as well.

> Not quite sure which work needed to maintain backward compatibility in placement in your mind?

Sorry, "backward compatibility" was the wrong term, this PR is backward compatible since it still only allows for NODE level to be selected. I meant to say the logic will need to be updated to allow for PO_CMP_TP_FAULT as well.

> I thought about another point of the necessary to always include NODE in pool map. For example if later we define the FAULT level at RACK (higher than NODE), if with NODE layer in pool map it is easier to control the data balance across different NODE for placement algorithm.

My instinct was that the user does not get to change the fault level of an existing pool, only to set it when creating the pool. Maybe it would be OK to change if no container has been created yet. Otherwise, if there are containers with data, wouldn't all that data need to be rebalanced if we consider the container to inherit redundancy level from the pool?

If the container maintains its own redundancy level as well, maybe it is a non-issue. I think we still would want to avoid changing the pool map though (i.e. changing what actual level FAULT applies to).

liuxuezhao:
@kjacque we discussed it at Chinese engineer weekly meeting. The thoughts are -

1. add pool property DAOS_PROP_PO_REDUN_LVL, user can set it when creating pool. Disallow to change it after pool creation. It can be set on FAULT/NODE/RANK (may consider to allow on TGT later).
2. all newly created container will inherit pool's redun_lvl when create cont, disallow to set the DAOS_PROP_CO_REDUN_LVL property on container level anymore. This is to simplify the RF model. To keep backward compatibility with old existing container, can internally keep the DAOS_PROP_CO_REDUN_LVL cont property (but cannot set it), so old existed container can still use its previously set property.
3. pool map with "ROOT, [PERF], [FAULT], NODE, RANK, TGT", the [] is optional, other layers are must needed. If DAOS_PROP_PO_REDUN_LVL on NODE/RANK then no FAULT layer. Main reasons are - 1st current placement algorithm we have NODE/RANK that physically corresponding to machine/engine, the NODE affect placement's domain selection when fault domain on RANK, so probably keep NODE/RANK in pool map is more clear; 2nd with NODE we can balance data on NODE level (the 2nd is not so important as if without NODE we still with RANK so can balance on RANK also).

@gnailzenh please correct me if the above note is not right.
@kjacque we can keep discussion, may schedule a meeting with liang if needed.

If we are on same page, I can help on some placement algorithm change (just a few code change in my mind), all other things like the property/control plane change may on your side. thanks.

liuxuezhao (approved):
LGTM, can change the DAOS_PROP_CO_REDUN_MAX to 3 in a following PR

src/common/fault_domain.c
```diff
@@ -15,27 +17,41 @@
...
+/**
+ * A compressed domain tree starts with values that encode metadata about the tree.
+ */
+#define FD_TREE_MD_LEN          1
+
+/**
+ * The compressed tree must contain metadata and at least a root node.
+ */
+#define FD_TREE_MIN_LEN         (FD_TREE_TUPLE_LEN + FD_TREE_MD_LEN)
+
```
liw, line 29: Just a question: To cope with this change of the compressed form, I need to make a test update along the following line?

```
diff --git a/src/pool/srv_util.c b/src/pool/srv_util.c
index e2118e8ef..40ba9836c 100644
--- a/src/pool/srv_util.c
+++ b/src/pool/srv_util.c
@@ -850,10 +850,16 @@ testu_create_domain_buf(d_rank_t *ranks, int n_ranks, uint32_t **domain_buf_out,
 
        D_ASSERT(n_domains * n_ranks_per_domain == n_ranks);
 
-       domain_buf_len = 3 /* root */ + 3 * n_domains + n_ranks;
+       domain_buf_len = 1 /* metadata */ + 3 /* root */ + 3 * n_domains + n_ranks;
        D_ALLOC_ARRAY(domain_buf, domain_buf_len);
        D_ASSERT(domain_buf != NULL);
 
+       /* The metadata. */
+       p = domain_buf;
+       D_ASSERT(domain_buf <= p && p + 1 <= domain_buf + domain_buf_len);
+       p[0] = 0;
+       p++;
+
        /* The root. */
        p = domain_buf;
        D_ASSERT(domain_buf <= p && p + 3 <= domain_buf + domain_buf_len);
```

liw (approved):

src/common/fault_domain.c
kjacque: Yes, these changes look correct for this test.

daltonbohning:
Ideally should get a review from @daos-stack/client-api-owners first, or I'll let another gatekeeper merge

mchaarawi:
> Ideally should get a review from @daos-stack/client-api-owners first, or I'll let another gatekeeper merge

the api changes are minimal (just adding some props, so lgtm).

PR #15888, opened 2025-02-12
DAOS-16355 pydaos: add dir object cache for Datasets

0xE0F:
Dataset reads a lot of samples that are stored under just a few common directories, to reduce computation load on lookup for each sample file, this commit introduces a cache of directory objects.

Features: PytorchDatasetsTest,PytorchCheckpointTest

### Gatekeeper:

### DIFF
--- a/src/client/pydaos/torch/torch_shim.c
+++ b/src/client/pydaos/torch/torch_shim.c
@@ -21,17 +21,74 @@
 #include <daos/common.h>
 #include <gurt/debug.h>
 #include <gurt/common.h>
+#include <gurt/hash.h>
 
 #define PY_SHIM_MAGIC_NUMBER (0x7A8B)
 #define EQ_POLL_BATCH_SIZE   (64)
 
 struct dfs_handle {
-	int           flags;
-	dfs_t        *dfs;
-	d_iov_t       global;
+	int                  flags;
+	dfs_t               *dfs;
+	d_iov_t              global;
 
-	daos_handle_t eq;
-	pid_t         eq_owner_pid;
+	daos_handle_t        eq;
+	pid_t                eq_owner_pid;
+	struct d_hash_table *dir_cache;
+};
+
+/* Cached directory object entry */
+struct dir_obj_cache_entry {
+	d_list_t   entry;
+	dfs_obj_t *obj;
+	char       name[PATH_MAX];
+};
+
+static inline struct dir_obj_cache_entry *
+dir_obj_cache_entry_from_link(d_list_t *rlink)
+{
+	return container_of(rlink, struct dir_obj_cache_entry, entry);
+}
+
+static bool
+dir_cache_key_cmp(struct d_hash_table *htable, d_list_t *rlink, const void *key, unsigned int ksize)
+{
+	struct dir_obj_cache_entry *h = dir_obj_cache_entry_from_link(rlink);
+
+	return (strcmp(h->name, (const char *)key) == 0);
+}
+
+static void
+dir_cache_rec_free(struct d_hash_table *htable, d_list_t *rlink)
+{
+	struct dir_obj_cache_entry *h = dir_obj_cache_entry_from_link(rlink);
+
+	int                         rc = dfs_release(h->obj);
+	if (rc) {
+		D_ERROR("Could not release object '%s': %s (rc=%d)", h->name, strerror(rc), rc);
+	}
+	free(h);
+}
+
+static bool
+dir_cache_rec_decref(struct d_hash_table *htable, d_list_t *rlink)
+{
+	return true;
+}
+
+static uint32_t
+dir_cache_rec_hash(struct d_hash_table *htable, d_list_t *rlink)
+{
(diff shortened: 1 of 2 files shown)


johannlombardi (dismissed):

jolivier23 (dismissed):

src/client/pydaos/torch/torch_shim.c
```diff
@@ -95,6 +152,50 @@ __shim_handle__module_fini(PyObject *self, PyObject *args)
 	return PyLong_FromLong(rc);
 }
 
+/* Hash table does not require lock: workers will be working on their own subsets of data
+and will have their own handlers, including the cache.
+*/
+static int
+__dir_cache_create(struct dfs_handle *hdl)
+{
+	int rc = d_hash_table_create(D_HASH_FT_EPHEMERAL | D_HASH_FT_NOLOCK | D_HASH_FT_LRU, 4,
```
mchaarawi, line 161: 4 means 2^4 (16) entries allowed in the hashtable. i thought for pytorch workloads, the number of dirs is usually much greater than that. do you want to reconsider this?
If you have done perf measurement where the number of directories in your dataset is greater than that, then i doubt you have seen any improvement in perf.

mchaarawi (changes requested):
no other changes requested other than verifying the size of the hashtable is sufficient.

src/client/pydaos/torch/torch_shim.c
enakta: I tested with two labels dataset: `/samples/train/{0,1}/<samples>`, which is default (and unet3d) layout for dlio_benchmark.
I don't think there would be one size that fits them all and the parameter in the constructor would be more flexible approach, indeed.

enakta: @mchaarawi added ability to passthrough cache size from the client code

mchaarawi: sorry i should have been more clear.
for the change that you just made, 2^256 for the dir cache size is too extreme and likely consumes too much memory. i would suggest to use something like 16, which should be ok for the default, or even for always.

in the example you posted, are all the datasets you tested with 2 directories under train? and all the files under those 2 directories? those should have fit in the cache and you should decrease the number of lookups. I have seen datasets that have thousands of directories (imagenet after being transformed, etc).  so just wanted to confirm.

enakta: > in the example you posted, are all the datasets you tested with 2 directories under train? and all the files under those 2 directories? those should have fit in the cache and you should decrease the number of lookups. 

Yes, it did. 

@mchaarawi it's not 2^256 tho,  there's conversion  with `uint32_t bits = ceil(log2(hdl->dir_cache_size))`.
I just didn't want to expose it to client code as 2^N to avoid confusion 😬

mchaarawi (approved):

src/client/pydaos/torch/torch_shim.c
```diff
@@ -262,32 +383,78 @@ __shim_handle__torch_worker_init(PyObject *self, PyObject *args)
...
+	}
+
+	rec->obj = *obj;
+	strncpy(rec->name, name, len);
+
+	rc = d_hash_rec_insert(hdl->dir_cache, rec->name, len, &rec->entry, false);
+	if (rc) {
+		D_ERROR("Failed to insert dir handle in hashtable: '%s': %s (rc=%d)", name,
+			d_errstr(rc), rc);
+		rc = daos_der2errno(rc);
```
daltonbohning, line 426: Should we instead print this as a system errno instead of d_err? I.e. convert der2errno and then print

daltonbohning (changes requested):
Code LGTM but this
```
Features: PytorchDatasetsTest,PytorchCheckpointTest
```
Means "run tests tagged with both PytorchDatasetsTest and PytorchCheckpointTest"
What you want is
```
Features: PytorchDatasetsTest PytorchCheckpointTest
```
Or, more simply:
```
Features: pytorch
```

src/client/pydaos/torch/torch_shim.c
enakta: `d_errstr` might give more specific details in case of the error, rather than more generic `errno`

enakta: @daltonbohning thanks, I've restarted checks with `Features: pytorch`

daltonbohning: I mean since all other printed errors at this layer are errno, it seems inconsistent to print d_err here

daltonbohning (approved):

src/client/pydaos/torch/torch_shim.c
enakta: ah I see, but there are heaps of other `d_errstr` in this layer: pretty much everything that does not start with `dfs_` prefix is printed with `d_errstr` and then converted to `errno` with `daos_der2errno`. Just to have a bit more detail when an error pops up - saved me a fair few grey hairs.

PR #15901, opened 2025-02-13
DAOS-16876 vos: discard invalid DTX when commit or abort - b26

Nasf-Fan:
When commit or abort a DTX, we will check whether it is a valid entry or not. For invalid case, we will discard it with warning message and increase related metrics counter.

It may be not perfect solution, but it is efficient to help the user to cleanup system efficiently.

### Gatekeeper:

### DIFF
--- a/src/vos/ilog.c
+++ b/src/vos/ilog.c
@@ -1,5 +1,7 @@
 /**
  * (C) Copyright 2019-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
+ * (C) Copyright 2025 Google LLC
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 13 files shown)


src/vos/ilog.c
```diff
@@ -984,8 +989,12 @@ ilog_modify(daos_handle_t loh, const struct ilog_id *id_in,
 		"%s in incarnation log " DF_X64 " status: rc=" DF_RC " tree_version: %d\n",
 		opc_str[opc], id_in->id_epoch, DP_RC(rc), ilog_mag2ver(lctx->ic_root->lr_magic));
 
-	if (rc == 0 && version != ilog_mag2ver(lctx->ic_root->lr_magic) &&
-	    (opc == ILOG_OP_PERSIST || opc == ILOG_OP_ABORT)) {
+	if (rc == 0 && opc != ILOG_OP_UPDATE) {
+		if (version == ilog_mag2ver(lctx->ic_root->lr_magic)) {
+			D_WARN("ilog entry on %s doesn't exist\n", opc_str[opc]);
+			return -DER_NONEXIST;
```
NiuYawei, line 995: I don't quite follow this.

src/vos/vos_dtx.c
```diff
@@ -573,7 +574,7 @@ dtx_ilog_rec_release(struct umem_instance *umm, struct vos_container *cont,
 
 	ilog_close(loh);
 
-	if (rc != 0)
+	if (rc != 0 && rc != -DER_NONEXIST)
```
NiuYawei, line 577: Isn't it too late to check DER_NONEXIST here? I suppose error is already returned when above ilog_open() failed, right?

src/vos/ilog.c
Nasf-Fan: As my understand, when we commit or abort the ilog via dtx_{commit,abort}, if commit/abort succeed, then the version will bump. Here, if the version does not bump, then must be not found. @jolivier23 , is it your expected?

src/vos/vos_dtx.c
Nasf-Fan: Such log message is no matter, its caller `do_dtx_rec_release()` will print.

src/vos/ilog.c
jolivier23: That is correct.   It means the entry is no longer in the ilog.  This code is shared between abort, persist, and update (or insert) but only the former two need to remove the ilog entry from the dtx record

daltonbohning (commented):
ftest LGTM

Nasf-Fan:
Please help to review the patch that is required for 2.6.3-rc4, thanks!

src/vos/ilog.c
NiuYawei: Ok, if we consider it as an error, I think we'd use D_ERROR instead of D_WARN.

src/vos/vos_dtx.c
```diff
@@ -648,6 +661,15 @@ do_dtx_rec_release(struct umem_instance *umm, struct vos_container *cont,
 		break;
 	}
 
+	if (unlikely(rc == -DER_NONEXIST)) {
+		struct vos_tls	*tls = vos_tls_get(false);
```
NiuYawei, line 665: Better to use vos_pool->vp_sysdb as parameter.

src/vos/vos_dtx.c
NiuYawei: It's better to add an error message in ilog_open() or after the ilog_open() call above.

NiuYawei (approved):

src/vos/vos_dtx.c
Nasf-Fan: Sure, it will be changed in subsequent patch.

gnailzenh (approved):

PR #15952, opened 2025-02-21
DAOS-17095 client: disable flock as default in dfuse

wiliamhuang:
dfuse needs argument "--enable-local-flock" to enable the support of node-wise flock. Otherwise, flock is not supported as default.

### Gatekeeper:


mchaarawi (dismissed):
lgtm

src/tests/ftest/dfuse/simul.py
```diff
@@ -103,22 +106,26 @@ def run_simul(self, include=None, exclude=None, raise_exception=True):
...
-        return run_command(cmd, output_check="combined", raise_exception=raise_exception)
+            for test_to_fail in faillist:
+                mpirun = Mpirun(f"{simul_cmd} -vv -d {dfuse.mount_dir.value} -i {test_to_fail}",
+                                mpi_type=mpi_type)
+                mpirun.assign_hosts(get_local_host(), dfuse.mount_dir.value)
+                mpirun.assign_processes(processes=num_proc, ppn=ppn)
+                out = mpirun.run(raise_exception=raise_exception)
+                # testing cases that are expected to fail
+                if "FAILED in simul" in out.stdout_text:
+                    self.log.info(f"Test {test_to_fail} failed as expected")
```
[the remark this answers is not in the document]
wiliamhuang: Thank you! I will fix it in next commit. I will wait for more feedbacks from other reviewers.

jolivier23 (approved):
Same comment on pylint as Mohamad

wiliamhuang:
@mchaarawi @jolivier23 Just fixed the pylint issue. Thank you!

src/tests/ftest/util/dfuse_utils.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, namespace, command, path=""):
         self.disable_wb_cache = FormattedParameter("--disable-wb-cache", False)
         self.multi_user = FormattedParameter("--multi-user", False)
         self.read_only = FormattedParameter("--read-only", False)
+        self.local_flock = FormattedParameter("--enable-local-flock", False)
```
daltonbohning, line 41: Nit - I'd recommend making the variable match the command line more closely so there is less guesswork
```suggestion
        self.enable_local_flock = FormattedParameter("--enable-local-flock", False)
```

src/tests/ftest/dfuse/simul.yaml
```diff
@@ -16,10 +16,11 @@ pool:
...
-    simul_path: /usr/lib64/openmpi/bin
-  mpich:
-    mpi_type: mpich
-    simul_path: /usr/lib64/mpich/bin
+mpi:
+  mpi_type: mpich
+  simul_path: /usr/lib64/mpich/bin
+client_processes:
+  np: 24
+  ppn: 12
```
daltonbohning, line 24: Better to just specify `ppn` and let the framework set `np` based on the number of clients. This lets the test scale
```suggestion
  ppn: 12
```

src/tests/ftest/dfuse/simul.py
```diff
@@ -103,22 +106,26 @@ def run_simul(self, include=None, exclude=None, raise_exception=True):
...
-        if include and not exclude:
-            cmd = "{0} -vv -d {1} -i {2}".format(simul_cmd, dfuse.mount_dir.value, include)
-        elif exclude and not include:
-            cmd = "{0} -vv -d {1} -e {2}".format(simul_cmd, dfuse.mount_dir.value, exclude)
+        self.log_step(f"Running simul on {mpi_type}")
+        if faillist is None:
+            mpirun = Mpirun(f"{simul_cmd} -vv -d {dfuse.mount_dir.value} -e {exclude}",
+                            mpi_type=mpi_type)
+            mpirun.assign_hosts(get_local_host(), dfuse.mount_dir.value)
+            mpirun.assign_processes(processes=num_proc, ppn=ppn)
```
daltonbohning, line 115: Here if you just give `ppn` the function will calculate `np` based on the number of hosts already assigned above
```suggestion
            mpirun.assign_processes(ppn=ppn)
```

src/tests/ftest/datamover/serial_large_posix.yaml
```diff
@@ -49,3 +49,4 @@ ddeserialize:
     np: 16
 dfuse:
   disable_caching: true
+  local_flock: true
```
daltonbohning, line 52: I'm curious about how tests were chosen to use `local_flock: true`? Was it just arbitrary to get some coverage?
FWIW this test serializes a container into dfuse in an HDF5 format.

daltonbohning (approved):
Nits

src/tests/ftest/datamover/serial_large_posix.yaml
wiliamhuang: @daltonbohning Actually, I ran all CI tests without "local_flock: true" at the beginning. Three tests failed. I discussed with Mohamad whether we could set "local_flock: true" for them.

src/tests/ftest/dfuse/simul.py
wiliamhuang: Thank you! Fixed as suggested.

src/tests/ftest/datamover/serial_large_posix.yaml
daltonbohning: So we set `local_flock: true` in certain tests because they fail without it? In a production environment, how would a user know to set this, since it's not the default?

mchaarawi: we do not support flock. enabling it by default through fuse gives the false sense that it is supported. so if one does flock across multiple nodes, both nodes can acquire the lock at the same time and get granted which can cause data corruption.
The idea is to disable this behavior since this is not what we want. so in production users would get errors when their apps use flock, as it should be.
This flag is to allow the use of flock in certain situation where the users know what they are doing, mainly running on a single node with a single dfuse instance. In that case, using flock is safe.

daltonbohning: Thanks for the explanation

mjmac: Would it be reasonable to only allow flock to be enabled if the container has been opened with the `DAOS_COO_EX` flag set?

daltonbohning (approved):

PR #15956, opened 2025-02-21
DAOS-17006 cart: Bypass telemetry init when not enabled

mjmac:
Fixes an oversight in the client where HG metrics were
causing warnings on startup when client telemetry is not
enabled.
  
Change-Id: I0144f432106e641ff7fb8f2ea0ab9c73c564c25d

### DIFF
--- a/src/client/api/event.c
+++ b/src/client/api/event.c
@@ -1,6 +1,5 @@
 /**
  * (C) Copyright 2016-2024 Intel Corporation.
- * (C) Copyright 2025 Google LLC
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -17,7 +16,6 @@
 #include "client_internal.h"
 #include <daos/mgmt.h>
 #include <daos/rpc.h>
-#include <daos/metrics.h>
 
 /** thread-private event */
 static __thread daos_event_t	ev_thpriv;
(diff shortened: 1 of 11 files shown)


src/client/api/event.c
```diff
@@ -1,6 +1,5 @@
 /**
  * (C) Copyright 2016-2024 Intel Corporation.
- * (C) Copyright 2025 Google LLC
```
jolivier23, line 3: this is an odd removal

jolivier23 (approved):
why is copyright removed?  Because it's just a revert?

src/client/api/init.c
```diff
@@ -203,19 +203,12 @@ daos_init(void)
...
-		DL_WARN(rc, "failed to initialize client telemetry");
-	}
-
 	/** get and cache attach info of default system */
 	rc = dc_mgmt_cache_attach_info(NULL);
 	if (rc != 0)
 		D_GOTO(out_job, rc);
 
-	crt_info = daos_crt_init_opt_get(false, 1, daos_client_metric);
+	crt_info = daos_crt_init_opt_get(false, 1);
```
jolivier23, line 211: minor nit, if we always pass in 1, should we should remove the extra arg?

src/client/api/event.c
mjmac: It's because of the revert commit.

src/client/api/init.c
mjmac: I think @frostedcmos made this change a little while ago... I'll let him comment on that.

PR #15963, opened 2025-02-24
DAOS-17006 cart: Publish Mercury counters as metrics (#15870)

mjmac:
When Mercury has been built with diagnostic RPC counters
enabled, CaRT will periodically republish the counters
as DAOS telemetry for consumption by monitoring
infrastructure. NB: Requires Mercury > 2.4.0.

Change-Id: I0232d0da8007374fd1d28d395c65544c7fa57bc1
Co-authored-by: Jeff Olivier <jeffolivier@google.com>
Co-authored-by: Nicholas Murphy <ncmurphy@google.com>


mjmac:
@soumagne: FYI, this is a PR suitable for master... I have chosen not to include the mercury patch, but I can adjust that if desired. Do you anticipate making a 2.4.1 release prior to DAOS 2.8.0? If so, then the new metrics will "just work" when DAOS is built against it.

soumagne:
ok that works, thanks! Yes I will be making a new release for DAOS 2.8.

daltonbohning (commented):
ftest LGTM but does this need `Features: telemetry`?

mjmac:
> ftest LGTM but does this need `Features: telemetry`?

The patch should basically be a no-op until we start building against the newer version of Mercury. Can you see if any of the telemetry tests were run as part of the usual PR set?

daltonbohning:
> > ftest LGTM but does this need `Features: telemetry`?
> 
> The patch should basically be a no-op until we start building against the newer version of Mercury. Can you see if any of the telemetry tests were run as part of the usual PR set?

These three are pr,telemetry. Maybe sufficient?
- control/dmg_telemetry_basic.py
- control/dmg_telemetry_io_basic.py:
- control/dmg_telemetry_nvme.py

mjmac:
> These three are pr,telemetry. Maybe sufficient?
> 
> * control/dmg_telemetry_basic.py

Yeah, that one runs this sub-test: https://github.com/daos-stack/daos/blob/master/src/tests/ftest/control/dmg_telemetry_basic.py#L73 ... Pretty sure that's the one that would fail if I forgot to update the metrics list.

jolivier23:
Need to fix the merge conflict

mjmac:
> Need to fix the merge conflict

Yeah, just waiting to see if I get any other review feedback before kicking off another CI run.

src/tests/ftest/util/telemetry_utils.py
```diff
@@ -851,6 +860,16 @@ class ClientTelemetryUtils(TelemetryUtils):
...
+    CLIENT_NET_METRICS = [
+        "client_net_hg_bulks",
+        "client_net_hg_req_recv",
+        "client_net_hg_extra_bulk_resp",
+        "client_net_hg_extra_bulk_req",
+        "client_net_hg_resp_sent",
+        "client_net_hg_resp_recv",
+        "client_net_hg_mr_copies",
+        "client_net_hg_req_sent",
+        "client_net_hg_active_rpcs"]
```
mchaarawi, line 872: isn't there supposed to be a test that collects the metrics and verifies the values against a certain workload?

mjmac: IMO that's kind of a low-value test in that it's really testing Mercury instead of DAOS code. Trying to determine "correct" counter values at this level also seems pretty error-prone and likely to introduce test flakiness.

mchaarawi: sure, i agree this is probably an internal test for those counters. IDK if such a test exists. @soumagne do you know? if those are being exposed, it would be a good idea to verify they are correct.

what i don't get is what flakiness have to do with counters. If it's a deterministic workload, the results should have no variance and should not be error prone. so not sure i get what you mean there.

soumagne: we currently do not have a test for these.

kjacque (approved):

src/tests/ftest/util/telemetry_utils.py
mjmac: > what i don't get is what flakiness have to do with counters. If it's a deterministic workload, the results should have no variance and should not be error prone. so not sure i get what you mean there.

Say we add a test today that writes 1MB of data to a container using IOR. We run the test, and then read the counters for the IOR client process. At the end of it we should see 0 active RPCs, 3 bulks, 13 req sent and 13 resp received (I'm just making numbers up for illustration).

The values of these counters are determined by a number of factors, including: object class, protocol query, collective RPCs enabled/disabled, etc. Now imagine that we change some of the client startup logic to somehow reduce or eliminate the protocol query stuff, or some other change is made to improve collective RPC efficiency, etc. And now the RPC count is off and the test will fail whenever it's run (probably not on the PR that made the change because that one wasn't directly modifying telemetry).

There are a lot of variables that will affect the final counts, and my concern is that this hypothetical test would wind up in the pile of perpetually-broken weekly tests that just get ignored and don't actually improve product quality.

We already have unit tests for the telemetry library itself. IMO, a better approach would be to implement some kind of unit test in Mercury itself to verify that counters are incremented as expected for a deterministic synthetic workload.

src/cart/crt_context.c
```diff
@@ -285,50 +286,47 @@ crt_context_provider_create(crt_context_t *crt_ctx, crt_provider_t provider, boo
...
 
 		ret = d_tm_add_metric(&ctx->cc_timedout_uri, D_TM_COUNTER,
 				      "Total number of timed out URI lookup "
 				      "requests", "reqs",
 				      "net/%s/uri_lookup_timeout/ctx_%u",
 				      prov, ctx->cc_idx);
 		if (ret)
-			D_WARN("Failed to create timed out uri req counter: "
-			       DF_RC"\n", DP_RC(ret));
+			DL_WARN(ret, "Failed to create timed out uri req counter:");
```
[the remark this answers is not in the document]
mjmac: Fixed, thanks.

src/tests/ftest/util/telemetry_utils.py
jolivier23: That's a fair concern.  Do we have sanity tests for for any of our metrics outside of the basic tests that they are produced?

jolivier23 (approved):

kjacque (approved):

PR #15975, opened 2025-02-25
DAOS-16936 dtx: rank range based DTX hints for coll_punch - b26

Nasf-Fan:
When handle collective punch RPC, the leader will generate hints array for related DTX RPC (abort or commit). Each involved engine will have one element in such array. To simplify RPC logic, such hints array is sparse and indexed with engine's rank#. Originally, we did not properly handled the case of incontinuous rank# in the pool map, as to when update some hints element with large rank#, the write maybe out of boundary and crash the others' space and cause kinds of DRAM corruption.

Similar situation can happen during handle collective DTX resync and cleanup.

This patch fixes such issue via building such sparse hints array based on related engines' rank# range instead of the ranks count in the pool map. Use relative ranks diff (real rank# - base rank#) as the index. That will avoid out of boundary access.

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate owners.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
  * [ ] Quick-build or Quick-functional is not used.
[...]

PR #15979, opened 2025-02-26
DAOS-16936 dtx: rank range based DTX hints for coll_punch

Nasf-Fan:
When handle collective punch RPC, the leader will generate hints array for related DTX RPC (abort or commit). Each involved engine will have one element in such array. To simplify RPC logic, such hints array is sparse and indexed with engine's rank#. Originally, we did not properly handled the case of incontinuous rank# in the pool map, as to when update some hints element with large rank#, the write maybe out of boundary and crash the others' space and cause kinds of DRAM corruption.

Similar situation can happen during handle collective DTX resync and cleanup.

This patch fixes such issue via building such sparse hints array based on related engines' rank# range instead of the ranks count in the pool map. Use relative ranks diff (real rank# - base rank#) as the index. That will avoid out of boundary access.

### Steps for the author:

### DIFF
--- a/src/dtx/dtx_coll.c
+++ b/src/dtx/dtx_coll.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2023-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -106,6 +107,43 @@ out:
 	D_ASSERT(rc == ABT_SUCCESS);
 }
 
+static void
+dtx_coll_find_rank_range(struct pl_map *map, struct pl_obj_layout *layout, uint32_t version,
+			 uint32_t *min_rank, uint32_t *max_rank)
+{
+	struct pool_target *target;
+	d_rank_t            my_rank = dss_self_rank();
+	int                 rc;
+	int                 i;
+
+	for (i = 0, *min_rank = -1, *max_rank = 0; i < layout->ol_nr; i++) {
+		if (layout->ol_shards[i].po_target == -1 || layout->ol_shards[i].po_shard == -1)
+			continue;
+
+		rc = pool_map_find_target(map->pl_poolmap, layout->ol_shards[i].po_target, &target);
+		D_ASSERT(rc == 1);
+
+		/* Skip current leader rank. */
+		if (target->ta_comp.co_rank == my_rank)
+			continue;
+
+		/* Skip the target that (re-)joined the system after the DTX. */
+		if (target->ta_comp.co_ver > version)
+			continue;
+
+		/* Skip non-healthy one. */
+		if (target->ta_comp.co_status != PO_COMP_ST_UP &&
+		    target->ta_comp.co_status != PO_COMP_ST_UPIN &&
+		    target->ta_comp.co_status != PO_COMP_ST_DRAIN)
+			continue;
+
+		if (*min_rank > target->ta_comp.co_rank)
+			*min_rank = target->ta_comp.co_rank;
+		if (*max_rank < target->ta_comp.co_rank)
+			*max_rank = target->ta_comp.co_rank;
+	}
+}
+
 int
 dtx_coll_prep(uuid_t po_uuid, daos_unit_oid_t oid, struct dtx_id *xid, struct dtx_memberships *mbs,
 	      uint32_t my_tgtid, uint32_t dtx_ver, uint32_t pm_ver, bool for_check, bool need_hint,
@@ -118,6 +156,7 @@ dtx_coll_prep(uuid_t po_uuid, daos_unit_oid_t oid, struct dtx_id *xid, struct dt
 	struct dtx_coll_target	*dct;
 	struct dtx_coll_entry	*dce = NULL;
 	struct daos_obj_md	 md = { 0 };
+	uint32_t                *ranks;
 	uint32_t		 rank_nr;
 	d_rank_t		 my_rank = dss_self_rank();
 	d_rank_t		 max_rank = 0;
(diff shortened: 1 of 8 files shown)


mchaarawi (commented):
i think the PR should re-enable collective punch as default if the fix was for it.
otherwise no test will be running with collective punch in CI anyway.

Nasf-Fan:
> i think the PR should re-enable collective punch as default if the fix was for it. otherwise no test will be running with collective punch in CI anyway.

Yes, collective punch has already been enabled on master, but disabled on release/2.6 by default. This patch is for master, so no need to change that.

Nasf-Fan:
https://github.com/daos-stack/daos/actions/runs/13547004393/job/37861024957?pr=15979
diff --git a/src/dtx/dtx_internal.h b/src/dtx/dtx_internal.h
index 6a7442c69..0250db7b9 100644
--- a/src/dtx/dtx_internal.h
+++ b/src/dtx/dtx_internal.h
@@ -77,16 +77,11 @@ CRT_RPC_DECLARE(dtx, DAOS_ISEQ_DTX, DAOS_OSEQ_DTX);
  * dci_hints is sparse array, one per engine, sorted against the rank ID.
  * It can hold more than 19K engines inline RPC body.
  */
-#define DAOS_ISEQ_COLL_DTX						\
-	((uuid_t)		(dci_po_uuid)		CRT_VAR)	\
-	((uuid_t)		(dci_co_uuid)		CRT_VAR)	\
-	((struct dtx_id)	(dci_xid)		CRT_VAR)	\
-	((uint32_t)		(dci_version)		CRT_VAR)	\
-	((uint32_t)		(dci_min_rank)		CRT_VAR)	\
-	((uint32_t)		(dci_max_rank)		CRT_VAR)	\
-	((uint32_t)		(dci_padding)		CRT_VAR)	\
-	((uint64_t)		(dci_epoch)		CRT_VAR)	\
-	((uint8_t)		(dci_hints)		CRT_ARRAY)
+#define DAOS_ISEQ_COLL_DTX                                                                         \
+	((uuid_t)(dci_po_uuid)CRT_VAR)((uuid_t)(dci_co_uuid)CRT_VAR)((struct dtx_id)(              \
+	    dci_xid)CRT_VAR)((uint32_t)(dci_version)CRT_VAR)((uint32_t)(dci_min_rank)CRT_VAR)(     \
+	    (uint32_t)(dci_max_rank)CRT_VAR)((uint32_t)(dci_padding)CRT_VAR)(                      \
+	    (uint64_t)(dci_epoch)CRT_VAR)((uint8_t)(dci_hints)CRT_ARRAY)

@daltonbohning ,
Honestly, I do not think that the suggested clang-format is friendly for both engineers and other readers. I am wondering why we have to follow such format? or is there any rule can be adjusted? Thanks!

Nasf-Fan:
> Test stage Functional Hardware Medium completed with status FAILURE. https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-15979/3/execution/node/1510/log

Failed for DAOS-17094.

daltonbohning:
> @daltonbohning , Honestly, I do not think that the suggested clang-format is friendly for both engineers and other readers. I am wondering why we have to follow such format? or is there any rule can be adjusted? Thanks!

The hook will format for you
https://github.com/daos-stack/daos/blob/master/utils/githooks/README.md

jolivier23:
> https://github.com/daos-stack/daos/actions/runs/13547004393/job/37861024957?pr=15979 diff --git a/src/dtx/dtx_internal.h b/src/dtx/dtx_internal.h index 6a7442c69..0250db7b9 100644 --- a/src/dtx/dtx_internal.h +++ b/src/dtx/dtx_internal.h @@ -77,16 +77,11 @@ CRT_RPC_DECLARE(dtx, DAOS_ISEQ_DTX, DAOS_OSEQ_DTX);
> 
> * dci_hints is sparse array, one per engine, sorted against the rank ID.
> * It can hold more than 19K engines inline RPC body.
>   */
>   -#define DAOS_ISEQ_COLL_DTX						\
> 
> * ((uuid_t)		(dci_po_uuid)		CRT_VAR)	\
> * ((uuid_t)		(dci_co_uuid)		CRT_VAR)	\
> * ((struct dtx_id)	(dci_xid)		CRT_VAR)	\
> * ((uint32_t)		(dci_version)		CRT_VAR)	\
> * ((uint32_t)		(dci_min_rank)		CRT_VAR)	\
> * ((uint32_t)		(dci_max_rank)		CRT_VAR)	\
> * ((uint32_t)		(dci_padding)		CRT_VAR)	\
> * ((uint64_t)		(dci_epoch)		CRT_VAR)	\
> * ((uint8_t)		(dci_hints)		CRT_ARRAY)
>   +#define DAOS_ISEQ_COLL_DTX                                                                         \
> 
> * ((uuid_t)(dci_po_uuid)CRT_VAR)((uuid_t)(dci_co_uuid)CRT_VAR)((struct dtx_id)(              \
> * ```
>     dci_xid)CRT_VAR)((uint32_t)(dci_version)CRT_VAR)((uint32_t)(dci_min_rank)CRT_VAR)(     \
>   ```
> * ```
>     (uint32_t)(dci_max_rank)CRT_VAR)((uint32_t)(dci_padding)CRT_VAR)(                      \
>   ```
> * ```
>     (uint64_t)(dci_epoch)CRT_VAR)((uint8_t)(dci_hints)CRT_ARRAY)
>   ```
> 
> @daltonbohning , Honestly, I do not think that the suggested clang-format is friendly for both engineers and other readers. I am wondering why we have to follow such format? or is there any rule can be adjusted? Thanks!

For cases where it sucks, you can add comments

/* clang-format off */
Your code section where it ends up ugly
/* clang-format on */

jolivier23:
to answer the other question, the format is controlled by .clang-format file in the daos root tree

Nasf-Fan:
Thanks @daltonbohning @jolivier23

Nasf-Fan:
Ping reviewers, thanks!

wangshilong (approved):
Changes looks direct and reasonable to me

jolivier23 (approved):
Is there a test for this change?

Nasf-Fan:
> Is there a test for this change?

We have collective operation test cases, but for most of cases, the corruption is silent. What we hit on CI (and Aurora) is random delayed DRAM corruption.

src/dtx/dtx_coll.c
```diff
@@ -106,6 +107,43 @@ dtx_coll_prep_ult(void *arg)
...
+
+		rc = pool_map_find_target(map->pl_poolmap, layout->ol_shards[i].po_target, &target);
+		D_ASSERT(rc == 1);
+
+		/* Skip current leader rank. */
+		if (target->ta_comp.co_rank == my_rank)
+			continue;
+
+		/* Skip the target that (re-)joined the system after the DTX. */
+		if (target->ta_comp.co_ver > version)
```
liuxuezhao, line 131: seems the co_ver only set when it initially added to pool map? when it be excluded+reinted seems only set co_in_ver.
in DTX code some places use co_ver, some other places use co_in_ver, probably need to unify them in a following PR.

And this comparison looks better to be after below L135 ~L138?

liuxuezhao (approved):

src/dtx/dtx_coll.c
Nasf-Fan: > seems the co_ver only set when it initially added to pool map? when it be excluded+reinted seems only set co_in_ver.
> in DTX code some places use co_ver, some other places use co_in_ver, probably need to unify them in a following PR.

Yes, use co_in_ver may filter out more unnecessary ranks, that is better.

> And this comparison looks better to be after below L135 ~L138?

Changing the comparison order will get the same result.

src/dtx/dtx_internal.h
```diff
@@ -76,14 +77,18 @@ CRT_RPC_DECLARE(dtx, DAOS_ISEQ_DTX, DAOS_OSEQ_DTX);
...
  * It can hold more than 19K engines inline RPC body.
  */
+/* clang-format off */
 #define DAOS_ISEQ_COLL_DTX						\
 	((uuid_t)		(dci_po_uuid)		CRT_VAR)	\
 	((uuid_t)		(dci_co_uuid)		CRT_VAR)	\
 	((struct dtx_id)	(dci_xid)		CRT_VAR)	\
 	((uint32_t)		(dci_version)		CRT_VAR)	\
+	((uint32_t)		(dci_min_rank)		CRT_VAR)	\
+	((uint32_t)		(dci_max_rank)		CRT_VAR)	\
```
mchaarawi, line 87: will this cause interoperability issue with different version of client / server?

mchaarawi (commented):
the PR should renable coll RPC by default for punch and stat too?
otherwise we also don't know since it's not tested in CI or other places

src/dtx/dtx_internal.h
Nasf-Fan: This is RPC among servers, will not cause interoperability issue between client and server.

> the PR should renable coll RPC by default for punch and stat too?
> otherwise we also don't know since it's not tested in CI or other places

Collective punch is disabled only on release/2.6, not on master. The patch for 2.6 (https://github.com/daos-stack/daos/pull/15975) enables collective punch explicitly.

mchaarawi: i see. thanks for explaining.

PR #15982, opened 2025-02-26
DAOS-10028 test: Update unit test helper

mjmac:
Reduce boilerplate in unit tests with test.MustLogContext(), which
takes care of creating a buffer-backed logger, injecting it into
a context, and then logging the contents of the buffer on test
failure.

### DIFF
--- a/src/control/common/test/utils.go
+++ b/src/control/common/test/utils.go
@@ -424,11 +424,36 @@ func Context(t *testing.T) context.Context {
 	return ctx
 }
 
-// MustLogContext returns a context containing the supplied logger.
+// MustLogContext returns a context containing a buffer-backed logger.
+// The contents of the buffer will be displayed on test failure. If the
+// optional parent context is not supplied, a default context is created.
 // Canceled when the test is done.
-func MustLogContext(t *testing.T, log logging.Logger) context.Context {
+func MustLogContext(t *testing.T, parents ...context.Context) context.Context {
 	t.Helper()
-	ctx, err := logging.ToContext(Context(t), log)
+
+	var parent context.Context
+	switch len(parents) {
+	case 0:
+		parent = Context(t)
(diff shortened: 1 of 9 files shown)


src/control/common/test/utils.go
```diff
@@ -424,11 +424,36 @@ func Context(t *testing.T) context.Context {
...
-	ctx, err := logging.ToContext(Context(t), log)
+
+	var parent context.Context
+	switch len(parents) {
+	case 0:
+		parent = Context(t)
+	case 1:
+		if parents[0] == nil {
+			// Pass through a nil context for tests that expect it.
+			return nil
```
kjacque, line 441: Do we have tests for nil contexts? I recall reading somewhere that a nil context is completely bad and unexpected behavior in Go, i.e. a "should never happen" situation.

kjacque (approved):
Nice job, good cleanup. Comment is nonblocking, at any rate it doesn't hurt anything.

src/control/common/test/utils.go
mjmac: Yeah, I usually always add a test case for it, and usually always remember to add a check for a nil context the code. Maybe we should stop doing that and just accept a segfault as a coding error that's hopefully caught in unit testing? Dunno, can see both sides of that argument.

kjacque: Yeah, I could go either way myself. I stopped creating test cases for a nil context a while ago specifically because of that though (I think we had a group discussion about it at the time). To be honest I think it'd be better to return an error than to just let the executable crash.

knard38 (approved):
LGTM

mjmac:
@daos-stack/daos-gatekeeper: PR only changes go unit tests -- can't be responsible for the hw stage failure. Please force-land.

PR #15994, opened 2025-02-28
DAOS-17142 rebuild: exit rebuild_tgt_status_check_ult when RPT stale

liuxuezhao:
rebuild_tgt_status_check_ult() ULT only can exit when rebuild abort or globally done. When rebuild done, the leader notifies each target server by IV LAZY SYNC, the notification lost due to some network issue and lead to dangling rebuild_tgt_status_check_ult.

### Steps for the author:

### DIFF
--- a/src/container/container_iv.c
+++ b/src/container/container_iv.c
@@ -473,14 +473,16 @@ again:
 				rc = cont_iv_snap_ent_create(entry, key);
 				if (rc == 0)
 					goto again;
-				D_DEBUG(DB_MD, "create cont snap iv entry failed "
-					""DF_RC"\n", DP_RC(rc));
+				D_DEBUG(DB_MD,
+					"cont " DF_UUID " create IV_CONT_SNAP iv entry failed\n",
+					DP_UUID(civ_key->cont_uuid));
 			} else if (class_id == IV_CONT_PROP) {
 				rc = cont_iv_prop_ent_create(entry, key);
 				if (rc == 0)
 					goto again;
-				D_ERROR("create cont prop iv entry failed "
-					""DF_RC"\n", DP_RC(rc));
+				DL_ERROR(rc,
+					 "cont " DF_UUID " create IV_CONT_PROP iv entry failed\n",
+					 DP_UUID(civ_key->cont_uuid));
 			} else if (class_id == IV_CONT_CAPA) {
 				struct container_hdl	chdl = { 0 };
 				int			rc1;
--- a/src/container/container_iv.c
+++ b/src/container/container_iv.c
@@ -473,15 +473,13 @@ again:
 				rc = cont_iv_snap_ent_create(entry, key);
 				if (rc == 0)
 					goto again;
-				D_DEBUG(DB_MD,
-					"cont " DF_UUID " create IV_CONT_SNAP iv entry failed\n",
-					DP_UUID(civ_key->cont_uuid));
+				DL_ERROR(rc, "cont " DF_UUID " create IV_CONT_SNAP iv entry failed",
+					 DP_UUID(civ_key->cont_uuid));
 			} else if (class_id == IV_CONT_PROP) {
 				rc = cont_iv_prop_ent_create(entry, key);
 				if (rc == 0)
 					goto again;
-				DL_ERROR(rc,
-					 "cont " DF_UUID " create IV_CONT_PROP iv entry failed\n",
+				DL_ERROR(rc, "cont " DF_UUID " create IV_CONT_PROP iv entry failed",
 					 DP_UUID(civ_key->cont_uuid));
 			} else if (class_id == IV_CONT_CAPA) {
 				struct container_hdl	chdl = { 0 };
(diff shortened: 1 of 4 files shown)


src/rebuild/srv.c
```diff
@@ -47,6 +47,44 @@ rebuild_pool_map_put(struct pool_map *map)
...
+
+	D_ASSERT(tls != NULL);
+	/* Only 1 thread will access the list, no need lock */
+	d_list_for_each_entry(pool_tls, &tls->rebuild_pool_list, rebuild_pool_list) {
+		if (uuid_compare(pool_tls->rebuild_pool_uuid, rpt->rt_pool_uuid) != 0)
+			continue;
+
+		if (rpt->rt_rebuild_ver == pool_tls->rebuild_pool_ver &&
+		    rpt->rt_rebuild_gen == pool_tls->rebuild_pool_gen)
+			found = true;
```
Nasf-Fan, line 72: Will we merge multiple rebuild instances for the same pool in the list?

liuxuezhao: Rebuild leader may try to merge to one rebuild, but in tgt server it will not merge the rpt for different versions/gens.
In a test we found orphan rebuild_tgt_status_check_ult left after rebuild one, reason not very clear, this PR is to add a kind of exiting method by a safe way.

Nasf-Fan (dismissed):

src/rebuild/srv.c
```diff
@@ -47,6 +47,44 @@ rebuild_pool_map_put(struct pool_map *map)
...
 
+/**
+ * Check whether the RPT is stale (new rebuild started).
+ * Only used in rebuild_tgt_status_check_ult(), the ULT only can exit when rebuild abort
+ * or globally done. When rebuild done, the leader notifies each target server by IV LAZY SYNC
+ * (see rebuild_leader_status_notify), so possibly the rebuild globally done but the async IV
+ * notification lost due to some network issue and lead to dangling rebuild_tgt_status_check_ult().
+ */
+bool
+rpt_stale(struct rebuild_tgt_pool_tracker *rpt)
```
kccain, line 58: With this new logic, the tgt ULT for a global done rebuild can be cleaned up the next time a new rebuild with a higher pool map version or generation number is started on the tgt engine? So most likely for a rebuild that is really globally done, this cleanup would happen with the Reclaim request, whenever it may be able to be received by this tgt engine (which may be having problems receiving RPCs from other engines)?

If we have a different sequence (outside scope of this change I guess?): rebuild (aborted) map version 285 gen 1, then a "Reclaim fail" version 268 gen 2, then Rebuild version 302 gen3, we expect that other existing logic in the tgt engine will recognize that the first rebuild is aborted, so that will get reaped? Because this logic here will not be able to detect that rebuild version 285 gen 1 is stale when "reclaim fail" version 268 gen 2 is in progress. It can be cleaned up I think when rebuild version 302 gen 3 arrives. Do I have that right?

kccain (approved):

kccain (changes requested):
Probably need to run this with Features: rebuild actually.

liuxuezhao:
> Probably need to run this with Features: rebuild actually.

the code change in this PR did not change core rebuild logic, it only check rpt_stale() when v.riv_pull_done which should be quite safe. The default testing already with lots of rebuild test cases, so I think unnecessary to add Features: rebuild to save CI resource.

kccain:
> > Probably need to run this with Features: rebuild actually.
> 
> the code change in this PR did not change core rebuild logic, it only check rpt_stale() when v.riv_pull_done which should be quite safe. The default testing already with lots of rebuild test cases, so I think unnecessary to add Features: rebuild to save CI resource.

OK that seems reasonable then.

The other comments in the patch were from GHA complaining about DL_ERROR() that has an unnecessary "\n". Will those be fixed in this PR?

kccain (dismissed):

src/container/container_iv.c
```diff
@@ -473,14 +473,16 @@ cont_iv_ent_fetch(struct ds_iv_entry *entry, struct ds_iv_key *key,
...
+				D_DEBUG(DB_MD,
+					"cont " DF_UUID " create IV_CONT_SNAP iv entry failed\n",
+					DP_UUID(civ_key->cont_uuid));
 			} else if (class_id == IV_CONT_PROP) {
 				rc = cont_iv_prop_ent_create(entry, key);
 				if (rc == 0)
 					goto again;
-				D_ERROR("create cont prop iv entry failed "
-					""DF_RC"\n", DP_RC(rc));
+				DL_ERROR(rc,
```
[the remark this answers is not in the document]
liuxuezhao: I did not try to change the log level, just to add the cont uuid to the log.
Probably can change L476 to D_ERROR, the extra "\n" can be removed, I'll change it in a following PR.

liuxuezhao:
> The other comments in the patch were from GHA complaining about DL_ERROR() that has an unnecessary "\n". Will those be fixed in this PR?

I'll fix the log format, may in a following PR.

daltonbohning (dismissed):
The lint check is required and the errors are introduced by this PR so IMO that should be fixed now.

liuxuezhao:
> The lint check is required and the errors are introduced by this PR so IMO that should be fixed now.

OK, fair enough, I'll fix it then.

kccain (approved):

Nasf-Fan (approved):

PR #15999, opened 2025-03-03
DAOS-10028 client: Fix sys-name handling

mjmac:
The pool bindings commit inadvertently introduced a regression
in the logic for handling the system name. If no system name is
supplied to daos_pool_connect(), then the system name obtained
from the agent via dc_mgmt_get_sys_info() should be used.

Includes a small fix to the dmg unit test helper to isolate it
from the environment of the user running the test.

Features: daos_cmd


kjacque (dismissed):

mchaarawi (approved):
i tested this and it resolves the issue.

tanabarr (approved):

kjacque (approved):

PR #16000, opened 2025-03-03
DAOS-17147 dfs: set dfs chunk size to a multiple of full stripe access

mchaarawi:
When the DFS object is using an EC oclass, and the dfs chunk size is not set by the user, user a better default for such files, where the chunk size would be at least the size of a full stripe access instead of the 1 MiB default.

### Steps for the author:

### DIFF
--- a/src/client/dfs/cont.c
+++ b/src/client/dfs/cont.c
@@ -23,7 +23,6 @@ suggest_dfs_cs(daos_handle_t poh, daos_prop_t *prop, uint64_t rf, daos_oclass_id
 	struct daos_oclass_attr *oc_attr;
 	struct daos_prop_entry  *dpe;
 	uint64_t                 ec_cell_size;
-	uint64_t                 val;
 	uint32_t                 nr_grps;
 	int                      rc;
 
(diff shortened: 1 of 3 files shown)


wangshilong (dismissed):

src/client/dfs/cont.c
```diff
@@ -15,6 +16,67 @@
...
+	dpe = daos_prop_entry_get(prop, DAOS_PROP_CO_EC_CELL_SZ);
+	if (dpe)
+		ec_cell_size = dpe->dpe_val;
+	else
+		ec_cell_size = DAOS_EC_CELL_DEF;
+
+	/** set the DFS chunk size to the 2 x the EC cell size x the number of data cells */
+	val = oc_attr->u.ec.e_k * ec_cell_size * 2;
+	if (val <= DFS_DEFAULT_CHUNK_SIZE)
+		*cs = DFS_DEFAULT_CHUNK_SIZE;
```
liuxuezhao, line 73: just confirm any reason why chunk size must >= 1MB?
For example if user define cell size as 100KB, for EC4P2, the "oc_attr->u.ec.e_k * ec_cell_size * 2" is 800K, why must set the chunk size to 1MB that is not full-stripe aligned?

src/client/dfs/cont.c
```diff
@@ -15,6 +16,67 @@
...
+		return 0;
+	}
+
+	/** query the EC cell size */
+	dpe = daos_prop_entry_get(prop, DAOS_PROP_CO_EC_CELL_SZ);
+	if (dpe)
+		ec_cell_size = dpe->dpe_val;
+	else
+		ec_cell_size = DAOS_EC_CELL_DEF;
```
liuxuezhao, line 68: if cell size not specified on dfs cont create, pool should with the cell size property that probably is not DAOS_EC_CELL_DEF.

src/client/dfs/cont.c
```diff
@@ -15,6 +16,67 @@
...
+	       daos_size_t *cs)
+{
+	struct daos_oclass_attr *oc_attr;
+	struct daos_prop_entry  *dpe;
+	uint64_t                 ec_cell_size;
+	uint64_t                 val;
+	uint32_t                 nr_grps;
+	int                      rc;
+
+	/** No EC above RF 2, use default CS */
```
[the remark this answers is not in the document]
mchaarawi: it is not well tested with 2.6 anyway, and we do not have it in the auto oclass selection either; so im fine to not include that here for now. we can update that later.

src/client/dfs/cont.c
mchaarawi: no reason. ill fix this.

src/client/dfs/cont.c
mchaarawi: ok ill check the pool property too in that case.

liuxuezhao (approved):

wangshilong (approved):

PR #16014, opened 2025-03-04
DAOS-17127 cart: Fix random rank calcultation

knard38:
### Description

When one or more ranks are excluded, the calculation for sending RPC to a service rank is invalid: it will eventually select a rank which has been excluded.
Moreover, if the system has been extended after evicting one or more ranks, then the new rank will never be used.

### Steps for the author:

### DIFF
--- a/docs/admin/administration.md
+++ b/docs/admin/administration.md
@@ -899,6 +899,31 @@ and then restarted to rejoin the system. An failed engine might also be excluded
 from the pools it hosted, please check the pool operation section on how to
 reintegrate an excluded engine.
 
+After one or more DAOS engines being excluded, the DAOS agent cache needs to be
+refresh.  For detailed information, please refer to the [1][System Deployment
+documentation].  Before refreshing the DAOS Agent cache, it should be checked
+that the exclusion information has been spread to the Management Service leader.
+This could be done thanks to the `dump-attachinfo` sub-command of the `daos_agent`
+executable:
+
+```bash
+daos_agent -o /tmp/daos_agent-tmp.yml dump-attachinfo
+```
+
+This usage of the `daos_agent` command needs a minimal DAOS agent configuration
+file `/tmp/daos_agent-tmp.yml` such as:
+
+```yaml
+name: daos_server
+access_points:
+- sertver-1
+port: 10001
+transport_config:
+  allow_insecure: true
+log_file: /tmp/daos_agent-tmp.log
+```
+
+
 ### Shutdown
 
 When up and running, the entire system can be shutdown.
@@ -1043,7 +1068,7 @@ dmg as follows:
 $ dmg storage format -l ${new_storage_node}
 ```
 
-new_storage_node should be replaced with the hostname or the IP address of the
+`new_storage_node` should be replaced with the hostname or the IP address of the
 new storage node (comma separated list or range of hosts for multiple nodes)
 to be added.
 
@@ -1055,6 +1080,11 @@ the system (this can be checked with `dmg system query -v`).
     nodes (if membership is not restricted on the dmg command line). That being
     said, existing pools won't be automatically extended to use the new servers.
     Please see the pool operation section for how to extend the pool membership.
+
+After extending the system, the cache of the `daos_agent` service of the client
+nodes needs to be refresh.  For detailed information, please refer to the
+[1][System Deployment documentation].
+
 
 ## Software Upgrade
 
@@ -1104,3 +1134,5 @@ Examples:
   * daos_server 2.4.0 is only compatible with daos_engine 2.4.0
   * daos_agent 2.6.0 is compatible with daos_server 2.4.0 (2.5 is a development version)
   * dmg 2.4.1 is compatible with daos_server 2.4.0
+
+[1]: <deployment.md#refresh-agent-cache>(Refresh DAOS Agent Cache)
--- a/docs/admin/administration.md
+++ b/docs/admin/administration.md
@@ -900,10 +900,10 @@ from the pools it hosted, please check the pool operation section on how to
 reintegrate an excluded engine.
 
 After one or more DAOS engines being excluded, the DAOS agent cache needs to be
-refresh.  For detailed information, please refer to the [1][System Deployment
+refreshed.  For detailed information, please refer to the [1][System Deployment
 documentation].  Before refreshing the DAOS Agent cache, it should be checked
 that the exclusion information has been spread to the Management Service leader.
-This could be done thanks to the `dump-attachinfo` sub-command of the `daos_agent`
+This could be done using the `dump-attachinfo` sub-command of the `daos_agent`
 executable:
 
 ```bash
@@ -1082,7 +1082,7 @@ the system (this can be checked with `dmg system query -v`).
     Please see the pool operation section for how to extend the pool membership.
 
 After extending the system, the cache of the `daos_agent` service of the client
-nodes needs to be refresh.  For detailed information, please refer to the
+nodes needs to be refreshed.  For detailed information, please refer to the
 [1][System Deployment documentation].
 
 
--- a/docs/admin/deployment.md
+++ b/docs/admin/deployment.md
@@ -1723,18 +1723,18 @@ Once the service file is installed and `systemctl daemon-reload` has been run to
 reload the configuration, the `daos_agent` can be started through systemd
 as shown above.
 
-#### Disable Agent Cache (Optional)
+#### Refresh Agent Cache
 
-In certain circumstances (e.g. for DAOS development or system evaluation), it
-may be desirable to disable the DAOS Agent's caching mechanism in order to avoid
-stale system information being retained across reformats of a system. The DAOS
-Agent normally caches a map of rank-to-fabric URI lookups as well as client network
-configuration data in order to reduce the number of management RPCs required to
-start an application. When this information becomes stale, the Agent must be
-restarted in order to repopulate the cache with new information.
-Alternatively, the caching mechanism may be disabled, with the tradeoff that
-each application launch will invoke management RPCs in order to obtain system
-connection information.
+In certain circumstances (e.g. [system extension][6] with new servers), it may
+be needed to refresh the DAOS Agent cache.  The DAOS Agent normally caches a map
+of rank-to-fabric URI lookups as well as client network configuration data in
+order to reduce the number of management RPCs required to start an application.
+When this information becomes stale, the cache could be refresh with sending the
+`SIGUSR2` to the `daos_agent` process.  Alternatively, the Agent could be
+restarted in order to repopulate the cache with new information.  Finally, the
+caching mechanism may be disabled, with the tradeoff that each application
+launch will invoke management RPCs in order to obtain system connection
+information.
 
 To disable the DAOS Agent caching mechanism, set the following environment
 variable before starting the `daos_agent` process:
@@ -1747,16 +1747,12 @@ the `[Service]` section before reloading systemd and restarting the
 
 `Environment=DAOS_AGENT_DISABLE_CACHE=true`
 
+It is also possible to disable the `daos_agent` cache with adding the following
+entry into the `daos_agent.yml` configuration file:
 
-[^1]: https://github.com/intel/ipmctl
-
-[^2]: https://github.com/daos-stack/daos/tree/master/utils/config
-
-[^3]: [https://www.open-mpi.org/faq/?category=running\#mpirun-hostfile](https://www.open-mpi.org/faq/?category=running#mpirun-hostfile)
-
-[^4]: https://github.com/daos-stack/daos/tree/master/src/control/README.md
-
-[^5]: https://github.com/pmem/ndctl/issues/130
+```yaml
+disable_caching: true
+```
 
 ## Multi-user DFuse setup
 
@@ -1771,3 +1767,16 @@ fuse and must be enabled by root before any user can use it.  To allow this then
 uncomment a line in `/etc/fuse.conf` to enable the `user_allow_other` setting.  The daos-client rpm
 does not do this automatically. An administrator must set this option on all nodes on which they
 want to provide a persistent multi-user dfuse service.
+
+
+[^1]: https://github.com/intel/ipmctl
+
+[^2]: https://github.com/daos-stack/daos/tree/master/utils/config
+
+[^3]: [https://www.open-mpi.org/faq/?category=running\#mpirun-hostfile](https://www.open-mpi.org/faq/?category=running#mpirun-hostfile)
+
+[^4]: https://github.com/daos-stack/daos/tree/master/src/control/README.md
+
+[^5]: https://github.com/pmem/ndctl/issues/130
+
+[6]: <administration.md#system-extension>(Extension of the DAOS system)
--- a/docs/admin/deployment.md
+++ b/docs/admin/deployment.md
@@ -1358,9 +1358,6 @@ engines:
 <end>
 ```
 
-There are a few optional providers that are not built by default. For detailed
-information, please refer to the [DAOS build documentation][6].
-
 !!! note
     DAOS Control Servers will need to be restarted on all hosts after updates to the server
     configuration file.
@@ -1369,9 +1366,6 @@ information, please refer to the [DAOS build documentation][6].
     include the hostnames or IP addresses (don't need to specify port) of those hosts.
 
     This will be the set of servers which host the replicated DAOS management service (MS).
-
->The support of the optional providers is not guarantee and can be removed
->without further notification.
 
 ### Network Configuration
 
@@ -1763,8 +1757,6 @@ the `[Service]` section before reloading systemd and restarting the
 [^4]: https://github.com/daos-stack/daos/tree/master/src/control/README.md
 
 [^5]: https://github.com/pmem/ndctl/issues/130
-
-[6]: <../dev/development.md#building-optional-components> (Building DAOS for Development)
 
 ## Multi-user DFuse setup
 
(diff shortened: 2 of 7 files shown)


src/mgmt/cli_mgmt.c
```diff
@@ -699,6 +718,7 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
 
 cleanup:
 	if (rc) {
+		D_FREE(g_serv_ranks);
```
mchaarawi, line 721: where would this be freed in the non-error case?

mchaarawi: probably should be in dc_mgmt_fini()?

mchaarawi (commented):
i was not aware of the agent maintaining a list of alive ranks. this would be a great improvement for an annoying issue we see when ranks get exluded.

src/mgmt/cli_mgmt.c
mchaarawi: actually NLT just failed because of that, so i think just freeing it in fini should resolve this.
```
{
   <insert_a_suppression_name_here>
   Memcheck:Leak
   match-leak-kinds: reachable
   fun:calloc
   fun:d_calloc
   fun:dc_mgmt_net_cfg
   fun:daos_init
   fun:_cgo_c695f4df3957_Cfunc_daos_init
   fun:runtime.asmcgocall.abi0
   obj:*
}
```

knard38: Indeed, there is some place where I need to free the allocation.
I am going to fix this asap.

- [ ] Fix memory leak

src/mgmt/cli_mgmt.c
```diff
@@ -604,6 +616,13 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
 	g_num_serv_ranks = resp->n_rank_uris;
 	D_INFO("Setting number of server ranks to %d\n", g_num_serv_ranks);
 
+	/* Save ranks id */
+	D_ALLOC_ARRAY(g_serv_ranks, g_num_serv_ranks);
```
frostedcmos, line 620: Either this function will need to be prevented from being called multiple times or previous g_serv_ranks allocation needs to be free-ed first here, or re-alloced

knard38: Yep, I was going to test if it could be called multiple times.  From my side, I will be in favour of realloacating (i.e. refreshing the rank table index).
@mchaarawi and @mchaarawi , I am also not sure if some special care should be taken with the function https://github.com/daos-stack/daos/blob/57701fa5147f96e63f066f712d2947efaf7cb28f/src/client/api/event.c#L101

src/client/api/rpc.c
```diff
@@ -107,14 +108,23 @@ struct rpc_proto {
...
+
 static void
 query_cb(struct crt_proto_query_cb_info *cb_info)
 {
 	struct rpc_proto	*rproto = (struct rpc_proto *)cb_info->pq_arg;
 	int			 rc;
 
 	if (daos_rpc_retryable_rc(cb_info->pq_rc)) {
-		rproto->ep.ep_rank = (rproto->ep.ep_rank + 1) % rproto->nr_ranks;
+		rproto->ep.ep_rank = get_rand_rank();
```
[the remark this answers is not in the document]
knard38: - [ ] Use round robin policies for retry

src/mgmt/cli_mgmt.c
frostedcmos: I dont think there is a concurrent access to this function; i think the main case would be as you mentioned earlier -- g_serv_ranks might be already allocated in the forked child, so just need to handle that case properly by handling re-alloc/re-use of this global buffer here

src/client/api/rpc.c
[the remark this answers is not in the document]
mchaarawi: correct me if im wrong, but technically if we get back the list of alive ranks only now, retry should rarely happen. Probably only in cases if while the agent was running and collected this list, then a server was excluded. so yes it's possible but not going to be as common as before.
not disagreeing with the roundrobin policy though, just calrifying.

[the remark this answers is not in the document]
frostedcmos: Yeah, I am more worried about constructed test scenarios where say we only have few engines to start with, kill 1 or 2, but then end up hitting random issues in CI due to bad luck/seed changing.

src/mgmt/cli_mgmt.c
knard38: > Indeed, there is some place where I need to free the allocation. I am going to fix this asap.
> 
> * [x]  Fix memory leak

Fixed with commit bae48ea068e35dd279bc347ba7b18f2befb62aa2

src/client/api/rpc.c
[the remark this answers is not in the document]
knard38: > * [x]  Use round robin policies for retry

Fixed with commit bae48ea068e35dd279bc347ba7b18f2befb62aa2

src/mgmt/cli_mgmt.c
knard38: > Yep, I was going to test if it could be called multiple times. From my side, I will be in favour of reallocating (i.e. refreshing the rank table index). @frostedcmos , do you know if we could have some concurrency issues with this function.
> 
> * [x]  Reallocate` g_serv_ranks`

Fixed with commit

knard38:
There is still some limitations related to the cache of the `daos_agent` and the update latency of the service rank pool map.
I am going to write some documentation on the way to deal with these limitations.
In a follow-up PR, I am also going to add a functional testing this issue.

src/mgmt/cli_mgmt.c
```diff
@@ -536,13 +537,35 @@ dc_mgmt_put_sys_info(struct daos_sys_info *info)
...
+
+	if (idx >= g_num_serv_ranks) {
+		D_ERROR("Invalid rank index: index=%d, ranks_num=%d\n", idx, g_num_serv_ranks);
+		return -DER_NONEXIST;
+	}
+
+	*rank =  g_serv_ranks[idx];
+
+	return -DER_SUCCESS;
+}
```
frostedcmos, line 567: We have a special CRT_NO_RANK rank value to indicate errors. With that you can eliminate need for separate rc with something like:
d_rank_t dc_mgmt_net_get_srv_rank(int idx) {
    return (idx < g_num_serv_ranks) ? g_serv_ranks[idx] : CRT_NO_RANK;
}

src/mgmt/cli_mgmt.c
```diff
@@ -697,6 +718,20 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
...
+	/* Save attached ranks id info */
+	g_num_serv_ranks = resp->n_rank_uris;
+	serv_ranks_tmp   = NULL;
+	if (g_num_serv_ranks > 0) {
+		D_ALLOC_ARRAY(serv_ranks_tmp, g_num_serv_ranks);
+		if (serv_ranks_tmp == NULL)
+			D_GOTO(cleanup, rc = -DER_NOMEM);
+		for (idx = 0; idx < g_num_serv_ranks; idx++)
+			serv_ranks_tmp[idx] = resp->rank_uris[idx]->rank;
+	}
```
frostedcmos, line 730: any reason for not using D_REALLOC_ARRAY instead of tmp array?

src/mgmt/cli_mgmt.c
```diff
@@ -697,6 +718,20 @@ dc_mgmt_net_cfg(const char *name, crt_init_options_t *crt_info)
...
+	D_FREE(g_serv_ranks);
+	g_serv_ranks = serv_ranks_tmp;
+	D_INFO("Setting number of attached server ranks to %d\n", g_num_serv_ranks);
```
frostedcmos, line 733: would be good to combine this print with line 714 above

src/mgmt/cli_mgmt.c
knard38: As we do not have concurrency issues it is not needed, however I prefer to always have a consistent value in the `g_serv_ranks` environment variable.  I also do not think that we could have performance constraints on this portion of code and thus `D_REALLOC_ARRAY` seems to not be mandatory.

src/mgmt/cli_mgmt.c
knard38: > would be good to combine this print with line 714 above

- [ ] Merge status comment

src/mgmt/cli_mgmt.c
knard38: > Did not know that special value. I will change it according to your comments.
> 
> * [x]  Use of `CRT_NO_RANK` for error management

Fixed with commit c8b0636

src/mgmt/cli_mgmt.c
knard38: > > would be good to combine this print with line 714 above
> 
> * [x]  Merge status comment

Fixed with commit c8b0636

frostedcmos (dismissed):

knard38:
@mchaarawi and @frostedcmos , please could you have check the documentation update.
Thanks in advance,
Cedric.

docs/admin/administration.md
```diff
@@ -899,6 +899,31 @@ and then restarted to rejoin the system. An failed engine might also be excluded
 from the pools it hosted, please check the pool operation section on how to
 reintegrate an excluded engine.
 
+After one or more DAOS engines being excluded, the DAOS agent cache needs to be
+refresh.  For detailed information, please refer to the [1][System Deployment
```
mchaarawi, line 903: refresh -> refreshed

docs/admin/administration.md
```diff
@@ -899,6 +899,31 @@ and then restarted to rejoin the system. An failed engine might also be excluded
...
+documentation].  Before refreshing the DAOS Agent cache, it should be checked
+that the exclusion information has been spread to the Management Service leader.
+This could be done thanks to the `dump-attachinfo` sub-command of the `daos_agent`
```
mchaarawi, line 906: this could be done using the ...

docs/admin/administration.md
```diff
@@ -1056,6 +1081,11 @@ the system (this can be checked with `dmg system query -v`).
     said, existing pools won't be automatically extended to use the new servers.
     Please see the pool operation section for how to extend the pool membership.
 
+After extending the system, the cache of the `daos_agent` service of the client
+nodes needs to be refresh.  For detailed information, please refer to the
```
mchaarawi, line 1085: refresh->refreshed

src/client/api/rpc.c
```diff
@@ -161,8 +164,16 @@ daos_rpc_proto_query(crt_opcode_t base_opc, uint32_t *ver_array, int count, int
...
-	rproto->nr_ranks   = dc_mgmt_net_get_num_srv_ranks();
-	rproto->ep.ep_rank = get_rand_rank();
+	nr_ranks = dc_mgmt_net_get_num_srv_ranks();
+	if (nr_ranks == 0) {
+		D_ERROR("failed to select an attached ranks: no attached ranks");
+		D_GOTO(out_free, -DER_NONEXIST);
+	}
+	rproto->rank_idx = d_rand() % nr_ranks;
+	rank             = dc_mgmt_net_get_srv_rank(rproto->rank_idx);
+	D_ASSERT(rank != CRT_NO_RANK);
```
mchaarawi, line 174: is there any chance this can be true in a situation where the agent is not running or so. asserts always scare me.

src/client/api/rpc.c
```diff
@@ -151,8 +164,16 @@ daos_rpc_proto_query(crt_opcode_t base_opc, uint32_t *ver_array, int count, int
 		D_GOTO(out_detach, rc = -DER_NOMEM);
 
 	/** select a random rank to issue the proto query rpc to */
-	rproto->nr_ranks   = dc_mgmt_net_get_num_srv_ranks();
-	rproto->ep.ep_rank = d_rand() % rproto->nr_ranks;
+	nr_ranks = dc_mgmt_net_get_num_srv_ranks();
```
mjmac, line 167: I know that you are just improving the existing logic here, but I wonder if it is really necessary to only use the MS ranks for the proto query. If I'm understanding correctly, clients will always do this proto query on startup, so in the case of a large number of clients all starting at the same time, a small number of ranks will have to service all of the proto queries. Maybe it would be better to pick from the entire set of valid ranks in order to spread the load?

mchaarawi: is this really what is happening here?
nr_ranks is all ranks not just ms ranks. we definitely do not want to just ping the ms ranks.

mjmac: Oops, you're right. I misunderstood the code. The `dc_mgmt_net_get_num_srv_ranks()` helper is returning the total number of ranks in the AttachInfo message. In any case, I think my point about sending the server protocol version in the AttachInfo payload instead of querying from every client still makes sense. But that's a different discussion that I'll bring up elsewhere.

docs/admin/administration.md
knard38: - [ ] Fix spelling

docs/admin/administration.md
knard38: - [ ] Fix frenglish

docs/admin/administration.md
knard38: - [ ] Fix spelling

src/client/api/rpc.c
knard38: It should never happen if `nr_ranks` is > 0 as the `nr_ranks` is equal to the number of elements in the conversion table used by the `dc_mgmt_net_get_srv_rank()` function.

knard38: The idea of this ASSERT and the one from line 126 is to detect some future regression which makes the previous properties not any more correct (i.e. for example concurrent access to the global variables `g_num_serv_ranks` and `g_serv_ranks`).

docs/admin/administration.md
knard38: > * [x]  Fix spelling

Fixed with commit 27abdcc47

docs/admin/administration.md
knard38: > * [x]  Fix frenglish

Fixed with commit 27abdcc47

docs/admin/administration.md
knard38: > * [x]  Fix spelling

Fixed with commit 27abdcc47

mchaarawi (dismissed):

docs/admin/deployment.md
```diff
@@ -1729,18 +1723,18 @@ Once the service file is installed and `systemctl daemon-reload` has been run to
...
+In certain circumstances (e.g. [system extension][6] with new servers), it may
+be needed to refresh the DAOS Agent cache.  The DAOS Agent normally caches a map
+of rank-to-fabric URI lookups as well as client network configuration data in
+order to reduce the number of management RPCs required to start an application.
+When this information becomes stale, the cache could be refresh with sending the
+`SIGUSR2` to the `daos_agent` process.  Alternatively, the Agent could be
+restarted in order to repopulate the cache with new information.  Finally, the
+caching mechanism may be disabled, with the tradeoff that each application
+launch will invoke management RPCs in order to obtain system connection
+information.
```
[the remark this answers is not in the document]
knard38: - [ ] Add third way to refresh the agent cache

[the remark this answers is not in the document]
knard38: > NP and thanks for the review:
> 
> * [x]  Add third way to refresh the agent cache

Fixed with commit 9008306f8

mjmac (approved):

mchaarawi (dismissed):

frostedcmos (approved):

knard38:
@daos-stack/daos-gatekeeper please could you lend this PR with the following message:
- title: `DAOS-17127 cart: Fix random rank calcultation`
- body:
```
Fix the code of the  DAOS lib calculating the target rank to send RPC.
Update the DAOS user documentation to properly handle system extension and rank exclusion.
```

knard38:
> > @daos-stack/daos-gatekeeper please could you lend this PR with the following message:
> > 
> > * title: `DAOS-17127 cart: Fix random rank calcultation`
> > * body:
> > 
> > ```
> > Fix the code of the  DAOS lib calculating the target rank to send RPC.
> > Update the DAOS user documentation to properly handle system extension and rank exclusion.
> > ```
> 
> PR has failed testing. can you please check?

Sure, thanks for the info.

knard38:
> > > @daos-stack/daos-gatekeeper please could you lend this PR with the following message:
> > > 
> > > * title: `DAOS-17127 cart: Fix random rank calcultation`
> > > * body:
> > > 
> > > ```
> > > Fix the code of the  DAOS lib calculating the target rank to send RPC.
> > > Update the DAOS user documentation to properly handle system extension and rank exclusion.
> > > ```
> > 
> > 
> > PR has failed testing. can you please check?
> 
> Sure, thanks for the info.

My apologies to the reviewers: I have missed some CI errors.

knard38:
All tests passed :-)

mjmac (commented):
Changes LGTM, but I'll defer to code owners for approvals.

frostedcmos (approved):

knard38:
@daos-stack/daos-gatekeeper please could you lend this PR with the following message:
- title: `DAOS-17127 cart: Fix random rank calcultation`
- body:
```
Fix the code of the  DAOS lib calculating the target rank to send RPC.
Update the DAOS user documentation to properly handle system extension and rank exclusion.
```

mchaarawi:
is there a 2.6 backport for the PR? i think it would be good to request for the ticket, if not in time for 2.6.3 at least for 2.6.4

knard38:
I have not asked for 2.6 back port.
I am going to reopen the ticket and ask for it.

knard38:
> is there a 2.6 backport for the PR? i think it would be good to request for the ticket, if not in time for 2.6.3 at least for 2.6.4

I have reopened the ticket https://daosio.atlassian.net/browse/DAOS-17251 and added the tags request for 2.6.3-rc3 and 2.6.4.

PR #16025, opened 2025-03-04
DAOS-XXX object: Improve firewall handling

jolivier23:
When a client is behind a firewall, rather than open it up to server connections, we can instead encode that information into the URI of the client.  In such cases where a connection would be needed, the server will return an error indicating such.  The client should ping servers involved in the request in such cases to establish a connection and then retry the request.

### Steps for the author:


jolivier23:
superceded by work on feature/firewall branch

PR #16045, opened 2025-03-06
DAOS-17094 pool: deep stack size for IV ULT

Nasf-Fan:
Use deep stack size for IV ULT to avoid stack overflow.

### Steps for the author:

### DIFF
--- a/src/pool/srv_iv.c
+++ b/src/pool/srv_iv.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2017-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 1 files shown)


liuxuezhao (approved):

mchaarawi (approved):

src/pool/srv_iv.c
```diff
@@ -1598,8 +1599,8 @@ ds_pool_iv_svc_fetch(struct ds_pool *pool, d_rank_list_t **svc_p)
 			D_GOTO(failed, rc);
 	} else {
 		/* create a ULT and schedule it on xstream-0 */
-		rc = dss_ult_execute(cont_pool_svc_ult, &ia, NULL, NULL,
-				     DSS_XS_SYS, DSS_ULT_DEEP_STACK, 0);
+		rc = dss_ult_execute(cont_pool_svc_ult, &ia, NULL, NULL, DSS_XS_SYS, 0,
+				     DSS_DEEP_STACK_SZ);
```
kccain, line 1603: I think this is wrong. DSS_DEEP_STACK_SZ=2 , it seems incorrect to pass the literal value 2 into this call stack, which will eventually invoke ABT_thread_attr_set_stacksize() as the stacksize argument.

Here is the one other example in the engine code that specifies a specific, nonzero stack_size argument, and it uses a number of bytes argument https://github.com/daos-stack/daos/blob/6dae36aad9d5d66ac39ce1a7aca9210d4d911808/src/object/srv_obj_migrate.c#L3264

kccain (changes requested):
Something seems incorrect, unless I am missing something.

src/pool/srv_iv.c
kccain: Actually, I stand corrected. I did not look closely enough to see that the code change removes the use of DSS_ULT_DEEP_STACK=2 and replaces it with DSS_DEEP_STACK_SZ=65536. Sorry about that.

kccain (approved):

src/pool/srv_iv.c
gnailzenh: I think ds_dtx_resync() has the same issue, could you also add the change to this PR?

Nasf-Fan: Up to now, `ds_dtx_resync()` does not need deep stack size. But for `cont_pool_svc_ult()`, the original author explicitly pointed out that it required deep stack, but used wrong parameter.

PR #16048, opened 2025-03-06
DAOS-15947 control: Enable replacement of rank after loss of metadata

tanabarr:
Add dmg storage format --replace flag to update UUID key in MS-db for
a given storage node engine. This enables a rank to be re-integrated
after SCM metadata has been lost e.g. due to a PMem DIMM failure.
Previously an excluded rank in this state would join with a new rank
after reformat.

Replacement will only succeed if the following are met:
- Identical control address, fault domain, fabric primary secondary
  IDs and contexts are supplied in the join request.
- Rank that matches joining engine is not enabled on any pools.
- Rank that matches joining engine is not in AdminExcluded state.

Replacement will also
Features: control
Allow-unstable-test: true

### Steps for the author:


Michael-Hennecke (commented):
LGTM, but a question on terminology: You are using `rejoin` in some contexts and `replace` in others. Should we unify this and use the same term everywhere? To me, `replace` conveys more clearly that we have indeed replaced some hardware components which causes this operation to be run, while `rejoin` sounds like a more "harmless" operation. Would like to see and comment on the how-to, so we can make all of this consistent.

tanabarr:
> LGTM, but a question on terminology: You are using `rejoin` in some contexts and `replace` in others. Should we unify this and use the same term everywhere? To me, `replace` conveys more clearly that we have indeed replaced some hardware components which causes this operation to be run, while `rejoin` sounds like a more "harmless" operation. Would like to see and comment on the how-to, so we can make all of this consistent.

Happy to switch everything to "replace", was using "rejoin" as that was the name of the flag originally suggested. The code around the membership change uses "replace" as "rejoin" doesn't make sense in that context hence the use of the different terms. I'm also adding some checks and balances as per suggestions in the ticket comments.

src/control/system/membership.go
```diff
@@ -198,11 +208,93 @@ func (m *Membership) Join(req *JoinRequest) (resp *JoinResponse, err error) {
...
+			return nil, errors.Wrap(err, "failed to get all members")
+		}
+
+		var memberToReplace *Member
+		for _, cm := range currentMembers {
+			// Only match identical member with different UUID.
+			switch {
+			case cm.Addr.String() != req.ControlAddr.String(),
+				cm.PrimaryFabricURI != req.PrimaryFabricURI,
+				!slices.Equal(cm.SecondaryFabricURIs,
```
knard38, line 235: From my understanding `SecondaryFabricURIs` and `SecondaryFabricContexts` are not sorted.  Maybe we should sort them before testing checking if they have the same content.

src/control/system/membership.go
```diff
@@ -198,11 +208,93 @@ func (m *Membership) Join(req *JoinRequest) (resp *JoinResponse, err error) {
...
+
+		if err := m.db.AddMember(memberToReplace); err != nil {
+			return nil, errors.Wrap(err, "adding new uuid member in rejoin")
+		}
+
+		newCount, err := m.Count()
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get member count")
+		}
+		if newCount != oldCount {
```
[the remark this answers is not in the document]
tanabarr: the problem with this is that the calls into svc.db contain locking themselves, maybe this test is unnecessary or maybe I could do it another way by verifying the old UUID is missing and the new one exists but then again maybe that's also necessary ? thoughts

src/control/system/membership.go
tanabarr: I was under the impression that these lists should be identical for a rejoining rank, @kjacque what do you think?

tanabarr:
@liw @wangshilong does https://github.com/daos-stack/daos/pull/16048/commits/50a6a8d8fba32baecd7f34fd5a85b6afa0b684e5 address the concerns regarding removing a Rank that is being used in pools? @mjmac @kjacque how does the approach look?

wangshilong:
> @liw @wangshilong does [50a6a8d](https://github.com/daos-stack/daos/commit/50a6a8d8fba32baecd7f34fd5a85b6afa0b684e5) address the concerns regarding removing a Rank that is being used in pools? @mjmac @kjacque how does the approach look?

I leave a question in the commit, by default poolQuery will not return enabled_ranks.

src/control/server/mgmt_system.go
```diff
@@ -209,7 +271,23 @@ func (svc *mgmtSvc) join(ctx context.Context, req *mgmtpb.JoinReq, peerAddr *net
...
+	}
+
+	if req.ReplaceMode {
+		rankToReplace, err := svc.membership.FindRankFromJoinRequest(joinReq)
+		if err != nil {
+			return nil, errors.Wrap(err, "join: replace existing rank")
+		}
+		if rankToReplace == ranklist.NilRank {
+			return nil, FaultJoinReplaceRankNotFound(joinReq)
+		}
```
kjacque, line 283: Maybe the "not found" fault should come from `svc.membership.FindRankFromJoinRequest(joinReq)`...

src/control/system/membership.go
kjacque: We identify the secondary providers by an index, so the order should be consistent.

src/control/lib/control/system.go
```diff
@@ -134,6 +134,7 @@ type SystemJoinReq struct {
 	InstanceIdx          uint32              `json:"idx"`
 	Incarnation          uint64              `json:"incarnation"`
 	CheckMode            bool                `json:"check_mode"`
+	ReplaceMode          bool                `json:"replace_mode"`
```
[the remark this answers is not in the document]
tanabarr: done

src/control/server/ctl_storage_rpc.go
```diff
@@ -852,6 +855,15 @@ func formatScm(ctx context.Context, req formatScmReq, resp *ctlpb.StorageFormatR
 		}
 	}
 
+	if req.rejoin {
+		// Only valid if single engine requires format.
+		if len(needFormat) != 1 {
```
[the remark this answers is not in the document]
tanabarr: The initial solution is focused on PMem failure which is extremely unlikely to happen on both ranks on the same storage node. In the case that one engine needs format for another reason other than PMem failure and the other needs replacement due to PMem failure, workaround could be to run format without `--replace` to  standard format both then wipe engine that needs replacement and try again with `--replace` after the standard format has been performed.

src/control/system/membership.go
```diff
@@ -198,11 +208,93 @@ func (m *Membership) Join(req *JoinRequest) (resp *JoinResponse, err error) {
...
+			return nil, errors.Errorf("expect the same number of members before and "+
+				"after rejoin, want %d got %d", oldCount, newCount)
+		}
```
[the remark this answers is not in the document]
tanabarr: removed

src/control/system/membership.go
```diff
@@ -198,11 +208,93 @@ func (m *Membership) Join(req *JoinRequest) (resp *JoinResponse, err error) {
...
+
+		return resp, err
+	}
 
 	if err := m.checkReqFaultDomain(req); err != nil {
 		return nil, err
 	}
 
+	// Process join request for a new rank.
```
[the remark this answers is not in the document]
tanabarr: separated out, done

src/control/lib/hostlist/numericlist_test.go
```diff
@@ -84,6 +85,38 @@ func TestHostList_NumericList(t *testing.T) {
 	}
 }
 
+func TestHostList_NumericList_Contains(t *testing.T) {
```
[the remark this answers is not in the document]
tanabarr: done

src/control/server/mgmt_system.go
tanabarr: done

src/control/server/ctl_storage_rpc.go
```diff
@@ -1007,6 +1019,7 @@ func (cs *ControlService) StorageFormat(ctx context.Context, req *ctlpb.StorageF
 		return resp, nil
 	}
 
+	// DAOS-15947: Do we need to pass in the replace flag here?
```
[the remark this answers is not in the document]
tanabarr: done

src/control/server/mgmt_system.go
```diff
@@ -180,6 +180,68 @@ func getPeerListenAddr(ctx context.Context, listenAddrStr string) (*net.TCPAddr,
...
+		svc.log.Debug("checking replace mode rank: zero pools to verify")
+		return nil // No pools to query.
+	}
+
+	// Verify rank to replace is not enabled on any pool.
+	for _, id := range poolIDs {
+		rs := poolRanks[id]
+		if rs.Count() == 0 {
+			return errors.Errorf("no ranks in map for pool %s", id)
+		}
```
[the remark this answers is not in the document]
tanabarr: yes basically an assert

src/control/server/mgmt_system.go
```diff
@@ -180,6 +180,68 @@ func getPeerListenAddr(ctx context.Context, listenAddrStr string) (*net.TCPAddr,
...
+
+	// Retrieve rank-to-pool mappings.
+	rl := ranklist.RankList{rankToReplace}
+	poolIDs, poolRanks, err := svc.getPoolsRanks(ranklist.RankSetFromRanks(rl))
+	if err != nil {
+		return err
+	}
+
+	if len(poolIDs) != len(poolRanks) {
+		return errors.New("nr poolIDs should be equal to poolRanks keys")
```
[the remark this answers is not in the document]
tanabarr: added details to error and moved into relevant function, done

tanabarr:
@kjacque have attempted to address your review comments, how does it look? If ok I will add the unit tests

src/control/server/ctl_storage_rpc.go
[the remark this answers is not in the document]
kjacque: Hmm, if we run format without replace on a repaired system, don't both engines join automatically with new ranks? Or do we have a way to prevent joining?

To be honest I'd prefer that replace work whether one or both engines need to be replaced.

src/control/server/ctl_storage_rpc.go
```diff
@@ -769,11 +769,16 @@ func (cs *ControlService) StorageScan(ctx context.Context, req *ctlpb.StorageSca
...
-func (cs *ControlService) formatMetadata(instances []Engine, reformat bool) (bool, error) {
+func (cs *ControlService) formatMetadata(instances []Engine, reformat, replace bool) (bool, error) {
 	// Format control metadata first, if needed
 	if needs, err := cs.storage.ControlMetadataNeedsFormat(); err != nil {
 		return false, errors.Wrap(err, "detecting if metadata format is needed")
 	} else if needs || reformat {
+		if replace == true {
+			return false, errors.New("control metadata formatting is not possible " +
+				"when rank replace is requested")
+		}
```
kjacque, line 780: Hmm, I guess I assumed you might do something special to support this case, like repair the affected engine's metadata subdirectory, instead of rejecting it outright. Since this change is to address a pmem problem, I'm fine with not supporting MD-on-SSD right away. Might be better to check if it's an MD-on-SSD config immediately after getting the format request, and deny any `--replace` request in that case.

tanabarr: I haven't worked out the details of the MD-on-SSD case yet as I haven't seen it in the wild, thinking about it if we have a media failure on a control_metadata storage device the whole thing would need to be rewritten anyway so not sure it's necessary to have the ability to rewrite individual engine control_metadata for this case. Maybe we could allow control_metadata format only in the case where both ranks require format and replace is requested. This would involve relaxing the restriction on replace not being allowed if both ranks require format (which is one of your other review comments) and then rely on the `joinReplace()` logic to reject inappropriate replacement requests.

src/control/server/ctl_storage_rpc.go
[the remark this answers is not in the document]
tanabarr: I don't think we have a way to selectively format in that case, I agree that we should relax dual-engine-replace-format case, the repercussion is that in the situation where we have one engine requiring replacement and one engine just requiring join (which would be unusual but still possible) then the replace format would succeed but join would fail for the non-replacement engine and the user would have to run normal format again for the non-replacement engine. I think this is acceptable for the edge case.

src/control/server/ctl_storage_rpc.go
kjacque: That makes sense to me.

kjacque (dismissed):
Changes LGTM

src/control/server/ctl_storage_rpc.go
```diff
@@ -852,6 +855,12 @@ func formatScm(ctx context.Context, req formatScmReq, resp *ctlpb.StorageFormatR
 		}
 	}
 
+	if req.replace && len(needFormat) == 0 {
+		// Only valid if at least one engine requires format.
+		return nil, nil, errors.New("format replace option only valid if at " +
+			"least one engine requires format but no engines need format")
```
knard38, line 861: This message is not very clear to me, do you mean:
```suggestion
			"least one engine requires format but no other engines need format")
```
If yes, the test should test `len(needFormat) != 1`

src/control/system/membership_test.go
```diff
@@ -860,6 +861,71 @@ func TestSystem_Membership_Join(t *testing.T) {
 				MapVersion: expMapVer,
 			},
 		},
+		//		"rejoin with existing UUID and nil rank; replace mode set in request": {
```
knard38, line 864: Not sure to understand why these tests are disabled.

knard38 (commented):
Mostly OK for me.
Just some points which are not clear to me.

tanabarr:
@mjmac @kjacque  @knard38 reviews please at your earliest convenience, critical priority issue

knard38 (approved):
Latest modifications LGTM

tanabarr:
@kjacque @mjmac please can I get a review on this P1 issue PR?

src/control/system/pool.go
```diff
@@ -88,51 +84,35 @@ func (pss *PoolServiceStorage) CreationRanks() []ranklist.Rank {
...
-
-func (pss *PoolServiceStorage) String() string {
-	if pss == nil {
-		return "no pool storage info available"
-	}
-	return fmt.Sprintf("total SCM: %s, total NVMe: %s",
-		humanize.Bytes(pss.TotalSCM()),
-		humanize.Bytes(pss.TotalNVMe()))
-}
+//// TotalSCM returns the total amount of SCM storage allocated to
```
kjacque, line 87: Did this need to be removed? If a future PR will re-enable, should add a comment pointing to the ticket.

kjacque (approved):
Since my only comment is comment-related, I am OK with including it in a follow-on.

tanabarr:
Behaviour is documented in the following wiki page: https://daosio.atlassian.net/wiki/spaces/DAOS/pages/12180848662/Replace+a+rank+when+metadata+is+lost

Gatekeeper please use the PR title and description as commit message when landing. TIA

src/control/system/pool.go
tanabarr: This needs to be removed, I will remove as part of a follow-on PR, thanks.

PR #16057, opened 2025-03-07
b/390177544 cart: Support dynamic mercury/libfabric loglevels (#15738)

mjmac:
Mercury and libfabric are much too spammy to enable at lower
than warning level by default. Mercury does support adjustment
of its log level and logged subsystems at runtime, so this
patch provides access to that functionality from the
existing `dmg server set-logmasks` tool.

This version of the patch adds support for writing mercury/libfabric
debug logs to a separate file, in order to avoid undue burden on
the cloud logging pipeline. To use it, set HG_DBG_LOG_FILE in the
engine config to a logfile path. NB: This is a "raw" log file, i.e.
no rotation is done and it's flushed once per second.

To enable debug logging for just the mercury/libfabric libraries:
  dmg server set-logmasks -m EXTERNAL=DEBUG

To disable debug logging again:
  dmg server set-logmasks -m EXTERNAL=ERR

Note that the CaRT/Mercury wrapper (crt_hg.c) still uses the normal
dlog logger with the "hg" facility. If debug logging in this file is
desired, then the mask HG=DEBUG may be used for set-logmasks.
Be aware that the debug logging produced by this file is rather
copious and may cause problems for cloud logging, however!

Change-Id: Ib6fb74c42507dbfba2c5984a92433ad43b6e2df3

### DIFF
--- a/src/cart/crt_hg.c
+++ b/src/cart/crt_hg.c
@@ -737,14 +737,12 @@ out:
 	return rc;
 }
 
-static FILE    *hg_dbg_log        = NULL;
-static uint64_t last_hg_dbg_flush = 0;
+static FILE *hg_dbg_log = NULL;
 
 static int
 crt_hg_log(FILE *stream, const char *fmt, ...)
 {
-	va_list		ap;
-	struct timeval  tv;
+	va_list         ap;
 	int		flags;
 
 	flags = d_log_check((intptr_t)stream);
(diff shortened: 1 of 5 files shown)


src/cart/crt_hg.c
```diff
@@ -737,16 +737,37 @@ crt_get_info_string(bool primary, crt_provider_t provider, int iface_idx,
...
+	/*
+	 * b/401315986: If a separate logfile has been defined for
+	 * Mercury debug logging, send DEBUG-level messages to it.
+	 */
+	if (hg_dbg_log != NULL && (flags & DLOG_PRIMASK) == DLOG_DBG) {
+		va_start(ap, fmt);
+		vfprintf(hg_dbg_log, fmt, ap);
+		va_end(ap);
+
+		(void)gettimeofday(&tv, 0);
```
jolivier23, line 763: I wonder if we should consider making this more generic.  In other words, what if we did this at the DAOS level (e.g D_LOG_DEBUG_FILE) ?  This would allow us to turn on debug generically without flooding log explorer if we so choose.

Otherwise, you lose the per message timestamp, no?

jolivier23: Another advantage is that we buffer messages below warn

mjmac:
Superseded by #16057

PR #16069, opened 2025-03-10
DAOS-16908 object: add client-side target pinging on update retry

karthjyojay:
In the case of update bulk transfers, the server targets may try to communicate with clients who have not yet established a connection with them. In this case, the connection operation hangs forever. In Google Parallelstore, this is an issue because we do not have control over the client firewalls. To deal with this issue, we will have the server tell the client that they cannot connect to the client due to this scenario and the client will ping the affected targets before retrying their update RPC.

The purpose of this PR is to introduce client-side logic to ping targets involved in an update RPC.

### Steps for the author:


src/object/obj_internal.h
```diff
@@ -838,7 +839,8 @@ obj_retry_error(int err)
 	       err == -DER_GRPVER || err == -DER_EXCLUDED || err == -DER_CSUM ||
 	       err == -DER_TX_BUSY || err == -DER_TX_UNCERTAIN || err == -DER_NEED_TX ||
 	       err == -DER_NOTLEADER || err == -DER_UPDATE_AGAIN || err == -DER_NVME_IO ||
-	       err == -DER_CHKPT_BUSY || err == -DER_OVERLOAD_RETRY || daos_crt_network_error(err);
+	       err == -DER_CHKPT_BUSY || err == -DER_OVERLOAD_RETRY || err == -DER_CLIENT_UNREACH ||
```
mchaarawi, line 842: i don't understand why unreachable is a retryable error. that usually means exactly that and it will not become reachable again.

src/object/srv_obj.c
```diff
@@ -1592,6 +1592,13 @@ obj_local_rw_internal(crt_rpc_t *rpc, struct obj_io_context *ioc, daos_iod_t *io
 	if (orw->orw_flags & ORF_CHECK_EXISTENCE)
 		goto out;
 
+	if (DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE) && obj_rpc_is_update(rpc)) {
+		/** Fault injection - client unreachable. */
+		D_INFO("enabled fault injection client unreachable");
+		rc = -DER_CLIENT_UNREACH;
+		goto out;
+	}
```
mchaarawi, line 1600: is this PR just for fault injection?
I have missed where we can return that error code anywhere else from the server..

src/object/cli_obj.c
```diff
@@ -1403,6 +1415,96 @@ obj_pool_query_task(tse_sched_t *sched, struct dc_object *obj,
...
+	a->obj      = obj;
+	a->obj_auxi = obj_auxi;
+
+	*task = t;
+	dc_pool_put(pool);
+	return 0;
+}
+
+int
+obj_tgt_ping_task(tse_sched_t *sched, struct dc_object *obj, struct obj_auxi_args *obj_auxi,
```
mchaarawi, line 1486: the entire sequence of this function does not seem necessary to me:
obj_tgt_ping_task -> create_ping_task -> ping_target
why not just put ping_target behind a task API and use that?

src/pool/cli.c
```diff
@@ -725,6 +726,100 @@ warmup_cb(const struct crt_cb_info *info)
 	sem_post(sem);
 }
 
+int
+ping_target(int tgt_id, daos_handle_t pool_hdl)
```
mchaarawi, line 730: why is this a pool client function and not in the object module?
is it anticipated to be used anywhere else?

mchaarawi (changes requested):
TBH im not sure what is the purpose of this PR. there is no description in the PR itself and the ticket looks like an unrelated or maybe just a parent ticket.
This is probably the first step to be able to give more meaningful feedback

jolivier23:
> TBH im not sure what is the purpose of this PR. there is no description in the PR itself and the ticket looks like an unrelated or maybe just a parent ticket. This is probably the first step to be able to give more meaningful feedback

@karthjyojay probably meant to make this a draft but yeah, you need to have a shorter title and a description in your PR comments.

For context, @mchaarawi the purpose of this change will be to support clients that are behind a firewall.  See https://daosio.atlassian.net/browse/DAOS-16906

Long story short, when mercury and libfabric changes are in and enabled, libfabric will return an error when an endpoint marked as behind a firewall can't be contacted.  This is to avoid user having to poke massive holes in their firewall for parallelstore to work.  Instead, the client will always ping to reestablish/create a connection if one does not already exist.

src/pool/cli.c
jolivier23: This is primarily because it's reusing the same RPC used by warmup_pool which is defined there.

karthjyojay:
@mchaarawi , my apologies. I meant to just give this to @jolivier23 and @wangdi1. I've put it back into draft.

src/object/obj_internal.h
jolivier23: with writes, the request goes from client to one engine (dtx leader) and is forwarded.  This is for the case where the non-leader tries to send a request to the client for bulk transfer but doesn't already have a connection and the client is behind a firewall so the server is unable to create a connection.   We want libfabric to return a special error in this case and we will return that error to the original client informing the client to ping the server targets (to establish connection from client side) before retrying the update.

src/object/cli_obj.c
```diff
@@ -1403,6 +1415,96 @@ obj_pool_query_task(tse_sched_t *sched, struct dc_object *obj,
...
+	obj                       = arg->obj;
+	obj_auxi                  = arg->obj_auxi;
+
+	grp_idx      = obj_dkey2grpidx(obj, obj_auxi->dkey_hash, obj->cob_version);
+	first_target = grp_idx * obj->cob_grp_size;
+
+	for (i = 0, target = first_target; i < obj->cob_grp_size; i++, target++) {
+		tgt_id = obj->cob_shards->do_shards[target].do_pl_shard.po_target;
+
+		rc = ping_target(tgt_id, pool_hdl);
```
jolivier23, line 1442: I think we will ultimately want to send these in parallel though that's an optimization and can probable be refined later.

src/object/srv_obj.c
jolivier23: Yes, there are other PRs in flight but the point was to get the client code working in parallel with those other changes.

src/object/cli_obj.c
wangdi1: We need a separate a task here to ping all necessary targets, then schedule retry here.  So it will need create a task anyway? 

you mean call ping_target directly in obj_retry_cb()? or you mean create task api for ping target(), which I do not see much difference? or I miss your comments?

src/include/daos/pool.h
```diff
@@ -200,6 +201,8 @@ int dc_pool_choose_svc_rank(const char *label, uuid_t puuid,
 int dc_pool_create_map_refresh_task(daos_handle_t pool_hdl, uint32_t map_version,
 				    tse_sched_t *sched, tse_task_t **task);
 void dc_pool_abandon_map_refresh_task(tse_task_t *task);
+int
+ping_target(int tgt_id, daos_handle_t pool_hdl);
```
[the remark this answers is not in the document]
karthjyojay: Done.

src/object/cli_obj.c
karthjyojay: I made b/402224475 to capture this.

src/object/obj_internal.h
mchaarawi: i would suggest a different error code then. -DER_NO_CONNECTION  or -DER_RECONNECT ?
since we already have DER_UNREACH, this one seems a little confusing

src/object/srv_obj.c
mchaarawi: maybe a good candidate for a feature branch then?

src/object/cli_obj.c
mchaarawi: in that case maybe change the ping_taget call to ping_targets (with an s) to for multiple RPC in-flight?
my thought was yes to create an task API for this (not a public API). in essence it's the same thing, but the code for creating a task is already implemented in dc_task_create(). 
anyway, I will that up to you. this way works fine too.

mchaarawi: > in that case maybe change the ping_taget call to ping_targets (with an s) that takes an array, for multiple RPC in-flight? my thought was yes to create an task API for this (not a public API). in essence it's the same thing, but the code for creating a task is already implemented in dc_task_create(). anyway, I will leave that up to you. this way works fine too.

karthjyojay: I will leave it for now because ping_target is working with one target, if that's ok with you.

src/object/srv_obj.c
karthjyojay: If we move to a feature branch, are you ok with moving to a feature branch after this PR is landed?

src/object/obj_internal.h
karthjyojay: Done.

src/object/cli_obj.c
karthjyojay: Resolving this, please reopen as needed.

src/pool/cli.c
karthjyojay: Resolving this, please reopen as needed.

src/object/srv_obj.c
karthjyojay: Resolving this, please reopen as needed.

jolivier23: I hate to use a feature branch for a small feature.   This code is the majority of the feature.  The rest is just mercury/libfabtic and this can be tested independently.

src/object/cli_obj.c
jolivier23: I think a single target ping is fine to start.  Before we roll this out, we will want to send multiple RPCs

jolivier23 (dismissed):

src/object/srv_obj.c
mchaarawi: well TBH this PR sounds a little less than half baked to land to master. Are the mercury and ofi changes landed?
i can't even tell how this is going to be used on the server side. you don't have to use a feature branch, just use this branch as a baseline..

karthjyojay: We can continue to use this branch as a baseline. For more features built on this, we can make new child PRs that are based on this branch as a baseline and then slowly merge onto this PR.

mchaarawi: the main advantage of a feature branch is that you can land PRs to it incrementally (have to be reviewed and passing tests same way as it it would be for master). then once all PRs in, the branch can be just landed to master without needing to review all the changes again. provided it passes CI testing of course.

karthjyojay: We made a feature branch and changed this PR to merge my dev branch into this feature branch.

karthjyojay: When you get a chance, @mchaarawi, can you please take a look again?

karthjyojay: @mchaarawi @jolivier23 , sorry, it looks like for some reason Jenkins is not able to build my PR. Jeff told me the way around this is making a new PR. Please hold off on re-reviewing this while I get a new PR ready.

PR #16074, opened 2025-03-11
DAOS-504 pool: Add backoffs to rsvc operations

liw:
Add backoffs to retries of pool and container rsvc operations.

Move tse_task_reinit calls to the end of each completion callback, in order to reduce the chance of new data races between a task and its reinitialized instance.

Features: pool container

### Steps for the author:


jolivier23 (approved):

src/pool/cli.c
```diff
@@ -696,12 +736,8 @@ pool_rsvc_client_complete_rpc(struct dc_pool *pool, const crt_endpoint_t *ep,
 				      &out->po_hint);
 	D_MUTEX_UNLOCK(&pool->dp_client_lock);
 	if (rc == RSVC_CLIENT_RECHOOSE ||
-	    (rc == RSVC_CLIENT_PROCEED && daos_rpc_retryable_rc(out->po_rc))) {
-		rc = tse_task_reinit(task);
-		if (rc != 0)
-			return rc;
+	    (rc == RSVC_CLIENT_PROCEED && daos_rpc_retryable_rc(out->po_rc)))
```
kccain, line 739: only the most minor note that is probably not worth changing - the comments preceding the function (here, and in container code function) indicate that if RSVC_CLIENT_RECHOOSE is returned, the task is reinited. Now with this change, I guess it means the task will be reinited by the caller.

src/pool/cli.c
```diff
@@ -1850,8 +1886,8 @@ pool_query_cb(tse_task_t *task, void *data)
 		D_INFO("retry with map buffer size required by server (%ul)\n",
 		       out_v5->pqo_map_buf_size);
 		pool->dp_map_sz = out_v5->pqo_map_buf_size;
-		rc = tse_task_reinit(task);
-		D_GOTO(out, rc);
+		reinit          = true;
+		D_GOTO(out, rc = 0);
```
kccain, line 1890: interesting approach to set rc = 0 here, and also in the -DER_AGAIN reinit=true case below. Is this just for clarity / no potential confusion reading the code? When we get to the out label, with reinit=true, rc will be reassigned to the result of the task reinit call.

kccain (approved):

src/pool/cli.c
liw: My bad. Will definitely fix the comment, probably in a follow-on doc-only PR to save time/CI resource.

src/pool/cli.c
liw: Yes, I had exactly the same thought when authoring this change. And yes, the unnecessary `rc = 0` is for to express the intention that "reinitializing the task implies that we want this function to return 0 instead of `-DER_TRUNC`".

src/pool/cli.c
liw: https://github.com/daos-stack/daos/pull/16174

liw:
Pinging @daos-stack/daos-gatekeeper...

PR #16081, opened 2025-03-12
DAOS-17271 pool: Fix handle_event -DER_NONEXISTs

liw:
When handling the exclusion of multiple ranks,
pool_svc_update_map_internal aborts the whole request and returns -DER_NONEXIST if any of the ranks is absent in the pool map. This is correct for the dmg case, but problematic for the handle_event case, where ranks not in the pool map should simply be ignored. (See the Jira ticket for more.)

### Steps for the author:

### DIFF
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -7033,7 +7033,7 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc, bool exclud
 							   inval_tgt_addrs);
 			if (rc != 0)
 				goto out_map;
-			if (inval_tgt_addrs->pta_number > 0) {
+			if (src == MUS_DMG && inval_tgt_addrs->pta_number > 0) {
 				/*
 				 * If any invalid ranks/targets were specified here,
 				 * abort the entire request. This will mean the
@@ -7604,8 +7604,8 @@ pool_svc_exclude_ranks(struct pool_svc *svc, struct pool_svc_event_set *event_se
 				 NULL, NULL, 0, &list, &inval_list_out, &map_version,
 				 NULL /* hint */, MUS_SWIM, false);
 
-	D_DEBUG(DB_MD, "Exclude pool "DF_UUID"/%u ranks %u: rc %d\n",
-		DP_UUID(svc->ps_uuid), map_version, n, rc);
+	D_DEBUG(DB_MD, DF_UUID ": exclude %u ranks: map_version=%u: " DF_RC "\n",
+		DP_UUID(svc->ps_uuid), n, rc == 0 ? map_version : 0, DP_RC(rc));
 
 	pool_target_addr_list_free(&inval_list_out);
 out:


src/pool/srv_pool.c
```diff
@@ -7033,7 +7033,7 @@ pool_svc_update_map_internal(struct pool_svc *svc, unsigned int opc, bool exclud
 							   inval_tgt_addrs);
 			if (rc != 0)
 				goto out_map;
-			if (inval_tgt_addrs->pta_number > 0) {
+			if (src == MUS_DMG && inval_tgt_addrs->pta_number > 0) {
```
liuxuezhao, line 7036: may leave a log if find invalid ranks for MUS_SWIM?

liuxuezhao (approved):

src/pool/srv_pool.c
liw: @liuxuezhao, is the following okay? If yes, I'll refresh this PR to include it.
```
			if (inval_tgt_addrs->pta_number > 0) {
				if (src == MUS_DMG) {
					/*
					 * If any invalid ranks/targets were specified here,
					 * abort the entire request. This will mean the
					 * operator needs to resubmit the request with
					 * corrected arguments, which will be easier without
					 * trying to figure out which arguments were accepted &
					 * started processing already.
					 */
					rc = -DER_NONEXIST;
					goto out_map;
				} else {
					int i;

					for (i = 0; i < inval_tgt_addrs->pta_number; i++)
						D_DEBUG(DB_MD,
							DF_UUID
							": ignore target: rank=%u target=%u\n",
							inval_tgt_addrs->pta_addrs[i].pta_rank,
							inval_tgt_addrs->pta_addrs[i].pta_target);
				}
			}
```

liuxuezhao: sure, LGTM, thanks.

liuxuezhao: maybe fine to leave it as it and change it in a following PR, to avoid test it again. just fyi

liw: Oh, okay, I'm almost about to push. Let's do this in another PR then. Thank you.

kccain: I guess it could be done either here, or in relevant caller, perhaps pool_svc_exclude_ranks() that currently doesn't seem to do much if inval_list_out has items?

kccain (approved):

src/pool/srv_pool.c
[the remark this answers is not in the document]
liw: OK.

liuxuezhao:
> https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-16081/1/execution/node/1557/log

maybe can belong to https://daosio.atlassian.net/browse/DAOS-16764

liw:
There were 3 PS leaders during this daos_rebuild_ec REBUILD46 run: rank 3 (term 1), rank 6 (term 2), and rank 7 term 3). The first leadership change was because the test killed rank 3, and the second the test killed rank 6 (and rank 0). I've checked the logs of rank 6 and 7, both of which seemed to have handled the crt events as expected. Hence, I'll assume that the test failure likely isn't due to this PR.

liuxuezhao:
> also, there was something similar in daos_test_ec.c, REBUILD45 (not REBUILD46 as in this case) in older ticket https://daosio.atlassian.net/browse/DAOS-16035

With some similar but should be different, 16035 was due to cont non-exist and returned data is all 0, this one return invalid non-zero data. May apply to DAOS-16764 first.

PR #16096, opened 2025-03-14
DAOS-17282 bio: clear io contexts for unplugged faulty device

NiuYawei:
When teardown an unplugged health device, we close blobs but keep the io contexts (for the up layer opened VOS pool), so that when the device is plugged back, blobs will be auto opened for these existing io contexts.

However, if the unplugged device is marked as FAULTY (after teardown completed), the upper layer VOS pools will be closed on faulty reaction, so we need to clear all the closed io contexts when setup device.

### Steps for the author:

### DIFF
--- a/src/bio/bio_monitor.c
+++ b/src/bio/bio_monitor.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2019-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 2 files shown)


src/bio/bio_monitor.c
```diff
@@ -735,16 +736,16 @@ auto_faulty_detect(struct bio_blobstore *bbs)
 	if (bbs->bb_state == BIO_BS_STATE_FAULTY)
```
tanabarr, line 736: I don't think you need this check now you have the one below (`!NORMAL && !OUT`) so this can be removed.

knard38:
The patch was validated at LRZ :-)

tanabarr (approved):

knard38 (approved):
Not sure to fully understand the patch.
However, I approve this PR as we were able to validate it on the LRZ cluster.

src/bio/bio_monitor.c
NiuYawei: Indeed, this could be removed now. I'll keep it in mind and cleanup it later.

PR #16117, opened 2025-03-18
DAOS-17278 build: Update golang.org/x/net to v0.37.0

mjmac:
Address a CVE. Requires an update to the minimum toolchain
version to 1.23.0. If the build toolchain is at least 1.21,
then a newer toolchain will automatically be downloaded.

Change-Id: I1331b764432e287eedc21898c49cb4086da0e6dc


jolivier23 (approved):

kjacque (commented):
I think this approach is the right one. We should port this to master, too, once the kinks are ironed out.

kjacque:
Pretty sure the RPM build failures are related to CI bugs that are patched for 2.6 with #16119 and #16125.

PR #16126, opened 2025-03-19
DAOS-17317 control: Quiet misleading server error

mjmac:
It's expected that querying the MS properties will
fail before the storage is formatted, so we shouldn't
log it as an error.


jolivier23 (approved):

tanabarr (approved):

kjacque (approved):
Was the logging just noisy?

mjmac:
> Was the logging just noisy?

Not just noisy, but misleading. When troubleshooting an engine start failure due to bad SPDK config or other issues, the repeated ERROR lines caused folks to look at the wrong thing for a while. It's not really an error that the MS isn't available yet, so it shouldn't be logged as such.

daltonbohning:
Copyright check failing only because the latest merge commit was not signed off.

PR #16136, opened 2025-03-23
DAOS-17323 common: do not need sgl bytes check for checksum

wangdi1:
client side checksum calculation is based on the request iod recxs, though the checksum is calculated by real IOM, so the input request bytes might be larger than the real fetched bytes in sgl, so let's ignore the size after the real fetched bytes.

### Steps for the author:

### DIFF
--- a/src/common/checksum.c
+++ b/src/common/checksum.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2019-2023 Intel Corporation.
+ * (C) Copyright 2025 Google LLC
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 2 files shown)


jolivier23 (approved):

src/common/checksum.c
```diff
@@ -612,18 +613,10 @@ calc_csum_recx_with_map(struct daos_csummer *obj, size_t *csum_nr,
 		}
 	}
 
-	if (consumed_bytes < recx->rx_nr * rec_len) {
-		/** Nothing mapped for recx or tail unmapped */
-		bytes_to_skip = (recx->rx_nr * rec_len) - consumed_bytes;
-		daos_sgl_processor(sgl, false, idx, bytes_to_skip,
-				   sgl_process_nop_cb, NULL);
```
liuxuezhao, line 619: you mean this daos_sgl_processor() possibly cause INFO log ""Requested more bytes than what's available in sgl"?
I am wondering how can it happen, for fetch, the code obj_iod_sgl_valid() -> obj_req_size_valid() already checked sgl's buffer is enough for iod, why here can find the sgl is not enough for the recx?

PR #16139, opened 2025-03-24
SRE-2832 build: unblock Build, Unit tests, Functional test (EL8) stages in FC Lab

grom72:
Set of changes required for FC Lab to enable RPM build, Unit Tests and Functional tests

All changes shall be harmless for RR Lab

### Steps for the author:


daltonbohning (approved):

JohnMalmberg (commented):
I want to test this in the Fort Collins lab before landing.

phender (dismissed):

Jenkinsfile
```diff
@@ -242,12 +242,21 @@ pipeline {
         booleanParam(name: 'CI_DEB_Ubuntu20_NOBUILD',
                      defaultValue: false,
                      description: 'Do not build DEB packages for Ubuntu 20')
+        booleanParam(name: 'CI_el8_BUILD',
```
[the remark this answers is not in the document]
grom72: I will clean up all params as soon as we have new lab infrastructure up and running - now it is just for easy customization manual build in FC.
The plan is to harmonize but also to remove parameters handling from skipStage() function.

Jenkinsfile
```diff
@@ -1098,7 +1117,6 @@ pipeline {
                         'Functional Hardware Medium': getFunctionalTestStage(
                             name: 'Functional Hardware Medium',
                             pragma_suffix: '-hw-medium',
-                            base_branch: 'master',
```
daltonbohning, line 1101: Why is this not needed anymore?

ftest.sh
```diff
@@ -56,8 +62,23 @@ cleanup() {
...
+        echo "All cleanup attempts failed." >&2
+        return 1
+    fi
+    return 0
 }
 
+ls ~/.ssh
+touch ~/.ssh/known_hosts
+chmod 600 ~/.ssh/known_hosts
+for host in "$(IFS=','; echo "${nodes[*]}")"; do
```
daltonbohning, line 75: shellcheck is failing

utils/githooks/pre-commit.d/30-Jenkinsfile.sh
```diff
@@ -20,6 +20,9 @@ if [ -z "$(_git_diff_cached_files "Jenkinsfile")" ] ; then
 fi
 
 : "${JENKINS_HOST:=build.hpdd.intel.com}"
+echo "Temporary no access to $JENKINS_HOST. Skipping"
```
daltonbohning, line 23: shellcheck is failing. But I think it only complains on files modified so maybe won't be an issue after merge

Jenkinsfile
JohnMalmberg: It only should be specified for daily/weekly branches to specify what branch to use.

If you specify it on master or release branches it should be a noop.

If you specify it on a PR, then it is not running test code in the PR as intended and you end up landing test code that has never been actually tested in a PR.

ftest.sh
JohnMalmberg: That code should have already been removed in one of the updates.   Apparently at some point in this PR a force push was done.

utils/githooks/pre-commit.d/30-Jenkinsfile.sh
JohnMalmberg: I think shell check is just complaining that some code is now unreachable.

daltonbohning: Right. But I'm not 100% sure whether shellcheck only runs against modified files or all shell files. So if we merge this, I'm not sure if other PRs will start seeing a failure in shellcheck

daltonbohning: Unless I'm missing something, it seems it runs against all shell files so would fail on PRs after merge

ftest.sh
JohnMalmberg: If this code is still actually present, it needs to be removed.

utils/githooks/pre-commit.d/30-Jenkinsfile.sh
JohnMalmberg: A simple suppression line should fix it.
CodeSpell and Shellcheck are now available as a plugin for Visual Studio Code so we should be catching these things earlier than in PRs.

Jenkinsfile
JohnMalmberg: That parameter swaps out everything checked out for the PR except the Jenkinsfile for the code in the master branch.

Any changes to any files in the repository that affect testing in the PR will actually just be using the master branch instead of the new code.

This was exposed in this PR because we have bugfixes in this PR to prevent CI from failing, and until this change is landed, all PRs being run are not actually running the test code that the PR author intended to be run.

JohnMalmberg (changes requested):
Potential of some uneeded code that is causing shellcheck errors may be present.

Code change to cause dead code needs a shellcheck suppression.

Jenkinsfile
daltonbohning: I don't doubt there is an issue here but
> Any changes to any files in the repository that affect testing in the PR

I don't think is correct. Otherwise we wouldn't be able to validate test fixes in PRs, which we do all the time. I can modify some files and my changes are in fact picked up in the functional test stages.

JohnMalmberg: Some test code is compiled and stored in DAOS RPMs, that code gets tests.
None of the code for doing CI provisioning that may have changed in the PR will be used in the PR.

ftest.sh
grom72: FIxed in 8a02226cb0c6403359067302488741deb3a7f52b
and removed in d929ddc18e80e16d84c3fb008137ed52a7242ad6

Jenkinsfile
```diff
@@ -17,6 +17,7 @@
 // To use a test branch (i.e. PR) until it lands to master
 // I.e. for testing library changes
 //@Library(value='pipeline-lib@your_branch') _
+@Library(value='pipeline-lib@sre-2932') _
```
daltonbohning, line 20: So it's intentional to merge with this?

daltonbohning (dismissed):

Jenkinsfile
JohnMalmberg: That also needs to be removed.

JohnMalmberg (changes requested):
pipeline-lib reference needs to be removed.

We should probably land the pipeline-lib pr with the debug messages removed before we land this one.

Jenkinsfile
daltonbohning: Maybe I misunderstood this
> DO NOT REMOVE BRANCH AFTER MERGING AS IT IS NEEDED FOR FC FURTHER VALIDATION

@grom72 do you mean to not remove this branch `grom72/sre-2832-fix-scratch-mount`?

JohnMalmberg (approved):

Jenkinsfile
JohnMalmberg: The FC Jenkins defaults to "@Library(value='pipeline-lib@sre-2932') _" if it not specified.
This fixes a minor issue about properly running groovy code in a sandbox and is what is needed to use any Jenkins other than in Rio Rancho.

The restriction is that pipeline-lib sre-2932 branch cannot be deleted when it is landed, or jobs will stop working on the FC Jenkins until that default is changed to normal.

phender (approved):

daltonbohning (approved):

PR #16144, opened 2025-03-24
b/406041958 control: Add configurable engine delay

mjmac:
When engine_start_delay is set to a non-zero value in
the daos_server yaml, delay engine start by that value
in seconds. Potentially useful in preventing use of PD
devices before they are ready.

Change-Id: I998aaf814114d6cb7383f936df6fa56a9cb33c34

### DIFF
--- a/src/control/server/config/server.go
+++ b/src/control/server/config/server.go
@@ -1,5 +1,6 @@
 //
 // (C) Copyright 2020-2024 Intel Corporation.
+// (C) Copyright 2025 Google LLC
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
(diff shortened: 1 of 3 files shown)


jolivier23 (approved):

jolivier23:
would be ideal if we could figure out a way to query the devices for readiness

mjmac:
> would be ideal if we could figure out a way to query the devices for readiness

Agreed, but this will allow us to work around it in the short term while we look for a better solution. It seems like a bit of a chicken/egg problem, in that if you try to access the device too soon in order to determine if it's ready, it gets into a bad state. At least, that's the current understanding.

PR #15668, opened 2024-12-28
DAOS-16630 control: improve dmg pool exclude

wangshilong:
When invoking “dmg pool exclude”, an administrator may not be aware of whether this will break the pool-wise RF or not. If this breaks the pool-wise RF, there may be data loss. Hence, we should perform the pool-wise RF check (i.e., the pw_rf check in pool_svc_update_map_internal) before changing the pool map.

On the other hand, the administrator may also want to exclude rank/targets even if there may be data loss. We should introduce a new, explicit option for the administrator to say that she’s aware of the danger and would like to bypass the pool-wise RF check.

### Before requesting gatekeeper:

* [ ] Two review approvals and any prior change requests have been resolved.
* [ ] Testing is complete and all tests passed or there is a reason documented in the PR why it should be force landed and forced-landing tag is set.
* [ ] `Features:` (or `Test-tag*`) commit pragma was used or there is a reason documented that there are no appropriate tags for this PR.
* [ ] Commit messages follows the guidelines outlined [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Any tests skipped by the ticket being addressed have been run and passed in the PR.

### Gatekeeper:

* [ ] You are the appropriate gatekeeper to be landing the patch.
* [ ] The PR has 2 reviews by people familiar with the code, including appropriate owners.
* [ ] Githooks were used. If not, request that user install them and check copyright dates.
* [ ] Checkpatch issues are resolved.  Pay particular attention to ones that will show up on future PRs.
* [ ] All builds have passed.  Check non-required builds for any new compiler warnings.
* [ ] Sufficient testing is done. Check feature pragmas and test tags and that tests skipped for the ticket are run and now pass with the changes.
* [ ] If applicable, the PR has addressed any potential version compatibility issues.
* [ ] Check the target branch.   If it is master branch, should the PR go to a feature branch?  If it is a release branch, does it have merge approval in the JIRA ticket.
* [ ] Extra checks if forced landing is requested
  * [ ] Review comments are sufficiently resolved, particularly by prior reviewers that requested changes.
  * [ ] No new NLT or valgrind warnings.  Check the classic view.
  * [ ] Quick-build or Quick-functional is not used.
* [ ] Fix the commit message upon landing. Check the standard [here](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments). Edit it to create a single commit. If necessary, ask submitter for a new summary.

PR #16164, opened 2025-03-27
DAOS-16630 control: improve dmg pool exclude (#15668)

wangshilong:
When invoking “dmg pool exclude”, an administrator may not be aware of whether this will break the pool-wise RF or not. If this breaks the pool-wise RF, there may be data loss. Hence, we should perform the pool-wise RF check (i.e., the pw_rf check in pool_svc_update_map_internal) before changing the pool map.

On the other hand, the administrator may also want to exclude rank/targets even if there may be data loss. We should introduce a new, explicit option for the administrator to say that she’s aware of the danger and would like to bypass the pool-wise RF check.

Test-tag: pr DmgPoolExcludeRanks

### Steps for the author:


daltonbohning (commented):
ftest LGTM

wangshilong:
> Just to note that I have a backport version of this PR in #16200 stack.

I closed this.

PR #16173, opened 2025-03-31
DAOS-17347 dtx: retry DTX resync for recoverable network failure

Nasf-Fan:
If the engine on which some DTX leader reside restart, then former in-DRAM committable DTX status will be lost. When (re)open related container, DTX resync will be triggered to resync and commit those former committable DTX entries. But during DTX resync, it may hit recoverable netwrok failure, such as "-DER_OOG" since lower layer network membership may be not ready yet. Under such case, wen need to retry resync for related DTX; otherwise related DTX will become non-committable status globally as to misguide subsequent IO logic to give wrong visibility decision.

On the other hand, if some non-leader triggers DTX refresh before DTX resync completed on the leader, it will get "-DER_INPROGRESS" and needs to retry.

### Steps for the author:


Nasf-Fan:
> Test stage Functional Hardware Large completed with status FAILURE. https://build.hpdd.intel.com//job/daos-stack/job/daos/view/change-requests/job/PR-16173/12/execution/node/1530/log

test_ec_online_rebuild_mdtest failed for DAOS-17161, not related with the patch.

src/dtx/dtx_internal.h
```diff
@@ -57,12 +57,16 @@ enum dtx_operation {
 #undef X
 
 /* DTX RPC input fields */
+/* clang-format off */
 #define DAOS_ISEQ_DTX							\
 	((uuid_t)		(di_po_uuid)		CRT_VAR)	\
 	((uuid_t)		(di_co_uuid)		CRT_VAR)	\
 	((uint64_t)		(di_epoch)		CRT_VAR)	\
+	((uint32_t)		(di_version)		CRT_VAR)	\
+	((uint32_t)		(di_padding)		CRT_VAR)	\
```
wangdi1, line 66: can you put di_version after di_flags, then you do not need padding?

btw: is this server-server RPC? does it need rpc version change?

src/dtx/dtx_resync.c
```diff
@@ -245,6 +245,13 @@ dtx_status_handle_one(struct ds_cont_child *cont, struct dtx_entry *dte, daos_un
 		 * let's commit the DTX globally.
 		 */
 		D_GOTO(out, rc = DSHR_NEED_COMMIT);
+	case -DER_OOG:
+	case -DER_HG:
+		D_WARN("Need retry resync for DTX " DF_DTI " because of " DF_RC "\n",
+		       DP_DTI(&dte->dte_xid), DP_RC(rc));
+		/* Yield to give more chance for network recovery. */
+		ABT_thread_yield();
```
wangdi1, line 253: please use dss_sleep() for dss scheduler purpose.

wangdi1 (dismissed):
Do you think we should add an unit test case for this?

src/dtx/dtx_resync.c
```diff
@@ -245,6 +245,13 @@ dtx_status_handle_one(struct ds_cont_child *cont, struct dtx_entry *dte, daos_un
 		 * let's commit the DTX globally.
 		 */
 		D_GOTO(out, rc = DSHR_NEED_COMMIT);
+	case -DER_OOG:
```
liuxuezhao, line 248: I am not quite sure if DER_OOG can always retry (for example when the target is not in Primary group will return DER_OOG, will it cause dead loop for some cases if always retry?), is there a chance to refresh pool map?

liuxuezhao (dismissed):
Please confirm if any problem for the commented case, thx

src/dtx/dtx_internal.h
Nasf-Fan: > can you put di_version after di_flags, then you do not need padding?
di_flags is array.


> btw: is this server-server RPC? does it need rpc version change?
It is server side RPC, no need version change since we do not support interoperability among different versions of servers.

src/dtx/dtx_resync.c
Nasf-Fan: > please use dss_sleep() for dss scheduler purpose.

DTX resync may block the other IO process, it is better to make it to be scheduled ASAP.

src/dtx/dtx_resync.c
Nasf-Fan: If the rank/target is not excluded, then since there is yield() for each cycle retry, then it will not always OOG as long as the group membership can be initialized finally; otherwise if it has already been excluded, then seems no matter even if the excluded one fall into dead loop?

Nasf-Fan:
Ping @wangdi1 @liuxuezhao , please help to review the patch again. Thanks!

wangdi1 (approved):

liuxuezhao (approved):

PR #16182, opened 2025-04-02
DAOS-17359 dfuse: Add special file for log mask

jolivier23:
It may be useful to interact with dfuse at runtime to perform certain tasks. This adds a generic
command parser capability and two initial commands as a reference.

### log command

Changing the log mask at runtime can be useful to debug certain runtime behaviors without
overwhelming the default log.  The log command allows us to do this.

```shell
cat << EOF > /mnt/dfuse/.dfuse_ctrl
log
log_mask=debug
streams=all
EOF
```
To see any errors from the command (or the output), cat the file
```shell
cat /mnt/dfuse/.dfuse_ctrl
```

### stats command

```shell
echo "stats" > /mnt/dfuse/.dfuse_ctrl; cat /mnt/dfuse/.dfuse_ctrl
```

This will produce the same output as `daos fs query /mnt/dfuse`

The daos command requires the client to be installed and this was not always the case when dfuse is mounted in a sidecar container.

Signed-off-by: Jeff Olivier <jeffolivier@google.com>

### Steps for the author:


jolivier23:
@mchaarawi any thoughts on the general concept here?

mchaarawi:
> @mchaarawi any thoughts on the general concept here?

not opposed to it. my question though is what is better, this approach or just add an ioctl and update a daos tool to pass the logging needed?
the latter does not need all those special handling of the ctrl file.

jolivier23:
> > @mchaarawi any thoughts on the general concept here?
> 
> not opposed to it. my question though is what is better, this approach or just add an ioctl and update a daos tool to pass the logging needed? the latter does not need all those special handling of the ctrl file.

The problem is that it's not always convenient to get the daos tool.   Most of our GKE customers just have a mount point and no control over it or daos client installed.

Though maybe we could provide some means that is relatively simple (I did previously write a python script to do an ioctl)

mchaarawi:
> > > @mchaarawi any thoughts on the general concept here?
> > 
> > 
> > not opposed to it. my question though is what is better, this approach or just add an ioctl and update a daos tool to pass the logging needed? the latter does not need all those special handling of the ctrl file.
> 
> The problem is that it's not always convenient to get the daos tool. Most of our GKE customers just have a mount point and no control over it or daos client installed.
> 
> Though maybe we could provide some means that is relatively simple (I did previously write a python script to do an ioctl)

not insisting on that really, just thought it would be cleaner?
but if there is a usage issue with that, then this is fine too.

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16182/8/testReport/

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16182/12/testReport/

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16182/13/testReport/

PR #16184, opened 2025-04-02
DAOS-17353 object: reduce noisy warning for DTX refresh

Nasf-Fan:
When process IO on DTX non-leader target, if hit prepared DTX, related retry logic may generate warning message for reminder before refresh related DTX. That is not fatal but some noisy.

This patch controls such warning message rate to avoid flood in log: one warning message per about 10 seconds.

### Steps for the author:

### DIFF
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -1,6 +1,7 @@
 /**
  * (C) Copyright 2016-2024 Intel Corporation.
  * (C) Copyright 2025 Google LLC
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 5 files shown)


liuxuezhao (dismissed):

mchaarawi (dismissed):
this look good to me. my only question is if we really just want to make those debug rather than just reduce frequency?

Nasf-Fan:
> this look good to me. my only question is if we really just want to make those debug rather than just reduce frequency?

We need this warning message since for some users, they may not want to enable debug in product environment, then these warning message will give us some hint if related application hung there.

mchaarawi:
please fix clang format, it is required for landing.
I can force land after you fix it without needing to wait for CI results. thanks

mchaarawi (approved):

PR #16192, opened 2025-04-03
DAOS-17361 vos: shrink DAOS_VOS_AGG_GAP default value

Nasf-Fan:
Change DAOS_VOS_AGG_GAP default value from 60 (seconds) to 25, then VOS aggregation can reclaim space more in time to reduce backend space pressure.

### Steps for the author:

### DIFF
--- a/src/vos/vos_internal.h
+++ b/src/vos/vos_internal.h
@@ -145,7 +145,7 @@ enum {
 extern uint32_t vos_agg_gap;
 
 #define VOS_AGG_GAP_MIN		20 /* seconds */
-#define VOS_AGG_GAP_DEF		60
+#define VOS_AGG_GAP_DEF         25
 #define VOS_AGG_GAP_MAX		180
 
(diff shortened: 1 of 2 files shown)


src/vos/vos_internal.h
```diff
@@ -145,7 +145,7 @@ enum {
 extern uint32_t vos_agg_gap;
 
 #define VOS_AGG_GAP_MIN		20 /* seconds */
-#define VOS_AGG_GAP_DEF		60
+#define VOS_AGG_GAP_DEF         25
```
NiuYawei, line 148: Can we change it back to original 20 seconds (or even smaller)? Even with 20 seconds (2.6), we've already hit false ENOSPACE in large I/O workload (for the vos pool with decent size).

I know that restart/resend could burden the engine further, but I think we can fix that by adjusting restart/resend logic, right?

Nasf-Fan:
Niu will handle related issue in another PR together with false enospace, close this one.

PR #16201, opened 2025-04-05
DAOS-17343 control: Validate SPDK config on engine start

mjmac:
Before starting an engine, first check to see if the on-disk
SPDK configuration file can be read and parsed by this
version of DAOS. If it's not found or can't be parsed, take
it as a signal to regenerate the file.

Change-Id: I3ccbd4c5ce036435c45f185f902f0aa6a679beb3

### DIFF
--- a/src/control/server/storage/bdev/backend.go
+++ b/src/control/server/storage/bdev/backend.go
@@ -1,5 +1,6 @@
 //
 // (C) Copyright 2019-2023 Intel Corporation.
+// (C) Copyright 2025 Google LLC
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -543,6 +544,29 @@ func (sb *spdkBackend) writeNvmeConfig(req storage.BdevWriteConfigRequest, confW
 
 func (sb *spdkBackend) WriteConfig(req storage.BdevWriteConfigRequest) (*storage.BdevWriteConfigResponse, error) {
 	return &storage.BdevWriteConfigResponse{}, sb.writeNvmeConfig(req, writeJsonConfig)
+}
+
+func (sb *spdkBackend) ReadConfig(req storage.BdevReadConfigRequest) (*storage.BdevReadConfigResponse, error) {
+	if req.ConfigPath == "" {
+		return nil, errors.New("empty SPDK config path")
+	}
+
+	r, err := os.Open(req.ConfigPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open SPDK config at %q", req.ConfigPath)
+	}
+	defer r.Close()
+
+	_, err = readSpdkConfig(r)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Reconstruct the WriteConfig request params? At the moment, all we care about is
+	// that we can read and parse the config file, but in the future it might be useful to
+	// be able to inspect its contents.
+	resp := &storage.BdevReadConfigResponse{}
+	return resp, nil
 }
 
 // UpdateFirmware uses the SPDK bindings to update an NVMe controller's firmware.
--- a/src/control/server/storage/bdev/backend.go
+++ b/src/control/server/storage/bdev/backend.go
@@ -542,10 +542,15 @@ func (sb *spdkBackend) writeNvmeConfig(req storage.BdevWriteConfigRequest, confW
 	return errors.Wrap(confWriter(sb.log, &req), "write spdk nvme config")
 }
 
+// WriteConfig writes the SPDK configuration file.
 func (sb *spdkBackend) WriteConfig(req storage.BdevWriteConfigRequest) (*storage.BdevWriteConfigResponse, error) {
 	return &storage.BdevWriteConfigResponse{}, sb.writeNvmeConfig(req, writeJsonConfig)
 }
 
+// ReadConfig reads the SPDK configuration file.
+// NB: Currently returns an empty response struct if the file is read
+// and parsed successfully, but may be extended to return the
+// parsed configuration.
 func (sb *spdkBackend) ReadConfig(req storage.BdevReadConfigRequest) (*storage.BdevReadConfigResponse, error) {
 	if req.ConfigPath == "" {
 		return nil, errors.New("empty SPDK config path")
--- a/src/control/server/storage/bdev/backend_class_test.go
+++ b/src/control/server/storage/bdev/backend_class_test.go
@@ -10,7 +10,6 @@ package bdev
 import (
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 
 	"github.com/dustin/go-humanize"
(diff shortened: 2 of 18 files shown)


src/control/server/storage/provider.go
```diff
@@ -666,6 +678,33 @@ func (p *Provider) UpdateBdevFirmware(req NVMeFirmwareUpdateRequest) (*NVMeFirmw
 	return p.bdev.UpdateFirmware(req)
 }
 
+// ValidateBdevConfig updates an existing SPDK bdev config, if necessary.
+func (p *Provider) ValidateBdevConfig(ctx context.Context, ctrlrs NvmeControllers) error {
```
tanabarr, line 682: this name looks misleading, validate doesn't intuitively imply an update

src/control/server/storage/provider.go
```diff
@@ -666,6 +678,33 @@ func (p *Provider) UpdateBdevFirmware(req NVMeFirmwareUpdateRequest) (*NVMeFirmw
...
+	if !fault.IsFaultCode(err, code.SpdkInvalidConfiguration) {
+		p.log.Errorf("Failed to read bdev config file: %s", p.engineStorage.ConfigOutputPath)
+		return err
+	}
+
+	// Otherwise, if the config file is there but we can't understand it, then
+	// it's probably from a different version of DAOS and we should just regenerate
+	// it based on our version of the configuration.
+	p.log.Notice("The bdev config file was unparsable; regenerating it.")
+	return p.WriteNvmeConfig(ctx, p.log, ctrlrs)
```
tanabarr, line 705: I'm unsure as to whether this is a safe approach, it seems like this scenario indicates something is wrong and maybe it is better to fail here. Format should be where all the checks and balances are accounted for, allowing ad hoc rewriting of config file seems risky. @kjacque thoughts?

src/control/server/storage/provider.go
mjmac: What do you suggest as an alternative verb? "Verify"? "UpdateIfNecessary" seems verbose.

tanabarr: CheckOrUpdate?

src/control/server/storage/provider.go
tanabarr: not necessarily, I'm more suggesting that rewriting the config might be hiding significant problems

tanabarr: in what situation would we want to rewrite the config, if metadata directory contains corrupt files don't we want to refuse to start?

tanabarr: Thanks for the extra context. In terms of upgrade path from 2.6->2.8 if we are supporting storage compatibility without reformat then this does sound reasonable as an approach.

tanabarr (approved):

src/control/server/storage/provider.go
kjacque: I think there is some value in checking whether the file is valid for spdk v22 before upgrading it. I.e. detecting the difference between an old file and a corrupt file. To be honest I think it is always safe to replace, but if we detect corruption, that would be potentially useful to log, to alert the admin.

src/control/server/storage/bdev/backend_class_test.go
```diff
@@ -120,60 +145,35 @@ func TestBackend_writeJSONFile(t *testing.T) {
 					DeviceList: storage.MustNewBdevDeviceList(test.MockPCIAddrs(1)...),
 				},
 			}).WithStorageEnableHotplug(true),
-			expOut: `
-{
-  "daos_data": {
-    "config": [
-      {
-        "params": {
-          "begin": 0,
```
kjacque, line 129: Is there a test verifying the JSON parameters? The lack of that is the only downside I could see to checking the format to the config. Checking the JSON output as a string (even for just a single happy path case) is a good sanity check for our structs' json fields being defined properly when, for example, we change the name of an underlying field like this.

src/control/server/storage/bdev/backend_test.go
```diff
@@ -1332,3 +1334,75 @@ func TestBackend_Prepare(t *testing.T) {
...
+			req:    storage.BdevReadConfigRequest{},
+			expErr: errors.New("empty SPDK config path"),
+		},
+		"bad config path": {
+			req: storage.BdevReadConfigRequest{
+				ConfigPath: "/bad/path",
+			},
+			expErr: errors.New("open /bad/path: no such file or directory"),
+		},
+		"good config path; bad config": {
```
kjacque, line 1355: how about a case with a well-formed config with bad fields?

src/control/server/storage/faults.go
```diff
@@ -265,6 +266,15 @@ func FaultPathAccessDenied(path string) *fault.Fault {
 	)
 }
 
+// FaultInvalidSPDKConfig creates a Fault for the case where SPDK configuration is invalid.
+func FaultInvalidSPDKConfig(err error) *fault.Fault {
+	return storageFault(
+		code.SpdkInvalidConfiguration,
+		fmt.Sprintf("unable to parse SPDK configuration: %s", err),
+		"regenerate the configuration and restart daos_server",
```
kjacque, line 274: suggestion - add some detail on how to do that or where in the admin guide to find this information

src/control/server/storage/provider.go
kjacque: Is there anything wrong with "UpdateBdevConfig"? If no update is needed, it cheerfully returns nil and moves on.

src/control/server/storage/bdev/backend_class_test.go
mjmac: I removed the JSON strings in favor of comparing the configuration structs as a way to test `ReadConfig()` and `UnmarshalJSON()`, and also to make these tests a bit more maintainable. Having to repeat the raw JSON strings constantly seems pretty toilsome, and at some point it seems like we're just testing the json package.

I do see your point that it might be useful to have a single test with all of the options so that we can see if some change results in output that wouldn't be parseable by the engine? That having been said, if we update SPDK and there are different config parameters in that new version, we're going to be broken here either way.

src/control/server/storage/bdev/backend_test.go
mjmac: Sure, I can add one just for completeness, although again, this seems like we're testing the json package.

src/control/server/storage/faults.go
mjmac: I don't expect this to be surfaced to the admin unless something is really screwed up. It has to be a fault so that it transits the daos_server <-> daos_server_helper boundary cleanly. I initially started with a simple sentinel error, but in that case the server would only see a stringified representation of the error.

In any case, I'm not sure what we could document here... There isn't a manual procedure for regenerating the config aside from storage format, so IMO if this fault somehow surfaces, it means something has gone wrong in the code and it's a support engagement. I'd prefer to keep the resolution string as-is because it's generic enough to not need updating later but specific enough to direct support toward a potential resolution.

src/control/server/storage/provider.go
mjmac: Sure, if that works for y'all I'll make that change.

src/control/server/storage/provider.go
mjmac: Hmm. As I looked into this, I found that there didn't seem to be a straightforward way to make this happen. There are layering concerns with either:
  * Pulling `bdev.SpdkConfig` and friends up into the storage layer so that we can get rid of the intermediate representation in `storage.BdevWriteConfigRequest` (and also more easily return the parsed configuration from `spdkConfig.ReadConfig()`)
  * Pulling the config validation logic down from `storage.Provider.UpgradeBdevConfig()`

Encapsulating all of the spdk config stuff in the `bdev` layer does seem to make sense. I was thinking that we might want more control over it in daos_server vs daos_server_helper, but maybe it doesn't matter.

src/control/server/storage/bdev/backend.go
```diff
@@ -545,6 +546,29 @@ func (sb *spdkBackend) WriteConfig(req storage.BdevWriteConfigRequest) (*storage
 	return &storage.BdevWriteConfigResponse{}, sb.writeNvmeConfig(req, writeJsonConfig)
 }
 
+func (sb *spdkBackend) ReadConfig(req storage.BdevReadConfigRequest) (*storage.BdevReadConfigResponse, error) {
```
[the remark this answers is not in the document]
kjacque: As a partial implementation that could be extended, but probably won't be in the near future, the name is misleading IMO, and may lead to confusion down the road. That said, as long as it is commented in the docstring, I will live with it.

src/control/server/storage/bdev/backend_class_test.go
kjacque: I'm fine with the general approach for most of the tests. But yeah I think it would be useful to have one unit test that verifies we didn't change the JSON structs unless we meant to. It's significantly easier to debug a Go unit test than a live system with crashing engines.

src/control/server/storage/faults.go
kjacque: Okay, that makes sense. I think my brain saw that it was a Fault and assumed it must become user visible at some point. Considering we always catch it, the vagueness is fine.

kjacque (approved):
Overall looks good. I think there would be some value in one test verifying the JSON output--not that it's valid JSON, but that we can interface with the expected SPDK API. I won't block on it but I do think a single test like that is low overhead, and changes to the test would only be triggered by changes we make to match the SPDK API.

daltonbohning:
Need one more review before gatekeeper?

mjmac:
> Need one more review before gatekeeper?

Oops, lost @tanabarr's review when I pushed an update.

tanabarr (approved):

mjmac:
@daltonbohning: OK, have two +1s now. Sorry for rushing that.

PR #16204, opened 2025-04-07
DAOS-17286 gurt: fix rank set string convertion

knard38:
Remove string maximal length limitation of the `d_rank_range_list_str()` convertion function.

### Steps for the author:


tanabarr (approved):
looks good , should there be an explicit test case added for the trailing commma? maybe I'm just missing it

knard38:
> looks good , should there be an explicit test case added for the trailing commma? maybe I'm just missing it

With this patch, trailing comma issue will not occur any more: It was occurring because the input list of ranks was too big to be serialized with the fixed output buffer, and thus some ranks were missing and the string list was containing a trailing comma.

Does it make sense ?

src/gurt/misc.c
```diff
@@ -1,5 +1,6 @@
 /*
  * (C) Copyright 2016-2024 Intel Corporation.
```
daltonbohning, line 2: Copyright check is complaining because commit is signed with intel
> Signed-off-by: Cedric Koch-Hofer <cedric.koch-hofer@intel.com>

daltonbohning (approved):

src/gurt/misc.c
knard38: Arghhh, I was not understanding why it was complaining as the new HPe copyright was added.
I am going to fix this asap.
Thanks for the info.

knard38:
@daos-stack/daos-gatekeeper , please could you lend this PR with the following message:
- title: `DAOS-17286 gurt: fix rank set string convertion`
- body:
```
Remove string maximal length limitation of the d_rank_range_list_str() convertion function.
```

PR #16228, opened 2025-04-10
DAOS-17229 cart: Add env var SWIM_SUBGROUP_SIZE

liw:
Make the number of iping targets, called the subgroup size, tunable by new environment variable SWIM_SUBGROUP_SIZE.

Log new SWIM suspicions at INFO level, including their origins. This may help debugging why a DEAD event happens.

### Steps for the author:

### DIFF
--- a/src/cart/swim/swim.c
+++ b/src/cart/swim/swim.c
@@ -2,6 +2,7 @@
  * Copyright (c) 2016 UChicago Argonne, LLC
  * (C) Copyright 2018-2024 Intel Corporation.
  * (C) Copyright 2025 Google LLC
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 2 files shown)


src/cart/swim/swim.c
```diff
@@ -48,6 +50,15 @@ swim_ping_timeout_default(void)
 	return val;
 }
 
+static inline int
+swim_subgroup_size_default(void)
+{
+	unsigned int val = SWIM_SUBGROUP_SIZE;
+
+	d_getenv_uint("SWIM_SUBGROUP_SIZE", &val);
```
frostedcmos, line 58: please add new env to CRT_ENV_LIST so that it gets dumped at the startup at info level. ideally we would change this (and other swim envs) to use crt_env_get() instead, but that would require include of crt_internal_types.h

frostedcmos (commented):
lgtm, minor note inline

src/cart/swim/swim.c
liw: Sure, added to `CRT_ENV_LIST`. Sounds like you don't want swim.c to include crt_internal_types.h, so I'm not changing to use `crt_env_get`.

frostedcmos (approved):

liuxuezhao (approved):

PR #16229, opened 2025-04-10
DAOS-17152 pool: seek to avoid false ENOSPACE

NiuYawei:
Seek to avoid false ENOSPACE by few adjustments:

- Revert the 60 seconds VOS_AGG_GAP_DEF back to 20 seconds.

- Aggregation ULTs don't voluntarily sleep when there is space pressure, otherwise, scheduler will think that there is nothing to be reclaimed and gives up on throttling I/O.

- Increase the reclaim ULTs CPU ratio for all pressure levels: bump the original [20, 30, 40, 50, 60] to [30, 45, 60, 75, 90].

- Shrink VEA flush/expire interval from 5/10 seconds to 2/3 seconds.

- Shrink the flush ULT sleep interval from 5/1 seconds to 500/50 milli-seconds.

### Steps for the author:

### DIFF
--- a/src/container/srv_target.c
+++ b/src/container/srv_target.c
@@ -471,8 +471,7 @@ cont_aggregate_interval(struct ds_cont_child *cont, cont_aggregate_cb_t cb,
 		/*
 		 * Sleep 2 seconds before next round when:
 		 * - Aggregation isn't runnable yet, or;
-		 * - Last round of aggregation failed, or;
-		 * - There is no space pressure;
+		 * - Last round of aggregation failed;
 		 */
 		uint64_t msecs = 2000;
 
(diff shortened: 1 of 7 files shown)


NiuYawei:
> > Seek to avoid false ENOSPACE by few adjustments:
> > 
> > * Revert the 60 seconds VOS_AGG_GAP_DEF back to 20 seconds.
> > * Aggregation ULTs don't voluntarily sleep when there is space pressure, otherwise, scheduler will think that there is nothing to be reclaimed and gives up on throttling I/O.
> > * Increase the reclaim ULTs CPU ratio for all pressure levels: bump the original [20, 30, 40, 50, 60] to [30, 45, 60, 75, 90].
> 
> This adjustment looks a bit aggressive compared to before, I am wondering in Jeff's scripts pool is small. do we really need so aggressive for a large pool size? or we add pool space into consideration?
> 
> > * Shrink VEA flush/expire interval from 5/10 seconds to 2/3 seconds.
> > * Shrink the flush ULT sleep interval from 5/1 seconds to 500/50 milli-seconds.

The changes to CPU ratio won't impact performance when there is no space pressure, and it's helpful in my local test: My test is FIO write in 256k IO size, each VOS pool has 20GB NVMe (data blob) size, each round of FIO write around 2GB per target and the per-target QD is 40).

jolivier23 (approved):

wangshilong (approved):

PR #16237, opened 2025-04-11
DAOS-17403 telem: Add non-shared memory flag

mjmac:
In cases where the client telemetry will not be
sampled via the agent (e.g. when dumping to a
CSV at process exit), it would be preferable to
avoid overhead and exposure to resource limits
imposed by the use of shared memory segments.

Setting the D_TM_NO_SHMEM flag in d_tm_init()
will result in process-local memory allocations
for the telemetry segments.

Change-Id: I60f8640aec5bc856c43b83c3bd691ae07276d457

### DIFF
--- a/src/gurt/telemetry.c
+++ b/src/gurt/telemetry.c
@@ -25,15 +25,15 @@
 #include <gurt/telemetry_producer.h>
 #include <gurt/telemetry_consumer.h>
 
-/** minimal list of shared memory regions with a global ID */
-struct shmem_region_list {
+/** minimal list of memory regions with a global ID */
+struct mem_region_list {
 	struct d_tm_node_t	*rl_link_node;
 	key_t			 rl_key;
 	d_list_t		 rl_link;
 };
 
-/** Header of a shared memory region */
-struct d_tm_shmem_hdr {
+/** Header of a memory region */
+struct d_tm_mem_hdr {
 	uint64_t		 sh_base_addr;	/** address of this struct */
 	key_t			 sh_key;	/** key to access region */
 	uint32_t                 sh_deleted : 1, /** marked for deletion */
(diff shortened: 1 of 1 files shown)


mjmac:
@kjacque: Wondering how much you hate this... :) We talked about it a while ago, and I decided to just make a proof-of-concept to see how it works. Functionally, this patch does work (I'm using it with https://github.com/daos-stack/daos/pull/16238), but it could use further refactoring and renaming to be clearer about shared vs non-shared memory management.

src/gurt/telemetry.c
```diff
@@ -81,13 +82,16 @@ static struct d_tm_shmem {
...
 	int			 id; /** Instance ID */
 } tm_shmem;
 
 /* Internal helper functions */
-static int allocate_shared_memory(int srv_idx, size_t mem_size,
-				  struct d_tm_shmem_hdr **shmem);
+static int
+allocate_memory_segment(int srv_idx, size_t mem_size, bool shared, struct d_tm_shmem_hdr **shmem);
+static void
+			   release_ctx_mem(struct d_tm_context **ctx, bool destroy_shmem);
```
kjacque, line 94: guessing clang-format did this... I know it can do some funky indentation, but any idea why the first function isn't indented but the second is?

src/gurt/telemetry.c
```diff
@@ -254,15 +258,37 @@ attach_shmem(key_t key, size_t size, int flags, struct d_tm_shmem_hdr **shmem)
 }
 
 static int
-new_shmem(key_t key, size_t size, struct d_tm_shmem_hdr **shmem)
+alloc_nsmem(size_t size, struct d_tm_shmem_hdr **shmem)
```
kjacque, line 261: tbh I'd rather have either `alloc_mem` or `alloc_non_shared_mem`. I can tell what `nsmem` means in the context of the PR, but outside of this diff, I think it's nonobvious.

src/gurt/telemetry.c
```diff
@@ -741,25 +783,27 @@ alloc_ctx(struct d_tm_context **ctx, struct d_tm_shmem_hdr *shmem, int shmid)
...
+	rc = allocate_memory_segment(key, size_bytes, shared, &shmem);
 	if (rc < 0)
 		return rc;
 
 	*new_shmid = rc;
 	rc = alloc_node(shmem, &shmem->sh_root, root_path);
 	if (rc != 0) {
-		destroy_shmem(*new_shmid);
+		if (shared)
+			destroy_shmem(*new_shmid);
```
kjacque, line 806: Need to free if it's non-shared?

kjacque (commented):
@mjmac I don't have a lot of issues with it actually... My main quibble is having structs named "shmem" which now no longer point to shared mem (though you cleaned up the function names). Some of the free/destroy logic is a bit repetitive as well and could probably be shared in a helper. But that's all polishing up stuff. The basic approach is cleaner than I expected it to be.

src/gurt/telemetry.c
mjmac: Dunno... I'll see if I can make it behave.

src/gurt/telemetry.c
mjmac: Yep, good catch. Fixed.

src/gurt/telemetry.c
```diff
@@ -4288,28 +4287,28 @@ validate_shmem_ptr(struct d_tm_shmem_hdr *shmem_root, void *ptr)
...
  * \param[in]	node	Node this pointer belongs to
  * \param[in]	ptr	The pointer to convert
  *
- * \return		A pointer to the item in the client's address
+ * \return		A pointer to the item in the consumer's address
  *			space
  *			NULL if the pointer is invalid
  */
 void *
 d_tm_conv_ptr(struct d_tm_context *ctx, struct d_tm_node_t *node, void *ptr)
```
kjacque, line 4302: An aside, you could probably just return the pointer directly in the non-shared case.

kjacque (approved):
Thanks! LGTM

jolivier23 (approved):

mchaarawi (approved):

mjmac:
@daos-stack/daos-gatekeeper: Passed all of the tests in the old flow.

PR #16246, opened 2025-04-11
DAOS-16908 object: add client-side compound RPC pinging RPC on DER_RE…

karthjyojay:
…CONNECT.

Like with the single update case, when the server returns DER_RECONNECT, it means that the server did not have an established TCP connection to the client it tried to do the bulk transfer from. We need to handle this logic in the compound RPC case, where each subrequest could result in this error.

### Steps for the author:


src/object/srv_obj.c
```diff
@@ -494,14 +494,24 @@ bulk_transfer_sgl(daos_handle_t ioh, crt_rpc_t *rpc, crt_bulk_t remote_bulk,
...
 	bool			async = true;
 	uint64_t		time = daos_get_ntime();
 
+	if (bulk_op == CRT_BULK_GET && dth != NULL && !(dth->dth_flags & DTE_LEADER) &&
+	    DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE)) {
+		/** Fault injection - client unreachable. */
+		firewall_blocking_flag = true;
+	}
+
+	if (firewall_blocking_flag) {
```
jolivier23, line 511: I would create a block for bulk_op == CRT_BULK_GET and put this in it

src/object/srv_obj.c
```diff
@@ -494,14 +494,24 @@ bulk_transfer_sgl(daos_handle_t ioh, crt_rpc_t *rpc, crt_bulk_t remote_bulk,
...
+		  int sgl_nr, int bulk_nr, struct obj_bulk_args *p_arg, struct dtx_handle *dth)
 {
 	struct obj_bulk_args	arg = { 0 };
 	int			i, rc, *status, ret;
 	int			skip_nr = 0;
 	bool			async = true;
 	uint64_t		time = daos_get_ntime();
 
+	if (bulk_op == CRT_BULK_GET && dth != NULL && !(dth->dth_flags & DTE_LEADER) &&
+	    DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE)) {
```
jolivier23, line 506: This check needs to be before leader check IMO.  Otherwise, it has unintended consequence of leader failing a subsequent update

src/object/srv_obj.c
karthjyojay: Done.

src/object/srv_obj.c
karthjyojay: Done.

src/object/srv_obj.c
```diff
@@ -494,14 +494,26 @@ bulk_transfer_sgl(daos_handle_t ioh, crt_rpc_t *rpc, crt_bulk_t remote_bulk,
...
 	int			i, rc, *status, ret;
 	int			skip_nr = 0;
 	bool			async = true;
 	uint64_t		time = daos_get_ntime();
 
+	if (bulk_op == CRT_BULK_GET) {
+		if (DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE) && dth != NULL &&
+		    !(dth->dth_flags & DTE_LEADER)) {
+			/** Fault injection - client unreachable. */
+			firewall_blocking_flag = true;
```
Nasf-Fan, line 509: Why not directly return -DER_RECONNECT?

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,127 @@ dtx_41(void **state)
...
+	char          fetch_buf[DTX_IO_BULK_TRANSFER];
+	daos_handle_t th = {0};
+	daos_obj_id_t oid;
+	struct ioreq  req;
+	int           i;
+	int           rc;
+
+	FAULT_INJECTION_REQUIRED();
+
+	test_set_engine_fail_loc(arg, CRT_NO_RANK, DAOS_CLIENT_UNREACHABLE | DAOS_FAIL_ONCE);
```
Nasf-Fan, line 3068: If you replace DAOS_FAIL_ONCE with DAOS_FAIL_ALWAYS, then do we still need firewall_blocking_flag?

karthjyojay: I don't think this strategy will work because DAOS_FAIL_ALWAYS will always return the DER_RECONNECT error when we only want it returned once on the first update before any ping has been made.

jolivier23: @Nasf-Fan See https://github.com/daos-stack/daos/pull/16248 which may be illustrative of why this doesn't work.

src/tests/suite/daos_dist_tx.c
```diff
@@ -3104,95 +3227,58 @@ dtx_sub_teardown(void **state)
...
+    {"DTX37: resync - leader failed during prepare", dtx_37, dtx_sub_setup, dtx_sub_teardown},
+    {"DTX38: resync - lost whole redundancy groups", dtx_38, dtx_sub_rf0_setup, dtx_sub_teardown},
+    {"DTX39: not restart the transaction with fixed epoch", dtx_39, dtx_sub_rf1_setup,
+     dtx_sub_teardown},
+
+    {"DTX40: uncertain check - miss commit with delay", dtx_40, NULL, test_case_teardown},
+    {"DTX41: uncertain check - miss abort with delay", dtx_41, NULL, test_case_teardown},
+    {"DTX42: client unreachable - retry - multiple SV update against the same obj", dtx_42, NULL,
+     test_case_teardown},
+    {"DTX42: client unreachable - retry - multiple SV update against multiple objs", dtx_43, NULL,
```
jolivier23, line 3280: DTX43 (I fixed this in my PR)

jolivier23 (commented):
minor nit.  I wonder if it would be better to merge my patch into this

src/tests/suite/daos_dist_tx.c
karthjyojay: Done.

src/object/cli_obj.c
```diff
@@ -1502,7 +1495,7 @@ ping_task(tse_task_t *task)
 		}
```
jolivier23, line 1495: The tgt_list won't get freed if this fails.  Maybe this should be a goto to clean up?

src/object/cli_obj.c
```diff
@@ -1462,35 +1462,28 @@ struct ping_task_arg {
...
-	obj                       = arg->obj;
+	tgt_list                   = &arg->tgt_list;
 
 	D_INIT_LIST_HEAD(&ping_task_list);
 
-	grp_idx      = obj_dkey2grpidx(obj, arg->dkey_hash, obj->cob_version);
-	first_target = grp_idx * obj->cob_grp_size;
-
-	for (i = 0, target = first_target; i < obj->cob_grp_size; i++, target++) {
+	d_list_for_each_entry(entry, tgt_list, link) {
```
jolivier23, line 1484: This really could be pop_entry here, no?  Though I suppose you still have to free the list at the end.

src/object/obj_internal.h
```diff
@@ -906,6 +911,11 @@ dc_sgl_out_set(d_sg_list_t *sgl, daos_size_t data_size)
 	}
 }
 
+int
+obj_gather_tgt_ids(d_list_t *head, struct dc_object *obj, uint64_t dkey_hash);
```
jolivier23, line 915: is clang format enforcing the multiple line prototype?  It doesn't match the two below it.  Definitely this is the style for the definition.

jolivier23 (changes requested):

src/object/obj_tx.c
```diff
@@ -1077,6 +1083,61 @@ dc_tx_commit_cb(tse_task_t *task, void *data)
...
+			dcsr = &tx->tx_req_cache[i + start];
+			if (dcsr->dcsr_opc == DCSO_UPDATE) {
+				int      tgt_ids_rc;
+				d_list_t tgt_list_head;
+				D_INIT_LIST_HEAD(&tgt_list_head);
+
+				obj = dcsr->dcsr_obj;
+
+				tgt_ids_rc =
+				    obj_gather_tgt_ids(&tgt_list_head, obj, dcsr->dcsr_dkey_hash);
```
wangdi1, line 1105: hmm, I think you should gather target ids from other reqs, then create the ping task all in once? i.e.

D_INIT_LIST_HEAD(&tgt_list_head);
for (i = 0; i < req_cnt; i++) {
           if (dscr->opc == UPDATE) {
               obj_gather_tgt_ids();  /* eliminate the duplicate target, I understand we decide to ignore that in the first step, but it should be easy to implement in this function anyway. */
             }
 }
 
 obj_tgt_ping_task();  /* hmm, you do not need create another independent required_task, just create those ping_tgt_task and make them dependent on the task directly? */

}

src/object/cli_obj.c
karthjyojay: Done.

src/object/cli_obj.c
karthjyojay: I left it as-is since we need to free the list at the end anyway. Let me know if you think I should change it.

src/object/obj_internal.h
karthjyojay: It looks like clang format is enforcing this.

src/object/obj_tx.c
karthjyojay: Thanks for the suggestion. I did a heavy refactoring to remove all the layers of tasks between the caller method and individual target ping tasks. I also did what you said by gathering all targets across all sub requests and then calling the ping task logic with those target ids.

src/object/cli_obj.c
```diff
@@ -1455,136 +1457,69 @@ ping_task_abort(tse_task_t *task, void *arg)
...
-ping_task(tse_task_t *task)
+int
+obj_create_ping_task(tse_sched_t *sched, daos_handle_t pool_hdl, d_list_t *tgt_list,
+		     tse_task_t **task)
 {
-	daos_handle_t          pool_hdl;
-	int                   rc = 0;
-	d_list_t              ping_task_list;
-	d_list_t              *tgt_list;
+	int                    rc;
```
wangdi1, line 1474: if the tgt_list is empty, which is possible if gather_tgt_ids failed, then it will return rc directly, so this rc need to be initialized as 0.

src/object/obj_tx.c
```diff
@@ -1092,48 +1091,37 @@ dc_tx_commit_cb(tse_task_t *task, void *data)
...
+		obj = dcsr->dcsr_obj;
+		D_ASSERT(obj != NULL);
+		D_ASSERT(obj->cob_pool != NULL);
+		dc_pool2hdl_noref(obj->cob_pool, &ph);
+
+		ping_task_rc = obj_create_ping_task(sched, ph, &tgt_list_head, &task);
+		if (ping_task_rc != 0) {
+			D_ERROR("Failed to create ping task for task %p: %d, %d\n", task,
+				ping_task_rc, rc);
+			tx->tx_status = TX_ABORTED;
```
wangdi1, line 1122: Hmm, this tx_status will be overwritten by following "TX_FAILED"? should tx->tx_status = TX_FAILED in else ?

wangdi1 (commented):
Looks much better, though 2 comments here, please have a look.

src/object/obj_tx.c
karthjyojay: Done.

src/object/cli_obj.c
karthjyojay: Done.

src/object/obj_tx.c
wangdi1: Hmm, if the ping task can not be created in this case, then it will retry or aborted? And also the following rc check will set 
                  tx->tx_retry = 1
if it suppose to abort the transaction? ( should it?), does this retry matter?

@Nasf-Fan  could you please check as well?

karthjyojay: To rephrase,

1) If the ping task cannot be created, should we set the transaction to abort or should we set it to failed and have the caller retry?
2) If we abort the transaction, will the abort be ignored and will the transaction be retried anyway?

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,127 @@ dtx_41(void **state)
...
+	char         *write_bufs[DTX_TEST_SUB_REQS];
+	char         *fetch_buf;
+	daos_handle_t th = {0};
+	daos_obj_id_t oids[DTX_TEST_SUB_REQS];
+	struct ioreq  reqs[DTX_TEST_SUB_REQS];
+	int           i;
+	int           rc;
+
+	FAULT_INJECTION_REQUIRED();
+
```
Nasf-Fan, line 3120: print_message("DTX43...");

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,127 @@ dtx_41(void **state)
 	dtx_uncertainty_miss_request(*state, DAOS_DTX_MISS_ABORT, true, true);
 }
 
+static void
+dtx_42(void **state)
+{
+	test_arg_t   *arg  = *state;
+	const char   *dkey = dts_dtx_dkey;
+	const char   *akey = dts_dtx_akey;
+	char          write_bufs[DTX_TEST_SUB_REQS][DTX_IO_BULK_TRANSFER];
```
Nasf-Fan, line 3058: Not a good practice to allocate too large buffer on stack.

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,127 @@ dtx_41(void **state)
...
+	}
+
+	rc = daos_tx_commit(th, NULL);
+	assert_rc_equal(rc, -DER_TX_RESTART);
+
+	// The first attempt at committing the transaction fails due to DER_RECONNECT.
+	MUST(daos_tx_restart(th, NULL));
+
+	for (i = 0; i < DTX_TEST_SUB_REQS; i++) {
+		oids[i] = daos_test_oid_gen(arg->coh, OC_EC_2P1G1, 0, 0, arg->myrank);
```
Nasf-Fan, line 3144: Please reuse former generated oids[i].

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,127 @@ dtx_41(void **state)
...
+		ioreq_init(&reqs[i], arg->coh, oids[i], DAOS_IOD_SINGLE, arg);
+
+		D_ALLOC(write_bufs[i], DTX_IO_BULK_TRANSFER);
```
Nasf-Fan, line 3147: Former allocated write_bufs[i] will be leaked, please reuse it.

src/object/srv_obj.c
```diff
@@ -494,14 +494,26 @@ bulk_transfer_sgl(daos_handle_t ioh, crt_rpc_t *rpc, crt_bulk_t remote_bulk,
...
 
+	if (bulk_op == CRT_BULK_GET) {
+		if (dth != NULL && DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE) &&
+		    !(dth->dth_flags & DTE_LEADER)) {
+			/** Fault injection - client unreachable. */
+			firewall_blocking_flag = true;
+		}
+
+		if (firewall_blocking_flag) {
+			return -DER_RECONNECT;
```
Nasf-Fan, line 513: Once the flag is set, then before ds_pool_tgt_warmup_handler() is triggered, all write style bulk transfer, it spite of it is from the client or not, in spite of it is for UPDATE or CPD RPC or others, in spite of current target is the leader or not, all will hit -DER_RECONNECT, does it the expected behavior?

src/object/obj_tx.c
```diff
@@ -1077,6 +1082,50 @@ dc_tx_commit_cb(tse_task_t *task, void *data)
...
+		for (i = 0; i < req_cnt; i++) {
+			dcsr = &tx->tx_req_cache[i + start];
+			if (dcsr->dcsr_opc == DCSO_UPDATE) {
+				obj = dcsr->dcsr_obj;
+				tgt_ids_rc =
+				    obj_gather_tgt_ids(&tgt_list_head, obj, dcsr->dcsr_dkey_hash);
+				if (tgt_ids_rc != 0) {
+					DL_ERROR(tgt_ids_rc, "Failed to gather target IDs");
+				}
+			}
```
Nasf-Fan, line 1108: If the CPD RPC body is too large (especially for punch large scaled object) to be held inside inline RPC, it will be transferred via bulk data from client to all related targets (not only leader, but also non-leaders). Under such case, if some non-leader(s) returned DER_RECONNECT for bulk transferred CPD RPC body, we also needs to tigger "ping" task for related target(s); otherwise, related CPD RPC after restart may hit DER_RECONNECT again.

src/object/obj_tx.c
Nasf-Fan: As my understand, the connect from the client and the server can be shared by multiple processes on the same client (to the same server), right? If yes, as long as one process establishes the connection with the server for some reason, then all the other subsequent bulk data transfer will not be rejected. So here:
1. If there is some in-flight ping RPC for the server, we should wait there instead of triggering more for the same server to avoid repeated (too much) ping RPCs to the same server.
2. For DER_RECONNECT case, we do not need to restart TX, instead, convert "rc = 0", that will make CPD RPC to be resent after "ping" task completed.
3. If failed to ping some server, it may be not fatal since the others may have already established the connection with the same server. So we can make current TX to be resent as above 2nd with some warning message.
4. Above case 3 may cause current process to be blocked for very long time since we do not know whether some other will help to establish related connection or not. So another option is to abort current transaction by convert "rc" as some non-retriable errno# and set "tx->tx_status = TX_ABORTED".

Nasf-Fan (changes requested):

src/object/srv_obj.c
jolivier23: ah, so should probably move the dth != NULL to the outer block

src/tests/suite/daos_dist_tx.c
jolivier23: Good catch.  Yeah, this should be either dynamic or a static buffer.

src/object/obj_tx.c
jolivier23: that is a good point

karthjyojay: If this case happens, will execution on the client side still go to where we've added the pinging logic? It sounds like our logic covers this? Are you saying we are missing this logic?

If we cover it, we can add a unit test for this.

jolivier23:
@Nasf-Fan what do you mean by multiple processes?  The connection is per cart endpoint per server.   So at best you can have multiple in-flight asynchronous RPCs from the same process/thread but definitely not across processes.  Waiting for those outstanding RPCs sounds like a follow-up optimization to me.

src/object/obj_tx.c
karthjyojay: To unit test this, I think we need to pass in a valid dtx handle inside `ds_obj_cpd_body_bulk` to the `obj_bulk_transfer` call. I'm currently not doing this.

src/object/obj_tx.c
karthjyojay: Trying to keep one thread for this. As @jolivier23 said in https://github.com/daos-stack/daos/pull/16246#issuecomment-2815856682, there is a connection between each client-side cart context and server-side cart context. The server is running a thread per target and there is one cart context per thread.
Multiple processes on the same client will not share the same connection to the server.

1) with what we're doing, at worst, we will ping the same target multiple times, but I think this is something we can optimize on later if that is ok with you.
2) Are you saying to do rc=0 instead of restarting the transaction?
3) Each client-side cart context shares a connection with the server-side cart context, so there is a connection to each target. If a ping fails, then I think this is fatal because the associated cart context will not be able to establish a connection with the server target cart context. This is because processes do not share a connection to each target. Therefore, I think we need to error on this.
4) I don't think this is relevant since processes are not shared.

What do you think?

src/object/obj_tx.c
karthjyojay: I added a unit test for this. Please take another look.

src/object/srv_obj.c
karthjyojay: We are using this `firewall_blocking_flag` just for tests, so I think the behavior we are getting out of it is ok. Please let me know what you think.

src/tests/suite/daos_dist_tx.c
karthjyojay: Done.

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,127 @@ dtx_41(void **state)
...
+	char          write_bufs[DTX_TEST_SUB_REQS][DTX_IO_BULK_TRANSFER];
+	char          fetch_buf[DTX_IO_BULK_TRANSFER];
+	daos_handle_t th = {0};
+	daos_obj_id_t oid;
+	struct ioreq  req;
+	int           i;
+	int           rc;
+
+	FAULT_INJECTION_REQUIRED();
+
```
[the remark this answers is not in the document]
karthjyojay: Done.

src/tests/suite/daos_dist_tx.c
karthjyojay: Made this a dynamic buffer.

src/tests/suite/daos_dist_tx.c
karthjyojay: Done.

src/tests/suite/daos_dist_tx.c
karthjyojay: Done.

Nasf-Fan:
> @Nasf-Fan what do you mean by multiple processes? The connection is per cart endpoint per server. So at best you can have multiple in-flight asynchronous RPCs from the same process/thread but definitely not across processes. Waiting for those outstanding RPCs sounds like a follow-up optimization to me.

If the connection if per-cart context based, then we can consider similar logic as tx_epoch_task does, that will be helpful to avoid server overloaded because of ping flood. On some degree, it is not simple efficiency improvement, instead, it may affect the system availability on large scaled cluster. But anyway, you may have better understanding about the users IO mode and workload. So please consider related optimization as necessary.

src/object/obj_tx.c
Nasf-Fan: > 1. with what we're doing, at worst, we will ping the same target multiple times, but I think this is something we can optimize on later if that is ok with you.

I am not sure for this, it depends on the application IO mode and workload. I am afraid that the ping flood may cause server to be overloaded.

> 2. Are you saying to do rc=0 instead of restarting the transaction?

I prefer to resend CPD RPC (tx_retry = 1) after ping completed instead of restart TX.

> 3. Each client-side cart context shares a connection with the server-side cart context, so there is a connection to each target. If a ping fails, then I think this is fatal because the associated cart context will not be able to establish a connection with the server target cart context. This is because processes do not share a connection to each target. Therefore, I think we need to error on this.

OK, then we may have to abort the TX if failed to establish the connection with the target. Under such case, set "tx->tx_status = TX_ABORTED" and "rc = -DER_RECONNECT".

> 4. I don't think this is relevant since processes are not shared.

karthjyojay: @wangdi1 @jolivier23 I think we discussed offline that we are ok with pinging the same target multiple times and optimize this later. Should we consider this a blocker before calling this feature done?

@Nasf-Fan I made the change to resend the compound rpc instead of restarting the transaction. Thank you for this suggestion. I also made the `rc=-DER_RECONNECT` change in the case the ping task failed.

src/object/srv_internal.h
```diff
@@ -194,6 +195,12 @@ enum latency_type {
 	VOS_LATENCY,
 };
 
+enum fi_leader {
+	UNSPECIFIED,
+	LEADER,
```
jolivier23, line 200: Add DTX_ prefix to these.

what is fi ?  I think you can just make this an anonymous enum and not name it.  There is no type checking in C anyway so can just use int parameter.   If you do keep it, fi_leader should probably be something like dtx_target_identity or something like that.

src/object/srv_obj.c
```diff
@@ -494,14 +494,42 @@ bulk_transfer_sgl(daos_handle_t ioh, crt_rpc_t *rpc, crt_bulk_t remote_bulk,
...
 	struct obj_bulk_args	arg = { 0 };
 	int			i, rc, *status, ret;
 	int			skip_nr = 0;
 	bool			async = true;
 	uint64_t		time = daos_get_ntime();
 
+	// Fault injection for the case where the compound RPC body bulk transfer may throw a
+	// DER_RECONNECT. dth will be NULL in this test case.
+	if (dth == NULL && bulk_op == CRT_BULK_GET) {
+		if (DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE_CPD_BODY) &&
```
jolivier23, line 509: if ((DAOS_FAIL_CHECK_DAOS_CLIENT_UNREACHABLE) || DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE_CPD_BODY)) && fi_leader_value == FOLLOWER) 

No need to have two blocks.  You can't set more than one fail loc at a time.

jolivier23: another option would be to land this (into your branch)
https://github.com/daos-stack/daos/pull/16248/

src/object/srv_internal.h
karthjyojay: "fi" was "fault injection" as we are using this enum for testing purposes. I've renamed it like you suggested.

src/object/srv_obj.c
```diff
@@ -495,38 +495,41 @@ int
...
+			if (dth != NULL) {
+				if (DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE) &&
+				    identity == DTX_FOLLOWER) {
+					/** Fault injection - client unreachable. */
+					firewall_blocking_flag = true;
+				}
 
-		if (firewall_blocking_flag) {
-			return -DER_RECONNECT;
+				if (firewall_blocking_flag) {
```
jolivier23, line 529: Minor, I think this can be moved outside of if/else since it's the same for both branches.

jolivier23 (approved):

jolivier23 (approved):

src/object/srv_obj.c
karthjyojay: Done.

src/object/cli_obj.c
```diff
@@ -1929,6 +1866,43 @@ dc_obj_retry_delay(tse_task_t *task, int err, uint16_t *retry_cnt, uint16_t *inp
...
+
+	// Get list of targets
+	grp_idx      = obj_dkey2grpidx(obj, dkey_hash, obj->cob_version);
+	first_target = grp_idx * obj->cob_grp_size;
+
+	for (i = 0, target = first_target; i < obj->cob_grp_size; i++, target++) {
+		D_ALLOC_PTR(entry);
+		if (entry == NULL)
+			D_GOTO(err, rc = -1);
```
wangdi1, line 1887: rc = -DER_NOMEM

wangdi1 (approved):

src/object/cli_obj.c
```diff
@@ -1456,134 +1458,69 @@ ping_task_abort(tse_task_t *task, void *arg)
...
+		D_DEBUG(DB_TRACE, "creating ping_tgt_task for tgt ID %d", entry->tgt_id);
 
 		rc = tse_task_create(ping_tgt_task, sched, NULL, &ping_task);
 		if (rc != 0) {
 			tse_task_list_traverse(&ping_task_list, ping_task_abort, &rc);
 
 			DL_ERROR(rc, "failed to create task");
 			tse_task_complete(ping_task, rc);
-			return rc;
+			goto cleanup;
```
Nasf-Fan, line 1493: Since tse_task_create() failed, why need tse_task_complete() for the failed task?

src/object/cli_obj.c
```diff
@@ -1929,6 +1866,43 @@ dc_obj_retry_delay(tse_task_t *task, int err, uint16_t *retry_cnt, uint16_t *inp
...
+	int                    target;
+	int                    tgt_id;
+	struct tgt_list_entry *entry;
+	int                    rc;
+
+	// Get list of targets
+	grp_idx      = obj_dkey2grpidx(obj, dkey_hash, obj->cob_version);
+	first_target = grp_idx * obj->cob_grp_size;
+
+	for (i = 0, target = first_target; i < obj->cob_grp_size; i++, target++) {
```
Nasf-Fan, line 1884: Why need establish connection for all the targets in the redundancy group? For replicated object, that is fine; but for EC object, it seems unnecessary for some non-touched target(s).

Nasf-Fan (changes requested):

src/object/srv_obj.c
```diff
@@ -520,14 +520,41 @@ bulk_transfer_sgl(daos_handle_t ioh, crt_rpc_t *rpc, crt_bulk_t remote_bulk,
 int
 obj_bulk_transfer(crt_rpc_t *rpc, crt_bulk_op_t bulk_op, bool bulk_bind, crt_bulk_t *remote_bulks,
 		  uint64_t *remote_offs, uint8_t *skips, daos_handle_t ioh, d_sg_list_t **sgls,
-		  int sgl_nr, int bulk_nr, struct obj_bulk_args *p_arg)
+		  int sgl_nr, int bulk_nr, struct obj_bulk_args *p_arg, struct dtx_handle *dth,
+		  enum dtx_target_identity identity)
```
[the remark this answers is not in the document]
karthjyojay: In our fault injection test for compound RPC body bulk transfer, we don't have a dtx handle. We do have a boolean for if we are the leader. I thought an enum is cleanest to give to the `obj_bulk_transfer` method. So if we have a `dth`, we are in `obj_bulk_transfer` doing a real update and if we have a NULL `dth`, then we are in the CPD body bulk transfer.

src/object/srv_obj.c
karthjyojay: I don't think this will work because `ds_obj_cpd_body_bulk` does not have a dth. To test the CPD body bulk data transfer case, `ds_obj_cpd_body_bulk` passes `obj_bulk_transfer` a NULL dth. We therefore use dth to determine if we are in the body bulk transfer or update bulk transfer case and use `identity` to know if it is a follower, so we can trigger the fail check.

src/object/cli_obj.c
karthjyojay: Done.

src/object/cli_obj.c
karthjyojay: @jolivier23 can you please correct me if I am wrong?

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,171 @@ dtx_41(void **state)
...
+	const char   *akey = dts_dtx_akey;
+	char         *fetch_buf;
+	daos_handle_t th = {0};
+	daos_obj_id_t oid;
+	struct ioreq  req;
+	int           i;
+	int           subrequests = 2;
+	char         *write_bufs[subrequests];
+
+	par_barrier(PAR_COMM_WORLD);
```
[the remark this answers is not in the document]
karthjyojay: Done.

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,171 @@ dtx_41(void **state)
...
+	/* The last value will be fetched. */
+	assert_memory_equal(write_bufs[subrequests - 1], fetch_buf, DTX_IO_BULK_TRANSFER);
+
+	for (i = 0; i < subrequests; i++) {
+		D_FREE(write_bufs[i]);
+	}
+	ioreq_fini(&req);
+	D_FREE(fetch_buf);
+
+	MUST(daos_tx_close(th, NULL));
```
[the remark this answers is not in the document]
karthjyojay: Done.

src/object/cli_obj.c
Nasf-Fan: For example, partial update against EC16P2 object, current update may only touches 1 data shard + 2 parity shards. So in theory, we only needs the connection from the client to such 3 targets. If we try to establish the connection for all the 18 targets in the RDG, then it is not only efficiency issue, but also potential failure for the connection with unrelated target(s). It is something like that you only wants to write shard_1, but the server tell you that you can write because the connection with shard_2 cannot be established. So I am not against to optimize the efficiency related issue, just concern whether we should allow IO failure because of unrelated target trouble.

src/object/srv_obj.c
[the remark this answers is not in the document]
Nasf-Fan: Then can we only use one parameter for such purpose? for example:
```
obj_bulk_transfer(crt_rpc_t *rpc, crt_bulk_op_t bulk_op, bool bulk_bind, crt_bulk_t *remote_bulks,
		  uint64_t *remote_offs, uint8_t *skips, daos_handle_t ioh, d_sg_list_t **sgls,
		  int sgl_nr, int bulk_nr, struct obj_bulk_args *p_arg, bool check_conn)
```
Then the caller itself can decide whether need to check the connection based on the dth flag or RPC input flag (for bulk transferring CPD body).

src/object/cli_obj.c
jolivier23: Agree this is an issue but I think we can leave it for future optimization.  The ping should be relatively rare in practice

wangdi1: It is indeed ideal if it only ping relevant targets, but it is hard to get RPC in this function(layer). Given the each target will only be pinged once for each client, even the ping may not be useful for this operation, it might be helpful for the following operation anyway. 

If there are some un-relevant targets can not be reached, the ping will timeout, it will still retry the original rpc. Anyway, IMHO, this can be done in the future if this is really necessary.

Nasf-Fan: As said before, establishing the connection is only used for special GCP case, it is OK for me to optimize related logic in the future. Pointing out the issue just for reminder.

Nasf-Fan:
@karthjyojay , I am fine for most of the implementation in the patch except the new parameters for obj_bulk_transfer(). It is unnecessary to introduce such redundancy for test purpose. One parameter seems enough for that. If you have chance to refresh your patch, I would suggest to reconsider related logic.

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,171 @@ dtx_41(void **state)
...
+dtx_43(void **state)
+{
+	test_arg_t   *arg  = *state;
+	const char   *dkey = dts_dtx_dkey;
+	const char   *akey = dts_dtx_akey;
+	char         *write_bufs[DTX_TEST_SUB_REQS];
+	char         *fetch_buf;
+	daos_handle_t th = {0};
+	daos_obj_id_t oids[DTX_TEST_SUB_REQS];
+	struct ioreq  reqs[DTX_TEST_SUB_REQS];
```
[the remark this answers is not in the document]
karthjyojay: Done.

src/tests/suite/daos_dist_tx.c
```diff
@@ -3047,6 +3049,171 @@ dtx_41(void **state)
...
+	char         *write_bufs[subrequests];
+
+	par_barrier(PAR_COMM_WORLD);
+
+	FAULT_INJECTION_REQUIRED();
+
+	test_set_engine_fail_loc(arg, CRT_NO_RANK, DAOS_CLIENT_UNREACHABLE | DAOS_FAIL_ONCE);
+
+	print_message("DTX42: client unreachable - retry - multiple SV update against the same obj "
+		      "with fault injection\n");
```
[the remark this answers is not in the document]
karthjyojay: Done.

src/object/obj_tx.c
```diff
@@ -1078,6 +1083,51 @@ dc_tx_commit_cb(tse_task_t *task, void *data)
...
 
+	// If the task's return code is DER_RECONNECT, then ping every update RPC's
+	// targets to establish connection between the client and server. We do not know
+	// which sub request resulted in this error.
+	if (rc == -DER_RECONNECT) {
+		struct daos_cpd_sub_req *dcsr;
+		struct dc_object        *obj;
+		int                      ping_task_rc;
+		start   = dc_tx_leftmost_req(tx, false);
+		req_cnt = tx->tx_read_cnt + tx->tx_write_cnt;
```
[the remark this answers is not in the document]
karthjyojay: Done.

src/object/obj_tx.c
```diff
@@ -1078,6 +1083,51 @@ dc_tx_commit_cb(tse_task_t *task, void *data)
...
+				}
+			}
+		}
+
+		dcsr = &tx->tx_req_cache[dc_tx_leftmost_req(tx, false)];
+		D_ASSERT(dcsr != NULL);
+		obj = dcsr->dcsr_obj;
+		D_ASSERT(obj != NULL);
+		D_ASSERT(obj->cob_pool != NULL);
+		dc_pool2hdl_noref(obj->cob_pool, &ph);
```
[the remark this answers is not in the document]
karthjyojay: Done.

src/object/srv_obj.c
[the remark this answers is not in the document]
karthjyojay: Done, I made a function `check_conn` that checks the dth or leader boolean flag. Im still using the existence of `dth` to know if we are in the `obj_bulk_transfer` due to cpd body transfer or because of the actual update, but moved this logic into the caller of `obj_bulk_transfer` to limit the number of new arguments.

karthjyojay:
@Nasf-Fan can you please take another look at how I introduced one boolean to `obj_bulk_transfer` and approve if you are ok with it? We are using the existence of `dth` to know if the caller is the cpd body transfer or normal update request and also `dth` or `leader` booleans to determine if we want the DTX member to return an error, as only the follower should return an error.

jolivier23 (approved):

Nasf-Fan (approved):

src/object/srv_obj.c
```diff
@@ -517,17 +517,42 @@ bulk_transfer_sgl(daos_handle_t ioh, crt_rpc_t *rpc, crt_bulk_t remote_bulk,
 	return rc;
 }
 
+bool
+check_conn(struct dtx_handle *dth, bool leader)
+{
+	// Fault injection for the case where the compound RPC body bulk transfer may throw
+	// a DER_RECONNECT. dth will be NULL in this test case.
+	if (dth == NULL) {
+		return DAOS_FAIL_CHECK(DAOS_CLIENT_UNREACHABLE_CPD_BODY) && !leader;
```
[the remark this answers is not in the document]
karthjyojay: I made a Google ticket to track this.

mchaarawi:
how was this merged without functional testing being run?
the PR itself modifies daos_test ..

karthjyojay:
> how was this merged without functional testing being run? the PR itself modifies daos_test ..

I'm not sure; was it supposed to block me from merging it? I talked with @jolivier23 offline and we thought it would be ok to submit this. Do you want me to revert this squashed commit?

jolivier23:
We will take on the task of making sure it is passing once CI comes back online.  This is only merged in the feature branch.

PR #16249, opened 2025-04-12
DAOS-17339 control: Avoid starting AdminExcluded ranks

kjacque:
- Filter out AdminExcluded ranks from fanout if explicit ranks are not requested.
- If AdminExcluded ranks are explicitly requested, return an error by default. The new --ignore-admin-excluded flag can be used to change the behavior to filter out the AdminExcluded ranks.
- Rename errors related to joining AdminExcluded ranks, for clarity.

### Steps for the author:


src/tests/ftest/control/dmg_system_start.yaml
```diff
@@ -0,0 +1,25 @@
+hosts:
+  test_servers: 1
+server_config:
+  name: daos_server
+  engines_per_host: 2
```
daltonbohning, line 5: This new test is working so not an issue. But just FYI this is now the only test using dual-engine on VMs (sharing `fabric_iface: eth0`). The only issue I could think of is that on our VMs, `targets: 4` * 2 might overload the VMs. There is only one other VM test using a total of 8 targets and it sets `DAOS_TARGET_OVERSUBSCRIBE=1`.  But this test didn't seem to need that.
Anyway, not an issue unless we run into something in the future. IMO it's great to user fewer nodes when possible

daltonbohning (commented):
ftest LGTM

mjmac (approved):
LGTM, good fix and UX improvement.

src/tests/ftest/control/dmg_system_start.yaml
kjacque: Hmm, okay. I actually wrote the file this way entirely so it would run on my solitary node in the lab. It doesn't necessarily need to be two engines on a single node, that was just all I had access to.

daltonbohning: Good point. IMO it's great to do that because it makes it easier for people to test locally

tanabarr (approved):

src/control/cmd/dmg/system.go
```diff
@@ -112,6 +112,11 @@ type baseRankListCmd struct {
 	rankListCmd
 }
 
+type liveRankListCmd struct {
```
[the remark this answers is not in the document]
kjacque: Yeah, I struggled with the terminology there. "nonAdminExcludedRanksCmd" felt like too much of a mouthful.

src/control/server/faults.go
```diff
@@ -200,6 +200,16 @@ func FaultJoinReplaceEnabledPoolRank(rank ranklist.Rank, poolIDs ...string) *fau
 	)
 }
 
+// FaultRankAdminExcluded indicates that the given rank list is administratively excluded.
+func FaultRankAdminExcluded(ranks ranklist.RankList) *fault.Fault {
+	return serverFault(
+		code.ServerRankAdminExcluded,
+		fmt.Sprintf("%s [%s] %s administratively excluded and cannot be operated on by this command", english.PluralWord(len(ranks), "rank", "ranks"), ranks.String(),
+			english.PluralWord(len(ranks), "is", "are")),
+		"re-run the command without requesting the administratively excluded rank(s)",
```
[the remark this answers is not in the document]
kjacque: Ah, you're right. Given that it's already nearly through CI, though, I'll fix it only if I need to repush.

kjacque:
Checking on the CI run, all tests actually passed. The appearance of stages that haven't run is due to some changes in which stages run in CI while this PR was percolating through. Requesting forced landing.

daltonbohning:
> Checking on the CI run, all tests actually passed. The appearance of stages that haven't run is due to some changes in which stages run in CI while this PR was percolating through. Requesting forced landing.

Right. This PR successfully ran the "old" requirements so is sufficient.

PR #16250, opened 2025-04-12
DAOS-15947 control: Unit test for PoolRanksReq type conversion

tanabarr:
### Steps for the author:

### DIFF
--- a/src/control/lib/control/pool_test.go
+++ b/src/control/lib/control/pool_test.go
@@ -235,7 +235,7 @@ func TestControl_PoolUpgrade(t *testing.T) {
 	}
 }
 
-func TestControl_PoolRankseReq_Convert(t *testing.T) {
+func TestControl_PoolRanksReq_Convert(t *testing.T) {
 	req := &PoolRanksReq{
 		ID:        "foo",
 		Ranks:     []ranklist.Rank{1, 2, 3},
--- a/src/control/lib/control/pool_test.go
+++ b/src/control/lib/control/pool_test.go
@@ -235,6 +235,55 @@ func TestControl_PoolUpgrade(t *testing.T) {
 	}
 }
 
+func TestControl_PoolRankseReq_Convert(t *testing.T) {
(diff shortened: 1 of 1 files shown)


src/control/lib/control/pool_test.go
```diff
@@ -235,6 +235,55 @@ func TestControl_PoolUpgrade(t *testing.T) {
 	}
 }
 
+func TestControl_PoolRankseReq_Convert(t *testing.T) {
```
kjacque, line 238: spelling nit
```suggestion
func TestControl_PoolRanksReq_Convert(t *testing.T) {
```

kjacque (commented):
Any reason not to make this a table-based test? The protobuf types are all different, but in the test case struct definition, you could use `proto.Message`, e.g.
```go
        for name, tc := range map[string]struct{
                req *PoolRanksReq
                protoReq proto.Message
                expProtoReq proto.Message
        }{
                "PoolDrainReq": {
                        req: &PoolRanksReq{
                                 // ...whatever fields need to be set...
                        },
                        protoReq: new(mgmtpb.PoolDrainReq),
                        expProtoReq: &mgmtpb.PoolDrainReq{
                                // ... etc...
                        },
                },
```
`cmp.Diff` should be smart enough to handle that correctly. Then I think these similar tests could use the same core logic.

tanabarr:
@kjacque this test has been committed to the 2.6 branch so I would prefer to keep the same on master

knard38 (commented):
functionality LGTM, however I am not go expert and thus do feel legitimate enough to define if non respecting the unit test table-based best practices is acceptable or not.

kjacque:
> @kjacque this test has been committed to the 2.6 branch so I would prefer to keep the same on master

Ah, sorry, didn't realize that. Yes in that case I think it is fine for now. We can make improvements on master in other commits.

knard38 (approved):
From my understanding, it sounds that these non table test is acceptable in this context.

PR #16257, opened 2025-04-14
DAOS-623 build: Fix protobuf generator Makefile

mjmac:
Also regenerate protobufs so that they've all been done with
the same versions of the tools.


kjacque (approved):
Thanks for fixing this. I was playing around with the Makefile while setting up a new machine recently, but hadn't quite gotten around to polishing it up.

jolivier23 (approved):

tanabarr (commented):
why do the proto-generated files need to be updated in this PR? where are we keeping track of the minimum/recommended versions for the protoc and gen packages so we don't keep flipping changes? is this maintained in the go mod files?

tanabarr (approved):

mjmac:
> why do the proto-generated files need to be updated in this PR? where are we keeping track of the minimum/recommended versions for the protoc and gen packages so we don't keep flipping changes? is this maintained in the go mod files?

in [src/control/go.mod](https://github.com/daos-stack/daos/blob/master/src/control/go.mod#L29) the current version is `v1.34.1`. The Makefile references this file to choose the tag when installing the toolchains.

PR #16279, opened 2025-04-17
DAOS-17427 control: Restart excluded rank after suicide

tanabarr:
When an engine detects that it has been removed from the system group
map by receiving a CART event, it will now notify its local control plane
with a RAS engine_self_terminated event before terminating its own
process. After receiving this self-termination event, the local
control plane will restart the engine so it can rejoin the system.

The goal of this change is to improve overall system resilience by
automatically recovering engines that are excluded because of
temporary issues such as network instability. Once the engines rejoin,
the rank will still need to be reintegrated into pools as a separate
follow‑up step.

Rate-limiting prevents restart storms: a configurable minimum delay
(default 300 seconds) between restarts per rank ensures system
stability. Two new server config file parameters control behavior:
disable_engine_auto_restart (boolean, default false) completely
disables automatic restarts, while engine_auto_restart_min_delay
(integer seconds) sets the minimum time between consecutive restart
attempts.

Functional tests for the automatic engine restart feature included
with cases to verify disabling, rate-limiting and configuration support.

Features: pool control

### Steps for the author:


src/engine/init.c
```diff
@@ -597,17 +597,21 @@ dss_crt_event_cb(d_rank_t rank, uint64_t incarnation, enum crt_event_source src,
...
+			 * Send RAS event to inform local server of intentional suicide before
+			 * raising a SIGKILL to ourselves. Local daos_server can then decide
+			 * whether to restart rank.
 			 */
+			rc = ds_notify_rank_suicide(rank, incarnation);
+			if (rc)
+				D_ERROR("failed to handle %u/%u event: " DF_RC "\n", src, type,
+					DP_RC(rc));
+
 			rc = kill(getpid(), SIGKILL);
```
liw, line 611: [Discussion] Looking at the code changes I think I'm back to the previously discussed point: Whether the engine kills itself, or only informs the server, who will decide what to do. If the engine kills itself, why do we need the RAS event... The server can simply decide whether to restart a killed engine, even without the RAS event, cannot it...

[Question] If the `ds_notify_rank_suicide` call fails, then the engine will still terminate, but will the server restart the engine?

liw (commented):
Not requesting changes.

knard38:
@tanabarr, I had the same issue with the CI regarding the "Unit test beds with memcheck".
At the end, I have merged with master and restarted the CI.
Now, this stage is successfully run.

src/engine/init.c
tanabarr: [Discussion] I implemented kill+ras mainly because it is guaranteed that the engine process will be killed, the RAS event signifies that it is a suicide as opposed to other termination, do we really want to always restart a terminated rank regardless of reason?

[Question] Yes, if the  notify call fails  then the engine will be terminated but not restarted.  if instead we don't terminate the engine on  group map change and just send ras and put the engine into a blocking state then if  a notify call fails  then the engine will  effectively hang.

liw: > do we really want to always restart a terminated rank regardless of reason?

Hmm, it's indeed hard to say for sure. I'd understand if we'd like to begin conservatively, only restarting for certain cases.

kjacque: I agree we should start conservatively. For engines that crash, or exit for some unknown reason, we don't know if the engine would be OK if we restarted it. Better to let the admin investigate and manually restart the ranks in that case.

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16279/5/execution/node/1303/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/5/testReport/

tanabarr:
PR CI run with `Features: pool control` to give an extra coverage in case restarting engine automatically causes any tests to fail. BoundaryTest and ListVerboseTest failures are unrelated.

@kjacque @knard38 @mjmac @liw can I get reviews please?

tanabarr:
conflicts resolved, doc-only change, CI run no. 5 is still relevant and should be used for PR review

tanabarr:
DAOS-18639 known issue fix should be landed to master

docs/admin/administration.md
```diff
@@ -49,6 +49,8 @@ severity, message, description, and cause.
 | engine\_died| STATE\_CHANGE| ERROR| DAOS engine <idx\> exited exited unexpectedly: <error\> | Indicates engine instance <idx\> unexpectedly. <error> describes the exit state returned from exited daos\_engine process.| N/A                          |
 | engine\_asserted| STATE\_CHANGE| ERROR| TBD| Indicates engine instance <idx\> threw a runtime assertion, causing a crash. | An unexpected internal state resulted in assert failure. |
 | engine\_clock\_drift| INFO\_ONLY   | ERROR| clock drift detected| Indicates CART comms layer has detected clock skew between engines.| NTP may not be syncing clocks across DAOS system.      |
+| engine\_suicide| INFO\_ONLY| NOTICE| excluded rank suicide detected| Indicates that a DAOS engine rank has performed a self-termination (suicide) due to having been excluded from the system's group map. The rank is automatically restarted by the control plane. | An engine was found to be in a transient non-functional state and excluded from the group map. The control plane monitors for this event and automatically restarts the affected engine so it can rejoin the system. |
```
kjacque: Terminology suggestion. I know it's tongue-in-cheek but calling it a "suicide" in our docs/UX feels a little anthropomorphic and maybe insensitive to some readers.

```suggestion
| engine\_self\_terminated| INFO\_ONLY| NOTICE| excluded rank self-termination detected| Indicates that a DAOS engine rank has performed a self-termination due to having been excluded from the system's group map. The rank is automatically restarted by the control plane. | An engine was found to be in a transient non-functional state and excluded from the group map. The control plane monitors for this event and automatically restarts the affected engine so it can rejoin the system. |
```

docs/overview/fault.md
```diff
@@ -84,3 +84,28 @@ can now read from the rebuilt object shards.
 
 This rebuild process is executed online while applications continue accessing
 and updating objects.
+
+### Engine Self-Termination and Automatic Restart
+
+A DAOS engine may be excluded from the group map because of inactivity
+for example. When an engine becomes aware of it's removal from the
+group map it will self-terminate (referred to as "suicide") to protect
```
kjacque: I'd suggest removing all remnants of the "suicide" terminology in docs/user-visible stuff. It's less of an issue in code, but I think it is better to use "self-terminate".

Docs otherwise look good.

src/control/lib/control/event.go
```diff
@@ -170,7 +171,8 @@ func newEventLogger(logBasic logging.Logger, newSyslogger newSysloggerFn) *Event
 }
 
 // NewEventLogger returns an initialized EventLogger capable of writing to the
-// supplied logger in addition to syslog.
+// supplied logger in addition to syslog. Should only be used in production code,
+// use MockEventLogger in unit tests.
```
kjacque, line 175: I dunno if this comment change is needed.

src/control/server/server_utils.go
```diff
@@ -777,6 +815,15 @@ func registerFollowerSubscriptions(srv *server) {
 	srv.pubSub.Reset()
 	srv.pubSub.Subscribe(events.RASTypeAny, srv.evtLogger)
 	srv.pubSub.Subscribe(events.RASTypeStateChange, srv.evtForwarder)
+	srv.pubSub.Subscribe(events.RASTypeInfoOnly,
```
kjacque: Leader and following subscriptions covers the MS... But don't all nodes in the system need to be subscribed to this engine event?

kjacque (changes requested):

src/control/server/server_utils.go
kjacque: Nevermind on this one... the `registerFollowerSubscriptions` function was poorly named. It's actually run on all servers. Maybe it should be called `registerLocalOnlySubscriptions`.

tanabarr: That isn't strictly accurate because in the function we register a subscription for an event forwarder which is not "LocalOnly" I will go with `registerNonLeaderSubscriptions` instead.

daosbuild3:
Test stage Unit Test completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/8/testReport/

src/control/lib/control/event.go
tanabarr: using it in tests inadvertently caught me out and we don't want unit tests raising syslog events

docs/admin/administration.md
tanabarr: done

docs/overview/fault.md
tanabarr: done

docs/overview/fault.md
```diff
@@ -89,23 +89,58 @@ and updating objects.
...
+membership state from transient engine failures without administrator
+intervention, improving overall system availability.
+
+#### Rate Limiting
+
+To prevent restart storms and ensure system stability, automatic engine restarts
+are rate-limited on a per-rank basis. By default, a minimum delay of 300 seconds
+(5 minutes) is enforced between consecutive restart attempts for the same rank.
+If an engine self-terminates again before this delay expires, the restart request
+is rejected and logged at NOTICE level.
```
kjacque: Will it retry after the time has passed? I.e. queueing the retries.

src/control/server/config/server.go
```diff
@@ -1,6 +1,6 @@
 //
-// (C) Copyright 2020-2024 Intel Corporation.
-// (C) Copyright 2025-2026 Hewlett Packard Enterprise Development LP
+// Copyright 2020-2024 Intel Corporation.
+// Copyright 2025-2026 Hewlett Packard Enterprise Development LP
```
kjacque: Probably going to have to revert this. I think the commit that updated the copyright githook was reverted.

src/control/server/server_utils.go
```diff
@@ -769,10 +774,16 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
...
-func handleEngineSuicide(ctx context.Context, srv *server, evt *events.RASEvent) error {
+// Handle local engine self termination and restart engine to rejoin system.
+func handleEngineSelfTerminated(ctx context.Context, srv *server, evt *events.RASEvent) error {
+
+	srv.log.Infof("handling engine self termination")
 
-	srv.log.Infof("handling suicide")
+	// Check if automatic restart is disabled
+	if srv.cfg.DisableEngineAutoRestart {
+		srv.log.Infof("automatic engine restart disabled by configuration")
```
kjacque: I could see this getting spammy. Maybe better to log at debug or trace level?

src/control/go.mod
```diff
@@ -5,7 +5,6 @@ module github.com/daos-stack/daos/src/control
 // - debian packaging version checks: debian/control
 // Scons uses this file to extract the minimum version.
 go 1.21
-toolchain go1.23.0
```
kjacque: Please revert this go.mod change

src/control/server/server_utils_test.go
```diff
@@ -2206,7 +2228,131 @@ func TestServer_handleEngineSuicide(t *testing.T) {
...
+	// Verify rate-limit log message
+	logOutput := buf.String()
+	if !strings.Contains(logOutput, "rate-limited") {
+		t.Errorf("expected log to contain 'rate-limited', got: %s", logOutput)
+	}
+
+	// Wait for the EngineAutoRestartMinDelay to expire
+	time.Sleep(time.Duration(srv.cfg.EngineAutoRestartMinDelay) * time.Second)
+
+	// Third restart after delay should succeed
```
kjacque: If the engine isn't restarted, why would we keep getting the event? It will self-terminate and then never try again. I think we will need to queue the restarts if they hit the rate limit.

kjacque (changes requested):
Thanks for cleaning up the terminology.

src/control/lib/control/event.go
kjacque: Okay, that's fair.

docs/overview/fault.md
tanabarr: not currently, it will just wait for the next RAS

src/control/server/config/server.go
tanabarr: thanks for the heads up

src/control/server/server_utils_test.go
tanabarr: so every time a RAS event is received for a given rank mark that rank as to be restarted and trigger restart as soon as timeout expires? it will only ever be a "queue" of one, right?

src/control/server/config/server.go
tanabarr: done

src/control/server/server_utils.go
tanabarr: my thinking was that administrator would want to know this and it will get lost in debug but I can change it

src/control/go.mod
tanabarr: done

src/control/server/server_utils_test.go
tanabarr: We wouldn't get an event if the engine isn't restarted, you are correct so that is a bug in the test but do we really want to restart if an event is received quickly after restarting? I would have thought that in that case we don't want to restart automatically and keep doing it every five minutes for a scenario where the engine can communicate but keeps being excluded from the group map. I will make the change but I'm on the fence about the behaviour choice.

src/control/server/server_utils.go
tanabarr: done

src/control/server/server_utils_test.go
kjacque: I agree if the engine restarts and then dies right away, it's probably better to leave it dead. I think we should raise a RAS event in that case though so the admin knows something is wrong.

tanabarr: How do we detect that specific case vs generic engine died? I presume you are suggesting a specific RAS event for this particular case.

kjacque: The control plane knows when this happens, right? In the logic handling the self-terminate RAS event, it can raise a new event if the engine has self-terminated before the cooldown timer runs out.

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16279/13/display/redirect

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16279/14/execution/node/421/log

src/control/server/server_utils_test.go
tanabarr: as discussed, rate-limiting has been implemented so a new event won't be added

src/control/server/server_utils_test.go
```diff
@@ -2312,22 +2314,44 @@ func TestServer_handleEngineSelfTerminated_RateLimiting(t *testing.T) {
...
 	}
 
-	// Wait for the EngineAutoRestartMinDelay to expire
+	// Verify pending timer exists
+	srv.rankRestartMu.Lock()
+	if _, exists := srv.rankRestartPending[evt.Rank]; !exists {
+		srv.rankRestartMu.Unlock()
+		t.Fatal("expected pending restart timer to exist")
+	}
+	srv.rankRestartMu.Unlock()
```
kjacque, line 2335: minor - might be better to do this in a function/closure so you can `defer srv.rankRestartMu.Unlock()` instead.

src/control/server/config/server.go
```diff
@@ -98,7 +98,9 @@ type Server struct {
 	Path string `yaml:"-"` // path to config file
 
 	// Behavior flags
-	AutoFormat bool `yaml:"-"`
+	AutoFormat                bool `yaml:"-"`
+	DisableEngineAutoRestart  bool `yaml:"disable_engine_auto_restart"`
+	EngineAutoRestartMinDelay int  `yaml:"engine_auto_restart_min_delay,omitempty"`
```
mjmac, line 103: There doesn't appear to be any validation of this parameter, e.g. in the `Validate()` method:

```go
       if cfg.EngineAutoRestartMinDelay < 0 {
               return errors.Errorf("engine_auto_restart_min_delay must be >= 0 (got %d)",
                       cfg.EngineAutoRestartMinDelay)
       }
```

src/control/server/server_utils.go
```diff
@@ -769,14 +774,126 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
...
 }
 
-// registerFollowerSubscriptions stops handling received forwarded (in addition
-// to local) events and starts forwarding events to the new MS leader.
-// Log events on the host that they were raised (and first published) on.
-// This is the initial behavior before leadership has been determined.
-func registerFollowerSubscriptions(srv *server) {
+// Handle local engine self termination and restart engine to rejoin system.
+func handleEngineSelfTerminated(ctx context.Context, srv *server, evt *events.RASEvent) error {
+
```
mjmac, line 779: Maybe add a check to guard against processing a non-local event that got here somehow?

```suggestion
       if evt.IsForwarded() || (evt.Hostname != "" && evt.Hostname != srv.hostname) {
               srv.log.Debugf("ignoring non-local engine_self_terminated event from %q", evt.Hostname)
               return
       }
```

docs/overview/fault.md
```diff
@@ -84,3 +84,68 @@ can now read from the rebuilt object shards.
 
 This rebuild process is executed online while applications continue accessing
 and updating objects.
+
+### Engine Self-Termination and Automatic Restart
+
+A DAOS engine may be excluded from the group map because of inactivity
+for example. When an engine becomes aware of it's removal from the
+group map it will self-terminate to protect data integrity and system stability.
```
liw: Not sure if "inactivity" is the right word. How about "difficulties in communicating with other engines" or something along that line? Then "when the engine becomes aware of its exclusion after the difficulties disappear, ...".

Do our user-facing docs and `dmg` help messages already present the concept of "group map"? If not, I wonder whether we should just say "excluded from the system"?

src/control/server/server_utils.go
```diff
@@ -769,14 +774,126 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to find instance for rank %d", evt.Rank)
+	}
+
+	if len(instances) == 0 {
+		return errors.Errorf("no instance found for rank %d", evt.Rank)
+	}
+	if len(instances) > 1 {
+		return errors.Errorf("multiple instances found for rank %d", evt.Rank)
+	}
```
kjacque, line 804: Might be inclined not to check this one -- or just log an error instead of failing hard here.

src/control/server/server_utils.go
```diff
@@ -769,14 +774,126 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
...
+					evt.Rank, engine.Index())
+
+				// Wait until engine is stopped
+				pollFn := func(e Engine) bool { return !e.IsStarted() }
+				if err := pollInstanceState(ctx, instances, pollFn); err != nil {
+					srv.log.Errorf("rank %d (instance %d) did not stop before deferred restart",
+						evt.Rank, engine.Index())
+					srv.rankRestartMu.Lock()
+					delete(srv.rankRestartPending, evt.Rank)
+					srv.rankRestartMu.Unlock()
```
kjacque, line 843: I wonder if it would be better to delegate the stuff that needs to be wrapped in the lock to a helper function. Then you could use `defer srv.rankRestartMu.Unlock()` within that scope, and not have to worry about unlocking before the various returns here.

kjacque (dismissed):
Comments are minor and not something I'd block on. I do think using `defer` makes it a lot easier to be sure we're handling mutexes safely.

tanabarr:
@kjacque I will address the nonblocking comments if I have to repush. @mjmac would you be able to review as I can't seem to find anyone else comfortable with reviewing? TIA

src/control/server/server_utils.go
```diff
@@ -769,14 +774,126 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
...
+		return errors.Wrapf(err, "failed to find instance for rank %d", evt.Rank)
+	}
+
+	if len(instances) == 0 {
+		return errors.Errorf("no instance found for rank %d", evt.Rank)
+	}
+	if len(instances) > 1 {
+		return errors.Errorf("multiple instances found for rank %d", evt.Rank)
+	}
+	engine := instances[0]
```
mjmac, line 805: nit: The variable `engine` shadows the imported package name. Might be safer to use `ei` like other areas of the code. I seem to recall that we had subtle problems caused by this in the past.

src/control/server/server.go
```diff
@@ -167,6 +167,10 @@ type server struct {
 	cbLock           sync.Mutex
 	onEnginesStarted []func(context.Context) error
 	onShutdown       []func()
+
+	rankRestartMu      sync.Mutex
+	rankRestartTimes   map[uint32]time.Time
+	rankRestartPending map[uint32]*time.Timer
```
mjmac, line 173: nit: Hmm. I know that `Event.Rank` is `uint32`, but maybe it should have been `ranklist.Rank`? We've been down this road before, where we used bare `uint32` to represent a rank because otherwise it meant type conversion, but IMO short-term convenience leads to longer-term decreases in code readability and quality. My preference would be to avoid `uint32` for rank in the control plane as much as possible.

src/control/server/server_utils.go
```diff
@@ -769,14 +774,126 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
...
+// the MS leader. Received events are logged on the host that they were raised (and first published)
+// on. This is the initial behavior for all servers and only changes when leadership has been
+// determined (when we change subscribers via registerLeaderSubscriptions). A handler is subscribed
+// for local engine self-termination events.
+func registerSubscriptions(srv *server) {
 	srv.pubSub.Reset()
 	srv.pubSub.Subscribe(events.RASTypeAny, srv.evtLogger)
 	srv.pubSub.Subscribe(events.RASTypeStateChange, srv.evtForwarder)
+	srv.pubSub.Subscribe(events.RASTypeInfoOnly,
+		events.HandlerFunc(func(ctx context.Context, evt *events.RASEvent) {
```
mjmac, line 889: Maybe this should be extracted into a helper used for both leader/follower paths? Seems like the code is identical in both places.

src/control/server/server_utils.go
```diff
@@ -769,14 +774,126 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
 	})
 }
 
-// registerFollowerSubscriptions stops handling received forwarded (in addition
-// to local) events and starts forwarding events to the new MS leader.
-// Log events on the host that they were raised (and first published) on.
-// This is the initial behavior before leadership has been determined.
-func registerFollowerSubscriptions(srv *server) {
+// Handle local engine self termination and restart engine to rejoin system.
+func handleEngineSelfTerminated(ctx context.Context, srv *server, evt *events.RASEvent) error {
```
mjmac, line 778: This function is really doing a lot, and it's quite difficult to reason about its behavior over time and in a concurrent environment. I strongly suspect that there is at least once race involving execution of the callbacks after the timer. The lock protects against multiple invocations of the handler, but I don't think it correctly protects against a handler making a decision based on a timestamp that's about to be updated by a callback (because the lock is only taken in the callback when it updates the timestamp).

I'm not 100% confident that it's wrong, but I'm also really not confident that it's right. I think it would be a lot cleaner to use a lockless approach and instead have the restart logic sitting out in its own thing that listens on a channel. The main logic can make restart requests all day long, but the restart manager only queues a new request if there isn't one already in progress. Aside from being easier to read and reason about, formalizing the restart logic as a concrete unit of code should also make it much easier to test thoroughly.

kjacque: This is a good idea.

src/control/server/server_utils.go
tanabarr: happy to add but should not be needed because these messages are not forwarded

src/control/server/server_utils.go
tanabarr: done

src/control/server/server.go
tanabarr: done

src/control/server/server_utils.go
tanabarr: done

src/control/server/server_utils.go
tanabarr: done

src/control/server/config/server.go
tanabarr: done

src/control/server/server_utils_test.go
tanabarr: done

src/control/server/server_utils.go
tanabarr: why would you ever want to have multiple instances returned for the same rank on a host? surely something is seriously broken in that case

src/control/server/server_utils.go
tanabarr: done

src/control/server/server_utils.go
tanabarr: I will do that as a separate Commit from the rest of the review comment changes

src/control/server/server_utils.go
kjacque: Yeah it would definitely be a sign of something very strange, but I'm wondering how we'd get into such a situation to begin with. Doesn't hurt to catch it I guess. I know sometimes we can be a little too stringent about vetting what's returned, making it so the user is basically stuck with no information, but that's more relevant to dmg and the daos tool.

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/19/testReport/

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/20/testReport/

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/22/testReport/

tanabarr:
CI failing on one dfuse known issue (seeing on other PRs)

tanabarr:
@mjmac @kjacque please could I have rereviews, I've refactored as per the review comments

tanabarr:
@mjmac @kjacque ping for reviews please

daosbuild3:
Test stage Unit Test completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/23/testReport/

daosbuild3:
Test stage Unit Test completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/25/testReport/

src/control/server/ctl_ranks_rpc.go
```diff
@@ -206,6 +206,22 @@ func (svc *ControlService) StopRanks(ctx context.Context, req *ctlpb.RanksReq) (
...
+		for _, ei := range instances {
+			rank, err := ei.GetRank()
+			if err == nil {
+				ranks = append(ranks, rank)
+			}
+		}
+		if len(ranks) > 0 {
+			svc.restartMgr.clearRankRestartHistory(ranks)
+		}
+	}
```
kjacque: minor - I'd be inclined to break this out into a helper function to reduce duplication. Would also make this logic easy to test.

src/control/server/server.go
```diff
@@ -167,6 +168,13 @@ type server struct {
 	cbLock           sync.Mutex
 	onEnginesStarted []func(context.Context) error
 	onShutdown       []func()
+
+	restartMgr *engineRestartManager
+
+	// Deprecated: use restartMgr instead
+	rankRestartMu      sync.Mutex
+	rankRestartTimes   map[ranklist.Rank]time.Time
+	rankRestartPending map[ranklist.Rank]*time.Timer
```
kjacque: Shouldn't this be removed from the PR?

src/control/server/server.go
```diff
@@ -283,6 +295,11 @@ func (srv *server) OnShutdown(fns ...func()) {
 }
 
 func (srv *server) shutdown() {
+	// Stop the restart manager first
+	if srv.restartMgr != nil {
```
kjacque: Is it allowed to be nil? Seems better to initialize it in the constructor, and then assume that it's initialized correctly so you don't have to do all these nil checks.

src/control/server/server_utils.go
```diff
@@ -769,14 +769,82 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
...
+
+	srv.log.Noticef("%s was notified @ %s of rank %d:%d (instance %d) self terminated", ts,
+		evt.Hostname, evt.Rank, evt.Incarnation, ei.Index())
+
+	rank := ranklist.Rank(evt.Rank)
+
+	if srv.restartMgr == nil {
+		return errors.Errorf("restart manager not initialized, cannot restart rank %d",
+			rank)
+	}
```
kjacque: Same question here about the nil checks... Maybe restartMgr should check if it's nil and return errors from its methods if so.

src/control/server/server_utils.go
```diff
@@ -769,14 +769,82 @@ func registerTelemetryCallbacks(ctx context.Context, srv *server) {
...
+	srv.restartMgr.requestRestart(rank, ei)
+
+	return nil
+}
+
+// subscribeEngineSelfTerminated creates a handler for engine self-termination events.
+func subscribeEngineSelfTerminated(srv *server) events.Handler {
+	return events.HandlerFunc(func(ctx context.Context, evt *events.RASEvent) {
+		switch evt.ID {
+		case events.RASEngineSelfTerminated:
```
kjacque: Any reason to use a `switch` here if there's only one allowed type?

src/control/server/instance_restart.go
```diff
@@ -0,0 +1,246 @@
...
+		mgr.log.Errorf("rank %d did not stop before restart: %s", rank, err)
+		return
+	}
+
+	mgr.log.Noticef("restarting rank %d", rank)
+	instance.requestStart(ctx)
+
+	// Record restart time and clear pending state on exit (deferred)
+	mgr.recordRestartTime(rank)
+	mgr.log.Noticef("recording rank %d", rank)
```
kjacque: Maybe shouldn't be Notice for this one. Seems like this is mostly useful for debug.

src/control/server/instance_restart.go
```diff
@@ -0,0 +1,246 @@
...
+	defer mgr.mu.Unlock()
+
+	// Cancel all pending restart timers
+	for rank, timer := range mgr.pendingRestart {
+		timer.Stop()
+		mgr.log.Debugf("cancelled pending restart for rank %d", rank)
+	}
+	mgr.pendingRestart = make(map[ranklist.Rank]*time.Timer)
+
+	close(mgr.stopChan)
```
kjacque, line 175: `start` should re-create the channel. Otherwise `stop` won't work again after closing it.

src/control/server/instance_restart_test.go
```diff
@@ -0,0 +1,748 @@
...
+
+func setupTestLogger(t *testing.T) (logging.Logger, *logging.LogBuffer) {
+	t.Helper()
+	log, buf := logging.NewTestLogger(t.Name())
+	t.Cleanup(func() {
+		test.ShowBufferOnFailure(t, buf)
+	})
+
+	return log, buf
+}
```
kjacque, line 31: Don't want to use `test.MustLogContext`/`logging.FromContext` to create the logger?

src/control/server/instance_restart_test.go
```diff
@@ -0,0 +1,748 @@
...
+	mgr.mu.RLock()
+	timer, exists := mgr.pendingRestart[testRank]
+	mgr.mu.RUnlock()
+
+	if !exists {
+		t.Fatal("expected pending restart timer to be created")
+	}
+
+	// Wait for timer to fire (with buffer)
+	time.Sleep(10 * time.Second)
```
kjacque: Probably don't need to wait this long if the timer is firing around 2 seconds. Maybe 3-5 sec is more reasonable?

src/control/server/instance_restart_test.go
```diff
@@ -0,0 +1,748 @@
...
+	mgr.mu.RUnlock()
+
+	if stillPending {
+		t.Error("expected pending restart to be cleared after execution")
+	}
+
+	// Cleanup
+	if timer != nil {
+		timer.Stop()
+	}
```
kjacque, line 411: For timers that need to stop at the end, you could use `defer` after you get them.

src/control/server/instance_restart_test.go
```diff
@@ -0,0 +1,748 @@
...
+		timer2.Stop()
+	}
+	mgr.clearPendingRestart(testRank)
+}
+
+func TestServer_NewEngineRestartManager(t *testing.T) {
+	cfg := &config.Server{
+		EngineAutoRestartMinDelay: 42,
+	}
+	mgr := setupTestManager(t, cfg)
```
kjacque: Should probably call `NewEngineRestartManager` directly here to verify only that behavior, not the test constructor.

src/control/server/server_utils_test.go
```diff
@@ -1981,3 +1988,753 @@ f0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
...
+				if !strings.Contains(logOutput, expStr) {
+					t.Errorf("expected log to contain %q, but it didn't\nLog:\n%s",
+						expStr, logOutput)
+				}
+			}
+		})
+	}
+}
+
+func TestServer_handleEngineSelfTerminated_RateLimiting(t *testing.T) {
```
kjacque, line 2274: Since the detailed behavior is now covered in the engineRestartManager, you can remove some of these higher level tests. You could just make sure the request is sent to the restart manager.

kjacque (changes requested):
First of all: nice work, and excellent job with all the test coverage. My comments are mostly minor cleanup stuff, but the stop/start behavior is something we may want to make behave better, or at least note that we only get to stop it once.

daosbuild3:
Test stage Functional on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16279/25/execution/node/983/log

src/control/server/ctl_ranks_rpc.go
tanabarr: done

src/control/server/server.go
tanabarr: done

src/control/server/server.go
tanabarr: done

src/control/server/server_utils.go
tanabarr: done

src/control/server/instance_restart.go
tanabarr: done

src/control/server/server_utils.go
tanabarr: removed the nil checks and will rely on constructor

src/control/server/instance_restart.go
tanabarr: the design is to only start once on service start and stop on  shutdown so this isn't necessary for practical purposes, I will add the capability anyway as it's nice to be re-entrant

src/control/server/instance_restart_test.go
tanabarr: the problem is that this Pattern doesn't enable me to expose the LogBuffer to be used in the tests for comparison purposes. maybe I'm missing a trick here? @mjmac

src/control/server/instance_restart_test.go
tanabarr: in this case we don't know whether the timer still exists at end of func, could do the nil check in defer func but I don't think that buys us much

src/control/server/instance_restart_test.go
tanabarr: done

src/control/server/server_utils_test.go
tanabarr: I'm going to leave them in for the moment unless you feel particularly strongly

src/control/server/ctl_ranks_rpc.go
```diff
@@ -206,21 +221,9 @@ func (svc *ControlService) StopRanks(ctx context.Context, req *ctlpb.RanksReq) (
...
-			rank, err := ei.GetRank()
-			if err == nil {
-				ranks = append(ranks, rank)
-			}
-		}
-		if len(ranks) > 0 {
-			svc.restartMgr.clearRankRestartHistory(ranks)
-		}
-	}
+	// clearly state history for stopped ranks, instances have already been filtered by
```
kjacque: typo? clearly -> clear

src/control/server/ctl_ranks_rpc.go
```diff
@@ -335,21 +338,9 @@ func (svc *ControlService) StartRanks(ctx context.Context, req *ctlpb.RanksReq)
...
-			rank, err := ei.GetRank()
-			if err == nil {
-				ranks = append(ranks, rank)
-			}
-		}
-		if len(ranks) > 0 {
-			svc.restartMgr.clearRankRestartHistory(ranks)
-		}
-	}
+	// clearly state history for started ranks, instances have already been filtered by
```
kjacque: typo? clearly -> clear

src/control/server/instance_restart_test.go
kjacque: Ah, I see! Never mind, I'm not sure how to extract the buffer in that case either. Might be nice to add this function to the `test` package. I suspect this will be useful in other tests (maybe call it `NewTestLoggerWithBuffer` or something like that?). Not requesting a change on this PR, just a suggestion.

src/control/server/instance_restart_test.go
kjacque: Looks like the test asserts if the timer wasn't set in the map... So it would have to be non-nil at this point, wouldn't it? Doesn't look like we're ever resetting that variable.

src/control/server/server_utils_test.go
kjacque: I don't feel that strongly, fine with leaving them.

src/control/server/ctl_ranks_rpc_test.go
```diff
@@ -76,17 +76,15 @@ func checkUnorderedRankResults(t *testing.T, expResults, gotResults []*sharedpb.
...
+	ei._superblock.Rank = ranklist.NewRankPtr(rank)
+
 	trc := &engine.TestRunnerConfig{}
 	if len(stopped) == 0 || !stopped[0] {
 		trc.Running.SetTrue()
-		srv.ready.SetTrue()
+		ei.ready.SetTrue()
 	}
-	srv.runner = engine.NewTestRunner(trc, engine.MockConfig())
-	srv.setIndex(idx)
```
kjacque, line 86: I guess no tests were using the index?

src/control/server/instance_restart.go
kjacque: As discussed, I agree it's fine as-is. I do think it'd be nice to leave a note to ourselves as a comment though.

src/control/server/instance_restart.go
kjacque: Oops, looks like this fix didn't get pushed.

kjacque (dismissed):
Nothing left that's blocking, just the minor typos and the Notice level logging should be fixed. I'm fine with that happening in a follow-on.

daosbuild3:
Test stage Functional on EL 9 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/26/testReport/

src/control/server/ctl_ranks_rpc_test.go
tanabarr: nope, so I removed it

src/control/server/ctl_ranks_rpc.go
tanabarr: done

src/control/server/ctl_ranks_rpc.go
tanabarr: done

src/control/server/server_utils.go
tanabarr: done

src/control/server/instance_restart.go
tanabarr: got inadvertently reverted in bd15522c203539d71553a25b25c6684f10f95067 , added back, done

src/control/server/instance_restart.go
tanabarr: comments added

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+        """Clean up after each test method."""
+        # Reset restart state for next test method
+        # This ensures clean state between sequential tests
+        try:
+            self.reset_engine_restart_state()
+        except Exception as error:
+            self.log.error("Failed to reset engine restart state: %s", error)
+            self.fail("tearDown failed to reset engine restart state: {}".format(error))
+        finally:
+            super().tearDown()
```
daltonbohning: We have a way to handle this in the framework by calling
```
self.register_cleanup(reset_engine_restart_state)
```
After whatever operation puts the system into the invalid state. So maybe instead of defining this tearDown, we can do that after calling `exclude_rank_and_wait_restart` the first time

src/control/server/instance_restart_test.go
tanabarr: have changed to improve clarity, done

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def test_auto_restart_basic(self):
+        """Test basic automatic engine restart after self-termination.
+
+        Test Description:
+            1. Exclude a rank from the system
+            2. Wait for rank to self-terminate
+            3. Verify rank automatically restarts and rejoins the system
+
+        :avocado: tags=all,pr,daily_regression
```
daltonbohning, line 36: Just confirming that we really do want to add this to PR testing?

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+        :avocado: tags=hw,medium
+        :avocado: tags=dmg,control,engine_auto_restart
+        :avocado: tags=EngineAutoRestartTest,test_auto_restart_basic
+        """
+        all_ranks = self.get_all_ranks()
+        if len(all_ranks) < 2:
+            self.skipTest("Test requires at least 2 ranks")
```
daltonbohning: It is better to fail because skipping will be silent and easily ignored
```suggestion
            self.fail("Test requires at least 2 ranks")
```

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+        # get initial incarnation number
+        initial_incarnation = self.get_rank_incarnation(test_rank)
+        if initial_incarnation is None:
+            self.fail(f"failed to get initial incarnation for rank {test_rank}")
+
+        self.log.info("rank %s initial incarnation: %s", test_rank, initial_incarnation)
+
+        restarted, final_state = self.exclude_rank_and_wait_restart(test_rank)
+
+        if not restarted:
```
daltonbohning, line 55: nit - intentional new line?
```suggestion
        restarted, final_state = self.exclude_rank_and_wait_restart(test_rank)
        if not restarted:
```

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+            self.fail(f"rank {test_rank} did not automatically restart. "
+                      f"final state: {final_state}")
+
+        # verify incarnation increased after restart
+        final_incarnation = self.get_rank_incarnation(test_rank)
+        if final_incarnation is None:
+            self.fail(f"failed to get final incarnation for rank {test_rank}")
```
daltonbohning: It would be better if get_rank_incarnation raised an exception instead of silently returning None

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+            3. Verify all automatically restart and rejoin
+
+        :avocado: tags=all,full_regression
+        :avocado: tags=hw,medium
+        :avocado: tags=dmg,control,engine_auto_restart
+        :avocado: tags=EngineAutoRestartTest,test_auto_restart_multiple_ranks
+        """
+        all_ranks = self.get_all_ranks()
+        if len(all_ranks) < 3:
+            self.skipTest("Test requires at least 3 ranks")
```
daltonbohning:
```suggestion
            self.fail("Test requires at least 3 ranks")
```

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+
+        # Exclude half the ranks
+        num_to_test = max(2, len(all_ranks) // 2)
+        test_ranks = self.random.sample(all_ranks, num_to_test)
+
+        self.log_step("Step 1: Excluding %s ranks: %s", (num_to_test, test_ranks))
```
daltonbohning: IMO it's better to not put `Step` in the string when using `log_step` because `log_step` already adds step numbers. Also, `log_step` does not support the `%s` syntax, which would raise an exception if this test was running in CI
```suggestion
        self.log_step(f"Exclude {num_to_test} ranks: {test_ranks}")
```

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+
+        incs = []
+        for rank in test_ranks:
+            initial_incarnation = self.get_rank_incarnation(rank)
+            if initial_incarnation is None:
+                self.fail(f"failed to get initial incarnation for rank {rank}")
+            incs.append(initial_incarnation)
+            self.dmg.system_exclude(ranks=[rank], rank_hosts=None)
+            time.sleep(1)  # small delay between exclusions
+            self.dmg.system_clear_exclude(ranks=[rank], rank_hosts=None)
```
daltonbohning, line 101: Where does the arbitrary 1s sleep come from? Historically, this kind of thing eventually fails and needs to be adjusted

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+        wait_time = 35
+
+        self.log_step("Step 3: Waiting %ss to verify all automatically restart", wait_time)
+        time.sleep(wait_time)
+
+        errors = []
+        end_incs = []
+        for rank in test_ranks:
+            failed = self.server_managers[0].check_rank_state(
+                ranks=[rank], valid_states=["joined"], max_checks=1)
```
daltonbohning: Instead of arbitrarily sleeping for 35 seconds, why not increase `check_rank_state(..., max_checks=1)` to something higher?

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+            end_incs.append(end_incarnation)
+
+        if errors:
+            self.fail("\n".join(errors))
+
+        # Show changes
+        for idx, (old, new) in enumerate(zip(incs, end_incs)):
+            actual_rank = test_ranks[idx]
+            if new > old:
+                self.log.debug(f"Rank {actual_rank}: {old} -> {new} (restarted)")
```
daltonbohning: > logging-fstring-interpolation, Use lazy % formatting in logging functions

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+        :avocado: tags=hw,medium
+        :avocado: tags=dmg,control,engine_auto_restart,pool
+        :avocado: tags=EngineAutoRestartTest,test_auto_restart_with_pool
+        """
+        all_ranks = self.get_all_ranks()
+        if len(all_ranks) < 4:
+            self.skipTest("Test requires at least 4 ranks")
+
+        # Create pool first
+        self.add_pool(connect=False)
```
daltonbohning: We're trying to not use that function for new tests. Instead:

```suggestion
        pool = self.get_pool(connect=False)
```

src/tests/ftest/control/engine_auto_restart.py
```diff
@@ -0,0 +1,201 @@
...
+        if final_incarnation is None:
+            self.fail(f"Failed to get final incarnation for rank {test_rank}")
+
+        if final_incarnation <= initial_incarnation:
+            self.fail(f"Rank {test_rank} incarnation did not increase. "
+                      f"Before: {initial_incarnation}, After: {final_incarnation}")
+
+        # Verify pool is still accessible
+        self.log_step("Verifying pool is still accessible after rank restart")
+        self.pool.query()
```
daltonbohning: And then use the local `pool` variable here
```suggestion
        pool.query()
```

src/tests/ftest/control/engine_auto_restart.yaml
```diff
@@ -0,0 +1,25 @@
+hosts:
+  test_servers: 2
+server_config:
```
daltonbohning, line 4: We typically put the timeout up here
```suggestion
hosts:
  test_servers: 2
timeout: 300
server_config:
```

src/tests/ftest/control/engine_auto_restart.yaml
```diff
@@ -0,0 +1,25 @@
...
+      log_file: daos_server1.log
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos1
+pool:
+  size: 2G
+timeout: 300
```
daltonbohning:
```suggestion
```

src/tests/ftest/util/control_test_base.py
```diff
@@ -46,3 +48,176 @@ def verify_dmg_storage_scan(self, verify_method):
...
+
+        Args:
+            rank (int): Rank number
+
+        Returns:
+            str: Current state of the rank
+        """
+        data = self.dmg.system_query(ranks="%s" % rank)
+        if data["status"] != 0:
+            self.fail("Cmd dmg system query failed")
```
daltonbohning: It's better for a helper function to raise an exception and let the test itself decide if it is a failure
```suggestion
        if data["status"] != 0:
            raise CommandFailure("dmg system query failed")
```

src/tests/ftest/util/control_test_base.py
```diff
@@ -46,3 +48,176 @@ def verify_dmg_storage_scan(self, verify_method):
...
+            int: Current incarnation number of the rank, or None if not found
+
+        Raises:
+            None - logs error and returns None on failure
+        """
+        try:
+            data = self.dmg.system_query(ranks=f"{rank}")
+            if data.get("status") != 0:
+                self.log.error("dmg system query failed for rank %s", rank)
+                return None
```
daltonbohning: IMO it's better to raise exception than silently skip
```suggestion
            if data.get("status") != 0:
                raise CommandFailure(f"dmg system query failed for rank {rank}")
```

src/tests/ftest/util/control_test_base.py
```diff
@@ -46,3 +48,176 @@ def verify_dmg_storage_scan(self, verify_method):
...
+                    self.log.error("No incarnation field for rank %s", rank)
+                    return None
+
+            self.log.error("Rank %s not found in system query response", rank)
+            return None
+
+        except Exception as error:  # pylint: disable=broad-exception-caught
+            # Catch all exceptions to prevent test framework crashes during rank queries
+            self.log.error("Exception getting incarnation for rank %s: %s", rank, error)
+            return None
```
daltonbohning: This comment is indicative of a larger problem. An unhandled exception should, IMO, cause the test to ultimately fail. We should not silently skip something that is broken

src/tests/ftest/util/control_test_base.py
```diff
@@ -46,3 +48,176 @@ def verify_dmg_storage_scan(self, verify_method):
...
+            Exception: If server stop/start fails or ranks fail to rejoin
+
+        Note:
+            This operation adds ~5-10 seconds per test due to server restart overhead,
+            but is necessary to ensure test isolation and reliable results.
+        """
+        self.log.info("Restarting servers to reset engine restart manager state")
+        self.server_managers[0].system_stop()
+        time.sleep(2)
+        self.server_managers[0].system_start()
```
daltonbohning: Where does the arbitrary 2s sleep come from? This will eventually be a problem and we will have to revisit

daltonbohning (commented):
It seems the new functional tests aren't even running in CI?
There are many lint errors in ftest that need to be resolved or possibly ignored

daosbuild3:
Test stage Functional on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16279/28/execution/node/942/log

src/tests/ftest/control/engine_auto_restart.py
tanabarr: So yes I think so it's a significant feature

src/tests/ftest/control/engine_auto_restart.py
tanabarr: done

src/tests/ftest/control/engine_auto_restart.py
tanabarr: done

src/tests/ftest/control/engine_auto_restart.py
tanabarr: done

src/tests/ftest/control/engine_auto_restart.py
tanabarr: done

src/tests/ftest/control/engine_auto_restart.py
tanabarr: it's to enable the propagation of exclude state within the control-plane before we clear the exclusion, closes a potential race where state is cleared before propagated which would prevent self-determination of engine process.

src/tests/ftest/util/control_test_base.py
tanabarr: done

src/tests/ftest/util/control_test_base.py
tanabarr: will remove if you don't think it necessary for a system restart

tanabarr: done

src/tests/ftest/control/engine_auto_restart.py
tanabarr: done

src/tests/ftest/util/control_test_base.py
```diff
@@ -46,3 +48,170 @@ def verify_dmg_storage_scan(self, verify_method):
...
+    def exclude_rank_and_wait_restart(self, rank, timeout=30):
+        """Exclude a rank and wait for it to self-terminate and potentially restart.
+
+        Args:
+            rank (int): Rank to exclude
+            timeout (int): Maximum seconds to wait for restart
+
+        Returns:
+            tuple: (restarted, final_state) - whether rank restarted and its final state
+        """
```
tanabarr, line 94: moving register_cleanup() to beginning of each test because not all of them call this helper

src/tests/ftest/control/engine_auto_restart.py
tanabarr: done

src/tests/ftest/control/engine_auto_restart.py
tanabarr: done

src/tests/ftest/util/control_test_base.py
tanabarr: done

src/tests/ftest/util/control_test_base.py
tanabarr: fixed

src/tests/ftest/control/engine_auto_restart.py
tanabarr: fixed

src/tests/ftest/control/engine_auto_restart.yaml
tanabarr: done

src/tests/ftest/control/engine_auto_restart.yaml
tanabarr: done

src/tests/ftest/control/engine_auto_restart.py
daltonbohning: How do we know 1s is enough?

src/tests/ftest/util/control_test_base.py
daltonbohning: If we don't need it at all then that is great, but I was more meaning that 2s seems arbitrary. And historically arbitrary sleeps have been an issue that we have to revisit later

src/tests/ftest/control/engine_auto_restart.yaml
```diff
@@ -0,0 +1,25 @@
...
+  engines_per_host: 2
+  engines:
+    0:
+      log_file: daos_server0.log
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos0
```
daltonbohning: We should use `storage: auto` for HW tests so the framework can set this up for both pmem and md on ssd
```suggestion
      storage: auto
```

src/tests/ftest/control/engine_auto_restart.yaml
```diff
@@ -0,0 +1,25 @@
...
+    1:
+      log_file: daos_server1.log
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos1
```
daltonbohning:
```suggestion
      storage: auto
```

src/tests/ftest/control/engine_auto_restart_advanced.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if actual_delay < min_delay:
+            self.fail(f"Restart too early: {actual_delay:.1f}s < {min_delay:.1f}s")
+        elif actual_delay > max_delay:
+            self.log.warning("Restart delayed beyond expected: %.1fs > %.1fs "
+                             "(may be acceptable depending on system load)",
+                             actual_delay, max_delay)
+        else:
+            self.log.info("SUCCESS: Restart delay within expected range [%.1fs, %.1fs]",
+                          min_delay, max_delay)
```
daltonbohning, line 154: Do we really want to just log a warning if `actual_delay > max_delay`? For example what if the actual is like 500% more? I'd recommend making the max delay higher and then strictly enforcing that the actual delay is within bounds. This might eventually fail, but then we can revisit and increase the max again if needed. Otherwise, we might silently skip over an extremely high delay

src/tests/ftest/control/engine_auto_restart_advanced.yaml
```diff
@@ -0,0 +1,26 @@
...
+  engine_auto_restart_min_delay: 30
+  engines:
+    0:
+      log_file: daos_server0.log
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos0
```
daltonbohning:
```suggestion
      storage: auto
```

src/tests/ftest/control/engine_auto_restart_advanced.yaml
```diff
@@ -0,0 +1,26 @@
...
+    1:
+      log_file: daos_server1.log
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos1
```
daltonbohning:
```suggestion
      storage: auto
```

src/tests/ftest/control/engine_auto_restart_disabled.py
```diff
@@ -0,0 +1,152 @@
...
+        restarted, _ = self.exclude_rank_and_wait_restart(test_rank, timeout=35)
+
+        if restarted:
+            self.fail("Rank {test_rank} unexpectedly restarted when auto-restart disabled!")
+
+        self.log.info("Confirmed: Rank %s did NOT automatically restart (as expected)", test_rank)
+
+        # Manually start the rank
+        self.log_step("Manually starting rank {test_rank}")
+        self.dmg.system_start(ranks=f"{test_rank}")
```
daltonbohning: I think this should handle an int?
```suggestion
        self.dmg.system_start(ranks=test_rank)
```

src/tests/ftest/control/engine_auto_restart_disabled.py
```diff
@@ -0,0 +1,152 @@
...
+        if errors:
+            self.fail("\n".join(errors))
+
+        self.log.info("Confirmed: None of %s automatically restarted", test_ranks)
+
+        # Manually restart all
+        self.log_step("Manually restart ranks")
+
+        for rank in test_ranks:
+            self.dmg.system_start(ranks=f"{rank}")
```
daltonbohning:
```suggestion
            self.dmg.system_start(ranks=rank)
```

src/tests/ftest/control/engine_auto_restart_disabled.yaml
```diff
@@ -0,0 +1,26 @@
...
+  disable_engine_auto_restart: true
+  engines:
+    0:
+      log_file: daos_server0.log
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos0
```
daltonbohning:
```suggestion
      storage: auto
```

src/tests/ftest/control/engine_auto_restart_disabled.yaml
```diff
@@ -0,0 +1,26 @@
...
+    1:
+      log_file: daos_server1.log
+      targets: 4
+      nr_xs_helpers: 0
+      storage:
+        0:
+          class: ram
+          scm_mount: /mnt/daos1
```
daltonbohning:
```suggestion
      storage: auto
```

src/tests/ftest/util/control_test_base.py
```diff
@@ -46,3 +49,165 @@ def verify_dmg_storage_scan(self, verify_method):
...
+        self.log_step(f"Excluding rank {rank}")
+        self.dmg.system_exclude(ranks=[rank], rank_hosts=None)
+
+        # Wait for rank to self-terminate (should go to AdminExcluded state)
+        self.log_step(f"Waiting for rank {rank} to self-terminate")
+        time.sleep(2)
+
+        # Check if rank is adminexcluded
```
daltonbohning, line 102: Similar to some other cases: how do we know 2s is enough? If there really is not a deterministic way to know at this code level, this is fine, but eventually this kind of thing needs to be revisited.

daosbuild3:
Test stage Functional on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16279/33/display/redirect

src/tests/ftest/control/engine_auto_restart_advanced.py
tanabarr: ok, done

src/engine/drpc_ras.c
mjmac, line 375: Should `wait_for_resp` be `true` now, in order to ensure that the event processing doesn't race with the SIGKILL? I don't really see a downside...

src/control/server/instance_restart.go
```diff
@@ -0,0 +1,246 @@
...
+			mgr.performRestart(ctx, rank, instance)
+		})
+
+		// Overwrite any existing pending restart
+		mgr.setPendingRestart(rank, timer)
+		return
+	}
+
+	// Can restart immediately
+	mgr.performRestart(ctx, rank, instance)
```
mjmac: There is still a race here, I think. If multiple restart requests come in at the same time, there's a window where the manager goroutine and the AfterFunc goroutine could both wind up calling `performRestart()`.

If you rewrite this method to make the decision under the lock, I think you eliminate the race and also easily deal with floods of restart requests:
```go
      rank, instance := req.rank, req.instance

      mgr.mu.Lock()
      if last, ok := mgr.lastRestart[rank]; ok {
          if elapsed := time.Since(last); elapsed < mgr.getMinDelay() {
              // Fast debounce for subsequent requests inside of the delay window
              if _, pending := mgr.pendingRestart[rank]; pending {
                  mgr.mu.Unlock()
                  mgr.log.Debugf("rank %d already has a deferred restart pending; dropping",
                      rank)
                  return
              }

              // First restart request inside of the delay window claims it
              remaining := mgr.getMinDelay() - elapsed
              mgr.pendingRestart[rank] = time.AfterFunc(remaining, func() {
                  mgr.requestRestart(rank, instance)
              })
              mgr.mu.Unlock()
              mgr.log.Noticef("rank %d restart rate limited: will restart in %s",
                  rank, remaining.Round(time.Second))
              return
          }
      }
      
      // If this is the first restart or it's outside of the delay window, start the process (over)
      mgr.lastRestart[rank] = time.Now()
      delete(mgr.pendingRestart, rank)
      mgr.mu.Unlock()

      if err := waitForEngineStopped(ctx, []Engine{instance}); err != nil {
          mgr.log.Errorf("rank %d did not stop before restart: %s", rank, err)
          return
      }
      mgr.log.Noticef("restart manager is restarting rank %d", rank)
      instance.requestStart(ctx)
```

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16279/45/execution/node/727/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16279/46/execution/node/1364/log

daltonbohning (commented):
ftest LGTM

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/48/testReport/

src/control/server/instance_restart.go
tanabarr: would be acceptable to address this as a follow-on PR? or do you think it is a blocking issue?

kjacque: I'd be fine with doing it as a follow-on, if only because I don't think a flood of near-simultaneous requests for the same node are likely.

kjacque (dismissed):
Only looked at Go and C. I think the issues can be addressed in a follow-on, if @mjmac agrees.

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16279/49/testReport/

mjmac (approved):
LGTM, thanks for reworking it

src/control/server/instance_restart.go
```diff
@@ -52,108 +52,50 @@ func (mgr *engineRestartManager) getMinDelay() time.Duration {
...
-	instance := req.instance
+	rank, instance := req.rank, req.instance
 
-	mgr.log.Debugf("processing restart request for rank %d", rank)
+	mgr.mu.Lock()
+	if last, ok := mgr.lastRestart[rank]; ok {
+		if elapsed := time.Since(last); elapsed < mgr.getMinDelay() {
+			// Fast debounce for subsequent requests inside of the delay window
+			if _, pending := mgr.pendingRestart[rank]; pending {
+				mgr.mu.Unlock()
```
kjacque, line 70: suggestion - I'd be inclined to wrap this section happening under lock in a helper function so you can use `defer`.

kjacque (approved):
Suggestion is nonblocking. Thanks!

daltonbohning (commented):
ftest LGTM

src/control/server/instance_restart.go
tanabarr: addressed in this change

src/control/server/instance_restart.go
tanabarr: will rework when I revisit this code

PR #16284, opened 2025-04-18
DAOS-17422 chk: enhance dmg check repair API - b26

Nasf-Fan:
When -f option is specified for "dmg check repair" command, we allows the user/admin to handle the same type of inconsistencies with the same action subsequently. Under such scenario, using the inconsistency-class is more clear instead of the seq-num. Then the SYNOPSIS will be as following:

dmg [OPTIONS] check repair [repair-OPTIONS] <seq-num|inconsistency-class>
        <interact-opt|action>
...
[repair command options]
        -f, --for-all   Take the same action for all (potential) inconsistencies
                        with the same class. If this option is specified, then
                        "inconsistency-class" and "action" will be accepted for
                        subsequent parameters; otherwise, "seq-num" and "interact-opt"
                        will be used.

Enhance test logic to make it to be workable after landing c9745d82982.

Test-tag: test_daos_cat_recov_core

### Steps for the author:

### DIFF
--- a/src/chk/chk_common.c
+++ b/src/chk/chk_common.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2022-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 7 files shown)


Nasf-Fan:
Not need it any longer.

PR #16287, opened 2025-04-18
DAOS-17135 cart: not exclude self when create corpc for IV - b26

Nasf-Fan:
Otherwise, it may misguide subsequent crt_get_filtered_grp_rank_list() and regard the IV root to be as non-exist in the ranks list, then fail related IV operation with -DER_NONEXIST.

It maybe not a perfected solution for current cart IV logic. But to be some temporary option, it makes CR to be workable when some ranks dead.

Add new test case to verfiy such corner case.

### Steps for the author:


Nasf-Fan:
Replaced by https://github.com/daos-stack/daos/pull/16358

PR #16300, opened 2025-04-23
DAOS-17438 chk: per engine based check query result

Nasf-Fan:
Currently, querying the checker progress for a pool will return per target based result. For large scaled pool, the results may be huge as to overflow related dRPC. Be as some temporary solution, we will merge the per target based result on check engine and only transfer per engine based checker progress to the leader and then to control plane. That will much reduce the query buffer usage. In the future, when we supports arbitrary sized dRPC, then we can reply per target based query results as required.

Test-tag: test_daos_cat_recov_core

### Steps for the author:


Nasf-Fan:
Not need it any longer.

PR #16301, opened 2025-04-23
DAOS-17447 control: Selective telemetry registration

mjmac:
Update daos_agent configuration and logic to allow for
selective enablement and registration of client telemetry.

Adds two new daos_agent configuration parameters:
  * telemetry_enabled_procs: [regex]
  * telemetry_disabled_procs: [regex]

Setting one or the other of these will specify a pattern
against which connecting client process names will be
matched to provide more control over behavior. If neither
are set, then all clients will have telemetry enabled,
assuming that the value of telemetry_enabled is true.

Includes some improvements to the client telemetry init
logic to more clearly implement different use cases.

Change-Id: Iaa8b98b008de32e9d3b2c2b8d7583a5155fff9de


src/control/cmd/daos_agent/config.go
```diff
@@ -40,6 +43,81 @@ func (rm refreshMinutes) Duration() time.Duration {
...
+	return (*regexp.Regexp)(cr).MatchString(s)
+}
+
+// TelemetryConfig defines the agent telemetry configuration.
+type TelemetryConfig struct {
+	Port       int           `yaml:"telemetry_port,omitempty"`
+	Enabled    bool          `yaml:"telemetry_enabled,omitempty"`
+	Retain     time.Duration `yaml:"telemetry_retain,omitempty"`
+	RegPattern *ConfigRegexp `yaml:"telemetry_enable_pattern,omitempty"`
+	IgnPattern *ConfigRegexp `yaml:"telemetry_disable_pattern,omitempty"`
```
kjacque, line 87: Do these need to be documented in the daos_agent.yml file?

I wonder about the idea of using regexes. They're flexible, of course, but can be challenging to write/edit even in the best of times. I feel like the most common approach for these settings would likely be a list of literal process names or something like "procname[0-9]*".

I also feel maybe the naming needs to be adjusted. If I understand correctly, it's not really enable/disable so much as keep telemetry in the agent vs. not. Right?

src/control/cmd/daos_agent/mgmt_rpc_test.go
```diff
@@ -317,6 +283,48 @@ func TestAgent_mgmtModule_getAttachInfo(t *testing.T) {
...
+			},
+			expResp: func(resp *mgmtpb.GetAttachInfoResp) *mgmtpb.GetAttachInfoResp {
+				resp.ClientNetHint.EnvVars = append(resp.ClientNetHint.EnvVars,
+					fmt.Sprintf("%s=1", telemetry.ClientMetricsEnabledEnv),
+					fmt.Sprintf("%s=1", telemetry.ClientMetricsRegisterEnv),
+				)
+				return resp
+			}(respWith(testResp, "test1", "dev1", defNumaMap)),
+		},
+		"client telemetry enabled; self not in register pattern": {
```
kjacque, line 300: How about the case where the self is in the register pattern?

src/control/cmd/daos_agent/config.go
mjmac: > Do these need to be documented in the daos_agent.yml file?
> 
Yes. Forgot to do that.

> I wonder about the idea of using regexes. They're flexible, of course, but can be challenging to write/edit even in the best of times. I feel like the most common approach for these settings would likely be a list of literal process names or something like "procname[0-9]*".
> 
I'm a bit wary of overengineering this to begin with. The need for this feature in our environment is that we'd like to enable agent-managed client telemetry for the dfuse process, but nothing else. I take your point that the most common use cases will likely just have literal process names, but I can imagine a scenario where there is a desire to enable telemetry for a family of processes matching a common ... pattern. Sure, we could invent a simplified pattern matching scheme, but IMO regexps aren't that difficult to write and should be well within the capabilities of your typical storage admin.

> I also feel maybe the naming needs to be adjusted. If I understand correctly, it's not really enable/disable so much as keep telemetry in the agent vs. not. Right?

Funny you point that out... I started out with register/ignore as the verbs, but felt like those focused too much on the implementation vs the intent. I don't feel strongly about it either way.

kjacque: `managed_telemetry_enabled_procs`/`managed_telemetry_disabled_procs` maybe? Or something similar to indicate the difference between enabling/disabling telemetry in general.

src/control/cmd/daos_agent/mgmt_rpc_test.go
mjmac: Done.

src/control/cmd/daos_agent/config.go
mjmac: Renamed and documented.

kjacque (dismissed):
Updates LGTM.

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16301/6/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16301/7/execution/node/1519/log

tanabarr (dismissed):

daosbuild3:
Test stage Build RPM on EL 8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16301/12/execution/node/328/log

daosbuild3:
Test stage Build RPM on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16301/12/execution/node/325/log

daosbuild3:
Test stage Build RPM on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16301/12/execution/node/341/log

kjacque (approved):

src/client/api/metrics.c
```diff
@@ -158,6 +166,8 @@ dc_tm_init(crt_init_options_t *crt_info)
 out:
 	if (rc != 0) {
 		daos_client_metric = false;
+		daos_client_metric_retain = false;
+		crt_info->cio_use_sensors = 0;
```
knard38, line 170: Not sure to understand why this modification is needed.

mjmac: It's fixing a bug that was introduced in an earlier commit. We enable this setting here: https://github.com/daos-stack/daos/blob/40aab803b6253c76cea36bffd7172501ed3a8cf3/src/client/api/metrics.c#L145 and it should be disabled if something failed during metrics initialization in order to avoid annoying error messages in the client log.

PR #16314, opened 2025-04-25
DAOS-17443 gurt: Always retain first .old log file

frostedcmos:
- Upon log rotation, the very first '.old' file is saved now as '.oldest'

### Steps for the author:

### DIFF
--- a/src/gurt/dlog.c
+++ b/src/gurt/dlog.c
@@ -1,5 +1,6 @@
 /*
  * (C) Copyright 2016-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -377,6 +378,13 @@ log_rotate(void)
 	int rc = 0;
 
 	if (!mst.log_old) {
+		rc = asprintf(&mst.log_old, "%s.oldest", mst.log_file);
+		if (rc < 0) {
+			dlog_print_err(errno, "failed to alloc name\n");
+			return -1;
+		}
+	} else {
+		free(mst.log_old);
 		rc = asprintf(&mst.log_old, "%s.old", mst.log_file);
 		if (rc < 0) {
 			dlog_print_err(errno, "failed to alloc name\n");


daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16314/3/execution/node/1427/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16314/3/execution/node/1382/log

Michael-Hennecke (approved):
LGTM

frostedcmos:
> I don't really mind either way but what is the incentive for keeping the first one and not the other ones ? are we just assuming that when we start rotating logs there are already so many errors that only the first log is interesting ? why don't we add a timestamp at the end of the filename and always use that to save the old log ? (instead of having old, oldest etc)

This request was based on real world usages on different clusters, and what we observe is that once we hit some bad error the logs tend to fill up rather quickly with repeated stuff making current and previous log files contain similar info. So, in practice only first log ends up being useful to see errors leading to repeated flood.

As for timestamps - we dont want to save every log rotation, but only  [current] [current -1] and [first]

mchaarawi (dismissed):

soumagne:
that's fine but imo it looks a bit clunky to have .oldest and .old naming, if I look at other services running on my linux box I don't see anybody doing that kind of things :)

soumagne (dismissed):

mchaarawi:
> that's fine but imo it looks a bit clunky to have .oldest and .old naming, if I look at other services running on my linux box I don't see anybody doing that kind of things :)

maybe rename oldest to something like .first or .initial ?

frostedcmos:
> > that's fine but imo it looks a bit clunky to have .oldest and .old naming, if I look at other services running on my linux box I don't see anybody doing that kind of things :)
> 
> maybe rename oldest to something like .first or .initial ?

renamed to .first

mchaarawi (approved):
we should get a review from a test engineer to verify this is not going to break CI in the case:
1) .first is also collected for logs
2) 3 logs are still able to be collected and we do not run out of space since we have an extra log (this is the case when the logs do overflow).

daltonbohning:
> we should get a review from a test engineer to verify this is not going to break CI in the case:
> 
> 1. .first is also collected for logs
> 2. 3 logs are still able to be collected and we do not run out of space since we have an extra log (this is the case when the logs do overflow).

ftest archives all `*log*`. Space can be a concern, and I don't think there are hard mechanisms in place to prevent that. Logs are stored in `/var/tmp` (not sure how much space we have on all the various clusters). Luckily, when archived, any files larger than 1M are compressed.

daltonbohning:
> > we should get a review from a test engineer to verify this is not going to break CI in the case:
> > 
> > 1. .first is also collected for logs
> > 2. 3 logs are still able to be collected and we do not run out of space since we have an extra log (this is the case when the logs do overflow).
> 
> ftest archives all `*log*`. Space can be a concern, and I don't think there are hard mechanisms in place to prevent that. Logs are stored in `/var/tmp` (not sure how much space we have on all the various clusters). Luckily, when archived, any files larger than 1M are compressed.

I talked with Phil and we think it's safe to move forward with this. Worst case if we notice issues in the future we can adjust the log rollover size per test or globally as needed.

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16314/5/testReport/

PR #16261, opened 2025-04-15
DAOS-15947 docs: Update admin guide dmg system command syntax

tanabarr:
Documentation updates relating to:
- New commands `dmg system drain|reintegrate`
- New command syntax for multiple --ranks
  `dmg pool drain|reintegrate|exclude –ranks 1-9,11-19`
- New command flag `dmg storage format --replace`

Doc-only: true

### Steps for the author:

* [ ] Commit message follows the [guidelines](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Appropriate [Features or Test-tag](https://daosio.atlassian.net/wiki/spaces/DC/pages/10984259629/Test+Tags) pragmas were used.
* [ ] Appropriate [Functional Test Stages](https://daosio.atlassian.net/wiki/spaces/DC/pages/12147556353/CI+Functional+Test+Stages) were run.
* [ ] At least two positive code reviews including at least one code owner from each category referenced in the PR.
* [ ] Testing is complete. If necessary, forced-landing label added and a reason added in a comment.

#### After all prior steps are complete:
* [ ] Gatekeeper requested (daos-gatekeeper added as a reviewer).

PR #16323, opened 2025-04-29
 DAOS-15947 docs: Update admin guide dmg system command syntax (#16261)

tanabarr:
Documentation updates relating to:
- New commands dmg system drain|reintegrate
- New command syntax for multiple --ranks dmg pool drain|reintegrate|exclude –ranks 1-9,11-19
- New command flag dmg storage format --replace

### Steps for the author:

### DIFF
--- a/docs/admin/administration.md
+++ b/docs/admin/administration.md
@@ -1043,7 +1043,8 @@ An examples workflow would be:
 - Storage server is rebooted after running `daos_server scm prepare` and command is run again.
 - Now PMem is intact, clear with `wipefs -a /dev/pmemX` where "X" refers to the repaired PMem ID.
 - `daos_server` can be started again. On start-up repaired engine prompts for "SCM format required".
-- Running `dmg storage format` now will create new "Joined" rank, instead run with `--replace` flag.
+- Run `dmg storage format --replace` to rejoin with existing rank (if --replace isn't used, a new
+  rank will be created).
 - Formatted engine will join using the existing (old) rank which is mapped to the engine's hardware.
 
 ### System Erase
--- a/docs/admin/administration.md
+++ b/docs/admin/administration.md
@@ -1029,6 +1029,23 @@ DAOS I/O Engines will be started, and all DAOS pools will have been removed.
     Then restart DAOS Servers and format.
 
 
+### Storage Format Replace
+
+If storage metadata for a rank is lost, for example after losing PMem contents after NVDIMM failure,
+storage for that rank will need to be formatted and rank metadata regenerated. If other hardware on
+the storage server has not changed the old rank can be "reused" by formatting using the
+`dmg storage format --replace` option.
+
+An examples workflow would be:
+- `daos_server` is running and PMem NVDIMM fails causing an engine to enter excluded state.
+- `daos_server` is stopped, storage server powered down, faulty PMem NVDIMM is replaced.
+- After powering up storage server, `daos_server scm prepare` command is used to repair PMem.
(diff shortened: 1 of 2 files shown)


docs/admin/administration.md
```diff
@@ -1029,6 +1029,23 @@ DAOS I/O Engines will be started, and all DAOS pools will have been removed.
...
+`dmg storage format --replace` option.
+
+An examples workflow would be:
+- `daos_server` is running and PMem NVDIMM fails causing an engine to enter excluded state.
+- `daos_server` is stopped, storage server powered down, faulty PMem NVDIMM is replaced.
+- After powering up storage server, `daos_server scm prepare` command is used to repair PMem.
+- Storage server is rebooted after running `daos_server scm prepare` and command is run again.
+- Now PMem is intact, clear with `wipefs -a /dev/pmemX` where "X" refers to the repaired PMem ID.
+- `daos_server` can be started again. On start-up repaired engine prompts for "SCM format required".
+- Running `dmg storage format` now will create new "Joined" rank, instead run with `--replace` flag.
```
Michael-Hennecke, line 1046: dont say what shouldn't be run, say what should be run (e.g. "run dmg storage format --replace to re-use the rank UUID and avoid creating a new rank UUID"

docs/admin/pool_operations.md
```diff
@@ -906,17 +971,34 @@ Target (rank 5 idx 1) is down.
...
 ```
 
 !!! warning
     While dmg pool query and list show how many targets are disabled for each
     pool, there is currently no way to list the targets that have actually
     been disabled. As a result, it is recommended for now to try to reintegrate
     all engine ranks one after the other via `for i in seq $NR_RANKs; do dmg
-    pool reintegrate --rank=$i; done`. This limitation will be addressed in the
+    pool reintegrate --ranks=$i; done`. This limitation will be addressed in the
     next release.
```
Michael-Hennecke, line 983: shouldn't we rewrite the for loop to construct a list of all ranks and pass it to one dmg pool reintegrate --ranks invocation a s a list (now that this is supported), instead of calling the reintegrate separately for each rank?

tanabarr: I think @johannlombardi is doing another pass over this area of the documentation soon, so rather than change the backport to be out of sync with the master commit maybe he can make this update then?

docs/admin/administration.md
tanabarr: I think @johannlombardi is doing another pass over this area of the documentation soon, so rather than change the backport to be out of sync with the master commit maybe he can make this update then?

Michael-Hennecke (dismissed):
didn't see that this is a backport, fine with me to land as-is, but my comments should then be addressed in both master and 2.6 (independent of Johann's edits which may or may not touch this)

docs/admin/pool_operations.md
tanabarr: This warning is confusing! Surely you don't want to go through every rank and instead you only want to go through the list of disabled ranks reported by pool query?

docs/admin/administration.md
tanabarr: changes made to this PR. master PR to update with review changes: https://github.com/daos-stack/daos/pull/16343

tanabarr:
> didn't see that this is a backport, fine with me to land as-is, but my comments should then be addressed in both master and 2.6 (independent of Johann's edits which may or may not touch this)

done

kjacque (approved):

PR #16331, opened 2025-05-01
DAOS-17504 build: Avoid unecessary patch download

jolivier23:
When --deps-as-gitmodules-subdir is used, we don't need to download patches as they are not applied.

### Steps for the author:

### DIFF
--- a/site_scons/prereq_tools/base.py
+++ b/site_scons/prereq_tools/base.py
@@ -1102,7 +1102,7 @@ class _Component():
         """Parse the patches variable"""
         patchnum = 1
         patchstr = self.prereqs.get_config("patch_versions", self.name)
-        if patchstr is None:
+        if patchstr is None or self.prereqs.deps_as_gitmodules_subdir:
             return {}
         patches = {}
         patch_strs = patchstr.split(",")


site_scons/prereq_tools/base.py
```diff
@@ -1102,7 +1102,7 @@ def _resolve_patches(self):
         """Parse the patches variable"""
         patchnum = 1
         patchstr = self.prereqs.get_config("patch_versions", self.name)
-        if patchstr is None:
+        if patchstr is None or self.prereqs.deps_as_gitmodules_subdir:
```
daltonbohning, line 1105: Could we just `return` here instead and skip this function altogether?
https://github.com/daos-stack/daos/blob/bd1bfe1a5077c5c6060552d6ac21b4eff30b1299/site_scons/prereq_tools/base.py#L1158-L1167

jolivier23: That line only sets up the retriever.  This line does the copy

https://github.com/daos-stack/daos/blob/bd1bfe1a5077c5c6060552d6ac21b4eff30b1299/site_scons/prereq_tools/base.py#L1176

daltonbohning: What I mean is why call `self.retriever.get` at all if we're just going to skip the logic inside?

daltonbohning: The logic is basically
```
if self.prereqs.deps_as_gitmodules_subdir:
    # setup
...
patches = self._resolve_patches()
# which does
if self.prereqs.deps_as_gitmodules_subdir:
    return {}
```
So there is no need to cal `self._resolve_patches` and modify it to look at this new boolean

jolivier23: We don't skip the logic in "get" though.  It's a "CopyRetriever" that has a get that does the copy.

jolivier23: We set the "retriever" to a CopyRetriever but pass an option to indicate it's already patched.  Eventually, get will invoke this code

https://github.com/daos-stack/daos/blob/bd1bfe1a5077c5c6060552d6ac21b4eff30b1299/site_scons/prereq_tools/base.py#L280

daltonbohning: Sorry, I think I see now

daltonbohning (approved):

PR #16354, opened 2025-05-08
DAOS-17531 object: Fix pl_map ref leak in EC agg

mjmac:
Call pl_map_decref() after pl_map_find() in order
to avoid a weird assertion on a long-running system.

Change-Id: Ia2a735856b14abede3ddf1b4cdd7a30cc0e9b1c2

### DIFF
--- a/src/placement/pl_map.c
+++ b/src/placement/pl_map.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2016-2023 Intel Corporation.
+ * (C) Copyright 2025 Google LLC
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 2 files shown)


src/placement/pl_map.c
```diff
@@ -350,6 +351,7 @@ pl_hop_rec_addref(struct d_hash_table *htab, d_list_t *link)
 
 	D_SPIN_LOCK(&map->pl_lock);
 	map->pl_ref++;
+	D_ASSERT(map->pl_ref > 0); /* detect overflow due to ref leak */
```
jolivier23, line 354: Can we make this D_ASSERTF and print the value?

jolivier23 (approved):

PR #16363, opened 2025-05-09
DAOS-17351 utils: add ofi tcp keepalive patch

wangdi1:
Enable keepalive in OFI to resolve the issue when connection is interrupted by engine restart.

### Steps for the author:


jolivier23 (dismissed):

utils/ofi_tcp_keepalive.patch
```diff
@@ -0,0 +1,139 @@
...
+ #define xnet_set_no_port(sock)
+ #endif
+ 
++int
++xnet_disable_keepalive(struct xnet_ep *ep)
++{
++	int optval = 0;
++	int ret;
++
++	ret = setsockopt(ep->bsock.sock, SOL_SOCKET, SO_KEEPALIVE, (char *)&optval,
```
jxiong, line 61: nic: should be `(const void *)&optval`.

utils/ofi_tcp_keepalive.patch
```diff
@@ -0,0 +1,139 @@
...
++}
++
+ int xnet_setup_socket(SOCKET sock, struct fi_info *info)
+ {
+ 	int ret, optval = 1;
+@@ -231,6 +291,8 @@ static int xnet_ep_connect(struct fid_ep *ep_fid, const void *addr,
+ 	if (!ep->addr)
+ 		return -FI_ENOMEM;
+ 
++	xnet_enable_keepalive(ep);
```
jxiong, line 122: Do you want to process error here?

btw, I tend to move this to the function `xnet_connect_done()`.

jxiong (approved):

utils/ofi_tcp_keepalive.patch
jxiong: can you please take care of this?

jxiong: I am not sure if keepalive timer would mess up with connecting timeout; I am too lazy to check the actual implementation, lol

utils/ofi_tcp_keepalive.patch
jolivier23: usually you don't need cast if target is void *

utils/ofi_tcp_keepalive.patch
wangdi1: Hmm, I thought it should still proceed to connect even if set keepalive failed. And also connection will be kept retrying, and doubling enable keepalive should be fine, so it seems fine to only disable it after the connection is setup.

Though I might miss some cases, what is your concern here? Thanks.

utils/ofi_tcp_keepalive.patch
wangdi1: ah, sure

utils/ofi_tcp_keepalive.patch
jxiong: it's more about uncertainties since I don't know how this works, like how keepalive behaves if the connection is not created yet. For example, if there are some hiccups in the network that makes connecting take much longer, and then keepalive mechanism kicks in, etc...

jolivier23 (approved):

jxiong (approved):

PR #16374, opened 2025-05-13
DAOS-17362 dtx: reduce backend TX size for batched commit

Nasf-Fan:
Change the batched commit DTX entries from 64 to 32 per backend TX. In the futher, if lower layer can return backend TX commit failure to the caller, we will adjust such count dynamically.

### Steps for the author:

### DIFF
--- a/src/dtx/dtx_internal.h
+++ b/src/dtx/dtx_internal.h
@@ -101,7 +101,7 @@ CRT_RPC_DECLARE(dtx, DAOS_ISEQ_DTX, DAOS_OSEQ_DTX);
 
 CRT_RPC_DECLARE(dtx_coll, DAOS_ISEQ_COLL_DTX, DAOS_OSEQ_COLL_DTX);
 
-#define DTX_YIELD_CYCLE		(DTX_THRESHOLD_COUNT >> 3)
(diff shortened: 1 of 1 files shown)


src/dtx/dtx_internal.h
```diff
@@ -101,7 +101,7 @@ CRT_RPC_DECLARE(dtx, DAOS_ISEQ_DTX, DAOS_OSEQ_DTX);
 
 CRT_RPC_DECLARE(dtx_coll, DAOS_ISEQ_COLL_DTX, DAOS_OSEQ_COLL_DTX);
 
-#define DTX_YIELD_CYCLE		(DTX_THRESHOLD_COUNT >> 3)
+#define DTX_YIELD_CYCLE         32
```
jxiong, line 104: I believe this will reduce the chance of this issue, however, is it possible to have a worst-case calculation for the size of tx to make sure it won't exceed `WAL_MAX_TRANS_BLKS`?

Nasf-Fan: In theory, we cannot avoid application to generate huge-sized transaction unless we explicitly claim that we do not support. This patch only tries to reduce the possibility of internally generating huge-sized transaction.

PR #16382, opened 2025-05-14
DAOS-17547 rebuild: error on stopped ds_pool_child

NiuYawei:
When a faulty SSD is replaced, reintegration will be auto triggered once local setup completed (ds_pool_child started).

Howerver, admin could manually run "dmg pool reintegrate" before the local setup done, then we need to return a retry-able error to make reintegration keep retry until the local ds_pool_child started.

### Steps for the author:

### DIFF
--- a/src/object/srv_obj_migrate.c
+++ b/src/object/srv_obj_migrate.c
@@ -493,8 +493,9 @@ migrate_pool_tls_create_one(void *data)
 
 	pool_child = ds_pool_child_lookup(arg->pool_uuid);
 	if (pool_child == NULL) {
-		D_ASSERTF(dss_get_module_info()->dmi_xs_id == 0,
-			  "Cannot find the pool "DF_UUIDF"\n", DP_UUID(arg->pool_uuid));
+		/* Local ds_pool_child isn't started yet, return a retry-able error */
+		if (dss_get_module_info()->dmi_xs_id != 0)
+			return -DER_STALE;
 	} else if (unlikely(pool_child->spc_no_storage)) {
 		D_DEBUG(DB_REBUILD, DF_UUID" "DF_UUID" lost pool shard, ver %d, skip.\n",
 			DP_UUID(arg->pool_uuid), DP_UUID(arg->pool_hdl_uuid), arg->version);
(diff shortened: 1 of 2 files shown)


liuxuezhao (dismissed):

src/object/srv_obj_migrate.c
```diff
@@ -493,8 +493,9 @@ migrate_pool_tls_create_one(void *data)
 
 	pool_child = ds_pool_child_lookup(arg->pool_uuid);
 	if (pool_child == NULL) {
-		D_ASSERTF(dss_get_module_info()->dmi_xs_id == 0,
-			  "Cannot find the pool "DF_UUIDF"\n", DP_UUID(arg->pool_uuid));
+		/* Local ds_pool_child isn't started yet, return a retry-able error */
+		if (dss_get_module_info()->dmi_xs_id != 0)
+			return -DER_STALE;
```
kccain, line 500: What are the implications of returning -DER_STALE here in migrate? Will the IO be retried (as opposed to having the entire rebuild operation get a -DER_STALE result)? If the IO is retried, does this also force a pool map refresh task to be run first before the retry (and is that appropriate for this case)?  I'm not as familiar with the object IO code, but see obj_comp_cb() setting pm_stale = true and invoking obj_retry_cb() that adds a pool map refresh task.

NiuYawei: In this case, the whole rebuild will be aborted and start over again. @liuxuezhao Please correct me if I'm wrong.

The pool map wasn't changed in this case, this issue (failed to find ds_pool_child) happens when a bad SSD is replaced by new SSD:
- Bad SSD is marked as faulty, the affected targets is auto excluded, rebuild done and the targets are in DOWNOUT
- New SSD is plugged and "dmg storage replace" is executed to replace the bad device with new one
- "dmg storage replace" will trigger local setup (setup the ds_pool_child) then reintegration in background (asynchronously).
- However, admin manually run"dmg pool reint" operation (which isn't necessary) before the local setup done, this assert will be triggered.

so we need to return a retry-able error before the local setup finished.

liuxuezhao: right the rebuild should be able to retry. 
To make sure it works, probably can add a fail injection to test that path.

NiuYawei: We've verified this fix by manual hotplug testing.

liuxuezhao (approved):

wangshilong (approved):

src/object/srv_obj_migrate.c
kccain: OK I see, looks like the error would be propagated through such that the tgt scan handler would return that error to the leader.

kccain (approved):

gnailzenh (approved):

PR #16396, opened 2025-05-16
DAOS-17300: RPC Protocol Versioning Enhancement for Rolling Upgrades 

wangshilong:
1. Added a generic daos_version_t type to encapsulate DAOS RPM version system RPC protocol.

2. Server's runtime daos_version_t is now returned in @daos_req_comm_out for future extensions.

3. Previously, servers assumed identical DAOS versions. This change introduces:
   - Runtime DAOS version tracking in each xstream
   - Each DAOS module maintains its version-to-RPC-protocol mapping
   - Modules use current system runtime protocol to determine correct RPC protocol
   - Currently supports single mapping (will expand to two entries for rolling upgrades)

4. Code refactored to separate RPC protocol selection logic between client and server.
5. Implement IV version awareness support by adding a version field to ds_iv_key structure.
The cache lookup will now use both key and version number to uniquely identify entries,
enabling proper handling of multiple cache versions for the same class during rolling upgrades.

### Steps for the author:


daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16396/2/execution/node/1341/log

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16396/2/display/redirect

daosbuild3:
Test stage Unit Test bdev on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16396/3/display/redirect

src/include/daos_srv/daos_engine.h
```diff
@@ -803,4 +805,32 @@ dss_chore_diy(struct dss_chore *chore);
...
+	{                                                                                          \
+		return dss_select_module_version(module_id, version);                              \
+	}
+
+#define DEFINE_RPC_PROTOCOL(module_prefix, module_id)                                              \
+	int module_prefix##_rpc_protocol(uint8_t *version)                                         \
+	{                                                                                          \
+		return dss_select_module_version(module_id, version);                              \
+	}
+
```
NiuYawei, line 849: What the difference of these two macros?

src/include/daos/rpc.h
```diff
@@ -64,6 +66,68 @@ enum daos_module_id {
...
+    .versions = {DAOS_VOS_VERSION, DAOS_MGMT_VERSION, DAOS_POOL_VERSION, DAOS_CONT_VERSION,
+		 DAOS_OBJ_VERSION, DAOS_REBUILD_VERSION, DAOS_RSVC_VERSION, DAOS_RDB_VERSION,
+		 DAOS_RDBT_VERSION, DAOS_SEC_VERSION, DAOS_DTX_VERSION, DAOS_PIPELINE_VERSION,
+		 DAOS_CHK_VERSION},
+    },
+    */
+};
+
+static inline int
+daos_get_module_version(int module_id, uint8_t protocol, uint8_t *version)
```
NiuYawei: Sounds like it's getting RPC version by 'module' and 'protocol'? daos_get_rpc_version() might be a better name.

src/include/daos_srv/daos_engine.h
```diff
@@ -88,6 +88,8 @@ struct dss_module_info {
 	int			dmi_ctx_id;
 	uint32_t		dmi_dtx_batched_started:1,
 				dmi_srv_shutting_down:1;
+	/* current daos system run-time version */
+	daos_version_t           dmi_version;
```
NiuYawei: Why do we make it per-xstream? Isn't it a global per-engine version?

src/engine/srv.c
```diff
@@ -416,6 +417,8 @@ dss_srv_handler(void *arg)
 	dmi->dmi_xs_id	= dx->dx_xs_id;
 	dmi->dmi_tgt_id	= dx->dx_tgt_id;
 	dmi->dmi_ctx_id	= -1;
+	/* TODO: get version from MS-DB */
+	daos_version_pack(&dmi->dmi_version);
```
NiuYawei: Each xstream query the version from MS-DB independently?

NiuYawei (commented):
LGTM

src/include/daos_srv/daos_engine.h
wangshilong: Yes, it should be per-engine, but there is a case engine version might be changed when rolling upgrade commit happen, make it per-xstream could avoid lock, or we could make this per-engine version as a atomic.(since it is actually u32).

src/engine/srv.c
wangshilong: No, it is per-engine, Engine query the version and pass it down to each XS.

src/include/daos/rpc.h
wangshilong: Yup, looks better.

src/include/daos_srv/daos_engine.h
wangshilong: It implement same function, i just implement it to follow different module's name style. Eg some module like pool/container, it prefer ds_pool/cont, some modules only used in the server side, function name avoid ds_prefix there..

daosbuild3:
Test stage NLT on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16396/9/execution/node/860/log

gnailzenh (dismissed):
One question, should we register for more than one RPC versions on client too? 
Of course, we can do this in later version like 3.0.

gnailzenh:
Sorry but there is conflict again...

daosbuild3:
Test stage NLT on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16396/10/execution/node/860/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16396/11/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16396/12/execution/node/1269/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16396/13/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16396/14/execution/node/1268/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16396/15/execution/node/457/log

gnailzenh (approved):

src/mgmt/srv_drpc.c
```diff
@@ -170,6 +170,13 @@ ds_mgmt_drpc_set_rank(Drpc__Call *drpc_req, Drpc__Response *drpc_resp)
 	if (rc != 0)
 		D_ERROR("Failed to set self rank %u: "DF_RC"\n", req->rank,
 			DP_RC(rc));
+	if (rc == 0) {
+		/* FIXME: get and set from req->daos_version */
```
tanabarr, line 174: I would prefer to document this "FIXME" in a ticket and reference it here so the task doesn't get lost. Even better would be to do that in this change so we aren't adding an unused protobuf definition. Given we don't use the `srv.proto daos_version` anywhere on the go side in this change, is there a reason it has to be added now? Could it instead be added when it's actually needed?

tanabarr (changes requested):

tanabarr (approved):
With regard to previous comment, it's not blocking and please address if you have to re-push. Is there a ticket you can reference/link in https://daosio.atlassian.net/browse/DAOS-17300 that covers the go changes required to use the new DaosVersion dRPC field?

src/mgmt/srv_drpc.c
wangshilong: I suppose it was protocol change between server side, so even it was not actually set now, i'd like to reserve it now.

wangshilong:
> With regard to previous comment, it's not blocking and please address if you have to re-push. Is there a ticket you can reference/link in https://daosio.atlassian.net/browse/DAOS-17300 that covers the go changes required to use the new DaosVersion dRPC field?

     I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change. And there is no ticket for that yet.

tanabarr:
> > With regard to previous comment, it's not blocking and please address if you have to re-push. Is there a ticket you can reference/link in https://daosio.atlassian.net/browse/DAOS-17300 that covers the go changes required to use the new DaosVersion dRPC field?
> 
> ```
>  I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change. And there is no ticket for that yet.
> ```

I think you should create the ticket then as part of the epic and link to the SRS then update the comment to reference and track the work.

src/include/daos/rpc.h
```diff
@@ -64,6 +66,68 @@ enum daos_module_id {
...
+		     DAOS_CHK_VERSION},
+    },
+    /* Please update DAOS_VERSION_PROTOCOL - 1 table when rolling upgrade is supported.
+    {
+    .protocol = DAOS_VERSION_PROTOCAL - 1;
+    .versions = {DAOS_VOS_VERSION, DAOS_MGMT_VERSION, DAOS_POOL_VERSION, DAOS_CONT_VERSION,
+		 DAOS_OBJ_VERSION, DAOS_REBUILD_VERSION, DAOS_RSVC_VERSION, DAOS_RDB_VERSION,
+		 DAOS_RDBT_VERSION, DAOS_SEC_VERSION, DAOS_DTX_VERSION, DAOS_PIPELINE_VERSION,
+		 DAOS_CHK_VERSION},
+    },
```
Nasf-Fan, line 105: It is better to define dyadic array "<module_id, version> [DAOS_NR_MODULE]" in `daos_protocol_table`, otherwise misplaced versions will cause trouble. That will also speedup daos_get_rpc_version().

wangshilong: Currently, init module_id equals array offset, this might be misused. i will change it to in the following PR.
> ``
+       .versions = {
+               [DAOS_VOS_MODULE]       = DAOS_VOS_VERSION,
                  .....
+       }
     },

wangshilong:
> > > With regard to previous comment, it's not blocking and please address if you have to re-push. Is there a ticket you can reference/link in https://daosio.atlassian.net/browse/DAOS-17300 that covers the go changes required to use the new DaosVersion dRPC field?
> > 
> > 
> > ```
> >  I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change. And there is no ticket for that yet.
> > ```
> 
> I think you should create the ticket then as part of the epic and link to the SRS then update the comment to reference and track the work.



> > > With regard to previous comment, it's not blocking and please address if you have to re-push. Is there a ticket you can reference/link in https://daosio.atlassian.net/browse/DAOS-17300 that covers the go changes required to use the new DaosVersion dRPC field?
> > 
> > 
> > ```
> >  I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change. And there is no ticket for that yet.
> > ```
> 
> I think you should create the ticket then as part of the epic and link to the SRS then update the comment to reference and track the work.

Sure, will do this.

wangshilong:
> > > With regard to previous comment, it's not blocking and please address if you have to re-push. Is there a ticket you can reference/link in https://daosio.atlassian.net/browse/DAOS-17300 that covers the go changes required to use the new DaosVersion dRPC field?
> > 
> > 
> > ```
> >  I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change. And there is no ticket for that yet.
> > ```
> 
> I think you should create the ticket then as part of the epic and link to the SRS then update the comment to reference and track the work.

@tanabarr I create the ticket here: https://daosio.atlassian.net/browse/DAOS-17896

kjacque:
>  I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change.

Just a thought: Looking at this PR, I actually think including the DAOS version in the protobuf struct may be superfluous. All communications at the control plane level include the DAOS version, including Join operations--we already use this for interop detection, so maybe there is a way to use what we already have, if we need to persist that information to track rolling upgrade.

If this is too painful maybe putting it in the JoinReq is still the right approach.

mjmac:
> > I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change.
> 
> Just a thought: Looking at this PR, I actually think including the DAOS version in the protobuf struct may be superfluous. All communications at the control plane level include the DAOS version, including Join operations--we already use this for interop detection, so maybe there is a way to use what we already have, if we need to persist that information to track rolling upgrade.
> 
> If this is too painful maybe putting it in the JoinReq is still the right approach.

I just happened to notice this in my feed... For the record all of the control plane RPCs are already automatically versioned: https://github.com/daos-stack/daos/blob/master/src/control/lib/control/interceptors.go#L80

There is also an interoperability framework for control plane components here: https://github.com/daos-stack/daos/blob/master/src/control/build/interop.go#L151

So, as Kris said, it is not necessary to add the version to protobuf messages if the goal is to define interoperability constraints between control plane components (dmg, daos_server, daos_agent). However, if there is a possibility that an engine could be a different version than its local daos_server process (this is currently not supported), then it may be useful to allow the engine to provide its own version via the upcall dRPC messages.

wangshilong:
> > > I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change.
> > 
> > 
> > Just a thought: Looking at this PR, I actually think including the DAOS version in the protobuf struct may be superfluous. All communications at the control plane level include the DAOS version, including Join operations--we already use this for interop detection, so maybe there is a way to use what we already have, if we need to persist that information to track rolling upgrade.
> > If this is too painful maybe putting it in the JoinReq is still the right approach.
> 
> I just happened to notice this in my feed... For the record all of the control plane RPCs are already automatically versioned: https://github.com/daos-stack/daos/blob/master/src/control/lib/control/interceptors.go#L80
> 
> There is also an interoperability framework for control plane components here: https://github.com/daos-stack/daos/blob/master/src/control/build/interop.go#L151
> 
> So, as Kris said, it is not necessary to add the version to protobuf messages if the goal is to define interoperability constraints between control plane components (dmg, daos_server, daos_agent). However, if there is a possibility that an engine could be a different version than its local daos_server process (this is currently not supported), then it may be useful to allow the engine to provide its own version via the upcall dRPC messages.

@mjmac 
      Currently we assume daos engines/daos_server always in the same version between different servers, the goal of this pr is to allow engines to be different versions, and during rolling upgrade process, system existed mixed engine versions(control plane too). current plan is to store rolling upgrade states in MS(include running version), when engine tried to join the system, it get that version.

@kjacque 
      We need check if there are better way, we could merge the logic here.

btw, here is design of rolling upgrade:https://daosio.atlassian.net/wiki/spaces/DC/pages/12289900560/DAOS+Rolling+Upgrade

wangshilong:
> > I thought when we implement new dmg rolling upgrade command, command will store DaosVersion in MS DB(as we discussed before), this code will be adjusted accordingly. so now it is more like a rpc protocol change.
> 
> Just a thought: Looking at this PR, I actually think including the DAOS version in the protobuf struct may be superfluous. All communications at the control plane level include the DAOS version, including Join operations--we already use this for interop detection, so maybe there is a way to use what we already have, if we need to persist that information to track rolling upgrade.

   Currently, the control plane context carries the DaosVersionHeader (DAOS build version) which is used for interoperability checking in unaryVersionInterceptor(). While this serves as a basic safety check, it proves insufficient for rolling upgrade scenarios.

   Consider a system with Component A (v2.8.0) and Component B (v3.0.0). For backward compatibility, Component B must communicate with Component A using protocols supported by the older version. This implies that when protocol changes exist between different versions, the system requires a negotiation mechanism between components.

   The current control plane implementation only performs basic sanity checks for interoperability. Proper protocol change support remains unimplemented. This PR introduces negotiation capability between different engine versions. Similar functionality should be implemented on the control plane side to fully support rolling upgrades.

PR #16404, opened 2025-05-18
DAOS-17550 object: refine dc_array_set_size processing

liuxuezhao:
1. use memcmp rather than strncmp in merge_key strncmp can cause some keys be incorrectly ignored
2. refine list dkey by providing more memory to save RPC round-trip.

### Steps for the author:


src/object/cli_obj.c
```diff
@@ -3573,7 +3573,7 @@ merge_key(struct dc_object *obj, d_list_t *head, char *key, int key_size)
 
 	d_list_for_each_entry(key_one, head, key_list) {
 		if (key_size == key_one->key.iov_len &&
-		    strncmp(key_one->key.iov_buf, key, key_size) == 0) {
+		    memcmp(key_one->key.iov_buf, key, key_size) == 0) {
```
mchaarawi, line 3576: yes that makes sense because integer dkeys are not strings

mchaarawi (approved):

mchaarawi:
i have not  tested on aurora yet, but will do soon

mchaarawi (changes requested):
im going to revert my +1 because the performance of the patch on aurora increases by 5x.

i just tested the memcmp fix, and that works as expected.
but the "improvement" in the array code made truncate worse my 5x:
With 4:26.38 minutes
Without 0:50.03 minutes

let's remove the other changes from the PR and just keep the memcmp fix please

liuxuezhao:
> im going to revert my +1 because the performance of the patch on aurora increases by 5x.
> 
> i just tested the memcmp fix, and that works as expected. but the "improvement" in the array code made truncate worse my 5x: With 4:26.38 minutes Without 0:50.03 minutes
> 
> let's remove the other changes from the PR and just keep the memcmp fix please

could you please test again with the refreshed PR? the original code only list 5 dkeys every time it looks not very good, you may tune the number of "ENUM_DESC_NR_MAX" to see which one is better on aurora. thanks

mchaarawi (approved):
verified functional and performance is back to normal

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16404/2/execution/node/1372/log

gnailzenh (approved):

PR #16475, opened 2025-06-04
DAOS-17643 control: Disallow most transitions from Excluded

kjacque:
Excluded ranks should be able to move only to the Joined or AdminExcluded states. Regardless of the process state, without re-joining successfully, the rank is still effectively excluded.

We saw a real life case where Excluded ranks were transitioning to the Errored state when they SIGKILLed themselves after discovering their exclusion.

Features: control

### Steps for the author:


liw (approved):
Seems to work in my simple test that used to cause "Joined" -> "Excluded" -> "Errored":

    DEBUG 2025/06/05 01:05:23.578552 membership.go:723: skipping illegal member state update for rank 3: Excluded->Errored

I'm surprised to find that without this PR the "Errored" rank (I think) causes a subsequent pool destroy to hang. This suggests that the analysis on "filtering" is very likely correct, and that it's a serious bug that we might want to consider backporting.

tanabarr (changes requested):
Seems like a reasonable change, what about the output of `dmg system query`? does the "reason" field in the table still get updated when the rank terminates? requesting changes just so we can verify before landing

liw:
> Seems like a reasonable change, what about the output of `dmg system query`? does the "reason" field in the table still get updated when the rank terminates? requesting changes just so we can verify before landing

I still have the `dmg system query -v` outputs from my tests. With this PR:

    3    92cfec45-8539-465c-8152-35d1ae36f949 10.214.208.181:31415 /hsw-252.daos.hpc.amslabs.hpecorp.net Excluded

Without this PR:

    1    9c03dcd1-d430-4ef2-b0a0-ea73826bdbe5 10.214.208.181:31415 /hsw-252.daos.hpc.amslabs.hpecorp.net Errored DAOS engine 0 exited unexpectedly: /home/liwei/daos/install/bin/daos_engine exited: signal: killed

mjmac:
> Seems like a reasonable change, what about the output of `dmg system query`? does the "reason" field in the table still get updated when the rank terminates? requesting changes just so we can verify before landing

It looks like the `Info` field of a member won't be updated if the state transition is not legal: https://github.com/daos-stack/daos/blob/master/src/control/system/membership.go#L556

While I agree that it would be nice to see the additional information in the system query output, IMO that's a more invasive change to do well and it would be better to handle it as a separate task. Getting the fix in for this improper state transition is a high priority, I think.

mjmac (approved):

tanabarr:
> > Seems like a reasonable change, what about the output of `dmg system query`? does the "reason" field in the table still get updated when the rank terminates? requesting changes just so we can verify before landing
> 
> I still have the `dmg system query -v` outputs from my tests. With this PR:
> 
> ```
> 3    92cfec45-8539-465c-8152-35d1ae36f949 10.214.208.181:31415 /hsw-252.daos.hpc.amslabs.hpecorp.net Excluded
> ```
> 
> Without this PR:
> 
> ```
> 1    9c03dcd1-d430-4ef2-b0a0-ea73826bdbe5 10.214.208.181:31415 /hsw-252.daos.hpc.amslabs.hpecorp.net Errored DAOS engine 0 exited unexpectedly: /home/liwei/daos/install/bin/daos_engine exited: signal: killed
> ```

Right so definitely a good fix which improves determinism and fixes the significant bug but we need to be aware that we are losing information important for debugging and that should be addressed as a follow-up task.

tanabarr (dismissed):
So do we want a unit test forr this?

kjacque:
> So do we want a unit test forr this?

I was debating with myself on that. While we have a number of complex MS Go tests, it doesn't seem like we have any Go tests verifying this state transition. Thought of writing one for the function itself -- that doesn't exist yet either.

@liw If you have a reproducer, can you provide the steps? Maybe we could create an ftest to exercise the full stack.

liw:
> @liw If you have a reproducer, can you provide the steps? Maybe we could create an ftest to exercise the full stack.

I used a trick:

1. Create a system of 4 engines.
2. Create a pool on only one engine, rank X.
3. On one server, invoke `gdb`, then in the GDB session, invoke `attach 12345`, where "12345" is the PID of the local daos_engine process, rank Y, where Y != X. This trick pauses the execution of rank Y.
4. Wait for the paused rank Y to be excluded from the system.
5. In the GDB session, invoke `q` and then type `y`. This resumes the execution of rank Y.
6. Observe that the resumed rank Y detects its exclusion from the system and commits suicide.
7. Destroy the pool.

Without this PR, in step 6 the engine state becomes "Errored", and in step 7 the pool destroy times out. Without this PR, in step 6 the engine state remains "Excluded", and in step 7 the pool destroy succeeds.

liw:
By the way, in the scenario of my test, I think the end result of "Excluded" and no "Reason" makes sense to me. Since the "reason" comes with the "Errored" update is a consequence of "Excluded", it would be confusing to be shown as the "reason" for "Excluded".

tanabarr:
> By the way, in the scenario of my test, I think the end result of "Excluded" and no "Reason" makes sense to me. Since the "reason" comes with the "Errored" update is a consequence of "Excluded", it would be confusing to be shown as the "reason" for "Excluded".

ok, let's leave it like that then.

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/1/execution/node/1399/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/1/execution/node/1413/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16475/1/testReport/

src/control/system/member.go
```diff
@@ -161,6 +162,14 @@ func (ms MemberState) isTransitionIllegal(to MemberState) bool {
...
 			MemberStateJoined: true,
 		},
+		MemberStateExcluded: {
+			MemberStateErrored:      true,
+			MemberStateStopping:     true,
+			MemberStateStopped:      true,
+			MemberStateStarting:     true,
+			MemberStateReady:        true,
+			MemberStateUnresponsive: true,
+		},
```
liw, line 172: Since @tanabarr has removed some of these in 2023, I think we can benefit from knowing whether that was for fixing an issue of just "user-friendly" messages or an issue of functionality.

liw (dismissed):

src/control/system/member.go
kjacque: Hmm, fair point. I didn't realize the history. It looks like the original issue was functional in nature, and maybe we're being a bit too strict internally about what we end up reporting to the user. I'll investigate this more tomorrow.

liw (approved):

src/control/system/member.go
liw: Sure, thanks. I've just approved for now.

kjacque: I tested with the reproducer steps on the ticket that prompted the 2023 changes, and wasn't able to reproduce the issue. So I think this is OK.

daosbuild3:
Test stage Build RPM on EL 8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/3/execution/node/364/log

daosbuild3:
Test stage Build RPM on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/3/execution/node/367/log

daosbuild3:
Test stage Build RPM on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/3/execution/node/326/log

src/control/system/member.go
tanabarr: The commit in question, 6d7df0ba4e850a7d1747f20c06eb3a3003abf9d0, related to a situation with PrepShutdown where failures conveyed through "MemberResult"s were not getting propagated correctly. This change is currently not very relevant because we basically disabled the PrepShutdown stage by forcing stop in the relatively recent changes to use SIGKILL. 

Thinking about it, the only transitions from Excluded I can think of as legal are AdminExcluded and Joined. Which is what you have in the PR. Should Excluded->CheckerStarted be allowed?

One then that I think we should do with this PR is to ensure that we are effectively logging failures that would otherwise get ignored with a refused illegal state transition. MemberResults received with an errored state should be printed when discarded. This should be performed within UpdateMemberStates() whenever discarding an errored MemberResult due to an illegal state transition.

In summary, change looks good and can't see any reason not to. The 2023 change doesn't have any subtleties as for as I can see that would make this a problem.

daosbuild3:
Test stage Build RPM on EL 8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/5/execution/node/339/log

daosbuild3:
Test stage Build RPM on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/5/execution/node/338/log

daosbuild3:
Test stage Build RPM on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/5/execution/node/352/log

src/control/system/member.go
tanabarr: Thinking about state transitions in another context, I think it's helpful for admins to know when a Rank is Stopped as opposed to Excluded. For example if we don't allow Excluded->Stopped then the user can't tell whether the relevant engine process is running. I would be in favour of allowing that transition and ensuring that rejoin suicides AdminExcluded ranks.

kjacque: Right now we do not allow Excluded->Stopped. I think it may require a bit more thought about how we handle that, as the "Excluded"/"AdminExcluded" states are currently special and affect the group map, while other errored/stopped states don't. I would prefer to be more strict now, and think about how to offer a more granular way to interrogate the state in the future.

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16475/6/execution/node/1623/log

kjacque:
There is only one test failure, and it is a known issue: https://daosio.atlassian.net/browse/DAOS-17657

Requesting forced landing.

PR #16476, opened 2025-06-05
DAOS-17644 gurt: Add server object I/O latency histogram

wangdi1:
Add server I/O latency histogram, and add %50 90% 99% latency to metrics.

### Steps for the author:

### DIFF
--- a/src/gurt/telemetry.c
+++ b/src/gurt/telemetry.c
@@ -1318,11 +1318,72 @@ d_tm_print_duration(struct timespec *tms, struct d_tm_stats_t *stats,
 		d_tm_print_stats(stream, stats, format);
 }
 
+static uint64_t
+d_tm_histogram_get_value(struct d_tm_histogram_t *histogram, uint64_t total_count,
+			 uint32_t percentile)
+{
+	double target_num;
+	double current_num = 0;
+	double prev_num    = 0;
+	int    i;
+
+	if (percentile == 0)
+		return histogram->dth_buckets[0].dtb_min;
+
+	if (percentile >= 100)
+		return histogram->dth_buckets[histogram->dth_num_buckets - 1].dtb_max;
+
+	target_num = total_count * ((double)percentile / 100);
+	for (i = 0; i < histogram->dth_num_buckets; i++) {
+		uint64_t current_val;
+
+		current_val = histogram->dth_buckets[i].dtb_bucket->dtn_metric->dtm_data.value;
+		prev_num    = current_num;
+		current_num += current_val;
+		if (target_num <= current_num) {
+			uint64_t low            = histogram->dth_buckets[i].dtb_min;
+			uint64_t high           = histogram->dth_buckets[i].dtb_max;
+			double   cur_percentage = (target_num - prev_num) / current_val;
+
+			return low + cur_percentage * (high - low);
+		}
+	}
+
+	return histogram->dth_buckets[histogram->dth_num_buckets - 1].dtb_max;
+}
+
+static void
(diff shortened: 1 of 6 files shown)


src/object/obj_utils.c
```diff
@@ -132,6 +133,11 @@ obj_latency_tm_init(uint32_t opc, int tgt_id, struct d_tm_node_t **tm, char *op,
 			D_WARN("Failed to create per-I/O size latency "
 			       "sensor: " DF_RC "\n",
 			       DP_RC(rc));
+
+		rc = d_tm_init_histogram(tm[i], path, 18, 256, 2, "ns");
```
mjmac, line 137: I wonder if we need this many buckets per histogram. Effectively what this does is add 18 new counters per latency gauge. Before this change, we have the following:

NR_LATENCY_BUCKETS: 16
OP_COUNT: 2 (fetch, update)
LAYER_COUNT: 4 (global, bulk, vos, bio)

16 * 2 * 4 = 128 * N targets

e.g. 16 targets = 2048 gauge metrics per engine

If we add an 18-bucket histogram for every one of those gauges, now we have: 2048 * 18 = 36864 additional counter metrics (one for each bucket). That's a pretty big jump. I checked this with the `daos_metrics` tool:

Before:
```
(rocky-8)mjmac@mjmac-daos-dev00:~/projects/daos$ daos_metrics | wc -l
3282
```

After:
```
(rocky-8)mjmac@mjmac-daos-dev00:~/projects/daos$ daos_metrics | wc -l
24018
```

Also, is the unit "ns" or "us"? The gauge unit above this is "us", i.e. microseconds. I think I/O latencies are usually measured in microseconds, not nanoseconds. If we are measuring in microseconds, then we probably don't really need fine-grained buckets for latencies greater than some large amount, e.g. 1048575us, right? That's 1.04 seconds, which is an eternity for an I/O.

The other question I have is whether or not it's going to be useful for telemetry consumers to have detailed histograms at each layer like this. I can see how it would be useful for debugging locally using `daos_metrics`, but it might be too much to store in a system like Prometheus. Maybe it's fine and we just document that consumers should filter out the low-level latencies if they don't want them.

src/gurt/telemetry.c
```diff
@@ -3792,6 +3874,64 @@ d_tm_get_gauge(struct d_tm_context *ctx, uint64_t *val,
...
+	if (unlikely(!node_is_readable(node)))
+		return -DER_AGAIN;
+
+	metric_data = conv_ptr(mem_hdr, node->dtn_metric);
+	if (metric_data != NULL) {
+		struct d_tm_histogram_t *dtm_histogram = NULL;
+
+		dtm_histogram = conv_ptr(mem_hdr, metric_data->dtm_histogram);
+		d_tm_node_lock(node);
+		if (has_stats(node) && histogram != NULL && dtm_histogram != NULL) {
```
mjmac, line 3904: nit: Maybe check `if (histogram == NULL)` above and exit with `-DER_INVAL`?

mjmac (commented):
Looks pretty good! I checked it out to play around with it. One thing I'm not sure about is if the Prometheus exporter should somehow expose the percentiles or if those should just be calculated via Prometheus query. Not necessary to address that in this PR though.

src/object/obj_utils.c
wangdi1: ok, let me change it to us?  Hmm, I will shrink the buck number to 6, and start from 128. Then each target will have 6 * 2 * 4 = 48 per target, 16 targets will have 768 gauge metrics per engine.

PR #16484, opened 2025-06-06
DAOS-17403 telem: Add non-shared memory flag (#16237)

mjmac:
In cases where the client telemetry will not be
sampled via the agent (e.g. when dumping to a
CSV at process exit), it would be preferable to
avoid overhead and exposure to resource limits
imposed by the use of shared memory segments.

Setting the D_TM_NO_SHMEM flag in d_tm_init()
will result in process-local memory allocations
for the telemetry segments.

Change-Id: I60f8640aec5bc856c43b83c3bd691ae07276d457

### DIFF
--- a/src/client/api/metrics.c
+++ b/src/client/api/metrics.c
@@ -106,7 +106,7 @@ dc_tm_init(crt_init_options_t *crt_info)
 {
 	struct d_tm_node_t *started_at;
 	pid_t               pid = getpid();
-	int                 metrics_tag;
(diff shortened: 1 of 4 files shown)


daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16484/2/testReport/

PR #16494, opened 2025-06-10
DAOS-17444 engine: reduce lock contention during rebuild

wangshilong:
Previously, the engine used a global lock for hash operations which created significant performance bottlenecks. This lock was shared across multiple target threads on the server side, leading to high contention during rebuild operations.

### Steps for the author:

### DIFF
--- a/src/engine/init.c
+++ b/src/engine/init.c
@@ -1,5 +1,6 @@
 /*
  * (C) Copyright 2016-2024 Intel Corporation.
+ * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
  * (C) Copyright 2025 Google LLC
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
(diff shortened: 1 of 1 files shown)


src/engine/init.c
```diff
@@ -753,10 +754,9 @@ server_init(int argc, char *argv[])
 		D_GOTO(exit_mod_init, rc);
 	D_INFO("Network successfully initialized\n");
 
-	rc = daos_hhash_init();
+	rc = daos_hhash_init_feats(D_HASH_FT_RWLOCK);
```
liuxuezhao, line 757: I almost forgot details, would "D_HASH_FT_RWLOCK | D_HASH_FT_LRU" be a little better?

liuxuezhao (approved):

src/engine/init.c
liuxuezhao: seems for server-side handle hash table, it could be per-thread local, and then need not thread lock at all. as we don't open one handle at thread X and use it at thread Y. but need some more check for this.

wangshilong: If D_HASH_FT_LRU is used, read will use write lock(because of moving caching entry..) I tested with/without LRU flag on the client side before, it looks disable LRU(make read-write lock usable) gives better performance....

wangshilong: > seems for server-side handle hash table, it could be per-thread local, and then need not thread lock at all. as we don't open one handle at thread X and use it at thread Y. but need some more check for this.

That will be even better IMO.

liuxuezhao: as discussed on teams, serverside use PTR type, so probably need not this PR.
see d_hhash_is_ptrtype related code.

gnailzenh: RW lock is not much better when there are more than handful threads, does it require a lot of changes if we switch to per-thread hash table? 
Do we use the hash table for object handle or container handle? I can't remember....

liuxuezhao: > RW lock is not much better when there are more than handful threads, does it require a lot of changes if we switch to per-thread hash table? Do we use the hash table for object handle or container handle? I can't remember....

for server-side the handle is just using pointer, bypassed hash-table but jut keep the same API as client.

PR #16511, opened 2025-06-14
DAOS-16170 control: Ignore EngineDied event for old incarnation

kjacque:
It is possible to be forwarded an EngineDied event late, after the engine has re-joined. This can incorrectly re-mark the rank as Errored.

- Include incarnation in engine-related events.
- Print incarnation in logs if provided.
- Do not update member if engine died event is for old incarnation.

Features: control

### Steps for the author:

### DIFF
--- a/src/control/server/instance_exec.go
+++ b/src/control/server/instance_exec.go
@@ -1,5 +1,6 @@
 //
 // (C) Copyright 2020-2023 Intel Corporation.
+// (C) Copyright 2025 Hewlett Packard Enterprise Development LP
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -105,12 +106,12 @@ func (ei *EngineInstance) finishStartup(ctx context.Context, ready *srvpb.Notify
 // createPublishInstanceExitFunc returns onInstanceExitFn which will publish an exit
 // event using the provided publish function.
 func createPublishInstanceExitFunc(publish func(*events.RASEvent), hostname string) onInstanceExitFn {
-	return func(_ context.Context, engineIdx uint32, rank ranklist.Rank, exitErr error, exPid int) error {
+	return func(_ context.Context, engineIdx uint32, rank ranklist.Rank, incarnation uint64, exitErr error, exPid int) error {
 		if exitErr == nil {
 			return errors.New("expected non-nil exit error")
 		}
 
-		evt := events.NewEngineDiedEvent(hostname, engineIdx, rank.Uint32(),
+		evt := events.NewEngineDiedEvent(hostname, engineIdx, rank.Uint32(), incarnation,
 			common.ExitStatus(exitErr.Error()), exPid)
 
 		// set forwardable if there is a rank for the MS to operate on
@@ -143,7 +144,7 @@ func (ei *EngineInstance) handleExit(ctx context.Context, exitPid int, exitErr e
 	// After we know that the instance has exited, fire off
 	// any callbacks that were waiting for this state.
 	for _, exitFn := range ei.onInstanceExit {
-		err := exitFn(ctx, engineIdx, rank, exitErr, exitPid)
+		err := exitFn(ctx, engineIdx, rank, ei.incarnation, exitErr, exitPid)
 		if err != nil {
 			ei.log.Errorf("onExit: %s", err)
 		}
--- a/src/control/system/membership.go
+++ b/src/control/system/membership.go
@@ -711,6 +711,11 @@ func (m *Membership) handleEngineFailure(evt *events.RASEvent) {
 		return
 	}
 
+	if evt.Incarnation < member.Incarnation {
+		m.log.Debugf("skipping event for incarnation = %d, current member incarnation = %d: %s", evt.Incarnation, member.Incarnation, evt.String())
+		return
+	}
+
 	// TODO DAOS-7261: sanity check that the correct member is being
 	//                 updated by performing lookup on provided hostname
 	//                 and matching returned addresses with the address
--- a/src/control/events/engine.go
+++ b/src/control/events/engine.go
@@ -1,5 +1,6 @@
 //
 // (C) Copyright 2020-2022 Intel Corporation.
+// (C) Copyright 2025 Hewlett Packard Enterprise Development LP
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -58,15 +59,16 @@ func EngineStateInfoToProto(rsi *EngineStateInfo) (*sharedpb.RASEvent_EngineStat
 }
 
 // NewEngineDiedEvent creates a specific EngineDied event from given inputs.
-func NewEngineDiedEvent(hostname string, instanceIdx uint32, rank uint32, exitErr common.ExitStatus, exPid int) *RASEvent {
+func NewEngineDiedEvent(hostname string, instanceIdx uint32, rank uint32, incarnation uint64, exitErr common.ExitStatus, exPid int) *RASEvent {
 	return fill(&RASEvent{
-		Msg:      fmt.Sprintf("DAOS engine %d exited unexpectedly: %s", instanceIdx, exitErr),
-		ID:       RASEngineDied,
-		Hostname: hostname,
-		Rank:     rank,
-		Type:     RASTypeStateChange,
-		Severity: RASSeverityError,
-		ProcID:   exPid, // pid of exited engine
+		Msg:         fmt.Sprintf("DAOS engine %d exited unexpectedly: %s", instanceIdx, exitErr),
+		ID:          RASEngineDied,
+		Hostname:    hostname,
+		Rank:        rank,
+		Incarnation: incarnation,
+		Type:        RASTypeStateChange,
+		Severity:    RASSeverityError,
+		ProcID:      exPid, // pid of exited engine
 		ExtendedInfo: &EngineStateInfo{
 			InstanceIdx: instanceIdx,
 			ExitErr:     exitErr,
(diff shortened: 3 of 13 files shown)


daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16511/1/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/1/execution/node/1427/log

phender:
@kjacque this PR needs to run with the `recovery` tag/feature.

kjacque:
> @kjacque this PR needs to run with the `recovery` tag/feature.

Good catch. I rebased on master and amended the commit pragma.

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/3/execution/node/627/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/3/execution/node/672/log

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/5/execution/node/1338/log

src/control/system/membership.go
```diff
@@ -711,6 +711,11 @@ func (m *Membership) handleEngineFailure(evt *events.RASEvent) {
 		return
 	}
 
+	if evt.Incarnation != 0 && evt.Incarnation < member.Incarnation {
```
tanabarr, line 714: when would incarnation be zero and why should we skip this clause if it is?

tanabarr (approved):
looks like a good change

src/control/system/membership.go
kjacque: If the engine did not successfully join before dying, it's possible that the control plane never got an incarnation for the engine, as it's generated by the engine itself. In that case we don't want to skip processing the EngineDied event.

kjacque: As I think about it more, I think we should record the incarnation ASAP after getting the "ready" response from the engine. Then we could probably avoid that case altogether.

liw, line 714: It's good that we use `<` instead of `==` here, since there's subtlety with our incarnation numbers today:
```
Engine X joins with incarnation 1
Engine X bumps its incarnation from 1 to 2 because it finds that someone suspects it (The control plane is not informed directly about this.)
Engine X terminates
MS sees events for engine X with incarnation 2
```

liw (dismissed):

src/control/system/membership.go
kjacque: Interesting! That's good to know, thanks.

kjacque: Fixed with the latest commit. Thanks!

liw (dismissed):

src/control/system/membership.go
tanabarr, line 714: with your most recent change can we not remove the zero-check here?

kjacque: We could -- I'd like to leave it in case there's still some gap somewhere. Logically there shouldn't be, but it also makes no sense to check incarnation at all if it's unset. 0 is never valid.

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/7/execution/node/1368/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/7/execution/node/1323/log

src/control/system/membership.go
tanabarr: Shouldn't we be returning a failure if it's not set? otherwise looks like we are masking an invalid state.
I understand the scenario where we don't want to create a failure case but covering up undesired scenarios will also bite us potentially.

kjacque: Understood... We actually don't return errors from this function. Maybe we should log it, though.

kjacque: Spent a little more time looking into this. There actually is a case where 0 would not be surprising. Consider this case:
- Admin starts daos_server.
- daos_engine process starts and dies before it can send a dRPC to the control plane including its incarnation.
- The RAS event includes incarnation = 0, because the control plane never got an incarnation.

This would only happen immediately after starting up the daos_server. We could conceivably have an incarnation in the member DB at that point, if the engine had run successfully in the past, but we would not have a current incarnation. That is definitely a corner case, and could leave an opening for the same bug to come through (essentially: engine fails to come up, event is sent, engine successfully comes up, forwarded event for incarnation 0 arrives, engine is incorrectly marked as errored).

kjacque: I spent a little time talking to @mjmac about this yesterday. I think I have a solution for this particular corner case and am working on it.

daosbuild3:
Test stage Functional on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16511/9/testReport/

tanabarr (approved):

tanabarr (approved):
probably want to mention superblock related updates somewhere in the commit message

kjacque:
> probably want to mention superblock related updates somewhere in the commit message

Good idea, I added a line to the description so it can be used as the commit message.

src/control/server/instance_exec.go
```diff
@@ -64,6 +81,10 @@ func (ei *EngineInstance) start(ctx context.Context) (chan *engine.RunnerExitInf
 		ei.log.Errorf("instance %d: unable to log SCM storage stats: %s", ei.Index(), err)
 	}
 
+	if err := ei.initIncarnationFromSuperblock(); err != nil {
```
liw, line 84: Related to [this comment](https://github.com/daos-stack/daos/pull/16511#discussion_r2190481128), when an engine exits before sending its incarnation to the local daos_server, this incarnation of the engine must have not joined the system yet, so the RAS event (actually, what RAS event that is?) with incarnation 0 may be safely ignored?

kjacque: So this is the EngineDied event. If the engine has never joined (incarnation 0), I think we can ignore EngineDied, you're right. There is no Member to update. If the engine has joined in the past and there is an incarnation in the superblock, the event is useful because it updates the Member status. If we use the last known incarnation, that should be OK.

liw: Sorry, my question is assuming that we don't use the last known incarnation, that is, without the newly-added superblock field. Using the last known incarnation seems to be lying (even though it might be okay for the current code). And to do so we are adding a new persistent field---I'm wondering if that's going too far, and if we should just ignore such an event, or I must be missing the point?

If we can't ignore the event, there may be alternative approaches. For instance, could daos_server generate the incarnation and pass it to daos_engine? This way, we wouldn't need to store it persistently in the superblock, and the event would become accurate (i.e., the incarnation in the event would be faithful). Again, that's _if we can't ignore the event_.

kjacque: Ah, I see. The issue with ignoring the event is potentially the following situation:
1. DAOS has been up and running for a while.
2. Admin shuts down daos_servers.
3. Admin brings daos_servers back up.
4. daos_engine begins to start up and then crashes.

In this case the incarnation would be 0. If we ignore the event, the state of the affected Member will *not* become "Errored". I'm not sure what the state would be.

I wouldn't be against the daos_server being the one who defines the incarnation. That requires a bit more surgery though. That also feels a little stranger in the case you described previously, where the engine bumps its own incarnation because it's considered suspect, but the daos_server doesn't know about that change.

liw: I think the incarnation bumps upon suspicions are largely a separate "problem". (We could easily notify daos_server upon those if there were good reasons.)

If we ignore the event in that case, and the member ends up in "starting" state, things will be too confusing. (If the member ends up in "stopped"/"excluded" state, I'd consider the state acceptable.) So it seems that we can't ignore the event...

OK, at least it sounds like the concern about the addition of the "incarnation" superblock field is well understood. Sorry for delaying this PR.

liw (approved):

src/control/server/instance_exec.go
kjacque: No worries. I appreciate the discussion and insights. Thanks!

knard38 (approved):

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/11/execution/node/1268/log

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/12/execution/node/793/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/12/execution/node/804/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/13/execution/node/1498/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/13/execution/node/1509/log

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16511/13/testReport/

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16511/14/execution/node/1464/log

kjacque:
Test failures are known issues:
- https://daosio.atlassian.net/browse/DAOS-17888
- https://daosio.atlassian.net/browse/DAOS-17751

I know we can't land anything until CI is up next week, but whenever things re-open, this one is ready.

PR #16299, opened 2025-04-22
DAOS-17381 rebuild: refine rebuild IO scheduling and rebuild IV

liuxuezhao:
1. refine rebuild IV detailed handling
2. for rebuild IO, apply SCHED_REQ_MIGRATE for scheduler throttling
3. refine yield for rebuild scanner
4. use d_list_del_init for rgt delete
5. skip orphan container for rebuild scanner

### Steps for the author:

* [ ] Commit message follows the [guidelines](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Appropriate [Features or Test-tag](https://daosio.atlassian.net/wiki/spaces/DC/pages/10984259629/Test+Tags) pragmas were used.
* [ ] Appropriate [Functional Test Stages](https://daosio.atlassian.net/wiki/spaces/DC/pages/12147556353/CI+Functional+Test+Stages) were run.
* [ ] At least two positive code reviews including at least one code owner from each category referenced in the PR.
* [ ] Testing is complete. If necessary, forced-landing label added and a reason added in a comment.

#### After all prior steps are complete:
* [ ] Gatekeeper requested (daos-gatekeeper added as a reviewer).

PR #16515, opened 2025-06-16
DAOS-17696 pool: discard data for pool extending

liuxuezhao:
Call pool_discard() to remove existing data for pool extending, to cleanup space.

### Steps for the author:

### DIFF
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -7520,45 +7520,9 @@ out:
 	return rc;
 }
 
-void
-ds_pool_extend_handler(crt_rpc_t *rpc)
-{
-	struct pool_extend_in	*in = crt_req_get(rpc);
-	struct pool_extend_out	*out = crt_reply_get(rpc);
-	struct pool_svc		*svc;
-	uuid_t			pool_uuid;
-	d_rank_list_t		rank_list;
-	uint32_t		ndomains;
-	uint32_t		*domains;
-	int			rc;
-
-	D_DEBUG(DB_MD, DF_UUID": processing rpc %p\n", DP_UUID(in->pei_op.pi_uuid), rpc);
-
-	uuid_copy(pool_uuid, in->pei_op.pi_uuid);
-	rank_list.rl_nr = in->pei_tgt_ranks->rl_nr;
-	rank_list.rl_ranks = in->pei_tgt_ranks->rl_ranks;
-	ndomains = in->pei_ndomains;
-	domains = in->pei_domains.ca_arrays;
-
-	rc = pool_svc_lookup_leader(in->pei_op.pi_uuid, &svc, &out->peo_op.po_hint);
-	if (rc != 0)
-		goto out;
-
-	rc =
-	    pool_svc_update_map(svc, pool_opc_2map_opc(opc_get(rpc->cr_opc)),
-				false /* exclude_rank */, &rank_list, domains, ndomains, NULL, NULL,
-				&out->peo_op.po_map_version, &out->peo_op.po_hint, MUS_DMG, true);
-
-	pool_svc_put_leader(svc);
-out:
-	out->peo_op.po_rc = rc;
-	D_DEBUG(DB_MD, DF_UUID ": replying rpc: %p " DF_RC "\n", DP_UUID(in->pei_op.pi_uuid), rpc,
-		DP_RC(rc));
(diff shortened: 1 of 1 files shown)


src/pool/srv_pool.c
```diff
@@ -7520,45 +7520,9 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
...
-out:
-	out->peo_op.po_rc = rc;
-	D_DEBUG(DB_MD, DF_UUID ": replying rpc: %p " DF_RC "\n", DP_UUID(in->pei_op.pi_uuid), rpc,
-		DP_RC(rc));
-	crt_reply_send(rpc);
-}
-
 static int
-pool_discard(crt_context_t ctx, struct pool_svc *svc, struct pool_target_addr_list *list)
+pool_discard(crt_context_t ctx, struct pool_svc *svc, struct pool_target_addr_list *list,
```
kccain, line 7524: Does ds_pool_tgt_discard_ult() that gets run as part of the corpc handling need to change so that its ex_status does not have PO_COMP_ST_NEW? To achieve the data discard for the extend case?

If I am reading the code correctly, this pool_discard() will issue a corpc before the pool map is updated (for exclude, the pool map update will change the new targets from NEW -> UP state). 

For the exclude case, then I assume that the targets are in the NEW state when pool_discard() happens. Given this, it seems the corpc handling will not discard data on the targets while they are in this NEW state (but the patch intends to discard data). ds_pool_tgt_discard_handler() -> ds_pool_tgt_discard_ult() -> ds_pool_thread_collective() seems that it will _**not**_ call pool_child_discard() for targets in the NEW state (since ex_status is assigned like this)
`	ex_status =
	    PO_COMP_ST_UP | PO_COMP_ST_UPIN | PO_COMP_ST_DRAIN | PO_COMP_ST_DOWN | PO_COMP_ST_NEW;`

So we might consider this assignment instead?
`ex_status =
	    PO_COMP_ST_UP | PO_COMP_ST_UPIN | PO_COMP_ST_DRAIN | PO_COMP_ST_DOWN;`

I guess this pool_discard() flow works for the reintegration case because those targets are already in the DOWNOUT state at the time of the pool_discard() invocation? And ex_status does not include PO_COMP_ST_DOWNOUT?

kccain: No change seems to be needed (but please confirm my understanding). I cooked a local patch to experiment. 

For reintegration, the rank's target states are PO_COMP_ST_DOWNOUT=16, so the rank does not get added to the array of targets when ds_pool_get_tgt_idx_by_state() calls pool_map_find_by_rank_status(). And therefore, discarding data occurs for these targets.

For exclude, I am concluding that the rank's target states are PO_COMP_ST_UNKNOWN=0 (you can't "match" on unknown=0 state value that I know of, and it the target states do not match DOWNOUT or the other target states provided in the status argument). So, similarly the rank does not get added to the array of targets as a result of the pool_map_find_by_rank_status() call. And so data discard will occur in this case, too.

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16515/3/execution/node/1369/log

kccain (approved):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16515/3/testReport/

src/pool/srv_pool.c
liuxuezhao: i did some test, current code can work because in the ds_pool_tgt_discard_ult() -> ds_pool_thread_collective() -> ds_pool_collective_reduce() -> ds_pool_get_tgt_idx_by_state(), for the in extending engine, ds_pool_lookup() can get the pool but "pool->sp_map" is still NULL. so it actually get zero exclude_tgt_nr then it can work even for NEW tgt.
But I think probably good to remove the NEW status from ex_status, some special cases maybe can make the pool->sp_map be valid ahead. The change will cause some conflict with PR https://github.com/daos-stack/daos/pull/16299, so I am going to merge to that PR, and discard this PR.

PR #16529, opened 2025-06-20
DAOS-17637 vos: support to mark corruption on object/{d,a}key

Nasf-Fan:
Sometimes. when rebuild, we may have not enough information to rebuild some object or {d,a}key. Under such case, we need to add some flags on the target, then subsequent operation, such as read data from such bad target can properly return errno to indicate some data lost. Otherwise it will cause sclient data corruption.

For such purpose, we introduce new VOS API: vos_obj_mark_corruption(). It allows server side caller to mark the specified object or {d,a}key as corrupted. Any subsequent online operation against such bad target, except discard, will get -DER_DATA_LOSS failure.

The user/admin can use ddb (rm) to remove corrupted target to release related space.

### Steps for the author:


daosbuild3:
Test stage Unit Test with memcheck on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16529/2/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16529/3/testReport/

Nasf-Fan:
EC_17/18 failed for DAOS-17656, not related with the patch.

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16529/4/testReport/

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16529/6/display/redirect

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16529/7/display/redirect

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16529/8/display/redirect

Nasf-Fan:
All required CI tests passed.

Nasf-Fan:
Ping reviewers, thanks!

src/vos/vos_tree.c
```diff
@@ -1236,6 +1269,55 @@ key_tree_punch(struct vos_object *obj, daos_handle_t toh, daos_epoch_t epoch,
...
+	if (rc != 0 || sub_toh == NULL)
+		return rc;
+
+	if (krec->kr_btr.tr_class != 0)
+		return dbtree_open_inplace_ex(&krec->kr_btr, uma, vos_cont2hdl(cont), obj, sub_toh);
+
+	ta = tree_find_attr_by_class(tclass);
+	D_ASSERT(ta != NULL);
+
+	return dbtree_create_inplace_ex(tclass, tree_obj2feats(obj, false), ta->ta_order, uma,
```
gnailzenh, line 1317: Do we really want to create subtree if it's already corrupted? Or this is just for keeping unified tree typology and consistent of implementation? In addition, I think VOS can support "flat KV", which means there is no akey tree, should this be considered?

Nasf-Fan: When we arrive here, `sub_toh` will be non-NULL, that means `akey_nr` is non-zero from the caller `vos_obj_mark_corruption()`.

```
int
vos_obj_mark_corruption(daos_handle_t coh, daos_epoch_t epoch, uint32_t pm_ver, daos_unit_oid_t oid,
                        daos_key_t *dkey, unsigned int akey_nr, daos_key_t *akeys)
{
...
        if (akey_nr != 0) {
                if (dkey == NULL)
                        D_GOTO(log, rc = -DER_INVAL);

                if (akeys == NULL)
                        D_GOTO(log, rc = -DER_INVAL);

                if (vos_obj_skip_akey_supported(cont, oid))
                        D_GOTO(log, rc = -DER_NO_PERM);
...
}
```

In above caller logic, if `akey_nr` is non-zero but `VOS_POOL_FEAT_FLAT_DKEY` is set, then will return `-DER_NO_PERM` when check `vos_obj_skip_akey_supported()` before calling `vos_tree_mark_corruption()`. That will avoid creating akey tree.

On the other hand, if the outside user requires to mark corruption on both dkey and some akey, then the logic in vos will mark corruption on both instead of just on dkey. Because if some day we support to clear corruption flag, then if the user only clear such flag from dkey, and if we did not set corruption on the akey, then we will lost corruption flag on the akey. That seems wrong.
Currently, it is the upper layer caller's duty to specify the parameter properly to avoid additional work.

gnailzenh (approved):

src/vos/ilog.c
```diff
@@ -1560,6 +1611,18 @@ ilog_aggregate(struct umem_instance *umm, struct ilog_df *ilog,
 	return empty;
 }
 
+bool
+ilog_is_corrupted(struct ilog_df *ilog_df)
+{
+	struct ilog_root *root  = (struct ilog_root *)ilog_df;
+	int               flags = ilog_mag2flags(root->lr_magic);
+
+	if (flags < 0)
```
liuxuezhao, line 1620: maybe a ERR/CRIT log is helpful here.

need it differentiate the cases of 1) invalid flags and 2) valid flag but with ILOG_FLAGS_CORRUPTED?

src/vos/vos_internal.h
```diff
@@ -1366,6 +1366,10 @@ key_tree_punch(struct vos_object *obj, daos_handle_t toh, daos_epoch_t epoch,
 	       struct vos_ilog_info *parent, struct vos_ilog_info *info);
 int
 key_tree_delete(struct vos_object *obj, daos_handle_t toh, d_iov_t *key_iov);
+int
+vos_tree_mark_corruption(struct vos_container *cont, struct vos_object *obj, daos_handle_t toh,
+			 enum vos_tree_class tclass, daos_epoch_t epoch, uint32_t pm_ver,
```
liuxuezhao, line 1371: just confirm the meaning of "pm_ver",
for example if calls with vos_tree_mark_corruption(..., 10, ..);
does it mean the obj corrupted start from pm ver 10, and is not corrupted for pm_ver < 10?
the ilog_is_corrupted() func without pm_ver arg, so seems also corrupted for pm_ver < 10

src/vos/vos_obj.c
```diff
@@ -803,6 +805,115 @@ vos_obj_del_key(daos_handle_t coh, daos_unit_oid_t oid, daos_key_t *dkey,
...
+		D_GOTO(log, rc = -DER_INVAL);
+
+	if (akey_nr != 0) {
+		if (dkey == NULL)
+			D_GOTO(log, rc = -DER_INVAL);
+
+		if (akeys == NULL)
+			D_GOTO(log, rc = -DER_INVAL);
+
+		if (vos_obj_skip_akey_supported(cont, oid))
```
liuxuezhao, line 836: i feel some confuse, "vos_obj_skip_akey_supported" means the feature VOS_POOL_FEAT_FLAT_DKEY supported or not (supported since 2.6.0), rather than the specific dkey is flat dkey or not.

not quite sure what's the purpose of this check. If marking a flat dkey's only akey corrupted, can it work?

liuxuezhao: hmm, it looks incorrect to me, vos_obj_skip_akey_supported() seems always true after 2.6.0?

src/vos/ilog.c
Nasf-Fan: Yes, that can be enhanced: `flags < 0` means `ilog` itself corrupted. Under such case, we have no way to mark corruption.

src/vos/vos_internal.h
Nasf-Fan: Once object/key is marked as corrupted (in spite of which which version), then it is corrupted for all versions because `ilog magic` have no way to distinguish `pm_ver` currently. So if we want to recovery related object/key that has been marked as corruption, we have to delete them firstly. We may support to clear such flag in the future, but have to consider how to guarantee no data lost.

src/vos/vos_obj.c
Nasf-Fan:
```
static inline bool
vos_obj_skip_akey_supported(struct vos_container *cont, daos_unit_oid_t oid)
{       
        struct vos_pool *pool = vos_cont2pool(cont);
        
        if ((pool->vp_feats & VOS_POOL_FEAT_FLAT_DKEY) == 0)
                return false;
        
        if (daos_is_array(oid.id_pub) || daos_is_kv(oid.id_pub))
                return true;
        
        return false;
}
```       
It does not only check `pool->vp_feats`, but also check `oid`.

liuxuezhao: looks you are right, sorry for noise.

PR #16495, opened 2025-06-10
DAOS-17666 pool: Fix property self_heal.exclude

liw:
Flag "exclude" of pool property "self_heal" has no effect, as shown by the experiment described in the Jira ticket. This patch checks the flag before excluding a rank from a pool, and restarts exclusion when "self_heal" is set to include "exclude".

### Steps for the author:

* [ ] Commit message follows the [guidelines](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Appropriate [Features or Test-tag](https://daosio.atlassian.net/wiki/spaces/DC/pages/10984259629/Test+Tags) pragmas were used.
* [ ] Appropriate [Functional Test Stages](https://daosio.atlassian.net/wiki/spaces/DC/pages/12147556353/CI+Functional+Test+Stages) were run.
* [ ] At least two positive code reviews including at least one code owner from each category referenced in the PR.
* [ ] Testing is complete. If necessary, forced-landing label added and a reason added in a comment.

#### After all prior steps are complete:
* [ ] Gatekeeper requested (daos-gatekeeper added as a reviewer).

PR #16538, opened 2025-06-25
DAOS-17727 pool: add a new option to pool reintegration

wangshilong:
During system maintenance windows, some ranks might be excluded due to switch reboots or network instability (while the number of failed ranks remains below the redundancy factor).

Although rebuild operations are triggered in such cases, there might be a potential data safety risk if additional ranks fail (e.g., due to SSD failures) during the rebuild process.

To mitigate this risk and avoid complexity for data recovery, we propose adding a new reintegration option(--no-migration) that skips data migration and directly brings the previously down ranks back online. Crucially, the rebuild reclaim phase must still be performed to free space from prior incomplete rebuild operations.This approach is safe only if we can confirm there is no inflight I/O during maintenance windows.

### Steps for the author:


daosbuild3:
Test stage Unit Test on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16538/1/testReport/

src/control/lib/control/pool.go
```diff
@@ -969,9 +973,53 @@ func poolReintegrateRank(ctx context.Context, rpcClient UnaryInvoker, req *PoolR
 	return getPoolRankResult(resp.Status, rank), nil
 }
 
+// PoolReintegrate will set pool targets on specified ranks back to up state which should
+// automatically start the reintegration process. Returns pool-ranks response and error.
+func PoolReintegrateRanks(ctx context.Context, rpcClient UnaryInvoker, req *PoolRanksReq) (*PoolRanksResp, error) {
```
wangshilong, line 978: @tanabarr This is not elegant, PR is for 2.6.5  but for this specific feature(new --no-migration), i need pass several ranks to engine at the same time and reintegration command should bring ranks to UPIN at the same time.

wangshilong:
@liuxuezhao @kccain

I tested the PR using 5 ranks, created a pool with rd_fac:2 (DAOS_POOL_RF=2), then terminated two ranks. After confirming rebuild initiation via 'dmg pool query -t' (dead_ranks/disabled_ranks appeared as expected), I terminated another rank. After restarting the first two ranks, I executed "dmg pool reintegration --no-migrations --ranks <>" and observed successful rebuild completion with verified data consistency.
Regarding the code behavior:
When rebuild completes, update_one_tgt() handles MAP_FINISH_REBUILD cases as follows:

case MAP_FINISH_REBUILD:
          switch (target->ta_comp.co_status) {
          case PO_COMP_ST_UPIN:
          case PO_COMP_ST_DOWNOUT:
          case PO_COMP_ST_NEW:
           /* Nothing to do */
          D_INFO(DF_MAP ": Skip FINISH_REBUILD " DF_TARGET "\n",
                         DP_MAP(pool_uuid, map), DP_TARGET(target));
           break;
Since reintegration sets targets to UPIN status, skipping the update is correct. The subsequent rebuild_task_complete_schedule() properly initiates RECLAIM.

Should we return an error here, or maintain the current behavior?

wangshilong:
Please hold on review, PR did not pass test yet.

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16538/10/execution/node/1338/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16538/11/testReport/

src/common/pool_map.c
```diff
@@ -3250,6 +3250,24 @@ pool_target_id_list_alloc(unsigned int num,
 	return 0;
 }
 
+int
```
kccain, line 3253: function is added but not used (as far as I can see). Should it be removed from the patch?

src/pool/srv_pool_map.c
```diff
@@ -152,9 +152,15 @@ update_one_tgt(uuid_t pool_uuid, struct pool_map *map, struct pool_target *targe
...
+					DP_MAP(pool_uuid, map), DP_TARGET(target));
+				target->ta_comp.co_status = PO_COMP_ST_UPIN;
+			} else {
+				D_DEBUG(DB_MD, DF_MAP ": change " DF_TARGET " to UP\n",
+					DP_MAP(pool_uuid, map), DP_TARGET(target));
+				target->ta_comp.co_status = PO_COMP_ST_UP;
+			}
 			target->ta_comp.co_in_ver = ++(*version);
 			if (print_changes)
 				D_PRINT(DF_MAP ": " DF_TARGET " start reintegration.\n",
```
kccain, line 166: maybe customize the D_PRINT/D_INFO message to indicate if reintegration is with or without migration?

src/pool/srv_pool.c
```diff
@@ -7459,11 +7571,44 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
...
+	bool                             safe_reint   = false;
+	bool                             no_data_sync = false;
+
+	rc = ds_pool_iv_prop_fetch(svc->ps_pool, &prop);
+	if (rc)
+		D_GOTO(out, rc);
+
+	entry = daos_prop_entry_get(&prop, DAOS_PROP_PO_SELF_HEAL);
+	D_ASSERT(entry != NULL);
+	if (!(entry->dpe_val & (DAOS_SELF_HEAL_AUTO_REBUILD | DAOS_SELF_HEAL_DELAY_REBUILD))) {
```
kccain, line 7583: if self_heal = "exclude" (instead of "exclude,rebuild") then I think the call to pool_svc_update_map_internal() is still needed below, to update the pool map. i.e. the code needs to look at the DAOS_SELF_HEAL_AUTO_EXCLUDE bit to know whether the pool map update is required (even if rebuild is disabled according to the DAOS_SELF_HEAL_DELAY_REBUILD bit not set). Do I have it right @liw ?

src/pool/srv_pool.c
```diff
@@ -7459,11 +7571,44 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
...
+	if (safe_reint || (no_data_sync && opc == MAP_REINT)) {
+		D_ASSERT(exclude_rank == false);
+		D_ASSERT(extend_rank_list == NULL);
+		D_ASSERT(extend_domains == NULL);
+		rc = ds_pool_tgt_safe_reint(svc, list, inval_list_out, map_version, hint, src,
+					    flags | POOL_TGT_UPDATE_NO_MIGRATION);
+		D_GOTO(out, rc);
+	}
+
+	if (no_data_sync) {
```
kccain, line 7603: no_data_sync is always so confusing, I think it is meant to be removed at some point?
Also, for opc like MAP_EXCLUDE and MAP_DRAIN, do we want to make the pool map changes (and maybe also do the ds_pool_tgt_exclude_out() like in the original code before this change?

src/pool/srv_pool.c
```diff
@@ -7404,7 +7405,7 @@ pool_update_map_internal(uuid_t pool_uuid, unsigned int opc, bool exclude_rank,
 
 	rc = pool_svc_update_map_internal(svc, opc, exclude_rank, NULL, 0, NULL, tgts, tgt_addrs,
 					  hint, p_updated, map_version_p, tgt_map_ver,
-					  inval_tgt_addrs, MUS_DMG, true);
+					  inval_tgt_addrs, MUS_DMG, 0);
```
kccain, line 7408: why does this not specify flags=POOL_TGT_UPDATE_SKIP_RF_CHECK in the final argument (to agree with skip_rf_check=true in the original code?

I realize there were 3 uses of pool_update_map_internal() previously, and one of them (related to no_data_sync) was removed in this patch - is that the only usage that required skip RF check?

src/pool/srv_pool.c
```diff
@@ -7438,6 +7439,117 @@ ds_pool_tgt_revert_rebuild(uuid_t pool_uuid, struct pool_target_id_list *list)
...
+	struct pool_target_id_list target_list = {0};
+	uint32_t                   tgt_map_ver = 0;
+	daos_epoch_t               rebuild_eph = d_hlc_get();
+	bool no_data_sync = (svc->ps_pool->sp_reint_mode == DAOS_REINT_MODE_NO_DATA_SYNC);
+
+	if (!no_data_sync) {
+		d_rank_list_t *dead_rank_list = NULL;
+		/*
+		 * Safe reintegration mode is only allowed when:
+		 * 1. Pool rebuild is in progress
```
kccain, line 7488: I'm trying to remember the exact scenario. There may be a case where pool rebuild is *not* running, but the admin will want to perform a reintegrate --no-migration to recover the pool.

let's say DAOS_POOL_RF=2, and we have failure event A with 2 fault domains affected - 2 nodes engines being excluded from the pool. This goes for some time, and admin decides to shut down the system before the rebulid finishes. Then, on system restart, a third node has an immediate failure (failure event B). I think in this condition, the pool_map_crit_prompt() function will emit the urgent message alerting the administrator that a rebuild will not be started for this latest pool map change. So, some admin command will be needed to cause a rebuild to be triggered.

kccain (commented):
first-pass review with some questions. Just based on the C code changes so far. This is a big/complex change.

src/pool/srv_pool.c
liw: @kccain, I agree. (In https://github.com/daos-stack/daos/pull/16495, I'm checking `exclude` of pool property `self_heal` inside `pool_svc_update_map_internal`.) Looks like the `if (no_data_sync)` statement on line 7603 has the same issue. Is that right?

src/pool/srv_pool.c
wangshilong: After restarting, system will resume event A's rebuild? or rebuild will not be triggered because of event B and it broke RF.

src/pool/srv_pool.c
wangshilong: You are right, POOL_TGT_UPDATE_SKIP_RF_CHECK should be added here.

src/pool/srv_pool.c
wangshilong: Yup, we intent to remove it, maybe i could do it in this PR.

src/pool/srv_pool_map.c
wangshilong: Good idea.

src/common/pool_map.c
wangshilong: Yes, should be removed.

src/pool/srv_pool.c
wangshilong: OK, i'll fix above issue.

src/pool/srv_pool.c
kccain: I think it will resume event A rebuild after restarting. However, IIRC, the scenario was that these rebuilds could fail (-DER_NOMEM failures occurred on rebuilds of certain pools). If this happens before dmg pool reintegrate --no-migration is run, then I guess the rebuild would not be triggered. Maybe this would require a restart and then having the admin run the reintegrate --no-migration while the event A excludes rebuild is running?

Maybe we can come up with a test exercise to see how this would work out.

src/control/lib/control/pool.go
tanabarr: I don't think I fully understand your comment but the previous change to implement multiple ranks in reint/exclude/drain was made in DAOS-6611 commit f4e401fb7b5dd1bfa9a85a885dce008b8fa550be intentionally performed iteration over ranks in the control-plane rather than in the engine. This decision was made after discussion with Mike and Niu and concerned error handling and management of complexity. Please don't revert these changes by adding ranks to the proto-definition before we have more discussion on the ramifications of the change and make consistent changes with the other drain and exclude dRPCs as well as altering the relevant control-plane code. Thanks

tanabarr (changes requested):

PR #16557, opened 2025-06-30
DAOS-17748 control: Add verification to raft-db Add/Remove wrappers

tanabarr:
Features: control

### Steps for the author:

### DIFF
--- a/src/control/system/membership.go
+++ b/src/control/system/membership.go
@@ -204,8 +204,8 @@ func (m *Membership) joinReplace(req *JoinRequest) (*JoinResponse, error) {
 	memberToReplace := &Member{}
 	*memberToReplace = *cm
 
-	m.log.Debugf("replace-rank: updating member with UUID %s->%s", memberToReplace.UUID,
-		req.UUID)
+	m.log.Debugf("replace-rank: updating member with UUID %s->%s, Addr %s",
+		memberToReplace.UUID, req.UUID, memberToReplace.Addr)
 
 	if err := m.db.RemoveMember(cm); err != nil {
 		return nil, errors.Wrap(err, "removing old member in replace-rank join request")
--- a/src/control/system/raft/database_members.go
+++ b/src/control/system/raft/database_members.go
@@ -1,5 +1,6 @@
 //
 // (C) Copyright 2020-2022 Intel Corporation.
+// (C) Copyright 2025 Hewlett Packard Enterprise Development LP
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 //
@@ -47,28 +48,68 @@ func (mrm MemberRankMap) MarshalJSON() ([]byte, error) {
 	return json.Marshal(jm)
 }
 
-func (mam MemberAddrMap) addMember(addr *net.TCPAddr, m *system.Member) {
-	if _, exists := mam[addr.String()]; !exists {
-		mam[addr.String()] = []*system.Member{}
+// Remove members with duplicate UUIDs from MemberAddrMap.
+func (mam MemberAddrMap) dedupMembers(addrStr string) {
+	//oldMembers []*system.Member) []*system.Member {
+	seen := make(map[string]struct{})
+	dupes := make(map[string]struct{})
+
+	for _, m := range mam[addrStr] {
+		us := m.UUID.String()
+
+		if _, exists := seen[us]; exists {
+			dupes[us] = struct{}{}
+		} else {
+			seen[us] = struct{}{}
+		}
 	}
-	mam[addr.String()] = append(mam[addr.String()], m)
+	if len(dupes) == 0 {
+		return
+	}
+
+	// Update with de-duplicated slice only if needed.
+	newMembers := []*system.Member{}
+	for _, m := range mam[addrStr] {
+		if _, exists := dupes[m.UUID.String()]; !exists {
+			newMembers = append(newMembers, m)
+		}
+	}
+
+	mam[addrStr] = newMembers
+}
+
+func (mam MemberAddrMap) addMember(addr *net.TCPAddr, m *system.Member) {
+	as := addr.String()
+
+	if _, exists := mam[as]; !exists {
+		mam[as] = []*system.Member{}
+	}
+
+	mam[as] = append(mam[as], m)
+	mam.dedupMembers(as)
 }
 
 func (mam MemberAddrMap) removeMember(m *system.Member) {
-	members, exists := mam[m.Addr.String()]
-	if !exists {
+	mas := m.Addr.String()
+
+	if _, exists := mam[mas]; !exists {
 		return
 	}
-	for i, cur := range members {
-		if m.UUID == cur.UUID {
-			// remove from slice
-			members = append(members[:i], members[i+1:]...)
-			break
+
+	mam.dedupMembers(mas)
+
+	newMembers := []*system.Member{}
+	for _, cur := range mam[mas] {
+		if m.UUID != cur.UUID {
+			nm := *cur
+			newMembers = append(newMembers, &nm)
 		}
 	}
-	if len(members) == 0 {
-		delete(mam, m.Addr.String())
+	if len(newMembers) == 0 {
+		delete(mam, mas)
 	}
+
+	mam[mas] = newMembers
 }
 
 // MarshalJSON creates a serialized representation of the MemberAddrMap.
@@ -170,7 +211,7 @@ func (mdb *MemberDatabase) updateMember(m *system.Member) {
 	mdb.addToFaultDomainTree(cur)
 }
 
-// removeMember is responsible for removing new Member and updating all
+// removeMember is responsible for removing Member and updating all
 // of the relevant maps.
 func (mdb *MemberDatabase) removeMember(m *system.Member) {
 	delete(mdb.Ranks, m.Rank)


tanabarr:
early reviews appreciated, still need to add unit tests but would like feedback on concept please

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16557/1/execution/node/1468/log

mjmac:
I'm confused about the approach in this PR. Based on the conversation in slack, I thought that we had clearly identified the issue as an oversight in the implementation of [MemberDatabase.updateMember](https://github.com/daos-stack/daos/blob/master/src/control/system/raft/database_members.go#L156), in that it needs to update the secondary map of member addresses `mdb.Addrs` when a member is updated in the database.

Unless I'm missing something, I think the fix should be as simple as just adding an `updateMember` method to the [MemberAddrMap](https://github.com/daos-stack/daos/blob/master/src/control/system/raft/database_members.go#L26) type and then calling it from `MemberDatabase.updateMember`.

src/control/system/membership.go
```diff
@@ -228,6 +268,10 @@ func (m *Membership) joinReplace(req *JoinRequest) (*JoinResponse, error) {
 		return nil, err
 	}
 
+	if err := m.db.ForceSnapshot(); err != nil {
+		return nil, errors.Wrap(err, "forcing snapshot after join-replace")
+	}
+
```
kjacque: After the more recent discussions about this, I don't think this is necessary. The snapshots aren't the problem, the core issue was that the maps were inconsistent.

src/control/system/membership.go
```diff
@@ -204,12 +241,15 @@ func (m *Membership) joinReplace(req *JoinRequest) (*JoinResponse, error) {
...
-		req.UUID)
+	m.log.Debugf("replace-rank: updating member with UUID %s->%s, Addr %s",
+		memberToReplace.UUID, req.UUID, memberToReplace.Addr)
 
 	if err := m.db.RemoveMember(cm); err != nil {
 		return nil, errors.Wrap(err, "removing old member in replace-rank join request")
 	}
+	if err := m.checkMemberRemoved(cm); err != nil {
+		return nil, errors.Wrap(err, "checking remove member in replace-rank join request")
+	}
```
kjacque: Instead of adding a check in production code, IMO we should create a unit test that calls RemoveMember and checks that it was removed from all maps.

src/control/system/raft/database.go
```diff
@@ -52,6 +53,7 @@ type (
 		LeadershipTransfer() raft.Future
 		Barrier(time.Duration) raft.Future
 		Shutdown() raft.Future
+		Snapshot() raft.SnapshotFuture
```
kjacque: Could probably be removed since ForceSnapshot is not needed.

src/control/system/raft/database_members.go
```diff
@@ -47,28 +48,55 @@ func (mrm MemberRankMap) MarshalJSON() ([]byte, error) {
 	return json.Marshal(jm)
 }
 
+// Remove members with duplicate UUIDs from MemberAddrMap.
+func (mam MemberAddrMap) dedupeMembers(addrStr string) {
```
kjacque: Huh. Does this become a problem, adding multiple members with the same UUID?

In a replace, wouldn't we want the new one to replace the old one?

src/control/system/raft/database_members.go
```diff
@@ -47,28 +48,55 @@ func (mrm MemberRankMap) MarshalJSON() ([]byte, error) {
...
+}
+
 func (mam MemberAddrMap) addMember(addr *net.TCPAddr, m *system.Member) {
-	if _, exists := mam[addr.String()]; !exists {
-		mam[addr.String()] = []*system.Member{}
+	as := addr.String()
+
+	if _, exists := mam[as]; !exists {
+		mam[as] = system.Members{}
 	}
```
kjacque: Probably not necessary, appending to a nil slice creates a new slice so it should be safe.

src/control/system/raft/raft.go
```diff
@@ -189,6 +190,16 @@ func (db *Database) ShutdownRaft() error {
 	})
 }
 
+// ForceSnapshot signals that the raft implementation should manually trigger snapshot.
+// Blocks until the snapshot is complete.
+func (db *Database) ForceSnapshot() error {
```
kjacque: Again, could probably be removed.

src/control/system/raft/database_members.go
```diff
@@ -47,28 +48,55 @@ func (mrm MemberRankMap) MarshalJSON() ([]byte, error) {
...
-		if m.UUID == cur.UUID {
-			// remove from slice
-			members = append(members[:i], members[i+1:]...)
-			break
+
+	mam.dedupeMembers(mas)
+
+	newMembers := system.Members{}
+	for _, cur := range mam[mas] {
+		if m.UUID != cur.UUID {
```
kjacque, line 66: Actually... Isn't this where we run into a problem? The UUID has changed, but it's an existing member we wish to remove/re-add (maybe this would fix the dedupe issue?). Maybe we should be matching by either UUID _or_ rank? Or maybe we really should be persisting the UUID between replacements, if we believe it is the same node.

tanabarr: The issue seems fixed with this code and I have verified on FTC clusters. I'd like to land as is then we can refactor after. It is a P1 issue and we need to get the fix in ASAP. @Michael-Hennecke do you think this is the right approach?

src/control/system/membership.go
tanabarr: ok will revert

src/control/system/raft/database.go
tanabarr: reverted

src/control/system/raft/database_members.go
tanabarr: true, was just trying to be extra explicit to fix the cause of the inconsistency, removed

src/control/system/raft/raft.go
tanabarr: reverted

src/control/system/raft/database_members.go
tanabarr: Regarding this code, `src/control/system/membership.go:joinReplace()` first calls `RemoveMember()` then  `AddMember()`. With this approach there should be no issue with UUID mappings as long as the remove and add operations are clean. @mjmac I don't think the fix is as simple as adding addr-map update function because `raftOpUpdateMember` doesn't get used in the `joinReplace()` function, only `raftOpRemoveMember` and `raftOpAddMember`. I would like to keep the fix intact, adjust some debug and revert the `ForceSlapshot()` change if that is blocking the review. As mentioned before I've verified the fix and would like to get it landed ASAP to unblock customer deployments.

tanabarr: I added `MemberAddrMap.updateMember()` for completeness and after reviewing the fix again I think it's safe performing checks using UUID as long as updates during join are made with Remove then Add.

tanabarr:
> I'm confused about the approach in this PR. Based on the conversation in slack, I thought that we had clearly identified the issue as an oversight in the implementation of [MemberDatabase.updateMember](https://github.com/daos-stack/daos/blob/master/src/control/system/raft/database_members.go#L156), in that it needs to update the secondary map of member addresses `mdb.Addrs` when a member is updated in the database.
> 
> Unless I'm missing something, I think the fix should be as simple as just adding an `updateMember` method to the [MemberAddrMap](https://github.com/daos-stack/daos/blob/master/src/control/system/raft/database_members.go#L26) type and then calling it from `MemberDatabase.updateMember`.

`updateMember()` doesn't get called during join-replace, see other comments.

src/control/system/raft/database_members.go
tanabarr: Somehow we end up with duplicate members in the previous implementation, simplifying things and being overly cautious in this change addresses the issue.  Yes in replace we do remove->verify->add.

src/control/system/membership.go
tanabarr: For the initial fix change I would like to do both, I've added the extensive testing for member database methods which improves the coverage and I'd like to leave this ugly but comprehensive check in (can be removed later when we are sure the problem has gone away).

knard38: From my side, I am in favor to keep this extra-check: From the sustaining point of view, it is better to know that something went wrong when we do the replace instead of the next system restart (as it occurred  at LRZ).

src/control/system/raft/database_members.go
```diff
@@ -47,28 +48,62 @@ func (mrm MemberRankMap) MarshalJSON() ([]byte, error) {
...
-		mam[addr.String()] = []*system.Member{}
+// dedupeMembers removes members with duplicate UUIDs from MemberAddrMap.
+func (mam MemberAddrMap) dedupeMembers(addrStr string) {
+	seen := make(map[string]struct{})
+
+	newMembers := system.Members{}
+	for _, m := range mam[addrStr] {
+		us := m.UUID.String()
+
+		if _, exists := seen[us]; !exists {
```
knard38: I am probably missing something but how do we know we are keeping the good member.

knard38 (approved):
Overall approach sounds good to me, but I am not familiar with this part of the code.

src/control/system/raft/database_members.go
tanabarr: The only plausible reason for having a duplicate is that identical member was added. We could verify all member fields before removing the duplicate and panic if not identical but I think this is overkill. @mjmac @kjacque thoughts?

src/control/system/membership.go
kjacque: OK, I think that's fair. Let's leave it in for now.

src/control/system/raft/database_members.go
kjacque: Since we're going by UUID here, I think this is OK as-is. The new member has a different UUID.
In a refactoring, I might instead structure the group of members as a map based on UUID. This would ensure no duplicates.

src/control/system/raft/database_members.go
kjacque: I think I got a little confused here about whether we were passing the old version of the member and the new. It looks like even in the old code, we were explicitly removing the old member with the old UUID. So I'm still at a loss for how the duplication was happening.

src/control/system/raft/database_members.go
```diff
@@ -22,8 +23,8 @@ type (
 	MemberRankMap map[ranklist.Rank]*system.Member
 	// MemberUuidMap provides a map of UUID->*system.Member.
 	MemberUuidMap map[uuid.UUID]*system.Member
-	// MemberAddrMap provides a map of string->[]*system.Member.
-	MemberAddrMap map[string][]*system.Member
+	// MemberAddrMap provides a map of string->system.Members.
+	MemberAddrMap map[string]system.Members
```
mjmac: Has this change been verified to work ok with snapshots written from before the change? `system.Members` is a type alias, so it will _probably_ work OK, but it would be a really good idea to generate a fixture with the older format and make sure that it unmarshals correctly on snapshot load. Otherwise you're going to be introducing an incompatibility (e.g. 2.6.4 server can't read 2.6.3 snapshots).

src/control/system/raft/database_members.go
```diff
@@ -47,28 +48,62 @@ func (mrm MemberRankMap) MarshalJSON() ([]byte, error) {
...
-	mam[addr.String()] = append(mam[addr.String()], m)
+
+	mam[addrStr] = newMembers
+}
+
+func (mam MemberAddrMap) addMember(addr *net.TCPAddr, m *system.Member) {
+	as := addr.String()
+
+	mam[as] = append(mam[as], m)
+	mam.dedupeMembers(as)
```
mjmac: I'm going to reiterate my comments from standup, here. Succinctly, it seems wrong to me that this dedupe functionality should be called as part of regular membership management rather than fixing the broken code that allows the duplicates to be introduced in the first place. This code addresses a specific symptom seen in this scenario, but there may be other symptoms that have not yet been encountered. Without understanding and addressing the root of the problem, how can you have confidence that there won't be other production outages?

I do think it would be useful to keep the dedupe and other repair functionality developed to address the bug as part of a CR command, though.

src/control/system/raft/database_test.go
```diff
@@ -431,6 +431,51 @@ func ignoreFaultDomainIDOption() cmp.Option {
 		}, cmp.Ignore())
 }
 
+func checkMemberDB(t *testing.T, mdb *MemberDatabase, expMember *Member, opts ...cmp.Option) {
```
mjmac, line 434: This could be exposed as part of a `daos_server ms check` command. An optional `--repair` flag would then attempt to fix any inconsistencies found. On the one hand, repairing silently is potentially more convenient, but on the other hand, there may be other serious problems that are hidden until they're not.

src/control/system/raft/database_members.go
johannlombardi: Sounds sensible. Let's discuss this in more details

tanabarr: The fix covers an observed problem that the DAOS raft library wrappers exhibit. The symptom is well defined and the join-replace functionality relatively simple. I have manually verified the fix extensively with @knard38 and observed that we can resolve the inaccuracies in the generated snapshot using a small update to the removeMember logic. I have verified the fix without the dedupe function. Both customers are eagerly awaiting the replace feature. The fix should be landed to 2.6.4 ASAP. @kjacque has also written a unit test which I have integrated into the PR, thanks! I have also removed the extra verification in join-replace as I don't think it's needed any more.

src/control/system/raft/database_members.go
tanabarr: reverted

kjacque (approved):
LGTM. Thanks for all the work on this!

knard38 (approved):
Nice simplification :-)

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16557/6/execution/node/1456/log

PR #16559, opened 2025-07-01
DAOS-17354 rebuild: interactive stop/start logic and testing

kccain:
Features: rebuild test_daos_rebuild_simple_interactive
Allow-unstable-test: true

### Steps for the author:


daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16559/2/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/2/execution/node/1400/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/2/execution/node/1344/log

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/2/execution/node/1541/log

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/3/execution/node/1455/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/3/execution/node/1545/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16559/6/testReport/

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16559/8/testReport/

kccain:
> Test stage NLT on EL 8.8 completed with status UNSTABLE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16559/8/testReport/

For this intermittent NLT failure seen so far just in this PR, I filed new ticket https://daosio.atlassian.net/browse/DAOS-17795 NLT valgrind invalid reads/writes in commands "daos fs fix-entry" and "daos fs check"

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/9/execution/node/1090/log

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/9/execution/node/1135/log

kccain:
> Test stage Functional Hardware Large MD on SSD completed with status FAILURE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/9/execution/node/1135/log

another instance of https://daosio.atlassian.net/browse/DAOS-17771 that has affected this PR in 2 different builds.
also see https://daosio.atlassian.net/browse/SRE-3197 with plans to take that cluster offline for the time being.

kccain:
> Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/9/execution/node/1090/log

test_daos_rebuild_ec overall test timeout reached, an instance of https://daosio.atlassian.net/browse/DAOS-17578

kccain:
At this point in testing the PR, remaining effort is to keep kicking off tests to try to avoid either CI cluster issues or known test failures as documented above. So I think it is reasonable now to open the PR up for reviews.

src/tests/ftest/daos_test/suite.py
```diff
@@ -302,6 +302,22 @@ def test_daos_rebuild_simple(self):
         """
         self.run_subtest()
 
+    def test_daos_rebuild_simple_interactive(self):
+        """Jira ID: DAOS-1568
```
daltonbohning: Nit - this gets copy pasted wrong all over but it should be tied to THIS PR. Or IMO you could leave it out since the git history has this info
```suggestion
        """Jira ID: DAOS-17354
```

src/tests/ftest/daos_test/suite.py
```diff
@@ -302,6 +302,22 @@ def test_daos_rebuild_simple(self):
...
+
+        Test Description:
+            Run daos_test -v --rebuild_interactive
+
+        Use cases:
+            Core tests for daos_test
+
+        :avocado: tags=all,full_regression
+        :avocado: tags=hw,medium,provider
+        :avocado: tags=daos_test,daos_core_test
```
daltonbohning: We aren't always on top of this, but since this is a weekly test, we should consider: how/when will this run in a PR?
Tagging with `rebuild` is probably the simplest so people can use `Features: rebuild` without thinking about THIS test in particular.
Anyway, I'll let you decide
```suggestion
        :avocado: tags=daos_test,daos_core_test,rebuild
```

daltonbohning (commented):
ftest LGTM

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16559/10/execution/node/541/log

src/tests/ftest/daos_test/suite.py
kccain: Yes I think adding the rebuild tag is a good idea here. It would also be good for test_daos_rebuild_simple and test_daos_rebuild_ec, but those are already covered by "pr" so there is less urgency I guess to add rebuild tag there.

daltonbohning: Less urgency I agree but FWIW you can add them if you find it helpful

src/include/daos_srv/pool.h
```diff
@@ -404,6 +404,10 @@ struct ds_pool *ds_pool_svc2pool(struct ds_pool_svc *ds_svc);
 struct cont_svc *ds_pool_ps2cs(struct ds_pool_svc *ds_svc);
 void ds_pool_disable_exclude(void);
 void ds_pool_enable_exclude(void);
+void
+ds_pool_rebuild_start(uuid_t pool_uuid, struct rsvc_hint *hint);
+void
+	    ds_pool_rebuild_stop(uuid_t pool_uuid, struct rsvc_hint *hint);
```
gnailzenh, line 410: unexpected indentation.

src/rebuild/rebuild_iv.c
```diff
@@ -10,7 +10,7 @@
  * This file contains the server API methods and the RPC handlers for rebuild
  * initiator.
  */
-#define D_LOGFAC	DD_FAC(rebuild)
```
gnailzenh: unintended change?

src/rebuild/rebuild_iv.c
```diff
@@ -30,7 +30,7 @@
 static int
 rebuild_iv_alloc_internal(d_sg_list_t *sgl)
 {
-	int	rc;
+	int rc;
```
gnailzenh: Looks like you are trying to unify the code format, I like this type of work but please avoid to piggyback them on complex patch like this in the future. 

We can either have a dedicated patch or piggyback these changes on simple fixes.

src/rebuild/rebuild_iv.c
```diff
@@ -229,66 +225,65 @@ rebuild_iv_ent_refresh(struct ds_iv_entry *entry, struct ds_iv_key *key,
...
-	.ivc_ent_refresh	= rebuild_iv_ent_refresh,
-	.ivc_value_alloc	= rebuild_iv_alloc,
+    .ivc_ent_init    = rebuild_iv_ent_init,
+    .ivc_ent_get     = rebuild_iv_ent_get,
+    .ivc_ent_put     = rebuild_iv_ent_put,
+    .ivc_ent_destroy = rebuild_iv_ent_destroy,
+    .ivc_ent_fetch   = rebuild_iv_ent_fetch,
+    .ivc_ent_update  = rebuild_iv_ent_update,
+    .ivc_ent_refresh = rebuild_iv_ent_refresh,
+    .ivc_value_alloc = rebuild_iv_alloc,
```
gnailzenh: use tabs please

src/pool/srv_pool.c
```diff
@@ -8234,6 +8234,74 @@ ds_pool_svc_stop_handler(crt_rpc_t *rpc)
...
+	struct pool_svc *svc;
+	int              rc;
+
+	D_INFO(DF_UUID ": received request to stop rebuild\n", DP_UUID(pool_uuid));
+
+	rc = pool_svc_lookup_leader(pool_uuid, &svc, hint);
+	if (rc != 0)
+		D_GOTO(out, rc);
+
+	ds_rebuild_admin_stop(pool_uuid);
```
gnailzenh: I'm not familiar with this part of code, but I think you might want to check the current rebuild status, for example, what if it has already been stopped, or return proper information back to admin if there is no in-progress rebuild?

src/pool/srv_pool.c
```diff
@@ -8234,6 +8234,74 @@ ds_pool_svc_stop_handler(crt_rpc_t *rpc)
...
+	}
+
+	rc = ds_pool_iv_prop_fetch(pool, &prop);
+	if (rc) {
+		daos_prop_fini(&prop);
+		DL_ERROR(rc, DF_UUID ": cannot fetch properties", DP_UUID(pool_uuid));
+		goto out_pool;
+	}
+
+	rc = ds_rebuild_regenerate_task(pool, &prop);
```
gnailzenh: same here, I think we should check status of the pool and report reasonable information back to admin if there operation can't be executed?

src/rebuild/srv.c
```diff
@@ -885,13 +927,28 @@ rebuild_leader_status_check(struct ds_pool *pool, uint32_t op,
...
+		int                         i;
+
+		if (!fi_admin_stop_fi)
+			fi_admin_stop_fi = DAOS_FAIL_CHECK(DAOS_REBUILD_ADMIN_STOP);
+		fi_admin_stop = (fi_admin_stop_fi && (op == RB_OP_REBUILD));
+
+		D_INFO(DF_RB ": fi_admin_stop_fi=%d, fi_admin_stop=%d, nsleeps=%u\n",
+		       DP_RB_RGT(rgt), fi_admin_stop_fi, fi_admin_stop, nsleeps);
+
+		if (fi_admin_stop && (nsleeps > 0)) {
```
gnailzenh, line 952: Could you explain what nsleeps does at here?

kccain:
Need engine code reviews still.

src/pool/srv_pool.c
```diff
@@ -8234,6 +8234,74 @@ ds_pool_svc_stop_handler(crt_rpc_t *rpc)
...
 }
 
+/* TODO: make a real RPC handler ds_pool_rebuild_stop_handler(crt_rpc *rpc) to call this. */
+void
+ds_pool_rebuild_stop(uuid_t pool_uuid, struct rsvc_hint *hint)
+{
+	struct pool_svc *svc;
+	int              rc;
+
+	D_INFO(DF_UUID ": received request to stop rebuild\n", DP_UUID(pool_uuid));
```
liw: [Nit] Since the `pool_svc_lookup_leader` call below may return `-DER_NOTLEADER` for 7 to 14 s, would it be more appropriate to use `D_DEBUG` here? If we'd like to log an `INFO` for the stop event, which makes sense to me, then a better place to do so might be somewhere in/during `ds_rebuild_admin_stop`?

src/pool/srv_pool.c
```diff
@@ -8234,6 +8234,74 @@ ds_pool_svc_stop_handler(crt_rpc_t *rpc)
...
+	if (rc != 0)
+		D_GOTO(out, rc);
+
+	ds_rebuild_admin_stop(pool_uuid);
+
+	/* TODO: ds_rsvc_set_hint(&svc->ps_rsvc, &out->hint); */
+	pool_svc_put_leader(svc);
+out:
+	/* TODO: RPC reply rc assign to rc */
+	D_INFO(DF_UUID ": rebuild stop result " DF_RC "\n", DP_UUID(pool_uuid), DP_RC(rc));
```
liw: [Nit] D_DEBUG?

src/pool/srv_pool.c
```diff
@@ -8234,6 +8234,74 @@ ds_pool_svc_stop_handler(crt_rpc_t *rpc)
...
+ */
+void
+ds_pool_rebuild_start(uuid_t pool_uuid, struct rsvc_hint *hint)
+{
+	struct pool_svc *svc;
+	struct ds_pool  *pool;
+	daos_prop_t      prop = {0};
+	int              rc;
+
+	D_INFO(DF_UUID ": received request to resume/start rebuild\n", DP_UUID(pool_uuid));
```
liw: [Nit] D_DEBUG?

src/pool/srv_pool.c
```diff
@@ -8234,6 +8234,74 @@ ds_pool_svc_stop_handler(crt_rpc_t *rpc)
...
+
+	daos_prop_fini(&prop);
+out_pool:
+	ds_pool_put(pool);
+out_svc:
+	/* TODO: ds_rsvc_set_hint(&svc->ps_rsvc, &out->hint); */
+	pool_svc_put_leader(svc);
+out:
+	/* TODO: RPC reply rc assign to rc */
+	D_INFO(DF_UUID ": rebuild start result " DF_RC "\n", DP_UUID(pool_uuid), DP_RC(rc));
```
liw: [Nit] D_DEBUG?

src/pool/srv_pool.c
```diff
@@ -8234,6 +8234,74 @@ ds_pool_svc_stop_handler(crt_rpc_t *rpc)
...
+	daos_prop_t      prop = {0};
+	int              rc;
+
+	D_INFO(DF_UUID ": received request to resume/start rebuild\n", DP_UUID(pool_uuid));
+
+	rc = pool_svc_lookup_leader(pool_uuid, &svc, hint);
+	if (rc != 0)
+		D_GOTO(out, rc);
+
+	rc = ds_pool_lookup(pool_uuid, &pool);
```
liw: Why not use `svc->ps_pool`?

src/rebuild/rebuild_iv.c
liw, line 1: So no change has been made to this file, is that right?

liw (dismissed):

src/pool/srv_pool.c
kccain: Yes, and actually I can probably just get rid of any logging in this function. I'm looking at implementing some real RPC handlers now, and those typically have the D_DEBUG logging for receiving the RPC request and replying. Presumably will do the same with the start operation too.

src/pool/srv_pool.c
kccain: Right, will do. I think it may have been left from an earlier prototyping form of the code where I had it implemented entirely within the rebuild module (which used ds_pool_svc_lookup_leader in the preceding call).

src/rebuild/rebuild_iv.c
kccain: That is an excellent question :). clang-format (or whatever technology "fixes" the formatting) went pretty aggressive on this file I think, haha.

I had to study the diff pretty carefully again to remember - it looks like I made one innocent change to rebuild_iv_fetch(), converting a D_ERROR() to a DL_ERROR(rc, ...). Ouch.

kccain: @daltonbohning I've merged with latest master, and abandoned any changes to src/rebuild/rebuild_iv.c in this patch.

But when I git commit -s / the git commit hooks insist on modifying this file with all of this unrelated change. I forget, is there a command-line option to git commit to avoid this unwanted formatting change? OK, trying --no-verify argument to avoid hooks running on my local machine in this instance.

daltonbohning: @kccain You can skip githooks on a commit with `git commit --no-verify`. Then do `git commit --amend` to run the checks

src/rebuild/rebuild_internal.h
```diff
@@ -166,8 +166,8 @@ struct rebuild_global_pool_tracker {
 	uint32_t	rgt_refcount;
 
 	uint32_t	rgt_opc;
-	unsigned int	rgt_abort:1,
-			rgt_init_scan:1;
+	unsigned int                    rgt_abort : 1, /* abort: kill rebuild */
+	    rgt_init_scan : 1, rgt_stop_admin : 1;     /* stop: kill rebuild (on all engines) */
```
gnailzenh: comments for abort and stop_admin are quite similar...

src/rebuild/srv.c
```diff
@@ -1635,6 +1706,12 @@ retry_rebuild_task(struct rebuild_task *task, int error, daos_rebuild_opc_t *opc
 		return;
 	}
 
+	/* Do not retry if the administrator manually stopped the rebuild */
```
gnailzenh: should we do this check before 
```
       /* Reclaim and exclude task needs to retry indefinitely */
        if (task->dst_rebuild_op == RB_OP_RECLAIM ||
            task->dst_rebuild_op == RB_OP_FAIL_RECLAIM) {
                DL_INFO(error, DF_UUID" opc %u/%u retry", DP_UUID(task->dst_pool_uuid),
                        task->dst_rebuild_op, task->dst_map_ver);
                *opc = task->dst_rebuild_op;
                return;
        }       

```

src/include/daos_srv/pool.h
kccain: All of this weird indentation is forced by commit hooks presumably. Definitely not how I would structure the code.

src/rebuild/rebuild_iv.c
kccain: not applicable, I removed changes to this file, since it was almost entirely due to commit hooks behavior. My single-line change was so minimal it wasn't needed actually.

src/rebuild/rebuild_iv.c
kccain: removed changes to this file from this patch

src/rebuild/rebuild_iv.c
kccain: n/a, changes removed

src/rebuild/srv.c
kccain: I want the reaction to the fault injection to occur once the rebuild has had at least a minimal chance to start running (nsleeps==1 would correspond to 2 seconds / likely in scanning phase). I don't want to make the daos_test code have to sleep/guess when the rebuild may get started, to know when it would then perform the fault injection to know that this ULT is started.

src/rebuild/rebuild_internal.h
kccain: Yeah, I can update the comments either in this patch or next one to indicate that rgt_abort kills the rebuild on the leader engine only (I think that was what I determined in a very detailed code investigation at the beginning of this work). I also was trying to rationalize the terminology "abort" versus "stop" (for this feature) - but with a lot of the existing code mixing and matching the terms in different function names, etc.

src/rebuild/srv.c
kccain: No, I don't think it matters. The calling function rebuild_task_complete_schedule() for a failed RB_OP_RECLAIM or RB_OP_FAIL_RECLAIM will already have called ds_rebuild_schedule(). So I'm not sure if that test in this function gets exercised.

Also my intention with this feature is to not stop any reclaim (or fail reclaim) rebuilds. In the usual case, just stop rebuilds with opcode RB_OP_REBUILD.

src/pool/srv_pool.c
kccain: the ds_rebuild_admin_stop() and ds_rebuild_admin_start() functions now have integer return codes. I made that change recently, because I am laying the groundwork for implementing pool service cart RPC handlers who will need an rc value to put in the RPC reply. So there is an opportunity to return errors for certain cases.

The ds_rebuild_admin_stop() function currently in this patch returns 0 in all cases. If a running rebuild is not found, the code currently returns 0 - nothing to do, so no harm if we return "success" is the idea. Is that alright? We may decide instead that the best way for an admin to check (before and after a "stop" command) is to perform a pool query to inspect the rebuild state. Rather than worry about returning various errors to the stop/start commands. On the other hand, if we want errors returned by stop/start commands, we could consider some other error for this case (candidates may be -DER_NONEXIST, -DER_ALREADY, -DER_NOTAPPLICABLE, -DER_OP_CANCELED).

Another thing about ds_rebuild_admin_stop(), maybe it needs to check that the running rebuild is of the expected opcode (rgt_opc == RB_OP_REBUILD) before setting rgt_abort / rgt_stop_admin = 1? So that we don't accidentally try to terminate an OP_RECLAIM or OP_RECLAIM_FAIL operation. Let me look into whether this is necessary (and if a nonzero rc is appropriate when an attempt to stop reclaim/fail_reclaim is made).

src/rebuild/srv.c
```diff
@@ -857,6 +873,31 @@ enum {
...
+static uuid_t test_stop_start_puuid;
+
+/* administrator (via dmg command) asks to stop the currently-running rebuild for pool */
+int
+ds_rebuild_admin_stop(struct ds_pool *pool)
+{
+	struct rebuild_global_pool_tracker *rgt;
+
+	/* look up the running rebuild and mark it as aborted (and by the administrator) */
+	rgt = rebuild_global_pool_tracker_lookup(pool->sp_uuid, -1 /* ver */, -1 /* gen */);
```
liuxuezhao, line 886: as discussed that for RB_OP_RECLAIM/RB_OP_FAIL_RECLAIM should not stop it by default right? may add a special option to allow stopping it.
Just confirm if we are on same page, is it plan to be done at a following PR later?

src/rebuild/rebuild_internal.h
liuxuezhao: "rgt_abort kills the rebuild on the leader engine only"
just to be clear, rgt_abort will cause rebuild_leader_status_check() break/exit and then inside rebuild_task_ult() the executing go ahead and by rebuild_leader_status_notify() to notify all other engines to stop through IV. So actually it also stops rebuild on all engines.

src/rebuild/srv.c
```diff
@@ -885,13 +929,28 @@ rebuild_leader_status_check(struct ds_pool *pool, uint32_t op,
...
+		char                        sbuf[RBLD_SBUF_LEN];
+		double                      now;
+		char                       *str;
+		d_rank_list_t               excluded      = {0};
+		bool                        rebuild_abort = false;
+		int                         i;
+
+		if (!fi_admin_stop_fi)
+			fi_admin_stop_fi = DAOS_FAIL_CHECK(DAOS_REBUILD_ADMIN_STOP);
+		fi_admin_stop = (fi_admin_stop_fi && (op == RB_OP_REBUILD));
```
liuxuezhao, line 950: RB_OP_UPGRADE is also a kind of rebuild, looks fine can be stopped same as normal rebuild, it can be helpful to diagnose when upgrade get stuck.
Admin can do the upgrade again later after stopped.

src/rebuild/srv.c
```diff
@@ -885,13 +929,28 @@ rebuild_leader_status_check(struct ds_pool *pool, uint32_t op,
...
+
+		D_INFO(DF_RB ": fi_admin_stop_fi=%d, fi_admin_stop=%d, nsleeps=%u\n",
+		       DP_RB_RGT(rgt), fi_admin_stop_fi, fi_admin_stop, nsleeps);
```
liuxuezhao: suppose this log is only for testing purpose for now and will be removed later after the RPC handler be ready.
same for the "test_stop_start_puuid".

src/rebuild/srv.c
```diff
@@ -943,9 +1002,17 @@ rebuild_leader_status_check(struct ds_pool *pool, uint32_t op,
...
-		    myrank == pool->sp_iv_ns->iv_master_rank)
+		/* leader send status to target engines if still running, or if admin stopped the
+		 * rebuild */
+		if (myrank == pool->sp_iv_ns->iv_master_rank) {
+			if ((!rgt->rgt_abort && !is_rebuild_global_done(rgt)) ||
+			    (rgt->rgt_abort && rgt->rgt_stop_admin))
+				D_INFO(DF_RB ": send status to tgts, abort=%d, stop_admin=%d, "
+					     "global_done=%d\n",
+				       DP_RB_RGT(rgt), rgt->rgt_abort, rgt->rgt_stop_admin,
+				       is_rebuild_global_done(rgt));
```
liuxuezhao: looks too much INFO log printed for every loop. probably fine to only log it for the rgt_stop_admin case.

another more important point - previous code will not call rebuild_leader_status_notify() here if it is aborted or global done, instead do it inside rebuild_task_ult() afer this ULT exit. Any intention to change to always call it here?
one special case is "the leader is stopped due to the leader change" (rgt->rgt_abort && rgt->rgt_status.rs_errno == 0) cannot notify other engines to abort at here, instead should let rebuild_task_ult() check that case.
This need to be fixed, I'll leave a -1 for it.

kccain:
> Need engine code reviews still.

Still awaiting the reviews. A more practical approach here is to go ahead with the reviews we do have, land this patch (no regressions in testing). Tweaks to the core logic can be done in subsequent patches (there are multiple development tickets for this work, and more PRs are planned).

src/pool/srv_pool.c
kccain: My plan for this, most likely to be included with subsequent planned patches for this work

- if there is nothing to stop, silently succeed (no change)
- prevent stop command from actually stopping any reclaim/fail_reclaim
- look into more elaborate logic change to handle stop during fail_reclaim, "remember" this state, and defer the scheduling of the retry of the original op:Rebuild until Fail_reclaim finishes. Experimenting in PR https://github.com/daos-stack/daos/pull/16653 for now (it needs some further changes, as failures were seen in testing).

src/rebuild/srv.c
```diff
@@ -1635,6 +1706,12 @@ retry_rebuild_task(struct rebuild_task *task, int error, daos_rebuild_opc_t *opc
 		return;
 	}
 
+	/* Do not retry if the administrator manually stopped the rebuild */
+	if (rgt->rgt_stop_admin) {
```
liuxuezhao: looks like this check need to move ahead, to be before above L1681 "retry with network error", because when admin stop the rebuild, even the rebuild failed due to network err, it should not retry.

src/rebuild/srv.c
liuxuezhao: I think can support to stop rebuild with RB_OP_UPGRADE also, because the upgrade possibly fails due to some reasons and we need a method to stop it for diagnose it.

src/rebuild/srv.c
```diff
@@ -1694,78 +1774,91 @@ rebuild_task_complete_schedule(struct rebuild_task *task, struct ds_pool *pool,
...
-		D_DEBUG(DB_REBUILD, DF_UUID" opc %u/%u, state %d error %d, update status.\n",
-			DP_UUID(task->dst_pool_uuid), task->dst_rebuild_op, task->dst_map_ver,
-			rgt->rgt_status.rs_state, rgt->rgt_status.rs_errno);
+		/* During and after an admin stop, pool query should show an indication.
+		 * Query rs_state from original Rebuild = IN_PROGRESS until Fail_reclaim finishes
+		 * later.
+		 */
+		if (rgt->rgt_stop_admin && rgt->rgt_abort)
+			rgt->rgt_status.rs_errno = -DER_OP_CANCELED;
+		update_rb_status = true;
```
liuxuezhao: what's the effect of "update_rb_status" flag? seems only will affect the L1854's log? ( the return code of this func is ignored).

src/rebuild/srv.c
```diff
@@ -1886,14 +1979,20 @@ rebuild_task_ult(void *arg)
...
 			 */
-			D_DEBUG(DB_REBUILD, DF_RB " Only stop the leader\n", DP_RB_RGT(rgt));
+			D_INFO(DF_RB ": stop rebuild due to admin command.\n", DP_RB_RGT(rgt));
+		} else if (rgt->rgt_abort && rgt->rgt_status.rs_errno == 0) {
+			/* If the leader rebuild is aborted due to a leader change, then do not
+			 * abort the real rebuild(scan/pull ults), because the new leader will
+			 * resend the scan requests, which will then become the new leader to track
+			 * the rebuild.
+			 */
+			D_DEBUG(DB_REBUILD, DF_RB " abort rebuild on leader only\n",
```
liuxuezhao: the log is incorrect, it indeed did not abort the rebuild, it just stop the leader. the rebuild is keep going ahead.
so the original log looks better.

src/rebuild/srv.c
```diff
@@ -1970,20 +2067,34 @@ rebuild_ults(void *arg)
...
 
+		fi_admin_start = DAOS_FAIL_CHECK(DAOS_REBUILD_ADMIN_START);
+		if (fi_admin_start) {
+			if (!uuid_is_null(test_stop_start_puuid)) {
+				D_INFO("fi_admin_start=%d, apply to pool " DF_UUID "\n",
+				       fi_admin_start, DP_UUID(test_stop_start_puuid));
+				ds_pool_rebuild_start(test_stop_start_puuid, NULL /* hint */);
+				uuid_clear(test_stop_start_puuid);
+			}
+		}
```
liuxuezhao, line 2088: suppose this section of code will be removed after the RPC handler added later?

src/rebuild/srv.c
```diff
@@ -2044,28 +2155,25 @@ rebuild_ults(void *arg)
...
-		d_list_for_each_entry(rpt, &rebuild_gst.rg_tgt_tracker_list, rt_list) {
-			if (uuid_compare(rpt->rt_pool_uuid, pool_uuid) == 0 &&
-			    (ver == (unsigned int)(-1) || rpt->rt_rebuild_ver == ver) &&
-			    (gen == (unsigned int)(-1) || rpt->rt_rebuild_gen == gen) &&
-			    (term == (uint64_t)(-1) || rpt->rt_leader_term == term)) {
-				D_INFO(DF_RB " try abort rpt %p\n", DP_RB_RPT(rpt), rpt);
-				rpt->rt_abort = 1;
-				aborted = false;
-			}
+		rpt = rpt_lookup(pool_uuid, -1 /* opc */, ver, gen);
```
liuxuezhao: why change the original code here?
as the original func take the "term" as the parameter, looks better to keep the original code.
although it should be same if called as "ds_rebuild_abort(pool->sp_uuid, -1, -1, -1)", but if call with not -1 term, the original code is better for that case.

src/rebuild/srv.c
```diff
@@ -2894,6 +3023,8 @@ init(void)
 {
 	int rc;
 
+	uuid_clear(test_stop_start_puuid);
```
liuxuezhao, line 3026: such code will be removed after the RPC handler be ready later?

liuxuezhao (changes requested):

liw (dismissed):

src/rebuild/srv.c
kccain: is "only abort the leader" a satisfactory log message? I'd like to change the word stop -> abort. So that we do not confuse the two things.

src/rebuild/srv.c
kccain: Yes, the D_INFO is only for my debugging purposes, will remove it.

src/rebuild/srv.c
kccain: not sure if we have any automated upgrade tests. I can look at making the core logic work, and perhaps we can arrange later for some manual testing of stopping an upgrade. I don't think I want to change the fault injection code right now to try to handle upgrade.

src/rebuild/srv.c
kccain: Correct, need to prevent stop with those two opcodes. I can update the code here to do that. We may later decide to implement a "force" option to stop a reclaim, that would be in a future PR if we decide to do it.

Also in a later patch, we will have to solve another challenge that is what to do if a stop request arrives during Fail_reclaim. For this, we will need to "remember" that the request arrived, but wait until Fail_reclaim completes and then delay the rescheduling of the retry of the original rebuild until that time. Right now, the code schedules the Fail_reclaim and the retry of the original rebuild when it is processing the error of the original rebuild.

src/rebuild/srv.c
kccain: Let's get rid of the D_INFO here, even for the rgt_stop_admin case.

The if (myrank == pool->sp_iv_ns->iv_master_rank) test above, does that confirm that the engine running this code branch is the leader, in which case it should be ok to invoke rebuild_leader_status_notify()?

But maybe the existing rebuild_leader_status_nofity() call in rebuild_task_ult() will be sufficient and I don't have to worry about the rgt_stop_admin case here? Unsure, I will have to try it and see.

src/rebuild/srv.c
kccain: Let me look at the ordering of events more closely (like when this patch sets the errno to -DER_OP_CANCELED). When admin stops the rebuild, the error code will be specifically set to -DER_OP_CANCELED which is not retryable.

src/rebuild/srv.c
kccain: Once we have the control plane command (dmg pool rebuild stop) and the PS RPC handler, we can consider changing the daos_test variants to use the dmg helpers library to invoke the dmg commands, instead of performing these fault injections. If it can work cleanly in the tests, we could consider removing fault injection. The timing of the "stop/start" commands might be tricky. The fault injection approach might be more precise to ensure that the stop/start occurs when we want it to.

src/rebuild/srv.c
kccain: Ah, I will take another look. Originally I modified rpt_lookup() to take a term argument, but you and I had an offline chat about this and you did not prefer that. So maybe I have to just restore the original form of the code here.

src/rebuild/srv.c
kccain: hopefully, once we have the control plane commands implemented, some dmg helpers for daos_test to call, etc. Depends on how easy it will be to structure the daos_test code to time the stop/start commands precisely.

kccain (commented):
Looking into code changes now, here are some initial replies to @liuxuezhao feedback. Let me know if any concerns with my ideas.

src/rebuild/srv.c
liuxuezhao: not sure if the pool->sp_iv_ns->iv_master_rank can be synced to this node when PS leader switch and rebuild leader stopping (rebuild_leader_stop).
the way of the original code looks safer anyway.

src/rebuild/srv.c
liuxuezhao: even if the code logic can ensure always set errno to DER_OP_CANCELED for rgt_stop_admin case, moving the check ahead looks easier to understand the logic.

src/rebuild/srv.c
liuxuezhao: the case here only stop the rebuild leader ULT as it was taken over by other leader, but the real rebuild jod did not stop/abort.
BTW, please ensure that for admin stop case, the rgt->rgt_status.rs_errno is set to non zero. thx

src/rebuild/rebuild_internal.h
kccain: I'll update the comment and avoid specifying "on all engines" language.
 
 rgt->rgt_abort can result in no rebuild_leader_status_notify() ncall, when (rgt->rgt_abort && rgt->rgt_status.rs_errno == 0) (rebuild aborted due to leader switch).

daltonbohning (commented):
ftest LGTM

src/rebuild/srv.c
kccain: will restore the original code.

src/rebuild/srv.c
kccain: will do

src/rebuild/srv.c
kccain: removing this boolean and replicating the DL_CLDEBUG() to be immediately after the status update calls.

src/rebuild/srv.c
kccain: will change the log text and will make sure rgt->rgt_status.rs_errno is set to -DER_OP_CANCELED for that case in this function.

kccain (commented):
pushed a candidate final change, for CI testing and hopefully final review.

src/rebuild/srv.c
```diff
@@ -857,6 +873,39 @@ enum {
...
+		       DP_UUID(pool->sp_uuid));
+		return 0;
+	}
+
+	/* admin stop command does not terminate reclaim/fail_reclaim jobs */
+	if ((rgt->rgt_opc == RB_OP_REBUILD) || (rgt->rgt_opc == RB_OP_UPGRADE)) {
+		D_INFO(DF_RB ": stopping rebuild opc %u(%s)\n", DP_RB_RGT(rgt), rgt->rgt_opc,
+		       RB_OP_STR(rgt->rgt_opc));
+		rgt->rgt_abort      = 1;
+		rgt->rgt_stop_admin = 1;
```
liuxuezhao, line 899: how about set "rgt->rgt_status.rs_errno = -DER_OP_CANCELED" here at the first place
the current PR's setting looks can work although.

liuxuezhao (approved):
thanks for addressing the questions.

src/rebuild/srv.c
kccain: I will look into it in a subsequent patch. Thanks for the review!

gnailzenh (approved):

PR #16568, opened 2025-07-04
DAOS-17656 object: several fix for EC object fetch processing

liuxuezhao:
1) For EC singv degraded fetch, don't directly start EC data recovery
   from client side, because cannot directly use any hlc from client.
   Instead it will do a degraded fetch from parity shard to get back
   the valid epoch for EC data recovery first.
2) For EC singv degraded fetch server handling, need not bulk
   transferring back the data, instead only need to do the bulk for
   data recovery.
3) For EC array size query, for the degraded fetch also need to convert
   the recx vos index to replica offset.

test code change - add degraded test for EC 18, and reset fail injection when init each test case.

### Steps for the author:

### DIFF
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -58,7 +58,7 @@ open_retry:
 		D_GOTO(unlock, rc = -DER_NONEXIST);
 	}
 
-	D_DEBUG(DB_TRACE, "Open object shard %d\n", shard);
+	D_DEBUG(DB_TRACE, "Open object shard %d, lock_upgraded %d\n", shard, lock_upgraded);
 
 	if (obj_shard->do_obj == NULL) {
 		daos_unit_oid_t	 oid;
@@ -1089,7 +1089,7 @@ obj_shard_tgts_query(struct dc_object *obj, uint32_t map_ver, uint32_t shard,
 			  grp_idx * daos_oclass_grp_size(obj_get_oca(obj_auxi->obj));
 
 		if (isclr(bitmap, tgt_idx)) {
-			D_DEBUG(DB_TRACE, DF_OID" shard %u is not in bitmap\n",
+			D_DEBUG(DB_IO, DF_OID " shard %u is not in bitmap\n",
 				DP_OID(obj->cob_md.omd_id), obj_shard->do_id.id_shard);
 			D_GOTO(close, rc = -DER_NONEXIST);
 		}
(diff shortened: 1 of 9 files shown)


src/object/cli_obj.c
```diff
@@ -1962,8 +1962,8 @@ obj_ec_recov_cb(tse_task_t *task, struct dc_object *obj,
 		 * singv recovery without fetch from server ahead - when
 		 * some targets un-available.
 		 */
```
liuxuezhao, line 1964: above comment should remove, I'll remove it at a following PR or when refresh this PR.

wangshilong (approved):
Looks reasonable for me, someone more familiar with client codes might take a look.

src/object/srv_obj.c
```diff
@@ -597,9 +600,14 @@ obj_bulk_transfer(crt_rpc_t *rpc, crt_bulk_op_t bulk_op, bool bulk_bind, crt_bul
...
-		rc = bulk_transfer_sgl(ioh, rpc, remote_bulks[i + skip_nr],
-				       remote_offs ? remote_offs[i] : 0,
-				       bulk_op, bulk_bind, sgl, i, p_arg);
+		/* OBJ_BULK_OFFSET_SKIP is for the case of EC single value degraded fetch,
+		 * unnecessary to bulk transfer here, the data will be transferred back when
+		 * data recovery (ORF_EC_RECOV).
+		 */
+		if (remote_offs == NULL || remote_offs[i] != OBJ_BULK_OFFSET_SKIP)
+			rc = bulk_transfer_sgl(ioh, rpc, remote_bulks[i + skip_nr],
+					       remote_offs ? remote_offs[i] : 0, bulk_op, bulk_bind,
```
NiuYawei, line 609: For old code, -DER_INVAL is returned when "remote_offs == NULL", I think we'd keep that, right?

NiuYawei (changes requested):
marked as "-1" before the concern is addressed.

src/object/srv_obj.c
liuxuezhao: the old code, when remote_offs == NULL it just use offset 0, that is the case for replica obj.
the change here is only for the case of long EC single value that must with remote_off != NULL and could set as OBJ_BULK_OFFSET_SKIP.
Would you please check again? thanks

liw (approved):

src/object/srv_obj.c
NiuYawei: yeah, I read it wrong. sorry for the noise.

NiuYawei (approved):

PR #16589, opened 2025-07-10
SRE-3194 build: Remove redhat-lsb-core from Dockerfile.mockbuild

ryon-jensen:
### Steps for the author:

### DIFF
--- a/utils/rpms/packaging/Dockerfile.mockbuild
+++ b/utils/rpms/packaging/Dockerfile.mockbuild
@@ -30,7 +30,7 @@ RUN chmod +x /tmp/repo-helper.sh &&                 \
 # This needs to be moved to a shell script like above in the future to
 # properly only remove the proxy variables only when they need to be removed
 RUN dnf -y install mock make                                        \
-                   rpm-build createrepo rpmlint redhat-lsb-core git \
+                   rpm-build createrepo rpmlint git                 \
                    python-srpm-macros rpmdevtools &&                \
     dnf -y clean all
 


utils/rpms/packaging/Dockerfile.mockbuild
```diff
@@ -30,7 +30,7 @@ RUN chmod +x /tmp/repo-helper.sh &&                 \
 # This needs to be moved to a shell script like above in the future to
 # properly only remove the proxy variables only when they need to be removed
 RUN dnf -y install mock make                                        \
-                   rpm-build createrepo rpmlint redhat-lsb-core git \
+                   rpm-build createrepo rpmlint git                 \
```
kccain, line 33: in general what is the relationship between the files in the daos repo here, and the files of the same name found in the packaging repo https://github.com/daos-stack/packaging ? Are they intended to be identical under typical circumstances?

kccain (approved):

utils/rpms/packaging/Dockerfile.mockbuild
JohnMalmberg: Yes, and this change was already supposed to have been landed in the packaging repo and the active packaging jobs because we saw this failure multiple packaging jobs.

kccain: I don't think this change has landed in packaging repo. And there are several differences between the packaging repo file contents and the ones in the daos repo - so I was just asking to figure out if this PR is the right way forward, or if some broader alignment/sync between packaging and daos repos is needed.

JohnMalmberg: Normally changes needed to the packaging "reference" repo are usually found when making updates to individual packaging jobs.

Once those PRs look stable then the packaging repo should get updated and then changes to it should get landed before landing the updated packaging jobs so that the PRs run with all stages green.

For the packaging PRs to pass, the master and release branches need to be stable, so we sometimes need to land fixes to master and the release branches first.

This PR is because we are having random PRs failing because the redhat-lsb-core package no longer exists in Fedora/42 which is being used to in the container for RPM packaging to make sure that the DAOS master and release branches are stable.

I have not been tracking the status of the PRs for the packaging repos and the packaging "reference" repo.

PR #16457, opened 2025-05-30
DAOS-17472 build: Enable new RPM builds

jolivier23:
Build RPMs using fpm based on utils/build.config for DAOS and its dependencies
Modify Jenkinsfile to use these new RPMs for testing.

### Steps for the author:

* [ ] Commit message follows the [guidelines](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Appropriate [Features or Test-tag](https://daosio.atlassian.net/wiki/spaces/DC/pages/10984259629/Test+Tags) pragmas were used.
* [ ] Appropriate [Functional Test Stages](https://daosio.atlassian.net/wiki/spaces/DC/pages/12147556353/CI+Functional+Test+Stages) were run.
* [ ] At least two positive code reviews including at least one code owner from each category referenced in the PR.
* [ ] Testing is complete. If necessary, forced-landing label added and a reason added in a comment.

#### After all prior steps are complete:
* [ ] Gatekeeper requested (daos-gatekeeper added as a reviewer).

PR #16451, opened 2025-05-29
DAOS-6344 build: move daemon logfiles from /tmp to /var/log/daos

ravalsam:
- daos_server and daos_client rpm install will create the /var/log/daos dir.
- rpm uninstall will not remove /var/log/daos dir
- It will also change the permission of /var/log/daos incase user is trying to access the same node for both client/server.

### Steps for the author:

* [ ] Commit message follows the [guidelines](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Appropriate [Features or Test-tag](https://daosio.atlassian.net/wiki/spaces/DC/pages/10984259629/Test+Tags) pragmas were used.
* [ ] Appropriate [Functional Test Stages](https://daosio.atlassian.net/wiki/spaces/DC/pages/12147556353/CI+Functional+Test+Stages) were run.
* [ ] At least two positive code reviews including at least one code owner from each category referenced in the PR.
* [ ] Testing is complete. If necessary, forced-landing label added and a reason added in a comment.

#### After all prior steps are complete:
* [ ] Gatekeeper requested (daos-gatekeeper added as a reviewer).

PR #16650, opened 2025-07-23
DAOS-17472 build: Enable new RPM builds

jolivier23:
Restores PR #16457 and fix the issues with it.

* isa-l build.config was out of sync with RPM builds
* Build appropriate -devel packages for backward compatibility
* Update isa-l to 2.31 (which is default version on el9 and gets used anyway there)
* Should also reduce docker cache size a bit

Co-authored-by: Kris Jacque <kris.jacque@hpe.com>

Skip-func-test-el9: false
Skip-func-test-leap15: false

### Steps for the author:


ryon-jensen:
Please hold off on landing this until we’re confident about its impact on CI. I’ll be out for a couple of weeks and won’t be able to review it myself. In the meantime, @JohnMalmberg and @grom72 can provide input, but they’ll be following Brad’s guidance on the level of risk we’re willing to take with this change.

kjacque (commented):
Go changeset looks unchanged from its previously-approved incarnation, so LGTM. Mind re-adding me as a co-author/signoff for the Go changes I contributed, please? Thanks.

jolivier23:
> Go changeset looks unchanged from its previously-approved incarnation, so LGTM. Mind re-adding me as a co-author/signoff for the Go changes I contributed, please? Thanks.

I just added it.  I had that thought too but figured I'd add it when if/when this lands.   Either way, it will be up to the Gatekeeper to make sure it persists.

kjacque (dismissed):
I only verified the Go stuff. LGTM

daltonbohning (changes requested):
-1 until we understand the impact to CI / packaging. Currently master and release/2.6 share RPM repos. So landing this means 2.6 will effectively have all of its dependencies updated for the next release.

jolivier23:
> -1 until we understand the impact to CI / packaging. Currently master and release/2.6 share RPM repos. So landing this means 2.6 will effectively have all of its dependencies updated for the next release.

Except for fuse and spdk which are different in master.

And most others are not an upgrade because nothing has changed.

daltonbohning:
> > -1 until we understand the impact to CI / packaging. Currently master and release/2.6 share RPM repos. So landing this means 2.6 will effectively have all of its dependencies updated for the next release.
> 
> Except for fuse and spdk which are different in master.

As far as I can see, fuse and spdk are not pinned to older versions in 2.6. So uploading new RPMs means 2.6 will use them.

> And most others are not an upgrade because nothing has changed.

If newer RPMs are uploaded to artifactory, 2.6 will use them. Unless there is logic in place to not upload when nothing has changed.

jolivier23:
> > > -1 until we understand the impact to CI / packaging. Currently master and release/2.6 share RPM repos. So landing this means 2.6 will effectively have all of its dependencies updated for the next release.
> > 
> > 
> > Except for fuse and spdk which are different in master.
> 
> As far as I can see, fuse and spdk are not pinned to older versions in 2.6. So uploading new RPMs means 2.6 will use them.
> 
> > And most others are not an upgrade because nothing has changed.
> 
> If newer RPMs are uploaded to artifactory, 2.6 will use them. Unless there is logic in place to not upload when nothing has changed.

We have this problem already, it's just less frequent.   If someone changes libfabric, it will cause release/2.6 to use new RPMs as well.  I don't see a big difference here other than updates happen more frequently.

The right answer if we want to avoid this is to branch the repo IMO.  It's crazy that we have persisted so long with a model that allows someone to update release branch components without merge approval.

daltonbohning:
> We have this problem already, it's just less frequent. If someone changes libfabric, it will cause release/2.6 to use new RPMs as well. I don't see a big difference here other than updates happen more frequently.

We don't currently rebuild and upload new RPMs for every dependency on landing a PR. This is the behavior change I'm referring to.

jolivier23:
> > We have this problem already, it's just less frequent. If someone changes libfabric, it will cause release/2.6 to use new RPMs as well. I don't see a big difference here other than updates happen more frequently.
> 
> We don't currently rebuild and upload new RPMs for every dependency on landing a PR. This is the behavior change I'm referring to.

yeah, I know.   But the point is release/2.6 is already affected by changes to master components.

jolivier23:
I just uploaded a change that will use the last change to deps or utils/build.config to identify the deps we build.

Can someone tell me what needs to happen to ensure we only push new RPMs if RPMs of the same name don't already exist?

tanabarr:
Sorry this may already be answered but why can we not pin all dependencies in the 2.6 spec and I think we are already doing it in build.config for the sauce builds. is there some restriction to the number of versions we can store in artifactory and this prevents us from pinning?

jolivier23:
> Sorry this may already be answered but why can we not pin all dependencies in the 2.6 spec and I think we are already doing it in build.config for the sauce builds. is there some restriction to the number of versions we can store in artifactory and this prevents us from pinning?

We can potentially pin the versions in 2.6 but there is some risk in doing so, particularly for client side.  I still think a separate repo will be the best path forward, particularly if this lands before 2.8 branch is cut.

tanabarr:
> > Sorry this may already be answered but why can we not pin all dependencies in the 2.6 spec and I think we are already doing it in build.config for the sauce builds. is there some restriction to the number of versions we can store in artifactory and this prevents us from pinning?
> 
> We can potentially pin the versions in 2.6 but there is some risk in doing so, particularly for client side. I still think a separate repo will be the best path forward, particularly if this lands before 2.8 branch is cut.

do we have the necessary resources on the team to do the repo branching before 2.8? are there risks involved or is the task well scoped? imagine this is a conversation with @JohnMalmberg and @ryon-jensen

daltonbohning:
FYI #16451 is landed so the log path would need to be updated here.

jolivier23:
> FYI #16451 is landed so the log path would need to be updated here.

done

daosbuild3:
Test stage Unit Test on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16650/6/display/redirect

daosbuild3:
Test stage Unit Test on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16650/7/display/redirect

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16650/8/execution/node/1558/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16650/8/execution/node/1569/log

kjacque (approved):
Go changes LGTM.

johannlombardi (approved):

johannlombardi:
@daltonbohning @phender are we good to land this patch?

daltonbohning:
> @daltonbohning @phender are we good to land this patch?

Isn't this still uploading all dependency RPMs for landing runs?

ryon-jensen:
> > @daltonbohning @phender are we good to land this patch?
> 
> Isn't this still uploading all dependency RPMs for landing runs?

This is my concern as well ... we don't fully understand how this will impact CI.

Jenkinsfile
```diff
@@ -22,6 +23,31 @@
...
+                              cd "$GIT_CHECKOUT_DIR"
+                          fi
+                          if git diff-index --name-only HEAD^ | grep -q TAG; then
+                              echo ""
+                          else
+                              echo ".$(git rev-list HEAD --count).g$(git rev-parse --short=8 HEAD)"
+                          fi''',
+                returnStdout: true).trim()
+    // If head is a TAG, keep it the same but otherwise, use the commit where deps changed
+    env.DAOS_DEPS_RELVAL = sh(label: 'get deps git tag',
```
jolivier23: @daltonbohning @ryon-jensen @johannlombardi we could potentially change it so it only uploads the new packages when DAOS_DEPS_RELVAL == DAOS_RELVAL.  That would indicate a change.   It would still upload everything when that happens but that would be a smaller diff than every landing.

We could also key on this as to whether or not we build them or install what we can (though fused and daos-spdk) don't have devel packages at present because I don't even package fused (static) and daos-spdk is internal only so I didn't bother.  One could certainly just create one with some effort.

I don't really have many cycles these days though.

jolivier23: Now that I think about it, I assume uploading would fail because rpms would always be named the same.  Maybe if we just handle that to start, it would suffice?

JohnMalmberg: Both Nexus and Artifactory expect that once an RPM/DEBIAN package with a name is uploaded, it will not be replaced with a different package of the same exact name with a different checksum or length.   There is an index file that needs to be maintained and updating that index file runs as a background task, and sometimes misses updating for a change, and then that requires manual intervention.

RPM packages with frequent version changes require special handling, because the time it takes for the RPM solver to determine which package to install is based on number of versions in the repository.   As the dependent packages normally have not been updating very frequently, they are in their own repositories for each distro and major version.

The dependencies are also in their own repository so that we can make the dependencies available for CI installs while blocking the CI installs from installing older versions of DAOS.   This also makes the install faster as it reduces the number of package versions that the installer needs to select when we only want to install the currently built packages.

The DAOS packages go into separate repositories that has automation that is only keeping packages in it that have been uploaded and downloaded in the past 30 days.   If we don't do this, it eventually takes too long to install DAOS.
The older packages are moved to an archive repository.

The size of the dependent package repositories may now be reaching a point where we need to start archiving the older packages to keep the RPM install time down.

So changes like this will require some infrastructure planning to determine what works best with the tools that we have available or are able to get.

jolivier23: So what if we used that "other repo" to install the version we need and uploaded to it on landing builds that build a new version.  We could push DAOS bits to one repo and other stuff (which won't change much to the other).  This doesn't seem like an insurmountable problem

jolivier23:
I've added a proof of concept for EL8....if it works, I'll make the same changes in the other builds.

johannlombardi (approved):

Jenkinsfile
jolivier23: I've taken a stab at this.  It will still upload dependencies to the daos repo (I need some help in Jenkins to avoid that) but it will only build/upload them if dependencies with the same stamp don't exist.  Otherwise, it will install the existing ones and avoid building them.

JohnMalmberg: The current system that builds the dependent RPMs separately tests those built RPMs with master and all active release builds, and at least one of the weekly or daily tests for those branches.

So we know that those RPMs are compatible both with master and all active release branches before they are landed.   So there should never be a problem with updating the version or patch level of a dependency and it breaking a release branch that is in progress.

JohnMalmberg: There are a lot of changes with this PR.   I do not see where the chroot environments are setup that make sure that the RPM build environments are setup correctly and consistently for RPMs that are intended to be distributed.

Is there some document that describes how this new workflow is suppose to work?

The 'product' parameter for uploading a package for pipeline-lib:/publishToRepository determines is passed to a site specific artifact uploader.    For the local site there are site specific rules which puts "daos*" project names in the daos-stack-daos.... repositories and everything else in the daos-stack-deps... repositories.

The publish_branch parameter is used to control the when artifacts get published or not, and default is only publish artifacts on master, "testbuild/..." and "release/..." branches.

jolivier23: > The current system that builds the dependent RPMs separately tests those built RPMs with master and all active release builds, and at least one of the weekly or daily tests for those branches.
> 
> So we know that those RPMs are compatible both with master and all active release branches before they are landed. So there should never be a problem with updating the version or patch level of a dependency and it breaking a release branch that is in progress.

The current system is broken, IMO.  We shouldn't be trying to update most of our dependencies against any release branch without explicit merge approval.  This method ensures (aside from 2.6) that we use the rpm version we have configured to use in DAOS and not just the latest version of the same.   Also, the current method does not really guarantee much of anything.  Because it only builds and runs a single test, it doesn't prevent an issue with a new version of an artifact.  This method runs the "new" artifact against everything any other DAOS change would run against.

jolivier23: > There are a lot of changes with this PR. I do not see where the chroot environments are setup that make sure that the RPM build environments are setup correctly and consistently for RPMs that are intended to be distributed.
> 
> Is there some document that describes how this new workflow is suppose to work?
> 
> The 'product' parameter for uploading a package for pipeline-lib:/publishToRepository determines is passed to a site specific artifact uploader. For the local site there are site specific rules which puts "daos*" project names in the daos-stack-daos.... repositories and everything else in the daos-stack-deps... repositories.
> 
> The publish_branch parameter is used to control the when artifacts get published or not, and default is only publish artifacts on master, "testbuild/..." and "release/..." branches.

This isn't using a traditional rpmbuild method so I'm not sure we can use chroot here.   What it is doing is unifying the "developer build" and the "rpm build" into a single build.  The rpms are produced after the build is finished based on the available bits that are built and moves files around as needed.  For example, when we build pmdk, we build it in $PREFIX/prereq/release/pmdk.  The rpmbuild for pmdk is done by utils/rpms/pmdk.sh.  This file will take the built bits from prereq/release/pmdk and produces rpms that mimic the old behavior (e.g. they will install headers in /usr/include, for example).   There is a fair bit of stuff done to make this transparent such as stripping rpath from the binaries put into the RPM.

We should investigate removing isa-l and isa-l_crypto from our builds on platforms where they are supplied by the distro.   None of our other dependencies are suitable for this option.

The one exception is the source RPM which would still use daos.spec but include all dependency code.  I am not presently producing this.

So it sounds like the dependencies will automatically go to the right place?

jolivier23: I don't have a document but I could probably produce one.

The basic flow is

1. User does build
2. User optionally produces rpms from build (utils/rpms/build_packages.sh)

There are a few environment variables that control how this script works including `DISTRO{el8, el9, suse.lp155, ubuntu}`, `DAOS_RELVAL={extention representing last change|empty on tagged commit}`, `DAOS_DEPS_RELVAL={extention representing last change to deps | empty on tagged commit}`, `OUTPUT_TYPE={deb|rpm}`

jolivier23: By default, the docker scripts will produce these RPMs (so anyone can produce some).   But in Jenkins, they are produced by the "Build on" steps

jolivier23: With the latest changes, we will only produce deps build if rpms with the same DAOS_DEPS_RELVAL string in the version don't exist in the repository.  So, for example, once this lands, the next PR after a landing build completes should not produce new RPMs

jolivier23: though all of the latest changes have not yet been tested in CI

JohnMalmberg: It is my understanding that the current system does not just run a single test and has not for quite a while when a dependency is updated. 

That change was made a while back to make sure that updating a dependency because of a need in either the master branch or an active release branch would not break something in the other branch.

I just checked the Jenkinsfile for one of the dependency builds:

It does run the master and release/2.6 and master branches, and not daily/weekly tests, and I see it setting some test tag parameters.  It runs all the functional tests, but it does skip the large test, and looks like it runs at least one medium test.

We test the prebuilt CI images with a more comprehensive test that includes running more tests, so we won't get any surprises when they are deployed.

If you are using a upload for each set of package RPMs as they are built, and setting product parameter, for them, then they will go in the right place.   If you are uploading the RPMs in one operation, and letting product default, then they will into the DAOS repository where we probably will not want them.

And the more RPMs you upload in one batch to a repository, the longer it may take for verifying the upload mainly because the index file is generated as a delayed background task that we have to wait for completion.   That will not show up in PRs, just landing builds.

I don't see an issue with building RPMs that are only used in a PR outside of a chroot if that makes it easier for developers, but it may be prudent to continue building the RPMs for landing in a chroot.

jolivier23: https://github.com/daos-stack/mercury/blob/e501d61efabf8278cf09013fe9c607613227ddb4/Jenkinsfile#L10

daosio runs a subset of the functional tests.  I think this is the most common tag we use.

>I don't see an issue with building RPMs that are only used in a PR outside of a chroot if that makes it easier for developers, but it may be prudent to continue building the RPMs for landing in a chroot.

This change is deprecrating the daos-stack repos that contain the spec files to do this and ending maintenance of the same.   One of the main problems I'm trying to solve with this change, something that we hit many times at Google, is that there are too many sources of truth and they don't always align.  This change essentially moves all of the truth into the DAOS repository.  For example, the release/2.6 utils/build.config never got updated after the move from Argobots 1.1 to 1.2 but the rpm builds grabbed it automatically.  Since Google builds the server using developer build, this means that upstream was building and testing against a different version of argobots in its releases than we were until we noticed the issue.

No more duplication of patches between those daos-stack repos and utils/build.config used by developers.  With this change, the latter becomes the source of truth and developers are guaranteed to be building the same thing that is being tested with RPMs.

This is also making it simpler to maintain these RPMs, at least for server only ones like spdk.   We build a daos specific rpm for this component and don't intend to maintain any compatibility with an upstream spdk (which nobody will ever produce as it's not really the way spdk is intended to be used).

jolivier23: Anyway, the idea is that if you have a commit from DAOS, you can reproduce exactly the set of dependencies used in the official release from the same commit even if you don't have those specific RPMs.

jolivier23: Another benefit of this method is for Ubuntu packages.   The debian packages are built from the same main source so we no longer need to update debian folder.  Instead, we just tell the script to produce deb files instead.  It also fixes some issues with the debian packages such as the installation directory of libraries which was /usr/lib64 previously.

JohnMalmberg: Right now I need nothing to break Jenkins CI at all.

I am concerned about the dependencies getting uploaded to the DAOS repository instead of the dependency repo.

Anything that increases how much disk space docker or the Jenkins workspace uses is possibly going to be a problem.

I also want to see how this PR works on the staging Jenkins with on one of the docker build agents that I am trying to bring up.

jolivier23: I don't believe this will break CI.  If there is a way in staging to test out what happens after it lands, that would be good.  

If we need to upload to separate repos, we can probably figure that out.

This should not increase disk space usage by a significant amount as changes to dependencies are relatively rare.

tanabarr (approved):
LGTM. This is a big step forward!

JohnMalmberg (changes requested):
For this to work as expected, the landing builds must put the dependent RPMs that are built have to uploaded to the dependency repo and not the DAOS repo.

Only the dependency repo should being looked at during the DAOS RPM build process, and if the updated dependency repos are not updated, this will always be rebuilding the dependencies.

We also have RPMs that are built that are dependent on pre-built DAOS RPMs that are used for testing.  How are they affected by this change?

jolivier23:
> For this to work as expected, the landing builds must put the dependent RPMs that are built have to uploaded to the dependency repo and not the DAOS repo.
> 
> Only the dependency repo should being looked at during the DAOS RPM build process, and if the updated dependency repos are not updated, this will always be rebuilding the dependencies.
> 
> We also have RPMs that are built that are dependent on pre-built DAOS RPMs that are used for testing. How are they affected by this change?

The code to upload to a repo for a landing build is entirely buried in pipeline-lib somewhere.   It would need to change to upload daos*.rpm to daos repo and everything else to dependency repos.  Is there a dependency repo for all dependencies or one for each dependency?   If you or someone can help me figure out what needs to change, I'd appreciate it.

I don't think this patch changes anything for RPMs that depend on DAOS rpms, does it?

jolivier23:
if I'm reading the code correctly, it looks like one would need to invoke publishToRepository twice, once for deps and once for DAOS and would need the rpms to be in separate directories.

JohnMalmberg:
Definitely needs the Publish to repository done differently for the dependencies and DAOS, would not be surprised if it needs separate directories.

jolivier23:
@JohnMalmberg Any way to really test this proposed change

https://github.com/daos-stack/daos/pull/16763 is a PR stacked on this one and this is a change to pipeline-lib: https://github.com/daos-stack/pipeline-lib/pull/475

The code won't really be exercised until a landing though though it is at least a proof of concept for the whether multiple artifact directories will work

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16650/19/execution/node/1458/log

Jenkinsfile
```diff
@@ -17,11 +18,36 @@
 
 // To use a test branch (i.e. PR) until it lands to master
 // I.e. for testing library changes
-//@Library(value='pipeline-lib@your_branch') _
+@Library(value='pipeline-lib@jvolivie/upload_fix') _
```
jolivier23, line 21: Before landing, this needs to be resolved.

jolivier23:
I've changed this to remove DAOS_DEPS_RELVAL.  It occurs to me that we can just bump the release version if we need a new version of a dependency to be built.  This should almost mimic what we have now.

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16650/20/testReport/

PR #16659, opened 2025-07-25
DAOS-17299 control: reserve rolling upgrade key prefix

wangshilong:
To leverage existing MS-DB code infrastructure, we propose storing rolling upgrade data in system attributes using a reserved key prefix (e.g., "ROLLING_UPGRADE").

We should protect certain system-reserved keywords (such as "rolling_upgrade.state") from being modified through standard dmg system set-attr/rm-attr commands. These protected attributes would be managed exclusively by the dmg system's rolling-upgrade subcommand.

### Steps for the author:

### DIFF
--- a/src/control/lib/control/system.go
+++ b/src/control/lib/control/system.go
@@ -42,6 +42,10 @@ const (
 	SystemJoinRetryTimeout = 10 * time.Second
 )
 
+const (
+	RollingUpgradePrefix = "ROLLING_UPGRADE"
+)
+
 var (
 	errMSConnectionFailure = errors.Errorf("unable to contact the %s", build.ManagementServiceName)
 )
@@ -1072,6 +1076,10 @@ type SystemSetAttrReq struct {
 	Attributes map[string]string
 }
 
+func IsSystemReservedKey(key string) bool {
+	return strings.HasPrefix(strings.ToUpper(key), RollingUpgradePrefix)
+}
+
 // SystemSetAttr sets system attributes.
 func SystemSetAttr(ctx context.Context, rpcClient UnaryInvoker, req *SystemSetAttrReq) error {
 	if req == nil {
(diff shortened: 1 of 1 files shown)


daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16659/1/testReport/

src/control/lib/control/system.go
```diff
@@ -1080,6 +1088,11 @@ func SystemSetAttr(ctx context.Context, rpcClient UnaryInvoker, req *SystemSetAt
 	if len(req.Attributes) == 0 {
 		return errors.New("attributes cannot be empty")
 	}
+	for key := range req.Attributes {
```
mjmac, line 1091: It would probably be better to do this in https://github.com/daos-stack/daos/blob/master/src/control/system/properties.go

There is already some framework there for distinguishing between properties that are settable/visible via `dmg system (set|get)-prop` and management properties. This seems like a good candidate to be a management property.

Is the goal to use this new property to control some behavior?

mjmac:
Adding on to my other comment... This pretty clearly seem to be a property vs. an attribute. Following the conventions used for pools and containers, attributes are generally user-defined and free-form vs. properties which have specific definitions and constraints.

kjacque (changes requested):
Agreed with @mjmac's comment. "mgmt" is the prefix used by general management properties, and it does seem to me that the MS controls this feature. You could do a naming convention like "mgmt.upgrade.<propname>" but we do have the case of non-settable props with the "mgmt" prefix covered.

src/control/lib/control/system.go
wangshilong: @mjmac thanks for suggestion, let's use prop for management property.

PR #16662, opened 2025-07-25
DAOS-17802 pool: Check RF for dc_pool_exclude

liw:
The dc_pool_exclude code path, used by NVMe fault handling, skips the RF check, potentially raising safety concerns. This patch drops flag POOL_TGT_UPDATE_SKIP_RF_CHECK in that code path, so that the RF check will apply.

The error messages logged when the RF check fails focus on ranks. To make them slightly more accurate, this patch updates the wording, and offers the list of DOWN targets when corresponding ranks are UPIN, in the draft pool map built by this POOL_EXCLUDE request. But we are unable to log the list of _all_ unavailable ranks/targets at the moment; the administrator needs to use `dmg pool query` and `dmg storage query list-devices`.

Since NVMe fault handling retries dc_pool_exclude upon -DER_RF, this patch tunes down the engine stdout/stderr output caused by those errors.

### Steps for the author:

### DIFF
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -7018,44 +7018,31 @@ pool_svc_schedule_reconf(struct pool_svc *svc, struct pool_map *map, uint32_t ma
 	return 0;
 }
 
-static int
-pool_map_crit_prompt(struct pool_svc *svc, struct pool_map *map)
+static void
+log_unavailable_targets(struct pool_svc *svc, struct pool_map *map)
 {
-	crt_group_t		*primary_grp;
 	struct pool_domain	*doms;
 	int			 doms_cnt;
-	int			 i;
-	int			 rc = 0;
+	int                      i;
 
-	D_DEBUG(DB_MD, DF_UUID": checking node status\n", DP_UUID(svc->ps_uuid));
 	doms_cnt = pool_map_find_ranks(map, PO_COMP_ID_ALL, &doms);
 	D_ASSERT(doms_cnt >= 0);
-	primary_grp = crt_group_lookup(NULL);
-	D_ASSERT(primary_grp != NULL);
 
-	D_CRIT("!!! Please try to recover these engines in top priority -\n");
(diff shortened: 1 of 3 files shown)


wangshilong (approved):

src/pool/srv_pool.c
```diff
@@ -7018,44 +7018,31 @@ pool_svc_schedule_reconf(struct pool_svc *svc, struct pool_map *map, uint32_t ma
...
-
-		rc = crt_rank_state_get(primary_grp, doms[i].do_comp.co_rank, &state);
-		if (rc != 0 && rc != -DER_NONEXIST) {
-			D_ERROR("failed to get status of rank %u: %d\n",
-				doms[i].do_comp.co_rank, rc);
-			break;
+		if (doms[i].do_comp.co_status & PO_COMP_ST_DOWN) {
+			D_ERROR(DF_UUID ":  rank %u\n", DP_UUID(svc->ps_uuid),
+				doms[i].do_comp.co_rank);
+		} else if (doms[i].do_comp.co_status & PO_COMP_ST_UPIN) { // XXX: ask Xuezhao
```
tanabarr, line 7036: resolve this comment before landing?

tanabarr (approved):

src/pool/srv_pool.c
liw: Good catch. I did forget about this. Will remove after asking Xuezhao. :)

liuxuezhao: may remove when refresh this PR or in a following pR, thanks.

liw: Thanks. I'll see if this PR could be landed as it is, with a follow-up, doc-only PR to remove this comment.

liw:
There's one code comment that should be removed. I'd like to do it in a doc-only PR after the current PR lands, if possible.

gnailzenh (approved):

PR #16672, opened 2025-07-28
DAOS-17845 container: keep CONT_TGT_SNAPSHOT_NOTIFY RPC errno

Nasf-Fan:
Return CONT_TGT_SNAPSHOT_NOTIFY RPC errno to the snap_oit_create() sponsor instead of converting it as -DER_IO. Then related client side logic can know what the failure is.

### Steps for the author:

### DIFF
--- a/src/container/srv_target.c
+++ b/src/container/srv_target.c
@@ -2233,7 +2233,8 @@ ds_cont_tgt_snapshot_notify_aggregator(crt_rpc_t *source, crt_rpc_t *result,
 
 	out_source = crt_reply_get(source);
 	out_result = crt_reply_get(result);
-	out_result->tso_rc += out_source->tso_rc;
+	if (out_result->tso_rc >= 0 && out_source->tso_rc < 0)
+		out_result->tso_rc = out_source->tso_rc;
 	return 0;
 }
 
--- a/src/container/srv_epoch.c
+++ b/src/container/srv_epoch.c
@@ -220,9 +220,8 @@ snap_oit_create(struct rdb_tx *tx, struct cont *cont, uuid_t coh_uuid,
 	out = crt_reply_get(rpc);
 	rc = out->tso_rc;
 	if (rc != 0) {
-		D_ERROR(DF_CONT": snapshot notify failed on %d targets\n",
-			DP_CONT(cont->c_svc->cs_pool_uuid, cont->c_uuid), rc);
-		rc = -DER_IO;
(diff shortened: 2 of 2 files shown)


daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16672/1/testReport/

rpadma2:
Tested on 2.6. Reports proper failure information now.

rpadma2 (approved):

src/container/srv_target.c
```diff
@@ -2233,7 +2233,8 @@ ds_cont_tgt_snapshot_notify_aggregator(crt_rpc_t *source, crt_rpc_t *result,
 
 	out_source = crt_reply_get(source);
 	out_result = crt_reply_get(result);
-	out_result->tso_rc += out_source->tso_rc;
+	if (out_result->tso_rc >= 0 && out_source->tso_rc < 0)
```
kccain, line 2236: No change requested - just curious about the following. I imagine the test "out_result->tso_rc >= 0 ..." in this code is safest (compared to "out_result->tso_rc == 0").

Is it expected that tso_rc will always be 0 or less than 0 through the entire chain (dss_pool_thread_collective on each engine, the aggregated per-thread rc values from all threads in one engine, and the aggregated tso_rc value here across multple engines)? It seems that the code for thread collective and bcast corpc here takes the first observed error and assigns it to the aggregated rc value. So I guess in theory the test here could possibly be "(out_result->tso_rc == 0") && ..." ?

kccain (approved):

src/container/srv_target.c
Nasf-Fan: Currently, it is expected that all the components in the entire chain can return 0 zero as success and negative value for failure, but I cannot guarantee that, especially consider the original expected summary semantics. So checking ">= 0" maybe more robust. On the other hand, in spite of how to check, we only can return one failure to the sponsor: either the first one or the last one. So under multiple different failures case, we still cannot exactly know what happened on targets side if without related targets logs. Anyway, the patch is helpful for our current situation.

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16672/2/execution/node/1495/log

PR #16687, opened 2025-07-30
DAOS-17422 chk: remove for_all option for check repair

Nasf-Fan:
To avoid confused semantics, we do not support "for_all" option when exectue "dmg check repair" command. Related functionality will be implemented via enhancing "dmg check set-policy" when check instance running in subsequent patch.

Test-tag: cat_recov

### Steps for the author:

### DIFF
--- a/src/mgmt/srv_internal.h
+++ b/src/mgmt/srv_internal.h
@@ -143,13 +143,14 @@ int ds_mgmt_check_start(uint32_t rank_nr, d_rank_t *ranks, uint32_t policy_nr,
 int ds_mgmt_check_stop(int pool_nr, char **pools);
 int ds_mgmt_check_query(int pool_nr, char **pools, chk_query_head_cb_t head_cb,
 			chk_query_pool_cb_t pool_cb, void *buf);
-int ds_mgmt_check_prop(chk_prop_cb_t prop_cb, void *buf);
+int
+ds_mgmt_check_prop(chk_prop_cb_t prop_cb, void *buf);
+int
+ds_mgmt_check_act(uint64_t seq, uint32_t act);
 int
 ds_mgmt_pool_rebuild_stop(uuid_t pool_uuid, uint32_t force, d_rank_list_t *svc_ranks);
 int
      ds_mgmt_pool_rebuild_start(uuid_t pool_uuid, d_rank_list_t *svc_ranks);
-int
-     ds_mgmt_check_act(uint64_t seq, uint32_t act);
 bool ds_mgmt_check_enabled(void);
 
 /** srv_query.c */


daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16687/3/testReport/

src/tests/suite/daos_cr.c
```diff
@@ -1323,6 +1324,7 @@ cr_engine_interaction(void **state)
 	cr_cleanup(arg, &pool, 1);
 }
 
+#if 0
```
kjacque: Need to fully remove this test case?

src/chk/chk_srv.c
```diff
@@ -144,7 +145,7 @@ ds_chk_act_hdlr(crt_rpc_t *rpc)
 	struct chk_act_out	*cao = crt_reply_get(rpc);
 	int			 rc;
 
-	rc = chk_engine_act(cai->cai_gen, cai->cai_seq, cai->cai_cla, cai->cai_act, cai->cai_flags);
+	rc = chk_engine_act(cai->cai_gen, cai->cai_seq, cai->cai_act);
```
kjacque, line 148: Hmm, do we really want to fully remove the flags field? I think think it's fine to do that as long as you don't foresee further uses for it.

src/tests/suite/daos_cr.c
Nasf-Fan: I prefer to mark it as "skip()" instead of "#if 0" or remove. Otherwise, `daos_test -F -u sub_tests xxx` will not work properly.

src/chk/chk_srv.c
Nasf-Fan: Related "flags" field is still kept in the `CHK_ACT` RPC. So if we has the requirement to use flags in the future, we can add them back since `chk_engine_act()` is check engine side local API, will not affect related RPC layout.

Nasf-Fan:
Ping reviewers, thanks!

src/tests/suite/daos_cr.c
```diff
@@ -1348,6 +1349,9 @@ cr_repair_forall_leader(void **state)
 	int				 rc;
 	int				 i;
 
+	/* Skip for DAOS-17422. */
+	skip();
```
kjacque: Hmm, if these test cases are for functionality that no longer exists, I think they need to be removed rather than skipped. If this causes a problem in the test behavior, let's fix that instead of leaving in dead test cases.

Nasf-Fan: If we directly remove the test case, then "daos_test -F -u sub_tests" will not work properly. For example, if we remove CR4 in the patch without changing the other test cases' index, then "daos_test -F -u 3-8" will run CR9 or CR10 unexpectedly. So if we want to make that works, when either keep CR4's index or decrease the index for all the subsequent test cases after CR4. But if we change the subsequent test cases' index, then it may cause inconvenience or confused result when search `CRx:` on Jira. That is why I prefer to keep CR4 with `skip()` instead of directly removing them.

kjacque: In that case would it be better to remove the old logic and mark them "reserved", like we would do with a struct with deprecated fields? I think it is best not to leave dead test code we know is no longer valid.

Nasf-Fan: So you mean the following:

print_message("Skip obsolete test\n");
skip();
#if 0
...
#endif

right? If yes, that is fine for me.

kjacque: I suggest taking it a little further: remove the logic between the `#if 0` and `#endif` altogether and renaming the test cases to "reserved[0-9]" so the test number can be easily identified for re-use later.

Leaving the old test case names and logic in any form, commented/ifdefed out or not, implies to me that maybe they will be restored later. I don't think that's the case here.

Nasf-Fan: If the test case may be restored, then only inserting "skip()" will be enough, right?
So using "if 0" to comment out the test case, from my understanding, the benefit is that:
1. It is some kinds of "remove".
2. It is possible that the patch will be landed to master instead of 2.6. Keeping the test source code will make the test source code to be consistent among different versions at the most degree. Then in the future, the other patch will be more easy to be applied on all related versions.
3. We still can trace related Jira issue with the old source code.

shimizukko (approved):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16687/9/execution/node/1454/log

Nasf-Fan:
Resolve merge conflict.

Nasf-Fan:
Passed all CI tests.

src/tests/suite/daos_cr.c
Nasf-Fan: In the subsequent patch for the same Jira ticket DAOS-17422 (https://github.com/daos-stack/daos/pull/16712), one of disabled CR tests will be improved and reused. So keeping the test source with `#if0` is always not meaningless. @kjacque , how do you think? Thanks!

gnailzenh (approved):

PR #16741, opened 2025-08-14
DAOS-17905 object: fixes excessive waiting time in system xstream dur…

wangshilong:
…ing object ULT creation

During migration, the system xstream only wakes up in ds_migrate_query_status() (called by rebuild_tgt_status_check_ult()), where it sleeps for RBLD_CHECK_INTV (2 seconds). This caused the system xstream to remain blocked for 2-second intervals without performing any operations, severely degrading rebuild performance.

Benchmark results demonstrate the improvement:

Without patch: 99,780 objects processed in 120 seconds

With patch: 225,136 objects processed in 11 seconds

The patch achieves approximately 20x performance improvement

### Steps for the author:


kccain:
probably a good idea to test the patch with commit pragma "Features: rebuild" I think

src/object/srv_obj_migrate.c
```diff
@@ -1964,9 +1964,7 @@ migrate_system_enter(struct migrate_pool_tls *tls, int tgt_idx, bool *yielded)
 		D_DEBUG(DB_REBUILD, DF_RB ": tgt%d:%u max %u\n", DP_RB_MPT(tls), tgt_idx, tgt_cnt,
 			tls->mpt_inflight_max_ult / dss_tgt_nr);
 		*yielded = true;
-		ABT_mutex_lock(tls->mpt_inflight_mutex);
-		ABT_cond_wait(tls->mpt_inflight_cond, tls->mpt_inflight_mutex);
-		ABT_mutex_unlock(tls->mpt_inflight_mutex);
+		dss_sleep(1);
```
kccain: just out of curiosity, do you have any sense  of whether adjusting this to other small delays > 1 millisecond has a significant effect on the rate of objects processed?

kccain (dismissed):

src/object/srv_obj_migrate.c
wangshilong: I changed it to dss_sleep(0) locally, did not see obvious performance differences..

wangshilong: > just out of curiosity, do you have any sense of whether adjusting this to other small delays > 1 millisecond has a significant effect on the rate of objects processed?

I suppose that depends on speed of handling migration ults? eg by default limit is 4096 ULT per rank, if one rank could handle more than 4Million ult per second, that means 1 msec is too large, if speed is lower than that, probably >  1millisecd reduce polling overhead. Anyway, as a first quick/simple code, probably it is fine now.

NiuYawei: It's great to improve so much by this quick fix, but I think we really need to improve further to make these limitations per-xstream (at least for 2.8 or later version).

With current quick fix, all waiting ULTs still need to repeatedly "wake -> check condition -> sleep", that'll significantly slow down everything when the number of waiting ULTs is large.

NiuYawei (dismissed):

src/object/srv_obj_migrate.c
wangshilong: Agreed, we should make it per-stream limit, Let's push another PR for that. (probably for 2.8). notice this wake->check condition->sleep logic is only for system xstream, and currently there will be only ULT for this polling.

liuxuezhao (dismissed):

gnailzenh (approved):

kccain (approved):

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16741/3/execution/node/1636/log

wangshilong:
> Test stage Functional Hardware Large MD on SSD completed with status FAILURE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16741/3/execution/node/1636/log

Failed tests are known As DAOS-17751 && DAOS-17725

gnailzenh (approved):

PR #16748, opened 2025-08-20
DAOS-17910 doc: maximum number of targets per SSD

NiuYawei:
Target number is limited by number of SSDs, specify this limitation in daos_server.yml

### Steps for the author:

### DIFF
--- a/utils/config/daos_server.yml
+++ b/utils/config/daos_server.yml
@@ -304,6 +304,12 @@
 #  #
 #  # The server should have sufficiently many physical cores to support the
 #  # number of targets, plus the additional service threads.
+#  #
+#  # Besides the number of CPU cores, this target number is limited by the number of
+#  # SSDs listed in the bdev_list as well. One NVMe SSD in the bdev_list can be assigned
+#  # to at most 64 targets in pmem mode, or 21 targets in md-on-ssd mode (when the SSD
+#  # has three roles), so the maximum target number will be 'number_of_SSD * 64' in
+#  # pmem mode or 'number_of_SSD * 21' in md-on-ssd mode.
 #
 #  targets: 16
 #


utils/config/daos_server.yml
```diff
@@ -304,6 +304,12 @@
 #  #
 #  # The server should have sufficiently many physical cores to support the
 #  # number of targets, plus the additional service threads.
+#  #
+#  # Besides the number of CPU cores, this target number is limited by the number of
+#  # SSDs listed in the bdev_list as well. One NVMe SSD in the bdev_list can be assigned
+#  # to at most 64 targets in pmem mode, or 21 targets in md-on-ssd mode (when the SSD
+#  # has three roles), so the maximum target number will be 'number_of_SSD * 64' in
+#  # pmem mode or 'number_of_SSD * 21' in md-on-ssd mode.
```
NiuYawei, line 312: @Michael-Hennecke , is it fine to document this in server.yml only? It looks to me it's not necessary to put such details in admin guide or release notes, what do you think?

Michael-Hennecke: Yes it's fine to only document this in the sample server.yml.

Clarification question: Is it always 21 for md-on-ssd because each SSD *could* have 3 roles? or is it "64 divided by the number of bdev_roles that the SSD *actually* has assigned to it"? I understand it's 21 if all SSDs have all 3 roles, but there could be other configs, for example some SSDs have only data role and some others have wal+meta.

I will add a clarification here: https://github.com/daos-stack/daos/blame/c4ddd511f2d0ee13ca775c0327e7322210693c17/docs/release/support_matrix.md#L389 when we refresh that document for 2.8.

NiuYawei: Yes, it's "64 divided by the number of bdev_roles for the SSD", one SSD can have at most 3 roles, so the 21 is minimal limitation in md-on-ssd mode.

Michael-Hennecke (approved):

PR #16755, opened 2025-08-21
DAOS-17772 rebuild: refine some rebuild/EC agg/EC deg fetch process

liuxuezhao:
1. add container RDB KV ds_cont_prop_ec_agg_eph for EC aggregation epoch boundary, store it when bump and load it after restart.
2. synchronize the ec agg boundary before rebuild
3. change ec agg boundary IV refresh to be EAGER sync
4. wait discard's completion in ds_pool_tgt_discard_handler()
5. fix a cart IV sync bug that ignored GRP_VER err case's err code
6. fetch ec agg boundary IV in fetch RPC handler when it is zero
7. fix an EC agg partial update stripe processing bug 
8. fix a sgl sg_nr_out and iov_len process bug 
9. fix a EC degraded fetch bug related with punch extent
10. some refines for rebuild upbound epoch
    1) rename rebuild stable epoch to rebuild upbound epoch
    2) select "current hlc + epsilon" (ds_rebuild_get_upbound_eph())
        as the rebuild upbound epoch.
    3) use the global consistent rebuild upbound epoch as rebuild fence
        to avoid VOS aggregation cross the epoch.
   

### Steps for the author:

### DIFF
--- a/src/container/container_iv.c
+++ b/src/container/container_iv.c
@@ -449,6 +449,7 @@ cont_iv_ent_fetch(struct ds_iv_entry *entry, struct ds_iv_key *key,
 		  d_sg_list_t *dst, void **priv)
 {
 	struct cont_iv_entry	*src_iv;
+	struct cont_iv_entry     iv_entry = {0};
 	daos_handle_t		root_hdl;
 	d_iov_t			key_iov;
 	d_iov_t			val_iov;
@@ -496,7 +497,6 @@ again:
 				rc1 = ds_cont_hdl_rdb_lookup(entry->ns->iv_pool_uuid,
 							     civ_key->cont_uuid, &chdl);
 				if (rc1 == 0) {
-					struct cont_iv_entry	iv_entry = { 0 };
 					daos_prop_t		*prop = NULL;
 					struct daos_prop_entry	*prop_entry;
 					struct daos_co_status	stat = { 0 };
@@ -546,8 +546,38 @@ again:
 				} else {
 					rc = -DER_NONEXIST;
 				}
+			} else if (class_id == IV_CONT_AGG_EPOCH_BOUNDRY) {
+				uint64_t ec_agg_eph;
+
+				rc = ds_cont_ec_agg_eph_rdb_lookup(entry->ns->iv_pool_uuid,
+								   civ_key->cont_uuid, &ec_agg_eph);
+				if (rc == 0) {
+					uuid_copy(iv_entry.cont_uuid, civ_key->cont_uuid);
+					iv_entry.iv_agg_eph.eph  = ec_agg_eph;
+					iv_entry.iv_agg_eph.rank = dss_self_rank();
+					d_iov_set(&val_iov, &iv_entry, sizeof(iv_entry));
+					rc = dbtree_update(root_hdl, &key_iov, &val_iov);
+					DL_CDEBUG(
+					    rc != 0, DLOG_ERR, DB_MD, rc,
+					    DF_CONT ": dbtree_update ec_agg_eph " DF_X64,
+					    DP_CONT(entry->ns->iv_pool_uuid, civ_key->cont_uuid),
+					    ec_agg_eph);
+					if (rc)
+						goto failed;
+					/* on master node will not trigger on_refresh callback,
+					 * so need to explicitly refresh its own cont child's
+					 * sc_ec_agg_eph_boundary especially for the case of
+					 * restart that the value is zero.
+					 */
+					rc = ds_cont_tgt_refresh_agg_eph(entry->ns->iv_pool_uuid,
+									 civ_key->cont_uuid,
+									 ec_agg_eph);
+					if (rc == 0)
+						goto again;
+				}
 			}
 		}
+failed:
 		D_DEBUG(DB_MGMT, DF_CONT "lookup cont: rc " DF_RC "\n",
 			DP_CONT(entry->ns->iv_pool_uuid, civ_key->cont_uuid), DP_RC(rc));
 		D_GOTO(out, rc);
@@ -653,8 +683,15 @@ cont_iv_ent_update(struct ds_iv_entry *entry, struct ds_iv_key *key,
 		} else if (entry->iv_class->iv_class_id ==
 						IV_CONT_AGG_EPOCH_BOUNDRY) {
 			rc = cont_iv_ent_agg_eph_refresh(entry, key, src);
-			if (rc)
+			if (rc) {
+				if (rc == -DER_NONEXIST) {
+					DL_INFO(
+					    rc, DF_CONT " cont_iv_ent_agg_eph_refresh ignore",
+					    DP_CONT(entry->ns->iv_pool_uuid, civ_key->cont_uuid));
+					rc = 0;
+				}
 				D_GOTO(out, rc);
+			}
 		}
 	}
 
@@ -1116,8 +1153,7 @@ cont_iv_ec_agg_eph_update(void *ns, uuid_t cont_uuid, daos_epoch_t eph)
 int
 cont_iv_ec_agg_eph_refresh(void *ns, uuid_t cont_uuid, daos_epoch_t eph)
 {
-	return cont_iv_ec_agg_eph_update_internal(ns, cont_uuid, eph,
-						  0, CRT_IV_SYNC_LAZY,
+	return cont_iv_ec_agg_eph_update_internal(ns, cont_uuid, eph, 0, CRT_IV_SYNC_EAGER,
 						  IV_CONT_AGG_EPOCH_BOUNDRY);
 }
 
--- a/src/container/srv_container.c
+++ b/src/container/srv_container.c
@@ -1691,6 +1691,7 @@ cont_ec_agg_alloc(struct cont_svc *cont_svc, uuid_t cont_uuid,
 	uuid_copy(ec_agg->ea_cont_uuid, cont_uuid);
 	ec_agg->ea_servers_num = rank_nr;
 	ec_agg->ea_current_eph = 0;
+	ec_agg->ea_rdb_eph     = 0;
 	for (i = 0; i < rank_nr; i++) {
 		ec_agg->ea_server_ephs[i].rank = doms[i].do_comp.co_rank;
 		ec_agg->ea_server_ephs[i].eph = 0;
@@ -1805,6 +1806,7 @@ cont_refresh_vos_agg_eph_one(void *data)
 	if (cont_child->sc_ec_agg_eph_boundary < arg->min_eph)
 		cont_child->sc_ec_agg_eph_boundary = arg->min_eph;
 
+	cont_child->sc_ec_agg_eph_valid = 1;
 	ds_cont_child_put(cont_child);
 	return rc;
 }
@@ -1814,6 +1816,8 @@ ds_cont_tgt_refresh_agg_eph(uuid_t pool_uuid, uuid_t cont_uuid,
 			    daos_epoch_t eph)
 {
 	struct refresh_vos_agg_eph_arg	arg;
+	static uint64_t                 cnt;
+	static uint64_t                 gap = 1;
 	int				rc;
 
 	uuid_copy(arg.pool_uuid, pool_uuid);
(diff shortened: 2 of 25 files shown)


src/pool/srv_target.c
```diff
@@ -2438,8 +2438,13 @@ cont_discard_cb(daos_handle_t ih, vos_iter_entry_t *entry,
 
 put:
 	ds_cont_child_put(cont);
+	/* don't destroy vos container, to avoid ds_cont_tgt_refresh_agg_eph() failure,
+	 * later depend on container recovery process to handle it.
+	 */
+#if 0
 	if (rc == 0)
 		rc = ds_cont_child_destroy(arg->tgt_discard->pool_uuid, entry->ie_couuid);
+#endif
```
NiuYawei, line 2447: I think this change will fail the test REBUILD35, and cause reint problem in some cases, right? see commit 167737b6dcd7d4502e95a351ea589b4d5aeebb34

src/container/container_iv.c
```diff
@@ -546,8 +546,38 @@ cont_iv_ent_fetch(struct ds_iv_entry *entry, struct ds_iv_key *key,
 				} else {
 					rc = -DER_NONEXIST;
 				}
+			} else if (class_id == IV_CONT_AGG_EPOCH_BOUNDRY) {
+				uint64_t ec_agg_eph;
+
+				rc = ds_cont_ec_agg_eph_rdb_lookup(entry->ns->iv_pool_uuid,
+								   civ_key->cont_uuid, &ec_agg_eph);
```
NiuYawei, line 553: In what case this will be triggered? On leader switch?

liuxuezhao: on rebuild it calls ds_cont_fetch_ec_agg_boundary for each cont when rebuilt the cont and possibly will go to here on PS leader.
One example is on full system restart and rebuild started, all local cache is invalid and the IV fetch ec_agg_boundary will go to PS leader, trigger the code here.

src/pool/srv_target.c
liuxuezhao: I did some test that the REBUILD35 can pass, due to my previous fix related with DER_CONT_NONEXIST err code handling.
It is not perfect, probably can refine it after container recovery ready later as the comment says.

NiuYawei (dismissed):

src/container/srv_container.c
```diff
@@ -1820,112 +1822,256 @@ ds_cont_tgt_refresh_agg_eph(uuid_t pool_uuid, uuid_t cont_uuid,
...
+	ABT_rwlock_wrlock(svc->cs_lock);
+	rc = cont_lookup(&tx, svc, cont_uuid, &cont);
+	if (rc != 0) {
+		D_ERROR(DF_CONT ": Failed to look container: %d\n",
+			DP_CONT(svc->cs_pool_uuid, cont_uuid), rc);
+		D_GOTO(out_lock, rc);
+	}
+
+	d_iov_set(&value, &ec_agg_eph, sizeof(ec_agg_eph));
+	rc = rdb_tx_update(&tx, &cont->c_prop, &ds_cont_prop_ec_agg_eph, &value);
```
liw: [Question] Not familiar with the EC aggregation logic, just in case: Do we need to check that the new epoch value is > any existing epoch value?

liuxuezhao: I thought about it, not very sure if it is good to lookup+compare before update each time, but before calling cont_agg_eph_store it checked "if (min_eph > ec_agg->ea_rdb_eph)" and only call it when it is true.

src/rebuild/srv.c
```diff
@@ -2180,7 +2186,7 @@ regenerate_task_of_type(struct ds_pool *pool, pool_comp_state_t match_states, ui
 
 /* Regenerate the rebuild tasks when changing the leader. */
 int
-ds_rebuild_regenerate_task(struct ds_pool *pool, daos_prop_t *prop)
+ds_rebuild_regenerate_task(struct ds_pool *pool, daos_prop_t *prop, uint64_t delay_sec)
```
liw, line 2188: [Question] Looks like `delay_sec` will always be 0. Is this change still necessary?

liuxuezhao: right, in previous version it passed in a delay for reboot, in this version i removed that change.
Did not remove the function parameter although, not sure if later can be used.

src/container/srv_container.c
liw: Hmm, okay... It's not easy to see that. For example, in `cont_agg_eph_sync`, line 1980 check if the value in the RDB is loaded or not (because at line 1961 we ignore errors loading the value), so it's not obvious why line 2003 does not check if the value in the RDB is loaded or not. If not, and if `ea_rdb_eph` is 0, then we are not _really_ checking any stored value... Not saying this will cause any problem; it's just not easy to answer.

liw (dismissed):

src/container/srv_container.c
liuxuezhao: agreed that is is safer to do lookup+compare before update, I'll refresh it, thanks.

liuxuezhao: agreed that is is safer to do lookup+compare before update, I'll refresh it, thanks.

src/container/srv_container.c
```diff
@@ -1912,15 +1917,30 @@ cont_agg_eph_store(struct cont_svc *svc, uuid_t cont_uuid, uint64_t ec_agg_eph)
...
 
-	cont_put(cont);
+	if (ec_agg_eph > old_eph) {
+		d_iov_set(&value, &ec_agg_eph, sizeof(ec_agg_eph));
+		rc = rdb_tx_update(&tx, &cont->c_prop, &ds_cont_prop_ec_agg_eph, &value);
+		if (rc == 0)
+			rc = rdb_tx_commit(&tx);
+	} else {
+		D_DEBUG(DB_MD, DF_CONT ": bypass rdb update ec_agg_eph "DF_X64", in rdb eph "DF_X64,
+			DP_CONT(svc->cs_pool_uuid, cont_uuid), ec_agg_eph, old_eph);
```
liw: [Question] In this case, `rc` is 0, and `cont_agg_eph_sync` will set `ea_rdb_eph` to the value of `ec_agg_eph` here, which <= `old_eph`, and then use the lower of the two values for refreshing. Is that okay?

src/pool/srv_target.c
Nasf-Fan, line 2447: Some confused. Before incremental reintegration being landed, if without discarding the old container, how can we guarantee the old/stale (punched) object/key can be removed after reintegration?

src/container/srv_container.c
liuxuezhao: in real test should not cause problem but in code logic it is not very good.
I'll check and refine it later after do some test for another issue, thx
...
refreshed, the changed code is in the 3rd commit.

src/pool/srv_target.c
liuxuezhao: all object already discarded in obj_discard_cb(), here just keep the vos container now.

daosbuild3:
Test stage Functional Hardware Large completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16755/5/execution/node/1349/log

liw (dismissed):

src/container/srv_container.c
liw: Thanks for taking the time to address this.

src/container/container_iv.c
```diff
@@ -653,8 +683,18 @@ cont_iv_ent_update(struct ds_iv_entry *entry, struct ds_iv_key *key,
 		} else if (entry->iv_class->iv_class_id ==
 						IV_CONT_AGG_EPOCH_BOUNDRY) {
 			rc = cont_iv_ent_agg_eph_refresh(entry, key, src);
-			if (rc)
+			if (rc) {
+				/* container non-exist possibly due to in reintegrate the container
+				 * discarded ahead. Ignore such err.
+				 */
+				if (rc == -DER_NONEXIST) {
```
Nasf-Fan, line 687: You have already changed `cont_discard_cb()` to prevent such case, so the container will not be discarded for reintegration case?

Nasf-Fan (dismissed):

src/container/container_iv.c
liuxuezhao: ah right, such code is harmless although, for some orphan container cases may be useful. The comment should change, will change it later if need to refresh for some other issue. Mohamad's test still with a few problems.

liuxuezhao:
adding the 4th commit for Mahamad's testing.

src/object/srv_obj.c
```diff
@@ -1484,6 +1503,33 @@ obj_local_rw_internal(crt_rpc_t *rpc, struct obj_io_context *ioc, daos_iod_t *io
 			get_parity_list = ec_recov && is_parity_shard &&
 					  ((orw->orw_flags & ORF_EC_RECOV_SNAP) == 0);
 		}
+		if ((ec_deg_fetch || (ec_recov && get_parity_list)) &&
+		    ioc->ioc_coc->sc_ec_agg_eph_boundary == 0 &&
+		    ioc->ioc_coc->sc_pool->spc_pool->sp_reclaim != DAOS_RECLAIM_DISABLED) {
```
NiuYawei: This reclaim property could be changed at runtime, so I think fetching EC agg boundary is always necessary no matter if the reclaim property is 'disabled' at this moment, right?

liuxuezhao: to avoid the reclaim disabled from begining case as we discussed.

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16755/9/execution/node/1479/log

liuxuezhao:
please reference the "commits" history to select the new code change to review, thanks.

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16755/11/execution/node/1132/log

src/container/srv_layout.h
```diff
@@ -137,6 +138,7 @@ extern d_iov_t ds_cont_prop_co_md_times;	/* co_md_times */
 extern d_iov_t ds_cont_prop_cont_obj_version;	/* uint32_t */
 extern d_iov_t ds_cont_prop_nhandles;		/* uint32_t */
 extern d_iov_t ds_cont_prop_oit_oids;		/* snapshot OIT OID KVS */
+extern d_iov_t ds_cont_prop_ec_agg_eph;         /* uint64_t */
```
kccain, line 141: Typically changing the pool/container metadata would involve bumping DAOS_POOL_GLOBAL_VERSION in src/include/daos/pool.h. But we can't do that on release/2.6 branch, whose current version is 3 - this is because master (for 2.8 release) is bumped to version 4. Is it safe to keep DAOS_POOL_GLOBAL_VERSION=3 on this branch, even though earlier versions on release/2.6 do not have this entry in the container property KVS?

src/container/srv_container.c
```diff
@@ -1820,112 +1822,256 @@ ds_cont_tgt_refresh_agg_eph(uuid_t pool_uuid, uuid_t cont_uuid,
...
+		D_ERROR(DF_CONT ": Failed to look container: %d\n",
+			DP_CONT(svc->cs_pool_uuid, cont_uuid), rc);
+		D_GOTO(out_lock, rc);
+	}
+
+	d_iov_set(&value, &ec_agg_eph, sizeof(ec_agg_eph));
+	rc = rdb_tx_update(&tx, &cont->c_prop, &ds_cont_prop_ec_agg_eph, &value);
+	if (rc == 0)
+		rc = rdb_tx_commit(&tx);
+	DL_CDEBUG(rc != 0, DLOG_ERR, DB_MD, rc, DF_CONT ": rdb_tx_update ec_agg_eph " DF_X64,
```
kccain: minor, unsure if it is worth changing. if rc !=0 the log message does not make it clear if it was the rdb_tx_update or the rdb_tx_commit that failed.

src/rebuild/rebuild_iv.c
```diff
@@ -201,12 +201,10 @@ rebuild_iv_ent_refresh(struct ds_iv_entry *entry, struct ds_iv_key *key,
...
 
-		if (rpt->rt_stable_epoch == 0)
-			rpt->rt_stable_epoch = dst_iv->riv_stable_epoch;
-		else if (rpt->rt_stable_epoch != dst_iv->riv_stable_epoch)
-			D_WARN("leader change stable epoch from "DF_U64" to "
-			       DF_U64 "\n", rpt->rt_stable_epoch,
-			       dst_iv->riv_stable_epoch);
+		D_ASSERT(rpt->rt_upbound_eph != 0);
+		if (rpt->rt_upbound_eph != dst_iv->riv_stable_epoch)
+			D_WARN("leader change stable epoch from " DF_U64 " to " DF_U64 "\n",
```
kccain, line 206: if you need to refresh patch, could consider renaming in the log message: stable -> upper-bound

src/rebuild/rpc.h
```diff
@@ -56,7 +57,6 @@ extern struct crt_proto_format rebuild_proto_fmt;
 	((uint32_t)		(rsi_layout_ver)	CRT_VAR)
 
 #define DAOS_OSEQ_REBUILD_SCAN	/* output fields */		 \
-	((uint64_t)		(rso_stable_epoch)	CRT_VAR) \
```
kccain, line 59: Since the rebuild RPC protocol version (DAOS_REBUILD_VERSION 4) is not being updated perhaps it is safest to keep the uint64_t member in this definition, but rename it to rso_reserved or something like that?

src/rebuild/rebuild_iv.c
kccain: also, will this branch ever be taken now that rpt->rt_stable_epoch does not get updated anymore with the result of the scan broadcast RPC reply rso_stable_epoch ?

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16755/13/execution/node/1302/log

src/container/srv_layout.h
liuxuezhao: yes, no problem, as it does not affect any compatibility

src/rebuild/rebuild_iv.c
liuxuezhao: rso_stable_epoch removed.

src/rebuild/rpc.h
liuxuezhao: as server rolling upgrade does not supported now, the DAOS_REBUILD_VERSION is not really make sense now. 
But I'll add a u64 reservation if need to refresh the PR later.

NiuYawei (dismissed):

src/rebuild/rebuild_iv.c
kccain: Understood. But does this branch ever get taken?

daosbuild3:
Test stage Functional Hardware Large completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16755/16/execution/node/1519/log

kccain (dismissed):

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16755/18/testReport/

src/rebuild/srv.c
```diff
@@ -1613,6 +1603,13 @@ rebuild_task_ult(void *arg)
 		D_GOTO(out_task, rc = -DER_NONEXIST);
 	}
 
+	rc = ds_cont_svc_refresh_agg_eph(task->dst_pool_uuid);
+	if (rc) {
+		D_ERROR(DF_UUID ": ds_cont_svc_refresh_agg_eph failed, " DF_RC "\n",
+			DP_UUID(task->dst_pool_uuid), DP_RC(rc));
+		goto out_task;
```
Nasf-Fan: `goto out_pool;`

src/rebuild/rebuild_iv.c
```diff
@@ -201,12 +201,10 @@ rebuild_iv_ent_refresh(struct ds_iv_entry *entry, struct ds_iv_key *key,
...
+			       rpt->rt_upbound_eph, dst_iv->riv_stable_epoch);
```
Nasf-Fan, line 207: Confused, which one is changed? The `rt_upbound_eph` is not related with `riv_stable_epoch`, right?

src/pool/srv_pool.c
```diff
@@ -7339,7 +7342,7 @@ pool_svc_update_map(struct pool_svc *svc, crt_opcode_t opc, bool exclude_rank,
 	bool				updated;
 	int				rc;
 	char				*env;
-	daos_epoch_t			rebuild_eph = d_hlc_get();
+	daos_epoch_t                     rebuild_eph = ds_rebuild_get_upbound_eph();
```
Nasf-Fan: Strictly speaking, it is still not safe to generate`rebuild_eph` via adding `epsilon` against current HLC before really refreshing the pool map. Because we do not know how long the refreshing pool map will take. If it takes more time than `epsilon`, then some new update may happen with newer epoch (than `rebuild_eph`) before refreshing pool map completed. And then current reintegration may not process those `new` data. That is unexpected.

On the other hand, since we have `epsilon` windows for reintegration, we need to double check VOS layer overwritten related logic to guarantee that if some data is `prepared` but not real `committed` yet, then the overwritten (from reintegration) with same epoch can be properly processed instead of returning failure.

src/object/srv_obj.c
```diff
@@ -1483,6 +1502,32 @@ obj_local_rw_internal(crt_rpc_t *rpc, struct obj_io_context *ioc, daos_iod_t *io
 			get_parity_list = ec_recov && is_parity_shard &&
 					  ((orw->orw_flags & ORF_EC_RECOV_SNAP) == 0);
 		}
+		if ((ec_deg_fetch || (ec_recov && get_parity_list)) &&
+		    ioc->ioc_coc->sc_ec_agg_eph_valid == 0) {
+			struct ec_agg_boundary_arg arg;
+
+			arg.eab_pool = ioc->ioc_coc->sc_pool->spc_pool;
+			uuid_copy(arg.eab_co_uuid, ioc->ioc_coc->sc_uuid);
+			rc = dss_ult_execute(obj_fetch_ec_agg_boundary, &arg, NULL, NULL,
```
Nasf-Fan, line 1511: What is the benefit of creating another ULT to fetch EC agg boundary instead of directly call `obj_fetch_ec_agg_boundary`()?

src/object/srv_ec_aggregate.c
```diff
@@ -1074,7 +1074,7 @@ agg_update_parity(struct ec_agg_entry *entry, uint8_t *bit_map,
 	buf  = entry->ae_sgl.sg_iovs[AGG_IOV_DATA].iov_buf;
 	diff = entry->ae_sgl.sg_iovs[AGG_IOV_DIFF].iov_buf;
 
-	for (i = 0, j = 0; i < cell_cnt; i++) {
+	for (i = 0, j = 0; i < cell_cnt; i++, j++) {
```
Nasf-Fan, line 1077: How to guarantee the subsequent `while (!isset(bit_map, j)) j++;` will not out of boundary?

src/object/cli_ec.c
```diff
@@ -2677,6 +2679,11 @@ obj_ec_sgl_copy(d_sg_list_t *sgl, uint64_t off, void *buf, uint64_t size)
 	/* to copy data from [buf, buf + size) to sgl */
 	rc = daos_sgl_processor(sgl, true, &sgl_idx, size, oes_copy, &arg);
 	D_ASSERT(rc == 0);
+	if (sgl_idx.iov_offset == 0)
+		sgl->sg_nr_out = sgl_idx.iov_idx;
+	else
+		sgl->sg_nr_out = sgl_idx.iov_idx + 1;
```
Nasf-Fan, line 2685: Would you please to comment for the reason? Thanks

src/container/srv_container.c
```diff
@@ -1820,116 +1822,274 @@ ds_cont_tgt_refresh_agg_eph(uuid_t pool_uuid, uuid_t cont_uuid,
...
 
-				if (d_rank_in_rank_list(&fail_ranks, rank)) {
-					D_DEBUG(DB_MD, DF_CONT" skip %u\n",
-						DP_CONT(svc->cs_pool_uuid,
-							ec_agg->ea_cont_uuid),
-						rank);
-					continue;
-				}
+		/* for reboot case the ea_rdb_eph possibly higher than min_eph */
+		if (min_eph < ec_agg->ea_rdb_eph)
```
Nasf-Fan, line 2005: Why reboot may make `min_eph` to be older? Because the subsequent `cont_agg_eph_store` may be failed but still go ahead?

src/object/cli_ec.c
liuxuezhao: The PR refreshed, in EC data recovery it copy the recovered data back to original buffer, so it is a chance to update sgl->sg_nr_out.

src/container/srv_container.c
liuxuezhao: When reboot, the EC agg will do from 0 and then possibly get a smaller global epoch than previously stored ea_rdb_eph right?

src/object/srv_ec_aggregate.c
liuxuezhao: it assume the bit_map is correctly set/updated before.
But you are right that we probably can add an assertion to make sure j will not jump out of boundary.

src/object/srv_obj.c
liuxuezhao: obj_fetch_ec_agg_boundary() can only be executed on system XS 0, but here it is IO service XS.

src/rebuild/rebuild_iv.c
liuxuezhao: riv_stable_epoch is useless now, I wanted to delete it actually, but did not delete it just leave it temporarily to see if it can be used later.

src/rebuild/srv.c
liuxuezhao: you are right, I'll fix it. thx

src/pool/srv_pool.c
liuxuezhao: we may need some offline discuss.

liuxuezhao:
@reviewers, sorry that as with too many original commits, I recombined some commits together to make 5 a little bit clearer PRs/commits so did not keep original commit history.

kccain (dismissed):

Nasf-Fan (approved):

NiuYawei (approved):

PR #16783, opened 2025-08-29
DAOS-17941 client: re-organize shared memory related header files

wiliamhuang:
to consolidate multiple shared memory related header files
replace d_hash_string_u32() with d_hash_murmur64()
make shm_mutex_lock() use the same argument like pthread_mutex_lock()
make parent process wait for child process to acquire lock in test_shm

### Steps for the author:


daosbuild3:
Test stage Unit Test with memcheck on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16783/1/display/redirect

daosbuild3:
Test stage Unit Test bdev with memcheck on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16783/1/display/redirect

daosbuild3:
Test stage Unit Test on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16783/1/display/redirect

src/gurt/shm_dict.c
```diff
@@ -621,13 +619,13 @@ shm_ht_rec_find(d_shm_ht_loc_t shm_ht_loc, const char *key, const int len_key,
 	}
 
 	ht_head    = shm_ht_loc->ht_head;
-	hash       = d_hash_string_u32(key, len_key);
-	idx        = hash & (ht_head->n_bucket - 1);
+	idx        = (unsigned int)(d_hash_murmur64((const unsigned char *)key, len_key, 0) &
```
mchaarawi, line 622: this doesn't make sense to me.. you changed to use a 64 bit hash, but then you cast it back to 32 bit?
what is going on?

src/gurt/shm_dict.c
```diff
@@ -687,13 +687,13 @@ shm_ht_rec_find_insert(d_shm_ht_loc_t shm_ht_loc, const char *key, const int len
...
```
mchaarawi, line 690: same here

mchaarawi (changes requested):

src/gurt/shm_dict.c
wiliamhuang: The distribution of d_hash_string_u32() is bad and d_hash_murmur64() shows more even distribution in my previous test.  If you do not like the change here, we can add murmur32 function and call murmur32() then. Please let me know. Thank you!

wiliamhuang: Here a 64-bit hash is used to mod with the number of buckets in hash table. The result of mod is supposed to be smaller than the number of buckets anyway.
idx = hash % n_bucket;

wiliamhuang: casting to 32 bit is applied to the result of modulous operation.

wiliamhuang:
@mchaarawi @knard38 Could you please review? Thank you very much!

src/gurt/shm_dict.c
mchaarawi: > The distribution of d_hash_string_u32() is bad and d_hash_murmur64() shows more even distribution in my previous test. If you do not like the change here, we can add murmur32 function and call murmur32() then. Please let me know. Thank you!

i do not understand this comment. what does this have to do with what i like and what i don't like?
this has nothing to do about distribution. yes you should use murmur64, but what does your commend have to do with my comment about not casting it back to 32 bit??

mchaarawi: > Here a 64-bit hash is used to mod with the number of buckets in hash table. The result of mod is supposed to be smaller than the number of buckets anyway.
idx = hash % n_bucket;

ok thanks this is a more helpful comment. i did not realize you had the & there.

knard38: From what I remember, using only the 32 first bits of a 64 bits hash function can have some bad effects on the distribution.  As suggested, by @wiliamhuang , using the 32bits version of the murmur has function should be a better option.  If we want to stick to the 64 bits implementation something such as the following example may be an option:
```
(uint32_t)( d_hash_murmur64((const unsigned char *)key, len_key, 0) >> 32)
```
However, this is some old souvenir and could be not true with the murmur has function.
@mchaarawi , any opinion on this ?

mchaarawi: sorry i am confused. maybe i did not read this properly.
is this being used to calculate the bucket index? how many max buckets do we have in this case?

wiliamhuang: @knard38 In the modulous operation "idx = hash % n_bucket", all 64 bit of the hash are used. The result of modulous is smaller than the number of buckets. (uint32_t) is applied on the final result of (hash % n_bucket).

wiliamhuang: > sorry i am confused. maybe i did not read this properly. is this being used to calculate the bucket index? how many max buckets do we have in this case?

@mchaarawi Yes. This is used to calculate the bucket index. "n_bucket" is defined as an int type. This code is in hash table implementation. We do not know the value here. Normally, it is much smaller than one million.

wiliamhuang: > From what I remember, using only the 32 first bits of a 64 bits hash function can have some bad effects on the distribution. As suggested, by @wiliamhuang , using the 32bits version of the murmur has function should be a better option. If we want to stick to the 64 bits implementation something such as the following example may be an option:
> 
> ```
> (uint32_t)( d_hash_murmur64((const unsigned char *)key, len_key, 0) >> 32)
> ```
> 
> However, this is some old souvenir and could be not true with the murmur has function. @mchaarawi , any opinion on this ?

@knard38 Only using lower 32-bit of d_hash_murmur64() and murmur32() seem having comparable distribution in my test.

wiliamhuang:
@mchaarawi Can we land this PR? After it is landed, I will refresh my LRU PR. Thank you very much!

PR #16785, opened 2025-08-29
DAOS-17823 client: implement LRU cache in shared memory

wiliamhuang:
Least Recently Used (LRU) cache designed for shared memory is implemented. It has a fixed capacity. Fine-grained lock is used by dividing cache into many smaller sub-caches to achieve better performance in high concurrency.

### Steps for the author:


daosbuild3:
Test stage Build on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/1/execution/node/410/log

daosbuild3:
Test stage Build on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/1/execution/node/417/log

daosbuild3:
Test stage Build on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/1/execution/node/420/log

daosbuild3:
Test stage Build RPM on EL 8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/1/execution/node/394/log

daosbuild3:
Test stage Build on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/2/execution/node/415/log

daosbuild3:
Test stage Build on Leap 15.5 with Intel-C and TARGET_PREFIX completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/2/execution/node/421/log

daosbuild3:
Test stage Build RPM on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/2/execution/node/365/log

daosbuild3:
Test stage Build RPM on EL 8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/2/execution/node/366/log

daosbuild3:
Test stage Build on Leap 15.5 with Intel-C and TARGET_PREFIX completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/3/execution/node/408/log

daosbuild3:
Test stage Build on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/3/execution/node/415/log

daosbuild3:
Test stage Build on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/3/execution/node/420/log

daosbuild3:
Test stage Build RPM on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/3/execution/node/384/log

daosbuild3:
Test stage Build on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/4/execution/node/412/log

daosbuild3:
Test stage Build on Leap 15.5 with Intel-C and TARGET_PREFIX completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/4/execution/node/416/log

daosbuild3:
Test stage Build on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/4/execution/node/420/log

daosbuild3:
Test stage Build RPM on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/4/execution/node/365/log

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,473 @@
...
+		    cache_list[i].off_keylist + (size_key_buf / n_subcache);
+
+		/* form a linked list for all free nodes. first_av points to the head node. */
+		node_list = (shm_lru_node_t *)((long int)cache + cache_list[i].off_nodelist);
+		for (j = 0; j < (capacity_per_subcache - 1); j++)
+			node_list[j].off_hnext = j + 1;
+
+		/* invalid node index */
+		node_list[j].off_hnext = -1;
+		if (lru_cache)
```
knard38: If `lru_cache` is `NULL`, then we will have a memory leak with `cache`.
If I am correct `lru_cache` should be tested at the beginning and an error should be return if it is `NULL`

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,473 @@
...
+		rc = shm_mutex_init(&cache_list[i].lock);
+		if (rc)
+			return rc;
+	}
+
+	return 0;
+}
+
+static void
+lru_free_dymaic_buff(shm_lru_cache_t *cache)
```
knard38: Not sure to understand the name of this function.
```suggestion
lru_free_dynamic_buff(shm_lru_cache_t *cache)
```

src/gurt/shm_lru.c
wiliamhuang: You are right. I will fix it in my next commit. Thank you!

src/gurt/shm_lru.c
wiliamhuang: A good catch! I will fix it soon. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Fixed as suggested.

src/gurt/shm_lru.c
wiliamhuang: Fixed as suggested.

wiliamhuang:
@mchaarawi Could you please review? All CI tests previously passed in build 8 and 9.
Only minor changes were made after that.
https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16785/8/pipeline-overview/
I will fix one format issue in next commit. We probably need to merge "feature/dfs_dcache" with latest master to avoid the new NLT issue. Thank you!

src/gurt/shm_dict.c
```diff
@@ -669,6 +669,7 @@ shm_ht_rec_find_insert(d_shm_ht_loc_t shm_ht_loc, const char *key, const int len
 	d_shm_ht_rec_t  rec_next = NULL;
 	char           *value    = NULL;
 	int             rc;
+	int             padding;
```
mchaarawi: the padding fix here sounds unrelated to this PR and should be split into a different PR?

src/gurt/shm_dict.c
```diff
@@ -717,21 +718,22 @@ shm_ht_rec_find_insert(d_shm_ht_loc_t shm_ht_loc, const char *key, const int len
 			off_next = rec->next;
 		}
 	}
+	/* add padding space to make sure value is aligned by SHM_MEM_ALIGN */
```
mchaarawi: why is padding required again here?

src/include/gurt/shm_utils.h
```diff
@@ -376,4 +376,139 @@ shm_ht_rec_data(d_shm_ht_rec_loc_t rec_loc, int *err);
...
+/* Node of entry in LRU cache */
+typedef struct shm_lru_node {
+	/* key size*/
+	uint32_t    key_size;
+	/* data size*/
+	uint32_t    data_size;
+	/* store the offset to key */
+	long int    key;
+	/* store the offset to data */
+	long int    data;
```
mchaarawi: rename to off_key and off_data

src/include/gurt/shm_utils.h
```diff
@@ -376,4 +376,139 @@ shm_ht_rec_data(d_shm_ht_rec_loc_t rec_loc, int *err);
...
+	/* the offset to shm_lru_cache_var_t */
+	long int off_cache_list;
+
+	char     pad[8];
+} shm_lru_cache_t;
+
+/**
+ * create LRU cache
+ *
+ * \param[in] n_subcache	the number of subcache. Each subcache owns a private lock
```
mchaarawi: can you explain what subcache means here? as a user of this API, i would have no clue what this means.

src/include/gurt/shm_utils.h
```diff
@@ -376,4 +376,139 @@ shm_ht_rec_data(d_shm_ht_rec_loc_t rec_loc, int *err);
...
+
+/**
+ * query LRU cache
+ *
+ * \param[in] cache		LRU cache
+ * \param[in] key		key
+ * \param[in] key_size		size of key in bytes
+ *
+ * \param[out] node_found	LRU cache node containing the key
+ * \param[out] val		LRU cache node containing the key
```
mchaarawi: is the val allocated by user or by the library?
how is it freed?

src/include/gurt/shm_utils.h
```diff
@@ -376,4 +376,139 @@ shm_ht_rec_data(d_shm_ht_rec_loc_t rec_loc, int *err);
...
+	/* the offset to the array of preallocated array of nodes */
+	long int      off_nodelist;
+	/* the offset to the array of preallocated array of keys */
+	long int      off_keylist;
+	/* the offset to the array of preallocated array of data */
+	long int      off_datalist;
+
+	d_shm_mutex_t lock;
+	char          pad[8];
+} shm_lru_cache_var_t;
```
mchaarawi: i don't see why t his needs to be in the public header. maybe i missed something?

src/gurt/tests/test_shm.c
```diff
@@ -35,6 +35,145 @@
 
 pthread_t thread_list[MAX_THREAD];
 
+void
+test_lrucache(void **state)
```
mchaarawi, line 74: the test is very simplistic.. shouldn't there be some multithreaded / multiprocess testing to validate?

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,475 @@
...
+		for (j = 0; j < (capacity_per_subcache - 1); j++)
+			node_list[j].off_hnext = j + 1;
+
+		/* invalid node index */
+		node_list[j].off_hnext = -1;
+		*lru_cache = cache;
+
+		rc = shm_mutex_init(&cache_list[i].lock);
+		if (rc)
+			return rc;
```
mchaarawi: no free on err?

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,475 @@
...
+	    (key_size > 0 && data_size <= LRU_ALLOC_SIZE_THRESHOLD) ? (key_size * capacity) : 0;
+
+	/* the space pre-allocated for the array of data of all records if data have a fixed size */
+	size_data_buf =
+	    (data_size > 0 && data_size <= LRU_ALLOC_SIZE_THRESHOLD) ? (data_size * capacity) : 0;
+
+	shm_lru_cache_t *cache = (shm_lru_cache_t *)shm_alloc(
+	    sizeof(shm_lru_cache_t) + sizeof(shm_lru_cache_var_t) * n_subcache +
+	    sizeof(int) * capacity + sizeof(shm_lru_node_t) * capacity + size_key_buf +
+	    size_data_buf);
```
mchaarawi: this is repeated later, so not optimal.
but i'd like to see a comment here to explain this allocation please.

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,475 @@
...
+}
+
+/* Put key-data into cache */
+int
+shm_lru_put(shm_lru_cache_t *cache, void *key, uint32_t key_size, void *data, uint32_t data_size)
+{
+	int                  rc;
+	uint64_t             hash         = d_hash_murmur64(key, key_size, 0);
+	uint32_t             idx_subcache = (uint32_t)((hash >> 32) % cache->n_subcache);
+	uint32_t             index        = (uint32_t)(hash % cache->capacity);
```
mchaarawi: do you have some data that  show this splitting of high 32 to subcache and low 32 to index in subcache is optimal?

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,475 @@
...
+	int *off_hash =
+	    (int *)((long int)cache + (long int)cache_list[idx_subcache].off_hashbuckets);
+	int             offset   = off_hash[index];
+	char           *buf_data = NULL;
+	shm_lru_node_t *node;
+	shm_lru_node_t *node_head;
+	shm_lru_node_t *node_new;
+
+	shm_mutex_lock(&cache_list[idx_subcache].lock);
+	while (offset) {
```
mchaarawi, line 299: is this to handle collisions?

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,475 @@
...
+
+	*node_found = NULL;
+	if (cache->key_size != 0)
+		D_ASSERT(cache->key_size == key_size);
+
+	while (offset) {
+		node = (shm_lru_node_t *)((long int)cache + (long int)offset);
+		if (key_cmp(cache, node, key, key_size) == 0) {
+			lru_move_to_head(cache, &cache_list[idx_subcache], node);
+			*val = ((void *)cache + (long int)node->data);
```
mchaarawi: i can cause segfault if i pass NULL or invalid val?

src/gurt/shm_dict.c
wiliamhuang: Right. This fix is not related to this PR. I will create a separate PR for it. Thank you!

src/gurt/shm_dict.c
wiliamhuang: In case of a robust mutex is stored in the data of hash table record, SHM_MEM_ALIGN (4-byte) aligned is required to make robust mutex working as expected. The padding is enforced to make sure data stored in hash table is 4 bytes aligned.

src/include/gurt/shm_utils.h
wiliamhuang: No memory allocation is involved in shm_lru_get(). *val points to the data buffer. shm_lru_get() increases the reference count by 1 to avoid the record is evicted. After user gets the needed data from the node, it should call shm_lru_node_dec_ref() to decrease the reference count by 1.

src/gurt/shm_lru.c
wiliamhuang: Correct. All nodes in current bucket can be traversed as a linked list.

src/include/gurt/shm_utils.h
wiliamhuang: Other than store all records in a large cache, we split records into multiple partitions. Each partition is called as a subcache and uses its own lock. The partition selection is based on the higher 32 bits of the hash value of the key. We could remove the argument here and always set n_subcache as the number of cores available on current client node to simplify the API.

src/include/gurt/shm_utils.h
wiliamhuang: You are right. I will fix it. Thank you!

src/include/gurt/shm_utils.h
wiliamhuang: Thank you! I will rename them.

src/gurt/shm_lru.c
mchaarawi: this was not what i had meant. im just wondering why a while loop is needed here.
if we hashed the bucket and the key in that bucket, we get to the record directly, no?
my only thought if it was a collision and we have multiple entries there in a list too? maybe i misunderstood your code.

wiliamhuang: Yes. Collision (multiple entries in the same bucket) are handled here. key from user is compared with the key in entries until the key is found or reaching the end of the list with the while loop.

src/gurt/shm_lru.c
wiliamhuang: Now "val" is checked. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Code was updated to pick sub-cache and bucket in sub-cache using 64-bit hash.
I did check the distribution of 1) only using high 32bits of the hash 2) only using low 32bits of the hash 3) using 64 bits of the hash. The differences are indistinguishable.

src/gurt/shm_lru.c
wiliamhuang: Data layout was improved a little bit. Comments were added. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Error handling was added. Thank you!

src/gurt/tests/test_shm.c
wiliamhuang: Multithreaded test code was added. Thank you!

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16785/13/testReport/

daosbuild3:
Test stage Unit Test with memcheck on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16785/13/testReport/

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+	/* Not found, remove one node near tail and create new node */
+	if (sub_cache->size >= cache->capacity) {
+		rc = lru_remove_near_tail(cache, sub_cache);
+		if (rc) {
+			shm_mutex_unlock(&sub_cache->lock);
+			return rc;
+		}
+	}
+
+	rc = lru_create_node(cache, sub_cache, key, key_size, data, data_size, &node_new);
```
knard38, line 356: Sound strange to me that the created node have a ref counter equal to zero.
It will be more logical for me to increment this counter after having add it and let the management of removing it to the user with explicitly calling `shm_lru_node_dec_ref()`.
Otherwise, from my understanding, nodes will be silently deleted by the function `lru_remove_near_tail()` when the LRU will be filled.

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+	shm_mutex_unlock(&sub_cache->lock);
+	return 0;
+}
+
+void
+shm_lru_node_dec_ref(shm_lru_node_t *node)
+{
+	D_ASSERT(node != NULL);
+
+	atomic_fetch_add(&node->ref_count, -1);
```
knard38, line 387: As for my comment for the function `shm_lru_put()`, it sound strange to me to not delete nodes when the ref count is equal to zero: from my point of view, the node deleted should be deterministic for the end user.

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+	}
+	subcache->off_head = (int)((long int)node - (long int)cache);
+
+	if (!subcache->off_tail)
+		subcache->off_tail = (int)((long int)node - (long int)cache);
+}
+
+/* Remove the least recently used node from tail with zero reference count */
+static int
+lru_remove_near_tail(shm_lru_cache_t *cache, shm_lru_cache_var_t *subcache)
```
knard38, line 128: From my understanding, using such lazy delete policy of the nodes have two drawbacks:
- the node deleter are not deterministic for the end user
- the insertion cost as a worst case of O(n) as you could have a loop to over all the entries to find the node to remove.

If I am correct, making the node delete to be explicit should remove these two drawbacks: insertion and deletion will have a constant cost of O(1).

src/gurt/shm_lru.c
wiliamhuang: The reference count is used to avoid an entry is evicted when copying data from cache to user's buffer, so the reference count is increased by one in shm_lru_get(). After copying data is finished, shm_lru_node_dec_ref() is called to decrease the reference count. Yes. The nodes are deleted only in function lru_remove_near_tail() as you noted. We want to share the cache across processes. We cannot predict when the data will be used again, so we keep data persistent (similar to the cache used in Linux kernel) until shared memory is destroyed.

Example running a serial of jobs in batch,

app_1
app_2
app_3

To make cache data available in all apps, we need to keep them all the time unless cache is full. If needed, we could support an expiration time too with LRU.

src/gurt/shm_lru.c
wiliamhuang: As stated above, we want to share the cache across processes. We cannot predict when the data will be used again, so we keep data persistent (similar to the cache used in Linux kernel) until cache is full.

src/gurt/shm_lru.c
wiliamhuang: We are implementing a cache that can be shared across many processes. We could not predict when the cached entries will be used next time by another application.
The user of cache needs to copy the needed data from cache entry and decrease its reference count asap. Normally the number of cache entries actively being accessed could be ~100 or smaller. When a cache with partitions is used, we actually split the cache into many sub-caches. The number of entries with non-zero refence count near cache tail should be not large at all (smaller than 10).

wiliamhuang:
@mchaarawi @knard38 Could you please review? Thank you very much!

src/gurt/tests/test_shm.c
```diff
@@ -31,10 +31,265 @@
...
+
+	/* insert entries */
+	for (i = param->start; i < param->end; i++) {
+		shm_lru_put(param->cache, &i, sizeof(int), &param->data[i], sizeof(int));
+	}
+
+	/* verify entries exist */
+	for (i = param->start; i < param->end; i++) {
+		rc = shm_lru_get(param->cache, &i, sizeof(int), &node_found, (void **)&addr_val);
+		assert(rc == 0);
```
mchaarawi: you should not use assert here. you have to use the cmocka asserts. this comment applies to all the tests.

src/gurt/tests/test_shm.c
```diff
@@ -31,10 +31,265 @@
...
+	/* cache with partitions for better performance in multi-threads */
+	rc = shm_lru_create_cache(true, capacity, sizeof(int), sizeof(int), &cache);
+	assert_true(rc == 0);
+
+	gettimeofday(&tm1, NULL);
+	for (i = 0; i < MAX_THREAD; i++) {
+		thread_param_list[i].cache = cache;
+		thread_param_list[i].start = i * size_per_thread;
+		thread_param_list[i].end   = thread_param_list[i].start + size_per_thread;
+		thread_param_list[i].data  = data;
```
mchaarawi, line 237: im not sure this test is sufficient. what happens if threads are inserting same keys? reading each other's data while insertions are happening, how do we guarantee consistency?

src/gurt/shm_lru.c
mchaarawi: im asking bec we are not having O(1) insertion here. so the question goes back to the hashing and how fair it is.

src/gurt/shm_lru.c
mchaarawi: @knard38 not quite sure i  follow you on how the insertion and deletion can be O(1) always even in worst case sceario, in this design of sharded hash buckets. even if you make the delete explicit, how is that correlated? maybe i misunderstood your point.

src/gurt/tests/test_shm.c
wiliamhuang: All assert() are replace with assert_true(). Thank you!

src/gurt/tests/test_shm.c
wiliamhuang: Now all keys in the full range are tested in all threads.
Currently, we assume that data are same for the same key if multiple put() are trying to insert/update the same key. There could be consistency issue if this assumption is false.
I think we do not need to worry about the consistency in our current use cases since only read-only mounts are considered at this moment.

src/gurt/shm_lru.c
wiliamhuang: Murmur hashing is one of the most common hash functions used in practice. I tried multiple hashing functions, murmur64(), rapidhash (https://github.com/Nicoshev/rapidhash), Jenkins hash function (https://en.wikipedia.org/wiki/Jenkins_hash_function). The distributions are indistinguishable. All distributions are even in tests.

src/gurt/shm_lru.c
knard38: Thanks for the information, it was a misunderstanding from my side.

src/gurt/shm_lru.c
knard38: It was a misunderstanding from my side on the way to use this cache.

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+		return 1;
+
+	return memcmp((char *)cache + node->off_key, key, key_size);
+}
+
+int
+shm_lru_get(shm_lru_cache_t *cache, void *key, uint32_t key_size, shm_lru_node_t **node_found,
+	    void **val)
+{
+	uint64_t hash = d_hash_murmur64(key, key_size, 0);
```
knard38: All the variable initialisation implying parameters `cache` and `key` should be done after the test of line 233.
Otherwise a SEGFAULT could be raised.

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+
+	return ENOENT;
+}
+
+/* Put key-data into cache */
+int
+shm_lru_put(shm_lru_cache_t *cache, void *key, uint32_t key_size, void *data, uint32_t data_size)
+{
+	int      rc;
+	uint64_t hash = d_hash_murmur64(key, key_size, 0);
```
knard38: All the variable initialisation implying parameters `cache` and `key` should be done after the test of line 282.
Otherwise a SEGFAULT could be raised.

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+				if (node->data_size > LRU_ALLOC_SIZE_THRESHOLD ||
+				    cache->data_size == 0)
+					shm_free((char *)cache + node->off_data);
+
+				if ((data_size > LRU_ALLOC_SIZE_THRESHOLD) ||
+				    (cache->data_size == 0)) {
+					/* data are too long or has unknown size,
+					 * dynamically allocate space for data
+					 */
+					buf_data       = shm_alloc(data_size);
```
knard38: `buf_data == NULL` should be managed

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+					node->off_data = (long int)buf_data - (long int)cache;
+					memcpy(buf_data, data, data_size);
+				} else {
```
knard38, line 328: From my understanding, If the previous data of the node was `node->data_size < LRU_ALLOC_SIZE_THRESHOLD` then this `memcpy()` will write on a deallocated memory space

src/gurt/shm_lru.c
knard38: Missunderstanding from my side

src/gurt/shm_internal.h
```diff
@@ -80,4 +80,80 @@ extern struct d_shm_hdr *d_shm_head;
...
+	d_shm_mutex_t lock;
+	char          pad[8];
+} shm_lru_cache_var_t;
+
+/* LRU Cache structure */
+struct shm_lru_cache {
+	/* the number of sub-cache to use */
+	uint32_t n_subcache;
+	/* max number of nodes to hold in each shard */
+	uint32_t capacity;
```
knard38: NIT, could be nice to explicitly indicate that it is per subcache
```suggestion
	uint32_t capacity_per_subcache;
```

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+	}
+
+	/* Not found, remove one node near tail and create new node */
+	if (sub_cache->size >= cache->capacity) {
+		rc = lru_remove_near_tail(cache, sub_cache);
+		if (rc) {
+			shm_mutex_unlock(&sub_cache->lock);
+			return rc;
+		}
+	}
```
knard38, line 353: NIT, could help to have an assert
```suggestion
	}
	D_ASSERT(sub_cache->size < cache->capacity);
```

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+	shm_lru_cache_var_t *sub_cache;
+	shm_lru_cache_t     *cache;
+	long int             num_cores;
+
+	if (lru_cache == NULL)
+		return EINVAL;
+
+	if (auto_partition) {
+		/* query the number of cores on current machine */
+		num_cores = sysconf(_SC_NPROCESSORS_ONLN);
```
knard38: From my understanding, the `_SC_NPROCESSORS_ONLN` is not a linux standard and was introduced in recent libc.
More critical, it does not return the number of physical core, but the number of logical core: when HT is activated the number of CPU will be overestimated.

One solution, could be to use the hwloc library, on which the rpm daos-client has already a dependency.
```c
#include <unistd.h>
#include <hwloc.h>
#include <stdio.h>

int
get_cpu_core(void)
{
        hwloc_topology_t topology;
        int cpu_count;

        // Allocate, initialize, and perform topology detection
        hwloc_topology_init(&topology);
        hwloc_topology_load(topology);

        // Try to get the number of CPU cores from topology
        int depth = hwloc_get_type_depth(topology, HWLOC_OBJ_CORE);
        if(depth == HWLOC_TYPE_DEPTH_UNKNOWN) {
                printf("*** The number of cores is unknown\n");
                cpu_count = -1;
                goto out;
        }

        cpu_count = hwloc_get_nbobjs_by_depth(topology, depth);

out:
        // Destroy topology object and return
        hwloc_topology_destroy(topology);
        return cpu_count;
}
```

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,513 @@
...
+			node = (shm_lru_node_t *)((long int)cache + (long int)node->off_prev);
+	}
+
+	/* node will be removed. Free dynamically allocated key/data space */
+	if (node->key_size > LRU_ALLOC_SIZE_THRESHOLD || cache->key_size == 0)
+		shm_free((char *)cache + node->off_key);
+	if (node->data_size > LRU_ALLOC_SIZE_THRESHOLD || cache->data_size == 0)
+		shm_free((char *)cache + node->off_data);
+
+	/* updata the linked list in the hash bucket */
```
knard38: NIT
```suggestion
	/* update the linked list in the hash bucket */
```

src/gurt/shm_lru.c
wiliamhuang: The rule for dynamic allocation of key and data in LRU cache is simplified. The issue should be resolved. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Fixed as you suggested. Thank you very much for your code!

src/gurt/shm_internal.h
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Fixed as you suggested. Thank you!

src/gurt/shm_alloc.c
```diff
@@ -47,14 +48,47 @@ static uint64_t       page_size;
...
+#define INVALID_NUM_CORE (0xFFFFFFFF)
+static unsigned int
+get_cpu_core(void)
+{
+	hwloc_topology_t topology;
+	unsigned int     cpu_count;
+	int              depth;
+
+	/* Allocate, initialize, and perform topology detection */
+	hwloc_topology_init(&topology);
```
knard38: My apologies to not have done it in the example, but the return code of the function calls to `hwloc_topology_init()` and `hwloc_topology_load()` should be checked.
From my understanding, the call to `hwloc_topology_destroy()` have to be done only if `hwloc_topology_init()` succeed.

knard38 (approved):
LGTM

src/gurt/shm_alloc.c
wiliamhuang: Thank you very much! Fixed as you suggested.

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16785/21/execution/node/1554/log

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,535 @@
...
+	shm_lru_node_t *node_list;
+	shm_lru_node_t *node;
+	char           *buf_key  = NULL;
+	char           *buf_data = NULL;
+	long int        idx_node;
+
+	*new_node = NULL;
+
+	/* make sure LRU is not full */
+	if (subcache->first_av < 0)
```
mchaarawi, line 33: I do not understand how an LRU can be full. by definition it should kick the oldest element out.
did i misunderstand this here?

src/gurt/shm_lru.c
```diff
@@ -0,0 +1,535 @@
...
+		 * space for key
+		 */
+		buf_key = shm_alloc(key_size);
+		if (buf_key == NULL)
+			return ENOMEM;
+		node->off_key = (long int)buf_key - (long int)cache;
+		memcpy(buf_key, key, key_size);
+	} else {
+		/* use preallocated buffer for key */
+		node->off_key = subcache->off_keylist + idx_node * (long int)cache->key_size;
```
mchaarawi, line 52: do we ever grow the lru size anywhere or it is always fixed?

mchaarawi (approved):

src/gurt/shm_lru.c
wiliamhuang: Yes. That's true for regular LRU cache.
In current implementation, data copy is not done in-place and the reference count is increased by 1 after we call shm_lru_get(). The caller gets the pointer of cache entry data. The reference count is decrease once user gets all needed data from the cache entry. This allows multiple consumers to read the same cache entry concurrently.
The entry with non-zero reference count cannot be evicted. When we search an entry starting from the tail for eviction, we skip such entries since users may be reading data from them.

src/gurt/shm_lru.c
wiliamhuang: LRU size is fixed.

wiliamhuang:
@mchaarawi The failure in Functional Hardware Medium MD on SSD / FTEST_io.IoConsistency.1-./io/io_consistency.py is not related. Can we force land this PR? Thank you very much!

src/gurt/shm_lru.c
mchaarawi: i see.. we probably should do some dynamic solution there.
meanwhile the put should document that on EBUSY one needs to  retry or just disregard putting the element as this is not fatal.

mchaarawi:
please add a description for landing..

wiliamhuang:
> please add a description for landing..

Done as you suggested. Thank you!

src/gurt/shm_lru.c
wiliamhuang: Thank you! I will add the document as you suggested in another PR.
Regarding "some dynamic solution", could you please provide more information if there is any suggestion in your mind? Thank you!

PR #16788, opened 2025-08-29
Modify DAOS agent to enable generic authentication

JonahRosenblumHPE:
The DAOS agent currently authenticates users using Unix Domain Sockets (UDS). While this works for many users, there is a growing demand for alternative authentication methods. This pull request reworks the DAOS agent code to make it simple to implement new sources of authentication for use in addition to/besides UDS. We also added a new example alternative authentication method in use at our company (part of a project soon to be open-sourced), and are happy to leave it in the codebase as an example.

### Steps for the author:

### DIFF
--- a/src/control/cmd/daos_agent/security_rpc.go
+++ b/src/control/cmd/daos_agent/security_rpc.go
@@ -10,11 +10,11 @@ package main
 import (
 	"context"
 	"fmt"
-	"net"
-	"os/user"
+	"slices"
 	"time"
 
 	"github.com/pkg/errors"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/daos-stack/daos/src/control/drpc"
 	"github.com/daos-stack/daos/src/control/lib/cache"
@@ -26,7 +26,7 @@ import (
 
 type (
 	// credSignerFn defines the function signature for signing credentials.
-	credSignerFn func(context.Context, *auth.CredentialRequest) (*auth.Credential, error)
+	credSignerFn func(context.Context, logging.Logger, auth.CredentialRequest) (*auth.Credential, error)
 
 	// credentialCache implements a cache for signed credentials.
 	credentialCache struct {
@@ -48,6 +48,8 @@ type (
 	securityConfig struct {
 		credentials *security.CredentialConfig
 		transport   *security.TransportConfig
+		infoCache   *InfoCache
+		sys         string
 	}
 
 	// SecurityModule is the security drpc module struct
@@ -55,23 +57,43 @@ type (
 		log            logging.Logger
 		signCredential credSignerFn
 		credCache      *credentialCache
-
-		config *securityConfig
+		config         *securityConfig
+		infoCache      *InfoCache
 	}
 )
 
 var _ cache.ExpirableItem = (*cachedCredential)(nil)
 
+func getCredReq(reqb []byte) (*auth.GetCredReq, error) {
+	credReq := new(auth.GetCredReq)
+
+	reqbSize := len(reqb)
+	if reqbSize == 0 {
+		credReq.Flavor = auth.AuthSysCredentialFactory{}.GetAuthFlavor()
+	} else {
+		if err := proto.Unmarshal(reqb, credReq); err != nil {
+			return nil, drpc.UnmarshalingPayloadFailure()
+		}
+	}
+
+	return credReq, nil
+}
+
+// Wrapper function, helps with fulfilling cache interface.
+func credentialRequestGetSigned(ctx context.Context, log logging.Logger, req auth.CredentialRequest) (*auth.Credential, error) {
+	return req.GetSignedCredential(log, ctx)
+}
+
 // NewSecurityModule creates a new module with the given initialized TransportConfig.
 func NewSecurityModule(log logging.Logger, cfg *securityConfig) *SecurityModule {
 	var credCache *credentialCache
-	credSigner := auth.GetSignedCredential
+	credSigner := credentialRequestGetSigned
 	if cfg.credentials.CacheExpiration > 0 {
 		credCache = &credentialCache{
 			log:          log,
 			cache:        cache.NewItemCache(log),
 			credLifetime: cfg.credentials.CacheExpiration,
-			cacheMissFn:  auth.GetSignedCredential,
+			cacheMissFn:  credentialRequestGetSigned,
 		}
 		credSigner = credCache.getSignedCredential
 		log.Noticef("credential cache enabled (entry lifetime: %s)", cfg.credentials.CacheExpiration)
@@ -82,11 +104,8 @@ func NewSecurityModule(log logging.Logger, cfg *securityConfig) *SecurityModule 
 		signCredential: credSigner,
 		credCache:      credCache,
 		config:         cfg,
+		infoCache:      cfg.infoCache,
 	}
-}
-
-func credReqKey(req *auth.CredentialRequest) string {
-	return fmt.Sprintf("%d:%d:%s", req.DomainInfo.Uid(), req.DomainInfo.Gid(), req.DomainInfo.Ctx())
 }
 
 // Key returns the key for the cached credential.
@@ -107,12 +126,12 @@ func (cred *cachedCredential) IsExpired() bool {
 	return time.Now().After(cred.expiredAt)
 }
 
-func (cc *credentialCache) getSignedCredential(ctx context.Context, req *auth.CredentialRequest) (*auth.Credential, error) {
-	key := credReqKey(req)
+func (cc *credentialCache) getSignedCredential(ctx context.Context, log logging.Logger, req auth.CredentialRequest) (*auth.Credential, error) {
+	key := req.GetKey()
 
 	createItem := func() (cache.Item, error) {
 		cc.log.Tracef("cache miss for %s", key)
-		cred, err := cc.cacheMissFn(ctx, req)
+		cred, err := cc.cacheMissFn(ctx, log, req)
 		if err != nil {
 			return nil, err
 		}
@@ -146,77 +165,71 @@ func newCachedCredential(key string, cred *auth.Credential, lifetime time.Durati
 	}, nil
 }
 
-// GetMethod gets the corresponding Method for a method ID.
-func (m *SecurityModule) GetMethod(id int32) (drpc.Method, error) {
-	if id == daos.MethodRequestCredentials.ID() {
-		return daos.MethodRequestCredentials, nil
-	}
-
-	return nil, fmt.Errorf("invalid method ID %d for module %s", id, m.String())
-}
-
-func (m *SecurityModule) String() string {
-	return "agent_security"
-}
-
 // HandleCall is the handler for calls to the SecurityModule
-func (m *SecurityModule) HandleCall(ctx context.Context, session *drpc.Session, method drpc.Method, body []byte) ([]byte, error) {
-	if method != daos.MethodRequestCredentials {
-		return nil, drpc.UnknownMethodFailure()
+func (m *SecurityModule) HandleCall(ctx context.Context, session *drpc.Session, method drpc.Method, reqb []byte) ([]byte, error) {
+	credReq, err := getCredReq(reqb)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse request body")
 	}
 
-	return m.getCredential(ctx, session)
+	switch method {
+	case daos.MethodRequestCredentials:
+		validAuthFlavors, err := m.retrieveAuthFromServer(ctx)
+		if err != nil || len(validAuthFlavors) == 0 {
+			return nil, errors.Wrap(err, "error in retrieving auth flavors from server")
+		}
+
+		if !slices.Contains(validAuthFlavors, credReq.Flavor) {
+			return nil, errors.Errorf("invalid authentication method: the method requested is not allowed by the server configuration.")
+		}
+		return m.getCredential(ctx, session, credReq)
+	case daos.MethodRequestValidFlavors:
+		return m.getValidAuthFlavors(ctx)
+	}
+
+	return nil, drpc.UnknownMethodFailure()
 }
 
-// getCredentials generates a signed user credential based on the data attached to
-// the Unix Domain Socket.
-func (m *SecurityModule) getCredential(ctx context.Context, session *drpc.Session) ([]byte, error) {
-	if session == nil {
-		return nil, drpc.NewFailureWithMessage("session is nil")
-	}
-
-	uConn, ok := session.Conn.(*net.UnixConn)
-	if !ok {
-		return nil, drpc.NewFailureWithMessage("connection is not a unix socket")
-	}
-
-	info, err := security.DomainInfoFromUnixConn(m.log, uConn)
+func (m *SecurityModule) retrieveAuthFromServer(ctx context.Context) ([]auth.Flavor, error) {
+	resp, err := m.infoCache.GetAttachInfo(ctx, m.config.sys)
 	if err != nil {
-		m.log.Errorf("Unable to get credentials for client socket: %s", err)
-		return m.credRespWithStatus(daos.MiscError)
+		return nil, errors.Wrap(err, "failed to get attach info")
 	}
 
+	validAuthFlavors := make([]auth.Flavor, len(resp.ValidAuthFlavors))
+	for i := 0; i < len(resp.ValidAuthFlavors); i++ {
+		validAuthFlavors[i] = auth.Flavor(resp.ValidAuthFlavors[i])
+	}
+
+	if len(validAuthFlavors) == 0 {
+		return nil, errors.Errorf("failed to receive valid authentication flavors from server.")
+	}
+
+	return validAuthFlavors, nil
+}
+
+// getCredentials generates a signed user credential based on the authentication method requested.
+func (m *SecurityModule) getCredential(ctx context.Context, session *drpc.Session, credReq *auth.GetCredReq) ([]byte, error) {
 	signingKey, err := m.config.transport.PrivateKey()
 	if err != nil {
-		m.log.Errorf("%s: failed to get signing key: %s", info, err)
+		m.log.Errorf("failed to get signing key: %s", err)
 		// something is wrong with the cert config
 		return m.credRespWithStatus(daos.BadCert)
 	}
 
-	req := auth.NewCredentialRequest(info, signingKey)
-	cred, err := m.signCredential(ctx, req)
+	req, err := auth.FlavorToFactory[credReq.Flavor].Init(m.log, m.config.credentials, session, credReq.Data, signingKey)
 	if err != nil {
-		if err := func() error {
-			if !errors.Is(err, user.UnknownUserIdError(info.Uid())) {
-				return err
-			}
-
-			mu := m.config.credentials.ClientUserMap.Lookup(info.Uid())
-			if mu == nil {
-				return user.UnknownUserIdError(info.Uid())
-			}
-
-			req.WithUserAndGroup(mu.User, mu.Group, mu.Groups...)
-			cred, err = m.signCredential(ctx, req)
-			if err != nil {
-				return err
-			}
-
-			return nil
-		}(); err != nil {
-			m.log.Errorf("%s: failed to get user credential: %s", info, err)
-			return m.credRespWithStatus(daos.MiscError)
+		if errors.Is(err, daos.MiscError) {
+			return m.credRespWithStatus(err.(daos.Status))
 		}
+		m.log.Errorf("Unable to get credentials for client socket: %s", err)
+		return nil, err
+	}
+
+	cred, err := m.signCredential(ctx, m.log, req)
+	if err != nil {
+		m.log.Errorf("failed to get user credential: %s", err)
+		return m.credRespWithStatus(daos.FailedSign)
 	}
 
 	resp := &auth.GetCredResp{Cred: cred}
@@ -226,6 +239,31 @@ func (m *SecurityModule) getCredential(ctx context.Context, session *drpc.Sessio
 func (m *SecurityModule) credRespWithStatus(status daos.Status) ([]byte, error) {
 	resp := &auth.GetCredResp{Status: int32(status)}
 	return drpc.Marshal(resp)
+}
+
+func (m *SecurityModule) getValidAuthFlavors(ctx context.Context) ([]byte, error) {
+	validAuthFlavors, err := m.retrieveAuthFromServer(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "error in retrieving auth flavors from server")
+	}
+
+	resp := &auth.GetValidFlavorsResp{ValidAuthFlavors: validAuthFlavors}
+	return drpc.Marshal(resp)
+}
+
+// GetMethod gets the corresponding Method for a method ID.
+func (m *SecurityModule) GetMethod(id int32) (drpc.Method, error) {
+	if id == daos.MethodRequestCredentials.ID() {
+		return daos.MethodRequestCredentials, nil
+	} else if id == daos.MethodRequestValidFlavors.ID() {
+		return daos.MethodRequestValidFlavors, nil
+	}
+
+	return nil, fmt.Errorf("invalid method ID %d for module %s", id, m.String())
+}
+
+func (m *SecurityModule) String() string {
+	return "agent_security"
 }
 
 // ID will return Security module ID
--- a/src/control/cmd/daos_agent/security_rpc_test.go
+++ b/src/control/cmd/daos_agent/security_rpc_test.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/daos-stack/daos/src/control/common/test"
 	"github.com/daos-stack/daos/src/control/drpc"
+	"github.com/daos-stack/daos/src/control/lib/cache"
+	"github.com/daos-stack/daos/src/control/lib/control"
 	"github.com/daos-stack/daos/src/control/lib/daos"
 	"github.com/daos-stack/daos/src/control/logging"
 	"github.com/daos-stack/daos/src/control/security"
@@ -33,7 +35,7 @@ func TestAgent_SecurityModule_ID(t *testing.T) {
 	log, buf := logging.NewTestLogger(t.Name())
 	defer test.ShowBufferOnFailure(t, buf)
 
-	mod := NewSecurityModule(log, defaultTestSecurityConfig())
+	mod := NewSecurityModule(log, defaultTestSecurityConfig(t, log, testInfoCacheParams{}))
 
 	test.AssertEqual(t, mod.ID(), daos.ModuleSecurityAgent, "wrong drpc module")
 }
@@ -57,7 +59,7 @@ func TestAgent_SecurityModule_GetMethod(t *testing.T) {
 			parent := test.MustLogContext(t)
 			log := logging.FromContext(parent)
 
-			mod := NewSecurityModule(log, defaultTestSecurityConfig())
+			mod := NewSecurityModule(log, defaultTestSecurityConfig(t, log, testInfoCacheParams{}))
 
 			method, err := mod.GetMethod(tc.methodID)
 
@@ -72,10 +74,27 @@ func newTestSession(t *testing.T, log logging.Logger, conn net.Conn) *drpc.Sessi
 	return drpc.NewSession(conn, svc)
 }
 
-func defaultTestSecurityConfig() *securityConfig {
+func defaultInfoCache(t *testing.T, log logging.Logger, params testInfoCacheParams) *InfoCache {
+	expSys := "GetAttachInfo-daos_server"
+	params.cachedItems = []cache.Item{
+		newCachedAttachInfo(0, expSys, nil,
+			func(_ context.Context, _ control.UnaryInvoker, _ *control.GetAttachInfoReq) (*control.GetAttachInfoResp, error) {
+				return &control.GetAttachInfoResp{ValidAuthFlavors: []auth.Flavor{auth.Flavor_AUTH_SYS}}, nil
+			}),
+	}
+	ic := newTestInfoCache(t, log, params)
+	ic.getAttachInfoCb = func(_ context.Context, _ control.UnaryInvoker, _ *control.GetAttachInfoReq) (*control.GetAttachInfoResp, error) {
+		return &control.GetAttachInfoResp{ValidAuthFlavors: []auth.Flavor{auth.Flavor_AUTH_SYS}}, nil
+	}
+
+	return ic
+}
+
+func defaultTestSecurityConfig(t *testing.T, log logging.Logger, params testInfoCacheParams) *securityConfig {
 	return &securityConfig{
 		transport:   &security.TransportConfig{AllowInsecure: true},
 		credentials: &security.CredentialConfig{},
+		infoCache:   defaultInfoCache(t, log, params),
 	}
 }
 
@@ -134,7 +153,7 @@ func TestAgentSecurityModule_RequestCreds_OK(t *testing.T) {
 	conn, cleanup := setupTestUnixConn(t)
 	defer cleanup()
 
-	mod := NewSecurityModule(log, defaultTestSecurityConfig())
+	mod := NewSecurityModule(log, defaultTestSecurityConfig(t, log, testInfoCacheParams{}))
 	respBytes, err := callRequestCreds(mod, t, log, conn)
 
 	if err != nil {
@@ -148,7 +167,7 @@ func TestAgentSecurityModule_RequestCreds_NotUnixConn(t *testing.T) {
 	log, buf := logging.NewTestLogger(t.Name())
 	defer test.ShowBufferOnFailure(t, buf)
 
-	mod := NewSecurityModule(log, defaultTestSecurityConfig())
+	mod := NewSecurityModule(log, defaultTestSecurityConfig(t, log, testInfoCacheParams{}))
 	respBytes, err := callRequestCreds(mod, t, log, &net.TCPConn{})
 
 	test.CmpErr(t, drpc.NewFailureWithMessage("connection is not a unix socket"), err)
@@ -167,7 +186,7 @@ func TestAgentSecurityModule_RequestCreds_NotConnected(t *testing.T) {
 	defer cleanup()
 	conn.Close() // can't get uid/gid from a closed connection
 
-	mod := NewSecurityModule(log, defaultTestSecurityConfig())
+	mod := NewSecurityModule(log, defaultTestSecurityConfig(t, log, testInfoCacheParams{}))
 	respBytes, err := callRequestCreds(mod, t, log, conn)
 
 	if err != nil {
@@ -189,6 +208,7 @@ func TestAgentSecurityModule_RequestCreds_BadConfig(t *testing.T) {
 	mod := NewSecurityModule(log, &securityConfig{
 		transport:   &security.TransportConfig{},
 		credentials: &security.CredentialConfig{},
+		infoCache:   defaultInfoCache(t, log, testInfoCacheParams{}),
 	})
 	respBytes, err := callRequestCreds(mod, t, log, conn)
 
@@ -207,8 +227,8 @@ func TestAgentSecurityModule_RequestCreds_BadUid(t *testing.T) {
 	conn, cleanup := setupTestUnixConn(t)
 	defer cleanup()
 
-	mod := NewSecurityModule(log, defaultTestSecurityConfig())
-	mod.signCredential = func(_ context.Context, _ *auth.CredentialRequest) (*auth.Credential, error) {
+	mod := NewSecurityModule(log, defaultTestSecurityConfig(t, log, testInfoCacheParams{}))
+	mod.signCredential = func(_ context.Context, _ logging.Logger, _ auth.CredentialRequest) (*auth.Credential, error) {
 		return nil, errors.New("LookupUserID")
 	}
 	respBytes, err := callRequestCreds(mod, t, log, conn)
@@ -217,7 +237,7 @@ func TestAgentSecurityModule_RequestCreds_BadUid(t *testing.T) {
 		t.Errorf("Expected no error, got %+v", err)
 	}
 
-	expectCredResp(t, respBytes, int32(daos.MiscError), false)
+	expectCredResp(t, respBytes, int32(daos.FailedSign), false)
 }
 
 type signCredentialResp struct {
@@ -226,17 +246,20 @@ type signCredentialResp struct {
 }
 
 func TestAgent_SecurityRPC_getCredential(t *testing.T) {
+	log, buf := logging.NewTestLogger(t.Name())
+	defer test.ShowBufferOnFailure(t, buf)
+
 	testCred := &auth.Credential{
 		Token:  &auth.Token{Flavor: auth.Flavor_AUTH_SYS, Data: []byte("test-token")},
 		Origin: "test-origin",
 	}
-	miscErrBytes, err := proto.Marshal(
+	failedSignErrBytes, err := proto.Marshal(
 		&auth.GetCredResp{
-			Status: int32(daos.MiscError),
+			Status: int32(daos.FailedSign),
 		},
 	)
 	if err != nil {
-		t.Fatalf("Couldn't marshal misc error: %v", err)
+		t.Fatalf("Couldn't marshal failed sign error: %v", err)
 	}
 	successBytes, err := proto.Marshal(
 		&auth.GetCredResp{
@@ -255,18 +278,18 @@ func TestAgent_SecurityRPC_getCredential(t *testing.T) {
 		expErr    error
 	}{
 		"lookup miss": {
-			secCfg: defaultTestSecurityConfig(),
+			secCfg: defaultTestSecurityConfig(t, log, testInfoCacheParams{}),
 			responses: []signCredentialResp{
 				{
 					cred: nil,
 					err:  user.UnknownUserIdError(unix.Getuid()),
 				},
 			},
-			expBytes: miscErrBytes,
+			expBytes: failedSignErrBytes,
 		},
 		"lookup OK": {
 			secCfg: func() *securityConfig {
-				cfg := defaultTestSecurityConfig()
+				cfg := defaultTestSecurityConfig(t, log, testInfoCacheParams{})
 				cfg.credentials.ClientUserMap = security.ClientUserMap{
 					uint32(unix.Getuid()): &security.MappedClientUser{
 						User: "test-user",
@@ -288,7 +311,7 @@ func TestAgent_SecurityRPC_getCredential(t *testing.T) {
 		},
 		"lookup OK, but retried request fails": {
 			secCfg: func() *securityConfig {
-				cfg := defaultTestSecurityConfig()
+				cfg := defaultTestSecurityConfig(t, log, testInfoCacheParams{})
 				cfg.credentials.ClientUserMap = security.ClientUserMap{
 					uint32(unix.Getuid()): &security.MappedClientUser{
 						User: "test-user",
@@ -306,7 +329,7 @@ func TestAgent_SecurityRPC_getCredential(t *testing.T) {
 					err:  errors.New("oops"),
 				},
 			},
-			expBytes: miscErrBytes,
+			expBytes: failedSignErrBytes,
 		},
 	} {
 		t.Run(name, func(t *testing.T) {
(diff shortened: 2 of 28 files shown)


JonahRosenblumHPE:
A [specification](https://gist.github.com/JonahRosenblumHPE/4a2bde936b652d21a96ed4bc5994dce4) we wrote about DAOS Agent authentication.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -52,41 +50,85 @@ type (
...
+			return nil, fmt.Errorf("found authentication method `%s`, which is not the expected length %d", authMethodString, auth.AuthTagSize)
+		}
+		method := (auth.AuthTag)([]byte(authMethodString))
+		_, found := auth.AuthRegister[method]
+		if !found {
+			return nil, fmt.Errorf("found authentication method `%s`, for which no implementation exists - check that `GetAuthName` is correct, and instance exists in the `CredentialRequests` structure", authMethodString)
+		}
+		validAuthMethods[method] = true
+	}
+	if len(cfg.credentials.ValidAuthMethods) == 0 {
```
tanabarr: So if ValidAuthMethods are specified in the config the default UDS method won't be automatically included and has to be specified within the config? Shouldn't we always provide UDS as a valid method regardless of what extra methods have been added?

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -52,41 +50,85 @@ type (
...
+		validAuthMethods[auth.CredentialRequestUnix{}.GetAuthName()] = true
+	}
+	return validAuthMethods, nil
+}
+
+func getAuthArgs(reqb []byte) (authArgs, error) {
+	var args authArgs
+	reqbSize := len(reqb)
+	if reqbSize == 0 {
+		args.tag = auth.CredentialRequestUnix{}.GetAuthName()
```
tanabarr: This seems to imply that UDS should always be available as an authentication option.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -52,41 +50,85 @@ type (
...
+	} else if reqbSize >= auth.AuthTagSize {
+		copy(args.tag[:], reqb)
+		args.reqBody = reqb[auth.AuthTagSize:]
```
tanabarr: How does the concatenation of the request body work?

src/control/security/auth/cred_am.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/pkg/errors"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/daos-stack/daos/src/control/drpc"
+	"github.com/daos-stack/daos/src/control/logging"
+	"github.com/daos-stack/daos/src/control/security"
+)
+
+type (
+	CredentialRequestAM struct {
```
tanabarr: This file seems to be missing comments to describe exported symbols.

src/control/security/auth/auth_sys.go
```diff
@@ -302,3 +78,59 @@ func AuthSysFromAuthToken(authToken *Token) (*Sys, error) {
...
+		CredReqKey() string
+	}
+)
+
+// CredentialRequests is a list of request types that can be used for authentication.
+// To add a new type of authentication simply implement the CredentialRequest interface and add
+// an instance to the CredentialRequests list.
+// The agent must be configured to allow an authentication method when it is initalized.
+// By default, only Unix authentication is enabled.
+var CredentialRequests = []CredentialRequestFactory{&CredentialRequestUnix{}, &CredentialRequestAM{}}
```
tanabarr: Comment "by default, only unix authentication is enabled" doesn't match the code. List also contains CredentialRequestAM.

src/control/security/auth/cred_unix.go
```diff
@@ -0,0 +1,349 @@
...
+	if IsNilish(req) {
+		return nil, errors.New("is nil")
+	}
+
+	cred, err := req.GetSignedCredentialInternal(ctx, req)
+	if err != nil {
+		if req.DomainInfo == nil {
+			return nil, err
+		}
+		if err := func() error {
```
tanabarr: nit: the style seems to introduce unnecessary nesting, I would prefer to add a `defer func() {log.Error(...)}` using the err variable defined in the function signature return types `(cred *Credential, err error)` to provide error-logging for each of the errors returned (which seems to be the purpose of the `func() error` syntax. Alternatively it would be fine to not supply a logger to the function and instead log the error in the caller of GetSignedCredential().

src/control/security/auth/auth_sys.go
```diff
@@ -302,3 +78,59 @@ func AuthSysFromAuthToken(authToken *Token) (*Sys, error) {
...
+		AllocCredentialRequest() CredentialRequest
+		// Returns a unique 4 character identifier (AuthTag) that refers to the CredentialRequest implementation.
+		GetAuthName() AuthTag
+	}
+
+	CredentialRequest interface {
+		// Using information stored in the agent's security config, the session/socket, request body, and the agent's signing key,
+		// initalize the state of the Credential Request so that it can sign a credential in the future. Return an error if
+		// a problem occurs during initializiation.
+		InitCredentialRequest(log logging.Logger, sec_cfg *security.CredentialConfig, session *drpc.Session, req_body []byte, key crypto.PrivateKey) error
```
tanabarr: Do we want to be passing a logger into credential methods, don't we want to just log the error in the caller. This might reduce the chance of logging information that might be useful to an attacker.

src/control/security/auth/auth_sys.go
```diff
@@ -302,3 +78,59 @@ func AuthSysFromAuthToken(authToken *Token) (*Sys, error) {
...
+		// Contact a source of authenticity (e.g. domain socket, access manager) using the initalized state of the CredentialRequest
+		// to construct a Credential object.
+		GetSignedCredential(log logging.Logger, ctx context.Context) (*Credential, error)
```
tanabarr: As above.

tanabarr (commented):
Looks like a great contribution thank you very much. The changes look good to me, I would like to see some unit testing including some basics on using the access manager. Some documentation on adding new authentication methods and a walk-through of using the access manager example would be good from a usability perspective. I think the original author of the code @kjacque should also take a look to see if I've missed anything.

src/control/cmd/daos_agent/security_rpc.go
JonahRosenblumHPE: I wanted this code to be backward compatible. "Authentication type" was never specified before this feature because the only option was UDS. By assuming that an empty body means the client wants to use UDS, this feature maintains compatibility with old clients. However, it is possible to disable UDS in the agent through the agent configuration file.

src/control/cmd/daos_agent/security_rpc.go
JonahRosenblumHPE: `So if ValidAuthMethods are specified in the config the default UDS method won't be automatically included and has to be specified within the config?`

Yes

`Shouldn't we always provide UDS as a valid method regardless of what extra methods have been added?`

Personally, I don't think so. If I create my own personal authentication method that I trust, and I don't want to consider "did I configure my Unix groups properly?", it would be nice to disable UDS. I am open to changing my opinion on this, but having more control over valid authentication methods for the agent seems better to me.

src/control/cmd/daos_agent/security_rpc.go
JonahRosenblumHPE: I'm not sure I follow the question, likely stemming from my inexperience with Golang. The purpose of this code is to parse the raw message into the `authTag`, which identifies the authentication method, and the body of the request into a struct.

I took an approach that is similar to `memcpy` in C, although I suspect there is a more "Golang-style" way of doing this.

If this does not answer your question, I apologize.

src/control/security/auth/auth_sys.go
JonahRosenblumHPE: `by default, only unix authentication is enabled`

I think this comment is misleading. `CredentialRequests` is a list containing every *implementation* of an authentication method.

[This method](https://github.com/JonahRosenblumHPE/daos_gen_auth/blob/master/src/control/cmd/daos_agent/security_rpc.go#L70) is where the agent determines which authentication is *enabled* and [this](https://github.com/JonahRosenblumHPE/daos_gen_auth/blob/master/src/control/cmd/daos_agent/security_rpc.go#L84) is where the "by default, only unix authentication is enabled" portion is relevant.

The reason for including this comment in the first place is to help developers, but I think they should instead refer to my [specification](https://gist.github.com/JonahRosenblumHPE/4a2bde936b652d21a96ed4bc5994dce4) for how to add new authentication and I should remove this portion of the comment. Do you agree?

src/control/security/auth/cred_unix.go
JonahRosenblumHPE: My design choice with the nesting in this function is related to how unit testing is done [here](https://github.com/JonahRosenblumHPE/daos_gen_auth/blob/master/src/control/cmd/daos_agent/security_rpc_test.go#L337). I'm not sure if your suggestion is compatible with this unit test, or if I should just rewrite the unit test.

src/control/security/auth/auth_sys.go
JonahRosenblumHPE: I don't have much experience with Golang logging. How can I refactor this to log the error in the caller - should I initialize a "default logger" in the caller body?

tanabarr: You already have a logger in the only call I can see: https://github.com/daos-stack/daos/pull/16788/files#diff-3aa160ba407c1ca20c5bc486aa45370dd185a0f0f1eb5eddbebd57293fe9e8a8R221 just log the error after the call if you want to (otherwise just return the error which should be sufficient in most cases, we don't tend to log an error and return it, normally just do one or the other and if returning error use `errors.Wrap()` et al. to add context to the returned error).

src/control/security/auth/cred_unix.go
tanabarr: Ok, I didn't realise the test was dependent on this paradigm. It's only a style related comment so can leave it as is, thanks.

src/control/security/auth/auth_sys.go
tanabarr: Yes that sounds good, thanks.

src/control/cmd/daos_agent/security_rpc.go
tanabarr: OK, that sounds okay as long as it's properly documented and understood.

src/control/cmd/daos_agent/security_rpc.go
tanabarr: Is there some unit test coverage that covers the cases where 1) empty body results in UDS (this line being executed) and  2) UDS is specified in the config auth types and so args.tag get set as UDS 3) UDS not specified in the config auth types therefore body is not ever empty and another type gets used? Basically it would be good to get decent unit test coverage on this method to ensure all existing authentication types work as expected.

src/control/cmd/daos_agent/security_rpc.go
tanabarr: I'm happy if it's got good unit test coverage to give us confidence in the implementation. You could probably use Marshal/Unmarshal for this kind of parsing.

src/control/cmd/daos_agent/security_rpc.go
JonahRosenblumHPE: This makes sense. Is `security_rpc_test.go` the appropriate location to add these unit tests?

src/control/cmd/daos_agent/security_rpc.go
JonahRosenblumHPE: I will look into using Marshal/Unmarshal and add unit tests.

src/control/security/auth/cred_unix.go
JonahRosenblumHPE: I didn't document this strange design choice in my code, which understandably raises eyebrows. I will add a short comment highlighting this for future developers.

src/include/daos/security.h
```diff
@@ -12,6 +12,18 @@
...
 #include <daos_prop.h>
 
+/** Constant that represents the size of an authentication method name in the DAOS agent. */
+#define AGENT_AUTH_METHOD_NAME_SIZE 4
+
+/** Constant that represents the name of the Unix authentication method. */
+#define AGENT_AUTH_METHOD_UNIX "unix"
+
+/** Constant that represents the name of the Unix authentication method. */
+#define AGENT_AUTH_METHOD_AM "acma"
```
kjacque: Looks like the comment wasn't updated. AM/acma == access manager? The name is a bit vague, but if that's what the project is called, no problem with the name.

src/security/cli_security.c
```diff
@@ -48,24 +48,56 @@ dc_sec_request_creds(d_iov_t *creds)
...
+
 static int
 request_credentials_via_drpc(Drpc__Response **response)
 {
 	Drpc__Call	*request;
 	struct drpc	*agent_socket;
 	int		rc;
+	uint8_t *token_buffer = NULL;
+	unsigned int token_buffer_length = 0;
+	char *delegation_token_path = getenv("TOKEN_PATH");
```
kjacque, line 136: This is something I'd expect the agent to know about (maybe part of the agent config?) and return to us in the credential payload.

Without knowing more about the Access Manager protocol, it is tricky to offer specific advice on how we should handle this and where. The agent is the secure component on the client side though. From the libdaos perspective, the agent hands us an appropriate credential, and we blindly send it along to the daos_engine. The daos_engine verifies the cred with the daos_server and stashes it.

src/security/cli_security.c
```diff
@@ -48,24 +48,56 @@ dc_sec_request_creds(d_iov_t *creds)
...
+	int use_access_manager = delegation_token_path != NULL && strcmp(delegation_token_path, "");
```
kjacque: The agent should return a structure based on the supported auth type. I don't think libdaos should care about details like this.

src/control/security/auth/auth_sys.go
```diff
@@ -66,228 +64,6 @@ func VerifyToken(key crypto.PublicKey, token *Token, sig []byte) error {
 	return errors.Wrap(err, "token verification Failed")
 }
 
-func sysNameToPrincipalName(name string) string {
```
kjacque: This file was intended to encapsulate the AUTH_SYS authentication token type (i.e. using unix usernames/groups). Rather than breaking this up, I think you should consider the flow for this new authentication type separately. If you're free to share any details on the protocol for the access manager, I'd be happy to collaborate.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -52,41 +50,85 @@ type (
...
+		tag     auth.AuthTag
+		reqBody []byte
 	}
+
+	AuthValidSet map[auth.AuthTag]bool
 )
 
 var _ cache.ExpirableItem = (*cachedCredential)(nil)
 
+func getValidAuthMethods(cfg *securityConfig) (AuthValidSet, error) {
```
kjacque: There is a dangling question here: Why would a single DAOS system want to depend on multiple auth types? Unless there's a compelling use case for allowing this, it seems unnecessarily complex, and maybe easier for a user to mess up. I think the auth type needs to be a single type defined in the server config, as a central source. The agent can get this information from the servers.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -146,77 +188,49 @@ func newCachedCredential(key string, cred *auth.Credential, lifetime time.Durati
...
-
-	return m.getCredential(ctx, session)
-}
-
-// getCredentials generates a signed user credential based on the data attached to
-// the Unix Domain Socket.
-func (m *SecurityModule) getCredential(ctx context.Context, session *drpc.Session) ([]byte, error) {
-	if session == nil {
-		return nil, drpc.NewFailureWithMessage("session is nil")
+func (m *SecurityModule) HandleCall(ctx context.Context, session *drpc.Session, method drpc.Method, reqb []byte) ([]byte, error) {
```
kjacque, line 169: The additions in this function seem like they really belong in the GetCredential logic, rather than in the overall SecurityModule.

src/control/cmd/daos_agent/security_rpc_test.go
```diff
@@ -317,19 +317,29 @@ func TestAgent_SecurityRPC_getCredential(t *testing.T) {
...
+								idx++
+							}
+						}()
+						t.Logf("returning response %d: %+v", idx, tc.responses[idx])
+						return tc.responses[idx].cred, tc.responses[idx].err
+					}
+				}()
+				r.GetSignedCredentialInternal = testing_fn
+				return r.GetSignedCredential(log, ctx)
+			}
```
kjacque, line 362: This feels a bit over-complex for a mock function. It seems like testing_fn is what you're really trying to inject here. Worth re-thinking the design approach here I think. Best to design the code to be easy to test.

src/control/security/auth/cred_unix.go
kjacque: Seconding the initial comment. I don't see why you can't define the closure as a named function (either within this function or outside it). You can pass `req` into it.

kjacque (changes requested):
Thanks for submitting this PR. It's a good start, but I think we should chat in-depth about the overall design. Right now it seems like the new AM authentication method is a bit jumbled up with the AUTH_SYS (unix) method, and changes are a bit less compartmentalized than I'd like. The initial design was intended to make auth types modular, down to the messages passed via dRPC (see `src/proto/security/auth.proto`). We also need to be very careful about how tokens and credentials are passed around and who they are accessible to.

If you have any public (or private, I'm at HPE too) specs for the sample access manager, I'd be happy to help with the high level design of this addition. We definitely need to ensure appropriate testing for all the new logic, too.

src/control/cmd/daos_agent/security_rpc.go
tanabarr: Yes it is but I think it might be best to arrange a sync-up with @kjacque before writing tests just in case any design changes need to be made.

src/proto/mgmt/svc.proto
```diff
@@ -8,102 +8,111 @@
...
+	uint32          idx             = 8;  // Instance index on server node.
+	uint64          incarnation     = 9;  // rank incarnation
+	repeated string secondary_uris  = 10; // URIs for any secondary providers
 	repeated uint32 secondary_nctxs = 11; // CaRT context count for each secondary provider
-	bool check_mode = 12; 		// rank started in check mode
+	bool            check_mode      = 12; // rank started in check mode
 	bool            replace         = 13; // Rank's engine instance metadata to be replaced
 }
 
-message JoinResp {
```
jonahrosenblum: Revert changes to this file - all whitespace.

src/proto/security/auth.proto
```diff
@@ -7,52 +7,73 @@
...
-	Token verifier = 2; // to verify integrity of the token
-	string origin = 3; // the agent that created this credential
+message Credential
+{
+	Token  token    = 1; // authentication token
+	Token  verifier = 2; // to verify integrity of the token
+	string origin   = 3; // the agent that created this credential
+}
+
+message AuthArgs
```
jonahrosenblum: This is `GetCredReq`.

src/proto/security/auth.proto
```diff
@@ -7,52 +7,73 @@
...
 
-option go_package = "github.com/daos-stack/daos/src/control/security/auth;auth";
+option  go_package = "github.com/daos-stack/daos/src/control/security/auth;auth";
 
 // Types of authentication token
 enum Flavor {
 	AUTH_NONE = 0;
-	AUTH_SYS = 1;
+	AUTH_SYS  = 1;
+	AUTH_AM   = 2;
```
jonahrosenblum: AM is vague; be more explicit (e.g. AccMan). Add comments along with the proto, it gets exported into generated code.

jonahrosenblum: Add detail for AUTH_SYS/AUTH_UNIX as well.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -10,11 +10,14 @@ package main
 import (
 	"context"
 	"fmt"
-	"net"
-	"os/user"
+<<<<<<< HEAD
+=======
```
jonahrosenblum: Fix merge issue!

src/security/cli_security.c
jonahrosenblum: The client should ping the agent to learn what authentication method(s) it should use. If multiple are provided, it will default to the first option.

Separate AUTH_SYS vs AUTH_ACCMAN into different functions.

Switch statement for flavor, make it easy to add new authentication.

src/control/server/init/setup_spdk.sh
```diff
@@ -58,7 +58,8 @@ thisscriptpath="$(dirname "$(readlink -f "$0")")"
 
 echo "start of script: $thisscriptpath/$thisscriptname"
```
jonahrosenblum, line 59: Revert this change.

src/control/server/mgmt_svc.go
```diff
@@ -83,9 +83,10 @@ type mgmtSvc struct {
 	serialReqs        batchReqChan
 	groupUpdateReqs   chan bool
 	lastMapVer        uint32
+	validAuthFlavors  []uint32
 }
 
-func newMgmtSvc(h *EngineHarness, m *system.Membership, s *raft.Database, c control.UnaryInvoker, p *events.PubSub) *mgmtSvc {
+func newMgmtSvc(h *EngineHarness, m *system.Membership, s *raft.Database, c control.UnaryInvoker, p *events.PubSub, v []uint32) *mgmtSvc {
```
jonahrosenblum: Change v to a (authentication!)

src/control/security/auth/security/auth.pb.go
```diff
@@ -0,0 +1,633 @@
...
+// (C) Copyright 2018-2021 Intel Corporation.
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+//
+
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.34.1
+// 	protoc        v3.21.12
+// source: security/auth.proto
```
jonahrosenblum: Remove this, added accidentally.

src/control/server/mgmt_svc.go
jonahrosenblum: Look into importing auth in the server, and then making this an auth.Flavor type.

src/control/security/auth/auth_sys.go
```diff
@@ -66,239 +70,83 @@ func VerifyToken(key crypto.PublicKey, token *Token, sig []byte) error {
...
-
-func (r *CredentialRequest) groupPrincipals() ([]string, error) {
-	if r.getGroupNames == nil {
-		return nil, errors.New("groupNames function not set")
-	}
-
-	groupNames, err := r.getGroupNames(r)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to get group names")
+func ParseValidAuthFlavors(authStrings []string) ([]uint32, error) {
```
jonahrosenblum: Rename file to just "auth.go", auth_sys is one flavor of auth not the entire thing anymore!

jonahrosenblum: Make this parsing case-insensitive.

jonahrosenblum: Another option: let them start their string with "auth" if they want.

strings.StartsWith("AUTH_") -> just pass the raw string in. Still handle case sensitivity for them.

jonahrosenblum: Also, update the example config to show how to specify auth type.

jonahrosenblum: For public methods, include a documentation comment explaining what the method does!

src/control/security/auth/auth_sys.go
```diff
@@ -66,239 +70,83 @@ func VerifyToken(key crypto.PublicKey, token *Token, sig []byte) error {
...
 }
 
-func (r *CredentialRequest) group() (*user.Group, error) {
-	if r.getGroup == nil {
-		return nil, errors.New("group lookup function not set")
-	}
-	return r.getGroup(strconv.Itoa(int(r.DomainInfo.Gid())))
+func CredentialRequestGetSigned(ctx context.Context, log logging.Logger, req CredentialRequest) (*Credential, error) {
+	return req.GetSignedCredential(log, ctx)
 }
```
jonahrosenblum, line 131: Probably move this to the agent since this is a workaround for an agent-specific issue.

src/control/server/security_rpc.go
```diff
@@ -64,7 +72,7 @@ func (m *SecurityModule) processValidateCredentials(body []byte) ([]byte, error)
 	}
 
 	// Check our verifier
-	err = auth.VerifyToken(key, cred.GetToken(), cred.GetVerifier().GetData())
+	err = auth.VerifyToken(key, cred.GetToken(), cred.GetVerifier().GetData(), m.validAuthFlavors)
```
jonahrosenblum: Could remove the parameter from VerifyToken and do the check of "is this flavor valid" in the body of this function.

src/control/server/server.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/daos-stack/daos/src/control/lib/hardware/defaults/topology"
 	"github.com/daos-stack/daos/src/control/logging"
 	"github.com/daos-stack/daos/src/control/security"
+	"github.com/daos-stack/daos/src/control/security/auth"
```
jonahrosenblum, line 36: You're including auth here, so why can't you use auth.Flavor?!

src/control/server/server.go
```diff
@@ -173,13 +176,19 @@ func newServer(log logging.Logger, cfg *config.Server, faultDomain *system.Fault
 
 	harness := NewEngineHarness(log).WithFaultDomain(faultDomain)
 
+	validAuthFlavors, err := auth.ParseValidAuthFlavors(cfg.TransportConfig.AuthenticationConfig.ValidAuth)
```
jonahrosenblum: Parse this once, find a way to pass this to the security service AND the management service.

src/include/daos/security.h
```diff
@@ -12,6 +12,9 @@
 #include <daos_types.h>
 #include <daos_prop.h>
 
+/** Constant that represents the upper bound size of a string encoded delegation credential token for the access manager. */
+#define MAX_DELEGATION_TOKEN_SIZE 10000
```
jonahrosenblum: Move to the function that the client uses to package an "access manager" request to the agent. It is not used anywhere else.

src/control/security/config.go
```diff
@@ -94,15 +94,36 @@ func (cm ClientUserMap) Lookup(uid uint32) *MappedClientUser {
 // CredentialConfig contains configuration details for managing user
 // credentials.
 type CredentialConfig struct {
-	CacheExpiration time.Duration `yaml:"cache_expiration,omitempty"`
-	ClientUserMap   ClientUserMap `yaml:"client_user_map,omitempty"`
+	CacheExpiration  time.Duration       `yaml:"cache_expiration,omitempty"`
+	ClientUserMap    ClientUserMap       `yaml:"client_user_map,omitempty"`
+	ValidAuthMethods []string            `yaml:"valid_auth_methods,omitempty"`
+	AMConfig         AccessManagerConfig `yaml:"access_manager_config,omitempty"`
```
jonahrosenblum, line 100: These should not be in the agent!

src/control/security/config.go
```diff
@@ -94,15 +94,36 @@ func (cm ClientUserMap) Lookup(uid uint32) *MappedClientUser {
...
+	BaseURL  string `yaml:"base_url,omitempty"`
 }
 
 // TransportConfig contains all the information on whether or not to use
 // certificates and their location if their use is specified.
 type TransportConfig struct {
 	AllowInsecure     bool `yaml:"allow_insecure"`
 	CertificateConfig `yaml:",inline"`
+	// authentication
+	AuthenticationConfig *AuthenticationConfig `yaml:"auth_config"`
```
jonahrosenblum: This should be directly part of the server config file!

src/control/security/config.go
```diff
@@ -94,15 +94,36 @@ func (cm ClientUserMap) Lookup(uid uint32) *MappedClientUser {
...
+}
+
+// AuthenticationConfig contains configuation details for valid authentication
+type AuthenticationConfig struct {
+	ValidAuth []string `yaml:"valid_auth"`
+}
+
+func DefaultAuthenticationConfig() *AuthenticationConfig {
+	return &AuthenticationConfig{
+		ValidAuth: []string{"SYS"},
```
jonahrosenblum: Use the actual default string "AUTH_SYS" generated by the protobuf.

jonahrosenblum: And move to the server file.

src/control/security/auth/auth_accman.go
jonahrosenblum, line 1: rename this to auth_accman.go, and the other to auth_sys.go. The current auth_sys.go is now the "generic auth" file so that should just be auth.go.

src/control/security/auth/cred_unix.go
```diff
@@ -0,0 +1,349 @@
...
+	GetSignedCredentialInternalFn func(ctx context.Context, req *CredentialRequestUnix) (*Credential, error)
+
+	getHostnameFn   func() (string, error)
+	getUserFn       func(string) (*user.User, error)
+	getGroupFn      func(string) (*user.Group, error)
+	getGroupIdsFn   func(*CredentialRequestUnix) ([]string, error)
+	getGroupNamesFn func(*CredentialRequestUnix) ([]string, error)
+
+	// CredentialRequest defines the request parameters for GetSignedCredential.
+	CredentialRequestUnix struct {
```
jonahrosenblum: Rename: AuthSysCredentialRequest. (Obviously change the acc man one too)

src/control/security/auth/auth_sys.go
```diff
@@ -66,239 +70,83 @@ func VerifyToken(key crypto.PublicKey, token *Token, sig []byte) error {
...
+		AllocCredentialRequest() CredentialRequest
+		// Returns a unique 4 character identifier (AuthTag) that refers to the CredentialRequest implementation.
+		GetAuthFlavor() Flavor
+	}
+
+	CredentialRequest interface {
+		// Using information stored in the agent's security config, the session/socket, request body, and the agent's signing key,
+		// initalize the state of the Credential Request so that it can sign a credential in the future. Return an error if
+		// a problem occurs during initializiation.
+		InitCredentialRequest(log logging.Logger, sec_cfg *security.CredentialConfig, session *drpc.Session, req_body []byte, key crypto.PrivateKey) error
```
jonahrosenblum: Rename this to just "init"

src/control/security/auth/auth_sys.go
```diff
@@ -66,239 +70,83 @@ func VerifyToken(key crypto.PublicKey, token *Token, sig []byte) error {
...
-
-	verifier, err := VerifierFromToken(req.SigningKey, &token)
-	if err != nil {
-		return nil, errors.WithMessage(err, "Unable to generate verifier")
+type (
+	AuthMap map[Flavor]CredentialRequestFactory
+
+	CredentialRequestFactory interface {
+		// Allocate, and return, an instance of the type that implements CredentialRequest interface.
+		AllocCredentialRequest() CredentialRequest
```
jonahrosenblum: See if factory can do the initialization! Seems redundant to do alloc then init.

src/control/security/auth/auth_sys.go
```diff
@@ -66,239 +70,83 @@ func VerifyToken(key crypto.PublicKey, token *Token, sig []byte) error {
...
+		// Returns a unique 4 character identifier (AuthTag) that refers to the CredentialRequest implementation.
```
jonahrosenblum: This comment is outdated, update.

src/control/security/auth/auth_sys.go
```diff
@@ -66,239 +70,83 @@ func VerifyToken(key crypto.PublicKey, token *Token, sig []byte) error {
...
+		// initalize the state of the Credential Request so that it can sign a credential in the future. Return an error if
+		// a problem occurs during initializiation.
+		InitCredentialRequest(log logging.Logger, sec_cfg *security.CredentialConfig, session *drpc.Session, req_body []byte, key crypto.PrivateKey) error
+		// Contact a source of authenticity (e.g. domain socket, access manager) using the initalized state of the CredentialRequest
+		// to construct a Credential object.
+		GetSignedCredential(log logging.Logger, ctx context.Context) (*Credential, error)
+		// Returns a key, as a string, representing a unique identifer specific to the request. This key is used by the cache
+		// to remember credentials. It is vital for security that this identifer cannot be forged or easily guessed by the client - 
+		// otherwise cached credentials can be "stolen".
+		CredReqKey() string
```
jonahrosenblum: Rename to "GetKey"

src/control/security/auth/auth_sys.go
```diff
@@ -66,239 +70,83 @@ func VerifyToken(key crypto.PublicKey, token *Token, sig []byte) error {
...
-	}
-
-	sysToken := &Sys{}
-	err := proto.Unmarshal(authToken.GetData(), sysToken)
-	if err != nil {
-		return nil, errors.Wrapf(err, "unmarshaling %s", authToken.GetFlavor())
-	}
-	return sysToken, nil
-}
+var CredentialRequests = []CredentialRequestFactory{&CredentialRequestUnix{}, &CredentialRequestAM{}}
```
jonahrosenblum: Revisit why I implemented it this way. Can this be simplified? Maybe just a function with a mapping/switch statement between flavors and factories.

jonahrosenblum: Split the factory and the request into two separate things. The requests should not double as factories; the factories should exist only to generate requests!

src/control/security/auth/auth_accman.go
```diff
@@ -0,0 +1,213 @@
...
+	for i := range groups {
+		var _, role, err = seperate(authInfo.Roles[i])
+		if err != nil {
+			return nil, err
+		}
+		groups[i] = role
+	}
+
+	// Craft AuthToken
+	sys := Sys{
```
jonahrosenblum, line 188: Add parsing on the server side to determine what to do with non-Sys tokens.

jonahrosenblum: Add new token for AccMan!

src/control/security/auth/cred_am.go
```diff
@@ -0,0 +1,213 @@
...
+		split_url[1] += "@am"
+		return split_url[0], split_url[1], nil
+	}
+
+	authInfo, err := req.validateAndParseDelegationCredential()
+	if err != nil {
+		return nil, err
+	}
+
+	machine_name, identity, err := seperate(authInfo.Identity)
```
jonahrosenblum: Use camel case for names.

src/control/security/auth/auth_accman.go
jonahrosenblum: In token definition, call things by their name in the AM (i.e. "roles" not "groups")

src/control/lib/control/network.go
```diff
@@ -241,6 +241,7 @@ type (
 		ClientNetHint           ClientNetworkHint     `json:"client_net_hint"`
 		AlternateClientNetHints []ClientNetworkHint   `json:"secondary_client_net_hints"`
 		BuildInfo               BuildInfo             `json:"build_info"`
+		ValidAuthFlavors		[]uint32			  `json:"valid_auth_flavors"`
```
jonahrosenblum: Try to change to auth.Flavor

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -48,46 +51,69 @@ type (
...
 	// SecurityModule is the security drpc module struct
 	SecurityModule struct {
-		log            logging.Logger
-		signCredential credSignerFn
-		credCache      *credentialCache
+		log              logging.Logger
+		signCredential   credSignerFn
+		credCache        *credentialCache
+		config           *securityConfig
+		validAuthFlavors []auth.Flavor
```
jonahrosenblum: This is in the info cache... why am I reading it here?

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -48,46 +51,69 @@ type (
...
+		infoCache        *InfoCache
+	}
 
-		config *securityConfig
+	authArgs struct {
```
jonahrosenblum: This should be removed.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -48,46 +51,69 @@ type (
...
+		tag     auth.AuthTag
+		reqBody []byte
 	}
+
+	AuthValidSet map[auth.AuthTag]bool
 )
 
 var _ cache.ExpirableItem = (*cachedCredential)(nil)
 
+func getAuthArgs(reqb []byte) (*auth.AuthArgs, error) {
```
jonahrosenblum: The caller should pass the flavor I am using, I can change the client logic to ping the agent and ask which flavor to use.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -48,46 +51,69 @@ type (
...
+		}
+	}
+	fmt.Println(args)
+
+	return args, nil
+}
+
 // NewSecurityModule creates a new module with the given initialized TransportConfig.
-func NewSecurityModule(log logging.Logger, cfg *securityConfig) *SecurityModule {
+func NewSecurityModule(log logging.Logger, cfg *securityConfig) (*SecurityModule, error) {
```
jonahrosenblum: This can no longer return errors.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -146,77 +172,79 @@ func newCachedCredential(key string, cred *auth.Credential, lifetime time.Durati
...
-		return nil, drpc.UnknownMethodFailure()
+func (m *SecurityModule) retrieveAuthFromServer(ctx context.Context) error {
+	resp, err := m.infoCache.GetAttachInfo(ctx, m.config.sys)
+	if err != nil {
+		return errors.Wrap(err, "failed to get attach info")
 	}
 
-	return m.getCredential(ctx, session)
-}
+	m.validAuthFlavors = make([]auth.Flavor, len(resp.ValidAuthFlavors))
```
jonahrosenblum: Should just look in the cache instead of using a second member variable.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -146,77 +172,79 @@ func newCachedCredential(key string, cred *auth.Credential, lifetime time.Durati
...
-	uConn, ok := session.Conn.(*net.UnixConn)
-	if !ok {
-		return nil, drpc.NewFailureWithMessage("connection is not a unix socket")
+	return nil
+}
+
+// getCredentials generates a signed user credential based on the authentication method requested.
+func (m *SecurityModule) getCredential(ctx context.Context, session *drpc.Session, args *auth.AuthArgs, req auth.CredentialRequest) ([]byte, error) {
+	if len(m.validAuthFlavors) == 0 {
+		err := m.retrieveAuthFromServer(ctx)
```
jonahrosenblum: Just ask the cache!

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -146,77 +172,79 @@ func newCachedCredential(key string, cred *auth.Credential, lifetime time.Durati
...
-			cred, err = m.signCredential(ctx, req)
-			if err != nil {
-				return err
-			}
-
-			return nil
-		}(); err != nil {
-			m.log.Errorf("%s: failed to get user credential: %s", info, err)
-			return m.credRespWithStatus(daos.MiscError)
+		if errors.Is(err, daos.MiscError) {
```
jonahrosenblum, line 222: Add new DAOS error code for this specific situation.

jonahrosenblum: include/daos_error.h
and
src/control/lib/daos/

and also documentation where needed.

src/control/cmd/daos_agent/security_rpc.go
```diff
@@ -228,6 +256,31 @@ func (m *SecurityModule) credRespWithStatus(status daos.Status) ([]byte, error)
 	return drpc.Marshal(resp)
 }
 
+func (m *SecurityModule) getValidAuthFlavors(ctx context.Context) ([]byte, error) {
+	if len(m.validAuthFlavors) == 0 {
```
jonahrosenblum: Get it from the cache!

src/control/cmd/daos_agent/mgmt_rpc.go
```diff
@@ -49,6 +49,8 @@ type mgmtModule struct {
 
 	numaGetter  hardware.ProcessNUMAProvider
 	providerIdx uint
+
+	validAuthFlavors []uint32
```
jonahrosenblum: Invesigate if this is necesary!

src/control/security/auth/auth_accman.go
JonahRosenblumHPE: My conjecture is that every auth flavor will implement a very similar (if not identical) token structure. I propose that there ought to be a single generic token, and each auth flavor is responsible for implementing `GetSignedCredential` in a way that conforms to the generic token. I am holding off on making this change until we agree on this design choice.

src/proto/mgmt/svc.proto
```diff
@@ -114,52 +123,60 @@ message BuildInfo
...
+	repeated uint32           ms_ranks            = 3; // Ranks local to MS replicas
+	ClientNetHint             client_net_hint     = 4; // Primary provider hint
+	uint64                    data_version        = 5; // Version of the system database.
+	string                    sys                 = 6; // Name of the DAOS system
+	repeated RankUri          secondary_rank_uris = 7; // Rank URIs for additional providers
+	repeated ClientNetHint    secondary_client_net_hints = 8; // Hints for additional providers
+	BuildInfo                 build_info = 9; // Structured server build information
+	repeated FabricInterfaces numa_fabric_interfaces =
+	    10; // Usable fabric interfaces by NUMA node (populated by agent)
+	repeated uint32 valid_auth_flavors = 11; // Authentication flavors allowed by the server
```
JonahRosenblumHPE, line 148: I spent hours trying to get this to be an auth.Flavor. I cannot for the life of me figure out the SConscript build system. I defer to the maintainers - it would be unproductive for me to try any longer to make this change myself.

src/proto/mgmt/svc.proto
kjacque: Looks like you still need to revert these whitespace changes. I think you can make clang-format ignore *.proto files, if that's what's catching you here.

src/proto/mgmt/svc.proto
kjacque, line 148: A bit hard to read this file's diff due to all the whitespace changes. But this is the only change to this file, right?

src/security/cli_security.c
```diff
@@ -48,16 +65,49 @@ dc_sec_request_creds(d_iov_t *creds)
...
+request_flavor_via_drpc(Drpc__Response **response)
 {
 	Drpc__Call	*request;
 	struct drpc	*agent_socket;
 	int		rc;
 
 	if (dc_agent_sockpath == NULL) {
-		D_ERROR("DAOS Socket Path is Uninitialized\n");
-		return -DER_UNINIT;
+		dc_agent_sockpath = DEFAULT_DAOS_AGENT_DRPC_SOCK;
```
kjacque, line 110: `dc_agent_sockpath` should be initialized by the time we hit this point. Please revert this change.

src/include/daos_errno.h
```diff
@@ -200,6 +201,8 @@ extern "C" {
 	ACTION(DER_CONTROL_INCOMPAT, One or more control plane components are incompatible)        \
 	/** No service available */                                                                \
 	ACTION(DER_NO_SERVICE, No service available)                                               \
+	/** Faild to sign credential */                                                            \
+	ACTION(DER_FAILED_SIGN, Faild to sign credential)                                          \
```
kjacque, line 205: Add any new errnos to the end of the list - otherwise it shifts the numbering.

src/control/server/server.go
```diff
@@ -509,14 +512,13 @@ func (srv *server) start(ctx context.Context) error {
...
-		transportConfig: srv.cfg.TransportConfig,
-		sysdb:           srv.sysdb,
-		events:          srv.pubSub,
-		client:          srv.controlClient,
-		msReplicas:      srv.cfg.MgmtSvcReplicas,
+		log:     srv.log,
+		sockDir: srv.cfg.SocketDir,
+		engines: srv.harness.Instances(),
+		tc:      srv.cfg.TransportConfig,
+		vaf:     srv.validAuthFlavors,
```
kjacque, line 519: `vaf` is a little too brief as a name at this scope. `validAuthFlavors` or `authFlavors` would work.

src/control/server/drpc.go
```diff
@@ -67,14 +67,13 @@ func checkSocketDir(sockDir string) error {
...
-	transportConfig *security.TransportConfig
-	sysdb           *raft.Database
-	events          *events.PubSub
-	client          *control.Client
-	msReplicas      []string
+	log     logging.Logger
+	sockDir string
+	engines []Engine
+	tc      *security.TransportConfig
+	vaf     []auth.Flavor
```
kjacque, line 74: `vaf` is a little too vague IMO. How about `authFlavors`?

src/control/server/config/server.go
```diff
@@ -53,6 +54,17 @@ type deprecatedParams struct {
 	EnableHotplug *bool    `yaml:"enable_hotplug,omitempty"` // deprecated in 2.8
 }
 
+// AuthenticationConfig contains configuation details for valid authentication
+type AuthenticationConfig struct {
+	ValidAuth []string `yaml:"valid_auth"`
```
kjacque, line 59: After thinking on it for all this time, I'm still reluctant to support multiple auth types in a DAOS deployment.

src/control/security/config.go
kjacque: Agreed.

src/control/security/config.go
```diff
@@ -94,8 +94,16 @@ func (cm ClientUserMap) Lookup(uid uint32) *MappedClientUser {
...
+	CacheExpiration  time.Duration       `yaml:"cache_expiration,omitempty"`
+	ClientUserMap    ClientUserMap       `yaml:"client_user_map,omitempty"`
+	ValidAuthMethods []string            `yaml:"valid_auth_methods,omitempty"`
+	AMConfig         AccessManagerConfig `yaml:"access_manager_config,omitempty"`
+}
+
+// AccessManagerConfig contains configuration details for managing access manager
+type AccessManagerConfig struct {
+	CallerID string `yaml:"caller_id,omitempty"`
+	BaseURL  string `yaml:"base_url,omitempty"`
```
kjacque, line 106: I'm still inclined to think this belongs on the server side config. Maybe we can abstract how it's returned in GetAttachInfo so the client can see the auth type and some kind of "auth config" message containing flavor-specific definitions like this.

src/control/security/auth/auth_accman.go
kjacque: I'm inclined to agree that the identity/groups paradigm is probably what all auth flavors will adhere to. I guess the question is how to handle the auth "flavors" in that context. We have a few options:

- Each "flavor" has a unique payload, but identity/groups are part of that payload, maybe as a member struct. I think it would be okay to do this, since control plane components are guaranteed to be at the same version. I suspect this would be more flexible and cleaner in the long term, as more auth flavor types are added.
- The payload is the same for all flavors, and "flavor" describes only how it was acquired. This may be the easiest adaptation.

kjacque (changes requested):
I didn't get through everything, but left a few comments. Let's meet up next week and discuss what's left to do.

Do you think it would make sense to break this into multiple smaller PRs? I found myself really wishing for the GetAuthFlavor stuff to be in a different patch, just for ease of reading and thinking about. Sometimes it can also help to write things in a more self-contained way.

PR #16793, opened 2025-09-01
DAOS-16796 container: server handle cleanup

NiuYawei:
Server handle cleanup to avoid potential deadlock & other issues.

- Never try to fetch server handle through IV_CONT_CAPA.
- Never try to lookup server handle in normal handle DB.
- Don't call ds_cont_tgt_open/close() for IV_POOL_HDL to avoid race.
- Use same ULT to periodically report EC agg epoch, fetch pool connection handles, server open handles.
- Don't put server handle in container handle hash, otherwise, the handle will be leaked on container stop/destroy.
- I/O request handler won't go through normal handle code path when the server handle is used. If the server handle isn't propagated to server, -DER_STALE will be returned for retry.
- Introduce ds_cont_child::sc_open_mutex to serialize cont open in the ds_cont_local_open().
- Fix error cleanup defect in ds_cont_local_open() which could call dtx_cont_close() mistakenly.

### Steps for the author:


daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16793/4/execution/node/1268/log

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16793/6/execution/node/1183/log

NiuYawei:
ping reviewers. @Nasf-Fan , @liw , @liuxuezhao , could you take a look? Thanks!

src/pool/srv_pool.c
```diff
@@ -2409,21 +2409,19 @@ pool_svc_step_up_cb(struct ds_rsvc *rsvc)
...
 		if (!uuid_is_null(svc->ps_pool->sp_srv_cont_hdl)) {
 			uuid_copy(srv_pool_hdl, svc->ps_pool->sp_srv_pool_hdl);
 			uuid_copy(srv_cont_hdl, svc->ps_pool->sp_srv_cont_hdl);
 		} else {
 			uuid_generate(srv_pool_hdl);
 			uuid_generate(srv_cont_hdl);
-			/* Only copy server handle to make is_pool_from_srv() check correctly, and
-			 * container server handle will not be copied here, otherwise
-			 * ds_pool_iv_refresh_hdl will not open the server container handle.
-			 */
```
liw, line 2529: Excellent!

src/include/daos_srv/pool.h
```diff
@@ -179,6 +172,7 @@ struct ds_pool_child {
 	struct sched_request	*spc_scrubbing_req; /* Track scrubbing ULT*/
 	struct sched_request    *spc_chkpt_req;     /* Track checkpointing ULT*/
 	d_list_t		spc_cont_list;
+	d_list_t                 spc_hdl_list; /* Single server cont handle */
```
liw: [Nit] Might be clearer to use a name that indicates srv cont hdl.

liw (dismissed):

src/container/srv_target.c
```diff
@@ -1196,7 +1206,9 @@ ds_cont_hdl_get(struct ds_cont_hdl *hdl)
 {
 	struct d_hash_table *hash = &dsm_tls_get()->dt_cont_hdl_hash;
 
-	cont_hdl_get_internal(hash, hdl);
+	/* It's global srv cont handle when the sch_cont == NULL */
+	if (hdl->sch_cont != NULL)
+		cont_hdl_get_internal(hash, hdl);
```
Nasf-Fan, line 1212: Why not hold reference for the global srv cont handle? Just because it is not in hash table? If not hold the reference, why related sponsor need to call `ds_cont_hdl_get()` against the global srv cont handle?

src/pool/srv_iv.c
```diff
@@ -814,24 +814,26 @@ pool_iv_ent_fetch(struct ds_iv_entry *entry, struct ds_iv_key *key,
...
+		/*
+		 * The potential handle change can only happen on the old pool (before 2.8),
+		 * following open call could be removed once 2.6 pool support is dropped
+		 */
+		D_WARN(DF_UUID ": Server cont hdl changed from " DF_UUID " to " DF_UUID "\n",
+		       DP_UUID(pool->sp_uuid), DP_UUID(pool->sp_srv_cont_hdl),
+		       DP_UUID(pih->pih_cont_hdl));
+
+		if (!uuid_is_null(pih->pih_cont_hdl))
+			rc = ds_pool_srv_open(pool->sp_uuid, pih->pih_cont_hdl);
```
Nasf-Fan, line 805: Do we need to close the old `sp_srv_cont_hdl` before open `pih_cont_hdl`?
If `ds_pool_srv_open` got failure, still need to subsequent `uuid_copy`?

src/pool/srv_target.c
```diff
@@ -1094,113 +1090,171 @@ ds_pool_put(struct ds_pool *pool)
...
+	struct cont_srv_open_arg arg = {0};
+
+	D_ASSERT(!uuid_is_null(cont_hdl_uuid));
+	uuid_copy(arg.pool_uuid, pool_uuid);
+	uuid_copy(arg.cont_hdl_uuid, cont_hdl_uuid);
+	arg.flags     = 0;
+	arg.sec_capas = ds_sec_get_rebuild_cont_capabilities();
+
+	return ds_pool_task_collective(pool_uuid,
+				       PO_COMP_ST_NEW | PO_COMP_ST_DOWN | PO_COMP_ST_DOWNOUT,
```
Nasf-Fan, line 1116: There seems race during reintegrating target which status maybe still `DOWN` or `DOWNOUT` before pool map being refreshed. If server handle is not opened for the target to be re-integrated in time, will that affect the subsequent reintegration? I am not sure, need @liuxuezhao for further confirmation.

src/pool/srv_target.c
```diff
@@ -1094,113 +1090,171 @@ ds_pool_put(struct ds_pool *pool)
...
+			}
+		}
 
-out:
-	D_INFO(DF_UUID": signaling done\n", DP_UUID(pool->sp_uuid));
-	ABT_mutex_lock(pool->sp_mutex);
-	ABT_cond_signal(pool->sp_fetch_hdls_done_cond);
-	ABT_mutex_unlock(pool->sp_mutex);
+		if (eph_report_exiting(pool))
+			break;
```
Nasf-Fan: Move the check `eph_report_exiting` inside `if (!conn_hdl_fetched)` section to avoid unnecessary check for each `while` cycle.

src/pool/srv_target.c
```diff
@@ -1094,113 +1090,171 @@ ds_pool_put(struct ds_pool *pool)
...
+				}
+			}
+		}
 
-static void
-tgt_track_eph_query_ult(void *data)
-{
-	ds_cont_track_eph_query_ult(data);
+		if (eph_report_exiting(pool))
+			break;
```
Nasf-Fan: ditto.

src/pool/srv_target.c
```diff
@@ -1094,113 +1090,171 @@ ds_pool_put(struct ds_pool *pool)
...
 
 static int
-ds_pool_start_track_eph_query_ult(struct ds_pool *pool)
+start_eph_report_ult(struct ds_pool *pool)
 {
 	struct sched_req_attr	attr;
 	uuid_t			anonym_uuid;
 
-	if (unlikely(ec_agg_disabled))
-		return 0;
```
Nasf-Fan, line 1153: Because your patch combines the fetching connection handle & `srv_cont_hdl` together with `agg_eph_report` logic together, then even if `ec_agg_disabled`, we still have to start `eph_report_ult` ult, that seems some confused. Maybe we can add some check inside `eph_report_ult` or elsewhere?

src/container/srv_target.c
NiuYawei: The global srv cont handle is per pool and immutable, so it's not necessary to hold reference for it.
In current implementation, the caller doesn't know if a handle is global handle or normal handle and uses same set of API for the container handle.

src/pool/srv_iv.c
NiuYawei: Server handle should only be closed when pool is destroyed or engine shutdown. 
Even if ds_pool_srv_open() failed, we should still propagate the global handles by following uuid_copy().

src/pool/srv_target.c
NiuYawei: When a target is in DOWN or DOWNOUT state, the 'ds_pool_child' could be not ready for this collective call.
I think the design of reintegration is that everything (server handles, EC epoch boundary, etc.) should be populated when target is UP, @liuxuezhao , is it correct?

src/pool/srv_target.c
NiuYawei: sure

src/pool/srv_target.c
NiuYawei: ok

src/pool/srv_target.c
NiuYawei: 'ec_agg_disabled' is for debug only, and it looks to me it's not quite related with starting the epoch reporting ULT, so I'd keep it simple.

src/container/srv_target.c
Nasf-Fan: But if we hold reference on that, it is harmful or not? If not, unconditional holding reference seems more simple.

src/pool/srv_iv.c
Nasf-Fan: I mean that the `pool->sp_srv_cont_hdl` has already been opened, do we need to close it before opening `pih->pih_cont_hdl`, otherwise it is overwritten?

src/pool/srv_target.c
Nasf-Fan: But in your patch, the `sp_srv_cont_hdl` is opened by the local logic after `ds_pool_iv_srv_hdl_fetch()`, that is usually once because `srv_hdl_fetched` check and set. So if some target is not `UP` at that time, then it will lost the chance to open such handle?

src/container/srv_target.c
NiuYawei: cont_hdl_get_internal() can't be called for server handle, the server handle isn't inserted in open handle hash.

src/pool/srv_iv.c
NiuYawei: Yes, server handle open is just for propagating the UUID to each xstream.

src/pool/srv_target.c
NiuYawei: Indeed, let me fix it. Thanks.

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16793/8/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16793/9/execution/node/1322/log

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16793/9/execution/node/1395/log

src/pool/srv_iv.c
Nasf-Fan: I mean that at L795, the `pih->pih_cont_hdl` does not match `pool->sp_srv_cont_hdl`, then we open the server handle with `pih->pih_cont_hdl`, and copy it into `pool->sp_srv_cont_hdl`. But before the copy, the old `pool->sp_srv_cont_hdl` may has been opened, and related `hdl` has already been into `pool_child->spc_srv_cont_hdl`, is it possible? If yes, should we close it firstly?

NiuYawei: no close required. "open srv handle" is just about copy the UUID somewhere, see above ds_cont_srv_open().

Nasf-Fan: Oh, right, thanks.

Nasf-Fan (approved):

src/pool/srv_target.c
```diff
@@ -1095,113 +1090,167 @@ ds_pool_put(struct ds_pool *pool)
...
-		D_INFO(DF_UUID" iv conn fetch %d\n", DP_UUID(pool->sp_uuid), rc);
-		D_GOTO(out, rc);
-	}
+	while (!eph_report_exiting(pool)) {
+		sleep_intvl = EPH_REPORT_INTVL;
+
+		/* Fetch pool connection handles */
+		if (!conn_hdl_fetched) {
+			D_INFO(DF_UUID ": Fetching connection handles.\n", DP_UUID(pool->sp_uuid));
+			rc = ds_pool_iv_conn_hdl_fetch(pool);
```
liuxuezhao, line 1160: not sure if I understood correctly.
but for the case that no user's pool connection, the ds_pool_iv_conn_hdls_update/ds_pool_iv_conn_hdl_update() both will not be called. 
So in eph_report_ult() it will always output INFO log -
2026/01/05 16:40:57.154149 aurora-daos-0602 DAOS[60581/0/1371] pool INFO src/pool/srv_iv.c:780 pool_iv_ent_fetch() 82868bf9 iv class id 3, master 5 is still stepping up.: DER_NOTLEADER(-2008): 'Not service leader'
2026/01/05 16:40:57.154155 aurora-daos-0602 DAOS[60581/0/1371] pool INFO src/pool/srv_iv.c:1333 ds_pool_iv_conn_hdl_fetch() 82868bf9 iv fetch failed DER_NOTLEADER(-2008): 'Not service leader'
2026/01/05 16:40:57.154162 aurora-daos-0602 DAOS[60581/0/1371] pool INFO src/pool/srv_target.c:1171 eph_report_ult() 82868bf9: Failed to fetch connection handles. DER_NOTLEADER(-2008): 'Not service leader'

Those logs will repeat forever if no one connect to the pool, that seems unnecessary?
Probably can be refined that inside pool_svc_step_up_cb(), if (iv_hdls == NULL) still update the IV with a special INVALID value to avoid it always fetch again and again in eph_report_ult() that just wasting IV RPC and produce useless INFO log? 
@NiuYawei

liuxuezhao: chatted with niu a little bit offline, I'll check some details and push a PR later. Not very urgent although.

PR #16798, opened 2025-09-02
DAOS-17607 sched: sched monitor

NiuYawei:
Introduce sched monitor, which uses SWIM xstream monitor the schedule activities of SYS and VOS xstreams, once any xstream is detected being inactive for an unexpected long time, warning message will be logged and the engine will be killed.

Set the env var DAOS_SCHED_INACTIVE_MAX (unit is ms) to tune the max inactive time (40000 ms by default), set it to zero will disable the monitor.

Set the env var DAOS_SCHED_MONITOR_KILL to 0 (the default value is 1) to disable engine killing (only warning message logged).

### Steps for the author:

### DIFF
--- a/src/bio/bio_wal.c
+++ b/src/bio/bio_wal.c
@@ -25,6 +25,7 @@ D_CASSERT(sizeof(struct wal_trans_tail) == WAL_CSUM_LEN);
 
 #define WAL_MIN_CAPACITY	(8192 * WAL_BLK_SZ)	/* Minimal WAL capacity, in bytes */
 #define WAL_MAX_TRANS_BLKS	4096			/* Maximal blocks used by a transaction */
+#define WAL_MAX_REPLAY_BLKS     (WAL_MAX_TRANS_BLKS * 2)
 #define WAL_HDR_BLKS		1			/* Ensure atomic header write */
 
 #define META_BLK_SZ		WAL_BLK_SZ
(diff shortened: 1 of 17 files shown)


wangshilong (approved):

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16798/2/execution/node/1268/log

src/engine/sched.c
```diff
@@ -2077,6 +2101,71 @@ sched_try_relax(struct dss_xstream *dx, ABT_pool *pools, uint32_t running)
...
+	/* Check schedule activities every two seconds */
+	hist = &cur_info->si_hist_seqs[cur_dx->dx_xs_id];
+	if ((hist->sm_last_ts + 2000) > cur_info->si_cur_ts)
+		return;
+
+	hist->sm_last_ts = cur_info->si_cur_ts;
+	if (!dss_sched_monitor_enter())
+		return;
+
+	/* Accessing other xstream data without locking */
```
liw, line 2129: [Nit] This comment makes it sounds like we are introducing data races, but after checking every access to `xstream_data`, I realize that you actually mean "these reads never race with the writes", thanks to `dss_sched_monitor_enter` and the "open barrier". Is that right?

liw (approved):

src/engine/sched.c
liw: Hmm, wait... Isn't the read from `si_cur_ts` racing with writes to the field from other xstream?

liw: Maybe the proper solution is to change `si_cur_ts` and `si_cur_seq` to `atomic_ullong`, and access them with `atomic_load`/`atomic_store`?

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16798/3/execution/node/1268/log

src/engine/sched.c
NiuYawei: Yes, the 'dss_sched_monitor_enter" only guarantee the 'dx' not being freed while monitor xs is accessing the 'dx'.

My comment is referring that the 64 bits integer 'si_cur_ts' & 'si_cur_seq' are updated by other xstream but read by this xstream, but I think this kind of race is tolerable, since no matter what value the reader got (old, new or even partial), it won't break the monitor logic, right?

I'm afraid that making them atomic could hurt performance.

liw: It may be okay, though I can't say for sure.

NiuYawei:
@daos-stack/daos-gatekeeper , all CI tests passed, but 3 failing checks are showed here due to CI infrastructure problem (the tests was re-triggered), this PR requires a force landing.

gnailzenh (approved):

PR #16835, opened 2025-09-10
DAOS-17929: ddb: make `vos_file_path` an optional argument

yunpeng-wang-panasas:
Note: This change generates a new ddb command style, and the corresponding ddb test cases may need to be updated.

### Steps for the author:


NiuYawei:
> Note: This change generates a new ddb command style, and the corresponding ddb test cases may need to be updated.

@yunpeng-wang-panasas If that's the case, all the related tests need be updated within same PR too, otherwise, this PR will never pass CI tests, right? I'm wondering if we can keep the old command style and just make this parameter as optional?

yunpeng-wang-panasas:
> > Note: This change generates a new ddb command style, and the corresponding ddb test cases may need to be updated.
> 
> @yunpeng-wang-panasas If that's the case, all the related tests need be updated within same PR too, otherwise, this PR will never pass CI tests, right? I'm wondering if we can keep the old command style and just make this parameter as optional?

@NiuYawei  I use a new method to keep the old command style,  I will check if the vos_file_path is valid, if not, I will add a nil vos_file_path and parse again.  Is this way ok?  BTW, the check logic in this commit is just find if the argument has specific character '/' .  If this logic has problem, Please tell me, Thanks.

NiuYawei:
> > > Note: This change generates a new ddb command style, and the corresponding ddb test cases may need to be updated.
> > 
> > 
> > @yunpeng-wang-panasas If that's the case, all the related tests need be updated within same PR too, otherwise, this PR will never pass CI tests, right? I'm wondering if we can keep the old command style and just make this parameter as optional?
> 
> @NiuYawei I use a new method to keep the old command style, I will check if the vos_file_path is valid, if not, I will add a nil vos_file_path and parse again. Is this way ok? BTW, the check logic in this commit is just find if the argument has specific character '/' . If this logic has problem, Please tell me, Thanks.

hmm, regarding an invalid path as no path specified sounds not correct to me. I'll look into the details and get back to you soon.

PR #16840, opened 2025-09-11
DAOS-17973 control: Allow empty values in property syntax

liw:
It is impossible to turn off all flags of pool property `self_heal`, because `dmg pool set-prop` disallows empty value syntax. This patch changes SetPropertiesFlag.UnmarshalFlag to allow an empty value, leaving its validity to the property type handler.

Features: control

### Steps for the author:

### DIFF
--- a/src/control/cmd/dmg/pool_test.go
+++ b/src/control/cmd/dmg/pool_test.go
@@ -180,12 +180,15 @@ func TestPoolCommands(t *testing.T) {
 	propWithVal := func(key, val string) *daos.PoolProperty {
 		hdlr := daos.PoolProperties()[key]
 		prop := hdlr.GetProperty(key)
-		if val != "" {
-			if err := prop.SetValue(val); err != nil {
-				panic(err)
-			}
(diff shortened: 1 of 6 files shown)


daosbuild3:
Test stage Unit Test on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16840/2/testReport/

liw:
Abandoning in favor of #16848. Here are my Slack messages explaining the details:

> The MS database eventually treats all system attributes (including properties) as map[string]string (see submitSystemAttrsUpdate), and it disallows empty strings as attribute values (see the v == "" check in applySystemUpdate). Hence, although empty string values work for pool properties, they do not work (directly) for system properties. I think we can't and shouldn't change applySystemUpdate because it's part of the Raft state machine.

> When using "", setting system self_heal to "" deleted the property. Then, dmg system get-prop self_heal returned the default value ("all flags turned on", contrary to what we want), rather than "" or "the property is not there".

PR #16863, opened 2025-09-15
DAOS-17982 cq: update GHA ubuntu to Noble

grom72:
Downgrade oracular (EOL) to Noble.

Only used in GHA.

Doc-only: true

### Steps for the author:

### DIFF
--- a/.github/workflows/ci2.yml
+++ b/.github/workflows/ci2.yml
@@ -26,7 +26,7 @@ jobs:
         include:
           - distro: ubuntu
             base: ubuntu
-            with: ubuntu:oracular
+            with: ubuntu:plucky
     env:
       DEPS_JOBS: 10
       COMPILER: clang
--- a/.github/workflows/ci2.yml
+++ b/.github/workflows/ci2.yml
@@ -26,7 +26,7 @@ jobs:
         include:
           - distro: ubuntu
             base: ubuntu
-            with: ubuntu:plucky
+            with: ubuntu:noble
     env:
       DEPS_JOBS: 10
       COMPILER: clang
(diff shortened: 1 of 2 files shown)


daltonbohning (dismissed):

daltonbohning:
@grom72 Should this be Noble now?

grom72:
> @grom72 Should this be Noble now?

With Plucky, we are prepared for the future and can identify some issues in the upcoming LTS release.
But on the other hand as it is only test environment we can use LTS (Noble) to decrease maintenance costs (Noble will have to be replaced soon).

daltonbohning (approved):

JohnMalmberg:
What are we actually trying to test here?

Ubuntu has two types of branches, LTS and not LTS.

We were test building on the future NON-LTS as a non-gating test since that will give us early warning of problems in future Ubuntu releases.

The previous build was oracular, which is a non-LTS release that has gone EOL.   It's replacement is plucky.

So to make this test do what it was intended to do, we also need be as part of the PR actually fixing DAOS to build on plucky.

Ideally this advisory build should automatically updated it self to the latest non-lts build automatically to give us the best early warning.

We should be test building also on the LTS version that could be a production target which is Nobile as a separate test and that should be a gating test.

daltonbohning:
> This totally changes what this workflow was doing from testing the most current interim release of Ubuntu to testing with an older current LTS release of Ubuntu.
> 
> I think we should have a workflow that tests the current LTS release of Ubuntu as a separate workflow.
> 
> Before we land this, I think we need to understand what we are really trying to test.

I'm not sure if we've thought about this much for Ubuntu since it's not yet fully supported or used.
The change at least to `.github/workflows/ci2.yml` I think we should land ASAP because that is a required GHA that is failing on PRs

PR #16872, opened 2025-09-16
DASO-17577 test: Skip certain functional test stages by default

phender:
Avoid utilizing excessive CI resources usage while manually starting CI jobs by disabling functional test stages that are normally skipped when running PRs.

Skip-unit-tests: true
Test-tag: always_passes

### Steps for the author:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -289,10 +289,10 @@ pipeline {
                      defaultValue: true,
                      description: 'Run the Functional on EL 8 test stage')
         booleanParam(name: 'CI_FUNCTIONAL_el9_TEST',
-                     defaultValue: true,
+                     defaultValue: false,
                      description: 'Run the Functional on EL 9 test stage')
         booleanParam(name: 'CI_FUNCTIONAL_leap15_TEST',
-                     defaultValue: true,
+                     defaultValue: false,
                      description: 'Run the Functional on Leap 15 test stage' +
                                   '  Requires CI_MORE_FUNCTIONAL_PR_TESTS')
         booleanParam(name: 'CI_FUNCTIONAL_ubuntu20_TEST',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -306,25 +306,25 @@ pipeline {
...
         booleanParam(name: 'CI_medium_verbs_provider_TEST',
                      defaultValue: true,
                      description: 'Run the Functional Hardware Medium Verbs Provider test stage')
         booleanParam(name: 'CI_medium_verbs_provider_md_on_ssd_TEST',
-                     defaultValue: true,
+                     defaultValue: false,
                      description: 'Run the Functional Hardware Medium Verbs Provider MD on SSD test stage')
         booleanParam(name: 'CI_medium_ucx_provider_TEST',
-                     defaultValue: true,
+                     defaultValue: false,
```
grom72, line 321: Do we use the UCX provider in any of 2.6 installation? If yes, we should keep this test as an active one.

grom72 (dismissed):

Jenkinsfile
phender: This change is to set the default stage execution such that it matches the default stages run in a PR - similar to what was implemented in the master branch.  The UCX provider stage can still be run via in PRs build params or commit pragmas.  It is still also run daily in the `daily-2.6-testing` branch.

phender: In other words, a PR normally skips running this stage and would require adding `Skip-func-hw-test-medium-ucx-provider: false` as a commit pragma to enable running it in the initial build.  This PR does not change this requirement.  It only changes the fact that the user will now need to explicitly check the `CI_medium_ucx_provider_TEST` box if they elect to start a build from Jenkins.

daltonbohning: @grom72 As Phil said, this really just turns the stage off by default when starting a new build within Jenkins, which is already the default when pushing a PR.

PR #16873, opened 2025-09-17
DAOS-17988 client: Include padding space during memory allocation

wiliamhuang:
Robust mutex needs the address of lock is 4-byte aligned. Otherwise, it cannot work as expected. This was observed on x86_64 platform. It has not been checked on other platforms yet. Since proper alignment normally is a good idea performance-wise, we enforce 4-byte alignment for the data stored in hash table record in shared memory to avoid such mis-alignment related issues. During the memory allocation for hash table record, the space used for padding needs to be included.

Allow-unstable-test: true

### Steps for the author:

### DIFF
--- a/src/gurt/shm_dict.c
+++ b/src/gurt/shm_dict.c
@@ -669,6 +669,7 @@ shm_ht_rec_find_insert(d_shm_ht_loc_t shm_ht_loc, const char *key, const int len
 	d_shm_ht_rec_t  rec_next = NULL;
 	char           *value    = NULL;
 	int             rc;
+	int             padding;
 	d_shm_ht_head_t ht_head;
 
 	if (created)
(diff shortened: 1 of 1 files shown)


mchaarawi:
on the other PR you mentioned:
"In case of a robust mutex is stored in the data of hash table record, SHM_MEM_ALIGN (4-byte) aligned is required to make robust mutex working as expected. The padding is enforced to make sure data stored in hash table is 4 bytes aligned."

can you explain why?

wiliamhuang:
> on the other PR you mentioned: "In case of a robust mutex is stored in the data of hash table record, SHM_MEM_ALIGN (4-byte) aligned is required to make robust mutex working as expected. The padding is enforced to make sure data stored in hash table is 4 bytes aligned."
> 
> can you explain why?

The address of the 32-bit integer of "robust futex" needs to be 4-bytes aligned to be robust. This is a fact I observed on x86_64 hardware. If it is not correctly aligned, a robust mutex is not robust (bit "FUTEX_OWNER_DIED" is not set by Linux kernel when a process/thread dies unexpectedly).

mchaarawi:
> > on the other PR you mentioned: "In case of a robust mutex is stored in the data of hash table record, SHM_MEM_ALIGN (4-byte) aligned is required to make robust mutex working as expected. The padding is enforced to make sure data stored in hash table is 4 bytes aligned."
> > can you explain why?
> 
> The address of the 32-bit integer of "robust futex" needs to be 4-bytes aligned to be robust. This is a fact I observed on x86_64 hardware. If it is not correctly aligned, a robust mutex is not robust (bit "FUTEX_OWNER_DIED" is not set by Linux kernel when a process/thread dies unexpectedly).

OK i guess that makes sense bec the kernel needs to do atomic operations and requires the alignment. what i still do not understand is this comment:
"If it is not correctly aligned, a robust mutex is not robust (bit "FUTEX_OWNER_DIED" is not set by Linux kernel when a process/thread dies unexpectedly)".
i would have thought there would just be an error like EINVAL or something..

mchaarawi (approved):

wiliamhuang:
> > > on the other PR you mentioned: "In case of a robust mutex is stored in the data of hash table record, SHM_MEM_ALIGN (4-byte) aligned is required to make robust mutex working as expected. The padding is enforced to make sure data stored in hash table is 4 bytes aligned."
> > > can you explain why?
> > 
> > 
> > The address of the 32-bit integer of "robust futex" needs to be 4-bytes aligned to be robust. This is a fact I observed on x86_64 hardware. If it is not correctly aligned, a robust mutex is not robust (bit "FUTEX_OWNER_DIED" is not set by Linux kernel when a process/thread dies unexpectedly).
> 
> OK i guess that makes sense bec the kernel needs to do atomic operations and requires the alignment. what i still do not understand is this comment: "If it is not correctly aligned, a robust mutex is not robust (bit "FUTEX_OWNER_DIED" is not set by Linux kernel when a process/thread dies unexpectedly)". i would have thought there would just be an error like EINVAL or something..

pthread_mutex_init() did not check the alignment and no error is reported when the address of mutex is not aligned. We do such check in shm_mutex_init. 
When I was testing robust mutex using un-aligned address, I did not observe any error reported by libpthread.

wiliamhuang:
@knard38 Could you please review? Thank you very much!

wiliamhuang:
@mchaarawi Can we force landing this PR? Thank you!

PR #16875, opened 2025-09-17
DAOS-17839 control: Add dmg system self-heal eval cmd

tanabarr:
Add self_heal evaluate command to trigger updates based on the system
self_heal property. Subcommand call to be handled server side by
MS-leader.

- If self_heal.exclude flag is set, call dRPC to trigger
ds_mgmt_drpc_get_group_status() to retrieve the latest CaRT PG SWIM
status from a local engine, compare with the system map and exclude
unavailable engines. 

- If pool_exclude or pool_rebuild is set on the system property,
similar to what has been implemented as part of dmg system rebuild
start, send the latest self_heal value to all PSs using
dsc_pool_svc_eval_self_heal(). The PSs will then handle the
reevaluation (updates related to values of self_heal property flags)
accordingly. 
 
Features: control

### Steps for the author:

### DIFF
--- a/src/control/lib/daos/system_prop.go
+++ b/src/control/lib/daos/system_prop.go
@@ -331,7 +331,7 @@ func (sp *SystemPropertyKey) FromString(val string) error {
 	return errors.Errorf("unknown system property key %q", val)
 }
 
-// SelfHealFlag defines a type to be used to represent self-heal flags.
+// SelfHealFlag defines a type to be used to represent an individual self-heal flag.
 type SelfHealFlag int
 
 // IsValid returns a boolean indicating whether or not the self-heal flag is valid.
@@ -546,8 +546,8 @@ func SystemPropertySelfHealHasFlag(value string, flag SelfHealFlag) bool {
 		return false
 	}
 
-	flags := strings.Split(value, ";")
-	for _, strFlag := range flags {
+	strFlags := strings.Split(value, ";")
+	for _, strFlag := range strFlags {
 		var f SelfHealFlag
 		if err := f.FromString(strFlag); err == nil && f == flag {
 			return true
@@ -587,10 +587,9 @@ func subsets(strings []string, sep string, empty string) []string {
 // SystemProperties returns the map of standard system properties.
 func SystemProperties() SystemPropertyMap {
 	poolProps := PoolProperties()
-	healFlagStrs := []string{
-		SelfHealFlagExclude.String(),
-		SelfHealFlagPoolExclude.String(),
-		SelfHealFlagPoolRebuild.String(),
+	healFlagStrs := []string{}
+	for i := selfHealFlagUnknown + 1; i < selfHealFlagMax; i++ {
+		healFlagStrs = append(healFlagStrs, i.String())
 	}
 	defSelfHealFlagsStr := strings.Join(healFlagStrs, ";")
 
--- a/src/control/server/mgmt_system.go
+++ b/src/control/server/mgmt_system.go
@@ -1398,7 +1398,7 @@ func (svc *mgmtSvc) SystemRebuildManage(ctx context.Context, pbReq *mgmtpb.Syste
 		return nil, errors.Errorf("nil %T", pbReq)
 	}
 
-	if err := svc.checkLeaderRequest(wrapCheckerReq(pbReq)); err != nil {
+	if err := svc.checkLeaderRequest(pbReq); err != nil {
 		return nil, err
 	}
 
@@ -1543,7 +1543,7 @@ func (svc *mgmtSvc) SystemSelfHealEval(ctx context.Context, pbReq *mgmtpb.System
 		return nil, errors.Errorf("nil %T", pbReq)
 	}
 
-	if err := svc.checkLeaderRequest(wrapCheckerReq(pbReq)); err != nil {
+	if err := svc.checkLeaderRequest(pbReq); err != nil {
 		return nil, err
 	}
 
--- a/src/mgmt/srv_drpc.c
+++ b/src/mgmt/srv_drpc.c
@@ -2170,7 +2170,7 @@ ds_mgmt_drpc_pool_self_heal_eval(Drpc__Call *drpc_req, Drpc__Response *drpc_resp
 		D_GOTO(out, rc = -DER_NOMEM);
 
 	// Convert string prop value to bitset.
-	D_INFO("self_heal=%s\n", req->prop_val);
+	D_DEBUG(DB_MGMT, "self_heal=%s\n", req->prop_val);
 	if (strcmp(req->prop_val, "none") != 0) {
 		for (p = strtok_r(req->prop_val, sep, &saveptr); p != NULL;
 		     p = strtok_r(NULL, sep, &saveptr)) {
--- a/src/mgmt/srv_pool.c
+++ b/src/mgmt/srv_pool.c
@@ -650,8 +650,8 @@ out:
  * \param[in]		force			boolean. force a rebuild in op:Fail_reclaim to stop.
  * \param[in]		svc_ranks		Ranks of pool svc replicas.
  *
- * \return		0			Success
- *			Negative value		Error
+ * \return			0				Success
+ *					Negative value	Error
  */
 int
 ds_mgmt_pool_rebuild_stop(uuid_t pool_uuid, uint32_t force, d_rank_list_t *svc_ranks)
@@ -668,8 +668,8 @@ ds_mgmt_pool_rebuild_stop(uuid_t pool_uuid, uint32_t force, d_rank_list_t *svc_r
  * \param[in]		pool_uuid		UUID of the pool.
  * \param[in]		svc_ranks		Ranks of pool svc replicas.
  *
- * \return		0			Success
- *			Negative value		Error
+ * \return			0				Success
+ *					Negative value	Error
  */
 int
 ds_mgmt_pool_rebuild_start(uuid_t pool_uuid, d_rank_list_t *svc_ranks)
@@ -687,8 +687,8 @@ ds_mgmt_pool_rebuild_start(uuid_t pool_uuid, d_rank_list_t *svc_ranks)
  * \param[in]		svc_ranks		Ranks of pool svc replicas.
  * \param[in]		sys_self_heal		Value of system property "self_heal"
  *
- * \return		0			Success
- *			Negative value		Error
+ * \return			0				Success
+ *					Negative value	Error
  */
 int
 ds_mgmt_pool_self_heal_eval(uuid_t pool_uuid, d_rank_list_t *svc_ranks, uint64_t sys_self_heal)
@@ -697,7 +697,6 @@ ds_mgmt_pool_self_heal_eval(uuid_t pool_uuid, d_rank_list_t *svc_ranks, uint64_t
 		"Sending request to evaluate self_heal system property for pool " DF_UUID "\n",
 		DP_UUID(pool_uuid));
 
-	// Shorter deadline used as this is called for each pool and should return quickly.
 	return dsc_pool_svc_eval_self_heal(pool_uuid, svc_ranks, mgmt_ps_call_deadline(),
 					   sys_self_heal);
 }
--- a/src/mgmt/tests/srv_drpc_tests.c
+++ b/src/mgmt/tests/srv_drpc_tests.c
@@ -2726,10 +2726,10 @@ test_drpc_pool_rebuild_start_mgmt_svc_fails(void **state)
 	Drpc__Response resp = DRPC__RESPONSE__INIT;
 
 	setup_rebuild_start_drpc_call(&call, TEST_UUID, "DaosSys");
-	ds_mgmt_pool_self_heal_eval_return = -DER_MISC;
+	ds_mgmt_pool_rebuild_return = -DER_MISC;
 
 	ds_mgmt_drpc_pool_rebuild_start(&call, &resp);
-	expect_drpc_rebuild_start_resp_with_status(&resp, ds_mgmt_pool_self_heal_eval_return);
+	expect_drpc_rebuild_start_resp_with_status(&resp, ds_mgmt_pool_rebuild_return);
 
 	D_FREE(call.body.data);
 	D_FREE(resp.body.data);
@@ -2920,10 +2920,10 @@ test_drpc_pool_self_heal_eval_mgmt_svc_fails(void **state)
 	Drpc__Response resp = DRPC__RESPONSE__INIT;
 
 	setup_self_heal_eval_drpc_call(&call, TEST_UUID, "DaosSys", "exclude;pool_rebuild");
-	ds_mgmt_pool_rebuild_return = -DER_MISC;
+	ds_mgmt_pool_self_heal_eval_return = -DER_MISC;
 
 	ds_mgmt_drpc_pool_self_heal_eval(&call, &resp);
-	expect_drpc_self_heal_eval_resp_with_status(&resp, ds_mgmt_pool_rebuild_return);
+	expect_drpc_self_heal_eval_resp_with_status(&resp, ds_mgmt_pool_self_heal_eval_return);
 
 	D_FREE(call.body.data);
 	D_FREE(resp.body.data);


daosbuild3:
Test stage Build on Leap 15.5 with Intel-C and TARGET_PREFIX completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/1/execution/node/316/log

daosbuild3:
Test stage Build on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/1/execution/node/318/log

daosbuild3:
Test stage Build on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/1/execution/node/326/log

daosbuild3:
Test stage Build on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/1/execution/node/413/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/2/execution/node/1379/log

daosbuild3:
Test stage Unit Test on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16875/6/testReport/

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16875/6/testReport/

daosbuild3:
Test stage Build on Leap 15.5 with Intel-C and TARGET_PREFIX completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/7/execution/node/301/log

daosbuild3:
Test stage Build on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/7/execution/node/309/log

daosbuild3:
Test stage Build on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/7/execution/node/317/log

daosbuild3:
Test stage Build on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/7/execution/node/405/log

tanabarr:
@kccain @liw please take a look to see is there are any glaring omissions or oversights, there are still some unit tests to be added and I've only just started the manual verification of the PR but I wanted early reviews to fix any problems. Functional features should be complete. TIA

daosbuild3:
Test stage Unit Test on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16875/8/testReport/

daosbuild3:
Test stage Unit Test with memcheck on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16875/8/testReport/

src/control/cmd/dmg/system.go
```diff
@@ -768,3 +768,40 @@ type systemRebuildStopCmd struct {
 func (cmd *systemRebuildStopCmd) Execute(_ []string) error {
 	return cmd.execute(control.PoolRebuildOpCodeStop, cmd.Force)
 }
+
+// systemSelfHealCmd represents the system self-heal auto recovery subcommand.
+type systemSelfHealCmd struct {
+	Eval systemSelfHealEvalCmd `command:"eval" description:"Trigger updates based on the system self_heal property"`
```
liw: [Nit] "Trigger self healing if necessary, based on system property self_heal and the current system status", perhaps?

src/control/lib/control/system.go
```diff
@@ -1312,3 +1312,55 @@ func SystemRebuildManage(ctx context.Context, rpcClient UnaryInvoker, req *Syste
...
+// SystemSelfHealEvalResp contains the response.
+type SystemSelfHealEvalResp struct {
+}
+
+// Errors returns collective error for response.
+func (resp *SystemSelfHealEvalResp) Errors() error {
+	return nil
+}
+
+// SystemSelfHealEval will apply interactive rebuild operation to all pools.
```
liw: [Nit] Forgot to update the comment?

src/control/server/mgmt_system.go
```diff
@@ -1442,6 +1442,112 @@ func (svc *mgmtSvc) SystemRebuildManage(ctx context.Context, pbReq *mgmtpb.Syste
...
+	for _, deadRank := range resp.DeadRanks {
+		if member, err := svc.membership.Get(ranklist.Rank(deadRank)); err != nil {
+			// Assume state transition to excluded is legal.
+			member.State = system.MemberStateExcluded
+
+			if err := svc.sysdb.UpdateMember(member); err != nil {
+				return err
+			}
+		}
+	}
```
liw, line 1493: [Question] Why not do what `registerLeaderSubscriptions` in server/server_utils.go does, as pasted below? The "forward" part probably does not apply though.

```go
// Mark the rank as unavailable for membership in
// new pools, etc. Do group update on success.
if err := srv.membership.MarkRankDead(ranklist.Rank(evt.Rank), evt.Incarnation); err != nil {
	srv.log.Errorf("failed to mark rank %d:%x dead: %s", evt.Rank, evt.Incarnation, err)
	if system.IsNotLeader(err) {
		// If we've lost leadership while processing the event,
		// attempt to re-forward it to the new leader.
		evt = evt.WithForwarded(false).WithForwardable(true)
		srv.log.Debugf("re-forwarding rank dead event for %d:%x", evt.Rank, evt.Incarnation)
		srv.evtForwarder.OnEvent(ctx, evt)
	}
	return
}
srv.mgmtSvc.reqGroupUpdate(ctx, false)
```

src/control/server/mgmt_system.go
```diff
@@ -1442,6 +1442,112 @@ func (svc *mgmtSvc) SystemRebuildManage(ctx context.Context, pbReq *mgmtpb.Syste
...
+
+	for _, id := range poolIDs {
+		req := &control.PoolSelfHealEvalReq{
+			ID:      id,
+			PropVal: propVal,
+		}
+		svc.log.Tracef("%T: %+v", req, req)
+
+		if err := control.PoolSelfHealEval(ctx, svc.rpcClient, req); err != nil {
+			return err
```
liw: [Question] Would it make sense to try to continue invoking on as many pools as possible (as `ctx` allows)?

src/mgmt/srv_pool.c
```diff
@@ -680,6 +680,27 @@ ds_mgmt_pool_rebuild_start(uuid_t pool_uuid, d_rank_list_t *svc_ranks)
...
+ *			Negative value		Error
+ */
+int
+ds_mgmt_pool_self_heal_eval(uuid_t pool_uuid, d_rank_list_t *svc_ranks, uint64_t sys_self_heal)
+{
+	D_DEBUG(DB_MGMT,
+		"Sending request to evaluate self_heal system property for pool " DF_UUID "\n",
+		DP_UUID(pool_uuid));
+
+	return dsc_pool_svc_eval_self_heal(pool_uuid, svc_ranks, mgmt_ps_call_deadline(),
```
liw, line 700: [Suggestion] Perhaps it would be more practical for now to use an earlier deadline than "now + 5 minutes" here, for the MS calls this for every pool. E.g., some deadline between "now + 14 s" and "now + 1 minutes"?

src/control/cmd/dmg/system.go
tanabarr: done

src/control/lib/control/system.go
tanabarr: done

src/control/server/mgmt_system.go
tanabarr: Yes that will be a better idea, the differences being that incarnation value gets verified and we do the group update after. Thanks. Done.

tanabarr: Hmm, we don't have a rank incarnation value though do we as the only version number we have in the get_group_status flow is the map version so most of the logic in MarkRankDead is irrelevant in this context. I think we just keep the logic the same but do the group update after?

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/8/execution/node/1347/log

src/control/server/mgmt_system.go
tanabarr: Sure but then we lose error info (all at least it's not returned to the user), is it acceptable that we just log errors in this case (for the moment we can obviously change to return per-pool results in the future)? Should any "last-error" be returned in case of failures or just return success and rely on logging to report Pool call failures?

My feeling with this kind of situation is that half measures will bite us, there is ambiguity in carrying on despite error(s) and there is a risk that we don't return them to the user in an effective fashion.

Options:
1. Fail fast and return on any Pool call failure - main disadvantage is that we get stuck if one pool fails
2. Carry on and just log errors - main disadvantage is that user may ignore logging and assume success
3. Report each pool result individually as we've done elsewhere - time pressure would like to get first iteration out
4. Another approach that we've used elsewhere, report "there were errors on pools X, Y, Z, check server log for details"

Shall we go for option nr. 4 for the moment or just go ahead and implement nr. 3? @liw @kccain

src/mgmt/srv_pool.c
tanabarr: done

src/control/server/mgmt_system.go
liw: Yes, let's do that for now. If you agree, we shall file a ticket for adding incarnation numbers to GetGroupStatusResp, once we have all the same basic functionality landed.

src/control/server/mgmt_system.go
liw: The "get stuck" part in option 1 led me to ask the original question---it's a concern. On the other hand, it's **not** acceptable to return success if there has been one or more errors. We don't need to implement option 3 in this PR or the initial version of our feature, I think. I'd vote for going ahead with option 4.

src/control/server/mgmt_system.go
liw: I've checked the C side: We don't use incarnation numbers when recovering events. Hence, forget about incarnation numbers here too---no need to file a ticket for adding them. Let's just do what your comment above suggested.

tanabarr: Although I've updated the MarkRankDead implementation and used it to centralise the "update Rank to excluded" Logic in https://github.com/daos-stack/daos/pull/16875/commits/eed5a7ed0bae30413d9bc04b4079a0776e4aaa02 . Let me know if those changes are not acceptable.

src/control/server/mgmt_system.go
tanabarr: Updated in line with option 4. Done

daosbuild3:
Test stage Build on Leap 15.5 with Intel-C and TARGET_PREFIX completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/11/execution/node/303/log

daosbuild3:
Test stage Build on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/11/execution/node/311/log

daosbuild3:
Test stage Build on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/11/execution/node/319/log

daosbuild3:
Test stage Build on Leap 15.5 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/11/execution/node/407/log

src/control/server/mgmt_system.go
tanabarr: fixing overflow issue

src/control/lib/daos/system_prop.go
```diff
@@ -331,11 +331,49 @@ func (sp *SystemPropertyKey) FromString(val string) error {
 	return errors.Errorf("unknown system property key %q", val)
 }
 
+// SelfHealFlag defines a type to be used to represent self-heal flags.
```
kccain:
```suggestion
// SelfHealFlag defines a type to be used to represent an individual self-heal flag.
```

src/control/lib/daos/system_prop.go
```diff
@@ -493,14 +541,15 @@ func pph2sp(key SystemPropertyKey, pph *PoolPropHandler, def string) SystemPrope
...
-func SystemPropertySelfHealHasFlag(value string, flag string) bool {
+func SystemPropertySelfHealHasFlag(value string, flag SelfHealFlag) bool {
 	if value == "none" {
 		return false
 	}
 
 	flags := strings.Split(value, ";")
-	for _, f := range flags {
-		if f == flag {
+	for _, strFlag := range flags {
```
kccain: Suggestion: rename the variable to strFlags, as it would reduce (momentary) confusion while reading the code, for this code reader at least.

kccain (commented):
first pass, couple of minor suggestions so far. Big patch, need more time. I spent more energy looking at the self_heal.pool_exclude/pool_rebuild evaluation and notification of PSes than self_heal.exclude handling.

src/control/cmd/dmg/system.go
```diff
@@ -768,3 +768,40 @@ type systemRebuildStopCmd struct {
 func (cmd *systemRebuildStopCmd) Execute(_ []string) error {
 	return cmd.execute(control.PoolRebuildOpCodeStop, cmd.Force)
 }
+
+// systemSelfHealCmd represents the system self-heal auto recovery subcommand.
+type systemSelfHealCmd struct {
+	Eval systemSelfHealEvalCmd `command:"eval" description:"Trigger self healing if necessary based on system property self_heal and the current system status"`
```
kjacque, line 774: The name of the command feels a little misleading to me. "eval" implies it will evaluate and tell you something about the state of the system. Maybe "run" would be better, indicating a process will do something if it needs to.

There's an implication in this structure that there will be more self-heal related commands--is that likely? Or would it be better for the command to be "dmg system self-heal" with nothing further added on?

src/control/lib/control/pool.go
```diff
@@ -1407,9 +1405,53 @@ func PoolRebuildManage(ctx context.Context, rpcClient UnaryInvoker, req *PoolReb
...
+// PoolSelfHealEvalReq contains pool-self-heal-eval operation parameters. ID identifies the pool on which the
+// operation should be run.
+type PoolSelfHealEvalReq struct {
+	poolRequest
+	ID      string
+	PropVal string
+}
+
+// PoolSelfHealEval sends a pool-self-heal-eval request to the pool service leader.
+func PoolSelfHealEval(ctx context.Context, rpcClient UnaryInvoker, req *PoolSelfHealEvalReq) error {
```
kjacque, line 1424: Same comment on the name. I think this should just be "PoolSelfHeal".

src/control/lib/control/system.go
```diff
@@ -1312,3 +1312,55 @@ func SystemRebuildManage(ctx context.Context, rpcClient UnaryInvoker, req *Syste
...
+type SystemSelfHealEvalResp struct {
+}
+
+// Errors returns collective error for response.
+func (resp *SystemSelfHealEvalResp) Errors() error {
+	return nil
+}
+
+// SystemSelfHealEval will trigger actions based on the value of the system self_heal property.
+func SystemSelfHealEval(ctx context.Context, rpcClient UnaryInvoker, req *SystemSelfHealEvalReq) (*SystemSelfHealEvalResp, error) {
```
kjacque, line 1332: Same comment on the name.

src/control/lib/daos/drpc.go
```diff
@@ -217,6 +219,10 @@ const (
 	MethodPoolRebuildStart MgmtMethod = C.DRPC_METHOD_MGMT_POOL_REBUILD_START
 	// MethodPoolRebuildStop defines a method stop an interactive pool rebuild
 	MethodPoolRebuildStop MgmtMethod = C.DRPC_METHOD_MGMT_POOL_REBUILD_STOP
+	// MethodGroupStatusGet defines a method for retrieving the group status of a system
+	MethodGroupStatusGet MgmtMethod = C.DRPC_METHOD_MGMT_GROUP_STATUS_GET
```
kjacque, line 223: naming nit - maybe should be "group get status" for consistency with how we name get/set methods elsewhere

src/control/lib/daos/system_prop.go
```diff
@@ -538,23 +587,36 @@ func subsets(strings []string, sep string, empty string) []string {
 // SystemProperties returns the map of standard system properties.
 func SystemProperties() SystemPropertyMap {
 	poolProps := PoolProperties()
+	healFlagStrs := []string{
+		SelfHealFlagExclude.String(),
+		SelfHealFlagPoolExclude.String(),
+		SelfHealFlagPoolRebuild.String(),
+	}
```
kjacque: Nothing particularly wrong with this, but it might be better to do this programmatically if it's intended to represent all the strings (not just the default subset).

src/proto/mgmt/pool.proto
```diff
@@ -368,3 +368,15 @@ message PoolRebuildStopReq
...
 
 // PoolRebuildStop uses DaosResp
+
+// PoolSelfHealEvalReq contains information about a request to evaluate the self_heal system
+// property value on a given pool.
+message PoolSelfHealEvalReq
+{
+	string          sys       = 1; // DAOS system identifier
+	string          id        = 2; // DAOS pool identifier
+	string          prop_val  = 4; // self_heal system property value
```
kjacque, line 378: Wonder if it would be better to define a shared set of enum flags instead of passing strings around.

src/mgmt/srv_pool.c
```diff
@@ -650,8 +650,8 @@ ds_mgmt_pool_get_prop(uuid_t pool_uuid, d_rank_list_t *svc_ranks,
  * \param[in]		force			boolean. force a rebuild in op:Fail_reclaim to stop.
  * \param[in]		svc_ranks		Ranks of pool svc replicas.
  *
- * \return			0				Success
- *					Negative value	Error
+ * \return		0			Success
+ *			Negative value		Error
```
kjacque: Something strange happened to the indentation... clang-format does this to me sometimes too, but doesn't complain if I fix it manually. Not sure why.

src/mgmt/srv_drpc.c
```diff
@@ -2131,6 +2131,85 @@ ds_mgmt_drpc_pool_rebuild_stop(Drpc__Call *drpc_req, Drpc__Response *drpc_resp)
...
+		DL_ERROR(rc, "Pool UUID is invalid");
+		goto out;
+	}
+
+	svc_ranks = uint32_array_to_rank_list(req->svc_ranks, req->n_svc_ranks);
+	if (svc_ranks == NULL)
+		D_GOTO(out, rc = -DER_NOMEM);
+
+	// Convert string prop value to bitset.
+	D_INFO("self_heal=%s\n", req->prop_val);
```
kjacque: Might be inclined to make this debug instead of info

src/mgmt/tests/srv_drpc_tests.c
```diff
@@ -2726,10 +2726,10 @@ test_drpc_pool_rebuild_start_mgmt_svc_fails(void **state)
 	Drpc__Response resp = DRPC__RESPONSE__INIT;
 
 	setup_rebuild_start_drpc_call(&call, TEST_UUID, "DaosSys");
-	ds_mgmt_pool_rebuild_return = -DER_MISC;
+	ds_mgmt_pool_self_heal_eval_return = -DER_MISC;
 
 	ds_mgmt_drpc_pool_rebuild_start(&call, &resp);
-	expect_drpc_rebuild_start_resp_with_status(&resp, ds_mgmt_pool_rebuild_return);
+	expect_drpc_rebuild_start_resp_with_status(&resp, ds_mgmt_pool_self_heal_eval_return);
```
kjacque: Typo or accidental replacement?

src/mgmt/tests/srv_drpc_tests.c
```diff
@@ -2842,6 +2842,108 @@ test_drpc_pool_rebuild_stop_success(void **state)
...
+}
+
+static void
+test_drpc_pool_self_heal_eval_mgmt_svc_fails(void **state)
+{
+	Drpc__Call     call = DRPC__CALL__INIT;
+	Drpc__Response resp = DRPC__RESPONSE__INIT;
+
+	setup_self_heal_eval_drpc_call(&call, TEST_UUID, "DaosSys", "exclude;pool_rebuild");
+	ds_mgmt_pool_rebuild_return = -DER_MISC;
```
kjacque: ds_mgmt_pool_rebuild_return? For the self_heal test?

src/control/server/mgmt_system.go
```diff
@@ -1442,6 +1443,140 @@ func (svc *mgmtSvc) SystemRebuildManage(ctx context.Context, pbReq *mgmtpb.Syste
 	return pbResp, nil
 }
 
+func (svc *mgmtSvc) selfHealExcludeRanks(ctx context.Context) error {
```
kjacque, line 1446: Reading through the code, I'm a little confused why this step is needed at all. Wouldn't the ranks already be marked excluded if they either crashed or were excluded by SWIM? Maybe I just need more details on how this is supposed to work. How would the previously-excluded ranks, which are still excluded in the group map, end up non-excluded?

src/control/server/server_utils.go
```diff
@@ -734,6 +734,37 @@ func registerFollowerSubscriptions(srv *server) {
...
+	ts, err := evt.GetTimestamp()
+	if err != nil {
+		srv.log.Errorf("bad event timestamp %q: %s", evt.Timestamp, err)
+		return
+	}
+	srv.log.Debugf("%s marked rank %d:%x dead @ %s", evt.Hostname, evt.Rank, evt.Incarnation,
+		ts)
+	if evt.Incarnation == 0 {
+		srv.log.Errorf("invalid zero incarnation value in event %+v", evt)
+	}
```
kjacque, line 747: Aren't you sending out an event with incarnation = 0?

src/control/system/membership.go
```diff
@@ -662,37 +662,49 @@ func (m *Membership) IsRankAdminExcluded(rank Rank) bool {
...
+		m.log.Debug(msg)
+		return false, nil
 	}
 
-	if member.State == MemberStateJoined && member.Incarnation > incarnation {
-		m.log.Debugf("ignoring rank dead event for previous incarnation of %d (%x < %x)", rank, incarnation, member.Incarnation)
-		return errors.Errorf("event is for previous incarnation of %d", rank)
+	// Perform check on rank incarnation number if valid value passed and a currently
+	// active/joined rank is going to be excluded.
+	if member.State == MemberStateJoined && incarnation > 0 &&
```
kjacque, line 694: I don't love using a magic incarnation value as an indirect "force" flag.

src/control/server/mgmt_system.go
```diff
@@ -1442,6 +1443,140 @@ func (svc *mgmtSvc) SystemRebuildManage(ctx context.Context, pbReq *mgmtpb.Syste
...
+}
+
+// SystemSelfHealEval triggers actions based on self_heal system property values.
+func (svc *mgmtSvc) SystemSelfHealEval(ctx context.Context, pbReq *mgmtpb.SystemSelfHealEvalReq) (*mgmtpb.DaosResp, error) {
+	if pbReq == nil {
+		return nil, errors.Errorf("nil %T", pbReq)
+	}
+
+	if err := svc.checkLeaderRequest(wrapCheckerReq(pbReq)); err != nil {
+		return nil, err
```
kjacque, line 1547: Huh. Does this all happen in checker mode? If so I think the dmg command should be "dmg check self-heal"

kjacque (changes requested):
Sorry, it's such a big patch I have a lot of comments. Some look like bugs, and some may just be that I don't fully understand the feature. Maybe we can chat briefly tomorrow if you have time.

src/control/cmd/dmg/system.go
tanabarr: @liw unless you feel strongly about making this change I would be inclined to leave as is. @kjacque we've sent some time discussing the syntax so unless you have a strong objection I would like to go ahead as is. we need to get the first version in soon. Also providing subcommands e.g. eval gives us space for iterating later on, if we remove eval and we need to add more command later on then we won't be able to make backwards compatible changes

src/control/lib/control/pool.go
tanabarr: I like to keep it as is for the moment.

src/control/lib/control/system.go
tanabarr: as above.

src/control/lib/daos/drpc.go
tanabarr: <noun+verb> is normally what we aim for, "GroupStatus" is the noun and "Get" is the verb. this is inconsistent with the dRPC Proto definitions but that's the state of the code base so I'm going with consistency with the current trends in the code base.

src/control/lib/daos/system_prop.go
tanabarr: Done.

src/control/lib/daos/system_prop.go
tanabarr: Done.

src/control/lib/daos/system_prop.go
tanabarr: Done.

src/control/server/mgmt_system.go
tanabarr: If the exclude policy hasn't been enabled in the past and it's been enabled in a recent system property change then the DeadRanks have to be excluded from the MS at the time of the "eval" command trigger.

src/control/server/mgmt_system.go
tanabarr: Hmm, there seems to be a number of commands in mgmt_system.go which are using this wrapper, which commands should be? I'm imagining it might be a copy-paste error. I will move from these new commands at least, let me know if I should remove others. Thanks for catching this!

tanabarr: Removed from RebuildManage and SelfHealEval functions, Done.

src/control/server/server_utils.go
tanabarr: Not that I'm aware of, from talking with @liw I took to understand that zero incarnation value are invalid. Thats why I can use a zero value in MarkRankDead to skip the check and that's why I check and log an invalid zero value here. Altering control flow based on a zero value here seemed like overkill.

src/control/system/membership.go
tanabarr: What would you prefer? skipIncarnationCheck flag and leave the previous behaviour where incarnation zero is accepted as valid?

src/mgmt/srv_drpc.c
tanabarr: ok, Done.

src/mgmt/srv_pool.c
tanabarr: I've tried to manually corrected but if the linter wants to change it again I'm not going to fight. I've decided that life is too short to fight with our project's self-imposed linters.

src/mgmt/tests/srv_drpc_tests.c
tanabarr: typo, good catch, thanks. Done.

src/proto/mgmt/pool.proto
tanabarr: would like to leave it as is for this iteration.

src/control/system/membership.go
tanabarr: Could this change can be made as a follow-up{?

src/mgmt/tests/srv_drpc_tests.c
tanabarr: good catch thanks, Done

src/control/lib/daos/drpc.go
kjacque: I guess I was thinking about how methods like pool_get_prop and pool_get_acl are formatted in a more natural English way, with sort of a component_action format. It's pretty mixed in the current code base.

src/proto/mgmt/pool.proto
kjacque: I think this decision is important, as it has interop implications. If there's a good reason to use strings, so be it, but we are stuck with it if this is what is used in 2.8.

src/control/cmd/dmg/system.go
kjacque: OK. Let's leave as-is for now. I think I need more background on the feature, maybe the name makes more sense in that context.

src/control/server/mgmt_system.go
kjacque: Some of it may be a copy/paste error, some may be intentional. According to the comments:
```
// checkerRequest is a wrapper around a request that is made on behalf of
// the checker or is otherwise allowed to be made while the checker is enabled.
```

So it's needed if it is OK to call the command in checker mode (not just checker commands, but stuff like system query can be called in either mode, for example). RebuildManage definitely makes no sense in checker mode, I'm not sure about self-heal though. @liw?

src/mgmt/srv_pool.c
kjacque: Fair. Yeah, if it changes it a second time I usually just leave it. I wish we could figure out what exactly it's doing when it misaligns previously good indentation though.

src/control/system/membership.go
kjacque: Fine with changing as a follow-up. I'd suggest just a "force" if that's what's intended, i.e. always update, don't bother checking anything.

src/control/lib/daos/system_prop.go
kjacque: Doesn't look like the change was pushed, if you updated this. I'm fine with deferring that if you want though.

src/control/server/server_utils.go
kjacque: Sorry, I think I see where I got confused on this. This is fine.

src/proto/mgmt/pool.proto
tanabarr: @liw we discussed this and we went with strings, do we keep it?

src/control/lib/daos/system_prop.go
tanabarr: huh, I thought I did

tanabarr: looks changed: https://github.com/daos-stack/daos/pull/16875/commits/fb1eafd310fe3f8c387816d12e9f64b5a79ae1af#diff-a3b56eae5cbaa9f3dfa34047c3ea1f66bf9d53f1a55f30755995876fc163a4abR590

kjacque: Ah, it comes up now. Maybe I was looking at it in a stale window. Thanks.

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/14/execution/node/1290/log

src/control/server/mgmt_system.go
liw: I think the check mode does not want any _automatic_ self healing. So it might be wise to disallow `eval` in the check mode. If in the future we want to allow it, the change will not break any use cases apparently.

src/control/server/mgmt_system.go
liw: @kjacque, Tom has a next ticket that will do something along this line (my prototype): https://github.com/daos-stack/daos/pull/16797/commits/a479a4f87dfb42c1fabb3767aed216aaa72d66aa. Hence, after re-enabling `self_heal.exclude`, one may want to invoke `eval`, which needs to "recover" the ignored dead events.

src/proto/mgmt/pool.proto
liw: Well, I don't have a strong opinion myself. But the latest decision is to use strings throughout, since in the end the MS DB only takes strings (contrary to what I thought when writing the design). So I was actually surprised to find this PR attempts to convert them to numerical values, but anyway, Kris and you know what the best way better than I do.

src/control/cmd/dmg/system.go
liw: Tom, as we discussed, although not ideal, `eval` seems more accurate than `run`. Because if the evaluation result is "no action required", then nothing new will "run" after a successful invocation of this command: If the name is `run`, some people may complain that nothing new is "running"---whatever that means.

src/proto/mgmt/pool.proto
liw, line 378: [Nit] Since this is in a pool context (i.e., this is a request sent to a PS), we should name `prop_val` differently, such as `sys_prop_val`. (`id` and `svc_ranks` are good, because they mean "pool ID" and "pool service ranks".)

tanabarr: I will update in a follow-up, where I will also improve unit test coverage.

src/control/server/mgmt_system.go
tanabarr: I removed the checker request wrapper for eval

tanabarr:
CI build issues attributed to known issue DAOS-18042

src/proto/mgmt/pool.proto
tanabarr: I guess we don't need to add the numeric flags, I can remove them in the follow-up if you want.

liw: I think either way should work, so better do what _you_ consider best.

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16875/14/execution/node/1470/log

tanabarr:
Thanks @phender for looking into the intermittent failures:
> The 23-./daos_test/suite.py:DaosCoreTest.test_daos_rebuild_ec [test timeout](https://jenkins.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16875/14/artifact/Functional%20Hardware%20Medium%20Verbs%20Provider%20MD%20on%20SSD/daos_test/suite.py) and related [missing cmocka](https://jenkins.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16875/14/testReport/FTEST_daos_test.DaosCoreTest/DAOS_Rebuild_EC/Test_Hardware___Functional_Hardware_Medium_Verbs_Provider_MD_on_SSD___23___daos_test_suite_py_DaosCoreTest_test_daos_rebuild_ec_run_agent_config_transport_config_daos_tests_args_daos_test_num_clients_pools_created_stopped_ranks_test_name_dmg_hosts_launch_pool_server_config_control_metadata_engines_0_storage_0_1_2_timeouts_1b1d/) file looks like https://daosio.atlassian.net/browse/DAOS-17578

> The 1-./pool/rf.py:PoolRedunFacProperty.test_rf_pool_property [failure](https://jenkins.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16875/14/artifact/Functional%20Hardware%20Large%20MD%20on%20SSD/pool/rf.py) is due to hdr-112 Available memory (RAM) insufficient for configured 168 GiB ram-disk size, want 151 GiB RAM but only have 145 GiB.  This test is running in a HW stage, yet has system_ram_reserved: 16 in the server config file.  Shouldn't this be removed for HW testing?  This test did pass in [this week's weekly build](https://jenkins.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/weekly-testing/19/artifact/Functional%20Hardware%20Large%20MD%20on%20SSD/pool/rf.py/) on a different cluster.

> The 2-./container/boundary.py:BoundaryTest.test_container_boundary [failure](https://jenkins.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16875/14/artifact/Functional%20Hardware%20Large%20MD%20on%20SSD/container/boundary.py) is due to ERROR: dmg: pool create failed: DER_NOSPACE(-1007): No space on storage target. This failure was also seen in this [week's weekly build](https://jenkins.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/weekly-testing/19/artifact/Functional%20Hardware%20Large%20MD%20on%20SSD/container/boundary.py).

src/control/system/membership.go
tanabarr: I don't want to skip verifying against illegal transitions though, should I use a forceIncarnation flag or just leave it?

PR #16857, opened 2025-09-12
DAOS-17981 build: Fix CI_BUILD_PACKAGES_ONLY use

phender:
Update the skip build commit pragma and build parameters to reflect the recent changes to the build stages.

Skip-unit-tests: true
Skip-fault-injection-test: true
Test-tag: test_setup_vm

### Steps for the author:

* [ ] Commit message follows the [guidelines](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [ ] Appropriate [Features or Test-tag](https://daosio.atlassian.net/wiki/spaces/DC/pages/10984259629/Test+Tags) pragmas were used.
* [ ] Appropriate [Functional Test Stages](https://daosio.atlassian.net/wiki/spaces/DC/pages/12147556353/CI+Functional+Test+Stages) were run.
* [ ] At least two positive code reviews including at least one code owner from each category referenced in the PR.
* [ ] Testing is complete. If necessary, forced-landing label added and a reason added in a comment.

#### After all prior steps are complete:
* [ ] Gatekeeper requested (daos-gatekeeper added as a reviewer).

PR #16881, opened 2025-09-18
DAOS-17993 ci: New stage name causes missing RPMs

grom72:
There were several ways to disable regular builds but to keep RPM builds. For example CI_BUILD_PACKAGES_ONLY disable regular builds. Now regular builds are the only way to get RPMs. We have to use stage name dedicated to RPM builds as skipping mechanism relay of stage names.

Allow-unstable-test: true

### Steps for the author:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -505,7 +505,7 @@ pipeline {
                 expression { !skipStage() }
             }
             parallel {
-                stage('Build on EL 8.8') {
+                stage('Build RPM on EL 8') {
                     when {
                         beforeAgent true
                         expression { !params.CI_el8_NOBUILD && !skipStage() }
@@ -559,7 +559,7 @@ pipeline {
                         }
                     }
                 }
-                stage('Build on EL 9') {
+                stage('Build RPM on EL 9') {
                     when {
                         beforeAgent true
                         expression { !params.CI_el9_NOBUILD && !skipStage() }
@@ -613,7 +613,7 @@ pipeline {
                         }
                     }
                 }
-                stage('Build on Leap 15.5') {
+                stage('Build RPM on Leap 15.5') {
                     when {
                         beforeAgent true
                         expression { !params.CI_leap15_NOBUILD && !skipStage() }


daltonbohning (dismissed):

JohnMalmberg (approved):

Jenkinsfile
```diff
@@ -731,7 +731,7 @@ pipeline {
                         }
                     }
                 }
-                stage('Unit Test bdev on EL 8.8') {
+                stage('Unit Test bdev on EL 8') {
```
daltonbohning: Same

Jenkinsfile
```diff
@@ -753,7 +753,7 @@ pipeline {
                         }
                     }
                 }
-                stage('NLT on EL 8.8') {
+                stage('NLT on EL 8') {
```
daltonbohning: Same

Jenkinsfile
```diff
@@ -793,7 +793,7 @@ pipeline {
                         }
                     }
                 }
-                stage('Unit Test with memcheck on EL 8.8') {
+                stage('Unit Test with memcheck on EL 8') {
```
daltonbohning: Same

Jenkinsfile
```diff
@@ -817,8 +817,8 @@ pipeline {
                             job_status_update()
                         }
                     }
-                } // stage('Unit Test with memcheck on EL 8.8')
-                stage('Unit Test bdev with memcheck on EL 8.8') {
+                } // stage('Unit Test with memcheck on EL 8')
+                stage('Unit Test bdev with memcheck on EL 8') {
```
daltonbohning: Same

Jenkinsfile
```diff
@@ -874,8 +874,8 @@ pipeline {
                             job_status_update()
                         }
                     }
-                } // stage('Functional on EL 8.8 with Valgrind')
-                stage('Functional on EL 8.8') {
+                } // stage('Functional on EL 8 with Valgrind')
+                stage('Functional on EL 8') {
```
daltonbohning: Same

Jenkinsfile
```diff
@@ -964,7 +964,7 @@ pipeline {
                         }
                     } // post
                 } // stage('Functional on Ubuntu 20.04')
-                stage('Fault injection testing on EL 8.8') {
+                stage('Fault injection testing on EL 8') {
```
daltonbohning: Same

Jenkinsfile
```diff
@@ -1094,7 +1094,7 @@ pipeline {
                 } // stage('Test RPMs on Leap 15.5')
             } // parallel
         } // stage('Test')
-        stage('Test Storage Prep on EL 8.8') {
+        stage('Test Storage Prep on EL 8') {
```
daltonbohning: This one is ok to change

Jenkinsfile
```diff
@@ -853,7 +853,7 @@ pipeline {
                 expression { !paramsValue('CI_FUNCTIONAL_TEST_SKIP', false) }
             }
             parallel {
-                stage('Functional on EL 8.8 with Valgrind') {
+                stage('Functional on EL 8 with Valgrind') {
```
daltonbohning: This one is okay to change

Jenkinsfile
```diff
@@ -709,7 +709,7 @@ pipeline {
                 expression { !skipStage() }
             }
             parallel {
-                stage('Unit Test on EL 8.8') {
+                stage('Unit Test on EL 8') {
```
daltonbohning: We cannot do this easily because the landing requirement explicitly requires 8.8. If you look at the pending stages you'll see
> build/Fault injection testing on EL 8.8 Expected — Waiting for status to be reported

grom72: But it seems to work in build [#3](https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16881/3/pipeline-overview/?selected-node=492).
As I understand it, we need to keep consistency in the naming of the stages, as the particular version is derived from there.

daltonbohning: Whether or not the stage runs, the requirements for landing include stage names with `8.8`. If you look at the checks at the bottom of the page you will see that the stages completed in Jenkins are still pending

Jenkinsfile
grom72: Revert

Jenkinsfile
grom72: Revert

Jenkinsfile
grom72: Revert

Jenkinsfile
grom72: Revert

Jenkinsfile
grom72: Revert

daltonbohning (approved):

ryon-jensen (approved):

daltonbohning:
@grom72 I think ultimately #16857 is a more robust solution because it moves the skip logic into `daos` instead of being buried in pipeline-lib. And IMO skip logic looking at the stage name has always been problematic

grom72:
> @grom72 I think ultimately #16857 is a more robust solution because it moves the skip logic into `daos` instead of being buried in pipeline-lib. And IMO skip logic looking at the stage name has always been problematic
@daltonbohning I agree but:
- now we have two independent solutions for skipping
- we have more and more legacy (unused) code in the pipeline lib.

grom72:
Will be solved by https://github.com/daos-stack/daos/pull/16857

PR #16907, opened 2025-09-25
DAOS-17772 rebuild: refine some rebuild/EC agg/EC deg fetch process

liuxuezhao:
1. add container RDB KV ds_cont_prop_ec_agg_eph for EC aggregation epoch boundary, store it when bump and load it after restart.
2. synchronize the ec agg boundary before rebuild
3. change ec agg boundary IV refresh to be EAGER sync
4. wait discard's completion in ds_pool_tgt_discard_handler()
5. fix a cart IV sync bug that ignored GRP_VER err case's err code
6. fetch ec agg boundary IV in fetch RPC handler when it is zero
7. fix an EC agg partial update stripe processing bug
8. fix a sgl sg_nr_out and iov_len process bug
9. fix a EC degraded fetch bug related with punch extent
10. some refines for rebuild upbound epoch
- rename rebuild stable epoch to rebuild upbound epoch
- select "current hlc + epsilon" (ds_rebuild_get_upbound_eph()) as the rebuild upbound epoch.
- use the global consistent rebuild upbound epoch as rebuild fence to avoid VOS aggregation cross the epoch.

### Steps for the author:

### DIFF
--- a/src/object/srv_ec_aggregate.c
+++ b/src/object/srv_ec_aggregate.c
@@ -1085,6 +1085,7 @@ agg_update_parity(struct ec_agg_entry *entry, uint8_t *bit_map,
 			goto out;
 		while (!isset(bit_map, j))
 			j++;
+		D_ASSERTF(j < k, "bad cell_idx %d, number of data shards %d\n", j, k);
 		agg_diff_preprocess(entry, diff, j);
 		ec_encode_data_update(cell_bytes, k, p, j,
 				      entry->ae_codec->ec_gftbls, diff,
@@ -1951,15 +1952,15 @@ out:
 			/* offload of ds_obj_update to push remote parity */
 			rc = agg_peer_update(entry, write_parity);
 			if (rc)
-				D_ERROR("agg_peer_update fail: "DF_RC"\n",
-					DP_RC(rc));
+				DL_ERROR(rc, "agg_peer_update failed, write_parity %d\n",
+					 write_parity);
 		}
 
 		if (rc == 0) {
 			rc = agg_update_vos(agg_param, entry, write_parity);
 			if (rc)
-				D_ERROR("agg_update_vos failed: "DF_RC"\n",
-					DP_RC(rc));
+				DL_ERROR(rc, "agg_update_vos failed, write_parity %d\n",
+					 write_parity);
 		}
 	}
 
--- a/src/object/srv_ec_aggregate.c
+++ b/src/object/srv_ec_aggregate.c
@@ -1074,7 +1074,7 @@ agg_update_parity(struct ec_agg_entry *entry, uint8_t *bit_map,
 	buf  = entry->ae_sgl.sg_iovs[AGG_IOV_DATA].iov_buf;
 	diff = entry->ae_sgl.sg_iovs[AGG_IOV_DIFF].iov_buf;
 
-	for (i = 0, j = 0; i < cell_cnt; i++) {
+	for (i = 0, j = 0; i < cell_cnt; i++, j++) {
 		old = &obuf[i * cell_bytes];
 		new = &buf[i * cell_bytes];
 		vects[0] = old;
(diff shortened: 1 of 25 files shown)


src/vos/vos_io.c
```diff
@@ -1104,14 +1105,44 @@ akey_fetch_recx(daos_handle_t toh, const daos_epoch_range_t *epr,
...
+				}
+				/* process this punch ext, for degraded fetch just skip it */
+				index = lo + nr;
+				holes = nr;
+				biov_set_hole(&biov, holes * inob);
+				/* skip the hole */
+				rc = iod_fetch(ioc, &biov);
+				if (rc != 0)
+					goto failed;
+				holes = 0;
```
NiuYawei, line 1138: So the holes (generated by punch records) above 'shadow_ep' won't be merged in 'with_shadow' case? Looks not a big deal though.

NiuYawei (approved):

src/vos/vos_io.c
liuxuezhao: I think you mean the contiguous punch records above shadow_ep, TBH I did not consider that case to make it merge (so can iod_fetch one time) before.
make it mergable seems need a status to differentiate the left hole is punched or normal hole, looks will make the code a little bit more complex.
The case seems not a normal case, we can consider to refine it on master branch later.
thanks for review.

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16907/4/execution/node/711/log

src/object/srv_ec_aggregate.c
```diff
@@ -1085,6 +1085,7 @@ agg_update_parity(struct ec_agg_entry *entry, uint8_t *bit_map,
 			goto out;
 		while (!isset(bit_map, j))
 			j++;
+		D_ASSERTF(j < k, "bad cell_idx %d, number of data shards %d\n", j, k);
```
Nasf-Fan, line 1088: Strictly speaking, if really out of boundary, above check cannot work. We may have to check `j` before `++`, means:
```
while (j < k && !isset(bit_map, j))
    j++;
D_ASSERTF(j < k, "bad cell_idx %d, number of data shards %d\n", j, k);
```
Anyway, we can adjust it when have other chance to refresh the patch.

Nasf-Fan (approved):

src/vos/vos_io.c
```diff
@@ -1104,14 +1105,44 @@ akey_fetch_recx(daos_handle_t toh, const daos_epoch_range_t *epr,
 			holes += lo - index;
 		}
 
-		/* Hole extent, with_shadow case only used for EC obj */
-		if (bio_addr_is_hole(&ent->en_addr) ||
-		    (with_shadow && (ent->en_epoch < shadow_ep))) {
+		if ((with_shadow && (ent->en_epoch < shadow_ep))) {
```
Nasf-Fan, line 1108: If `shadow_ep` is newer, is it still hole? As my understand, under such case, current `ent` should be ignored since `shadow` will cover this `ent`, right?

So here, you may want express: `ent->en_epoch > shadow_ep` ?

On the other hand, we also need to consider the minor epoch if the major epoch are the same.

liuxuezhao: as discussed offline, need not consider minor epoch now. thanks.

Nasf-Fan: Discussed with Xuezhao, the logic can work.

Nasf-Fan (approved):

src/object/srv_ec_aggregate.c
liuxuezhao: right, can change it when need to refresh it later, or by other PR, thanks

PR #16922, opened 2025-09-29
DAOS-18018 test: recovery/pool_membership.py fails on MD-on-SSD cluster

shimizukko:
The test expects that the mount points are removed when system is stopped on MD-on-SSD cluster. However, the test intermittently fails because after the system is stopped at step 3, the mount points aren't removed. Thus, it causes some unexpected behavior and fails.

The solution is to add sleep after dmg system stop.

Skip-unit-tests: true
Test-tag: test_dangling_rank_entry
Test-repeat: 6

### Steps for the author:


daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16922/3/execution/node/797/log

src/tests/ftest/recovery/pool_membership.py
```diff
@@ -333,15 +333,18 @@ def test_dangling_rank_entry(self):
...
+                hosts=self.hostlist_servers, filename=pool_directory, directory=True)
+            if not pool_directory_result[0]:
+                msg = ("MD-on-SSD cluster. Contents under mount point are removed by "
+                       "control plane after system stop.")
+                self.log.info(msg)
+                dmg_command.system_start()
+                # return results in PASS.
+                return
+            self.log.info("Pool dir exists under mount. Wait 3 sec. count = %s", count)
+            time.sleep(3)
```
daltonbohning, line 347: What causes this to fail if we check 5 times and it's still there?

daltonbohning: Do we need to fail here if the dir still exists after 5 tries?

shimizukko:
I'm updating pool_membership.py tests to support MD-on-SSD, so don't need this PR anymore.

PR #16927, opened 2025-09-29
DAOS-18023 tests: Add rebuild scenario to interop testing under ftest

saurabhtandan:
Adding rebuild for interop test
Trigger rebuild after server/client upgrade
Wait for rebuild to complete
Perform read/write
Reintegrate the ranks
Wait for rebuild to complete again.

Skip-test: true

Signed-off-by: Saurabh Tandan <saurabh.tandan@hpe.com>

### Steps for the author:

### DIFF
--- a/src/tests/ftest/interoperability/upgrade_downgrade_base.py
+++ b/src/tests/ftest/interoperability/upgrade_downgrade_base.py
@@ -422,13 +422,12 @@ class UpgradeDowngradeBase(IorTestBase):
     
         self.log.info("Successfully handled excluded ranks: %s and rebuild completed", ranks)
     
-    def pool_upgrade(self, pool, with_fault=False, with_rebuild=False):
+    def pool_upgrade(self, pool, with_fault=False):
         """Run and verify dmg pool upgrade.
     
         Args:
             pool (TestPool): pool to be upgraded
             with_fault (bool): whether to use and verify fault injection
-            with_rebuild (bool): whether to trigger and wait for rebuild after upgrade
         """
         # Make sure an upgrade isn't already in progress
         self.wait_for_pool_upgrade(pool, ["not started", "completed"], fail_on_status=["failed"])
@@ -469,11 +468,6 @@ class UpgradeDowngradeBase(IorTestBase):
                 'Expected pool_layout_ver to increase. '
                 f'pre={pre_pool_layout_ver}, post={post_pool_layout_ver}')
         # TODO verify they are EQUAL
-    
-        # Trigger rebuild only if with_fault is True and with_rebuild is True
-        if with_fault and with_rebuild:
-            self.log.info("Triggering rebuild as with_fault and with_rebuild are both set to True")
-            self.trigger_and_wait_for_rebuild(pool, rank="2")
 
     def verify_write_read(self, clients, container, write=True, read=True):
         """Verify write and/or read with IOR.
@@ -640,6 +634,9 @@ class UpgradeDowngradeBase(IorTestBase):
 
         self.show_daos_version(all_hosts, hosts_client)
 
+        self.log_step("Trigger and wait for rebuild completion")    
+        self.trigger_and_wait_for_rebuild(pool, rank="2")
+
         self.log_step("Verify client/server communication - new client, new server, old pool")
         self.verify_server_client_comm(pool)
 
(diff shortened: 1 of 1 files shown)


src/tests/ftest/interoperability/upgrade_downgrade_base.py
```diff
@@ -371,12 +371,64 @@ def wait_for_pool_upgrade(self, pool, status, fail_on_status=None, timeout=300):
...
+        self.log.info("Reintegrating ranks: %s into pool %s", ranks, pool.identifier)
+        self.get_dmg_command().pool_reintegrate(pool=pool.identifier, ranks=ranks)
+    
+        # Step 4: Wait for rebuild to complete
+        self.log.info("Waiting for rebuild to complete after reintegration")
+        pool.wait_for_rebuild(False)
+    
+        self.log.info("Successfully handled excluded ranks: %s and rebuild completed", ranks)
+    
+    def pool_upgrade(self, pool, with_fault=False, with_rebuild=False):
```
daltonbohning: I don't think we should tie `dmg pool upgrade` to the rebuild scenario. Those are different components to test so IMO they should be different helper functions.

We should probably also get clarification on whether we need to test rebuild before and/or after pool upgrade

saurabhtandan: Indeed. in my other test I did it after server/client upgrades as per the suggestions.
I have made the same changes here. Please review as per convenience.

src/tests/ftest/interoperability/upgrade_downgrade_base.py
```diff
@@ -418,8 +469,6 @@ def pool_upgrade(self, pool, with_fault=False):
                 f'pre={pre_pool_layout_ver}, post={post_pool_layout_ver}')
         # TODO verify they are EQUAL
 
-        # TODO also check [rebuild][state]
-
```
daltonbohning, line 424: Please keep this comment. It still needs to be addressed with respect to pool upgrade

src/tests/ftest/interoperability/upgrade_downgrade_base.py
```diff
@@ -371,9 +371,60 @@ def wait_for_pool_upgrade(self, pool, status, fail_on_status=None, timeout=300):
...
+        self.log.info("Clearing exclusion for ranks: %s", ranks)
+        self.get_dmg_command().system_clear_exclude(ranks=ranks)
+    
+        # Step 2: Start the specified ranks
+        self.log.info("Starting ranks: %s", ranks)
+        self.get_dmg_command().system_start(ranks=ranks)
+    
+        # Step 3: Reintegrate the ranks into the pool
+        self.log.info("Reintegrating ranks: %s into pool %s", ranks, pool.identifier)
+        self.get_dmg_command().pool_reintegrate(pool=pool.identifier, ranks=ranks)
```
daltonbohning: FYI
```suggestion
       pool.reintegrate(ranks=ranks)
```

src/tests/ftest/interoperability/upgrade_downgrade_base.py
```diff
@@ -585,6 +634,9 @@ def upgrade_and_downgrade(self, fault_on_pool_upgrade=False):
 
         self.show_daos_version(all_hosts, hosts_client)
 
+        self.log_step("Trigger and wait for rebuild completion")
```
daltonbohning: I think we should be a little more specific here and also we should use a random rank instead of hardcoding 2. Something like this?
```suggestion
        self.log_step("Exclude 1 random rank and wait for rebuild completion")
        exclude_rank = self.random.choice(list(self.server_managers[0].ranks.keys()))
```

src/tests/ftest/interoperability/upgrade_downgrade_base.py
```diff
@@ -602,6 +654,10 @@ def upgrade_and_downgrade(self, fault_on_pool_upgrade=False):
         self.verify_write_read(hosts_client, tmp_cont, write=True, read=True)
         tmp_cont.destroy()
 
+        self.log_step("Handle excluded ranks and wait for rebuild")
```
daltonbohning: Similarly, instead of saying "handle" we should say what we are doing
```suggestion
        self.log_step("Start and reintegrate excluded ranks and wait for rebuild")
```

daltonbohning (commented):
There are a lot of lint errors that need to be addressed

PR #16943, opened 2025-10-03
DAOS-18017 build: update mercury to 2.4.1

soumagne:
remove libfabric dependency on DAOS rpm

### Steps for the author:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,7 @@
 
 // To use a test branch (i.e. PR) until it lands to master
 // I.e. for testing library changes
-@Library(value='pipeline-lib@soumagne/mercury_libfabric') _
+//@Library(value='pipeline-lib@your_branch') _
 
 /* groovylint-disable-next-line CompileStatic */
 job_status_internal = [:]


daosbuild3:
Test stage Unit Test on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/3/testReport/

daosbuild3:
Test stage Unit Test with memcheck on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/3/testReport/

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/5/testReport/

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/5/execution/node/1032/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/6/testReport/

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/6/execution/node/1005/log

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/6/execution/node/1029/log

daosbuild3:
Test stage Build on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/7/execution/node/296/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/11/testReport/

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/13/execution/node/959/log

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/13/execution/node/1011/log

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/15/execution/node/930/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/16/testReport/

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/16/execution/node/974/log

Jenkinsfile
```diff
@@ -18,7 +18,7 @@
 
 // To use a test branch (i.e. PR) until it lands to master
 // I.e. for testing library changes
-//@Library(value='pipeline-lib@your_branch') _
+@Library(value='pipeline-lib@soumagne/mercury_libfabric') _
```
soumagne: @phender this is the best and easiest solution I think for now (using the PR https://github.com/daos-stack/pipeline-lib/pull/480), this way there is only minimal changes required to the Jenkinsfiles.

utils/rpms/daos.sh
```diff
@@ -62,7 +62,7 @@ install_list+=("${tmp}${sysconfdir}/daos/certs=${sysconfdir}/daos")
 
 EXTRA_OPTS+=("--rpm-attr" "0755,root,root:${sysconfdir}/daos/certs")
 
-DEPENDS=( "mercury >= ${mercury_full}" "${libfabric_lib} >= ${libfabric_full}" )
+DEPENDS=( "mercury >= ${mercury_version}" )
```
daltonbohning, line 71: DAOS no longer depends on libfabric directly? So mercury is the only access to libfabric?

utils/rpms/daos.spec
```diff
@@ -659,6 +651,10 @@ fi
 %endif
 
 %changelog
+* Mon Oct 27 2025  Jerome Soumagne <jerome.soumagne@hpe.com> 2.7.101-17
+- Drop libfabric-devel build requirement
+- Drop libfabric requirement that is already provided by mercury-libfabric
+
```
daltonbohning: FYI this needs to be moved to daos.changelog instead, per https://github.com/daos-stack/daos/commit/abfa0bc2271ea2b56049929cc294a7bffacdba48

utils/rpms/mercury.sh
```diff
@@ -36,14 +37,17 @@ list_files files "${SL_MERCURY_PREFIX}/lib64/lib*.so.*"
...
+
 TARGET_PATH="${libdir}/mercury"
 list_files files "${SL_MERCURY_PREFIX}/lib64/mercury/libna_plugin_ofi.so"
 clean_bin "${files[@]}"
 append_install_list "${files[@]}"
 
 ARCH="${isa}"
-DEPENDS=("${libfabric_lib} >= ${libfabric_version}")
-build_package "mercury"
+DEPENDS=("${libfabric_lib} >= 1.20")
```
daltonbohning: This says 1.20, but package_info.sh still has 1.22?

Jenkinsfile
daltonbohning: Reminder to ourselves that we need to land the pipeline-lib change first

utils/rpms/daos.sh
soumagne: that's right, only the mercury plugin does. The idea is that we need to be able to install DAOS on systems that only have UCX so we don't want to always a pull a libfabric dependency (DAOS itself does not need it anymore anyway).

utils/rpms/daos.spec
soumagne: thanks looks like I missed that one...

utils/rpms/mercury.sh
soumagne: right, this is a different version, maybe I could define a new env that says min_version instead. The min version is different from the version that we ship. You want to be able to give some room when generating rpms for folks that do not necessarily have the latest libfabric (and it's not always possible to use the one that we ship).

utils/rpms/daos.spec
soumagne: done

utils/rpms/mercury.sh
soumagne: I added a separate env to define the min version

Jenkinsfile
soumagne: to be frank, I'm not completely sure of the order that should go in, if I look at the comment it says:
```
// The @library line needs to be edited in a PR for adding a test.
// Adding a test requires two PRs.  One to add the test.
// That PR should be landed with out deleting the PR branch.
// Then a second PR submitted to comment out the @Library line, and when it
// is landed, both PR branches can be deleted.
```
so in theory I think we want to land that PR first so that it uses the branch when it lands, then land the pipeline-lib PR without deleting the branch, and finally merge a new PR that comments out that line again so it uses the default branch. That's how I see it but maybe I'm missing something?

daltonbohning: Usually we land the pipeline-lib PR (and delete the branch). Then update this daos PR back to the default / uncommented pipeline-lib. Then land this PR if testing is okay

soumagne: well imo if you do that then you are going to break all PRs / CIs until this PR has landed :)

soumagne: but maybe not :)

daltonbohning: I see. So the problem is that this PR and the pipeline-lib PR are co-dependent and we can't land either without breaking CI.
https://github.com/daos-stack/pipeline-lib/pull/480/files
And if we were able to implement John's suggestion this wouldn't be a problem:
> This type of logic should not be in pipeline-lib. It should be handled as a parameter or a script / data file in the DAOS job that is causing the stage to be run.

And I don't think it's a good idea to land this daos PR with a reference to a non-master pipeline-lib branch so IMO the ideal is to have the daos Jenkinsfile control the package as John suggests. Do you know how much work that would be?

soumagne: John's idea is unfortunately not great and would require an update to all Jenkinsfiles, including the ones used for daily and weekly testing. This has to be treated the same way we treat mercury-ucx or we should rework our Jenkins pipeline to no longer parse the functional hardware stages arguments this way.

soumagne: And to be clear I completely agree with him that this type of thing should be handled outside of pipeline-lib but this is the way it is handled currently so we should be consistent with mercury-ucx and the way we parse providers when running functional tests. We can either change one line in pipeline-lib that defaults to mercury-libfabric or change 30 different places for each Jenkinsfile to add that additional rpm.

daltonbohning: Another problem is that if landing the pipeline-lib PR breaks CI without this daos PR, then all inflight PRs will be broken until they merge latest master to pick up the changes on this PR. And furthermore, release and feature branches would also be broken until the changes are picked up.
So the pipeline-lib PR needs to be backwards compatible in some way.

soumagne: that's why this PR needs to be merged first imo :) this way you get the new rpms picked up by all branches

daltonbohning: But if this PR lands, then wouldn't in-flight daos PRs still try to install `mercury` and `libfabric`? And then landing the pipeline-lib PR means daos PRs try to ALSO install `mercury-libfabric`?

soumagne: I'm not sure what in-flight really means, for sure PRs will have to be updated / rebased if we merge this PR first. If trying to install `mercury-libfabric` when it does not exist does not create any failures maybe it's fine ? that's why I said "maybe not" in my earlier comment. Either way there might be some potential issue that will resolve on itself once everything lands. At the same time, given the state of our CI I'm not that concerned if I break it for 2 minutes as long as we are aware of it :)

daltonbohning: The pipeline-lib PR is failing, which means it does break CI without this PR.
And I'm not confident CI would only be broken for a few minutes.
By "in-flight" I mean that when this PR lands, all other PRs do not automatically pick up the changes. Sure, the new RPM is uploaded and available, but that doesn't mean all other PRs work. And if all PRs have to merge master, then all PRs are broken

soumagne: When this PR lands, it will update the mercury rpm package which will then be picked up by all branches so all current test PRs will fail if they don't install mercury-libfabric. I think that's just inevitable, given the way we globally propagate rpm updates. And this is true in both cases, if the change is limited to that PR or also to pipeline-lib. In my opinion the pipeline-lib change remains the cleanest and easiest solution, I'm open to better ideas though.

daltonbohning: Ok, I misunderstood what this PR was doing. I did some testing locally and for example trying to install this new mercury with 2.6.4 works as expected
```
sudo dnf install mercury-2.4.1~rc5-1.el8 daos-2.6.4
...
Installing:
 daos     2.6.4-7.3224.g8a20f44f.el8
 mercury  2.4.1~rc5-1.el8
```

So, PRs that do NOT have this change will be okay because their version of `daos` still depends on `mercury` and on `libfabric`.
Then, when pipeline-lib says to install `mercury-libfabric` it doesn't actually matter for those PRs because `mercury-libfabric` really just installs `mercury` and `libfabric`.

daltonbohning: So I see two options. Either what you mentioned about leaving the pipeline-lib reference in the landing of this PR, OR something like this:
1. Remove the pipeline-lib reference in this PR.
2. Keep the `daos -> libfabric` dependency for now for compatibility
3. Land this PR, since with (1) and (2) it should still work.
4. Land the pipeline-lib PR
5. New daos PR to remove the `daos -> libfabric` dependency

IMO I prefer this approach so we don't have to leave the pipeline-lib reference in place

soumagne: > PRs that do NOT have this change will be okay

you are allowed to install mercury without mercury-libfabric or mercury-ucx but all tests will fail because the providers / plugins won't be found. What this PR does is separate the libfabric mercury plugin that allows mercury to interface with libfabric. Without it, mercury cannot use libfabric.

> Keep the daos -> libfabric dependency for now for compatibility

No, this does not help in any way and no with (1) and (2) it won't work because you do need mercury-libfabric. This is not a simple alias of some sort, to expand on what I just mentioned above,  it installs a separate library that is similar to mercury-ucx (you can look at that as an example). This library is then dynamically  loaded by mercury.

daltonbohning: Then I don't know how to land any of this without breaking CI

soumagne: Well that's what I've just been trying to explain :) I'm glad we're reaching the same conclusion. The real issue is that rpms are globally visible once merged, so because of that, any change of that sort will face the same issues.

soumagne:
@phender @daltonbohning please review this PR. I'm also happy to discuss offline to walk you through it and think about better solutions if there are any. It is passing the CI and I will need to get that merged as there are a number of fixes we need to pick for customers.
For background you can also read more about the need for separating mercury-libfabric in that ticket: https://daosio.atlassian.net/browse/DAOS-13244

soumagne:
From discussion with @mchaarawi since this PR will affect 2.6 also, should get 2.6 merge approval before moving further.

jolivier23 (approved):

frostedcmos (approved):

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/24/testReport/

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/25/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16943/25/execution/node/1427/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/26/testReport/

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16943/26/display/redirect

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16943/26/testReport/

soumagne:
PRs should be merged in the following order:
- This PR first
- Next, the pipeline-lib PR: https://github.com/daos-stack/pipeline-lib/pull/480
- Finally the 2.6 PR: https://github.com/daos-stack/daos/pull/17221

Jenkinsfile
jolivier23: should this be reverted and land both at the same time?

jolivier23 (approved):

Jenkinsfile
soumagne: I think to avoid breaking master's CI, it is best to land it with that change, land the pipeline-lib change, and then revert that change. Or one could also land the pipeline-lib change, "break" the master CI, and land that PR after :) to be frank, yes I think you're right it may not make a lot of differences if everything is done at the same time :)

soumagne: thinking more about it, I think I like your suggestion and just before this gets merged, we could land the pipeline-lib change, push a patch to revert that change and land this PR immediately after to not have to open another PR to revert something that is temporary.

jolivier23: Probably when @daos-stack/daos-gatekeeper lands the pipeline-lib change they will then want to avoid deleting the branch (or undeleting it right after landing as it gets deleted automatically) and then delete it after making the subsequent change in daos master branch.

ryon-jensen: Agree that landing the pipeline-lib change first is probably the better approach.

PR #16945, opened 2025-10-05
DAOS-17772 rebuild: avoid potential races between rebuild and EC aggregation

gnailzenh:
- Increment the map_version for reclaim tasks to ensure they operate on the latest
   pool map. Without this, reclaim tasks might act on an outdated pool map, potentially
   causing data corruption.

- Always recalculate parity for partial overwrites, disabling the optimized path to
   ensure data consistency.

- Ignore parity update errors during EC aggregation. This is a temporary workaround
   because DAOS does not use DTX to update parities during EC aggregation.
   Without DTX, parity shards may become inconsistent if an engine failure occurs
   alongside a rebuild. This approach significantly increases the likelihood of parity
   shards remaining consistent when EC aggregation interacts with rebuild operations.

- Standardize the parity target status checks for both full stripe and hole processing
   during EC aggregation.

- Introduce additional barriers to minimize the chances of overlapping EC aggregation
   and rebuild operations.

### Steps for the author:

### DIFF
--- a/src/object/srv_ec_aggregate.c
+++ b/src/object/srv_ec_aggregate.c
@@ -1224,6 +1224,7 @@ agg_process_partial_stripe(struct ec_agg_entry *entry)
 				    entry->ae_cur_stripe.as_stripenum,
 				    &full_cell_cnt);
 
+	has_old_replicas = true; /* XXX force full recalculation */
 	if (full_cell_cnt >= k / 2 || cell_cnt == k || has_old_replicas) {
 		stripe_ud.asu_recalc = true;
 		cell_cnt = full_cell_cnt;
--- a/src/object/srv_ec_aggregate.c
+++ b/src/object/srv_ec_aggregate.c
@@ -1352,17 +1352,18 @@ agg_peer_update_ult(void *arg)
 	uint32_t		 peer, peer_shard;
 	struct dc_object	*obj = NULL;
 	crt_rpc_t		*rpc = NULL;
-	int			 ult_rc = 0, rc = 0;
+	int			 peer_rc = 0, rc = 0;
+	int			 peer_updated = 0;
 	uint64_t		 enqueue_id;
 	uint32_t		 max_delay = 0;
 
 	if (unlikely(DAOS_FAIL_CHECK(DAOS_FORCE_EC_AGG_PEER_FAIL)))
-		D_GOTO(out, ult_rc = -DER_TIMEDOUT);
+		D_GOTO(out, rc = -DER_TIMEDOUT);
 
 	agg_param = container_of(entry, struct ec_agg_param, ap_agg_entry);
-	ult_rc = agg_peer_check_avail(agg_param, entry);
-	if (ult_rc != 0)
-		D_GOTO(out, ult_rc);
+	rc = agg_peer_check_avail(agg_param, entry);
+	if (rc != 0)
+		D_GOTO(out, rc);
 
 	csummer = ec_agg_param2csummer(agg_param);
 	iod.iod_type = DAOS_IOD_ARRAY;
@@ -1383,7 +1384,7 @@ agg_peer_update_ult(void *arg)
 			D_ERROR(DF_UOID" pidx %d to peer %d, rank %d tag %d obj_req_create "
 				DF_RC"\n", DP_UOID(entry->ae_oid), pidx, peer,
 				tgt_ep.ep_rank, tgt_ep.ep_tag, DP_RC(rc));
-			ult_rc = ult_rc ?: rc;
+			peer_rc = peer_rc ?: rc;
 			/* XXX no way to rollback parity writes today, so just continue */
 			continue;
 		}
@@ -1459,6 +1460,7 @@ agg_peer_update_ult(void *arg)
 			D_CDEBUG(rc == 0, DB_TRACE, DLOG_ERR,
 				 "update parity[%d] to %d:%d, status = "DF_RC"\n",
 				 peer, tgt_ep.ep_rank, tgt_ep.ep_tag, DP_RC(rc));
+			peer_updated += rc == 0;
 		}
 	next:
 		if (bulk_hdl)
@@ -1474,12 +1476,16 @@ agg_peer_update_ult(void *arg)
 			dss_sleep(daos_rpc_rand_delay(max_delay) << 10);
 			goto retry;
 		}
-		ult_rc = ult_rc ?: rc;
+		peer_rc = peer_rc ?: rc;
 	}
 out:
 	if (obj)
 		obj_decref(obj);
-	ABT_eventual_set(stripe_ud->asu_eventual, (void *)&ult_rc, sizeof(ult_rc));
+	/* XXX: before switching to DTX, any successful parity write is deemed as
+	 * success of all parity shards.
+	 */
+	rc = peer_updated > 0 ? 0 : peer_rc;
+	ABT_eventual_set(stripe_ud->asu_eventual, (void *)&rc, sizeof(rc));
 }
 
 /* Invokes helper function to send the generated parity and the stripe number
@@ -1552,7 +1558,8 @@ agg_process_holes_ult(void *arg)
 	uint64_t		 ext_tot_len = 0;
 	uint32_t		 pidx = ec_age2pidx(entry);
 	uint32_t		 peer;
-	int			 rc = 0, ult_rc = 0;
+	int			 rc = 0, peer_rc = 0;
+	int			 peer_updated = 0;
 	uint32_t		 max_delay = 0;
 	uint64_t		 enqueue_id;
 
@@ -1581,8 +1588,8 @@ agg_process_holes_ult(void *arg)
 		goto out;
 
 	agg_param = container_of(entry, struct ec_agg_param, ap_agg_entry);
-	ult_rc = agg_peer_check_avail(agg_param, entry);
-	if (ult_rc != 0)
+	rc = agg_peer_check_avail(agg_param, entry);
+	if (rc != 0)
 		goto out;
 
 	obj = obj_hdl2ptr(entry->ae_obj_hdl);
@@ -1610,7 +1617,7 @@ agg_process_holes_ult(void *arg)
 				   NULL, DIOF_FOR_EC_AGG, NULL, NULL);
 		if (rc) {
 			D_ERROR("dsc_obj_fetch failed: "DF_RC"\n", DP_RC(rc));
-			D_GOTO(out, ult_rc = rc);
+			D_GOTO(out, rc);
 			goto out;
 		}
 
@@ -1619,7 +1626,7 @@ agg_process_holes_ult(void *arg)
 					    NULL, -1, &stripe_ud->asu_iod_csums);
 		if (rc != 0) {
 			D_ERROR("setting up iods csums failed: "DF_RC"\n", DP_RC(rc));
-			D_GOTO(out, ult_rc = rc);
+			D_GOTO(out, rc);
 			goto out;
 		}
 	}
@@ -1633,7 +1640,7 @@ agg_process_holes_ult(void *arg)
 		if (rc) {
 			D_ERROR(DF_UOID" crt_bulk_create returned: "DF_RC"\n",
 				DP_UOID(entry->ae_oid), DP_RC(rc));
-			D_GOTO(out, ult_rc = rc);
+			D_GOTO(out, rc);
 			goto out;
 		}
 	}
@@ -1655,7 +1662,7 @@ agg_process_holes_ult(void *arg)
 		if (rc) {
 			D_ERROR(DF_UOID" obj_req_create failed: "DF_RC"\n",
 				DP_UOID(entry->ae_oid), DP_RC(rc));
-			ult_rc = ult_rc ?: rc;
+			peer_rc = peer_rc ?: rc;
 			continue;
 		}
 		ec_rep_in = crt_req_get(rpc);
@@ -1693,6 +1700,7 @@ agg_process_holes_ult(void *arg)
 			D_CDEBUG(rc == 0, DB_TRACE, DLOG_ERR,
 				 DF_UOID" parity[%d] er_status = "DF_RC"\n",
 				 DP_UOID(entry->ae_oid), peer, DP_RC(rc));
+			peer_updated += rc == 0;
 		}
 		crt_req_decref(rpc);
 		rpc = NULL;
@@ -1700,7 +1708,7 @@ agg_process_holes_ult(void *arg)
 			dss_sleep(daos_rpc_rand_delay(max_delay) << 10);
 			goto retry;
 		}
-		ult_rc = ult_rc ?: rc;
+		peer_rc = peer_rc ?: rc;
 	}
 
 out:
@@ -1709,7 +1717,11 @@ out:
 	if (obj)
 		obj_decref(obj);
 	entry->ae_sgl.sg_nr = AGG_IOV_CNT;
-	ABT_eventual_set(stripe_ud->asu_eventual, (void *)&ult_rc, sizeof(ult_rc));
+	/* XXX: before switching to DTX, any successful parity write is deemed as
+	 * success of all parity shards.
+	 */
+	rc = peer_updated > 0 ? 0 : peer_rc;
+	ABT_eventual_set(stripe_ud->asu_eventual, (void *)&rc, sizeof(rc));
 }
 
 static int
--- a/src/object/srv_ec_aggregate.c
+++ b/src/object/srv_ec_aggregate.c
@@ -1356,7 +1356,6 @@ agg_peer_update_ult(void *arg)
 	crt_rpc_t		*rpc = NULL;
 	int                      peer_rc = 0, rc = 0;
 	int                      peer_updated = 0;
-	uint64_t		 enqueue_id;
 	uint32_t		 max_delay = 0;
 
 	if (unlikely(DAOS_FAIL_CHECK(DAOS_FORCE_EC_AGG_PEER_FAIL)))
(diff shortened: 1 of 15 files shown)


src/cart/crt_rpc.h
```diff
@@ -511,7 +511,7 @@ CRT_RPC_DECLARE(crt_iv_fetch, CRT_ISEQ_IV_FETCH, CRT_OSEQ_IV_FETCH)
 	((uint32_t)		(padding)		CRT_VAR)
 
 #define CRT_OSEQ_IV_UPDATE	/* output fields */		 \
-	((uint64_t)		(rc)			CRT_VAR) \
+	((uint64_t)		(ivo_rc)		CRT_VAR) \
```
frostedcmos, line 514: what is 'o' here? should be ivu_rc (iv_update), no?

frostedcmos (commented):
Cart change looks fine to me except for ivo_* naming as it should be ivu_ instead

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16945/2/execution/node/1522/log

src/cart/crt_rpc.h
gnailzenh: thanks Alex, I will let Xuezhao know, this PR is just for testing.

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-16945/5/testReport/

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16945/5/display/redirect

daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-16945/5/display/redirect

src/cart/crt_rpc.h
liuxuezhao: it is just because took the same prefix with the other original field "ivo_iv_sgl".
"u" or "o" looks either is fine as far as it is unique and easy to grep, "o" probably means output.
@frostedcmos

src/object/srv_ec_aggregate.c
```diff
@@ -837,21 +838,16 @@ agg_update_vos(struct ec_agg_param *agg_param, struct ec_agg_entry *entry,
...
-					   &ext->ae_recx);
-		if (err)
-			D_ERROR("array_remove fails: "DF_RC"\n", DP_RC(err));
-		if (!rc && err)
-			rc = err;
-	}
+	recx.rx_idx = entry->ae_cur_stripe.as_stripenum * ec_age2ss(entry);
+	recx.rx_nr  = ec_age2ss(entry);
+	epoch_range.epr_lo = ap->ap_epr.epr_lo;
+	epoch_range.epr_hi = entry->ae_cur_stripe.as_hi_epoch;
```
liuxuezhao, line 843: maybe this is the idea change? @NiuYawei 
original code use "epoch_range.epr_lo = epoch_range.epr_hi = ext->ae_epoch;" supose that is incorrect?

NiuYawei: AFAIK, the original code is to delete the partial data extents in 'entry-><ae_cur_stripe.as_dextents' one by one, so the "epoch_range.epr_lo = epoch_range.epr_hi = ext->ae_epoch;" looks correct to me.

Here we changed it to delete all partial data extents within an epoch range. If the data extents within this epoch range are exactly same to the extents in 'entry->ae_cur_stripe.as_dextents', then I think the outcome should be same.

Is there a defect in original code?

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16945/7/execution/node/1381/log

daosbuild3:
Test stage Build on Leap 15.5 with Intel-C and TARGET_PREFIX completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16945/8/execution/node/439/log

src/object/srv_ec_aggregate.c
liuxuezhao: seems the changed code is better?
consider the following case:
cell size 1MB, EC4P1
with such data -
[0-4M]@epoch_100 (full-stripe)
[0-4K]@epoch_110, [0-2K]@epoch_120, [2K-4K]@epoch_130

in EC agg, it will see [0-2K]@epoch_120, [2K-4K]@epoch_130 and [0-4M]@epoch_100, and cannot see [0-4K]@epoch_110.
So original code cannot remove the [0-4K]@epoch_110.
Commonly it looks will not cause problem as the EC agg will generate new parity ext @epoch_130,
but leaving the [0-4K]@epoch_110 there seems still possible cause problem, for example rebuild possibly will rebuild it at a higher epoch (see __migrate_fetch_update_parity will rebuild the replica ext using a high epoch) that possibly will be higher than the parity ext epoch_130 and cause corruption. @NiuYawei

NiuYawei: Hmm, it looks to me the vos_obj_array_remove() removes only visible extents, so the covered extent will be missed out no matter what approach is used?

No matter if a covered data extent is missed out or not, I don't think rebuild should blindly rebuild a data extent with epoch lower than parity epoch. That part of rebuild code might need be re-organized in the future. (start rebuilding stripe only after everything within the stripe are prepared).

NiuYawei: Sorry, I read the code closer this time, looks vos_obj_array_remove() removes both visible & covered extents. (I overlooked that evt_find_visible() does special processing for remove caller). So this fix should be able to fix some potential issues as you described.

gnailzenh: I think it's problematic to delete them one by one, because enumerate all return visible part of those extents, it means we will punch visible parts of those extents even if there are overlaps? In addition, I feel this is a more efficient way to delete, and we do this on remote parity shards anyway (please see ds_obj_ec_agg_handler)

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16945/10/execution/node/1412/log

wangshilong (dismissed):

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16945/11/execution/node/1394/log

src/object/srv_ec_aggregate.c
```diff
@@ -1331,14 +1377,16 @@ agg_peer_update_ult(void *arg)
...
 		rc = obj_req_create(dss_get_module_info()->dmi_ctx, &tgt_ep,
 				    DAOS_OBJ_RPC_EC_AGGREGATE, &rpc);
 		if (rc) {
 			D_ERROR(DF_UOID" pidx %d to peer %d, rank %d tag %d obj_req_create "
 				DF_RC"\n", DP_UOID(entry->ae_oid), pidx, peer,
 				tgt_ep.ep_rank, tgt_ep.ep_tag, DP_RC(rc));
-			goto out;
+			peer_rc = peer_rc ?: rc;
+			/* XXX no way to rollback parity writes today, so just continue */
+			continue;
```
liuxuezhao: I'm thinking that if no any peer updated yet (peer_updated == 0), for this kind of failure maybe better to fail the peer update rather than ignore it?

src/object/srv_ec_aggregate.c
```diff
@@ -1379,26 +1427,21 @@ agg_peer_update_ult(void *arg)
...
 					     &sgl, CRT_BULK_RW, &bulk_hdl);
 			if (rc) {
-				D_ERROR(DF_UOID" pidx %d to peer %d, "
-					"crt_bulk_create "DF_RC"\n",
-					DP_UOID(entry->ae_oid), pidx, peer,
-					DP_RC(rc));
-				goto out_rpc;
+				D_ERROR(DF_UOID" pidx %d to peer %d, crt_bulk_create "DF_RC"\n",
+					DP_UOID(entry->ae_oid), pidx, peer, DP_RC(rc));
+				goto next;
```
liuxuezhao, line 1444: same, if no any peer updated yet (peer_updated == 0), for this kind of failure maybe better to fail the peer update rather than ignore it?

src/object/srv_ec_aggregate.c
```diff
@@ -1379,26 +1427,21 @@ agg_peer_update_ult(void *arg)
...
-							    &iod_csums);
+				rc = daos_csummer_calc_iods(csummer, &sgl, &iod, NULL, 1, false,
+							    NULL, 0, &iod_csums);
 				if (rc) {
-					D_ERROR("daos_csummer_calc_iods "
-						"failed: "DF_RC"\n", DP_RC(rc));
-					goto out;
+					D_ERROR("daos_csummer_calc_iods failed: "DF_RC"\n",
+						DP_RC(rc));
+					goto next;
```
liuxuezhao, line 1454: same, if no any peer updated yet (peer_updated == 0), for this kind of failure maybe better to fail the peer update rather than ignore it?

src/object/srv_ec_aggregate.c
```diff
@@ -1407,46 +1450,42 @@ agg_peer_update_ult(void *arg)
...
-				DF_RC"\n", DP_UOID(entry->ae_oid), pidx, peer,
-				DP_RC(rc));
-			goto out_bulk;
+			D_ERROR(DF_UOID" pidx %d to parity[%d], dss_rpc_send "
+				DF_RC"\n", DP_UOID(entry->ae_oid), pidx, peer, DP_RC(rc));
+		} else {
+			crt_req_get_timeout(rpc, &max_delay);
+			ec_agg_out = crt_reply_get(rpc);
+			rc = ec_agg_out->ea_status;
+			enqueue_id = ec_agg_out->ea_comm_out.req_out_enqueue_id;
```
liuxuezhao: not quite follow why set enqueue_id here?
for example RPC to peer_1 succeed, then why use the enqueue_id replied from peer_1 for following peer_2's update? may need confirm with @wangshilong ?

the original code seems also not quite correct, in the case of peer_1 update get DER_OVERLOAD_RETRY, then will use the enqueue_id replied from peer_1 to retry update for peer_1, and also for following's update for peer_2? @wangshilong

src/object/srv_ec_aggregate.c
```diff
@@ -1685,7 +1663,8 @@ agg_process_holes_ult(void *arg)
 		if (rc) {
 			D_ERROR(DF_UOID" obj_req_create failed: "DF_RC"\n",
 				DP_UOID(entry->ae_oid), DP_RC(rc));
-			goto out;
+			peer_rc = peer_rc ?: rc;
+			continue;
```
liuxuezhao: same as the peer update ult

src/object/srv_ec_aggregate.c
```diff
@@ -2787,12 +2774,7 @@ cont_ec_aggregate_cb(struct ds_cont_child *cont, daos_epoch_range_t *epr,
 	agg_clear_extents(&ec_agg_param->ap_agg_entry);
 	agg_reset_entry(&ec_agg_param->ap_agg_entry, NULL, NULL);
 
-	if (daos_handle_is_valid(ec_agg_param->ap_agg_entry.ae_obj_hdl)) {
-		dsc_obj_close(ec_agg_param->ap_agg_entry.ae_obj_hdl);
-		ec_agg_param->ap_agg_entry.ae_obj_hdl = DAOS_HDL_INVAL;
-	}
-
```
liuxuezhao, line 2794: why remove this code section, is it intended?

src/object/srv_ec_aggregate.c
gnailzenh: indeed, it can de-risk the chance of running into intermittent failure but still proceed the aggregation, I can fix this in a upcoming patch.

src/object/srv_ec_aggregate.c
gnailzenh: agreed

src/object/srv_ec_aggregate.c
gnailzenh: agreed

src/object/srv_ec_aggregate.c
gnailzenh: right, we should reset queue_id each time switching to a new peer, good catch

src/object/srv_ec_aggregate.c
gnailzenh: It has been done by reset_entry already

liuxuezhao: i see, agg_reset_entry already closed it, sorry for the noise.

src/object/srv_ec_aggregate.c
gnailzenh: It's actually set to zero in the next loop, so I think it's OK?

src/object/srv_ec_aggregate.c
```diff
@@ -1407,46 +1453,47 @@ agg_peer_update_ult(void *arg)
...
+			ec_agg_out = crt_reply_get(rpc);
+			rc         = ec_agg_out->ea_status;
+			if (rc == -DER_OVERLOAD_RETRY) {
+				enqueue_id = ec_agg_out->ea_comm_out.req_out_enqueue_id;
+				overloaded = true;
+			}
+			D_CDEBUG(rc == 0, DB_TRACE, DLOG_ERR,
+				 "update parity[%d] to %d:%d, status = " DF_RC "\n", peer,
+				 tgt_ep.ep_rank, tgt_ep.ep_tag, DP_RC(rc));
+			peer_updated += rc == 0;
```
liuxuezhao, line 1478: minor, here the rc already checked is 0.

liuxuezhao (dismissed):

wangshilong (dismissed):

src/object/srv_ec_aggregate.c
gnailzenh: rc is overwritten by ea_status in reply right?

liuxuezhao: ah, right. thx

src/object/srv_obj.c
```diff
@@ -3363,8 +3372,14 @@ obj_local_enum(struct obj_io_context *ioc, crt_rpc_t *rpc,
 		goto failed;
 	}
 
-	if (oei->oei_flags & ORF_FOR_MIGRATION)
+	if (oei->oei_flags & ORF_FOR_MIGRATION) {
+		/* just in case ds_pool::sp_rebuilding is not set, pause my local EC aggregation
+		 * by setting this flag.
+		 * NB: it's a lockess write to shared data structure and it's harmless.
+		 */
+		ioc->ioc_coc->sc_pool->spc_pool->sp_rebuild_scan = 1;
```
NiuYawei, line 3380: hmm. it looks not safe to me to have multiple writers for single 32 bits integer. (single writer, multiple readers looks good to me)

I don't quite see the purpose of this 'sp_rebuild_scan'? Why can't we rely on the 'sp_rebuilding' solely?

daosbuild3:
Test stage Functional Hardware Medium completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16945/12/execution/node/1349/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16945/12/execution/node/1479/log

src/object/srv_obj.c
gnailzenh: As we discussed, this is a workaround because adding another barrier is too much works for 2.6, I will address this in 2.8.x. It's safe to overwrite with the same value (1), as long as there is a single writer to set it to (0).

liuxuezhao (approved):

liuxuezhao (dismissed):

wangshilong (dismissed):

src/tests/ftest/util/server_utils_params.py
```diff
@@ -436,6 +436,7 @@ class EngineYamlParameters(YamlParameters):
         "common": [
             "D_LOG_FILE_APPEND_PID=1",
             "DAOS_POOL_RF=4",
+            "DAOS_REBUILD_WAIT_EC_PAUSE=1",
```
daltonbohning, line 439: Do we need to also set this on Aurora, etc.? Why only in CI?

gnailzenh: without this,  all rebuild tests may run longer time, we should not use this on Aurora because we do want to wait for longer time for large scale system.

daltonbohning: Thanks for explaining. I think that's okay then

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16945/15/execution/node/1380/log

liuxuezhao (approved):

wangshilong (approved):

PR #16970, opened 2025-10-13
DAOS-17358 rebuild: rebuilding test in obj_inflight_io_check

kccain:
This change removes the check of incoming client IO with the ORF_REBUILDING_IO flag set.

Before the change, the intent of the check was to temporarily disallow such IO while a rebuild was starting, when the PS leader engine first distributes a fence / epoch value to all engines. The check caused the IO to get an error -DER_UPDATE_AGAIN, causing the client to retry.

With forthcoming features like interactive/explicit rebuild control, the rebuild *stop* case is negatively affected by this test, causing all subsequent client IO after rebuild stops to block indefinitely, until the rebuild is restarted by the administrator.

Other intermittent test failures, unrelated to the new feature, are also being seen. Discussion with other developers have suggested that this test in the engine code is not required.

Allow-unstable-test: true

### Steps for the author:

### DIFF
--- a/src/object/srv_obj.c
+++ b/src/object/srv_obj.c
@@ -2457,19 +2457,6 @@ obj_inflight_io_check(struct ds_cont_child *child, uint32_t opc,
 		return -DER_UPDATE_AGAIN;
 	}
 
-	/* All I/O during rebuilding, needs to wait for the rebuild fence to
-	 * be generated (see rebuild_prepare_one()), which will create a boundary
-	 * for rebuild, so the data after boundary(epoch) should not be rebuilt,
-	 * which otherwise might be written duplicately, which might cause
-	 * the failure in VOS.
-	 */
-	if ((flags & ORF_REBUILDING_IO) &&
-	    (is_pool_rebuild_allowed(child->sc_pool->spc_pool, false) &&
-	     child->sc_pool->spc_rebuild_fence == 0)) {
-		D_ERROR("rebuilding "DF_UUID" retry.\n", DP_UUID(child->sc_pool->spc_uuid));
-		return -DER_UPDATE_AGAIN;
-	}
-
 	return 0;
 }
 


daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-16970/1/execution/node/1421/log

2 engines asserted during test_daos_rebuild_ec execution. I noted the symptoms in existing issue [DAOS-17091](https://daosio.atlassian.net/browse/DAOS-17091) (though I'm not totally sure those failures always involve raft asserts). I believe it is unrelated to this change however.

Rerunning from test hardware stage

kccain:
This PR is ready for review. To be clear, I've organized the extensive testing of this central object IO change across multiple PRs that make this same change:
- (for landing) This PR for "pr" tagged tests: all tests passed after the most recent build, and the one failure from an earlier build documented in the earlier comment
- (test-only) [PR-16971](https://github.com/daos-stack/daos/pull/16971) for "rebuild" tagged tests beyond the "pr" ones (e.g., getting daily_regression and full_regression rebuild tests) - one known test failure documented in that PR (test timeout too low) for the in-development interactive rebuild feature
- (test-only) [PR-16972](https://github.com/daos-stack/daos/pull/16972) for [DAOS-17937](https://daosio.atlassian.net/browse/DAOS-17937) for running all osa/online_drain.py cases (that can suffer from -DER_UPDATE_AGAIN due to this same issue) in a Test-Repeat fashion. a 5x repeat run passed in one build, and a 10x repeat passed in a subsequent build.

liuxuezhao (approved):

wangshilong (approved):
As my understanding originally it tried to fix following issue:

For example, suppose the original data is stored on shard1, shard2, and shard3. If shard1 fails, the data needs to be migrated to shard4. Assume rebuild starts at epoch 10, so we need to rebuild all data before epoch 10. Before the rebuild fence is set, no new writes should be allowed. If new writes are allowed, they may be written to shard2, shard3, and shard4. Then, during the rebuild, the system will think some data needs to be rewritten to shard2/shard3/shard4, 

This might cause problem in vos, as it did not support overwrite with same epoch. but if we changed that, this is probably fine..

src/object/srv_obj.c
```diff
@@ -2457,19 +2457,6 @@ obj_inflight_io_check(struct ds_cont_child *child, uint32_t opc,
...
-	 * the failure in VOS.
-	 */
-	if ((flags & ORF_REBUILDING_IO) &&
-	    (is_pool_rebuild_allowed(child->sc_pool->spc_pool, false) &&
-	     child->sc_pool->spc_rebuild_fence == 0)) {
-		D_ERROR("rebuilding "DF_UUID" retry.\n", DP_UUID(child->sc_pool->spc_uuid));
-		return -DER_UPDATE_AGAIN;
-	}
-
 	return 0;
```
NiuYawei, line 2460: I suppose our ultimate goal is to remove this function entirely?

NiuYawei (approved):

src/object/srv_obj.c
liuxuezhao: can try to remove more later when it is safe, cannot remove more now.

PR #17064, opened 2025-10-31
DAOS-623 build: git commit template

grom72:
The proposed git commit template provides all available pragmas for easy and quick Jenkins build configuration/tuning.
It is always used when the `-m` parameter is not provided to `github commit` command.

Run `git config commit.template ./.gitcommittemplate.txt` to install the template.

### Steps for the author:


.gitcommittemplate.txt
```diff
@@ -0,0 +1,74 @@
+#<ticket> <component>: <title> 
+# Valid components are: 'agent', 'build', 'ci', 'csum', 'doc', 'gha', 'il',
+#                       'md', 'mercury', 'packaging', 'pil4dfs', 'swim',
+#                       'test', 'tools', 'ddb'
+
+# See instruction at the end of the file
```
daltonbohning: What instructions is this referring to?

grom72: Installation instruction.
See c114c84cab8cd67a40e61113304192fab02a9e52

ryon-jensen (dismissed):

.gitcommittemplate.txt
```diff
@@ -0,0 +1,75 @@
+#<ticket> <component>: <title> 
+# Valid components are: 'agent', 'build', 'ci', 'csum', 'doc', 'gha', 'il',
+# 'md', 'mercury', 'packaging', 'pil4dfs', 'swim', 'test', 'tools', 'ddb'
+# Please refer to the installation instruction at the end of the file.
```
daltonbohning: Is this necessary then? Once the template is installed, this line is no longer useful but will show up in every commit.

grom72: It will be ignored because it's a comment (`# ...`).
but an alternative can be to add this to https://github.com/daos-stack/daos/blob/master/docs/dev/contributing.md#commit-comments
see: f9c56f4404a84f802b8978b9d9ca87b26e379b09

ryon-jensen:
I like the idea of this, but a concern that I have is that it creates another place that needs to be maintained when pragmas and components change, which could lead to it becoming stale.  I think it's a great tool, but would it be better simply documented in a wiki and shared with the team? Just a thought.

grom72:
> I like the idea of this, but a concern that I have is that it creates another place that needs to be maintained when pragmas and components change, which could lead to it becoming stale. I think it's a great tool, but would it be better simply documented in a wiki and shared with the team? Just a thought.

I think that the problem is we have too many places where we document Pragmas, Jenkins parameters, etc. And the wiki itself is too far from the source code.
The main purpose of this PR is to provide a simple checklist to disable/enable particular tests in a very easy way.

In the future anyone who modify the default tests configuration (I guess in Jenkins file) should also adjust this template file.

daltonbohning:
> > I like the idea of this, but a concern that I have is that it creates another place that needs to be maintained when pragmas and components change, which could lead to it becoming stale. I think it's a great tool, but would it be better simply documented in a wiki and shared with the team? Just a thought.
> 
> I think that the problem is we have too many places where we document Pragmas, Jenkins parameters, etc. And the wiki itself is too far from the source code. The main purpose of this PR is to provide a simple checklist to disable/enable particular tests in a very easy way.
> 
> In the future anyone who modify the default tests configuration (I guess in Jenkins file) should also adjust this template file.

I think part of the problem is most pragmas are controlled by pipeline-lib. So when they change, presumably there is a pipeline-lib PR, and then someone has to come update this repo as well. Which is in addition to already having to update the other locations.

daltonbohning (approved):
I agree with Ryon's point but since this is optional and I'm not the one personally maintaining it, I'm okay with it.

PR #17130, opened 2025-11-15
DAOS-18170 object: more delay before retry obj rpc

Nasf-Fan:
If an update RPC is timeout, it is unnecessary to retry the update RPC immediately. Because the original one maybe blocked on some of targets under some heave load cases. Under such case, the retried RPC will get -DER_INPROGRESS and cause RPC retry again and again. That will further increase server load and make the situation to be worse.

This patch introduces more delay before retrying RPC under such cases. It also add more delay for collective object RPC.

### Steps for the author:

### DIFF
--- a/src/object/cli_obj.c
+++ b/src/object/cli_obj.c
@@ -1717,35 +1717,43 @@ dc_obj_layout_refresh(daos_handle_t oh)
 }
 
 uint32_t
-dc_obj_retry_delay(tse_task_t *task, uint32_t opc, int err, uint16_t *retry_cnt,
-		   uint16_t *inprogress_cnt, uint32_t timeout_sec)
+dc_obj_retry_delay(tse_task_t *task, uint32_t opc, int err, uint32_t *retry_cnt,
+		   uint32_t timeout_sec, bool long_delay)
 {
 	uint32_t delay = 0;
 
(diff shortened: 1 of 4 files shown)


daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17130/2/execution/node/1291/log

liuxuezhao (approved):

src/object/cli_obj.c
```diff
@@ -1717,35 +1717,43 @@ dc_obj_layout_refresh(daos_handle_t oh)
...
+			case DAOS_OBJ_RPC_CPD:
+				/* 8 times of the delay for compounded RPC. */
+				delay <<= 3;
+				break;
+			default:
+				break;
+			}
+
+			/* Increase delay after multiple times retry. */
+			if (*retry_cnt >= 5)
```
wangshilong, line 1750: Do you think it makes sense to output some warning message if a RPC is retied more than 5 times?

wangshilong (approved):

src/object/cli_obj.c
Nasf-Fan: The caller has the warning message output.

PR #17267, opened 2025-12-12
DAOS-18192 rebuild: global resource control for rebuild

gnailzenh:
Previously, the resource controls within the rebuild system—such as the limits on ULT (user-level thread) counts and DMA buffer usage—were scoped per pool.

This meant that when rebuild or migration operations occurred across multiple pools, each pool operated within its own local resource boundaries. As a result, simultaneous rebuilds in several pools could lead to excessive system-wide consumption of threads and memory, potentially impacting performance and system stability.

This patch reworks the rebuild migration resource management to introduce global limits on the number of ULTs and DMA buffers the rebuild system can use across all pools. A centralized migration resource manager is established per target to coordinate these resources across all active pools, preventing overallocation and minimizing resource contention.

### Steps for the author:

### DIFF
--- a/src/object/srv_internal.h
+++ b/src/object/srv_internal.h
@@ -77,16 +77,13 @@ struct migrate_pool_tls {
 	 */
 	uint32_t                 mpt_tgt_obj_ult_cnt;
 	uint32_t                 mpt_tgt_dkey_ult_cnt;
+	/* The current in-flight data size */
 	uint64_t                 mpt_inflight_size;
 
 	struct migr_res_manager *mpt_rmg;
 
 	/* reference count for the structure */
-	uint64_t		mpt_refcount;
-
-	/* The current in-flight iod, mainly used for controlling
-	 * rebuild in-flight rate to avoid the DMA buffer overflow.
-	 */
+	uint64_t                 mpt_refcount;
 	uint32_t		mpt_opc;
 
 	/* The new layout version for upgrade job */
--- a/src/object/srv_internal.h
+++ b/src/object/srv_internal.h
@@ -26,6 +26,8 @@
 
 extern struct dss_module_key obj_module_key;
 
+struct migr_res_manager;
+
 /* Per pool attached to the migrate tls(per xstream) */
 struct migrate_pool_tls {
 	/* POOL UUID and pool to be migrated */
@@ -75,6 +77,9 @@ struct migrate_pool_tls {
 	 */
 	uint32_t                 mpt_tgt_obj_ult_cnt;
 	uint32_t                 mpt_tgt_dkey_ult_cnt;
+	uint64_t                 mpt_inflight_size;
+
+	struct migr_res_manager *mpt_rmg;
 
 	/* reference count for the structure */
 	uint64_t		mpt_refcount;
@@ -82,11 +87,6 @@ struct migrate_pool_tls {
 	/* The current in-flight iod, mainly used for controlling
 	 * rebuild in-flight rate to avoid the DMA buffer overflow.
 	 */
-	uint64_t		mpt_inflight_size;
-	uint64_t		mpt_inflight_max_size;
-	ABT_cond		mpt_inflight_cond;
-	ABT_mutex		mpt_inflight_mutex;
-	uint32_t		mpt_inflight_max_ult;
 	uint32_t		mpt_opc;
 
 	/* The new layout version for upgrade job */
@@ -147,6 +147,10 @@ struct obj_tgt_punch_args {
 
 void
 migrate_pool_tls_destroy(struct migrate_pool_tls *tls);
+int
+obj_migrate_init(void);
+void
+obj_migrate_fini(void);
 
 struct obj_tls {
 	d_sg_list_t		ot_echo_sgl;
--- a/src/object/srv_obj_migrate.c
+++ b/src/object/srv_obj_migrate.c
@@ -35,10 +35,55 @@
 /* Set the total in-flight size to be 25% of MAX DMA size for
  * the moment, will adjust it later if needed.
  */
-#define MIGRATE_MAX_SIZE	(1 << 28)
-/* Max migrate ULT number on the server */
-#define MIGRATE_DEFAULT_MAX_ULT	4096
+#define MIGR_TGT_INF_DATA       (1 << 28)
+
 #define ENV_MIGRATE_ULT_CNT	"D_MIGRATE_ULT_CNT"
+
+/* Number of migration ULTs per target */
+#define MIGR_TGT_ULTS_MIN       100
+#define MIGR_TGT_ULTS_DEF       500
+#define MIGR_TGT_ULTS_MAX       2000
+
+/* 1/3 object ults, 2/3 key ULTs */
+#define MIGR_OBJ_ULT_PERCENT    33
+
+#define MIGR_TGT_OBJ_ULTS(ults) ((ults * MIGR_OBJ_ULT_PERCENT) / 100)
+#define MIGR_TGT_KEY_ULTS(ults) (ults - MIGR_TGT_OBJ_ULTS(ults))
+
+enum {
+	MIGR_OBJ = 0,
+	MIGR_KEY,
+	MIGR_DATA,
+	MIGR_MAX,
+};
+
+/* resource consumed by migration */
+struct migr_resource {
+	const char *res_name;
+	/* upper limit of the resource */
+	long        res_limit;
+	/* resource amount in "unit" */
+	long        res_units;
+	/* number of waiters on this resource */
+	int         res_waiters;
+	/* ABT_cond for waiters */
+	ABT_cond    res_cond;
+};
+
+/* migration resources manager */
+struct migr_res_manager {
+	ABT_mutex            rmg_mutex;
+	struct migr_resource rmg_resources[MIGR_MAX];
+};
+
+struct migr_engine_res {
+	/* total ULTs per target, it a tunable which can be set by admin */
+	unsigned int             er_ults;
+	struct migr_res_manager *er_rmgs;
+};
+
+static struct migr_engine_res migr_eng_res;
+
 struct migrate_one {
 	daos_key_t		 mo_dkey;
 	uint64_t		 mo_dkey_hash;
@@ -391,10 +436,6 @@ migrate_pool_tls_destroy(struct migrate_pool_tls *tls)
 		D_FREE(tls->mpt_svc_list.rl_ranks);
 	if (tls->mpt_done_eventual)
 		ABT_eventual_free(&tls->mpt_done_eventual);
-	if (tls->mpt_inflight_cond)
-		ABT_cond_free(&tls->mpt_inflight_cond);
-	if (tls->mpt_inflight_mutex)
-		ABT_mutex_free(&tls->mpt_inflight_mutex);
 	if (daos_handle_is_valid(tls->mpt_root_hdl))
 		obj_tree_destroy(tls->mpt_root_hdl);
 	if (daos_handle_is_valid(tls->mpt_migrated_root_hdl))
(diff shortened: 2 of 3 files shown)


daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17267/2/testReport/

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17267/3/testReport/

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17267/3/execution/node/1355/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17267/4/testReport/

wangshilong (commented):
Looks good, better to add "Features: rebuild" in the commit to verify no regression for rebuild.

wangshilong (dismissed):

src/object/srv_obj_migrate.c
```diff
@@ -1775,77 +1804,98 @@ migrate_one_destroy(struct migrate_one *mrone)
...
-		if (tls->mpt_fini)
-			D_GOTO(out, rc = migrate_pool_tls_get_status(tls));
+	res = &rmg->rmg_resources[res_type];
+	while (1) {
+		if (tls->mpt_fini) {
+			rc = migrate_pool_tls_get_status(tls);
+			D_GOTO(out, rc);
+		}
+
+		if (res->res_units + units <= res->res_limit) {
```
liuxuezhao: limit default value is 256MB, this code will cause some extreme case always cannot do the migrate.
for example one big ext > 256MB cannot be migrated forever.
The original code allow such case by allow it to go if res->res_units is zero, despite of how big for single pulling ext, may need similar handling here

src/object/srv_obj_migrate.c
```diff
@@ -35,10 +35,55 @@
 /* Set the total in-flight size to be 25% of MAX DMA size for
  * the moment, will adjust it later if needed.
  */
-#define MIGRATE_MAX_SIZE	(1 << 28)
-/* Max migrate ULT number on the server */
-#define MIGRATE_DEFAULT_MAX_ULT	4096
+#define MIGR_TGT_INF_DATA       (1 << 28)
```
liuxuezhao: BTW may set it as larger size as now it becomes global size compared to previous per-pool control, may confirm with @NiuYawei

src/object/srv_obj_migrate.c
```diff
@@ -1919,7 +1957,7 @@ migrate_one_ult(void *arg)
 out:
 	migrate_one_destroy(mrone);
 	if (tls != NULL) {
-		migrate_tgt_exit(tls, DKEY_ULT);
+		migrate_res_release(tls, MIGR_KEY, 1);
```
liuxuezhao: I just thought about another potential problem, for the case of NULL tls it will not call migrate_res_release().
NULL tls possibly happen when the ULT scheduled after (or at the same time of) aborting the rebuild, this will cause the "res->res_units" leaked.
Current code also with the problem, I ever pushed https://github.com/daos-stack/daos/pull/17149 to fix that but later found that actually will not cause problem for current code so did not ask to land it. 
But with this PR it looks will cause potential problem because the global counter "res->res_units" will be leaked for that case.

src/object/srv_obj_migrate.c
gnailzenh: I've confirmed with Niu, I think this is a bug in original code, DMA buffer is 1GB per target and our original intention is only allows 256MB for rebuild (of all pools), but the implement actually breaks the limit.

src/object/srv_obj_migrate.c
gnailzenh: can we just call migrate_pool_tls_get() each time we create a ULT and guarantee it wouldn't disappear?

src/object/srv_obj_migrate.c
gnailzenh: or we just always allow one and only one large I/O (>= 128M), so it can always proceed? I can add a counter for this kind of large I/O which ignores the limit, and only allow one at a moment.

src/object/srv_obj_migrate.c
liuxuezhao: right, if the to be created ULT is on the same XS seem fine to take one extra refcount to ensure tls is valid, if so may assert tls is valid in the ULT. Or just reuse the already taken refcount (don't drop it if the ULT creation succeed). just FYI.

src/object/srv_obj_migrate.c
liuxuezhao: seems fine to me.

src/object/srv_obj_migrate.c
```diff
@@ -35,10 +35,55 @@
...
+
+/* migration resources manager */
+struct migr_res_manager {
+	ABT_mutex            rmg_mutex;
+	struct migr_resource rmg_resources[MIGR_MAX];
+};
+
+struct migr_engine_res {
+	/* total ULTs per target, it a tunable which can be set by admin */
+	unsigned int             er_ults;
```
kccain:
```suggestion
	unsigned int             er_max_ults;
```

src/object/srv_obj_migrate.c
```diff
@@ -35,10 +35,55 @@
...
+	struct migr_res_manager *er_rmgs;
```
kccain, line 97:
```suggestion
	struct migr_res_manager *er_rmgs; /* dss_tgt_nr items */
```

src/object/srv_obj_migrate.c
```diff
@@ -35,10 +35,55 @@
...
+	/* resource amount in "unit" */
+	long        res_units;
+	/* number of waiters on this resource */
+	int         res_waiters;
+	/* ABT_cond for waiters */
+	ABT_cond    res_cond;
+};
+
+/* migration resources manager */
+struct migr_res_manager {
```
kccain, line 88:
```suggestion
struct migr_tgt_res_manager {
```

src/object/srv_obj_migrate.c
```diff
@@ -35,10 +35,55 @@
...
+
+enum {
+	MIGR_OBJ = 0,
+	MIGR_KEY,
+	MIGR_DATA,
+	MIGR_MAX,
+};
+
+/* resource consumed by migration */
+struct migr_resource {
```
kccain, line 71:
```suggestion
struct migr_tgt_resource {
```

src/object/srv_internal.h
```diff
@@ -75,18 +77,16 @@ struct migrate_pool_tls {
...
 	uint32_t                 mpt_tgt_obj_ult_cnt;
 	uint32_t                 mpt_tgt_dkey_ult_cnt;
+	uint64_t                 mpt_inflight_size;
+
+	struct migr_res_manager *mpt_rmg;
 
 	/* reference count for the structure */
 	uint64_t		mpt_refcount;
 
 	/* The current in-flight iod, mainly used for controlling
```
kccain: this comment is now misplaced after moving/removing structure members below it.

src/object/srv_obj_migrate.c
```diff
@@ -1775,77 +1804,98 @@ migrate_one_destroy(struct migrate_one *mrone)
...
+	/* per-pool counters for rebuild status tracking */
+	if (res_type == MIGR_OBJ)
 		tls->mpt_tgt_obj_ult_cnt++;
-	else
+	else if (res_type == MIGR_KEY)
 		tls->mpt_tgt_dkey_ult_cnt++;
+	else
+		tls->mpt_inflight_size += units;
+
+	D_DEBUG(DB_REBUILD, DF_RB " migrate %s: hold units=%lu, used=%lu, limit=%lu, waited=%d\n",
```
kccain: should the debug print also convey mpt_tgt_obj_ult_cnt and mpt_tgt_dkey_ult_cnt?

src/object/srv_obj_migrate.c
```diff
@@ -1775,77 +1804,98 @@ migrate_one_destroy(struct migrate_one *mrone)
...
+	D_ASSERT(rmg != NULL);
 
-static void
-migrate_tgt_exit(struct migrate_pool_tls *tls, int ult_type)
-{
-	D_ASSERT(dss_get_module_info()->dmi_xs_id != 0);
-	if (ult_type == OBJ_ULT) {
+	res = &rmg->rmg_resources[res_type];
+
+	D_DEBUG(DB_REBUILD, DF_RB "migrate %s: release units=%lu, used=%lu, limit=%lu\n",
```
kccain: include ult counts in this debug log also?

src/object/srv_obj_migrate.c
```diff
@@ -1775,77 +1804,98 @@ migrate_one_destroy(struct migrate_one *mrone)
...
 		D_ASSERT(tls->mpt_tgt_dkey_ult_cnt > 0);
 		tls->mpt_tgt_dkey_ult_cnt--;
+	} else {
+		D_ASSERT(tls->mpt_inflight_size >= units);
+		tls->mpt_inflight_size -= units;
+	}
+
+	D_ASSERT(res->res_units >= units);
+	res->res_units -= units;
+	if (res->res_waiters > 0) {
```
kccain, line 1937: should the test of res->res_waiters be done instead while the rmg->rmg_mutex lock is held? It seems like it should be evaluated while holding the lock.

kccain (commented):
initial pass review

src/object/srv_obj_migrate.c
gnailzenh: all ULTs changing res_waiters are on the same xstream so it's not required right? The only reason we have mutex is because ABT_cond requires it.

src/object/srv_obj_migrate.c
gnailzenh: I actually avoid to add the tgt_prefix at here, because I think we might need it for other cases in the future.

src/object/srv_obj_migrate.c
gnailzenh: let's keep it as now because I'm thinking about use this data structure for other resources.

src/object/srv_obj_migrate.c
```diff
@@ -31,15 +31,76 @@
...
-/* Max in-flight data size per xstream */
-/* Set the total in-flight size to be 25% of MAX DMA size for
+/* Max in-flight transfer size per xstream */
+/* Set the total in-flight size to be 50% of MAX DMA size for
  * the moment, will adjust it later if needed.
  */
-#define MIGRATE_MAX_SIZE	(1 << 28)
-/* Max migrate ULT number on the server */
-#define MIGRATE_DEFAULT_MAX_ULT	4096
+#define MIGR_TGT_INF_DATA       (1 << 29)
```
gnailzenh, line 38: I made this change based on Niu's feedback, he thinks ULT can throttle resource usage most of the time.

src/object/srv_obj_migrate.c
```diff
@@ -2687,21 +2756,24 @@ migrate_start_ult(struct enum_unpack_arg *unpack_arg)
...
+
+		migrate_pool_tls_get(tls);
+		mrone->mo_tls = tls;
+
+		D_ASSERT(arg->tgt_idx == dss_get_module_info()->dmi_tgt_id);
+		rc = dss_ult_create(migrate_one_ult, mrone, DSS_XS_SELF, 0, MIGRATE_STACK_SIZE,
+				    NULL);
 		if (rc) {
-			migrate_tgt_exit(tls, DKEY_ULT);
+			migrate_res_release(tls, MIGR_KEY, 1);
```
wangshilong, line 2771: migrate_pool_tls_put() missed if ult creation fail?

gnailzenh: No, it's actually called by migrate_one_destroy, because mrone->mo_tls = tls has taken over the refcount

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17267/7/testReport/

wangshilong (approved):

src/object/srv_obj_migrate.c
```diff
@@ -31,15 +31,76 @@
...
+/* Set the total in-flight size to be 50% of MAX DMA size for
  * the moment, will adjust it later if needed.
  */
-#define MIGRATE_MAX_SIZE	(1 << 28)
-/* Max migrate ULT number on the server */
-#define MIGRATE_DEFAULT_MAX_ULT	4096
+#define MIGR_TGT_INF_DATA       (1 << 29)
+
+/* Threshold for very large transfers.
+ * This may exceed the MIGR_TGT_INF_DATA limit to prevent starvation.
```
liuxuezhao, line 41: minor, actually with  "D_CASSERT(MIGR_TGT_INF_DATA > MIGR_INF_DATA_HULK);"

liuxuezhao (approved):

kccain (approved):

PR #17380, opened 2026-01-14
DAOS-18161 object: load container write stamp from VOS

gnailzenh:
After restart, load the latest epoch of writes that require aggregation from VOS instead of setting it to zero, otherwise EC aggregation has to scan the container again even it's not changed after previous aggregation.

### Steps for the author:

### DIFF
--- a/src/object/srv_ec_aggregate.c
+++ b/src/object/srv_ec_aggregate.c
@@ -1,7 +1,7 @@
 /**
  * (C) Copyright 2020-2024 Intel Corporation.
  * (C) Copyright 2025 Google LLC
- * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
+ * (C) Copyright 2025-2026 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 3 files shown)


src/object/srv_ec_aggregate.c
```diff
@@ -2734,19 +2734,27 @@ cont_ec_aggregate_cb(struct ds_cont_child *cont, daos_epoch_range_t *epr,
...
 	}
 
+	if (cont->sc_ec_agg_eph == 0) {
+		D_INFO(DF_CONT ": update cont->sc_ec_agg_eph to " DF_X64,
+		       DP_CONT(cont->sc_pool->spc_uuid, cont->sc_uuid),
+		       cont->sc_ec_agg_eph_boundary);
+		cont->sc_ec_agg_eph = cont->sc_ec_agg_eph_boundary;
+	}
+
+	if (cont->sc_ec_update_timestamp == 0) {
```
NiuYawei, line 2750: This `sc_ec_update_timestamp` could be removed.

NiuYawei (approved):

src/object/srv_ec_aggregate.c
liuxuezhao: if without it need to vos_cont_query() everytime? so maybe fine to keep it?

liuxuezhao (approved):

src/object/srv_ec_aggregate.c
gnailzenh: Loading it is very cheap, I think it's OK to remove it but not for 2.6 to reduce change, we can make this change on 2.8

PR #17462, opened 2026-01-27
DAOS-18487 rebuild: disallow reint/extend when with DOWN targets

liuxuezhao:
If not in delay_rebuild mode disallow reint/extend if with DOWN targets, user should try later after rebuild done.

Features: rebuild

### Steps for the author:

### DIFF
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -7926,7 +7926,6 @@ ds_pool_reint_extend_allowed(struct pool_svc *svc)
 	struct ds_pool     *pool          = svc->ps_pool;
 	struct pool_target *targets       = NULL;
 	uint32_t            down_tgts_cnt = 0;
-	uint64_t            sys_self_heal = 0;
 	int                 rc;
 	bool                ret = true;
 
@@ -7948,27 +7947,6 @@ ds_pool_reint_extend_allowed(struct pool_svc *svc)
 		D_DEBUG(DB_REBUILD,
 			DF_UUID " with DOWN tgt in delay_rebuild node, "
 				"REINT/EXTEND allowed",
-			DP_UUID(pool->sp_uuid));
-		goto out;
-	}
-
-	rc = ds_mgmt_get_self_heal_policy(pool_svc_abort_gshp, svc, &sys_self_heal);
-	if (rc != 0) {
-		DL_ERROR(rc, DF_UUID ": failed to get self-heal policy", DP_UUID(svc->ps_uuid));
-		/*
-		 * Another PS replica might be able reach the MS. If I'm
-		 * not the PS leader of the specified term, this
-		 * rdb_resign call does nothing.
-		 */
-		rdb_resign(svc->ps_rsvc.s_db, svc->ps_rsvc.s_term);
-		ret = false;
-		goto out;
-	}
-
-	if (!(sys_self_heal & DS_MGMT_SELF_HEAL_POOL_REBUILD)) {
-		D_DEBUG(DB_REBUILD,
-			DF_UUID " with DOWN tgt, pool_rebuild disabled in "
-				"sys_self_heal, REINT/EXTEND allowed",
 			DP_UUID(pool->sp_uuid));
 		goto out;
 	}
--- a/src/pool/srv_pool.c
+++ b/src/pool/srv_pool.c
@@ -1,6 +1,6 @@
 /*
  * (C) Copyright 2016-2024 Intel Corporation.
- * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
+ * (C) Copyright 2025-2026 Hewlett Packard Enterprise Development LP
  * (C) Copyright 2025 Google LLC
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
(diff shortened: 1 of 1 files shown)


src/pool/srv_pool.c
```diff
@@ -7920,6 +7920,68 @@ pool_discard(crt_context_t ctx, struct pool_svc *svc, struct pool_target_addr_li
...
+
+	rc = ds_mgmt_get_self_heal_policy(pool_svc_abort_gshp, svc, &sys_self_heal);
+	if (rc != 0) {
+		DL_ERROR(rc, DF_UUID ": failed to get self-heal policy", DP_UUID(svc->ps_uuid));
+		/*
+		 * Another PS replica might be able reach the MS. If I'm
+		 * not the PS leader of the specified term, this
+		 * rdb_resign call does nothing.
+		 */
+		rdb_resign(svc->ps_rsvc.s_db, svc->ps_rsvc.s_term);
```
wangshilong: I am not sure we need call rdb_resign here? it looks copied from pool_svc_update_map(), maybe @liw could confirm.

wangshilong (dismissed):

src/pool/srv_pool.c
liw: I'm not sure why reint and extend need to check the self-heal policy: The only automatic recovery (i.e., self-heal) is the SWIM-based exclusion case; all reint/extend are manual. Am I missing @liuxuezhao's point, I wonder?

liuxuezhao: if pool_rebuild is enabled, if with DOWN tgt should disallow REINT.
But if pool_rebuild is disabled, even with DOWN tgt, cannot disallow REINT, that is the difference right?

If don't check pool_rebuild, you mean always allow or disallow REINT if with DOWN tgt already?

liw: Yes, I think the self-heal policy can't tell us anything about whether there is an ongoing rebuild task or not. Also, because reint/extend are manual, they are not managed by the self-heal policy. Hence, it sounds like we don't need to look at the self-heal policy here.

liw: Example 1: The rebuild flags in both the system and pool self-heal properties are disabled, but the administrator invokes `dmg pool exclude ...`---a rebuild task is in progress while self-heal rebuild is disabled.

Example 2: The rebuild flags in both the system and pool self-heal properties are enabled, an engine is excluded, but the administrator happens to disable one of the rebuild flags---a rebuild task is in progress while self-heal rebuild is disabled.

liuxuezhao: so you tend to always disallow REINT if with DOWN tgt already, despite of pool_rebuild policy setting?

Then how could it work for the case that user disable pool_rebuild, so SWIM exclusion will not trigger rebuild. And following REINT will not trigger rebuild either, that is not what we want, no?

liw: Right, we do want the reint to happen. What I'm saying is only "it seems looking at self-heal policy is not the right way". Where is the fundamental issue? Is it about "have there been any rebuild attempt for the DOWN targets?

liuxuezhao: OK, let's remove the pool_rebuild policy check now, later can refine it if needed. thanks.

wangshilong (approved):

liw (approved):

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17462/3/testReport/

NiuYawei (approved):

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17462/3/execution/node/1355/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17462/3/testReport/

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17462/3/testReport/

src/pool/srv_pool.c
kccain: I can't see the full context of the code/conversation here due to the subsequent force push. For the case that a user has disabled rebuild in a pool's `self_heal` property, a SWIM exclusion occurs and does not trigger rebuild - then, if a reintegrate command is issued, with this patch it will not be permitted. So, if the administrator actually wants to reintegrate the targets:

- first the pool `self_heal` property would need to be modified to enable rebuild, and `dmg system self-heal eval` command would be issued. Those actions will allow the `DOWN` targets to be rebuilt to `DOWN_OUT` state.
- After this, another reintegrate command could be issued, which would then be permitted and a reintegration rebuild would run.
Do I have it right? If so, that sounds OK to me.

kccain (approved):

PR #17495, opened 2026-02-03
DAOS-18564 control: Check kernel config for mount options

mjmac:
Set ramdisk mount options based on kernel configuration
(e.g. CONFIG_NUMA=y) instead of assuming. Allows DAOS to
run on more platforms.

Features: control
Signed-off-by: Michael MacDonald <github@macdonald.cx>


daosbuild3:
Test stage Functional on EL 8.8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17495/1/execution/node/1011/log

daosbuild3:
Test stage Test RPMs on EL 8.6 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17495/1/execution/node/1086/log

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17495/2/testReport/

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17495/4/execution/node/1315/log

src/control/server/storage/scm/provider.go
```diff
@@ -96,11 +97,20 @@ func DefaultProvider(log logging.Logger) *Provider {
 
 // NewProvider returns an initialized *Provider.
 func NewProvider(log logging.Logger, backend Backend, sys SystemProvider, mounter storage.MountProvider) *Provider {
+	// Parse kernel config once at provider init; it doesn't change at runtime.
+	// NB: If kernel config is needed outside of the scm provider, consider
+	// moving to a sync.Once-based accessor in the system package instead.
+	kernelCfg, err := system.ParseKernelConfig()
+	if err != nil {
+		log.Debugf("failed to parse kernel config: %s", err)
```
tanabarr: shouldn't this be printed at info level to prevent it getting lost in debug noise?

src/control/server/storage/scm/provider.go
```diff
@@ -438,30 +448,56 @@ func (p *Provider) mountDcpm(device, target string) (*storage.MountResponse, err
...
+	} else {
+		p.log.Debug("NUMA kernel support not detected; skipping tmpfs mpol option")
 	}
 	if params.Size > 0 {
-		opts = append(opts, fmt.Sprintf("size=%dg", params.Size))
+		mountOpts = append(mountOpts, fmt.Sprintf("size=%dg", params.Size))
 	}
-	if !params.DisableHugepages {
-		opts = append(opts, "huge=always")
+	if !params.DisableHugepages && system.IsKernelConfigEnabled(kernelCfg, "CONFIG_TRANSPARENT_HUGEPAGE") {
```
tanabarr: If the kernel has support but the global policy (/sys/kernel/mm/transparent_hugepage/enabled) is set to never, you can still use huge=always on a specific tmpfs mount to enable them just for that instance. This should be considered here as it could result in confusing hugepage usage if the global runtime policy is disabled.

I would prefer to return an error in the following cases:
- !params.DisableHugepages && !CONFIG_TRANSPARENT_HUGEPAGE
- !params.DisableHugepages && /sys/kernel/mm/transparent_hugepage/enabled=never

This will prevent unexpected behaviour from confusing configuration combinations. The user should be prevented from using the scm_hugepages_disabled=false setting in config unless the environment is configured to support it.

Alternatively leave the above changes to be  implemented in this dedicated PR (https://github.com/daos-stack/daos/pull/17526)  rather than here.

tanabarr (changes requested):

src/control/server/storage/scm/provider.go
mjmac: Hmm. This is a bit clunky, tbh. If it fails here, it will fail again loudly down in `mountRamdisk()`. I added this as an eager load for the happy path. It would be better to have `scm.NewProvider()` return an error if it fails, but that had wide ripple effects.

src/control/server/storage/scm/provider.go
mjmac: We probably need to extract the logic in your PR and this PR into a helper that can be tested directly. e.g. `scm.Provider.ramdiskShouldUseHugepages()` or something like that. I can take care of that after your PR is merged.

tanabarr: it's been merged

mjmac: OK, I've added a commit to address this feedback via a new helper. I decided not to prohibit `huge=always` if the global policy is disabled because it is technically supported at the system level. The server will emit a warning in this case, though.

src/control/provider/system/kernel_config.go
```diff
@@ -0,0 +1,115 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+// KernelConfig is a parsed kernel configuration, mapping option names
+// (e.g. "CONFIG_TRANSPARENT_HUGEPAGE") to their values (e.g. "y", "m", "n").
+type KernelConfig map[string]string
```
kjacque, line 23: If this only applies to Linux, might be good to clarify that in the naming of the file. I think the type/function names are fine.

kjacque (approved):
Comment isn't blocking. Nice work.

src/control/server/storage/scm/provider.go
```diff
@@ -436,6 +452,27 @@ func (p *Provider) formatDcpm(req storage.ScmFormatRequest) (*storage.ScmFormatR
...
 
+// ramdiskHugepageOpt checks whether hugepages can be used for a tmpfs mount
+// and returns the appropriate mount option. It returns an error if hugepages
+// are requested but not available.
+func (p *Provider) ramdiskHugepageOpt(kernelCfg system.KernelConfig) (string, error) {
+	if !kernelCfg.IsEnabled("CONFIG_TRANSPARENT_HUGEPAGE") {
+		return "", FaultHugepagesNotSupported
+	}
+
+	if p.thpDetector != nil {
```
tanabarr, line 463: Why do we skip the check if thpDetector == nil? Wouldn't it be safer to return an error if nil and always make sure that the detector is intentionally provided.

src/control/server/storage/scm/provider.go
```diff
@@ -436,6 +452,27 @@ func (p *Provider) formatDcpm(req storage.ScmFormatRequest) (*storage.ScmFormatR
...
+		enabled, err := p.thpDetector.IsTHPEnabled()
+		if err != nil {
+			p.log.Noticef("unable to check transparent hugepage status: %s", err)
+		} else if !enabled {
+			p.log.Notice("transparent hugepages globally disabled but per-mount huge=always will still work; " +
```
tanabarr, line 468: This is better, thanks!

src/control/server/storage/scm/provider.go
```diff
@@ -486,10 +523,14 @@ func (p *Provider) doMountRamdisk(target string, params *storage.RamdiskParams,
...
-	if !params.DisableHugepages && system.IsKernelConfigEnabled(kernelCfg, "CONFIG_TRANSPARENT_HUGEPAGE") {
-		mountOpts = append(mountOpts, "huge=always")
-	} else if !params.DisableHugepages {
-		p.log.Debug("transparent hugepage kernel support not detected; skipping tmpfs huge option")
+	if !params.DisableHugepages {
+		hugeOpt, err := p.ramdiskHugepageOpt(kernelCfg)
+		if err != nil {
+			return nil, err
+		}
+		if hugeOpt != "" {
```
tanabarr, line 531: This check is unnecessary, hugeOpt should never return empty string if err == nil.

tanabarr (approved):
nonblocking comments

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17495/12/execution/node/1346/log

src/control/server/storage/scm/provider.go
mjmac: Hmm, I suppose I was being a bit lazy here. I'll fix in a follow-up.

mjmac:
@daos-stack/daos-gatekeeper Please force this one. The test failures are in dfuse; unrelated to these control plane changes. Contributor email should be github@macdonald.cx, thanks.

daltonbohning:
This did also fail `ior/small`, which is probably unrelated but I have not seen that test failing for a while so I repushed to make sure.

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17495/14/testReport/

daltonbohning:
> This did also fail `ior/small`, which is probably unrelated but I have not seen that test failing for a while so I repushed to make sure.

It passed this time so I'm assuming it was a network issue

PR #17535, opened 2026-02-10
DAOS-18487 placement: fix a bug in shard remapping chain

liuxuezhao:
For the case of 2nd remap, if the spare target is DOWN2UP need to set fs_down2up flag, to make it be able to set shard's po_rebuilding flag at the end.
One example case -
Target A is DOWN, rebuild completed and status changed to DOWNOUT
Target B is DOWN, rebuild started but not completed but admin do the reint,
its status change to UP and with DOWN2UP flag.

In object layout calculation, one shard firstly located in Target A, but 1st remap to Target B, but still need to do 2nd remap. In this case should set fs_down2up flag which is not set in the 1st remap, to avoid not be able to set shard's po_rebuilding flag so will cause read from it (invalid place).

This bug could cause data corruption (mostly like with cause shard losing).

### Steps for the author:

### DIFF
--- a/src/placement/pl_map_common.c
+++ b/src/placement/pl_map_common.c
@@ -1,5 +1,6 @@
 /**
  * (C) Copyright 2016-2024 Intel Corporation.
+ * (C) Copyright 2026 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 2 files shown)


wangshilong (approved):

src/placement/pl_map_common.c
```diff
@@ -378,6 +379,22 @@ determine_valid_spares(struct pool_target *spare_tgt, struct daos_obj_md *md,
...
+		 * Target B is DOWN, rebuild started but not completed but admin do the reint,
+		 * its status change to UP and with DOWN2UP flag.
+		 *
+		 * In object layout calculation, one shard firstly located in Target A, but 1st
+		 * remap to Target B, but still need to do 2nd remap. In this case should set
+		 * fs_down2up flag which is not set in the 1st remap, to avoid not be able to set
+		 * shard's po_rebuilding flag so will cause read from it (invalid place).
+		 */
+		if (pool_target_is_down2up(spare_tgt))
+			f_shard->fs_down2up = 1;
```
NiuYawei, line 396: so the "f_shard->fs_down2up = 1;" in remap_alloc_one() could be removed?

NiuYawei (approved):

src/placement/pl_map_common.c
liuxuezhao: I think that one cannot be removed, 
remap_alloc_one()'s "fs_down2up = 1;" is for the normal case that only one time remap
here is that for the case of first time remap did not set the flag, but 2nd time remap need to set it.

kccain (approved):
this area of the code is a little new to me, but changes seem reasonable after some preliminary study of the code.

daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17535/2/testReport/

liuxuezhao:
replaced by https://github.com/daos-stack/daos/pull/17789

PR #17536, opened 2026-02-10
DAOS-18487 placement: fix a bug in shard remapping chain

liuxuezhao:
For the case of 2nd remap, if the spare target is DOWN2UP need to set fs_down2up flag, to make it be able to set shard's po_rebuilding flag at the end.
One example case -
Target A is DOWN, rebuild completed and status changed to DOWNOUT Target B is DOWN, rebuild started but not completed but admin do the reint, its status change to UP and with DOWN2UP flag.

In object layout calculation, one shard firstly located in Target A, but 1st remap to Target B, but still need to do 2nd remap. In this case should set fs_down2up flag which is not set in the 1st remap, to avoid not be able to set shard's po_rebuilding flag so will cause read from it (invalid place).

This bug could cause data corruption (mostly like with cause shard losing).

### Steps for the author:

### DIFF
--- a/src/object/obj_layout.c
+++ b/src/object/obj_layout.c
@@ -115,8 +115,8 @@ obj_dump_grp_layout(daos_handle_t oh, uint32_t shard)
 	}
 
 	grp_idx = shard / obj->cob_grp_size;
-	D_INFO(DF_OID " shard %d, grp_idx %d, grp_size %d", DP_OID(obj->cob_md.omd_id), shard,
-	       grp_idx, obj->cob_grp_size);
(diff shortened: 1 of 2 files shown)


daosbuild3:
Test stage NLT on EL 8.8 completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17536/1/testReport/

wangshilong (approved):

NiuYawei (approved):

liuxuezhao:
replaced by https://github.com/daos-stack/daos/pull/17789

PR #17644, opened 2026-03-04
DAOS-17946 doc: update space_rb default in comment

daltonbohning:
Per #17032, default space_rb is now 5%.

Doc-only: true

### Steps for the author:

### DIFF
--- a/src/include/daos_prop.h
+++ b/src/include/daos_prop.h
@@ -1,6 +1,6 @@
 /**
  * (C) Copyright 2015-2023 Intel Corporation.
- * (C) Copyright 2025 Hewlett Packard Enterprise Development LP
+ * (C) Copyright 2025-2026 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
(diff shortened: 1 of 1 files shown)


src/include/daos_prop.h
```diff
@@ -38,7 +38,7 @@ enum daos_pool_props {
 	DAOS_PROP_PO_ACL,
 	/**
 	 * Reserve space ratio: amount of space to be reserved on each target
-	 * for rebuild purpose. default = 0%.
+	 * for rebuild purpose. default = 5%.
```
daltonbohning, line 41: I wonder if we should actually just remove the default in this comment so it doesn't get out of sync. And most other props in this file don't mention default values either.

kjacque: I think it would be fine to remove it.

daltonbohning: Thanks. I'll just do that then to reduce maintenance and confusion

daltonbohning: Hmm but then I wonder if we should remove default from all of these or add/update them all to be consistent at least

kjacque: I'd be inclined to remove all of them. Definitely will reduce the maintenance burden.

kjacque (approved):
I'll approve as-is, it's correct after all. I do think it'd be better to remove these defaults from comments so we don't have to keep them in sync with in-code definitions.

johannlombardi (approved):

PR #17696, opened 2026-03-13
DAOS-18633 rebuild: throttle rebuild status logs to reduce overhead

wangshilong:
Currently, each target dumps its rebuild progress to the log every 2 seconds unconditionally. In a large-scale scenario where a system has 100 pools rebuilding concurrently across 16 targets per rank, running for 10 hours can generate massive amounts of log data (several GBs per rank). This continuous, high-frequency logging (around 50 logs per second per xstream) causes severe I/O contention and negatively impacts overall I/O performance and ULT scheduling.

There is no necessary reason to print background progress logs this frequently. This patch throttles the rebuild status log dumping from 2 seconds to 5 minutes. The final status will still be printed immediately if a rebuild completes or aborts, ensuring that we retain sufficient visibility for debugging while avoiding log storms.

### Steps for the author:


NiuYawei (dismissed):

daosbuild3:
Test stage NLT completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17696/1/testReport/

src/rebuild/srv.c
```diff
@@ -1062,7 +1064,7 @@ rebuild_leader_status_check(struct ds_pool *pool, uint32_t op,
 				}
 			}
 
-			if (now - last_print > 20)
+			if (now - last_print > RBLD_LOG_INTV)
```
liuxuezhao: for some quick rebuild less than RBLD_LOG_INTV, it will cause cannot see this log forever. then it is hard to find which set of ranks caused the rebuild.
How about change to firstly use RBLD_CHECK_INTV (or log it at beginning), then double it every time until reach to RBLD_LOG_INTV.

src/rebuild/srv.c
```diff
@@ -3021,11 +3025,15 @@ rebuild_tgt_status_check_ult(void *arg)
 			}
 		}
 
-		D_INFO(DF_RB " obj " DF_U64 " rec " DF_U64 " size " DF_U64 " scan done %d "
-			     "pull done %d scan gl done %d gl done %d status %d abort %s\n",
-		       DP_RB_RPT(rpt), iv.riv_obj_count, iv.riv_rec_count, iv.riv_size,
-		       rpt->rt_scan_done, iv.riv_pull_done, rpt->rt_global_scan_done,
-		       rpt->rt_global_done, iv.riv_status, rpt->rt_abort ? "yes" : "no");
+		if (check_cnt % RBLD_LOG_INTV_CNT == 0 || rpt->rt_global_done || rpt->rt_abort) {
```
liuxuezhao: ditto

src/rebuild/srv.c
wangshilong: Good idea, let me refresh.

liuxuezhao: acutally last_print init as 0 so will log it at the first time.
But double the interval till reach to RBLD_LOG_INTV can help to provide some progress for short rebuild so maybe good.
thanks.

liuxuezhao (approved):

kccain (approved):
any reason the leader loop doesn't use the same modulo approach as implemented in the tgt status loop?

wangshilong:
> any reason the leader loop doesn't use the same modulo approach as implemented in the tgt status loop?

I tried to avoid double when updating leader parts and over-optimize to avoid % operations, but forgot to update tgt status loop, but I guess this is not a big deal..

NiuYawei (approved):

PR #17710, opened 2026-03-16
DAOS-18685 object: fix a bug of tag set when retry in agg_peer_update…

liuxuezhao:
…_ult

Should get ep_tag from layout rather than directly reuse previous RPC's value because ds_obj_req_create() will calculate/change it every time.

### Steps for the author:

### DIFF
--- a/src/object/srv_ec_aggregate.c
+++ b/src/object/srv_ec_aggregate.c
@@ -1428,10 +1428,10 @@ agg_peer_update_ult(void *arg)
 
 		if (peer == pidx)
 			continue;
+retry:
 		D_ASSERT(entry->ae_peer_pshards[peer].sd_rank != DAOS_TGT_IGNORE);
 		tgt_ep.ep_rank = entry->ae_peer_pshards[peer].sd_rank;
 		tgt_ep.ep_tag  = entry->ae_peer_pshards[peer].sd_tgt_idx;
-retry:
 		peer_retry = false;
 		rc = ds_obj_req_create(dss_get_module_info()->dmi_ctx, &tgt_ep,
 				       DAOS_OBJ_RPC_EC_AGGREGATE, &rpc);


src/object/srv_ec_aggregate.c
```diff
@@ -1428,10 +1428,10 @@ agg_peer_update_ult(void *arg)
 
 		if (peer == pidx)
 			continue;
+retry:
 		D_ASSERT(entry->ae_peer_pshards[peer].sd_rank != DAOS_TGT_IGNORE);
 		tgt_ep.ep_rank = entry->ae_peer_pshards[peer].sd_rank;
 		tgt_ep.ep_tag  = entry->ae_peer_pshards[peer].sd_tgt_idx;
```
wangshilong, line 1434: [Question] I thought this might be not safe, if this happens it means during parity updating (retry might yeild), parity changed. considering we had 2 peers. If we changed two peers and first peer has been updated, that means this agg_peer_update_ult update inconsistency parity(second peer update to new parity)?  more safe way is if we detected parity changed.

Aborting ult or continue agg peer updating here looks both unsafe for me in this case..

A more question is could we prevent parity changes during agg update or we abort before starting?

daosbuild3:
Test stage Unit Test completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17710/1/testReport/

src/object/srv_ec_aggregate.c
liuxuezhao: The failed case in the test is peer addr is not changed (rank 9, tag 5), it get DER_STALE and retry.
But the problem is ds_obj_req_create() will change ep_tag as 7. If retry from original label, then it will send to wrong tag (data corruption). In the test, it retry again and again to tag 9, 11, 13 until assert it exceed MAX tag index.

Peer addr change or not is another problem, as EC agg use RO mode layout(PRE_REBUILD), it most likely will not change peer addr. 
Currently, we only retry if first peer update succeed, then in retry even if 2nd peer addr changed, looks fine to update to new peer addr. The previous failed update to old 2nd peer suppose will be cleaup by reclaim later, layout will not access the old 2nd peer although.
For more complex case, I think we have to depend on DTX to handle it in future, to leverage its abort ability.

wangshilong (approved):

src/object/srv_ec_aggregate.c
wangshilong: > The failed case in the test is peer addr is not changed (rank 9, tag 5), it get DER_STALE and retry. But the problem is ds_obj_req_create() will change ep_tag as 7. If retry from original label, then it will send to wrong tag (data corruption). In the test, it retry again and again to tag 9, 11, 13 until assert it exceed MAX tag index.
> 

I got what you tried to fix, good catch!


> Peer addr change or not is another problem, as EC agg use RO mode layout(PRE_REBUILD), it most likely will not change peer addr. Currently, we only retry if first peer update succeed, then in retry even if 2nd peer addr changed, looks fine to update to new peer addr. The previous failed update to old 2nd peer suppose will be cleaup by reclaim later, layout will not access the old 2nd peer although. For more complex case, I think we have to depend on DTX to handle it in future, to leverage its abort ability.

gnailzenh (approved):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17710/5/execution/node/1366/log

gnailzenh:
CI is not very stable lately, since we've landed this on 2.6 (it's all green), and this is clean cherry-pick of one line change, I'm landing it.

PR #17829, opened 2026-03-30
DAOS-18607 placement: generate partial layout for rebuild and EC aggregation

gnailzenh:
Rebuild and EC aggregation do not require the full object layout;
instead, they only need the localities of the shards within the
same redundancy group as the current shard.

This patch stops the layout computation after generating the layout
for the requested redundancy group, saving a significant amount of
computation resources.
### Steps for the author:

### DIFF
--- a/src/placement/jump_map.c
+++ b/src/placement/jump_map.c
@@ -1,7 +1,7 @@
 /**
  *
- * (C) Copyright 2016-2024 Intel Corporation.
- * (C) Copyright 2025-2026 Hewlett Packard Enterprise Development LP
+ * Copyright 2016-2024 Intel Corporation.
+ * Copyright 2025-2026 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -703,12 +703,17 @@ get_object_layout(struct pl_jump_map *jmap, uint32_t layout_ver, struct pl_obj_l
 					D_GOTO(out, rc = -DER_NOMEM);
 
 				remap_add_one(&remap_list, shard);
+				/* remapping requires full layout */
+				if (i == md->omd_grp_spec)
+					md->omd_grp_spec = PL_GRP_MAX;
 			} else {
 				layout_set_shard_flags(layout, k, remap_flags);
 				if (domain != NULL)
 					setbit(dom_cur_grp_real, domain - root);
 			}
 		}
+		if (i >= md->omd_grp_spec)
+			break; /* caller doesn't require the full layout */
 	}
 
 	if (fail_tgt_cnt > 0)
@@ -776,6 +781,9 @@ obj_layout_alloc_and_get(struct pl_jump_map *jmap, uint32_t layout_ver,
 			DP_RC(rc));
 		return rc;
 	}
+
+	if (gen_mode != PRE_REBUILD || !(md->omd_flags & PL_FL_GRP_SPEC))
+		md->omd_grp_spec = PL_GRP_MAX; /* full layout */
 
 	rc = get_object_layout(jmap, layout_ver, *layout_p, jmop, allow_version, gen_mode, md);
 	if (rc) {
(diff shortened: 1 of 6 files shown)


src/placement/jump_map.c
```diff
@@ -777,6 +782,9 @@ obj_layout_alloc_and_get(struct pl_jump_map *jmap, uint32_t layout_ver,
 		return rc;
 	}
 
+	if (gen_mode != PRE_REBUILD || !(md->omd_flags & PL_FL_GRP_SPEC))
+		md->omd_grp_spec = PL_GRP_MAX; /* full layout */
```
liuxuezhao: This PR looks can work. Just a few concern that the generated layout actually with partial valid data compared to the definition -
 73 struct pl_obj_layout {                                                                              
 74 >       uint32_t>       >        ol_ver;                                                            
 75 >       uint32_t>       >        ol_grp_size;                                                       
 76 >       uint32_t>       >        ol_grp_nr;                                                         
 77 >       uint32_t>       >        ol_nr;                                                             
 78 >       /* number of peer targets, this field is added for reint/extension/drain placement */       
 79 >       unsigned int             ol_shard_peers;                                                    
 80 >       struct pl_obj_shard>    *ol_shards;                                                         
 81 };         

seems 2 options -
1) set the ol_nr, ol_grp_nr as actual value, or
2) set ol_nr, ol_grp_nr same as current code (pl_obj_layout_alloc(jmop->jmop_grp_size, jmop->jmop_grp_nr) above), but assert it will not exceed valid range when using it in srv_ec_aggregate.c?

And seems for obj_reclaim() it can also use similar method?

wangshilong (approved):

src/placement/jump_map.c
```diff
@@ -703,12 +717,24 @@ get_object_layout(struct pl_jump_map *jmap, uint32_t layout_ver, struct pl_obj_l
...
 			}
 		}
+
+		if (i >= grp_spec)
+			break; /* caller doesn't require the full layout */
+	}
+
+	if (md->omd_grp_spec != PL_GRP_MAX) {
+		/* may have generated the full layout, but ignore other parts */
+		layout->ol_grp_nr = md->omd_grp_spec + 1;
```
liuxuezhao, line 736: md->omd_grp_spec may be set as a value > jmop->jmop_grp_nr - 1,
seems that either check that case and return failure,
or here set "layout->ol_grp_nr = min(md->omd_grp_spec + 1, jmop->jmop_grp_nr)"?

src/placement/jump_map.c
```diff
@@ -1102,7 +1131,8 @@ jump_map_obj_find_diff(struct pl_map *map, uint32_t layout_ver, struct daos_obj_
 		D_GOTO(out, rc);
 
 	obj_layout_dump(md->omd_id, reint_layout);
-	rc = layout_find_diff(jmap, layout, reint_layout, &reint_list, true);
+	rc = layout_find_diff(jmap, layout, reint_layout, &reint_list, true,
+			      (md->omd_flags & PL_FL_GRP_SPEC) ? md->omd_grp_spec : PL_GRP_MAX);
```
liuxuezhao, line 1135: using PL_FL_GRP_SPEC in jump_map_obj_find_diff() seems dangerous, and current code seems possibly with problem.
For example if L1123 and L1128 two obj_layout_alloc_and_get(), one call remapped, another call does not remap that will generate different layout size, may cause some trouble for layouf_find_diff?
as the new version code did not change md->omd_grp_spec

liuxuezhao: checked again that should not a problem.

src/placement/jump_map.c
liuxuezhao: probably can change it in a following PR, if it is correct.

liuxuezhao (approved):

src/placement/jump_map.c
gnailzenh: yes,  would be nice to add this check in the future.
I will add it while applying this to dsc_obj_open()

wangshilong (approved):

PR #17850, opened 2026-03-31
DAOS-18690 vos: emergency buffer for GC

gnailzenh:
GC may require extra space for TX snapshots, and it can't reclaim any space
if it failed to get the space for snapshots.

This patch preallocates space for TX snapshots, GC switches to emergency
mode and use the preallocated buffer in case of ENOSPACE.

### Steps for the author:

### DIFF
--- a/src/common/mem.c
+++ b/src/common/mem.c
@@ -1,6 +1,6 @@
 /**
- * (C) Copyright 2016-2024 Intel Corporation.
- * (C) Copyright 2025-2026 Hewlett Packard Enterprise Development LP
+ * Copyright 2016-2024 Intel Corporation.
+ * Copyright 2025-2026 Hewlett Packard Enterprise Development LP
  *
  * SPDX-License-Identifier: BSD-2-Clause-Patent
  */
@@ -1003,6 +1003,23 @@ pmem_tx_stage(void)
 	return pmemobj_tx_stage();
 }
 
+static int
+pmem_tx_set_snapbuf(struct umem_instance *umm, umem_off_t snapbuf, size_t size)
+{
+	void *buf = umem_off2ptr(umm, snapbuf);
+	int   rc;
+
+	rc = pmemobj_tx_log_append_buffer(TX_LOG_TYPE_SNAPSHOT, buf, size);
+	if (rc)
+		return rc;
+
+	rc = pmemobj_tx_log_auto_alloc(TX_LOG_TYPE_SNAPSHOT, 0);
+	if (rc)
+		return rc;
+
+	return 0;
+}
+
 static umem_off_t
 pmem_reserve(struct umem_instance *umm, void *act, size_t size, unsigned int type_num,
 	     unsigned int unused)
@@ -1135,27 +1152,27 @@ umem_tx_add_cb(struct umem_instance *umm, struct umem_tx_stage_data *txd,
 	return 0;
 }
 
-static umem_ops_t	pmem_ops = {
-	.mo_tx_free		= pmem_tx_free,
-	.mo_tx_alloc		= pmem_tx_alloc,
-	.mo_tx_add		= pmem_tx_add,
-	.mo_tx_xadd		= pmem_tx_xadd,
-	.mo_tx_add_ptr		= pmem_tx_add_ptr,
-	.mo_tx_abort		= pmem_tx_abort,
-	.mo_tx_begin		= pmem_tx_begin,
-	.mo_tx_commit		= pmem_tx_commit,
-	.mo_tx_stage		= pmem_tx_stage,
-	.mo_reserve		= pmem_reserve,
-	.mo_defer_free		= pmem_defer_free,
-	.mo_cancel		= pmem_cancel,
-	.mo_tx_publish		= pmem_tx_publish,
-	.mo_atomic_copy		= pmem_atomic_copy,
-	.mo_atomic_alloc	= pmem_atomic_alloc,
-	.mo_atomic_free		= pmem_atomic_free,
-	.mo_atomic_flush	= pmem_atomic_flush,
-	.mo_tx_add_callback	= umem_tx_add_cb,
+static umem_ops_t pmem_ops = {
+    .mo_tx_free         = pmem_tx_free,
+    .mo_tx_alloc        = pmem_tx_alloc,
+    .mo_tx_add          = pmem_tx_add,
+    .mo_tx_xadd         = pmem_tx_xadd,
+    .mo_tx_add_ptr      = pmem_tx_add_ptr,
+    .mo_tx_abort        = pmem_tx_abort,
+    .mo_tx_begin        = pmem_tx_begin,
+    .mo_tx_commit       = pmem_tx_commit,
+    .mo_tx_stage        = pmem_tx_stage,
+    .mo_tx_set_snapbuf  = pmem_tx_set_snapbuf,
+    .mo_reserve         = pmem_reserve,
+    .mo_defer_free      = pmem_defer_free,
+    .mo_cancel          = pmem_cancel,
+    .mo_tx_publish      = pmem_tx_publish,
+    .mo_atomic_copy     = pmem_atomic_copy,
+    .mo_atomic_alloc    = pmem_atomic_alloc,
+    .mo_atomic_free     = pmem_atomic_free,
+    .mo_atomic_flush    = pmem_atomic_flush,
+    .mo_tx_add_callback = umem_tx_add_cb,
 };
-
 
 /** BMEM operations (depends on dav) */
 
--- a/src/include/daos/mem.h
+++ b/src/include/daos/mem.h
@@ -1,6 +1,6 @@
 /**
- * (C) Copyright 2016-2024 Intel Corporation.
- * (C) Copyright 2025-2026 Hewlett Packard Enterprise Development LP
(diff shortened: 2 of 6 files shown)


gnailzenh:
This is just a draft PR, I'd like to get some feedbacks

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17850/1/execution/node/1287/log

gnailzenh:
@janekmi is this reasonable to you?

src/include/daos/mem.h
```diff
@@ -1080,6 +1082,13 @@ bool umem_tx_none(struct umem_instance *umm);
 
 int umem_tx_errno(int err);
 
+static inline void
```
NiuYawei, line 1085: error code should be returned

src/common/mem.c
```diff
@@ -1003,6 +1003,23 @@ pmem_tx_stage(void)
...
+	void *buf = umem_off2ptr(umm, snapbuf);
+	int   rc;
+
+	rc = pmemobj_tx_log_append_buffer(TX_LOG_TYPE_SNAPSHOT, buf, size);
+	if (rc)
+		return rc;
+
+	rc = pmemobj_tx_log_auto_alloc(TX_LOG_TYPE_SNAPSHOT, 0);
+	if (rc)
+		return rc;
```
janekmi, line 1018: As far as I can tell, this API is dedicated for testing whether the provided user buffer is actually sufficient for the size of transaction you expect it to handle.

My understanding is that the user-provided buffer is populated first anyway. So, by using it you prevent PMDK from creating an extra allocation internally even if it would be possible.

Maybe it would make a nice debug feature assuming you are confident the provided user buffer is big enough to handle what you plan to put there.

src/include/daos/mem.h
```diff
@@ -751,6 +751,8 @@ typedef struct {
 	int		 (*mo_tx_commit)(struct umem_instance *umm, void *data);
 
 #ifdef DAOS_PMEM_BUILD
+	/** Set emergency buffer for transaction snapshot */
```
janekmi, line 754: Nitpick. Currently you use it as an emergency feature but in general it can also be used to improve performance.

```suggestion
	/** Set a user buffer for the transaction snapshots */
```

src/vos/vos_layout.h
```diff
@@ -116,14 +116,21 @@ struct vos_gc_bkt_df {
...
+#define VOS_GC_SNAPBUF_EMERG                    (1 << 19)
 
 /* VOS pool durable format extension */
 struct vos_pool_ext_df {
 	/* Extension for GC bucket */
 	struct vos_gc_bkt_df	ped_gc_bkt;
 	/* Memory file size for md-on-ssd phase2 pool */
 	uint64_t                ped_mem_sz;
+	/* emergency buffer for GC */
+	umem_off_t              ped_gc_emerg;
```
janekmi, line 133: Should we bump `VOS_POOL_DF_*`?

src/vos/vos_pool.c
```diff
@@ -1444,6 +1444,13 @@ vos_pool_create_ex(const char *path, uuid_t uuid, daos_size_t scm_sz, daos_size_
 	/* pd_ext is newly allocated, no need to call tx_add_ptr() */
 	pd_ext_df             = umem_off2ptr(&umem, pool_df->pd_ext);
 	pd_ext_df->ped_mem_sz = scm_sz;
+	pd_ext_df->ped_gc_emerg = umem_zalloc(&umem, VOS_GC_SNAPBUF_EMERG);
+	if (UMOFF_IS_NULL(pd_ext_df->ped_gc_emerg)) {
+		D_ERROR("Failed to allocate pool GC emergency buffer.\n");
```
janekmi, line 1449: Nitpick.

```suggestion
		D_ERROR("Failed to allocate the pool GC emergency buffer.\n");
```

src/vos/vos_pool.c
```diff
@@ -1444,6 +1444,13 @@ vos_pool_create_ex(const char *path, uuid_t uuid, daos_size_t scm_sz, daos_size_
...
+		rc = -DER_NOSPACE;
+		goto end;
```
janekmi, line 1451: Unnecessary.

```suggestion
```

src/vos/vos_gc.c
```diff
@@ -876,6 +877,16 @@ gc_reclaim_pool(struct vos_pool *pool, int *credits, bool *empty_ret)
 		goto done;
 	}
 
+	if (pool->vp_gc_nospc != 0) {
+		struct vos_pool_ext_df *ext;
```
janekmi, line 881: Nitpick. I think it makes sense to keep both this and `struct vos_pool_df *pool_df` in the same scope either here or in the scope of the function. They serve the same purpose.

src/vos/vos_gc.c
```diff
@@ -876,6 +877,16 @@ gc_reclaim_pool(struct vos_pool *pool, int *credits, bool *empty_ret)
...
+
+		ext = umem_off2ptr(&pool->vp_umm, pool_df->pd_ext);
+		if (!UMOFF_IS_NULL(ext->ped_gc_emerg)) {
+			/* ignore the returned value */
+			umem_tx_set_snapbuf(&pool->vp_umm, ext->ped_gc_emerg, VOS_GC_SNAPBUF_EMERG);
```
janekmi, line 886: Nitpick. I guess this is a typical way of discarding the returned value. No need for a comment.

```suggestion
			(void) umem_tx_set_snapbuf(&pool->vp_umm, ext->ped_gc_emerg, VOS_GC_SNAPBUF_EMERG);
```

src/vos/vos_gc.c
```diff
@@ -2036,8 +2048,15 @@ vos_gc_pool(daos_handle_t poh, int credits, int (*yield_func)(void *arg),
...
 			D_ERROR("GC pool failed: " DF_RC "\n", DP_RC(rc));
-			if (rc == -DER_NOSPACE)
-				pool->vp_gc_nospc = 1;
+			if (rc == -DER_NOSPACE) {
+				if (pool->vp_gc_nospc == 0) {
+					D_INFO("Enter GC emergency mode (pool=" DF_UUID ")\n",
+					       DP_UUID(pool->vp_id));
+					pool->vp_gc_nospc = gc_emerg_cycles;
+				} else { /* switch back, nothing else we can do */
+					pool->vp_gc_nospc = 0;
```
janekmi, line 2057: I think in this case we need another error message to indicate we run out of options.

src/common/mem.c
```diff
@@ -1003,6 +1003,23 @@ pmem_tx_stage(void)
 	return pmemobj_tx_stage();
 }
 
+static int
+pmem_tx_set_snapbuf(struct umem_instance *umm, umem_off_t snapbuf, size_t size)
+{
+	void *buf = umem_off2ptr(umm, snapbuf);
+	int   rc;
+
+	rc = pmemobj_tx_log_append_buffer(TX_LOG_TYPE_SNAPSHOT, buf, size);
```
janekmi, line 1012: Are we 100% sure the problem is with the space required for snapshots (`TX_LOG_TYPE_SNAPSHOT`) vs the space required for intents (`TX_LOG_TYPE_INTENT`)?

From what I see in the GC code and in the destroy/free paths it calls there are not many huge snapshots happening. Unless I missed something?

If anything there is possibly a large number of small snapshots. But if the problem is with even small snapshots I can imagine it is with intents as well (they are small by nature).

janekmi (changes requested):

src/vos/vos_gc.c
```diff
@@ -2036,8 +2048,15 @@ vos_gc_pool(daos_handle_t poh, int credits, int (*yield_func)(void *arg),
 
 		if (rc) {
 			D_ERROR("GC pool failed: " DF_RC "\n", DP_RC(rc));
-			if (rc == -DER_NOSPACE)
-				pool->vp_gc_nospc = 1;
+			if (rc == -DER_NOSPACE) {
+				if (pool->vp_gc_nospc == 0) {
+					D_INFO("Enter GC emergency mode (pool=" DF_UUID ")\n",
+					       DP_UUID(pool->vp_id));
+					pool->vp_gc_nospc = gc_emerg_cycles;
```
Nasf-Fan, line 2055: It seems `gc_emerg_cycles` will never be refreshed, then why define it as a `static` variable? why not direct set `pool->vp_gc_nospc` as `32`?

src/common/mem.c
```diff
@@ -1003,6 +1003,23 @@ pmem_tx_stage(void)
...
+	if (rc)
+		return rc;
+
+	rc = pmemobj_tx_log_auto_alloc(TX_LOG_TYPE_SNAPSHOT, 0);
```
Nasf-Fan, line 1016: I am some confused about why use `0` (for disable) as the `on_off` parameter for `pmemobj_tx_log_auto_alloc()` under space pressure.

src/vos/vos_pool.c
janekmi: Resolved here: https://github.com/daos-stack/daos/pull/18039/changes#diff-17fc005c898cffbb0cdd39361cec1cf295fb2ec9af7e60a39f12f6839e9eba05

PR #17861, opened 2026-04-01
DAOS-18633 rebuild: global barrier for ec agg

wangshilong:
Pause EC agg globally before triggering rebuild and set rebuild fence after we confirmed all ec agg has been paused.

After rebuild, we don't need trigger full scan for ec agg any more, because we gurantee there is no ec aggregation after rebuild fence.

### Steps for the author:


src/rebuild/srv.c
```diff
@@ -3119,20 +3136,62 @@ rebuild_prepare_one(void *data)
...
+	struct ds_pool_child            *dpc;
+
+	dpc = ds_pool_child_lookup(rpt->rt_pool_uuid);
+	if (dpc == NULL) {
+		D_INFO(DF_UUID ": Local VOS pool isn't ready yet.\n", DP_UUID(rpt->rt_pool_uuid));
+		return -DER_STALE;
+	}
+
+	if (!dpc->spc_no_storage) {
+		dpc->spc_ec_agg_pause_gate = d_hlc_get();
```
NiuYawei, line 3171: Instead of acquiring a new HLC, can't we use the 'rebuild version' (pool version) to identify this rebuild?

src/rebuild/srv.c
```diff
@@ -3119,20 +3136,62 @@ rebuild_prepare_one(void *data)
...
+		D_INFO(DF_RB ": set local EC agg pause gate token " DF_U64 "\n", DP_RB_RPT(rpt),
+		       dpc->spc_ec_agg_pause_gate);
+		ds_cont_child_wait_ec_agg_pause(dpc, rebuild_wait_ec_pause);
```
NiuYawei, line 3174: This could lead to RPC timeout.

See agg_rate_ctl(), it should check the 'spc_ec_agg_pause_gate' as well? BTW, "rebuild_ver" looks a better name.

src/rebuild/srv.c
```diff
@@ -1793,14 +1792,33 @@ rebuild_leader_start(struct ds_pool *pool, struct rebuild_task *task,
...
+				    task->dst_rebuild_op, RB_SCAN_PREPARE);
 	if (rc)
+		goto cancel_pause;
+
+	D_INFO(DF_RB ": global EC agg pause completed\n", DP_RB_RGT(*p_rgt));
+
+	/* phase-2: broadcast scan RPC to all targets */
+	D_INFO(DF_RB ": start rebuild scan\n", DP_RB_RGT(*p_rgt));
+	rc = rebuild_scan_broadcast(pool, *p_rgt, &task->dst_tgts, task->dst_new_layout_version,
+				    task->dst_rebuild_op, RB_SCAN_START);
```
NiuYawei, line 1814: Given that we turned to use two rounds of scan broadcast, getting the global 'rebuild epoch' could be easier now, @liuxuezhao ?

src/rebuild/rebuild_internal.h
```diff
@@ -373,6 +380,12 @@ rebuild_tgt_status_check_ult(void *arg);
 int
 rebuild_tgt_prepare(crt_rpc_t *rpc, struct rebuild_tgt_pool_tracker **p_rpt);
 
+int
+rebuild_tgt_prepare_pause(crt_rpc_t *rpc, uint64_t *stable_epoch);
+
+int
+rebuild_tgt_cancel_pause(crt_rpc_t *rpc);
```
NiuYawei: rebuild_tgt_stop_agg()/rebuild_tgt_resume_agg() sounds more straightforward.

src/container/srv_target.c
```diff
@@ -381,38 +382,15 @@ cont_child_aggregate(struct ds_cont_child *cont, cont_aggregate_cb_t agg_cb,
 	if (cont->sc_snapshots_nr + 1 < MAX_SNAPSHOT_LOCAL) {
 		snapshots = snapshots_local;
 	} else {
-		D_ALLOC(snapshots, (cont->sc_snapshots_nr + 1) *
-			sizeof(daos_epoch_t));
+		D_ALLOC(snapshots, (cont->sc_snapshots_nr + 1) * sizeof(daos_epoch_t));
```
liuxuezhao: as rebuild_fence removed now, so looks need not + 1, for both L382 and 385

src/rebuild/srv.c
wangshilong: rebuild_version is not safe, as retied might reuse same rebuild_version.

src/container/srv_target.c
```diff
@@ -992,11 +970,13 @@ void
...
 {
 	uint64_t start_time = daos_wallclock_secs();
-	int      wait_intv  = 10;
+	int      wait_intv  = 1;
 	int      waited     = 0;
 
 	D_DEBUG(DB_MD, DF_UUID "[%d]: wait for pausing EC aggregation\n",
 		DP_UUID(pool_child->spc_uuid), dss_get_module_info()->dmi_tgt_id);
+	D_INFO(DF_UUID "[%d]: wait for local EC aggregation to pause\n",
+	       DP_UUID(pool_child->spc_uuid), dss_get_module_info()->dmi_tgt_id);
```
liuxuezhao, line 978: looks fine to use INFO, then L976 can remove

src/rebuild/rebuild_internal.h
wangshilong: OK

src/rebuild/srv.c
wangshilong: Problem is we don't guarantee rebuild_ver always bump for retrying rebuild. I don't want to change that Semantics. Or I bump it locally to ensure it increase every time.

src/rebuild/srv.c
```diff
@@ -2864,20 +2882,19 @@ rebuild_fini_one(void *arg)
...
+	/* Reset EC agg pause gate if this rebuild still owns it, then
+	 * set the aggregation end HLC so aggregation can restart.
 	 */
-	D_ASSERT(rpt->rt_rebuild_fence != 0);
-	if (rpt->rt_rebuild_fence == dpc->spc_rebuild_fence) {
-		dpc->spc_rebuild_fence = 0;
-		dpc->spc_rebuild_end_hlc = d_hlc_get();
+	if (rpt->rt_ec_pause_token == dpc->spc_ec_agg_pause_gate) {
+		dpc->spc_ec_agg_pause_gate = 0;
+		dpc->spc_rebuild_end_hlc   = d_hlc_get();
```
liuxuezhao: spc_rebuild_end_hlc is no usage now, looks fine to remove it also.

src/rebuild/srv.c
liuxuezhao: adding spc_ec_agg_pause_gate check to ds_pool_is_rebuilding() can make it affect agg_rate_ctl().

src/rebuild/srv.c
liuxuezhao: right, I also thought that, I can check some details, may push another PR or another commit to this PR later (will confirm with shilong which one is better).

liuxuezhao (dismissed):

NiuYawei (approved):

wangshilong:
Replace by https://github.com/daos-stack/daos/pull/17941

PR #17866, opened 2026-04-01
DAOS-18641 object: properly assemble obj shard ID when handle CPD RPC - b26 (#17865)

Nasf-Fan:
The old CPD RPC handler does not correctly assemble object shard ID when punch multiple shards that belong to the same object but locating on the same VOS target, as to some vos object shard will be repeatedly punched, but some others are leaked there.

Features: NvmePoolExtend
Allow-unstable-test: true

### Steps for the author:


Nasf-Fan:
The failure during fault injection testing on EL 8.8 is not related with this patch.

liuxuezhao (approved):

PR #17952, opened 2026-04-09
DAOS-18742 object: fix rebuild failure on container destroy

wangshilong:
Container destroy can cause fetch to return DER_SHUTDOWN, which may make rebuild fail. Normalize this case to DER_CONT_NONEXIST and let rebuild skip the error.

### Steps for the author:


liuxuezhao (dismissed):

src/object/srv_obj.c
```diff
@@ -2212,6 +2212,13 @@ obj_ioc_init(uuid_t pool_uuid, uuid_t coh_uuid, uuid_t cont_uuid, crt_rpc_t *rpc
 	/* load VOS container on demand for rebuild */
 	rc = ds_cont_child_lookup(pool_uuid, cont_uuid, &coc);
 	if (rc) {
+		if (rc == -DER_NONEXIST || rc == -DER_SHUTDOWN) {
```
liw: [Question] Is it possible that an engine performs a "graceful" shutdown during a rebuild task, produces a `-DER_SHUTDOWN`, causes the rebuild task to skip this container (do I understand this correctly?), and results in data loss (I've no specific idea how)? In other words, does `-DER_SHUTDOWN` always indicate that the container is being destroyed?

wangshilong: ds_cont_child_lookup is a pure memory lookup(create container is false), the only reason for DER_SHUTDOWN comes from container is being destroying state.

liw: Is it possible that `sc_stopping` is nonzero but `sc_detroying` is zero? E.g., is that possible in the "graceful" shutdown case I mentioned above?

wangshilong: you are right, let's differ that case and only ignore destroying case.

src/include/daos_srv/container.h
```diff
@@ -175,6 +175,24 @@ ds_cont_child_ec_aggregating(struct ds_cont_child *cont)
...
+{
+	return rc == -DER_CONT_DESTROYING;
+}
+
+static inline bool
+ds_cont_rc_is_ignorable_for_rebuild(int rc)
+{
+	return ds_cont_rc_is_nonexist(rc) || ds_cont_rc_is_destroying(rc);
+}
+
```
liw: [Nit] `ds_cont_rc_is_ignorable_for_rebuild` makes some sense, though it should perhaps belong to the `rebuild` module instead of the `ds_cont`. `ds_cont_rc_is_nonexist` and `ds_cont_rc_is_destroying` seem unnecessary, since either include only one obvious error code.

src/dtx/dtx_coll.c
```diff
@@ -58,17 +58,10 @@ dtx_coll_prep_ult(void *arg)
...
-			DF_DTI" with opc %u: "DF_RC"\n",
-			DP_UUID(dci->dci_po_uuid), DP_UUID(dci->dci_co_uuid),
-			DP_DTI(&dci->dci_xid), opc, DP_RC(dcpa->dcpa_result));
-		/*
-		 * Convert the case of container non-exist as -DER_IO to distinguish
-		 * the case of DTX entry does not exist. The latter one is normal.
-		 */
-		if (dcpa->dcpa_result == -DER_NONEXIST)
-			dcpa->dcpa_result = -DER_IO;
-
```
liw, line 71: @Nasf-Fan ^

src/object/srv_obj_migrate.c
```diff
@@ -1840,26 +1840,26 @@ migrate_get_cont_child(struct migrate_pool_tls *tls, uuid_t cont_uuid,
 		 */
 		rc = ds_cont_child_open_create(tls->mpt_pool_uuid, cont_uuid, false, &cont_child);
 		if (rc != 0) {
-			if (rc == -DER_SHUTDOWN || (cont_child && cont_child->sc_stopping)) {
+			if (rc == -DER_SHUTDOWN || rc == -DER_NONEXIST) {
```
liw: [Question] Why are we still treating `-DER_SHUTDOWN` as "destroying" here and in the `else` below? I thought this patch let `-DER_SHUTDOWN` indicate "stopping" and `-DER_CONT_DESTROYING` indicate "destroying"? Sorry, I'm not following.

wangshilong: Sorry, this is a leftover.. i think we don't need extra check here.

src/dtx/dtx_coll.c
wangshilong: Yup, since now we will return DER_CONT_NONEXIST and DER_CONT_DESTROYING for container, i think we don't need differ trick here.

src/include/daos_srv/container.h
wangshilong: OK, let me cleanup this as well.

src/object/srv_obj_migrate.c
```diff
@@ -3530,7 +3521,7 @@ migrate_obj_ult(void *data)
 	if (arg->epoch == DAOS_EPOCH_MAX)
 		tls->mpt_obj_count++;
 
-	if (rc == -DER_NONEXIST) {
+	if (rc == -DER_CONT_NONEXIST) {
```
liuxuezhao, line 3524: here should it also check/process the case of rc == -DER_CONT_DESTROYING?

wangshilong: Indeed, we should check to ignore DER_CONT_DESTROYING for above cases as well.

liw: [Question] Why not returning `-DER_CONT_NONEXIST` in the case where this patch currently returns `-DER_CONT_DESTROYING`? Does returning `-DER_CONT_DETROYING` help in some case?

wangshilong: Yup, i thought about that...Maybe we could return DER_CONT_NONEXIST for that case for now. and change it if we need DER_CONT_DESTROYING in the future.

liuxuezhao (approved):

liw (approved):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17952/15/testReport/

daosbuild3:
Test stage Functional Hardware Large MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17952/15/execution/node/1340/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17952/16/testReport/

wangshilong:
> Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17952/16/testReport/


@daos-stack/daos-gatekeeper  only failure is known as  https://daosio.atlassian.net/browse/SRE-3734
https://jenkins-3.daos.hpc.amslabs.hpecorp.net/blue/organizations/jenkins/daos-stack%2Fdaos/detail/PR-17952/16/tests

PR #17967, opened 2026-04-10
DAOS-18304 ddb: Introduce DdbAPI interface for Go unit testing

knard38:
#### Overview

Preliminary refactoring to enable Go unit test coverage for the ddb command layer.
The actual tests are added in the follow-up PR #18086.

A `DdbAPI` Go interface is introduced to decouple the command layer from the C VOS
implementation, so that tests can use a stub without requiring a live VOS environment.

This patch is also solving the ticket [DAOS-18581](https://daosio.atlassian.net/browse/DAOS-18581)

#### Go Refactoring

- **New `DdbAPI` interface**: all `ddbXxx()` free functions are converted to
  `(ctx *DdbContext) Xxx()` methods; `DdbAPI` is propagated through
  `addAppCommands()`, `createGrumbleApp()` and `parseOpts()`.
- **`DdbContext` lifecycle**: `Init()` becomes a method on a caller-allocated
  `DdbContext`, giving the caller (or a test stub) full control over the object.
- **Remove `dc_pool_path` / `dc_db_path` from `ddb_ctx`**: paths are now
  passed as explicit function arguments, removing the need for the `SetCString()`
  helper.
- **`parseOpts()` testability**: `os.Exit()` calls are removed; all errors are
  returned up to `main()`, making `parseOpts()` fully unit-testable.
- **Auto-open command list**: replace the `HasPrefix` loop with a `noAutoOpen`
  map for exact command-name matching.

#### C API Simplification

- **`dv_pool_open` / `dv_pool_destroy`**: signatures simplified from
  `(path, path_parts, poh, flags, write_mode)` to `(path, db_path, ctx, flags)`;
  `vos_file_parts` allocation is moved into a new internal
  `create_vos_file_parts()` helper.
- **VOS/DB path size**: `DB_PATH_SIZE` and `VOS_PATH_SIZE` raised from 256 to
  `PATH_MAX`; the original VOS path is now stored in `vf_vos_file_path` and its
  length is validated in `parse_vos_file_parts()`.

#### Bug Fixes

- **`ddb_run_feature()`**: `dc_poh` / `dc_write_mode` state reset is now guarded
  by the `close` flag, preventing corruption when the pool was already open;
  add missing error message on `dv_pool_open()` failure.
- **Non-interactive mode**: errors from `runCmdStr()` and `runFileCmds()` are now
  propagated instead of silently discarded.
- **`--version`**: prints directly instead of synthesising a fake command.
- **`--db_path` without `--vos_path`**: returns an explicit error instead of
  being silently ignored.
- **Pool close**: consolidated into a `closePoolIfOpen()` helper with `defer`.
- **`Feature()`**: fix memory leak on enable/disable C strings.
- **`DtxStat()`**: fix `defer freeString()` ordering.
- **Unknown command errors**: typed `unknownCmdError` struct for type-safe
  detection; `exitWithError()` now prints the command list on unknown commands
  and fixes the duplicated `ERROR:` prefix in fault resolution messages.
- **`rm_pool`**: make `--vos_path` mandatory; pass `db_path` explicitly.
- **Test fixes**: correct VOS tree path indices in `ls`, `value_dump` and
  `ilog_dump` tests; update `dtx_stat` message; restore pool handle in
  `helper_stat_open_modify_close_stat()` after close; fix `assert_memory_equal`
  in `read_only_vs_write_mode_test()` to compare PRE vs POST digest.

### Steps for the author:

### DIFF
--- a/src/control/cmd/ddb/commands_wrapper.go
+++ b/src/control/cmd/ddb/commands_wrapper.go
@@ -37,33 +37,35 @@ func freeString(s *C.char) {
 	C.free(unsafe.Pointer(s))
 }
 
-func SetCString(out **C.char, s string) func() {
-	cstr := C.CString(s)
-	*out = cstr
-
-	return func() {
-		C.free(unsafe.Pointer(cstr))
-	}
-}
-
-// InitDdb initializes the ddb context and returns a closure to finalize it.
-func InitDdb(log *logging.LeveledLogger) (*DdbContext, func(), error) {
-	// Must lock to OS thread because vos init/fini uses ABT init and finalize which must be called on the same thread
-	runtime.LockOSThread()
-
-	if err := daosError(C.ddb_init()); err != nil {
-		runtime.UnlockOSThread()
-		return nil, nil, err
-	}
-
-	ctx := &DdbContext{}
-	C.ddb_ctx_init(&ctx.ctx) // Initialize with ctx default values
-	ctx.log = log
-
-	return ctx, func() {
-		C.ddb_fini()
-		runtime.UnlockOSThread()
-	}, nil
+type DdbAPI interface {
+	Init(log *logging.LeveledLogger) (func(), error)
+	PoolIsOpen() bool
+	Ls(path string, recursive bool, details bool) error
+	Open(path string, dbPath string, writeMode bool) error
+	Version() error
+	Close() error
+	SuperblockDump() error
+	ValueDump(path string, dst string) error
+	Rm(path string) error
+	ValueLoad(src string, dst string) error
+	IlogDump(path string) error
+	IlogCommit(path string) error
+	IlogClear(path string) error
+	DtxDump(path string, active bool, committed bool) error
+	DtxCmtClear(path string) error
+	SmdSync(nvmeConf string, dbPath string) error
+	VeaDump() error
+	VeaUpdate(offset string, blkCnt string) error
+	DtxActCommit(path string, dtxID string) error
+	DtxActAbort(path string, dtxID string) error
+	Feature(path, dbPath, enable, disable string, show bool) error
+	RmPool(path string, dbPath string) error
+	DtxActDiscardInvalid(path string, dtxID string) error
+	DevList(dbPath string) error
+	DevReplace(dbPath string, oldDevid string, newDevid string) error
+	DtxStat(path string, details bool) error
+	ProvMem(dbPath string, tmpfsMount string, tmpfsMountSize uint) error
+	DtxAggr(path string, cmtTime uint64, cmtDate string) error
 }
 
 // DdbContext structure for wrapping the C code context structure
@@ -72,11 +74,30 @@ type DdbContext struct {
 	log *logging.LeveledLogger
 }
 
-func ddbPoolIsOpen(ctx *DdbContext) bool {
+// Init initializes the ddb context and returns a closure to finalize it.
+func (ctx *DdbContext) Init(log *logging.LeveledLogger) (func(), error) {
+	// Must lock to OS thread because vos init/fini uses ABT init and finalize which must be called on the same thread
+	runtime.LockOSThread()
+
+	if err := daosError(C.ddb_init()); err != nil {
+		runtime.UnlockOSThread()
+		return nil, err
+	}
+
+	C.ddb_ctx_init(&ctx.ctx) // Initialize with ctx default values
+	ctx.log = log
+
+	return func() {
+		C.ddb_fini()
+		runtime.UnlockOSThread()
+	}, nil
+}
+
+func (ctx *DdbContext) PoolIsOpen() bool {
 	return bool(C.ddb_pool_is_open(&ctx.ctx))
 }
 
-func ddbLs(ctx *DdbContext, path string, recursive bool, details bool) error {
+func (ctx *DdbContext) Ls(path string, recursive bool, details bool) error {
 	/* Set up the options */
 	options := C.struct_ls_options{}
 	options.path = C.CString(path)
@@ -87,33 +108,34 @@ func ddbLs(ctx *DdbContext, path string, recursive bool, details bool) error {
 	return daosError(C.ddb_run_ls(&ctx.ctx, &options))
 }
 
-func ddbOpen(ctx *DdbContext, path string, write_mode bool) error {
+func (ctx *DdbContext) Open(path string, dbPath string, writeMode bool) error {
 	/* Set up the options */
 	options := C.struct_open_options{}
 	options.path = C.CString(path)
 	defer freeString(options.path)
-	options.db_path = ctx.ctx.dc_db_path
-	options.write_mode = C.bool(write_mode)
+	options.db_path = C.CString(dbPath)
+	defer freeString(options.db_path)
+	options.write_mode = C.bool(writeMode)
 	/* Run the c code command */
 	return daosError(C.ddb_run_open(&ctx.ctx, &options))
 }
 
-func ddbVersion(ctx *DdbContext) error {
+func (ctx *DdbContext) Version() error {
 	/* Run the c code command */
 	return daosError(C.ddb_run_version(&ctx.ctx))
 }
 
-func ddbClose(ctx *DdbContext) error {
+func (ctx *DdbContext) Close() error {
 	/* Run the c code command */
 	return daosError(C.ddb_run_close(&ctx.ctx))
 }
 
-func ddbSuperblockDump(ctx *DdbContext) error {
+func (ctx *DdbContext) SuperblockDump() error {
 	/* Run the c code command */
 	return daosError(C.ddb_run_superblock_dump(&ctx.ctx))
 }
 
-func ddbValueDump(ctx *DdbContext, path string, dst string) error {
+func (ctx *DdbContext) ValueDump(path string, dst string) error {
 	/* Set up the options */
 	options := C.struct_value_dump_options{}
 	options.path = C.CString(path)
@@ -124,7 +146,7 @@ func ddbValueDump(ctx *DdbContext, path string, dst string) error {
 	return daosError(C.ddb_run_value_dump(&ctx.ctx, &options))
 }
 
-func ddbRm(ctx *DdbContext, path string) error {
+func (ctx *DdbContext) Rm(path string) error {
 	/* Set up the options */
 	options := C.struct_rm_options{}
 	options.path = C.CString(path)
@@ -133,7 +155,7 @@ func ddbRm(ctx *DdbContext, path string) error {
 	return daosError(C.ddb_run_rm(&ctx.ctx, &options))
 }
 
-func ddbValueLoad(ctx *DdbContext, src string, dst string) error {
+func (ctx *DdbContext) ValueLoad(src string, dst string) error {
 	/* Set up the options */
 	options := C.struct_value_load_options{}
 	options.src = C.CString(src)
@@ -144,7 +166,7 @@ func ddbValueLoad(ctx *DdbContext, src string, dst string) error {
 	return daosError(C.ddb_run_value_load(&ctx.ctx, &options))
 }
 
-func ddbIlogDump(ctx *DdbContext, path string) error {
+func (ctx *DdbContext) IlogDump(path string) error {
 	/* Set up the options */
 	options := C.struct_ilog_dump_options{}
 	options.path = C.CString(path)
@@ -153,7 +175,7 @@ func ddbIlogDump(ctx *DdbContext, path string) error {
 	return daosError(C.ddb_run_ilog_dump(&ctx.ctx, &options))
 }
 
-func ddbIlogCommit(ctx *DdbContext, path string) error {
+func (ctx *DdbContext) IlogCommit(path string) error {
 	/* Set up the options */
 	options := C.struct_ilog_commit_options{}
 	options.path = C.CString(path)
@@ -162,7 +184,7 @@ func ddbIlogCommit(ctx *DdbContext, path string) error {
 	return daosError(C.ddb_run_ilog_commit(&ctx.ctx, &options))
 }
 
-func ddbIlogClear(ctx *DdbContext, path string) error {
+func (ctx *DdbContext) IlogClear(path string) error {
 	/* Set up the options */
 	options := C.struct_ilog_clear_options{}
 	options.path = C.CString(path)
@@ -171,7 +193,7 @@ func ddbIlogClear(ctx *DdbContext, path string) error {
 	return daosError(C.ddb_run_ilog_clear(&ctx.ctx, &options))
 }
 
-func ddbDtxDump(ctx *DdbContext, path string, active bool, committed bool) error {
+func (ctx *DdbContext) DtxDump(path string, active bool, committed bool) error {
 	/* Set up the options */
 	options := C.struct_dtx_dump_options{}
 	options.path = C.CString(path)
@@ -182,7 +204,7 @@ func ddbDtxDump(ctx *DdbContext, path string, active bool, committed bool) error
 	return daosError(C.ddb_run_dtx_dump(&ctx.ctx, &options))
 }
 
-func ddbDtxCmtClear(ctx *DdbContext, path string) error {
+func (ctx *DdbContext) DtxCmtClear(path string) error {
 	/* Set up the options */
 	options := C.struct_dtx_cmt_clear_options{}
 	options.path = C.CString(path)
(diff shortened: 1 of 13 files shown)


daosbuild3:
Test stage Unit Test completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17967/1/testReport/

daosbuild3:
Test stage Unit Test completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-17967/3/testReport/

daosbuild3:
Test stage Functional on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17967/4/execution/node/1032/log

daosbuild3:
Test stage Unit Test bdev with memcheck completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-17967/7/display/redirect

daosbuild3:
Test stage Unit Test with memcheck completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos/job/PR-17967/7/display/redirect

daosbuild3:
Test stage Build on EL 9 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17967/11/execution/node/272/log

daosbuild3:
Test stage Build on EL 8 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17967/11/execution/node/280/log

daosbuild3:
Test stage Build on Leap 15 completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17967/11/execution/node/288/log

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17967/14/execution/node/1335/log

daosbuild3:
Test stage Functional Hardware Medium Verbs Provider MD on SSD completed with status FAILURE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net//job/daos-stack/job/daos/view/change-requests/job/PR-17967/14/execution/node/1356/log

src/control/cmd/ddb/commands_wrapper.go
```diff
@@ -37,46 +37,67 @@ func freeString(s *C.char) {
 	C.free(unsafe.Pointer(s))
 }
 
-func SetCString(out **C.char, s string) func() {
-	cstr := C.CString(s)
-	*out = cstr
+type DdbAPI interface {
```
tanabarr, line 40: https://go-proverbs.github.io/ "the bigger the interface, the weaker the abstraction", we definitely haven't always followed this advice in DAOS but it's worth noting

tanabarr (approved):
LGTM, not sure if we really need the large interface and there might be a better way of achieving the goal but that's not a blocking comment

src/control/cmd/ddb/commands_wrapper.go
kjacque: Agreed with Tom, but seeing how it's used, this seems like an acceptable intermediate step to enable better unit testing.

src/control/cmd/ddb/main.go
```diff
@@ -147,37 +167,25 @@ func printGeneralHelp(app *grumble.App, generalMsg string) {
...
+		return &unknownCmdError{cmd: opts.Args.RunCmd}
 	}
-	return true
+	return errHelpRequested
 }
 
 // Prints either general or command-specific help message.
 // Returns a reasonable return code in case the caller chooses to terminate the process.
-func printHelp(generalMsg string, opts *cliOptions) int {
+func printHelp(opts cliOptions, generalMsg string) error {
```
kjacque, line 179: Why pass opts by value instead of by reference?

src/control/cmd/ddb/main.go
```diff
@@ -216,7 +224,7 @@ func strToLogLevels(level string) (logging.LogLevel, engine.LogLevel, error) {
 	}
 }
 
-func newLogger(opts *cliOptions) (*logging.LeveledLogger, error) {
+func newLogger(opts cliOptions) (*logging.LeveledLogger, error) {
```
kjacque, line 227: Same question

src/control/cmd/ddb/main.go
```diff
@@ -261,94 +269,83 @@ func newLogger(opts *cliOptions) (*logging.LeveledLogger, error) {
...
-	p.ShortDescription = "daos debug tool"
-	p.LongDescription = ddbLongDescription
+func closePoolIfOpen(api DdbAPI, log *logging.LeveledLogger) {
+	if !api.PoolIsOpen() {
+		return
+	}
 
-	// Set the traceback level such that a crash results in
-	// a coredump (when ulimit -c is set appropriately).
-	debug.SetTraceback("crash")
```
kjacque, line 273: Is it OK to remove this? Looks like it's no longer in `parseOpts` below.

src/utils/ddb/tests/ddb_commands_tests.c
```diff
@@ -89,7 +89,7 @@ ls_cmd_tests(void **state)
 
 	/* printing a recx works */
 	dvt_fake_print_called = 0;
-	opt.path = "/[0]/[0]/[0]/[0]/[0]";
+	opt.path              = "/[0]/[0]/[0]/[1]/[0]";
```
kjacque, line 92: Why are we changing behavior here? IMO it is best to do the refactor and any desired behavior changes separately to ensure the refactor is transparent.

kjacque (changes requested):
This is a big patch doing several different things. I would like to see this split into 3 different PRs, one for each of the categories referenced by the description. For a refactor I would not expect to see any changes in tests, for example--we want to ensure the refactor did NOT change behavior. And I would like the C API improvements to be separated from the bugfixes if possible.

kjacque:
Additionally, any bugfixes touching Go code post-refactor provide you with a great opportunity to write new Go unit tests.

knard38:
> Additionally, any bugfixes touching Go code post-refactor provide you with a great opportunity to write new Go unit tests.

A first set of Go unit tests could be found in the follow-up PR #18086.
It covers, among others, the `unknownCmdError` handling and the `feature` command no-auto-open fix, taking advantage of the new `DdbAPI` stub.

The `Feature()` memory leak and `DtxStat()` defer ordering fixes are not specifically covered:
- the memory leak involves C string deallocation which cannot be easily observed from Go unit tests,
- the defer ordering fix is a code structure correction without easy way to assert on.

src/control/cmd/ddb/commands_wrapper.go
knard38: Agreed on the fact thata large interface weakens the abstraction.
However, since each sub-command maps to one method, the interface cannot easily be split further without losing the ability to stub the entire VOS layer in tests.
As kjacque noted, this is an intentional intermediate step: the follow-up PR #18086 shows exactly how the interface is used in unit tests.

src/control/cmd/ddb/main.go
knard38: `debug.SetTraceback("crash")` was not removed: it was moved up to `main()` (now line 367). This is intentional and actually better: it gets called as early  as possible before any other logic, and it keeps parseOpts() as a pure argument-parsing function with no side effects (which also makes parseOpts()
  unit-testable)

src/control/cmd/ddb/main.go
knard38: `cliOptions` is a small flat struct (7 simple fields + a nested struct).
`printHelp`, `printCmdHelp`, and `newLogger` are read-only (they never mutate opts).
From my understanding, passing a small non-mutated struct by value is idiomatic.                                                                                                                                                                                                                                              From what I have understand, it makes the function signature self-documenting (no mutation), eliminates nil-dereference risk, etc.                                                                                                                                                            The old pointer semantics were necessary because the old `parseOpts` modified `opts` in-place.
Now that `parseOpts` owns and creates `opts` locally, passing it by value to read-only helpers sounds to me a more idiomatic choice

src/control/cmd/ddb/main.go
knard38: See my previous comment.

src/utils/ddb/tests/ddb_commands_tests.c
knard38: I agree with you. While writing this PR I also found some bugs and fixed them along the way, which is why they ended up in the same patch.

However, the following changes are tightly coupled to the Go refactoring and cannot be split without leaving the code or tests in a broken intermediate state.
Removing `dc_pool_path`/`dc_db_path` from `ddb_ctx` is required by the Go layer which now passes paths explicitly; this forces the `dv_pool_open` / `dv_pool_destroy` signature change.
The signature change in turn requires adding `vf_vos_file_path` and raising path size constants in `ddb_parse`, which then cascades into:
- `ddb_commands.c` simplifications.
- The `dtx_stat` message change (the old message printed `ctx->dc_pool_path` which no longer exists).
- All test call-site updates, including the path index corrections (caused by moving `dvt_vos_insert_2_records_with_dtx` to `dcv_suit_setup` as part of the test setup rewrite).

The following parts are standalone bug fixes that could go in a separate PR:
- [ ] `ddb_run_feature` out-label guard (`ddb_commands.c`): `if (close)` braces preventing `dc_poh`/`dc_write_mode` from being reset when the pool  was not opened by this call.
- [ ] `assert_memory_equal` comparing PRE vs PRE (`ddb_vos_tests.c`): `read_only_vs_write_mode_test` was comparing the pre-digest against itself  instead of the post-digest.
- [ ] `tctx->dvt_poh` not restored (`ddb_vos_tests.c`): `helper_stat_open_modify_close_stat` was leaving the pool handle in an invalid state for the rest of the test suite.

Does it sounds reasonable to you ?

kjacque:
> > Additionally, any bugfixes touching Go code post-refactor provide you with a great opportunity to write new Go unit tests.
> 
> A first set of Go unit tests could be found in the follow-up PR #18086. It covers, among others, the `unknownCmdError` handling and the `feature` command no-auto-open fix, taking advantage of the new `DdbAPI` stub.
> 
> The `Feature()` memory leak and `DtxStat()` defer ordering fixes are not specifically covered:
> 
> * the memory leak involves C string deallocation which cannot be easily observed from Go unit tests,
> * the defer ordering fix is a code structure correction without easy way to assert on.

Cool, good to know there's another PR in the works with test coverage. I'd still like this PR split to separate the transparent refactoring from the behavior changes though.

src/control/cmd/ddb/main.go
kjacque: Makes sense. Thanks for the explanation.

src/control/cmd/ddb/main.go
kjacque: Ah, I see it now. Thanks. Yes, I agree that's better. :)

src/utils/ddb/tests/ddb_commands_tests.c
kjacque: If some bugfixes must be left in, could you separate them out into a different commit in the same PR? That would make it easier to review. Thanks.

knard38:
> > > Additionally, any bugfixes touching Go code post-refactor provide you with a great opportunity to write new Go unit tests.
> > 
> > 
> > A first set of Go unit tests could be found in the follow-up PR #18086. It covers, among others, the `unknownCmdError` handling and the `feature` command no-auto-open fix, taking advantage of the new `DdbAPI` stub.
> > The `Feature()` memory leak and `DtxStat()` defer ordering fixes are not specifically covered:
> > 
> > * the memory leak involves C string deallocation which cannot be easily observed from Go unit tests,
> > * the defer ordering fix is a code structure correction without easy way to assert on.
> 
> Cool, good to know there's another PR in the works with test coverage. I'd still like this PR split to separate the transparent refactoring from the behavior changes though.

I will do my best to make split this PR in smaller ones as small as possible.

src/utils/ddb/tests/ddb_parse_tests.c
```diff
@@ -127,6 +142,7 @@ parse_vos_file_parts_test_success(void **state)
 	/* Test with root path */
 	rc = parse_vos_file_parts("/" MOCKED_POOL_UUID_STR "/vos-0", NULL, &parts);
 	assert_rc_equal(rc, DER_SUCCESS);
+	assert_string_equal("/" MOCKED_POOL_UUID_STR "/vos-0", parts.vf_vos_file_path);
```
janekmi, line 145: Please apply everywhere Ctrl+F will find a match.


```suggestion
	assert_string_equal(MOCKED_VOS_PATH_STR, parts.vf_vos_file_path);
```

src/utils/ddb/ddb_parse.c
```diff
@@ -217,6 +218,15 @@ parse_vos_file_parts(const char *vos_path, const char *db_path,
 	}
 	D_ASSERT(rc == 0);
 
+	vos_path_len = strnlen(vos_path, VOS_PATH_SIZE);
+	if (vos_path_len >= VOS_PATH_SIZE) {
+		D_ERROR("VOS path '%s' too long: got=%zu, max=%d\n", vos_path, vos_path_len,
+			VOS_PATH_SIZE - 1);
+		rc = -DER_EXCEEDS_PATH_LEN;
+		goto out_preg;
```
janekmi, line 226: Since you are using `out_preg` label for both `vos_path` and `db_path` now it seems you have a memory leek when `vos_path` passes the test and get copied into the `vfp_tmp`.

src/utils/ddb/tests/ddb_vos_tests.c
```diff
@@ -1200,7 +1198,7 @@ read_only_vs_write_mode_test(void **state)
 	/** In read‑only mode, the pool contents remain unchanged, and its mtime stays the same. */
 	helper_stat_open_modify_close_stat(tctx, fs, false /** read-only */);
 	assert_int_equal(fs[FILE_STATE_PRE].stat.st_mtime, fs[FILE_STATE_POST].stat.st_mtime);
-	assert_memory_equal(fs[FILE_STATE_PRE].digest, fs[FILE_STATE_PRE].digest,
+	assert_memory_equal(fs[FILE_STATE_PRE].digest, fs[FILE_STATE_POST].digest,
```
janekmi, line 1201: This is interesting. I assume it changed after daos-stack/daos#17533 has landed.
How come this test has not failed previously?

src/control/cmd/ddb/main.go
```diff
@@ -261,94 +269,83 @@ func newLogger(opts *cliOptions) (*logging.LeveledLogger, error) {
...
+			"feature":     true,
+			"open":        true,
+			"close":       true,
+			"prov_mem":    true,
+			"smd_sync":    true,
+			"rm_pool":     true,
+			"dev_list":    true,
+			"dev_replace": true,
+		}
+		if !noAutoOpen[opts.Args.RunCmd] {
```
janekmi, line 331: Since there is no record with `false` value and it looks like not all commands actually have records in the `noAutoOpen` map, it seems a list would be a better solution. What do you think?

```suggestion
		noAutoOpen := []string{
			"feature",
			"open",
			"close",
			"prov_mem",
			"smd_sync",
			"rm_pool",
			"dev_list",
			"dev_replace",
		}
		if !slices.Contains(noAutoOpen, opts.Args.RunCmd) {
```

janekmi (changes requested):

src/utils/ddb/tests/ddb_vos_tests.c
knard38: Hard to fail when comparing the same memory segments ;-)
Hopefully it is not failing with POST.

src/control/cmd/ddb/main.go
knard38: Agree with you that using a dictionary is overkill in this situation and probably less efficient with this number of elements.

- [x] Use slice instead of dictionary

src/utils/ddb/tests/ddb_parse_tests.c
knard38: - [x] Use relevant macro `"/" MOCKED_POOL_UUID_STR "/vos-0"` → `MOCKED_VOS_PATH_STR`

src/utils/ddb/tests/ddb_vos_tests.c
janekmi: Oh. Now I see. Right.

src/utils/ddb/ddb_parse.c
knard38: Not sure to understand where you see some memory leak:

`vf_vos_file_path` and `vf_db_path` are fixed-size `char[]` arrays embedded directly inside `struct vos_file_parts` (see `ddb_parse.h` lines 20–21). The `memcpy` calls on lines 228 and 238 copy data into the struct's own storage — they do not allocate any additional heap memory.

The single `D_ALLOC_PTR(vfp_tmp)` at line 203 allocates one block that covers the entire struct including both string buffers, and the single `D_FREE(vfp_tmp)` at line 265 releases it all. Every exit path after the allocation either reaches `out_preg` (which calls `D_FREE`) or is `out` (reached only when
  `vfp_tmp` is NULL).

Let me know if I'm missing something!

src/utils/ddb/tests/ddb_parse_tests.c
knard38: Fixed with commit 7b3c6ac5b1750f5a9ef94c6a7531a3cfc31381d0

src/control/cmd/ddb/main.go
knard38: Fixed with commit b0f720750d

knard38:
> > > > Additionally, any bugfixes touching Go code post-refactor provide you with a great opportunity to write new Go unit tests.
> > > 
> > > 
> > > A first set of Go unit tests could be found in the follow-up PR #18086. It covers, among others, the `unknownCmdError` handling and the `feature` command no-auto-open fix, taking advantage of the new `DdbAPI` stub.
> > > The `Feature()` memory leak and `DtxStat()` defer ordering fixes are not specifically covered:
> > > 
> > > * the memory leak involves C string deallocation which cannot be easily observed from Go unit tests,
> > > * the defer ordering fix is a code structure correction without easy way to assert on.
> > 
> > 
> > Cool, good to know there's another PR in the works with test coverage. I'd still like this PR split to separate the transparent refactoring from the behavior changes though.
> 
> I will do my best to make split this PR in smaller ones as small as possible.

Fixed with creating PRs:
- #18125
- #18124

knard38:
As requested during review, this PR has been split into two smaller ones to
separate the bug fixes from the refactoring:

- **[#18125](https://github.com/daos-stack/daos/pull/18125)** — bug fixes only (memory leaks, defer ordering, pool handle corruption, test assertions)
- **[#18124](https://github.com/daos-stack/daos/pull/18124)** — main refactoring (`DdbAPI` interface, C `dv_pool_open` API simplification, `db_path` fix)

PR #18020, opened 2026-04-15
DAOS-18782 cart: Change mrc settings

frostedcmos:
- for all providers except verbs - disable mrc on servers, enable on clients
- for verbs -- disable mrc on servers and clients

### Steps for the author:

### DIFF
--- a/src/cart/crt_init.c
+++ b/src/cart/crt_init.c
@@ -567,13 +567,16 @@ prov_settings_apply(bool primary, crt_provider_t prov, crt_init_options_t *opt)
 	if (prov == CRT_PROV_OFI_TCP || prov == CRT_PROV_OFI_TCP_RXM)
 		file_limit_bump();
 
-	if (prov == CRT_PROV_OFI_CXI)
-		mrc_enable = 1;
+	/* Enable mrc on clients for all providers except verbs; disable on servers */
+	mrc_enable = 1;
+	if (crt_is_service() || prov == CRT_PROV_OFI_VERBS_RXM)
+		mrc_enable = 0;
 
 	crt_env_get(CRT_MRC_ENABLE, &mrc_enable);
 	if (mrc_enable == 0) {
-		D_INFO("Disabling MR CACHE (FI_MR_CACHE_MAX_COUNT=0)\n");
+		D_INFO("Disabling MR CACHE\n");
 		d_setenv("FI_MR_CACHE_MAX_COUNT", "0", 1);
+		d_setenv("UCX_RCACHE_ENABLE", "n", 1);
 	}
 
 	g_prov_settings_applied[prov] = true;


src/cart/crt_init.c
```diff
@@ -567,13 +567,16 @@ prov_settings_apply(bool primary, crt_provider_t prov, crt_init_options_t *opt)
 	if (prov == CRT_PROV_OFI_TCP || prov == CRT_PROV_OFI_TCP_RXM)
 		file_limit_bump();
 
-	if (prov == CRT_PROV_OFI_CXI)
-		mrc_enable = 1;
+	/* Enable mrc on clients for all providers except verbs; disable on servers */
+	mrc_enable = 1;
+	if (crt_is_service() || prov == CRT_PROV_OFI_VERBS_RXM)
+		mrc_enable = 0;
```
Michael-Hennecke, line 573: hmm, wouldn't the CRT_MRC_ENABLE evaluation below re-enable this on the servers if it's set in the engine env? is this intentional, or do we want to prevent that this is overridden on the server side?

src/cart/crt_init.c
```diff
@@ -567,13 +567,16 @@ prov_settings_apply(bool primary, crt_provider_t prov, crt_init_options_t *opt)
...
 
 	crt_env_get(CRT_MRC_ENABLE, &mrc_enable);
 	if (mrc_enable == 0) {
-		D_INFO("Disabling MR CACHE (FI_MR_CACHE_MAX_COUNT=0)\n");
+		D_INFO("Disabling MR CACHE\n");
 		d_setenv("FI_MR_CACHE_MAX_COUNT", "0", 1);
+		d_setenv("UCX_RCACHE_ENABLE", "n", 1);
 	}
```
Michael-Hennecke, line 580: do we want an "else" clause here to make sure we explicitly enable it when CRT_MRC_ENABLE is set to true, rather than relying on the assumption that the default is "enabled"?
(it's just me being paranoid of silent changes to defaults...)

frostedcmos: the CRT_MRC_ENABLE  variable is really a misnomer and probably should have been called CRT_MRC_DISABLE from the beginning.

enabling cache is more complicated than just setting those bits back to true. in ofi case a proper cache monitor needs to be set, along with proper max cache count and possibly other settings

so in reality we either disable mr cache or leave it unchanged from how system has it already set up

Michael-Hennecke: ok

src/cart/crt_init.c
frostedcmos: correct, we only control a default behavior, but allow people to override it via env

Michael-Hennecke (approved):

soumagne (approved):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-18020/1/testReport/

frostedcmos:
One failure due to curl download issue: https://daosio.atlassian.net/browse/DAOS-18813

PR #18023, opened 2026-04-15
DAOS-17306 doc: self-healing properties, interactive rebuild

kccain:
For the DAOS version 2.8 release, add two major sections to the DAOS Administrator's Guide:
- self-healing properties / policy controls (DAOS-17306)
- explicit / interactive rebuild control (DAOS-17281)

Doc-only: true

Signed-off-by: Kenneth Cain <kenneth.cain@hpe.com>
Signed-off-by: Tom Nabarro <thomas.nabarro@hpe.com>

### Steps for the author:


tanabarr:
had to force push to clean up the commit list, Hope that's okay

docs/admin/self_healing.md
```diff
@@ -80,6 +80,89 @@ refers roughly to the amount of time for:
...
+Pool health info:
+- Rebuild busy, 42 objs, 21 recs
+- Data redundancy: degraded
+```
+
+### Field Values
+
+| Value | Meaning |
+|-------|---------|
+| `normal` | All targets are UP and accessible; data redundancy is intact |
```
kccain, line 113:
```suggestion
| `normal` | No targets are DOWN; data redundancy is intact |
```

Reasoning is for targets that are being reintegrated, their state is UP (but not yet UP_IN). The data redundancy would likely be intact in this case, if a previous exclude rebuild finished successfully.

docs/admin/self_healing.md
```diff
@@ -80,6 +80,89 @@ refers roughly to the amount of time for:
...
+
+The `Data redundancy` field appears in the pool health information section:
+
+```
+Pool health info:
+- Rebuild idle
+- Data redundancy: normal
+```
+
+or when targets are excluded:
```
kccain, line 101:
```suggestion
or when targets are excluded and a corresponding rebuild has not yet completed:
```

docs/admin/self_healing.md
```diff
@@ -80,6 +80,89 @@ refers roughly to the amount of time for:
...
+
+```
+Pool health info:
+- Rebuild busy, 42 objs, 21 recs
+- Data redundancy: degraded
+```
+
```
kccain, line 108: Something I thought of, but no change requested
there are other cases like drain that would show Rebuild busy but Data redundancy: normal
And I guess extend might show something similar. But it's probably not worth enumerating too many cases actually.

docs/admin/self_healing.md
```diff
@@ -80,6 +80,89 @@ refers roughly to the amount of time for:
...
+Pool 6f450a68-8c7d-4da9-8900-02691650f6a2, ntarget=8, disabled=1, state=TargetsExcluded
+Pool health info:
+- Disabled ranks: 3
+- Rebuild idle
+- Data redundancy: degraded  ← Confirms exclusion completed
+```
+
+The combination of:
+- `Disabled ranks: 3` — Shows which rank was excluded
+- `Rebuild idle` — No automatic rebuild triggered (expected with exclude-only policy)
```
kccain, line 144:
```suggestion
- `Rebuild idle` or `Rebuild done`— idle if pool has not been rebuilt since system start, done if it has been rebuild for a prior change. No new automatic rebuild triggered for this exclusion (expected with exclude-only policy)
```

docs/admin/self_healing.md
```diff
@@ -0,0 +1,646 @@
...
+- `Data redundancy: degraded` — Confirms data redundancy is impaired and manual
+  intervention is needed
+
+To restore redundancy, manually trigger rebuild:
```
kccain, line 148:
```suggestion
To restore redundancy for the excluded targets, manually trigger rebuild. Refer to [Detailed Pool Operations: Interactive Rebuild Controls](rebuild_controls.md) for more information about rebuild start commands.
```

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,336 @@
...
+successfully finished.
+
+```bash
+$ dmg pool reintegrate --ranks=3 p1
+```
+
+Run `dmg pool query` in a loop (with short sleeps between commands):
+
+```bash
+$ dmg pool query --health-only p1
```
tanabarr, line 148: might be good to add comments like above to describe what's happening

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,336 @@
...
+```bash
+$ dmg system list-pools -v
+Label UUID                                 State           SvcReps   SCM Size SCM Used SCM Imbalance NVME Size NVME Used NVME Imbalance Disabled UpgradeNeeded? Rebuild State
+----- ----                                 -----           -------   -------- -------- ------------- --------- --------- -------------- -------- -------------- -------------
+p1    cdf27ec1-ed97-4aa6-a766-39a2ed2136a1 Ready           [0-1,4-6] 105 GB   37 GB    28%           0 B       0 B       0%             0/64     None           done
+p2    dcfe63e0-e10f-464d-b18c-0915e52e048c TargetsExcluded [0-2,6-7] 49 GB    4.6 GB   62%           0 B       0 B       0%             8/64     None           busy
+```
+
+**2.** Stop all pool rebuilds with one command, and confirm they have stopped.
+
```
tanabarr, line 229: the description of what we are expecting to see in the display output of the commands could be expanded on here for clarity

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,336 @@
...
+Pool dcfe63e0-e10f-464d-b18c-0915e52e048c, ntarget=64, disabled=8, leader=7, version=46, state=TargetsExcluded
+Pool health info:
+- Disabled ranks: 3
+- Rebuild stopped (state=idle, status=-2027)
+- Data redundancy: degraded
+```
+
+**3.** Restart the affected engine, reintegrate engine and monitor rebuilds
+until finished.
+
```
tanabarr, line 256: could explain that we are waiting for rejoin after restarting engine and show this in `dmg system query` before continuing. again could use a bit more description of what we are looking for in the list-pools command output

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,336 @@
...
+----- ----                                 ----- -------   -------- -------- ------------- --------- --------- -------------- -------- -------------- -------------
+p1    cdf27ec1-ed97-4aa6-a766-39a2ed2136a1 Ready [0-1,4-6] 120 GB   42 GB    31%           0 B       0 B       0%             0/64     None           done
+p2    dcfe63e0-e10f-464d-b18c-0915e52e048c Ready [0-2,6-7] 56 GB    4.9 GB   62%           0 B       0 B       0%             0/64     None           done
+```
+
+
+## Rebuild Stop Command Errors
+
+The rebuild stop command may return errors when it is issued at a time that it
+may not be able to handle the request. For example:
```
tanabarr, line 283: "For example:" looks odd here, maybe "The following subsections show some examples"?

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,336 @@
...
+When no rebuild is currently running, the command will report a "nonexist"
+error:
+
+```bash
+# system command applying to 2 pools
+$ dmg system rebuild stop
+System-rebuild stop request succeeded on 0 pools
+ERROR: dmg: system rebuild stop failed: pool-rebuild stop failed on pool p1: pool-rebuild stop failed: DER_NONEXIST(-1005): The specified entity does not exist, pool-rebuild stop failed on pool p2: pool-rebuild stop failed: DER_NONEXIST(-1005): The specified entity does not exist
+
+# single pool command
```
tanabarr, line 296: "single pool command for missing pool"?

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,336 @@
...
+### Rebuild Finished, Reclaim in Progress
+
+When the rebuild stage has successfully finished and is in its reclaim cleanup
+stage, `dmg` will report a busy error. For example when pools `p1` and `p2` are
+both done rebuilding and in the reclaim stage:
+
+```bash
+$ dmg system rebuild stop
+System-rebuild stop request succeeded on 0 pools
+ERROR: dmg: system rebuild stop failed: pool-rebuild stop failed on pool p1: pool-rebuild stop failed: DER_BUSY(-1012): Device or resource busy, pool-rebuild stop failed on pool p2: pool-rebuild stop failed: DER_BUSY(-1012): Device or resource busy
```
tanabarr, line 310: would it be useful for the administrator to print something about it being busy in reclaim?

docs/admin/self_healing.md
```diff
@@ -0,0 +1,646 @@
...
+
+Pool-level properties apply to pool-specific behavior and each pool has its own
+set of pool-properties, which can be read and written by the system
+administrator using `dmg pool get-prop` / `dmg pool set-prop`.
+
+### Self-heal Properties
+
+The `self_heal` property describes behavior associated with self-healing and can
+be set at both the system and pool levels. For a given pool to perform the
+recovery actions (map change, rebuild), the policy needs to be enabled at both
```
tanabarr, line 36: may be explain that "map change"=="exclude" for clarity

tanabarr (approved):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-18023/5/testReport/

docs/admin/rebuild_controls.md
kccain: Done

docs/admin/rebuild_controls.md
kccain: Done

docs/admin/rebuild_controls.md
kccain: Added the system query and explanations

docs/admin/rebuild_controls.md
kccain: Fixed.

docs/admin/rebuild_controls.md
kccain: I've revised the text - the nonexist refers to the condition that there is no running rebuild for the pool.

docs/admin/rebuild_controls.md
kccain: That would require a more specific new error code. Probably too late to do such a thing for the 2.8 release, and not sure it has significant benefit, since the stakes are low in this case (the rebuild that the admin wants to stop has actually already finished, and is just cleaning up before "success").

docs/admin/self_healing.md
kccain: added some words to explicitly mention exclude

tanabarr (dismissed):

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+* When a pool has insufficient free space to accommodate relocation of affected
+  data upon engine(s) excluded/drained. For example, if a rebuild fails with
+  `status=-1007` (`-DER_NOSPACE`) (that will likely repeat in its automatic
+  retries). Stopping such a rebuild allows an administrator to perform alternate
+  actions (e.g., directly reintegrate the lost engine(s); and/or pool expansion
+  to more engines).
+* A system with many tens of pools are all rebuilding simultaneously (requiring
+  substantial CPU and network resources in the DAOS system). And the
+  administrator deciding in such cases to stagger an overall recovery by
+  manually commanding pools to rebuild in smaller batches.
```
mchaarawi, line 42: just a thought for future improvement: this is a nice use case for having dmg pool rebuild stop/start accepting a list of pools.

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+not a pause/resume interface).
+
+Some circumstances where rebuild stop/start controls may be helpful are:
+
+* When a pool has insufficient free space to accommodate relocation of affected
+  data upon engine(s) excluded/drained. For example, if a rebuild fails with
+  `status=-1007` (`-DER_NOSPACE`) (that will likely repeat in its automatic
+  retries). Stopping such a rebuild allows an administrator to perform alternate
+  actions (e.g., directly reintegrate the lost engine(s); and/or pool expansion
+  to more engines).
```
mchaarawi: what if you get DER_NOSPACE on reintegrate rebuild ?

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+Upon stopping a pool's rebuild, its rebuild state as reported by `dmg pool query`
+will be an idle state, and an error status=`-2027` (`-DER_OP_CANCELED` DAOS error code).
+
+The effect of a `rebuild stop` command is "one shot", meaning only a pool's
+currently-running rebuild is stopped and there is no persistent effect on future
+operations. Subsequent self-healing automatic recovery, or administrator command
+(e.g., system stop, system/pool exclude, reintegrate, drain, pool extend) can
+trigger new rebuild(s). Also, if a rebuild is stopped, whatever progress it had
+made in reconstructing the data in the pool is not retained — a subsequent
+"rebuild start" command will start the rebuild from the beginning (i.e., this is
```
mchaarawi: what if a pool has failed rebuild with something random, let's say DER_STALE, or some network issue.
can someone just call rebuild start and it should restart, or would one need to call rebuild stop then rebuild start? or is that sort of use case not handled by this?

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+1. Run `op:Rebuild`
+2. Run `op:Fail_reclaim` to clean up
+    * If `Fail_reclaim` *itself* failed, retry `Fail_reclaim`
+    * If `Fail_reclaim` succeeded, retry the original `op:Rebuild`
+
+The `rebuild stop` commands are not typically allowed to terminate a rebuild in
+the `op:Reclaim` and `op:Fail_reclaim` phases — instead the command must be
+issued during the `op:Rebuild` execution. An exception is available with the
+`--force` option to `rebuild stop`, intended to be applied for rebuilds that
+repeatedly fail and possibly may even be looping `Fail_reclaim` operations.
```
mchaarawi: is there a reason to do it this way? why not just make it always stop (the force being the default)?
just from experience, usually when you want to stop a rebuild, there is usually always something wrong going on either in the rebuild phase, or  in the reclaim_fail phase.

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+
+Because of these details, carefully timing the execution of `rebuild stop`
+commands is needed, which can be facilitated with pool rebuild state querying
+with `dmg pool query`. See the section
+[Rebuild Stop Command Errors](#rebuild-stop-command-errors) for examples of
+errors returned by "rebuild stop" in different timing circumstances.
```
mchaarawi: There is a race condition here, and you cannot expect an admin to "time" commands to be issued.

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+**1.** Observe the pool `p1` is rebuilding after the fault is detected.
+
+The pool state reflects 8 disabled targets (`disabled=8`) corresponding to the exclusion
+of engine rank 3. Rebuild is underway (`busy` rebuild state).
+```bash
+$ dmg pool query --health-only p1
+Pool cdf27ec1-ed97-4aa6-a766-39a2ed2136a1, ntarget=64, disabled=8, leader=6, version=77, state=TargetsExcluded
+Pool health info:
+- Disabled ranks: 3
+- Rebuild busy, 1 objs, 0 recs
```
mchaarawi: just for the sake of this example, i would put something in recs and not 0. usually if it is 0 and objs is not 0, it means it is in reclaim or reclaim_fail.

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+**2.** Stop the rebuild and monitor with multiple `pool query` commands until the stop is confirmed
+(`stopped, state=idle, status=-2027`). Notice some rebuild state changes while waiting.
+
+Using a command at single-pool scope, no output is expected if the request is successfully
+sent to the storage system.
+```bash
+$ dmg pool rebuild stop p1
+```
+
+Run `dmg pool query` in a loop (with short delays between commands).
```
mchaarawi: this is really not a good practice. other than pool query sometimes hangs or takes a long time when rebuild is running, this also a pretty ugly way to administer a system.
also one can just cause a problem running multiple queries in large loops. 

anyway, i just think that we should not be asking anyone to do any pool operations in a loop.

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+rebuild is stopped. This is the final state reflecting that the rebuild is stopped
+(`state=idle, status=-2027`).
+```bash
+$ dmg pool query --health-only p1
+Pool cdf27ec1-ed97-4aa6-a766-39a2ed2136a1, ntarget=64, disabled=8, leader=6, version=77, state=TargetsExcluded
+Pool health info:
+- Disabled ranks: 3
+- Rebuild stopped (state=idle, status=-2027)
+- Data redundancy: degraded
+```
```
mchaarawi, line 154: an alternate outcome is that fail_reclaim fails or actually stuck.. what is the course of action in this case?
stop the rebuild with --force and start it again?
will reclaim_fail restart first in this case before rebuild?
maybe add that example here?

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,414 @@
...
+
+**1.** Observe the engine is excluded from the system, and pools begin
+rebuilding after the fault is detected.
+
+```bash
+$ dmg system query
+Rank      State
+----      -----
+[0-2,4-7] Joined
+3         Stopped
```
mchaarawi, line 251: Stopped will transition to Excluded and this is when rebuild kicks in.

mchaarawi (commented):
reviewed the rebuild_controls file only. 
will get to self_healing next.

docs/admin/self_healing.md
```diff
@@ -0,0 +1,646 @@
...
+
+This section describes some of the mechanisms using the `dmg` tool that can be
+employed to manage DAOS' self-healing feature. As a reminder, the system
+consists of a set of DAOS engine instances that join the system when they are
+started. An engine is assigned a numeric rank upon initially joining the system.
+This document may frequently interchange the term "rank" and "engine" to refer
+to a DAOS engine.
+
+By default, the system works as previous versions (per-pool `self_heal`
+property) if one doesn't modify the "new" system `self_heal` property. An
```
liw: [Nit] This assumes the current release to 2.8, which is not the case for master. Might reword 'the "new" system `self_heal` property' to something like 'the system `self_heal` property introduced in 2.8', and "previous versions" to something like "pre-2.8 versions".

docs/admin/self_healing.md
```diff
@@ -0,0 +1,646 @@
...
+
+System-level properties apply to system behavior and examples include system
+name and pool scrub mode. System-level properties are shared and centralized in
+the control-plane management service and can be read by the code in the DAOS
+engines as needed. Some properties are also writable. The administrator can read
+and write system properties using `dmg system get-prop` / `dmg system set-prop`.
+
+Pool-level properties apply to pool-specific behavior and each pool has its own
+set of pool-properties, which can be read and written by the system
+administrator using `dmg pool get-prop` / `dmg pool set-prop`.
```
liw, line 30: [Nit] Might want to remind users about the need to escape `;` (as in `exclude\;rebuild`).

docs/admin/self_healing.md
```diff
@@ -0,0 +1,646 @@
...
+needs to be enabled at both the system and per-pool level.
+
+The system level property has flags `exclude;pool_exclude;pool_rebuild` which
+indicate the following:
+
+* **`exclude`** — Whether engines get excluded from the system membership based
+  on (SWIM) activity detection.
+* **`pool_exclude`** — Whether when engine/target states change (e.g., as a
+  result of engine exclusion from the system), all relevant affected pools' pool
+  maps will be automatically updated (or not).
```
liw, line 51: [Nit] For both "exclude"s, I feel that for users an important info is to know this: When disabled, future operations (pool create, object update, rebuild tasks, etc.) may time out and retry until either the exclusion happens or the unavailable targets/engines become available again.

docs/admin/self_healing.md
```diff
@@ -0,0 +1,646 @@
...
+  `pool.exclude` are set to true.
+* **`pool.rebuild`** — Related to the `system.pool_rebuild` property and
+  together they determine if rebuild activity will be automatically triggered
+  upon a pool map update. Automatic rebuild on a pool will happen iff
+  `system.pool_rebuild` and `pool.rebuild` are set to true.
+* **`delay_rebuild`** — Mutually exclusive with `rebuild` and specifies that
+  rebuild does not necessarily trigger automatically and can be delayed based on
+  user requirements. Delay rebuild is mostly out of scope for this section.
+
+On starting a DAOS system and pool creation, default `self_heal` flags will be
```
liw: Hmm, you mean "on **creating** a DAOS system"? If one modifies system `self_heal` to, say, `exclude`, then restarting the system will not "reset" this property to `exclude;pool_exclude;pool_rebuild`.

docs/admin/self_healing.md
```diff
@@ -0,0 +1,646 @@
...
+* Two SWIM protocol periods to occur,
+* Plus a SWIM suspicion timeout to occur,
+* Plus a `CRT_EVENT_DELAY` number of seconds to elapse, during which failure
+  events can be aggregated into a logical event,
+* Plus some margin of error.
+
+
+## Pool Query Data Redundancy Status
+
+**Available in:** DAOS 2.6+
```
liw: [Nit] 2.6 doesn't have it, IIRC. So perhaps clearer to say something like "2.8 and later" or ">= 2.8".

docs/admin/self_healing.md
```diff
@@ -0,0 +1,646 @@
...
+
+The `dmg pool query` command displays the pool's data redundancy status as part
+of the health information output. This field provides a clear indication of
+whether the pool has sufficient target availability to maintain data redundancy.
```
liw: [Question] Is it really accurate? My understanding is that there may be sufficient targets available, but the data redundancy can still be degraded, for example, because the rebuild task is still in progress, or is temporarily disabled.

docs/admin/rebuild_controls.md
kccain: Sorry, I think I got the documentation wrong here and in the last section describing command errors. I'll fix it up. When `rebuild stop` occurs during `Fail_reclaim`, it succeeds silently (no error returned to the administrator), though it does allow the `Fail_reclaim` to continue to its own (hopefully successful) completion. After which, any normal retry mechanism will be suppressed (it remembers that the command was issued during `Fail_Reclaim`).

The more pathological case of `Fail_reclaim` itself failing and retrying itself is what the `--force` option is for. This one will only allow the `Fail_reclaim` to be forcibly terminated if it has failed at least once (i.e, an admin can't stop the first `Fail_reclaim` before we even know if that one will fail, and get into the pathological case).

I was being very conservative in the implementation to make sure to run any reclaim behavior during stop, to avoid causing potential further problems due to not freeing up space after a rebuild. One idea for the future may be to always force stop `Fail_reclaim`, with the idea that it "should be ok" to combine its cleanup work with a reclaim following the next rebuild. Obviously, would require some more careful code analysis and testing to be sure.

docs/admin/rebuild_controls.md
kccain: rebuild start invokes the same logic as a pool leader switch or system restart — it launches rebuilds for any targets found in relevant states (DOWN, DRAINING, or UP). But I don't think it's needed here.
			
For transient errors like -DER_STALE or network errors, the retry is automatic. No pool map revert occurs, so the retry finds targets in the same state as the original operation (possibly with additional targets in new states from pool map changes that occurred during the failed rebuild). No rebuild start needed.

For other (general) errors, the pool map is reverted after failure (excluded targets stay DOWN; reintegrating UP → DOWN/DOWNOUT; draining DRAIN → UPIN; extending UP → NEW). A retry rebuild is also launched automatically, but for reintegrate/drain/extend the targets are no longer in a rebuildable state, so the retry quickly self-cancels. To redo those operations, the original dmg command (reintegrate, drain, extend) would need to be reissued.

docs/admin/self_healing.md
tanabarr: addressed

docs/admin/self_healing.md
tanabarr: addressed

docs/admin/self_healing.md
tanabarr: addressed

docs/admin/self_healing.md
tanabarr: re. "exclusion happens" isn't this dependent on the exclusion flag in question being enabled? "or the unavailable targets/engines become available again" doesn't immediately make sense to me in the context of the "exclude" flags, isn't that just a general statement?

docs/admin/self_healing.md
tanabarr: addressed

docs/admin/self_healing.md
tanabarr: addressed

docs/admin/rebuild_controls.md
kccain: I'm not sure what to expect. I imagine if it is a direct reintegration of the excluded targets shortly before, then getting `-DER_NOSPACE` may be unlikely (since the data was placed there originally)?

docs/admin/rebuild_controls.md
kccain: In the typical case (`Rebuild` fails, `Fail_reclaim` runs successfully, Rebuild is retried) - if the stop command is received during `Fail_reclaim` it will be handled gracefully without undue burden on the administrator to time the command. I'll update the text to fix my inaccurate description.

docs/admin/rebuild_controls.md
kccain: will fix to emphasize pool query can be used to inspect rebuild state, and with advice to not run it too frequently, and to consider using `--health-only` option.

docs/admin/rebuild_controls.md
mchaarawi: it definitely is likely and ran into that issue several times :-)
probably a bug, but sometimes we get aroudn that issue by restarting the system or reintegrate only 1 pool at a time.

docs/admin/rebuild_controls.md
mchaarawi: i would not agree that all network errors the retry is automatic. we have seen some failure of rebuild with DER_HG or DER_NOMEM (which come s from the network).

docs/admin/rebuild_controls.md
mchaarawi: i do not think you need to get into that. just avoid really saying that one needs to query in a loop.

docs/admin/rebuild_controls.md
kccain: The action is to run rebuild stop with `--force` and start it again. When restarted, it will start from the beginning with `Rebuild` (not `Fail_reclaim`).

I can try to put a small example in for this, though it may be thin on command outputs
and may just have to describe the conditions seen in pool query output.

docs/admin/rebuild_controls.md
kccain: added in latest push

docs/admin/rebuild_controls.md
kccain: This is my best understanding of the code. It's based on looking at the behavior of  `retry_rebuild_task()` and the context in which it is invoked. I'd have to see particular failure instances in which a rebuild retry was not actually launched as a result of one of these failures, to be able to better see why it is not true.

docs/admin/rebuild_controls.md
kccain: done in latest push

tanabarr (dismissed):

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,457 @@
...
+made in reconstructing the data in the pool is not retained — a subsequent
+`rebuild start` command will start the rebuild from the beginning (i.e., this is
+not a pause/resume interface).
+
+Some circumstances where rebuild stop/start controls may be helpful are:
+
+* When a pool has insufficient free space to accommodate relocation of affected
+  data upon engine(s) excluded/drained. For example, if a rebuild fails with
+  `status=-1007` (`-DER_NOSPACE`) (that will likely repeat in its automatic
+  retries). Stopping such a rebuild allows an administrator to perform alternate
```
liuxuezhao: this is mismatch with real processing, it will not automatically retry for ENOSPACE err.
you may change the descripted case to temporarily network error or sth else retryable err.

docs/admin/rebuild_controls.md
liuxuezhao: if reint got DER_NOSPACE it should fail and revert pool map change of reint and will not automatically retry, admin can redo the reintegrate later.

"it definitely is likely and ran into that issue several times"
probably it was other err caused retry, and finally hit ENOSPACE? from code reading it should not automatically retry ENOSPACE case anyway.

docs/admin/rebuild_controls.md
```diff
@@ -0,0 +1,457 @@
...
+
+1. Run `op:Rebuild`
+2. Run `op:Reclaim` to clean up
+
+For a **failed rebuild**, the sequence is:
+
+1. Run `op:Rebuild`
+2. Run `op:Fail_reclaim` to clean up
+    * If `Fail_reclaim` *itself* failed, retry `Fail_reclaim`
+    * If `Fail_reclaim` succeeded, retry the original `op:Rebuild`
```
liuxuezhao: + if the rebuild was failed with retryable error.

kccain: Done in the latest commit.

docs/admin/rebuild_controls.md
kccain: OK I have rewritten this item to focus on the case of ongoing network failures that will cause all types of rebuilds to (actually) be retried. This is a simpler case to explain than other more general errors like `-DER_NOSPACE` (that are *effectively* not retried for drain/reintegrate/extend variants, but that actually are retried for exclude).

docs/admin/self_healing.md
liw: > re. "exclusion happens" isn't this dependent on the exclusion flag in question being enabled?

Yes.

> "or the unavailable targets/engines become available again" doesn't immediately make sense to me in the context of the "exclude" flags, isn't that just a general statement?

Hmm, I’m not sure if this is specific or general, but let me try again: If auto exclusions are disabled (either at the pool or the system level), then when an engine / target becomes unavailable, operations may time out and retry until the engine / target is excluded (either manually or by re-enabling auto exclusions) or until the engine or target becomes available again. I hope this expanded version makes more sense.

tanabarr (approved):

liw (approved):

liuxuezhao (approved):

PR #18039, opened 2026-04-18
DAOS-18690 vos: handle DTX commit under space pressure

Nasf-Fan:
If we cannot normally allocate space to hold committed DTX table, then release some old DTX entries from current container to hold new committed ones.

The patch also preallocates some space for TX snapshots. Related logic, such as DTX commit and maybe GC, will switch to emergency mode and use the preallocated buffer in case of space pressure.

### Steps for the author:


daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-18039/3/testReport/

Nasf-Fan:
> Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE. https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-18039/3/testReport/

test_dfuse_daos_build_wb failed for DAOS-18813, not related with the patch.

janekmi:
I understand this PR incorporates the solution created here: https://github.com/daos-stack/daos/pull/17850
But also gives a chance the system to recover in case we run out of space already. Is this the primary reason why we decided not to stop at the solution proposed by #17850 or is there some additional reasoning?

Nasf-Fan:
> I understand this PR incorporates the solution created here: #17850 But also gives a chance the system to recover in case we run out of space already. Is this the primary reason why we decided not to stop at the solution proposed by #17850 or is there some additional reasoning?

Yes, I discussed with Liang, the consideration is that: PR#17850 is some purpose drat for idea feedback, not sure when can be completed. On the other hand, DTX is the key point on the space pressure cycle. Because we often hit the cases like that VOS aggregation is blocked by some non-committed DTX entries, as to it cannot merge some records to release space, then GC will be further affected. And DTX commit logic is also blocked since no space can be released via VOS aggregation and GC. This patch tries to break such bad cycle via DTX internal logic and the pre-allocation idea from PR#17850.

After this patch done, we will rework PR#17850 to handle GC related things in further, that will be relatively easy then.

src/include/daos/mem.h
```diff
@@ -1074,12 +1086,38 @@ umem_tx_end(struct umem_instance *umm, int err)
 	return umem_tx_end_ex(umm, err, NULL);
 }
 
+static inline void
+umem_tx_set_failure_behavior(struct umem_instance *umm, enum umem_tx_failure_behavior behavior)
+{
+	if (umm->umm_ops->mo_tx_set_failure_behavior)
+		return umm->umm_ops->mo_tx_set_failure_behavior(behavior);
```
janekmi: This interface returns just void.


```suggestion
		umm->umm_ops->mo_tx_set_failure_behavior(behavior);
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1039,169 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+		 * committed DTX blob from current container. That maybe unfair since some other
+		 * container may have more old committed DTX entries, but scanning the target to
+		 * find out the pool/container that have the oldest DTX entry may take sometime,
+		 * depends on the pool/container count on the target. Under space pressure, such
+		 * behavior may be not worth, we just choose current container as the victim to
+		 * release space. It may be optimized in the future. DAOS-18690
+		 */
+
+		if (unlikely(cont_df->cd_dtx_committed_head == cont_df->cd_dtx_committed_tail)) {
+			D_ERROR("No space when extend commit DTX table.\n");
```
janekmi: Nitpick.

```suggestion
			D_ERROR("No space when extending the committed DTX table.\n");
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1039,169 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+	dbd->dbd_magic = DTX_CMT_BLOB_MAGIC;
+	dbd->dbd_cap   = (DTX_CMT_BLOB_SIZE - sizeof(struct vos_dtx_blob_df)) /
+		       sizeof(struct vos_dtx_cmt_ent_df);
+	dbd->dbd_count = 0;
+	dbd->dbd_index = 0;
+
+out:
+	if (allocated)
+		DL_CDEBUG(rc == 0, DB_IO, DLOG_ERR, rc,
+			  "Allocate DTX committed blob %p (" UMOFF_PF ") for cont " DF_UUID, dbd,
```
janekmi:
```suggestion
			  "Allocated DTX committed blob %p (" UMOFF_PF ") for cont " DF_UUID, dbd,
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1039,169 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+			  UMOFF_P(dbd_off), DP_UUID(cont->vc_id));
+	else
+		DL_CDEBUG(rc == 0, DLOG_WARN, DLOG_ERR, rc,
+			  "SPACE PRESSURE! reuse DTX blob %p (" UMOFF_PF ") for cont " DF_UUID, dbd,
```
janekmi: 1. I am not sure about this all caps part. 🤔
2. Nitpick.

```suggestion
			  "SPACE PRESSURE! Reusing DTX blob %p (" UMOFF_PF ") for cont " DF_UUID, dbd,
```

src/vos/vos_layout.h
```diff
@@ -116,14 +116,21 @@ struct vos_gc_bkt_df {
 /** 2.8 features */
 #define VOS_POOL_FEAT_2_8			(VOS_POOL_FEAT_GANG_SV)
 
-#define VOS_POOL_EXT_DF_PADDING_SIZE            53
+#define VOS_POOL_EXT_DF_PADDING_SIZE            52
+
+/* 512K preallocated buffer for TX snapshot in case of space emergency.
```
janekmi:
```suggestion
/* Preallocated 512 KB buffer for transaction snapshots under space pressure.
```

src/vos/vos_layout.h
```diff
@@ -116,14 +116,21 @@ struct vos_gc_bkt_df {
...
+ */
+#define VOS_SNAPBUF_EMERG                       (1 << 19)
 
 /* VOS pool durable format extension */
 struct vos_pool_ext_df {
 	/* Extension for GC bucket */
 	struct vos_gc_bkt_df	ped_gc_bkt;
 	/* Memory file size for md-on-ssd phase2 pool */
 	uint64_t                ped_mem_sz;
+	/* emergency buffer for GC */
```
janekmi:
```suggestion
	/* emergency buffer for GC TX snapshots */
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1039,169 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+		 * We almost run out of space. Under such case, we have to release some space to
+		 * make current DTX commit to go ahead; otherwise, non-committed DTX will affect
+		 * VOS aggregation as to no more space can be released via VOS aggregation. That
+		 * will make things to be worse. The most direct solution is to release some old
+		 * committed DTX blob from current container. That maybe unfair since some other
+		 * container may have more old committed DTX entries, but scanning the target to
+		 * find out the pool/container that have the oldest DTX entry may take sometime,
+		 * depends on the pool/container count on the target. Under space pressure, such
+		 * behavior may be not worth, we just choose current container as the victim to
+		 * release space. It may be optimized in the future. DAOS-18690
```
janekmi: Wording refinement suggestion:

> Space is nearly exhausted. In this case, we must reclaim space to let the current DTX commit proceed; otherwise, uncommitted DTXs can block VOS aggregation and prevent further space recovery. The most direct approach is to reclaim an old committed DTX blob from the current container. This may be unfair because another container could hold older committed DTX entries. However, scanning all pools/containers on the target to find the globally oldest entry can be expensive under space pressure. For now, we select the current container as the victim. This can be optimized later. DAOS-18690.

src/common/mem.c
```diff
@@ -997,6 +997,56 @@ pmem_defer_free(struct umem_instance *umm, umem_off_t off, void *act)
...
+	behavior = pmemobj_tx_get_failure_behavior();
+
+	switch (behavior) {
+	case POBJ_TX_FAILURE_ABORT:
+		return TX_FAILURE_ABORT;
+	case POBJ_TX_FAILURE_RETURN:
+		return TX_FAILURE_RETURN;
+	default:
+		D_ERROR("Unknown TX failure behavior %d\n", behavior);
+		return -DER_INVAL;
```
janekmi: It is very very unlikely but when it would happen it means something is very very wrong. I think we should stop at this point.


```suggestion
		D_ASSERTF(0, "Unknown TX failure behavior %d\n", behavior);
```

src/common/mem.c
```diff
@@ -997,6 +997,56 @@ pmem_defer_free(struct umem_instance *umm, umem_off_t off, void *act)
...
+
+static int
+pmem_tx_set_snapbuf(struct umem_instance *umm, umem_off_t snapbuf, size_t size)
+{
+	void *buf = umem_off2ptr(umm, snapbuf);
+	int   rc;
+
+	rc = pmemobj_tx_log_append_buffer(TX_LOG_TYPE_SNAPSHOT, buf, size);
+	if (rc != 0)
+		return rc;
```
janekmi:
```suggestion
		return umem_tx_errno(rc);
```

src/common/mem.c
```diff
@@ -997,6 +997,56 @@ pmem_defer_free(struct umem_instance *umm, umem_off_t off, void *act)
...
+
+	rc = pmemobj_tx_log_auto_alloc(TX_LOG_TYPE_SNAPSHOT, 0);
+	if (rc != 0)
+		return rc;
```
janekmi: IMHO this is more of a debug feature. Designed to make sure the `pmemobj_tx_log_append_buffer()` call actually works. I think we should not use it.

Please note you either would like it use extra memory if necessary or you do not care whether it failed because it run out of the memory provided by you or the total process memory. Unless you introduce a test which makes sure the provided buffer is actually in use, this call does not make much sense to me.

```suggestion
```

src/tests/suite/daos_base_tx.c
```diff
@@ -1024,6 +1024,83 @@ dtx_23(void **state)
...
+	 */
+	for (i = 'N'; i <= 'Z'; i++) {
+		snprintf(akey, TSIZE - 1, "akey-%c", i);
+		for (j = 'a'; j <= 'm'; j++) {
+			memset(wbuf, j, TSIZE);
+			lookup_single(dkey, akey, (j - 'a') * TSIZE, rbuf, TSIZE, DAOS_TX_NONE,
+				      &req);
+			assert_memory_equal(wbuf, rbuf, TSIZE);
+		}
+	}
```
janekmi, line 1098: In the comment you say the lookup will fail but I cannot see the lookup failure in the loop. I probably do not understand something. Can you please explain what I do not see here?

src/vos/vos_layout.h
```diff
@@ -116,14 +116,21 @@ struct vos_gc_bkt_df {
...
+	umem_off_t              ped_emerg_buf;
```
janekmi, line 133: Should we bump the pool layout version? 🤔

src/dtx/dtx_srv.c
```diff
@@ -196,19 +196,12 @@ dtx_handler(crt_rpc_t *rpc)
 		if (unlikely(din->di_epoch == 1))
 			D_GOTO(out, rc = -DER_IO);
 
-		while (i < din->di_dtx_array.ca_count) {
-			if (i + count > din->di_dtx_array.ca_count)
-				count = din->di_dtx_array.ca_count - i;
+		rc1 = dtx_commit_large(cont->sc_hdl, (struct dtx_id *)(din->di_dtx_array.ca_arrays),
```
janekmi: Nitpick. But it took me some time to understand you are not calling a function here just doing a type casting. 🤣

```suggestion
		rc1 = dtx_commit_large(cont->sc_hdl, (struct dtx_id *)din->di_dtx_array.ca_arrays,
```

src/dtx/dtx_srv.c
```diff
@@ -196,19 +196,12 @@ dtx_handler(crt_rpc_t *rpc)
...
+				       din->di_dtx_array.ca_count, false, NULL);
```
janekmi: You are passing `uint64_t ca_count` value as `int cnt` paramter. Two problems:

- mixing signed and unsigned values
- missing explicitly sized and architecture-dependent type

Please make it straight.

src/vos/vos_dtx.c
```diff
@@ -420,6 +420,31 @@ static btr_ops_t dtx_committed_btr_ops = {
...
+vos_dtx_add_ptr(struct vos_pool *pool, void *ptr, size_t size)
+{
+	struct umem_instance *umm = &pool->vp_umm;
+	int                   rc;
+
+	rc = umem_tx_add_ptr(umm, ptr, size);
+#ifdef DAOS_PMEM_BUILD
+	if (unlikely(rc == -DER_NOSPACE)) {
+		struct vos_pool_ext_df *ext_df   = umem_off2ptr(umm, pool->vp_pool_df->pd_ext);
+		int                     behavior = umem_tx_get_failure_behavior(umm);
```
janekmi, line 433: In the `pmem_tx_get_failure_behavior()` you assume it can return an error. IMHO it shouldn't. Either make sure it does not return an error or check for errors here.

src/vos/vos_dtx.c
```diff
@@ -420,6 +420,31 @@ static btr_ops_t dtx_committed_btr_ops = {
...
+
+		if (ext_df != NULL && !UMOFF_IS_NULL(ext_df->ped_emerg_buf) &&
+		    behavior == TX_FAILURE_RETURN) {
+			rc = umem_tx_set_snapbuf(umm, ext_df->ped_emerg_buf, VOS_SNAPBUF_EMERG);
```
janekmi: I was a little bit confused here since the constant's name does indicate it is actually a size. Feel free to name it differently but I would like to have the `_SIZE` suffix here.

```suggestion
			rc = umem_tx_set_snapbuf(umm, ext_df->ped_emerg_buf, VOS_SNAPBUF_EMERG_SIZE);
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1039,169 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+
+		/* Current @head will be reused, move it after @tail, @dbd will be the new head. */
+
+		dbd->dbd_prev                  = head->dbd_prev;
+		head->dbd_prev                 = cont_df->cd_dtx_committed_tail;
+		tail->dbd_next                 = cont_df->cd_dtx_committed_head;
+		cont_df->cd_dtx_committed_head = head->dbd_next;
+		head->dbd_next                 = UMOFF_NULL;
+		cont_df->cd_dtx_committed_tail = dbd_off;
+		dbd                            = head;
```
janekmi: IMHO this bit ought to be a separate function.

src/vos/vos_dtx.c
janekmi: Are you sure this buffer won’t be used by more than one ULT thread at the same time?

janekmi (changes requested):

src/include/daos/mem.h
Nasf-Fan: fixed.

src/vos/vos_dtx.c
Nasf-Fan: Fixed.

src/vos/vos_dtx.c
Nasf-Fan: Done.

src/vos/vos_dtx.c
Nasf-Fan: Fixed.

src/vos/vos_layout.h
Nasf-Fan: Done.

src/vos/vos_layout.h
Nasf-Fan: Done.

src/vos/vos_dtx.c
Nasf-Fan: Reworded.

src/common/mem.c
Nasf-Fan: Done.

src/common/mem.c
Nasf-Fan: Done.

src/common/mem.c
Nasf-Fan: Removed.

src/tests/suite/daos_base_tx.c
Nasf-Fan: As you can lookup_single() return `void`, so even if lookup failure, we cannot directly check that via return value, instead, the content of `rbuf` will not be the expected characters. That is why we compare it with `wbuf`. So if the content in `rbuf` and `wbuf` are the same, then means lookup succeed and means former batched commit succeed; otherwise if former batched commit failed, then related lookup will fail as to content in `rbuf` and `wbuf` will be different.

src/dtx/dtx_srv.c
Nasf-Fan: Fixed.

src/vos/vos_dtx.c
Nasf-Fan: The subsequent logic guarantees that error case will not be accepted:
`if (ext_df != NULL && !UMOFF_IS_NULL(ext_df->ped_emerg_buf) && behavior == TX_FAILURE_RETURN)`

src/vos/vos_dtx.c
Nasf-Fan: Done.

src/vos/vos_dtx.c
Nasf-Fan: Split.

src/vos/vos_dtx.c
Nasf-Fan: There will be no CPU yield during PMEM based VOS transaction. So nobody can share `ext_df->ped_emerg_buf` until current PMEM TX committed or aborted. It is harmless if the buffer is used by others after current PMEM TX committed, right?

src/dtx/dtx_srv.c
Nasf-Fan: Fixed.

src/vos/vos_layout.h
Nasf-Fan: That seems unnecessary. If it is new created pool, then the `ped_emerg_buf` will be assigned. Otherwise, if it is an existing old pool, `ped_emerg_buf` will be `UMEM_NULL` since the whole area `ped_paddings[VOS_POOL_EXT_DF_PADDING_SIZE]` was initialized when creating such pool. So even if we do not bump the version, it will not cause any confusing.

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-18039/7/testReport/

Nasf-Fan:
> * continuous-integration/jenkins/pr-head

test_dfuse_daos_build_wb failed for DAOS-18813, not related with the patch.

src/vos/vos_dtx.c
```diff
@@ -420,6 +420,32 @@ static btr_ops_t dtx_committed_btr_ops = {
...
 
+static inline int
+vos_dtx_add_ptr(struct vos_pool *pool, void *ptr, size_t size)
+{
+	struct umem_instance *umm = &pool->vp_umm;
+	int                   rc;
+
+	rc = umem_tx_add_ptr(umm, ptr, size);
+#ifdef DAOS_PMEM_BUILD
+	if (unlikely(rc == -DER_NOSPACE)) {
```
knard38, line 431: From my understanding, `behavior == TX_FAILURE_RETURN` is redundant: in `TX_FAILURE_ABORT` mode, PMDK aborts the TX internally before returning `-DER_NOSPACE` making any recovery here a silent no-op. If I am correct, an `assert()` should better documents the invariant and catches future callers that forget to set `TX_FAILURE_RETURN`.

```c
   static inline int
   vos_dtx_add_ptr(struct vos_pool *pool, void *ptr, size_t size)
   {
          struct umem_instance *umm = &pool->vp_umm;
          int                   rc;

          rc = umem_tx_add_ptr(umm, ptr, size);
   #ifdef DAOS_PMEM_BUILD
          if (unlikely(rc == -DER_NOSPACE)) {
                  struct vos_pool_ext_df *ext_df = umem_off2ptr(umm, pool->vp_pool_df->pd_ext);

                  /* In TX_FAILURE_ABORT mode PMDK already aborted; -DER_NOSPACE
                   * is only actionable when TX_FAILURE_RETURN was set. */
                  D_ASSERT(umem_tx_get_failure_behavior(umm) == TX_FAILURE_RETURN);

                  /* Use pre-allocated emergency undo-log buffer and retry.
                   * ext_df/ped_emerg_buf are NULL on pre-patch pools. */
                  if (ext_df != NULL && !UMOFF_IS_NULL(ext_df->ped_emerg_buf)) {
                          rc = umem_tx_set_snapbuf(umm, ext_df->ped_emerg_buf,
                                                   VOS_SNAPBUF_EMERG_SIZE);
                          if (rc == 0)
                                  rc = umem_tx_add_ptr(umm, ptr, size);
                          else
                                  rc = -DER_NOSPACE;
                  }
          }
   #endif
          return rc;
   }
```

src/vos/vos_dtx.c
```diff
@@ -2691,6 +2852,8 @@ vos_dtx_abort_internal(struct vos_container *cont, struct vos_dtx_act_ent *dae,
 	if (rc != 0)
 		goto out;
 
+	umem_tx_set_failure_behavior(umm, TX_FAILURE_RETURN);
```
knard38, line 2855: From my understanding, the `umem_tx_begin() + umem_tx_set_failure_behavior(TX_FAILURE_RETURN)` pair appears in all 4 DTX entry points that use `vos_dtx_add_ptr()`. Combining them into a helper should eliminates the risk of forgetting the second call (which would silently disable the emergency buffer fallback under space pressure).

If I am correct, introducing this new helper could make some sense:
```c
   /* Begin a DTX PMEM transaction with TX_FAILURE_RETURN mode, required
    * for the emergency undo-log buffer fallback in vos_dtx_add_ptr(). */
   static inline int
   vos_dtx_tx_begin(struct umem_instance *umm)
   {
          int rc;

          rc = umem_tx_begin(umm, NULL);
          if (rc == 0)
                  umem_tx_set_failure_behavior(umm, TX_FAILURE_RETURN);
          return rc;
   }
```

Then, the 4 following calls would become:
```c
   /* vos_dtx_commit() */
   rc = vos_dtx_tx_begin(vos_cont2umm(cont));

   /* vos_dtx_abort_internal() */
   rc = vos_dtx_tx_begin(umm);

   /* vos_dtx_set_flags() */
   rc = vos_dtx_tx_begin(umm);

   /* dtx_blob_aggregate() */
   rc = vos_dtx_tx_begin(umm);
```

More over, this also makes the `D_ASSERT` in `vos_dtx_add_ptr()` a stronger guarantee: any caller not using `vos_dtx_tx_begin()` would be immediately caught.

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+	struct vos_dtx_blob_df    *tail;
+	struct vos_dtx_cmt_ent_df *dce_df;
+	daos_epoch_t               epoch;
+	umem_off_t                 dbd_off = UMOFF_NULL;
+	d_iov_t                    kiov;
+	uint32_t                   count;
+	int                        rc;
+	int                        i;
+
+	/*
```
knard38, line 1059: Fully agree that a full scan of all containers to find the globally oldest blob would be too costly. Do you think a bounded scan with an early exit could be acceptable? Something like the following — capped at MAX_CONTAINERS_TO_SCAN — would keep the cost effectively O(1) while still covering the common case where a viable victim exists nearby:
```C
   #define MAX_CONTAINERS_TO_SCAN 4

   /* Fallback for vos_dtx_reuse_cmt_blob() when the current container
    * has at most one committed blob. Scans at most MAX_CONTAINERS_TO_SCAN
    * peer containers on the same pool (all xstream-local, no locking needed).
    * Best-effort: may miss victims beyond the scan limit.
    */
   int vos_dtx_steal_cmt_blob(struct vos_container *cont) {
       struct vos_pool      *pool = cont->vc_pool;
       struct vos_container *victim;
       int                   scanned = 0;

       pool_for_each_container(pool, victim) {
           if (victim == cont)
               continue;
           if (scanned++ >= MAX_CONTAINERS_TO_SCAN)
               break;
           if (victim->cd_dtx_committed_head != victim->cd_dtx_committed_tail)
               return vos_dtx_reuse_cmt_blob(victim);
       }
       return -DER_NOSPACE;
   }
```

One drawback is that iteration always starts from the list head, so the same containers are checked first on every call (potential fairness issue). A round-robin starting point would improve this, though I am not sure it is worth the added complexity.

src/dtx/dtx_srv.c
```diff
@@ -196,19 +196,13 @@ dtx_handler(crt_rpc_t *rpc)
 		if (unlikely(din->di_epoch == 1))
 			D_GOTO(out, rc = -DER_IO);
 
-		while (i < din->di_dtx_array.ca_count) {
-			if (i + count > din->di_dtx_array.ca_count)
-				count = din->di_dtx_array.ca_count - i;
+		/* The count of DTX entries will not exceed DTX_THRESHOLD_COUNT. */
+		rc1 = dtx_commit_large(cont->sc_hdl, (struct dtx_id *)din->di_dtx_array.ca_arrays,
+				       (int)din->di_dtx_array.ca_count, false, NULL);
+		if (rc1 < 0)
```
knard38, line 202: From my understanding, when `dtx_commit_large()` returns an error on the participant side, some DTXs in the batch may already have been locally committed. However, this should not be harmful as the leader will retry the full batch.
If this is correct, it could be helpful to log a warning in this case to make it easier to debug space pressure issues.

```C
   if (rc1 < 0) {
       /*
        * Some DTXs in this batch may already be committed before the error.
        * The leader will retry the full batch; re-committing is safe (idempotent).
        */
       D_WARN("Partial DTX commit on participant, count %d: "DF_RC"\n",
              (int)din->di_dtx_array.ca_count, DP_RC(rc1));
       D_GOTO(out, rc = rc1);
   }
```

src/dtx/dtx_rpc.c
knard38, line 877: From my side, it took me some time to understand why we use `PARTIAL_COMMITTED + retry` instead of just reverting the partial commit.
If I understand correctly, the remote participants commit before the leader, so by the time the leader's local commit partially fails, the data is already visible on remote nodes. Thus, aborting an already-committed DTX there would corrupt it?

For contributors less familiar with the DTX commit protocol, like me, an expanded comment could be helpful:
```C
    /*
     * Remote participants committed before the leader (see ordering comment above).
     * If the leader's local commit partially fails (e.g., -DER_NOSPACE), reverting
     * remote participants is not possible: aborting an already-committed DTX would
     * corrupt data visible on those nodes. Instead, mark all entries in this batch
     * as PARTIAL_COMMITTED so the next batched commit retries them. Re-committing
     * an already-committed DTX is always safe.
     */
```

This is just a suggestion to improve readability: no issue to keep as-is if you think it is not needed.

src/vos/vos_dtx.c
Nasf-Fan: Sorry, I am not sure I catch your idea. Here is the call trace:

vos_dtx_add_ptr => pmemobj_tx_add_range_direct => pmemobj_tx_add_common => ...

```
static inline int
obj_tx_fail_err(int errnum, uint64_t flags)
{
        if ((flags & POBJ_FLAG_TX_NO_ABORT) == 0)
                obj_tx_abort(errnum, 0);
        errno = errnum;
        return errnum;
}

static int
pmemobj_tx_add_common(struct tx *tx, struct tx_range_def *args)
{   
...

        if (ret != 0) {
                ERR_WO_ERRNO("out of memory");
                return obj_tx_fail_err(ENOMEM, args->flags);
        }

        return 0;
}
```

So even if `POBJ_FLAG_TX_NO_ABORT` is set, it still will not prevent `ENOMEM` (that is converted as `-DER_NOSPACE` by DAOS) being returned to the caller in VOS, right? From the caller perspective, it only can know whether backend TX has been aborted via check the `behavior` instead of the return value, or I missed anything?

Please note that, obj_tx_abort() will `longjum()` only when the caller specify `tx_data::env` when start the PMDK TX, but for DAOS usage, it is empty.

src/vos/vos_dtx.c
Nasf-Fan: I will refresh the patch for that.

src/vos/vos_dtx.c
Nasf-Fan: OK, I will optimize the algorithm for the victim a bit, maybe not the same as your method.

src/dtx/dtx_srv.c
Nasf-Fan: Your understand is correct. But such message may fill the log file if some targets down or network trouble last for long time. That maybe complained.

src/dtx/dtx_rpc.c
Nasf-Fan: Currently, there is no way to revert partial commit, since we do not know whether someone has already read related partial committed data on related targets. I will add some comment to make things more clear.

src/common/mem.c
```diff
@@ -997,6 +997,49 @@ pmem_defer_free(struct umem_instance *umm, umem_off_t off, void *act)
...
+	behavior = pmemobj_tx_get_failure_behavior();
+
+	switch (behavior) {
+	case POBJ_TX_FAILURE_ABORT:
+		return TX_FAILURE_ABORT;
+	case POBJ_TX_FAILURE_RETURN:
+		return TX_FAILURE_RETURN;
+	default:
+		D_ASSERTF(0, "Unknown TX failure behavior %d\n", behavior);
+		return -DER_INVAL;
```
janekmi, line 1029: No return after unconditional assert.

```suggestion
```

src/vos/vos_layout.h
janekmi: Have you considered adding an ability to upgrade already existing pools or is it unlikely a feature like this would be actually helpful?

src/vos/vos_dtx.c
janekmi: > It is harmless if the buffer is used by others after current PMEM TX committed, right?

Yes.

src/dtx/dtx_srv.c
```diff
@@ -196,8 +196,9 @@ dtx_handler(crt_rpc_t *rpc)
 		if (unlikely(din->di_epoch == 1))
 			D_GOTO(out, rc = -DER_IO);
 
-		rc1 = dtx_commit_large(cont->sc_hdl, (struct dtx_id *)(din->di_dtx_array.ca_arrays),
-				       din->di_dtx_array.ca_count, false, NULL);
+		/* The count of DTX entries will not exceed DTX_THRESHOLD_COUNT. */
```
janekmi, line 199:
```suggestion
		D_ASSERT(din->di_dtx_array.ca_count <= DTX_THRESHOLD_COUNT);
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+	 * Space is almost exhausted. Under such case, we must reclaim space to make current
+	 * DTX commit to be proceed; otherwise, uncommitted DTX may block VOS aggregation as
+	 * to prevent further space release. The most direct approach is to reclaim some old
+	 * blob from current container's committed DTX table. It may be unfair because other
+	 * containers could hold older committed DTX entries. However, it maybe not worth to
+	 * scan all pools/containers on the target to find the globally oldest committed DTX
+	 * blob under space pressure. For now, we select current container as the victim.
+	 *
+	 * This can be optimized later. DAOS-18690.
+	 */
```
janekmi, line 1069: Nitpick. This comment seems to apply to the whole `vos_dtx_reuse_cmt_blob()` function now. Considering it length it would be good to move it there.

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+
+	if (unlikely(cont_df->cd_dtx_committed_head == cont_df->cd_dtx_committed_tail)) {
+		D_ERROR("No space when extending commit DTX table.\n");
```
janekmi, line 1072:
```suggestion
		D_ERROR("Cannot reuse a blob from the committed DTX table. There are no blobs in this table.\n");
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+		d_iov_set(&kiov, &dce_df->dce_xid, sizeof(dce_df->dce_xid));
+		rc = dbtree_delete(cont->vc_dtx_committed_hdl, BTR_PROBE_EQ, &kiov, NULL);
+		if (rc == 0)
+			count++;
+		if (rc == -DER_NONEXIST)
+			rc = 0;
+		if (unlikely(rc != 0)) {
+			D_ERROR("Failed to remove DTX entry " DF_DTI
+				" for urgent DTX aggregation " UMOFF_PF ": " DF_RC "\n",
+				DP_DTI(&dce_df->dce_xid), UMOFF_P(dbd_off), DP_RC(rc));
```
janekmi, line 1091: 1. The "for urgent DTX aggregation" bit seems to describe the wider context of the this procedure. It is a risky thing to write since this function may be used in other places soon and then this message will become outdated. The context of this message has to be read from other messages.
2. It is not written explicitly what is the relation between `DF_DTI` and `UMOFF_PF`. I tried to express it in my suggestion. Feel free to adjust as necessary.

```suggestion
			D_ERROR("Failed to remove DTX entry " DF_DTI
				" belonging to the " UMOFF_PF " blob: " DF_RC "\n",
				DP_DTI(&dce_df->dce_xid), UMOFF_P(dbd_off), DP_RC(rc));
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+	if (unlikely(cont_df->cd_dtx_committed_head == cont_df->cd_dtx_committed_tail)) {
+		D_ERROR("No space when extending commit DTX table.\n");
+		return -DER_NOSPACE;
+	}
+
+	dbd_off = cont_df->cd_dtx_committed_head;
+	head    = umem_off2ptr(umm, dbd_off);
+	D_ASSERT(head != NULL);
+
+	for (i = 0, epoch = 0, count = 0; i < head->dbd_count; i++) {
```
janekmi, line 1080: Not `epoch` nor `count` are actually counters controlling this loop. You can initialize them at the beginning of this function.


```suggestion
	for (i = 0; i < head->dbd_count; i++) {
```

src/dtx/dtx_srv.c
knard38: Make sense

src/dtx/dtx_rpc.c
knard38: Thanks, really think it will help new comer such as me.

src/vos/vos_dtx.c
knard38: Nothing mandatory, just wanted to have your opinion on this for being sure that I have properly understood.

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+		}
+
+		if (epoch < dce_df->dce_epoch)
+			epoch = dce_df->dce_epoch;
+	}
+
+	tail = umem_off2ptr(umm, cont_df->cd_dtx_committed_tail);
+	D_ASSERT(tail != NULL);
+
+	dbd = umem_off2ptr(umm, head->dbd_next);
```
janekmi, line 1102: `dbd` and `dbd_off` are not related despite sharing the `dbd` part of their name which makes following this function quite confusing to me. Can you please make it straight?

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+
+	rc = vos_dtx_add_ptr(cont->vc_pool, head, DTX_CMT_BLOB_SIZE);
+	if (rc != 0)
+		goto out;
+
+	/* dbd_next is next to dbd_prev */
+	rc = vos_dtx_add_ptr(cont->vc_pool, &head->dbd_prev,
+			     sizeof(head->dbd_prev) + sizeof(head->dbd_next));
+	if (rc != 0)
+		goto out;
```
janekmi, line 1113: It looks like these two overlaps and the first one seems a little bit excesive. Or am I missing something?

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+	head->dbd_next                 = UMOFF_NULL;
+	cont_df->cd_dtx_committed_tail = dbd_off;
+
+	dbd            = head;
+	dbd->dbd_count = 0;
+	dbd->dbd_index = 0;
+
+	if (count > 0) {
+		D_ASSERTF(cont->vc_dtx_committed_count >= count,
+			  "Unexpected committed DTX entries count for " DF_UUID ": %u/%u\n",
```
janekmi, line 1145:
```suggestion
			  "Unexpected committed DTX entries count for " DF_UUID ": %"PRIu32"/%"PRIu32"\n",
```

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+			  DP_UUID(cont->vc_id), cont->vc_dtx_committed_count, count);
+
+		cont->vc_dtx_committed_count -= count;
+		cont->vc_pool->vp_dtx_committed_count -= count;
```
janekmi, line 1149: I think the counter on the pool level could use the same assert as you wrote for the container level counter.

src/vos/vos_dtx.c
```diff
@@ -1018,6 +1040,179 @@ vos_dtx_extend_act_table(struct vos_container *cont)
...
+		goto out;
+
+	/* Current @head will be reused, move it after @tail, @dbd will be the new head. */
+
+	dbd->dbd_prev                  = head->dbd_prev;
+	head->dbd_prev                 = cont_df->cd_dtx_committed_tail;
+	tail->dbd_next                 = cont_df->cd_dtx_committed_head;
+	cont_df->cd_dtx_committed_head = head->dbd_next;
+	head->dbd_next                 = UMOFF_NULL;
+	cont_df->cd_dtx_committed_tail = dbd_off;
```
janekmi, line 1137: IMHO how it stand right now is rather hard to follow. Can we break it into two steps like this:

```suggestion
	/* Move current @head after @tail. */
	head->dbd_next = UMOFF_NULL;
	head->dbd_prev = cont_df->cd_dtx_committed_tail;
	tail->dbd_next = cont_df->cd_dtx_committed_head;
	cont_df->cd_dtx_committed_tail = cont_df->cd_dtx_committed_head;
	
	/* Make @dbd the new head. */
	dbd->dbd_prev = UMOFF_NULL;
	cont_df->cd_dtx_committed_head = dbd_off; /* XXX assuming dbd and dbd_off are related */
```

janekmi (changes requested):

src/vos/vos_dtx.c
knard38: Thanks for the clarification!

I have to admit I didn't look deep enough into the code — I wasn't aware of the PMDK `longjmp` mechanism and had naively assumed that in `TX_FAILURE_ABORT` mode, PMDK was calling `abort()`, stopping the whole program. I therefore thought `-DER_NOSPACE` could never reach `vos_dtx_add_ptr` in that mode, making the behavior   check look redundant.

Thanks for your explanation of the PMDK behavior, which makes much more sense now!

src/common/mem.c
Nasf-Fan: Some static analysis tools may warning as "miss return" or similar for such case.

src/dtx/dtx_srv.c
Nasf-Fan: We cannot directly assertion check the value that is from network. If the sender offered invalid value, in spite of by wrong or intentionally, then `dtx_commit_large()` will handle that.

src/vos/vos_dtx.c
Nasf-Fan: I will re-implement vos_dtx_extend_cmt_table() and fix all related issues related with comment, code style, log message.

src/vos/vos_dtx.c
Nasf-Fan: Right, I will fix it.

src/vos/vos_dtx.c
Nasf-Fan: `pool->vp_dtx_committed_count` will be always not less than `cont->vc_dtx_committed_count`, so such check is redundant.

Nasf-Fan:
> Have you considered adding an ability to upgrade already existing pools or is it unlikely a feature like this would be actually helpful?

That is good point. We prepare to land a small patch of reserving buffer for space emergency to 2.8, then we do not need to consider handing existing pool without such extension when upgrade from 2.8 to 2.8.x/master. As for the pool from 2.6, there are too much pool layout difference, such as 2.6 pool does not has the whole
`vos_pool_ext_df`, as to we do not have conclusion about how to upgrade yet.

Nasf-Fan:
> > Have you considered adding an ability to upgrade already existing pools or is it unlikely a feature like this would be actually helpful?
> 
> That is good point. We prepare to land a small patch of reserving buffer for space emergency to 2.8, then we do not need to consider handing existing pool without such extension when upgrade from 2.8 to 2.8.x/master. As for the pool from 2.6, there are too much pool layout difference, such as 2.6 pool does not has the whole `vos_pool_ext_df`, as to we do not have conclusion about how to upgrade yet.

So this patch will be split into two parts, the small part is https://github.com/daos-stack/daos/pull/18137

Nasf-Fan:
> > > Have you considered adding an ability to upgrade already existing pools or is it unlikely a feature like this would be actually helpful?
> > 
> > 
> > That is good point. We prepare to land a small patch of reserving buffer for space emergency to 2.8, then we do not need to consider handing existing pool without such extension when upgrade from 2.8 to 2.8.x/master. As for the pool from 2.6, there are too much pool layout difference, such as 2.6 pool does not has the whole `vos_pool_ext_df`, as to we do not have conclusion about how to upgrade yet.
> 
> So this patch will be split into two parts, the small part is #18137

The small part has already been landed to master. Another part is https://github.com/daos-stack/daos/pull/18141. This one will be closed.

PR #17979, opened 2026-04-10
DAOS-18086 control: Fix data race in drpc unit test

kjacque:
This issue was only affecting unit tests.

### Steps for the author:

* [x] Commit message follows the [guidelines](https://daosio.atlassian.net/wiki/spaces/DC/pages/11133911069/Commit+Comments).
* [x] Appropriate [Features or Test-tag](https://daosio.atlassian.net/wiki/spaces/DC/pages/10984259629/Test+Tags) pragmas were used.
* [x] Appropriate [Functional Test Stages](https://daosio.atlassian.net/wiki/spaces/DC/pages/12147556353/CI+Functional+Test+Stages) were run.
* [x] At least two positive code reviews including at least one code owner from each category referenced in the PR.
* [x] Testing is complete. If necessary, forced-landing label added and a reason added in a comment.

#### After all prior steps are complete:
* [x] Gatekeeper requested (daos-gatekeeper added as a reviewer).

PR #18060, opened 2026-04-20
DAOS-18086 control: Fix data race in drpc unit test (#17979)

kjacque:
This issue was only affecting unit tests.

### Steps for the author:

### DIFF
--- a/src/control/drpc/drpc_server_test.go
+++ b/src/control/drpc/drpc_server_test.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


mjmac (dismissed):

kjacque:
Merged in a clean cherry-pick for the secondary issue (#18090), which affects the 2.8 branch along with master. I'm awaiting merge approval for this modified fix, will add the label when I get it.

tanabarr (approved):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-18060/3/testReport/

src/control/drpc/drpc_server_test.go
```diff
@@ -283,35 +284,51 @@ func TestServer_RegisterModule(t *testing.T) {
...
 
-	lis := newMockListener()
-	lis.setNumConnsToAccept(3)
-	dss, _ := NewDomainSocketServer(log, "dontcare.sock", testFileMode)
-	dss.listener = lis
+			dss.Listen(ctx)
 
-	dss.Listen(test.Context(t)) // will return when error is sent
+			// Time for session goroutines to settle
+			time.Sleep(500 * time.Microsecond)
```
knard38, line 325: Not sure I fully understand the intent of this sleep. From my understanding, `GetNumSessions()` is protected by the same mutex as `Listen()`, which already fully populates the sessions map before returning.
`closeSession()` should also never be called here: `SetReadOutputBytesToResponse()` makes `mockConn.Read()` return the same valid chunk indefinitely.
Thus, `ProcessIncomingMessage()` should never return an error and never trigger a `closeSession()`.
If I am correct, the sleep therefore appears unnecessary.

One danger of this sleep for me is: if the behavior of `SetReadOutputBytesToResponse()` is changed to stop returning indefinitely, such a short 500us sleep would be too small to reliably prevent a race condition on our CI. This would make failures intermittent and hard to debug.

Is there something I'm missing?

kjacque:
This PR affects unit tests only. The only failure is a known issue with a dfuse functional HW test. Requesting forced landing.

src/control/drpc/mocks_test.go
```diff
@@ -179,7 +180,9 @@ func (l *mockListener) Accept() (net.Conn, error) {
 	if l.acceptCallCount > l.acceptNumConns {
 		return nil, l.acceptErr
 	}
-	return newMockConn(), nil
+	c := newMockConn()
+	c.SetReadOutputBytesToResponse(l.t, &Response{})
```
knard38, line 184: From my understanding, `SetReadOutputBytesToResponse(l.t, &Response{})` sets `mockConn.ReadOutputBytes` to one valid header-only chunk, but `mockConn.ReadOutputError` remains `nil`.  As a result, each session goroutine loops forever: `recvMsg()` always succeeds, `ProcessIncomingMessage()` never errors, and `closeSession()` is never called.
 Thus the 3 session goroutines are never terminated: they spin in a busy loop for the entire test binary run, leaking goroutines.

 If I am correct, this could be fixed with setting `ReadOutputError: io.EOF` on the mock conn so goroutines exit cleanly after one round-trip. Then, a `sync.WaitGroup` signalled when each session goroutine exits would ensure goroutines are properly cleaned up before the test binary finishes, instead of relying on a sleep.

knard38 (commented):
I know that the PR has been already merged, but I could have find some issues which could make some sense to fix in a follow-up PR.

src/control/drpc/drpc_server_test.go
kjacque: > `closeSession()` should also never be called here: `SetReadOutputBytesToResponse()` makes `mockConn.Read()` return the same valid chunk indefinitely. Thus, `ProcessIncomingMessage()` should never return an error and never trigger a `closeSession()`. If I am correct, the sleep therefore appears unnecessary.

Correct, it's unnecessary *if* we assume we never trigger a `closeSession()`. The problem is that we were accidentally doing this, which led to the racy behavior. Even when locking around the variables, although the race detector was no longer triggered, it was possible to get different results for the number of sessions depending on when it is read vs. when the sessions died out.

> One danger of this sleep for me is: if the behavior of `SetReadOutputBytesToResponse()` is changed to stop returning indefinitely, such a short 500us sleep would be too small to reliably prevent a race condition on our CI. This would make failures intermittent and hard to debug.

Fair point, this was the type of problem the PR fixes. But even waiting a small amount of time for things to settle allowed the bug to reproduce much more consistently. I left the sleep in because it serves as evidence that my fix works. I could make it longer but was concerned about allocating excessive time unnecessarily to the tests.

src/control/drpc/mocks_test.go
kjacque: One cool thing about the context created by `test.MustLogContext` is that it cancels when the test exits. Cancelling the context causes the sessions to abort, so they are in fact cleaned up when the test exits. I was trying to think of a better way to synchronize and/or detect the incorrectly failed sessions, but it everything I thought of felt like overengineering for a problem with test code.

knard38: Missed this part with `test.MustLogContext`.  Thanks for the explanation.

src/control/drpc/drpc_server_test.go
knard38: Not sure to fully understand, how it could be accidentally triggered, however it seems that it was quite troublesome.  Thanks for the explanation.

kjacque: Oops, thought I had replied to this. Just to add a little more detail, what was happening:

- Each session was failing and closing immediately, because the test didn't format the dRPC packet correctly anymore after the dRPC code was changed to chunk its messages. Before that change, this test succeeded all the time because the sessions stayed open until the context was canceled.
- The race was between the sessions closing and the test checking for active sessions. The sessions closed very quickly, but the test was usually (but not always) a little faster.

PR #18062, opened 2026-04-21
DAOS-18607 aggregation: reduce EC sluggish warning

gnailzenh:
- disable the sluggish warning when rebuild is running
- in other cases, log it every 10 minutes

### Steps for the author:

### DIFF
--- a/src/container/srv_container.c
+++ b/src/container/srv_container.c
@@ -2000,7 +2000,7 @@ static void
 cont_agg_eph_sync(struct ds_pool *pool, struct cont_svc *svc)
 {
 	bool                warn_sluggish;
-	int                 warn_slug_ranks = 8; /* 8 ranks at most */
+	int                 warn_slug_ranks = 0;
 	d_rank_list_t       fail_ranks = {0};
 	struct cont_ec_agg *ec_agg;
 	struct cont_ec_agg *tmp;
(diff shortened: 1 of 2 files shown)


src/container/srv_container.c
```diff
@@ -2049,11 +2050,14 @@ cont_agg_eph_sync(struct ds_pool *pool, struct cont_svc *svc)
 			}
 
 			if (pool->sp_reclaim != DAOS_RECLAIM_DISABLED &&
-			    cur_ts > ec_agg->ea_server_ephs[i].ee_update_ts + 600)
+			    !ds_pool_is_rebuilding(pool) && (cur_ts - warn_sluggish_ts) >= 600 &&
+			    (cur_ts - ec_agg->ea_server_ephs[i].ee_update_ts) >= 600) {
```
Nasf-Fan: Can we find out the oldest one and only log message for that?

liuxuezhao: right, I thought to only log 3 or 8 such warn msg for each round. (may need not to be oldest).

gnailzenh: warn_sluggish_ts always starts from zero right? It means we still log the first warning

liuxuezhao (dismissed):

wangshilong (commented):
After we landed fix for race between ec agg and rebuild, i think we probably could just skip cont_agg_eph_sync if rebuild is trigggered?

liuxuezhao:
> After we landed fix for race between ec agg and rebuild, i think we probably could just skip cont_agg_eph_sync if rebuild is trigggered?

when ds_pool_is_rebuilding(pool) is true, possibly still with one valid ec agg epoch need to be synced.

wangshilong:
> > After we landed fix for race between ec agg and rebuild, i think we probably could just skip cont_agg_eph_sync if rebuild is trigggered?
> 
> when ds_pool_is_rebuilding(pool) is true, possibly still with one valid ec agg epoch need to be synced.

There is possibly that ec agg epoch is a bit older, not a big deal? Looking at master we actually tried to break loop if ds_pool_is_rebuilding() is called in container iterations loop. I think we'd better unify this logic in the master/2.8 branch.

wangshilong (dismissed):
Good enough for 2.6

liuxuezhao:
> There is possibly that ec agg epoch is a bit older, not a big deal? Looking at master we actually tried to break loop if ds_pool_is_rebuilding() is called in container iterations loop. I think we'd better unify this logic in the master/2.8 branch.

the break looks weird, it is there from 2023. probably just because we forgot to remove it in previous PRs, I'll check some details.

src/container/srv_container.c
gnailzenh: thanks, refreshed the PR

wangshilong (dismissed):

liuxuezhao (approved):

wangshilong (approved):

PR #18064, opened 2026-04-21
DAOS-18849 vos: re-check in vos_cont_destroy()

NiuYawei:
In md-on-ssd mode, vos_cont_destroy() chould yield on flushing WAL or starting local tx, so should re-check if the container is re-opened by rebuild after tx started.

This additional check could be removed once rebuild no longer do on-demand container open/create.

### Steps for the author:


liuxuezhao (approved):

src/vos/vos_container.c
```diff
@@ -654,8 +654,23 @@ vos_cont_destroy(daos_handle_t poh, uuid_t co_uuid)
...
-	d_iov_set(&iov, &key, sizeof(struct d_uuid));
-	rc = dbtree_delete(pool->vp_cont_th, BTR_PROBE_EQ, &iov, NULL);
+	/*
+	 * Temporary workaround: Re-check if the container is re-opened by rebuild
+	 * when this function yield on flushing header or starting local tx.
+	 *
+	 * This additional check should be removed once rebuild no longer do on-demand
+	 * container open/create.
+	 */
+	rc = cont_lookup(&key, &pkey, &cont, pool->vp_sysdb);
```
Nasf-Fan, line 664: The semantics seems wrong. Because when logic comes here, the destroy for related container shards may have already happened and completed on other targets, if we refuse destroy on current target, it may leave orphan shard on current target. Is it possible to add some flags (such as in `vos_cont_df`) before `vos_flush_wal_header()` to refuse subsequent container open during the yield?

NiuYawei: That won't solve the issue you described, see the former reference check in this function, it could return -DER_BUSY as well. (I observed those DER_BUSY error messages in the engine logs), furthermore, changing vos_conf_df requires local tx (that would yield too), I don't think we should complicate things for a temporary workaround. 

The long-term solution is to avoid rebuild do on-demand open/create, I think we've reached a consensus on this?

Nasf-Fan: > That won't solve the issue you described, see the former reference check in this function, it could return -DER_BUSY as well. (I observed those DER_BUSY error messages in the engine logs), furthermore, changing vos_conf_df requires local tx (that would yield too), I don't think we should complicate things for a temporary workaround.
> 
> The long-term solution is to avoid rebuild do on-demand open/create, I think we've reached a consensus on this?

I am not sure I caught your idea, return `-DER_BUSY` will cause current container destroy to be failed on current target, but we have no way to rollback such destroy on other targets, right? So we only can fail subsequent container open.

Nasf-Fan: After discussion with @NiuYawei and @liuxuezhao, we got the agreement that land this workaround firstly, and then try to remove the logic about open/create container on demand from rebuild logic in 2.8.x or later.

Nasf-Fan (approved):

daosbuild3:
Test stage Functional Hardware Medium MD on SSD completed with status UNSTABLE.  https://jenkins-3.daos.hpc.amslabs.hpecorp.net/job/daos-stack/job/daos//view/change-requests/job/PR-18064/3/testReport/

NiuYawei:
`10:34:19 DEBUG|     Errors during downloading metadata for repository 'artifactory.daos.hpc.amslabs.hpecorp.net_artifactory_mellanox-proxy_doca_3.2.1_rhel9_x86_64':
10:34:19 DEBUG|       - Curl error (28): Timeout was reached for https://artifactory.daos.hpc.amslabs.hpecorp.net/artifactory/mellanox-proxy/doca/3.2.1/rhel9/x86_64/repodata/repomd.xml [Operation timed out after 30000 milliseconds with 0 out of 0 bytes received]
10:34:19 DEBUG|     Error: Failed to download metadata for repo 'artifactory.daos.hpc.amslabs.hpecorp.net_artifactory_mellanox-proxy_doca_3.2.1_rhel9_x86_64': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
10:34:19 INFO | Command export PATH=/tmp/daos_dfuse_test_dfuse_daos_build_wb_1/venv/bin:$PATH;export VIRTUAL_ENV=/tmp/daos_dfuse_test_dfuse_daos_build_wb_1/venv;export COVFILE=/tmp/test.cov;export HTTPS_PROXY=http://proxy.houston.hpecorp.net:8080/; sudo -E NO_OPENMPI_DEVEL=1 /tmp/install.sh -y completed in 2:00 (10% of timeout)
10:34:19 INFO | Failure detected - debug information:
10:34:19 ERROR| BuildDaos Test Failed`

@daos-stack/daos-gatekeeper single test failure due to lab connectivity issue, it requires force landing.

PR #18070, opened 2026-04-21
DAOS-18880 cq: specify min python version in lint requirements

daltonbohning:
Specify the minimum python version in the linting requirements file. This will produce a warning instead of a generic error when installing.

Doc-only: true

### Steps for the author:

### DIFF
--- a/utils/cq/requirements.txt
+++ b/utils/cq/requirements.txt
@@ -2,11 +2,11 @@
 pyenchant
 ## flake8 6 removed --diff option which breaks flake precommit hook.
 ## https://github.com/pycqa/flake8/issues/1389 https://github.com/PyCQA/flake8/pull/1720
-flake8==7.3.0
-isort==8.0.1
-pylint==4.0.5
-yamllint==1.38.0
-codespell==2.4.2
+flake8==7.3.0;python_version>="3.9"
+isort==8.0.1;python_version>="3.10"
+pylint==4.0.5;python_version>="3.10"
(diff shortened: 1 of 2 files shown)


utils/cq/requirements.txt
```diff
@@ -2,11 +2,11 @@
...
-codespell==2.4.2
+flake8==7.3.0;python_version>="3.9"
+isort==8.0.1;python_version>="3.10"
+pylint==4.0.5;python_version>="3.10"
+yamllint==1.38.0;python_version>="3.10"
+codespell==2.4.2;python_version>="3.9"
 # Used by ci/jira_query.py which pip installs it standalone.
 jira
-torch>=2.2.0
+torch>=2.2.0;python_version>="3.8"
```
daltonbohning: These are the current minimum versions that are necessary for each package. I opted for this instead of requiring 3.11 for everything so it would be less intrusive for people.

daltonbohning: These could definitely change in the future but that generally won't happen very often

phender: FYI, for torch specifically: https://github.com/daos-stack/daos/pull/18080/changes

daltonbohning: Ah, thanks. I'll wait for that version bump first, then this will by >=3.10

phender (approved):

ryon-jensen (approved):

